PR #14040, opened 2015-09-28
Fix qgis.rb for removed Cask: 'matplotlib'

MatzFan:
Removed cask is available as Python formula


MatzFan:
Ooops - will resubmit with correct syntax..

PR #14063, opened 2015-09-29
adding colwiz

seblemaguer:
I propose to add this colwiz cask :)

### DIFF
--- a/Casks/colwiz-desktop.rb
+++ b/Casks/colwiz-desktop.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'colwiz-desktop' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://www.colwiz.com/update?app=colwiz.installer.v2.mac'
+  name 'Colwiz'
+  homepage 'https://www.colwiz.com'
+  license :closed
+
+  app 'Colwiz Desktop.app'
+end


Casks/colwiz.rb
```diff
@@ -0,0 +1,12 @@
+# coding: utf-8
```
vitorgalvao: Remove this.

Casks/colwiz.rb
```diff
@@ -0,0 +1,12 @@
...
+cask :v1 => 'colwiz' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'colwiz-desktop' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `colwiz-desktop.rb`.

vitorgalvao:

Also, please [squash your commits](http://davidwalsh.name/squash-commits-git).

seblemaguer:
Normally I have done. thanks for the squashing link

vitorgalvao:
Thank you for the contribution. It was merged directly as 1434e3976a5d161074cb02d69e83d86e665dbe1d to keep commit history cleaner. Your contribution is still credited to you.

PR #14070, opened 2015-09-29
Fix sha256 mismatch error for Shapes app

vigosan:


vitorgalvao:
Thank you for the contribution. It was merged directly as 2d17a1b5f3d757f8aa24e00a867feb5606fd1a0a to keep commit history cleaner. Your contribution is still credited to you.

PR #14081, opened 2015-09-29
Fixed app name for Zend Studio 13.0.0

shpoont:


vitorgalvao:
In that case, the top line need to be changed to `cask :v1 => 'zend studio' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates. The file’s name also needs to be changed to `zend studio.rb`.

shpoont:
Done

vitorgalvao:
Thank you for the contribution. It was merged directly as 7338e7d42165da7aca25a7a0a33855ee11ade547 to keep commit history cleaner. Your contribution is still credited to you.

PR #14099, opened 2015-09-30
Add Stem Creator v1.0.0.5

danielbayley:
Stem Creator tool from Native Instruments for the new [Stems audio format](http://stems-music.com).

Signed-off-by: Daniel Bayley daniel.bayley@me.com

### DIFF
--- a/Casks/stemcreator.rb
+++ b/Casks/stemcreator.rb
@@ -1,0 +1,15 @@
+cask :v1 => 'stemcreator' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://stems-music.com/downloads/StemCreator_Installer.dmg'
+  name 'Stem Creator'
+  homepage 'http://stems-music.com/stem-creator-tool'
+  license :gratis
+
+  app 'StemCreator.app'
+
+  artifact 'Documentation', :target => Pathname.new(File.expand_path('~')).join('Library/Application Support/Native Instruments/Stem Creator/Documentation')
+
+  zap :delete => '~/Library/Application Support/Native Instruments/Stem Creator'
+end


danielbayley:
There's one slight problem with this; the `artifact` doesn't seem to work as it should, and I'm not sure why?

Casks/stem-creator.rb
```diff
@@ -0,0 +1,15 @@
...
+  url 'http://stems-music.com/downloads/StemCreator_Installer.dmg'
+  name 'Stem Creator'
+  homepage 'http://stems-music.com/stem-creator-tool'
+  license :gratis
+
+  app 'StemCreator.app'
+
+  artifact 'Documentation', :target => Pathname.new(File.expand_path('~')).join("Library/Application Support/Native Instruments/#{name}/Documentation")
+
+  zap :delete => '~/Library/Application Support/Native Instruments/#{name}'
```
vitorgalvao: Don’t use `#{name}`, write the full path.

Casks/stem-creator.rb
```diff
@@ -0,0 +1,15 @@
+cask :v1 => 'stem-creator' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'stemcreator' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `stemcreator.rb`.

Casks/stem-creator.rb
```diff
@@ -0,0 +1,15 @@
...
+  version :latest
+  sha256 '78f2b082c447bc049c620b2896670c43c4431f3e81953c7c668074f1e76b26da'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/master/Casks/acorn.rb).

vitorgalvao:
It seems like you’re not using it for its purpose, for one. Are you trying to copy a directory? Why?

danielbayley:
Ok, I've made those tweaks.

> It seems like you’re not using it for its purpose, for one. Are you trying to copy a directory? Why?

I was… as the `.dmg` contains a `Documentation` folder (but no `.pkg` to put it anywhere). I own a bunch of Native Instruments products and they all store things in `~/Library/Application Support/Native Instruments/`… so I thought it should go in there. I was going off a previous cask formula (`touchosc-editor`);


```ruby
artifact "touchosc-editor-#{version}-osx/layouts", :target => Pathname.new(File.expand_path('~')).join('Library/Application Support/TouchOSCEditor/layouts')
```

Casks/stem-creator.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  url 'http://stems-music.com/downloads/StemCreator_Installer.dmg'
+  name 'Stem Creator'
+  homepage 'http://stems-music.com/stem-creator-tool'
+  license :gratis
+
+  app 'StemCreator.app'
+
+  artifact 'Documentation', :target => Pathname.new(File.expand_path('~')).join("Library/Application Support/Native Instruments/#{name}/Documentation")
```
vitorgalvao: Don’t use `#{name}`, write the full path.

vitorgalvao:
Changes aren’t made. Did you forget to push?

danielbayley:
Nope; I just hadn't gotten round to re-testing after the local changes. Sorted, and all working now, thanks.

vitorgalvao:
Thank you for the contribution. It was merged directly as 17fad16dadf3d28cce5353f948b3886a6e408a13 to keep commit history cleaner. Your contribution is still credited to you.

PR #14122, opened 2015-10-01
Update HyperDock

keith:
It seems like this app's download URLs have become unversioned, this is
the newest one along with its sha256.


vitorgalvao:
Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`.

keith:
Sounds good, updated.

vitorgalvao:
Thank you for the contribution. It was merged directly as 8f2a94eeec42727b42feb320f81415d2a60542a5 to keep commit history cleaner. Your contribution is still credited to you.

PR #14130, opened 2015-10-01
revert hyperdock to versioned URL and update to latest 1.6.0.1

adidalal:
Found the new versioned URL, so changing the cask back to that.
Updates https://github.com/caskroom/homebrew-cask/pull/14122 and also fixes the stylecop errors.


victorpopkov:
Thank you for the contribution. It was merged directly as de762f8f1cd193bdeb9d1a8d962998abab7c4238 to keep commit history cleaner. Your contribution is still credited to you.

PR #14132, opened 2015-10-01
Added XGuardian v0.1

alexweber:


Casks/xguardian.rb
```diff
@@ -0,0 +1,16 @@
+cask :v1 => 'xguardian' do
+  version '0.1'
+  sha256 '5abb7b4fe1599829f5fbf57c8220060fa75c6d4cd493db63e0f2d34b1096bc07'
+
+  url "https://github.com/openscanner/XGuardian/releases/download/#{version}/XGuardian.app.zip"
```
vitorgalvao: The URL I see on the homepage is `http://xara.openscanner.cc/download`. Where is this one from?

alexweber:
@vitorgalvao It's from the README.md, happy to change it.

alexweber:
@vitorgalvao I'm actually having trouble squashing the commit for the updated homepage, doesn't look like adding a new file plays well with this :-1:

vitorgalvao:
> @vitorgalvao It's from the README.md, happy to change it.

The big requirement is that the download URL can be found on the app’s homepage. If there’s one for github there, we can accept it.

> @vitorgalvao I'm actually having trouble squashing the commit for the updated homepage, doesn't look like adding a new file plays well with this :-1:

Why adding a new file? What’s the issue? If you can do it without squashing, just do it that way and I’ll see if I can do the squashing on this end.

alexweber:
@vitorgalvao phew, got it! :)   I was forgetting to force the push after squashing...

vitorgalvao:
It still looks the same, though.

Casks/xguardian.rb
```diff
@@ -0,0 +1,16 @@
...
+  name 'XGuardian'
+  homepage 'http://xara.openscanner.cc/download'
```
alexweber: @vitorgalvao does it? :)

vitorgalvao: Oh, I see. I should clarify. I was talking about `url`, not `homepage`, `homepage` was fine. Perhaps [this part of the documentation explains it](https://github.com/caskroom/homebrew-cask/blob/7c010b42bd7ac679c1f002a0368b36bb3f495612/doc/CASK_LANGUAGE_REFERENCE.md#when-url-and-homepage-hostnames-differ-add-a-comment)).

vitorgalvao: The problem there is there is no link on the homepage to the download on Github, so we can’t even add the comment, and `url` should be just `http://xara.openscanner.cc/download`and use `version :latest` and `sha256 :no_check`.

*Unless* there is a link to github on the homepage, but I need to know where.

alexweber:
@vitorgalvao Ok, got it, is this ok now? thanks

Casks/xguardian.rb
```diff
@@ -0,0 +1,16 @@
+cask :v1 => 'xguardian' do
+  version ':latest'
```
vitorgalvao: No quotes here and on `:no_check`.

alexweber:
@vitorgalvao k

Casks/xguardian.rb
```diff
@@ -0,0 +1,16 @@
+cask :v1 => 'xguardian' do
+  version :latest
+  sha256 :no_check
+
+  url "http://xara.openscanner.cc/download"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes (it’s what is breaking the build).

alexweber:
@vitorgalvao thanks for your patience & pointers, _should_ be good now :)

vitorgalvao:
Thank you for all the work. Looks great.

PR #14145, opened 2015-10-02
Add MenuMeters v1.9 (yujitach El Capitan port)

ranek:
This is a fork of Alex Harper’s MenuMeters, by Yuji Tachikawa. 
It adds OS X El Capitan support to the utility.
The fork is in development at https://github.com/yujitach/MenuMeters.

### DIFF
--- a/Casks/yujitach-menumeters.rb
+++ b/Casks/yujitach-menumeters.rb
@@ -1,0 +1,14 @@
+cask :v1 => 'yujitach-menumeters' do
+  version '1.9'
+  sha256 'ecfda83c213c5562b9be1b9fd430c78303d5b246d92e3e31822dcbe01acf3107'
+
+  url "http://member.ipmu.jp/yuji.tachikawa/MenuMetersElCapitan/zips/MenuMeters_#{version}.zip"
+  name 'MenuMeters'
+  name 'MenuMeters El Capitan Port'
+  homepage 'http://member.ipmu.jp/yuji.tachikawa/MenuMetersElCapitan/'
+  license :gpl
+
+  prefpane 'MenuMeters.prefPane'
+
+  zap :delete => '~/Library/Preferences/com.ragingmenace.MenuMeters.plist'
+end


vitorgalvao:
@ranek Is this your first contribution? If so, welcome! Also, I think it’s the first time I see a newcomer getting a submission of a fork correct on the first cask.

vitorgalvao:
Thank you for the contribution. It was merged directly as eb9cfe8cd775ed0c00e1998ed828cc005ae3423c to keep commit history cleaner. Your contribution is still credited to you.

ranek:
Thanks, @vitorgalvao. This was actually my second contribution, but your contributor guidelines were easy to follow both times!

As it turns out, the download domain for my previous contribution (https://github.com/caskroom/homebrew-cask/pull/5939) lapsed in registration and the cask went bad. I'm curious, does Homebrew Cask have guidelines on:
* When Wayback Machine links are appropriate as cask URLs?
* Whether a fork of a project, which is no longer available from its original vendor, gets a vendor prefix?
* How to include alternate links (e.g. to a mirror or associated GitHub repo) in cask metadata?

vitorgalvao:
> When Wayback Machine links are appropriate as cask URLs?

Never in the main repo. They can be in [caskroom/unofficial](https://github.com/caskroom/homebrew-unofficial).

> Whether a fork of a project, which is no longer available from its original vendor, gets a vendor prefix?

If there is nowhere where the original still exists officially, then the fork does not need the prefix. Prefixes are meant to prevent confusion — if only one version exists, there is no confusion. That said, if the fork is in a github repo (or similar) and labeled as a fork, then it would still need the prefix. The goal is that it should be clear.

> How to include alternate links (e.g. to a mirror or associated GitHub repo) in cask metadata?

What do you mean by “alternate”? We cannot have more than one `url` per cask. If you mean a case where the homepage and download url are on different domains, [there is a rule for that](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#when-url-and-homepage-hostnames-differ-add-a-comment).

PR #14176, opened 2015-10-03
Fix diskmaker-x version naming scheme

kevinoconnor7:
Looks like they only using the major version number in the actual app name. No promises that this revision will work forever though.

### DIFF
--- a/Casks/diskmaker-x.rb
+++ b/Casks/diskmaker-x.rb
@@ -7,5 +7,5 @@ cask :v1 => 'diskmaker-x' do
   homepage 'http://diskmakerx.com/'
   license :gratis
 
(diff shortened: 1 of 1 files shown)


Casks/diskmaker-x.rb
```diff
@@ -7,5 +7,5 @@
   homepage 'http://diskmakerx.com/'
   license :gratis
 
-  app "DiskMaker X #{version}.app"
+  app "DiskMaker X #{version.split(/\./).first}.app"
```
vitorgalvao: Use `#{version.to_i}` instead.

kevinoconnor7:
Cleaned up and squashed the PR.

PR #14183, opened 2015-10-03
mono-mdk: downgrade El Capitan for 4.0.4.4 hotfix

tko:
http://www.mono-project.com/docs/about-mono/releases/4.0.4.4/

Installs binaries in /usr/local/bin rather than /usr/bin

### DIFF
--- a/Casks/mono-mdk.rb
+++ b/Casks/mono-mdk.rb
@@ -1,8 +1,8 @@
 cask :v1 => 'mono-mdk' do
-  version '4.2.0'
+  version '4.0.4.4'
   sha256 :no_check # required as upstream package is updated in-place
 
-  url "http://download.mono-project.com/archive/#{version}/macos-10-x86/MonoFramework-MDK-#{version}.macos10.xamarin.x86.pkg"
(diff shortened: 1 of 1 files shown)


tko:
On related note "[Mono 4.2.0 is available in the alpha channel.](http://www.mono-project.com/docs/about-mono/releases/4.2.0/)" – I wonder if it would be better to keep everyone on 4.0.4/stable series by default.

vitorgalvao:
>  I wonder if it would be better to keep everyone on 4.0.4/stable series by default.

Yes, it would. It is a mistake that this is here, since [unstable versions belong elsewhere](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#beta-unstable-development-nightly-or-legacy-versions). Care to fix that as well?

tko:
Updated.

victorpopkov:
Thank you for the contribution. It was merged directly as 857db92c300bc3fdf9065a755f80f21000bdea29 to keep commit history cleaner. Your contribution is still credited to you.

PR #14190, opened 2015-10-03
Add LibreOffice - DE v5.0.2

fkoehler:
Adds the german language pack of libre office for the old (4.4) and new (5.0) version of LibreOffice.
The different version is detected by the arch of mac used like in the LibreOffice cask itself.

One must still manually run the installer unfortunately. Could not find a better solution.


fanquake:
Please submit this to https://github.com/caskroom/homebrew-versions

vitorgalvao:
And it requires [`uninstall`](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#uninstall-stanza-details), or it cannot be merged.

PR #14194, opened 2015-10-04
ios7-screensaver: updated homepage and description

specious:
The homepage has moved to http://bodysoulspirit.weebly.com/ios-screensaver-for-osx.html.


fanquake:
Please squash your commits.

fanquake:
Actually, now that I've seen your adding the newer version in #14195, instead of merging this, we'd also remove the older version, and put in the versions repo instead.

specious:
Squashed.

The iOS 8 lockscreen screensaver in #14195 is a fork and is quite different.  The iOS 7 version works on 10.9 Mavericks, while iOS 8 does not.  They are different programs.

vitorgalvao:
This one has been discontinued, so lets just remove it. It is quite confusing otherwise, since the fork is actually now the official one.

PR #14210, opened 2015-10-05
Updated gitbook cask to point to new download

jayshao:
Previous cask pointed at discontinued github repo vs. new download


jayshao:
will fix the style notes and submit separately

PR #14214, opened 2015-10-05
Update iTerm2 to 2.1.4

cabbarb:


cabbarb:
Already requested #14213

PR #14232, opened 2015-10-05
update source for Slack.app

taboularasa:
https://github.com/caskroom/homebrew-cask/issues/14219


victorpopkov:
At the moment, there are no exact preferences of which official download URLs to use. In most cases, however, the one from the vendor homepage is more preferred.

Usually, the download URL taken from appcast is used only if the URL on vendor homepage is not versioned. In that case, the one from appcast is more preferred. However, it's not a mistake to use any of those, especially, if the download hosts in both cases are different. On the other hand if there is an option, the ones that have the same host like the homepage should be preferred.

PR #14244, opened 2015-10-06
Update appcast checksum for casks

linc01n:


victorpopkov:
Thank you for the contribution. However, there is no point in updating those because the appcast checksums are not used anywhere yet. You can update them, of course, for example when you upgrade the cask version, but technically there is no point in that.

If the appcast checksum verification will be implemented, by that time many of those will become outdated, and there will be a need to change most of them again.

fanquake:
We also wouldn't merge a commit that touches so many files at once. Ideally commits should only contain changes to a single file. The case for multiple files is when your updating casks which are either slighty different/interrelated versions of the same software, or they should all be updated at the same time i.e navicat

vitorgalvao:
Thank you anyway, @linc01n, but closing on account of what @victorpopkov and @fanquake already pointed out.

linc01n:
Understood. Thanks everyone.

PR #12344, opened 2015-07-03
Add Infinit Latest

adelinofaria:

Issue #11894, opened 2015-06-14
brew cask install infinit broken


wcarlsen:
Launching Infinit.app after install with brew cask, you run into the problem showed in the attached picture below. I worked around it by opening the disk image downloaded by brew cask in the /Library/Caches/Homebrew folder. But if this work around is need for this application, I see no need for this particular piece of software to be included into brew cask. It defy the purpose of brew cask being easy and smart.

![infinit_error](https://cloud.githubusercontent.com/assets/2537399/8149072/17fab522-12b8-11e5-8f3e-4a9c578bfad5.png)

vitorgalvao:
Agreed. Will remove it.

PR #14254, opened 2015-10-07
Add Infinit.app v0.9.43

digitalknk:


adityadalal924:
Should probably [squash your commits](http://davidwalsh.name/squash-commits-git) to keep things cleaner, fyi.

digitalknk:
Thank you @adityadalal924 , that's exactly what I was looking for!

vitorgalvao:
#11894
#12344

Is that fixed upstream, @digitalknk? If it is, please report back so we can reopen.

digitalknk:
@vitorgalvao The dmg I am using applies the correct infinit.app and it does not have this issue. This dmg is directly from infinit as well. Thanks for bringing this up, I was wondering why infinit never was added to Cask.

vitorgalvao:
Thank you. It still needs to use

```ruby
version :latest
sha256 :no_check
```

though, since without it the cask would break on updates.

digitalknk:
@vitorgalvao done and thank you again :-)

vitorgalvao:
Thank you.

PR #14278, opened 2015-10-08
added openzfs caveat for el capitan

### LINKED ISSUE #14256: OpenZFS needs a caveat about El Capitan installation.
See https://github.com/caskroom/homebrew-cask/issues/14182#issuecomment-146029731.

adidalal:
Fixes #14256 

@vitorgalvao not sure how to make this caveat conditional, not a ruby guy. Feel free to fix and merge as needed

### DIFF
--- a/Casks/openzfs.rb
+++ b/Casks/openzfs.rb
@@ -16,5 +16,10 @@ cask :v1 => 'openzfs' do
     pkg "OpenZFS on OS X #{version.sub(%r{-.*},'')} Yosemite or higher.pkg"
   end
 
+  caveats <<-EOS.undent
+    #{token} requires SIP to be disabled on El Capitan, see
+      https://openzfsonosx.org/wiki/Install for further details.
+  EOS
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
We could do it with [a simple conditional](https://github.com/caskroom/homebrew-cask/blob/0ac5dfaf53d760824b6f581f697ad201dee570bc/Casks/texpad.rb#L3) but I have no quarrel with it appearing in every version. I actually think it might be better in this case, so no one is caught off guard.

ilovezfs:
Please remove any caveats. We are now SIP compliant.

https://github.com/ilovezfs/openzfsonosx-installer/commit/fe930aa0add6cb4c709cbab4d828101ceed13889
https://twitter.com/openzfsonosx/status/656095561490132994
https://openzfsonosx.org/wiki/Downloads
https://openzfsonosx.org/forum/viewtopic.php?f=20&t=2353

PR #14281, opened 2015-10-08
Release conductor 1.2.2

keith:


vitorgalvao:
You need to revert back the quotes. Double quotes are reserved for places where string interpolation is needed.

keith:
Sounds good! Updated

keith:
Thanks!

PR #14290, opened 2015-10-08
QMapShack.app v1.3.1

kribe:


Casks/qmapshack.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'qmapshack' do
+  version '1.3.1'
+  sha256 :no_check
```
vitorgalvao: This needs to be set.

kribe: Do you mean the SHA Check? According to documentation is that optional.

jawshooah: It is used only in cases where an actual shasum is not useful. I.e. with `version :latest`, or if the upstream file is updated in-place without changing the URL.

kribe:
added sha check sum.

vitorgalvao:
Looks good. Can you please squash your commits?

kribe:
commits are squashed now.

vitorgalvao:
Thank you.

PR #14291, opened 2015-10-08
Upgrade Prepros version

tjfwalker:


vitorgalvao:
Thank you for the submission, but we prefer versioned URLs. Also, that comment is standard for us, and should not be changed to free form.

tjfwalker:
@vitorgalvao,

To clarify

- What's the point of `:latest` then as can be found in the documentation https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md ?
- Who's "we" in "...we prefer..."?
- Do you think that that detail about not changing the comment should be mentioned in the documentation (https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md) if it's not already mentioned. If it is already, please point it out to me.

Thanks

vitorgalvao:
> What's the point of `:latest`

It’s for use with apps that are *exclusively* distributed with an unversioned URL. I.e., it is to be used when there is no choice of a versioned URL.

> Who's "we" in "...we prefer..."?

[The project as a whole (including its users)](https://github.com/caskroom/homebrew-cask/issues/1021).

> Do you think that that detail about not changing the comment ought to be mentioned in the documentation.

Most of the comments we have are standard to us. That consistency helps with many things, including with triaging when necessary. The comment is question is [talked about in the reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#when-url-and-homepage-hostnames-differ-add-a-comment), but [there are other](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#target-should-only-be-used-in-select-cases) (check the links in that section).

tjfwalker:
Cool, gotcha. Thanks for the clarifications, @vitorgalvao.

PR #14295, opened 2015-10-09
Add Serial.app v1.2.2

samdoran:


Casks/serial.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'serial' do
+  version '1.2.2'
+  sha256 '683e9cff5962f8871909315ed0f8b56d2e710046c054343866f84ef4dac1e486'
+
+  url 'https://www.decisivetactics.com/lib/click?u=/products/serial/dl/Serial_1.2.2.zip'
```
fanquake: You need to use #{version} in the url

samdoran: Sorry about that. I have corrected it and pushed changes.

PR #14301, opened 2015-10-09
Improve Ruby version check in the formula

jgonera:
Previously if a person was using RVM or Rbenv and set a Ruby < 2.0 as
their global Ruby, cask would use brew to install a newer Ruby even
though a new enough Ruby could be present already.

### DIFF
--- a/bin/brew-cask
+++ b/bin/brew-cask
@@ -88,6 +88,7 @@ ensure_executable () {
     fi
 }
 
+# NOTE: Keep in sync with `Ruby20Requirement` in `/brew-cask.rb`.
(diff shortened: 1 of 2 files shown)


vitorgalvao:
Fine by me. What do you think @jawshooah?

I’d just change the backticks (they’re hard as hell to see, even if we use them in other parts) to `%x(…)`.

jawshooah:
LGTM as well. I don't mind the backticks, but if you find the other syntax more readable, we should go with that.

vitorgalvao:
If you don’t mind them, lets leave them for now.

PR #14304, opened 2015-10-09
Bitscope

scontini76:
BitScope Application Suite: DSO, Meter, Logic, Chart

### DIFF
--- a/Casks/bitscope-chart.rb
+++ b/Casks/bitscope-chart.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'bitscope-chart' do
+  version '1.1.DC01G'
+  sha256 'e7c2900281cf289e6ae458a04c8435024aa526ef2df8d0b17874e71f9c92c3a8'
+
+  url "http://bitscope.com/download/files/bitscope-chart_#{version}.app.tgz"
+  name 'BitScope Chart'
+  homepage 'http://www.bitscope.com'
+  license :gratis
+
+  app 'bitscope-chart.app'
+end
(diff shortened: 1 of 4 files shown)


Casks/bitscope-chart.rb
```diff
@@ -0,0 +1,19 @@
+cask :v1 => 'bitscope-chart' do
+  version '1.1.DC01G'
+  sha256 'e7c2900281cf289e6ae458a04c8435024aa526ef2df8d0b17874e71f9c92c3a8'
+
+  url 'http://bitscope.com/download/files/bitscope-chart_1.1.DC01G.app.tgz'
```
fanquake: You should be using #{version} in the url

scontini76: Yeah, sorry for the mistake, I'm going to fix it, thank you! :)

vitorgalvao:
Please remove the caveats, and these are good to go.

scontini76:
It's not a problem for me to remove the caveats, but I put them because you are invited to read the Terms and Conditions in order to downloads those applications from the BitScope site (http://my.bitscope.com/download/). Maybe I misunderstand the use of caveats, I'm going to remove them as you asked!

vitorgalvao:
Please remove them, yes.

Caveats should be used only for things that pertain to the installation itself. This kind os agreement acceptance is something we’re still thinking about.

scontini76:
Ok, thank you very much for the explanation!

vitorgalvao:

Issue #14298, opened 2015-10-09
Java uninstall another issue.


AlexandrButenko:
So, guys. We have another problem. 2 bad strings was fixed in last commit, but we have another one:

```
'/usr/lib/java/libjdns_sd.jnilib',
```
can't be deleted, coz operation don't permitted.

vitorgalvao:
Can you confirm, @bdhess?

bdhess:
I actually haven't converted to El Capitan yet, but I assume that the line that @AlexandrButenko shared is also verboten per System Integrity Protection.

On a Yosemite system with `java` installed, as well as `java6` and `java7` from `-versions`, `/usr/lib/java` is actually empty.  It's difficult for me to ascertain the reason that this line exists in `java.rb` given that none of the installers seems to produce this file.

bdhess:
In the absence of a rationale for the existence of the removal of this file in the first place, I'd be happy to contribute a patch that restricts this behavior to `<= mavericks` only.

AlexandrButenko:
So, guys, i'am not an expert. But i have problem connected with this file. U can look at my logs.

```
[~]$ brew cask uninstall java --force
==> Running uninstall process for java; your password may be necessary
==> Removing launchctl service com.oracle.java.Helper-Tool
Password:
==> Removing launchctl service com.oracle.java.Java-Updater
==> Quitting application ID com.oracle.java.Java-Updater
==> Quitting application ID net.java.openjdk.cmd
==> Removing files from pkgutil Bill-of-Materials
==> Removing files: ["/Library/Internet Plug-Ins/JavaAppletPlugin.plugin", "/Library/Java/JavaVirtualMachines/jdk1.8.0_60.jdk/Contents", "/Library/PreferencePanes/JavaControlPanel.p
==> rm: /usr/lib/java/libjdns_sd.jnilib: Operation not permitted
Error: Command failed to execute!

==> Failed command:
["/usr/bin/sudo", "-E", "--", "/bin/rm", "-rf", "--", "#<Pathname:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin>", "#<Pathname:/Library/Java/JavaVirtualMachines/jdk1.8.0_60.jdk/Contents>", "#<Pathname:/Library/PreferencePanes/JavaControlPanel.prefPane>", "#<Pathname:/Library/Java/Home>", "#<Pathname:/usr/lib/java/libjdns_sd.jnilib>"]

==> Output of failed command:


[...]

PR #14312, opened 2015-10-10
Fix for #14298

bdhess:


vitorgalvao:
Wouldn’t it make more sense to put both under the same conditional?

bdhess:
Cool-- didn't realize that nested arrays were allowed.

vitorgalvao:
Have you confirmed it still works correctly, though? I don’t have my computer, so I can’t.

bdhess:
Yup, I've confirmed that it works correctly.
Commits squashed.

vitorgalvao:
Thank you for your work.

PR #14313, opened 2015-10-10
Bump stack to 0.1.5.0

wereHamster:


Casks/stack.rb
```diff
@@ -1,14 +1,14 @@
...
 
-  url "https://github.com/commercialhaskell/stack/releases/download/v#{version}/stack-#{version}-x86_64-osx.gz"
+  url "https://github.com/commercialhaskell/stack/releases/download/v#{version}/stack-#{version}-x86_64-osx.tar.gz"
   appcast 'https://github.com/commercialhaskell/stack/releases.atom'
   name 'stack'
   homepage 'https://github.com/commercialhaskell/stack'
   license :bsd
 
-  binary "stack-#{version}", :target => 'stack'
+  binary "stack-#{version}-x86_64-osx/stack", :target => 'stack'
```
vitorgalvao: You can remove `, :target => 'stack'`. No longer needed.

wereHamster:
Fixed.

wereHamster:
Apparently the `set_permissions` path doesn't exist anymore, and on a second box installing stack fails (because chmod can't find the file). But on my laptop it works without problems. Any idea why?

wereHamster:
Is set_permissions in postflight needed? It works without.

jawshooah:
The target of the `set_permissions` call was the binary file. Since the binary file has changed as of this PR, that target needs to change as well.

wereHamster:
I figured that much. But is `set_permissions` even needed? The file in the tarball already has +x permissions.

jawshooah:
If that's the case, then no.

PR #14314, opened 2015-10-10
Add QLPlayground v1.0

jakubpetrik:


vitorgalvao:
If the plugin is clearly called (by yourself, even) `qlplayground`, why are you giving the cask a different name?

A strict [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

Since there is already a cask with this name, this one needs [special naming](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#forks-and-apps-with-conflicting-names). Call it `inloop-qlplayground` please [squash your commits](http://davidwalsh.name/squash-commits-git)?

jakubPetrik:
Sorry, first timer here. Should be correct now.

Casks/inloop-qlplayground.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'inloop-qlplayground' do
+  version '1.0'
+  sha256 '44c25a7da0dc3748b01deb0c01634044ccbc625b4266b4fea0630cbedb773929'
+
+  url 'https://github.com/inloop/qlplayground/releases/download/v1.0/inloop-qlplayground.v1.0.zip'
+  name 'inloop-qlplayground'
+  homepage 'https://github.com/inloop/qlplayground'
+  license :oss
```
vitorgalvao: `:mit`.

Casks/inloop-qlplayground.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'inloop-qlplayground' do
+  version '1.0'
+  sha256 '44c25a7da0dc3748b01deb0c01634044ccbc625b4266b4fea0630cbedb773929'
+
+  url 'https://github.com/inloop/qlplayground/releases/download/v1.0/inloop-qlplayground.v1.0.zip'
```
vitorgalvao: This can also be changed to `url 'https://github.com/inloop/qlplayground/releases/download/v#{version}/inloop-qlplayground.v#{version}.zip"` to facilitate future updates.

jakubPetrik:
Many thanks for the tips. 👍

Casks/inloop-qlplayground.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'inloop-qlplayground' do
+  version '1.0'
+  sha256 '44c25a7da0dc3748b01deb0c01634044ccbc625b4266b4fea0630cbedb773929'
+
+  url 'https://github.com/inloop/qlplayground/releases/download/v#{version}/inloop-qlplayground.v#{version}.zip'
```
vitorgalvao: Need to be double quotes, here, or it won’t work.

jakubPetrik:
And what about the sha256 checksum? It won't be valid with the next version, or not?

jakubPetrik:
I have replaced the `sha256` checksum with `:no_check`

jawshooah:
@jakubPetrik No, please don't remove the checksum. We only use `:no_check` in one of two cases: when the URL is not versioned (in which case we also use `version :latest`), and when the file is updated in-place despite having a versioned URL.

Neither of those appear to be the case here, so we need the checksum.

jakubPetrik:
ok. will revert.

jawshooah:
Thank you for the contribution. It was merged directly as 628cfe49a3dfc5f5e2a3fe8f2b8037b8c704d7a5 to keep commit history cleaner. Your contribution is still credited to you.

PR #14318, opened 2015-10-11
Add url command

kevintraver:
Add url command for showing the homepage


vitorgalvao:
Calling this command `url` when it actually shows `homepage` is confusing enough, but it also isn’t needed (or very useful).

There are multiple commands to get this, already (like `brew cask cat`), but even if you need just this information, you can get it with `brew cask _stanza {{whatever stanza you want to read}}`.

Thank you anyway for the work, though, it is appreciated.

kevintraver:
Awesome, thanks!

PR #14337, opened 2015-10-11
add Daylite v5.0.9 (build 6632)

disenchant:


Casks/daylite.rb
```diff
@@ -0,0 +1,13 @@
+cask :v1 => 'daylite' do
+  version '5.0.9 (6632)'
```
victorpopkov: The build number is not needed here. You should use only 5.0.9 here for the easier use of string interpolation in `url` stanza.

Casks/daylite.rb
```diff
@@ -0,0 +1,13 @@
...
+  sha256 '44f1f0fcf20d10581b7ee412ed83d8afced0b673087bf33e69f47ec0b6098f37'
+
+  url 'https://www.marketcircle.com/daylite/download/daylitedma/'
```
victorpopkov: The redirect URL should be used here which includes the version number and it's better to use the string interpolation here for easier version updates:
```ruby
  url "https://download.marketcircle.com/daylite/daylitedma#{version.delete('.')}.dmg"
```

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 0e377b6415b1d9857c944804b1997a5bc2d46fea. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #14348, opened 2015-10-11
Add enpass.rb

Eitot:
Password manager

### DIFF
--- a/Casks/enpass.rb
+++ b/Casks/enpass.rb
@@ -1,0 +1,23 @@
+cask :v1 => 'enpass' do
+  version '5.0.2'
+  sha256 '62bfc6da69a43800d429013c6426181df085334e58d9142a7824cf917c0ab319'
+
+  # sinew.in is the official download host per the vendor homepage
+  url "http://dl.sinew.in/mac/package/Enpass-#{version}.zip"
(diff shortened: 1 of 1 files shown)


victorpopkov:
You should check your Git configurations (`git config`). At the moment the email used in your commits doesn't point to your GitHub account so your contribution won't be credited to you.

Eitot:
@victorpopkov: I modified it with `git config`, but how can I change or override the commit?

victorpopkov:
Yes, the email address will be exposed. If you don't want to expose it, you could use the `<your_username>@users.noreply.github.com`.

Eitot:
@victorpopkov: That’s indeed what I did. I wonder about my previous commits though. Didn’t realise that Github uses my email address unless I set it to private.

Eitot:
I think I fixed it now, for both pull requests. Past commits don’t seem to have exposed my email address. Thanks for the heads-up.

vitorgalvao:
Thank you, but this a [walled build](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#finding-a-home-for-your-cask).

We cannot accept it here. You may submit it to [caskroom/unofficial](https://github.com/caskroom/homebrew-unofficial), though.

PR #14367, opened 2015-10-12
Add Konica Minolta Bizhub C220/C280/C360 Driver 3.9.0, Update Localization of Adobe CC Applications, Add SaveHollywood Screen Saver

reitermarkus:


reitermarkus:
Can this be merged?

vitorgalvao:
Just saw this PR.

We can’t merge it as it is, no. You should submit a PR for each app, as this type of submissions most of the time has some problem with an app that stops the other from being included. In this case, the problem is with the Adobe apps. Many people prefer their system in one language (many times their native one) and their apps in another (mostly english), so we won’t install a different language automatically, that is what [caskroom/versions](https://github.com/caskroom/homebrew-versions) is for.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 3e3a5c1771691fc21a8c1b4546419704ccc9d907. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

reitermarkus:
Thank you, so should I resubmit the Adobe ones to caskroom/versions as a whole `adobe-photoshop-cc-localized` or only for my language `adobe-photoshop-cc-de`?

PR #14369, opened 2015-10-12
Update Private Eye.app download URL

samdoran:
Previous download URL was grabbing an older version. This is the new download
URL per the homepage.

Add launchctl key to uninstall stanza.

Verified workaround is still neded to unload kext.

If Private Eye is running during uninstall, the uninstall will fail becase the kext unload happens too soon after the quit. Running the uninstall a second time works or quitting the app before running unistall works. I'm not sure of a good way to solve this. Open to suggestions.

### DIFF
--- a/Casks/private-eye.rb
+++ b/Casks/private-eye.rb
@@ -2,12 +2,12 @@ cask :v1 => 'private-eye' do
   version :latest
   sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Merged as fb64eb5c80790456ed5b6f0a7c469eac7f455d86.

PR #14370, opened 2015-10-12
Add Radio Silence.app

samdoran:
Similar to PR #1802, an extra stanza is required so the uninstall does not fail. Both of these apps are from the same developer and using the same kext, hence the similar behavior.

Also, similar to PR #14369, the uninstall task will fail if the app is running. I'm not sure of a good way to fix that


vitorgalvao:
Merged as faf1a76edab0205d3c707fed2380ff53303a62f8.

PR #14383, opened 2015-10-13
Fixed Haskell-Platform 7.10.2-a uninstall script

coreymcdermott:
Fixed Haskell-Platform 7.10.2-a uninstall script

### DIFF
--- a/Casks/haskell-platform.rb
+++ b/Casks/haskell-platform.rb
@@ -9,7 +9,7 @@ cask :v1 => 'haskell-platform' do
 
   pkg "Haskell Platform #{version} 64bit-signed.pkg"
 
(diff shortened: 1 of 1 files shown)


michaelsbradleyjr:
After running `brew cask uninstall haskell-platform` I find that `haskell-platform` is still given in the output of `brew cask list`, despite the uninstall process having completed without error. I have done `cask doctor` and `cask cleanup`, in combination with cycling through an uni/install of the package, to no avail. I'm all up-to-date with respect to `brew update`.

I'm not sure how to fix it, nor whether the problem is related to this pull request, but any suggestions would be appreciated re: how to fix it.

vitorgalvao:
@michaelsbradleyjr https://github.com/caskroom/homebrew-cask/issues/14058.

PR #14386, opened 2015-10-13
change panoply to use the version number, which is now part of the URL.

hollandjg:


fanquake:
If you switch to using a version number, you also need to provide a SHA256

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 1fb9f960099b584947ac3c8fb6757842966cbc6d. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #14390, opened 2015-10-13
Use open application syntax.

knpwrs:
Not sure if this has been suggested before. I tried searching but the results were too generic.


vitorgalvao:
Though the suggestion is appreciated (personally that’s how I’d do it), we’re not exactly suggesting you run that command. Rather, we’re illustrating how it can work, and in that sense using the full path is more powerful (i.e. “look, it’s right there were we said it would be”).

PR #14392, opened 2015-10-13
fontforge: change dependency to flexible x11

### LINKED ISSUE #14389: FontForge hard Cask-XQuartz dep?
Hey Folks,

Was wondering why the FontForge Cask has a `depends_on :cask => 'xquartz'` line instead of the more flexible `depends_on :x11 => true` line?

My understanding is the latter checks if a user already has X11 available and installed, whereas the `depends_on :cask` syntax seems to be trying to force me to install the Cask's XQuartz on top of my already existing XQuartz non-brewed installation.

There wasn't any explanation in https://github.com/caskroom/homebrew-cask/commit/dc122dc64b5af4a9a8e8b5a9be9dcef16664deab, but apologies if I'm missing something previously documented.

Cheers.

DomT4:
Closes #14389.

### DIFF
--- a/Casks/fontforge.rb
+++ b/Casks/fontforge.rb
@@ -10,5 +10,5 @@ cask :v1 => 'fontforge' do
 
   app 'FontForge.app'
 
-  depends_on :cask => 'xquartz'
+  depends_on :x11 => true
 end


vitorgalvao:
Thank you!

DomT4:
Thanks for the quick merge! :)

Issue #1021, opened 2013-09-15
Prefer checksummed versioned or no_checksum latest casks?


phinze:
This issue comes from a long discussion in #1011 between @gregstallings @vitorgalvao @NanoXD and @passcod.

If anybody believes the core issue to be different than the one I've titled; please let me know and I'm happy to edit.

----

Here's the way I put it over there:

> For a given cask, if we have available both latest / no_checksum and versioned / checksummed URLs, which should we prefer?

So to lay out some differences...

### **Preferring latest**

 * :+1: Maintainers don't have to deal with PRs for every cask update.
 * :+1: / :-1: For apps with autoupdate, users don't interact with homebrew-cask for upgrades. (Some users may prefer this usage model; other may prefer homebrew-cask to be a one-stop-shop package manager.)
 * :-1: There's no checksum protecting the download from MITM.

### **Preferring checksummed**

 * :+1: Checksums protect the downloads from tomfoolery and skullduggery :skull:.
 * :+1: / :-1: For upgrades, homebrew-cask manages different versions.
 * :-1: Maintainers + contributors need to process update PRs more often.

-----

There are a few things I'd like to highlight as we continue this conversation:

### Users and Maintainers are Different

There is an important distinction between *user benefit* and *maintainer benefit*. I believe we should strive for both, but in situations where they are at odds where should always prefer users over maintainers (within reason). Obviously hours of maintainer work for a minor user benefit is a bad idea. But something somewhat inconvenient for maintainers that provides a major benefit to users deserves to be considered.

Mostly I want to make sure we distinguish the two in our conversations about features and stuff, so we can properly talk about the needs of the two groups.

### There was one more thing but now I've forgotten it.

Maybe it will come back to me. :blush:

phinze:
Closing this as I think the core maintainers are all on the same page. @rolandwalker's last update accurately represents the state of the union. Anybody who disagrees is welcome to chime in and we can talk about reopening. :ok_hand:

Yann-R:
@rolandwalker 
> There will always be some cases where 'latest' is the only sensible approach, or the only possible approach. So we will never lose it entirely. If those remaining 'latest' Casks are more difficult to upgrade, that's just a compromise we will have to accept. The perfect is the enemy of the good.

Do you mean that when the download URL of the binary contains no version (e.g. http://company.com/download/latest_version.dmg) we must go to unversioned formula?

[...]

PR #14441, opened 2015-10-15
Note preference for versioned URLs in CONTRIBUTING.md

jawshooah:
This is the only place remaining in the documentation where we express a preference for unversioned URLs. Following #1021 and other conversations, we have reached a consensus that _versioned_ URLs are preferred, and that _unversioned_ URLs should be used only when no versioned URL exists.

[Review on Reviewable](https://reviewable.io/reviews/caskroom/homebrew-cask/14441)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ cask :v1 => 'unity' do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thumbs up from me.

PR #14453, opened 2015-10-16
Update chefdk for El Capitan

LoreleiAurora:


adityadalal924:
This should really be a conditional - they have downloads for multiple OS X versions
(at https://downloads.chef.io/chef-dk/mac/)

Sample cask to model after - [Graphviz](https://github.com/caskroom/homebrew-cask/blob/master/Casks/graphviz.rb)

Also, fyi, for El Capitan, the actual url is actually `https://opscode-omnibus-packages.s3.amazonaws.com/mac_os_x/10.11/x86_64/chefdk-0.9.0-1.dmg`

adityadalal924:
@LoreleiAurora looked again, my mistake, it seems like all the downloads all point to the same `.dmg`

This cask actually doesn't need updating (`shasum` still valid, and so are download urls); however, the url can be changed to have `10.11` if you also want to be "technically" correct.

vitorgalvao:
> This cask actually doesn't need updating (`shasum` still valid, and so are download urls); however, the url can be changed to have `10.11` if you also want to be "technically" correct.

Agreed. If we’re changing anything, might as well do that. @LoreleiAurora mind making the change?

adityadalal924:
@LoreleiAurora Please [squash your commits](http://davidwalsh.name/squash-commits-git) and then your PR is good to go!

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 0e1708b9241196b0e14961da67de60c1da72deaa. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #14455, opened 2015-10-16
Add RemoteViewer

rimenes:
A OS X client for the spice protocol

### DIFF
--- a/Casks/remoteviewer.rb
+++ b/Casks/remoteviewer.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'remoteviewer' do
+  version '0.5.7'
+  sha256 '96ec9eb33422ac8a717238a7f99fded7c50503198d7302028a09f4e536b02849'
+
+  url 'http://people.freedesktop.org/~teuf/spice-gtk-osx/dmg/0.5.7/RemoteViewer-0.5.7-1.dmg'
+  name 'Remote Viewer'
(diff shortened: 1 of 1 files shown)


adityadalal924:
Please use a versioned url, i.e. `url "http://people.freedesktop.org/~teuf/spice-gtk-osx/dmg/#{version}/RemoteViewer-#{version}-1.dmg"`

Also needs a comment about that download url being official

vitorgalvao:
[The comment in question must follow this form](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#when-url-and-homepage-hostnames-differ-add-a-comment).

rimenes:
Thanks. I've made contributions in the past but a lot of things have changed since then.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 9576cebf065f651828e31f817038e2f0961c9d49. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

adityadalal924:
@vitorgalvao I think the fixes you made actually broke the cask - there's no folder with `0.5.7-1`in the download url, it has to be just the version part before the `-` (the actual dmg name is still correct, though)

vitorgalvao:
@adityadalal924 You’re right, hadn’t noticed `#{version}` was used twice. Will correct shortly.

PR #14464, opened 2015-10-16
flash 19.0.0.226 and pepper-flash 19.

andyli:
Updated flash and pepper-flash.

I notice that pepper flash has its sha256 check disabled. I'm not sure if that is good or not. But should we do the same for the flash, flash-player, and flash-player-debugger formulae since they are in the same situation?


vitorgalvao:
No. What we should do is instead set `pepper-flash` with a full version and `sha256`.

andyli:
Done. : )

vitorgalvao:
Thank you.

PR #14467, opened 2015-10-16
Update subler.rb

Eitot:
• Update version
• Add zap stanza

### DIFF
--- a/Casks/subler.rb
+++ b/Casks/subler.rb
@@ -1,14 +1,21 @@
 cask :v1 => 'subler' do
-  version '1.0.3'
-  sha256 '631495e84b6dcda68e589b1c7172fb5ad6f41f76cbd73e3b739f25f741b1a1a6'
+  version '1.0.4'
+  sha256 'f8e95ea92eef150246d3a2f3da0472e4ef53e2781ae2e4b9d5ad29c16f5bd40e'
 
   # bitbucket.org is the official download host per the vendor homepage
   url "https://bitbucket.org/galad87/subler/downloads/Subler-#{version}.zip"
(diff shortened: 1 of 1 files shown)


Casks/subler.rb
```diff
@@ -1,14 +1,22 @@
...
-  sha256 '631495e84b6dcda68e589b1c7172fb5ad6f41f76cbd73e3b739f25f741b1a1a6'
+  version '1.0.4'
+  sha256 'f8e95ea92eef150246d3a2f3da0472e4ef53e2781ae2e4b9d5ad29c16f5bd40e'
 
   # bitbucket.org is the official download host per the vendor homepage
   url "https://bitbucket.org/galad87/subler/downloads/Subler-#{version}.zip"
   appcast 'http://subler.org/appcast/appcast.xml',
-          :sha256 => 'c2997d3e545fb02306552377ac6a3fd0dc25d7d9761465397982f2904aaf7876'
+          :sha256 => 'f8e95ea92eef150246d3a2f3da0472e4ef53e2781ae2e4b9d5ad29c16f5bd40e',
+          :format => :sparkle
```
vitorgalvao: No need. This is the default.

Eitot: What does the separate sha256 line actually do, is it needed? I don’t see it all that often.

vitorgalvao: It’s so we know the appcast changed. However, `appcast` isn’t yet functional, so there’s no need to update it, either way.

fanquake:
I'll merge this now, but in future please use a descriptive commit message, i.e "Subler 1.0.4" rather than the GitHub default.

PR #14473, opened 2015-10-16
Ftdi vcp driver load kext fix

JoeNyland:
The `ftdi-vcp-driver` cask did not load the kext that it installs on installation. This means that the driver will not work until the next reboot. The user was not informed as a caveat that they need to reboot to use the driver after installation.

I figured it would be better to load the kext for the user after installation so that the user can begin using the driver right away.

On uninstall, the kext was being unloaded correctly and `pkgutil` was clearing up it's side of things, but the kext was still left in `/Library/Extensions` so I fixed that too, by deleting it at uninstall.

Added caveats section to warn users that they need to disconnect/reconnect their FTDI cable for it to show up in `/dev` if they had it connected prior to installation.

### DIFF
--- a/Casks/ftdi-vcp-driver.rb
+++ b/Casks/ftdi-vcp-driver.rb
@@ -15,6 +15,30 @@ cask :v1 => 'ftdi-vcp-driver' do
   license :gratis
 
   uninstall :pkgutil => 'com.FTDI.ftdiusbserialdriverinstaller.*',
-            :kext    => 'com.FTDI.driver.FTDIUSBSerialDriver'
+            :kext    => 'com.FTDI.driver.FTDIUSBSerialDriver',
+            :delete  => '/Library/Extensions/FTDIUSBSerialDriver.kext'
+
+  caveats do
+    reboot
+
+    <<-EOC.undent
+      If you don't want to reboot, you can load the driver using the following
+      command:
+
+        sudo /sbin/kextload -b com.FTDI.driver.FTDIUSBSerialDriver
+
+      Once you've rebooted or loaded the driver, you can connect your FTDI
+      based cable to a USB port and it will show up in /dev, usually like this:
+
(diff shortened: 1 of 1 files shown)


jawshooah:
We generally do not load kexts on install. Instead, we have a [mini-DSL](https://github.com/caskroom/homebrew-cask/blob/fdcf14aafe2bfd18e19cb572937fdd027056862e/doc/CASK_LANGUAGE_REFERENCE.md#caveats-mini-dsl) for `caveats` that includes a `reboot` message ([example](https://github.com/caskroom/homebrew-cask/blob/fdcf14aafe2bfd18e19cb572937fdd027056862e/Casks/osxfuse.rb#L20)). Please use this instead.

Also, for the rest of the `caveats`, you should use `<<-EOC.undent` to clean up the formatting.

After both these changes, there should be no `postflight` block, and your `caveats` should look like this:

```ruby
caveats do
  reboot

  <<-EOC.undent
    Once the driver is installed, you can connect your FTDI based cable to a
    USB port and it will show up in /dev, usually like this:

      /dev/tty.usbserial-XXXXXXXX

    where XXXXXXXX is a random ID, based on the serial number of your FTDI
    based cable.

    NOTE: If your FTDI based cable was already connected before you installed
    the driver, you'll need to unplug the cable from the USB port and
    reconnect it for it to show up in /dev.
  EOC
end
```

MasterRoot24:
@jawshooah Thanks for the pointers.

I've made the requested changes, squashed and pushed my fork.

I've added a note in the caveats to say that reboot can be avoided by manually loading the kext (see [here](https://github.com/MasterRoot24/homebrew-cask/blob/683d3399b3ed517d661bb2e24420e9f67a68c782/Casks/ftdi-vcp-driver.rb#L25:L28)). I trust this is acceptable.

PR #14475, opened 2015-10-16
Add ArgoUML.app v0.34

williamboman:
Depends on Java SE 6, not sure how to set up that dependency.

### DIFF
--- a/Casks/argouml.rb
+++ b/Casks/argouml.rb
@@ -1,0 +1,17 @@
+cask :v1 => 'argouml' do
+  version '0.34'
+  sha256 '42a2e7659c4c4b4fcae4e093f65d83b13ff363c9594920c7c044601fcbb0cf24'
+
+  url "http://argouml-downloads.tigris.org/nonav/argouml-#{version}/ArgoUML-#{version}.app.tar.gz"
+  name 'ArgoUML'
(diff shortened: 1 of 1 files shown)


adityadalal924:
You can use `depends_on :java`, but it's [just a stub right now](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#all-depends_on-keys)

williamboman:
@adityadalal924 It's Java 6 though. Would `depends_on :java6` work?

adityadalal924:
I'm pretty sure any line you write doesn't have an impact, at this point (there's no backend code to enforce it) - probably best to just not add anything

williamboman:
Isn't there a specification for it though?

vitorgalvao:
> Isn't there a specification for it though?

Yes, sort of. Not officially. Until we get java dependencies working (likely not soon), you should add a caveat:

````ruby
caveats <<-EOS.undent
  #{token} requires Java 6 like any other IntelliJ-based IDE.
  You can install it with

    brew cask install caskroom/homebrew-versions/java6
EOS
```

williamboman:
@vitorgalvao Alright, updated.

PR #14480, opened 2015-10-16
Add PhotoScanPro.app v1.1.6

Gerg:
Note that this is a different application than the existing 'photoscan'
cask


Casks/photoscan-pro.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'photoscan-pro' do
```
vitorgalvao: Name must be `photoscanpro`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) has to be followed).

Gerg:
Updated to conform to token reference.

PR #14485, opened 2015-10-17
Add LogicSniffer.app v0.9.7.2

xyb:


vitorgalvao:
[Already present](https://github.com/caskroom/homebrew-cask/blob/master/Casks/logicsniffer.rb). The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed exactly to prevent this types of duplicates.

PR #14491, opened 2015-10-17
Add WeChat.app v6.0.0

wayou:


fanquake:
Something has gone wrong here. You've added the cask, then deleted it? Could you please fix that, and squash everything into a single commit.

fanquake:
Actually, we can close this, please read [here](https://github.com/caskroom/homebrew-cask/issues/10897#issuecomment-97450434) for an explanation as to why we don't have WeChat in homebrew-cask.

PR #14498, opened 2015-10-18
Update istat-menus sha256 for v5.11

adambware:
Updating to correct sha256 checksum for istat-menus v5.11


cjs:
:+1: I just ran into this

PR #14501, opened 2015-10-18
Update ti-connect-ce to 5.1.0.68.

SushiDude:


adityadalal924:
That's a "bad" download URL. I can get it from `http://edudownload.ti.com/downloads/files/cdn/ti-connect/5.1/TIConnectCE-5.1.0.68.dmg`

so, the actual stanza needs some regex magic, but it should be something like
```
url "http://edudownload.ti.com/downloads/files/cdn/ti-connect/#{version__regexhere__}/TIConnectCE-#{version}.dmg"

vitorgalvao:
@adityadalal924 is correct. The submitted `url` looks like it’d need changing at every version. Small tip, no need for a regex, here, converting it to a float works.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 2c7a50027ee06f001f3eb569cb91a12c9fe5e47d. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #14525, opened 2015-10-19
change JOSM to fetch latest stable

danbjoseph:
The latest stable version of the JOSM app is available at: https://josm.openstreetmap.de/download/macosx/josm-macosx.zip

Is it better to point there instead of manually updating the cask each time?


vitorgalvao:
[We prefer them versioned](https://github.com/caskroom/homebrew-cask/issues/1021).

danbjoseph:
:+1:

PR #14528, opened 2015-10-19
Fix openzfs-1.4.5 installation

aktau:
Bug introduced by PR #14526 (commit c54cf295f2). Tested locally on El
Capitan.


adityadalal924:
Whoops, missed fixing the conditional.
Thanks!

PR #14533, opened 2015-10-20
Added fldigi v3.23.04

cedwardsmedia:


Casks/fldigi.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'fldigi' do
+  version '3.23.04'
+  sha256 '162bff2fa2abd62230286eedfa627e4e3199a4c8c68a26094dd6075925026e69'
+
+  url "http://downloads.sourceforge.net/project/fldigi/fldigi/fldigi-#{version}_i386.dmg"
+  name 'fldigi'
+  homepage 'http://sourceforge.net/projects/fldigi/files/fldigi/'
+  license :gpl
+
+  app 'fldigi-3.23.04.app'
```
fanquake: You should be using #{version} here also

Casks/fldigi.rb
```diff
@@ -0,0 +1,12 @@
...
+  app 'flarq-4.3.5.app'
```
vitorgalvao: What is this other app, and why is it included here as well?

cedwardsmedia: Fldigi allows an ordinary computer's sound card to be used as a simple two-way data modem.
Flarq is a file transfer application developed by the same person that is packaged along with fldigi. Fldigi is virtually useless without something to transmit and flarq relies on fldigi in order to transmit.

Casks/fldigi.rb
cedwardsmedia: Good point. Thanks for catching it.

cedwardsmedia:
I've updated the formula to use #{version} in the app stanza as well.

Casks/fldigi.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'fldigi' do
+  version '3.23.04'
+  sha256 '162bff2fa2abd62230286eedfa627e4e3199a4c8c68a26094dd6075925026e69'
+
+  url "http://downloads.sourceforge.net/project/fldigi/fldigi/fldigi-#{version}_i386.dmg"
+  name 'fldigi'
+  homepage 'http://sourceforge.net/projects/fldigi/files/fldigi/'
+  license :gpl
+
+  app 'fldigi-#{version}.app'
```
vitorgalvao: It needs to use double quotes here (`app "fldigi-#{version}.app"`, or it won’t work).

vitorgalvao:
Please [squash your commits](http://davidwalsh.name/squash-commits-git).

Casks/fldigi.rb
cedwardsmedia: So every time you use #{version}, the surrounding string needs double quotes, correct?

vitorgalvao: Yes.

cedwardsmedia: Gotcha. Thanks.

jawshooah: @cedwardsmedia Just FYI, this is a feature of the Ruby language, not homebrew-cask itself. More info [here](https://rubymonk.com/learning/books/1-ruby-primer/chapters/5-strings/lessons/31-string-basics).

PR #14547, opened 2015-10-20
bump vagrant-manager to 2.5.0 and github-desktop to 2.1.5

lourinaldi:


adityadalal924:
I think we prefer one cask change per PR, so if you could split them back up that'd be fantastic

PR #14570, opened 2015-10-21
Update the hashes for Amethyst 0.9.5 again

ianyh:
The bundled app name was incorrect, so linking was not working correctly. Can
confirm that it installs correctly now.


vitorgalvao:
Thank you for the contribution. It was merged directly as 43211b381de4458411c292d0ff30db07eaf180ea to keep commit history cleaner. Your contribution is still credited to you.

PR #14573, opened 2015-10-21
Add HotSLogs UploaderFX 1.3

thorstenschaefer:


eivindveg:
Please update this tap and make sure the app name matches the maven artifactId found in the uploader's pom.xml

vitorgalvao:
@eivindveg Even you being the owner of the app, its name when included in this project is not up to you. We have a strict [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) that must be adhered to, to avoid duplicates. We’ve had a huge issue with duplicates in the past, and forging a consistent naming scheme is the only way to keep it in check.

If, for whatever reason, that is unacceptable to you and you’d rather we not include your app here, we’ll honour that. However, changing the name of this cask (the one @thorstenschaefer submitted is correct) is non-negotiable, or it’ll make our lives worse (I can already see someone submitting this in the future with the correct name and it slipping by).

If you are fine with it, please let me know, and we can merge it.

eivindveg:
I've created an issue in my repository to discuss a more sane, final name for my application, and would very much like to have this settled, then updated here before being added to the repositories. I think that would be the best for everyone. See [#48 on eivindveg/HotSUploader](https://github.com/eivindveg/HotSUploader/issues/48) for the discussion that hopefully leads somewhere.

vitorgalvao:
Sounds good. As long as the name of the cask matches the name of the app bundle, we’re fine with whatever name you pick.

vitorgalvao:
Note for maintainers: This is on hold until https://github.com/eivindveg/HotSUploader/issues/48 is closed and a definitive app name is picked.

eivindveg:
We've released version 2.0 of the application, which all in all should be a more sanely structured app. @zhedar will take over my handling/nagging on this issue as he can actually test and submit new releases to you.

adityadalal924:
Closing in favor of https://github.com/caskroom/homebrew-cask/pull/15562

PR #14575, opened 2015-10-21
Alfred: removed settings and added accessibility_access true

vitorgalvao:
As for removing settings, not only doesn’t it make a ton of sense to have it (if Alfred’s developers wanted you to see it as a separate app, they would have packaged it as such), it would break when we go off the linking model.


vitorgalvao:
Merged as 76af7e46ebf631e02be654b5fad61120158c70dd.

PR #14576, opened 2015-10-21
Add Atom Beta.app v1.1.0-beta1

williamboman:


Casks/atom-beta.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  app 'Atom Beta.app'
+  binary 'Atom.app/Contents/Resources/app/apm/node_modules/.bin/apm', :target => 'apm'
+  binary 'Atom.app/Contents/Resources/app/atom.sh', :target => 'atom'
+
+  postflight do
+    suppress_move_to_applications
+  end
+
+  zap :delete => [
```
williamboman: This might be a bit dangerous since it's the beta version. People might uninstall it but still have the regular `Atom.app` installed. Same goes for vice verse. Remove?

vitorgalvao: `zap` should remove as much as possible. It’d be one thing if it were deleting stuff from an unrelated app, but in this case, it is OK.

vitorgalvao:
[Regarding betas](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#beta-unstable-development-nightly-or-legacy-versions).

Casks/atom-beta.rb
williamboman:
```sh
brew cask install atom
# 6 months later, with a nicely configured ~/.atom directory
# oh, look - a beta channel
brew cask install atom-beta
# ugh that worked terribly
brew cask uninstall atom-beta
# wtf?
```

vitorgalvao: No, that will never happen. `uninstall` does not run `zap`, uninstalling means getting rid of the app, zap means getting rid of the app and its preferences.

williamboman: Ahh, TIL.

vitorgalvao: Also, `zap` will move to trash instead of deleting, in the future.

williamboman:
Sorry, opened https://github.com/caskroom/homebrew-versions/pull/1361.

PR #14585, opened 2015-10-22
Add CouchPotato cask.

jefbarn:


adityadalal924:
Found a versioned url, looks like the source lives at https://github.com/RuudBurger/CouchPotatoServer/

Thus, download url can (and should) be changed to something like `"https://github.com/RuudBurger/CouchPotatoServer/releases/download/build%2F#{version}/CouchPotato-#{version}.macosx-10_6-intel.zip"`; and the version and shasum updated accordingly. You will also need a comment above url, and you can add in an appcast for good measure.

Sample cask -> https://github.com/caskroom/homebrew-cask/blob/master/Casks/aether.rb

jefbarn:
Ok, thanks will do.
I wasn’t sure if the versioned url was preferred over the ‘latest’ url style.

PR #14586, opened 2015-10-22
Add LuxMark 3.1.

jefbarn:


adityadalal924:
Looks like you made a small typo in `url`
Would appreciate it if you could fix and [squash your commits](http://davidwalsh.name/squash-commits-git)

Thanks!

PR #14590, opened 2015-10-22
Update Jumpshare.app to version 2.1.1

vinkla:

PR #14593, opened 2015-10-22
Update Jumpshare.app to latest version

vinkla:
Replaces #14590


vitorgalvao:
Thank you, but we prefer versioned casks. Could you resubmit the other one? It was better in that regard.

vinkla:
I've reopened the other one.

PR #14604, opened 2015-10-22
Delete goagentx.rb

vitorgalvao:
Seems like app disappeared.


vitorgalvao:
Merged as 63b954182c705c808b06124217ec9045128fbb74.

PR #14646, opened 2015-10-25
Add UnofficialWhatsApp.app v1.1.0

OndroNR:


Casks/unofficial-whatsapp-desktop.rb
```diff
@@ -0,0 +1,18 @@
+cask :v1 => 'unofficial-whatsapp-desktop' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'unofficialwhatsapp' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `unofficialwhatsapp.rb`.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 29cde07ca55cecf73d0dca22a11f7d72bfc55237. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

OndroNR:
Thanks for renaming and accepting.

PR #14659, opened 2015-10-26
automatically download beta version of Proxifier if running on currently-not-supported El Capitan

realdimas:
Last stable version of Proxifier was released in December 2014 (see https://www.proxifier.com/mac_download.htm).
Its kernel module is incompatible with El Capitan.
Vendor offers a beta version to El Capitan users since September 10th.

I think we should at least warn user if they try to install defunct version, if not automatically switch to Beta version.
This PR changes download URL for El Capitan users and adds a caveat.
I think that would be the most beneficial to the user.

I'll submit a revert once vendor release stable version supporting El Capitan.

### DIFF
--- a/Casks/proxifier.rb
+++ b/Casks/proxifier.rb
@@ -5,7 +5,7 @@ cask :v1 => 'proxifier' do
   url 'https://www.proxifier.com/distr/ProxifierMac.zip'
   name 'Proxifier'
   homepage 'https://www.proxifier.com/mac/'
-  license :unknown    # todo: change license and remove this comment; ':unknown' is a machine-generated placeholder
+  license :commercial
 
   app 'Proxifier.app'
 end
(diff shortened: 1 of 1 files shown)


Casks/proxifier.rb
```diff
@@ -2,10 +2,22 @@
   version :latest
   sha256 :no_check
 
-  url 'https://www.proxifier.com/distr/ProxifierMac.zip'
+  if MacOS.release == :el_capitan
```
vitorgalvao: Conditional should be reversed. `else` should always point to the newest version.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 48ca3f764d1384078f30f26fd0bb710a54eae97a. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #14687, opened 2015-10-27
Update Paragon-ntfs sha256

ghost:
the sha256 of the paragon-ntfs is error

### DIFF
--- a/Casks/paragon-ntfs.rb
+++ b/Casks/paragon-ntfs.rb
@@ -1,6 +1,6 @@
 cask :v1 => 'paragon-ntfs' do
   version '14'
-  sha256 'eb25919b05d37dbde9da3555db3f2fb97d54038e8e8c98394e33116e2c40861a'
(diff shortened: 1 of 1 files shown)


adityadalal924:
@vitorgalvao The commit is okay but the commit message should be updated (to something like `update shasum for paragon-ntfs`); what script should I use for cases like this?

vitorgalvao:
@dezon `prfixmaster` works here as well. Grab it and [amend](http://stackoverflow.com/a/179147/1661012).

Also keep in mind you don’t *have* to use those scripts (but are very free to do so, naturally).

vitorgalvao:
@adityadalal924 `prfixmaster` works here as well. Grab it and [amend](http://stackoverflow.com/a/179147/1661012).

Also keep in mind you don’t *have* to use those scripts (but are very free to do so, naturally).

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 0d898fef1bcef5a2377e0c9f52d2d14d3855e309. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #14707, opened 2015-10-27
add caveat for firefox

adidalal:
Adds in a 1Password caveat to firefox as well (currently the same as chrome). Wording is tentative, because it's not just 1Password that will have this problem, it's all externally installed addons.

Ref: https://github.com/caskroom/homebrew-cask/issues/13351

Thoughts? (I'll fix the rubocop error before merge)

### DIFF
--- a/Casks/firefox.rb
+++ b/Casks/firefox.rb
@@ -15,4 +15,11 @@ cask :v1 => 'firefox' do
                   '~/Library/Application Support/Firefox',
                   '~/Library/Caches/Firefox',
                  ]
+
+ caveats <<-EOS.undent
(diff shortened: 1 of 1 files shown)


vitorgalvao:
You can leave the wording as is. The warning was added for our own sanity, as we got constant reports of it, so no need to inform of other cases. It’s even better to have it be specific to 1Password instead of a more generic warning, so we’re really sure users understand it.

vitorgalvao:
Also keep in mind we’ll be able to get rid of it when moving behaviour is implemented, which is nice.

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in 6ffc1ac18a73387c965dc932fd031225c2049610. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #14709, opened 2015-10-27
Upgrade WavTap to 0.3.0 (retina icons)

johnhamelink:
Hey there - first time submitting a PR to homebrew-cask.

Just a quick update to an app that's very useful to Livestreamers, which was previously almost 3 years out of date. I've updated the version number, URL, and sha256 hash.

Best,
JH

### DIFF
--- a/Casks/wavtap.rb
+++ b/Casks/wavtap.rb
@@ -1,8 +1,8 @@
 cask :v1 => 'wavtap' do
-  version '0.2.0'
-  sha256 'ae4bbdb82c2b04ab7fa9995ed688b59d79e91f67122ed405409f19d74e7f60e5'
+  version '0.3.0'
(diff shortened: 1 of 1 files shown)


adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in 719b143b2af2cc24bd97f7ea1ed05eaef44a2842. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

adityadalal924:
cc @vitorgalvao 

Looks like this app won't work in 10.10+ without `kext-dev-mode` enabled, is this something that needs to be in a conditional caveat?

Ref: https://github.com/pje/WavTap#installation

vitorgalvao:
> is this something that needs to be in a conditional caveat?

Yes, it’d be nice to have it.

PR #14712, opened 2015-10-27
fix broken rubocop on ff caveat

adidalal:
Pushed wrong version of https://github.com/caskroom/homebrew-cask/pull/14707

Fixing rubocop errors on FF cask


adityadalal924:
Merged as f754bcb7dd1c4c20cd6282bfa1389d032b644dc7.

PR #14715, opened 2015-10-27
CONTRIBUTING.md: add rubocop instructions

vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/issues/14713.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -255,6 +255,16 @@ If everything looks good, you’ll also want to make sure your Cask passes audit
 brew cask audit my-new-cask --download
 ```
 
+You should also check stylistic details with the [`rubocop`](https://github.com/bbatsov/rubocop) gem:
+
+```bash
+cd "$(brew --repository)/Library/Taps/caskroom/homebrew-cask"
+bundle install
+bundle exec rubocop Casks/my-new-cask.rb
+```
+
+Keep in mind all of these checks will be made when you submit your PR, so by doing them in advance you’re saving everyone a lot of time and trouble.
+
 If your application and homebrew-cask do not work well together, feel free to [file an issue](https://github.com/caskroom/homebrew-cask/issues) after checking out open issues.
 
 ## Finding a Home For Your Cask


jawshooah:
This won't work as expected, since `rubocop` will use the default ruleset if it can't find one in the current working directory.

Better would be something like this:

```
cd "$(brew --repository)/Library/Taps/caskroom/homebrew-cask"
bundle install
bundle exec rubocop Casks/my-cask.rb
```

vitorgalvao:
Are you sure? Because my tests suggest it does work.

I have a global `.rubocop` in my home directory. Among other rules, it prefers `%x(…)` over backticks. If I’m in my home directory, move a cask that uses backticks ([`jce-unlimited-strength-policy.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/jce-unlimited-strength-policy.rb)) to `/tmp` and run `rubocop` on it, it finds the backticks as an error. However, if (still inside my home directory) I `rubocop "$(brew --repository)/Library/Taps/caskroom/homebrew-cask/Casks/jce-unlimited-strength-policy.rb", then it does not find an error.

This seems to suggest rather than just looking for a ruleset in the current directory, it looks in the directory of the file you’re checking and its parent directories.

jawshooah:
Strange:

```
$ cd ~
$ rubocop /usr/local/Library/Taps/caskroom/homebrew-cask/Casks/adobe-air.rb
Inspecting 1 file
C

Offenses:

/usr/local/Library/Taps/caskroom/homebrew-cask/Casks/adobe-air.rb:22:13: C: Unnecessary spacing detected.
      :rmdir  => '~/Library/Application Support/Adobe/'
            ^

1 file inspected, 1 offense detected

$ cd /usr/local/Library/Taps/caskroom/homebrew-cask
$ bundle exec rubocop /usr/local/Library/Taps/caskroom/homebrew-cask/Casks/adobe-air.rb
Inspecting 1 file
Warning: unrecognized cop Style/CommandLiteral found in /usr/local/Library/.rubocop.yml
.

```

vitorgalvao:
I can reproduce that. Not sure why I could do it the other way, but I can reproduce that.

Either way, we can’t have a solution if not everyone can reproduce it. Will change to your suggestion.

jawshooah:
You could also leave out the bit about having `rubocop` installed, since `bundle install` takes care of that.

vitorgalvao:
Right. Changed.

jawshooah:
:+1:

PR #14718, opened 2015-10-28
Zap thunderbird configuration and cache files.

renard:


Casks/thunderbird.rb
```diff
@@ -9,4 +9,12 @@
   tags :vendor => 'Mozilla'
 
   app 'Thunderbird.app'
+
+  zap :delete => [
+        '~/Library/Thunderbird',
+        '~/Library/Caches/Thunderbird',
+        '~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/org.mozilla.thunderbird.sfl',
+        '~/Library/Saved Application State/org.mozilla.*',
```
vitorgalvao: `:delete` doesn’t support expansion. You’ll need to be more specific.

You’re being too broad, anyway, this looks like it’ll delete saved states for other Mozilla apps, such as Firefox.

renard: Ok so line 17 `'~/Library/Saved Application State/org.mozilla.*',` could be removed from patch.
Do you want me to fix that?

vitorgalvao: Yes, one of them should go (why are you repeating it?) and the other should point to a specific file.

renard: I am not sure to understand your meaning: "one of them should go (why are you repeating it?) and the other should point to a specific file"

There are actually several things to remove:
- `~/Library/Caches/Thunderbird/`
- `~/Library/Thunderbird/`
- `~/Library/Saved Application State/org.mozilla.thunderbird.savedState/`
- `~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/org.mozilla.thunderbird.sfl`


This is found using:
`find ~/Library | grep -i thunderbird`

vitorgalvao: That’s exactly what it should be. That last comment.

renard: ok should be good now.

PR #14721, opened 2015-10-28
Add Guild Wars 2 installer

d8uv:


Casks/guild-wars-2.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'guild-wars-2' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'guild-wars' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `guild-wars.rb`.

Casks/guild-wars-2.rb
```diff
@@ -0,0 +1,12 @@
...
+  version :latest
+  sha256 :no_check
+
+  # it's the official cdn for guild wars 2
```
vitorgalvao: Comment needs to be `# amazonaws.com is the official download host per the vendor homepage`. The same template needs to be used for every cask.

Where on the page is the download for this, though, so I can confirm it?

vitorgalvao:
It’s missing `uninstall`. There are [scripts](https://github.com/caskroom/homebrew-cask/tree/master/developer/bin) and [documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#uninstall-stanza-details) geared for that.

Casks/guild-wars-2.rb
d8uv: Guild Wars is an entirely different game from Guild Wars 2. Changing the name to Guild Wars would be confusing and would guarantee duplicates.

Casks/guild-wars-2.rb
d8uv: The actual download page is hidden behind some horrible sign-in malarky. You have to go to guildwars2.com, click on "My Account" under the top "Services" bar. This will send you to account.arena.net (ArenaNet is the developer of the game). After you sign in, you have to go to the "Download Game Clients" page. Then, you click the "Download for OSX" button, which sends you to a url on account.arena.net which is just a redirect to the .dmg hosted on amazonCDN.

I wanted an easier way than doing all of that. :(

d8uv:
I used Backblaze.rb as a template... though I really should have used battle-net.rb as the template. I'll look into it after work.

Casks/guild-wars-2.rb
vitorgalvao: How would having a more restrictive name guarantee duplicates? Makes no sense.

Either way, then it needs to be `guild-wars2` (no `-` before numbers). Also, needs [a comment like this](https://github.com/caskroom/homebrew-cask/blob/ccaf00a189cdbe50109ba6bf294c2b72aecba784/Casks/c3.rb#L2).

Casks/guild-wars-2.rb
vitorgalvao: Then it is a [walled build](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#unofficial-vendorless-and-walled-builds), and we cannot accept it here. Submit to [caskroom/unofficial](https://github.com/caskroom/homebrew-unofficial) instead.

PR #8523, opened 2014-12-30
attempt to clean up launchctl plists on uninstall

rolandwalker:
separate `launchctl unload` and comment that this command was intended for unloading jobs by pathname rather than bundle ID. Leave that undocumented for now as it is untested.

PR #14730, opened 2015-10-28
Remove launchjob files during uninstall, even if they are not loaded

### LINKED ISSUE #14728: Launchctl uninstall do not clean-up users LaunchAgents .plist files
(moving discussion from #14723)

launchctl uninstall seems to be designed to delete LaunchAgents .plist files as well, but it does not work on users LaunchAgents.
[/lib/hbc/artifact/uninstall_base.rb#L215-L223](../blob/c0ed120631bd6ab5607d99dfa0efff82b7398959/lib/hbc/artifact/uninstall_base.rb#L215-L223):
```ruby
if result.success?
  paths = ["/Library/LaunchAgents/#{service}.plist",
           "/Library/LaunchDaemons/#{service}.plist"]
  paths.each { |elt| elt.prepend('~') } unless with_sudo
  paths = paths.map { |elt| Pathname(elt) }.select(&:exist?)
  paths.each do |path|
    @command.run!('/bin/rm', :args => ['-f', '--', path], :sudo => with_sudo)
  end
end
```

Issue seems to be related to FileTest.exist not supporting tilde:
```ruby
irb(main):001:0> FileTest.exist?('./')
=> true
irb(main):002:0> FileTest.exist?('~/')
=> false
```

CC @vitorgalvao, @jawshooah

jawshooah:
This fixes the original implementation in #8523 to correctly remove user launchjob files, and to do so even when the jobs have not been loaded.

Fixes #14728

[Review on Reviewable](https://reviewable.io/reviews/caskroom/homebrew-cask/14730)

### DIFF
--- a/lib/hbc/artifact/uninstall_base.rb
+++ b/lib/hbc/artifact/uninstall_base.rb
@@ -212,16 +212,14 @@ class Hbc::Artifact::UninstallBase < Hbc::Artifact::Base
           plist_status = @command.run('/bin/launchctl', :args => ['list', service], :sudo => with_sudo, :print_stderr => false).stdout
           if %r{^\{}.match(plist_status)
             result = @command.run!('/bin/launchctl', :args => ['remove', service], :sudo => with_sudo)
-            if result.success?
-              paths = ["/Library/LaunchAgents/#{service}.plist",
-                       "/Library/LaunchDaemons/#{service}.plist"]
-              paths.each { |elt| elt.prepend(ENV['HOME']) } unless with_sudo
-              paths = paths.map { |elt| Pathname(elt) }.select(&:exist?)
-              paths.each do |path|
-                @command.run!('/bin/rm', :args => ['-f', '--', path], :sudo => with_sudo)
-              end
-            end
             sleep 1
+          end
+          paths = ["/Library/LaunchAgents/#{service}.plist",
+                   "/Library/LaunchDaemons/#{service}.plist"]
+          paths.each { |elt| elt.prepend(ENV['HOME']) } unless with_sudo
+          paths = paths.map { |elt| Pathname(elt) }.select(&:exist?)
+          paths.each do |path|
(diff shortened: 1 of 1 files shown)


lib/hbc/artifact/uninstall_base.rb
```diff
@@ -209,18 +209,16 @@ def dispatch_uninstall_directives(stanza, expand_tilde=true)
...
+          paths = ["/Library/LaunchAgents/#{service}.plist",
+                   "/Library/LaunchDaemons/#{service}.plist"]
+          paths.each { |elt| elt.prepend(ENV['HOME']) } unless with_sudo
+          paths = paths.map { |elt| Pathname(elt) }.select(&:exist?)
+          paths.each do |path|
+            @command.run!('/bin/rm', :args => ['-f', '--', path], :sudo => with_sudo)
+          end
           plist_status = @command.run('/bin/launchctl', :args => ['list', service], :sudo => with_sudo, :print_stderr => false).stdout
           if %r{^\{}.match(plist_status)
             result = @command.run!('/bin/launchctl', :args => ['remove', service], :sudo => with_sudo)
```
forkbomber: @jawshooah any specific reason to rm .plist file and only after it run `launchctl remove`?
This is a reversal of previous flow and feels unnatural.
I'd assume it works just fine, but we can't safely assume that another launchctl rewrite at Apple wouldn't break our sequence (i. e. first `rm .plist` and only then `launchctl remove`).

jawshooah: Good point. I tested locally and there were no problems, but there's no guarantee that won't change. Thanks for the review!

vitorgalvao:
Looks fine to me. We should cut a release after this is in (there were other small fixes in the meantime as well)

vitorgalvao:
Are you cutting the new release, or do you want me to do it?

jawshooah:
Took care of it myself.

PR #14749, opened 2015-10-29
Update UninstallPKG to 1.0.18

buo:


vitorgalvao:
Though there’s no harm in it, you don’t need to waste time updating the shasum of appcasts. We don’t yet take advantage of the feature, so they’ll all likely need changing when we do, anyway,

vitorgalvao:
Thank you for the contribution. It was merged directly as 3a85f7da61322779cda73e3bc9b455190b4d35d4 to keep commit history cleaner. Your contribution is still credited to you.

PR #14811, opened 2015-10-30
Move semulov back to using :latest

Amorymeltzer:
Updated version, dev appears to no longer be offering versioned dl links
See #6084


vitorgalvao:
Thank you for the contribution. It was merged directly as f1cb2670d18c4f781c81163cf361c0f5d882e582 to keep commit history cleaner. Your contribution is still credited to you.

PR #14821, opened 2015-10-31
add BOINC 7.6.12

jm3:
add Berkeley Open Infrastructure for Network Computing, aka BOINC!]

if anyone can confirm whether I did the pkg installer the right way, that would be helpful.

### DIFF
--- a/Casks/boinc.rb
+++ b/Casks/boinc.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'boinc' do
+  version '7.6.12'
+  sha256 '194c91040807d995a5f54574c207d12b8970008319e146abb8fcfa13f75e39c6'
+
+  url "http://boinc.berkeley.edu/dl/boinc_#{version}_macOSX_x86_64.zip"
+  name 'Berkeley Open Infrastructure for Network Computing (BOINC)'
+  homepage 'https://boinc.berkeley.edu/'
+  license :gpl
+  installer :script => "boinc_#{version}_macOSX_x86_64/BOINC Installer.app/Contents/MacOS/BOINC Installer"
+  uninstall :pkgutil => 'edu.berkeley.boinc'
+end


Casks/boinc.rb
```diff
@@ -0,0 +1,10 @@
+cask :v1 => 'boinc' do
+  version '7.6.12'
+  sha256  '194c91040807d995a5f54574c207d12b8970008319e146abb8fcfa13f75e39c6'
+
+  url      'http://boinc.berkeley.edu/dl/boinc_7.6.12_macOSX_x86_64.zip'
```
fanquake: Use #{version} in the url

vitorgalvao:
Don’t space things like that. Look at other casks and follow their structure.

That installation method won’t work. You’re just calling the binary inside the bundle, which is useless. It’ll still call a GUI. Use [`installer :manual`](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#installer-manual) instead.

Also, it’s missing `uninstall`. There are [scripts](https://github.com/caskroom/homebrew-cask/tree/master/developer/bin) and [documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#uninstall-stanza-details) geared for that.

jm3:
thanks for the feedback! updated the PR with working uninstall stanza, un-aligned variables, and `#{version}` substitution in the URL + installer path strings. let me know if there's anything else i should change.

@vitorgalvao calling the installer binary inside the bundle works fine for me; i learned that directly from the homebrew-cask contribution instructions.

vitorgalvao:
> calling the installer binary inside the bundle works fine for me; i learned that directly from the homebrew-cask contribution instructions.

I find that unlikely. Where exactly did you see that? We don’t like if it calls a GUI, and never did.

vitorgalvao:
> calling the installer binary inside the bundle works fine for me

It calls a GUI, so it doesn’t work fine.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 6413ecc69a304036813d1df47ba7d5dedda9b084. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

vitorgalvao:
I’d still like to know where you got that idea, though, so we can fix the documentation.

jm3:
@vitorgalvao k,  i changed the formula so it runs entirely on the command-line now:
`pkg "boinc_#{version}_macOSX_x86_64/BOINC Installer.app/Contents/Resources/BOINC.pkg"`

re: the documentation: yup, i knew that it launched the GUI installer, but when I *didn't* invoke it that way, the app doesn't get installed. That style of invoking the script INSIDE the app bundle (or the raw pkg file inside the app bundle, which I've since changed it to), was the only way that it seemed to work, which i got by trying all the things in the CONTRIBUTION docs, under pkg instructions and `installer :script`. when you unzip the result of the BOINC URI, you get an .app bundle, but making homebrew cask run that directly with `pkg` or `app` didn't install the app. can you point me to which section of the docs would have showed me the right way to do it?

since the headless installer is a bit slow, with no terminal feedback to the user, is there a simple way I can output a status message to the terminal so that the user knows it's running successfully and hasn't died? or is that poor form?

vitorgalvao:
> i changed the formula so it runs entirely on the command-line now:
pkg "boinc_#{version}_macOSX_x86_64/BOINC Installer.app/Contents/Resources/BOINC.pkg"

That works, thank you. Already updated the cask.

> since the headless installer is a bit slow, with no terminal feedback to the user

Not that slow, and it has a bouncing icon, so not that big of a deal.

> yup, i knew that it launched the GUI installer, but when I didn't invoke it that way, the app doesn't get installed.

It also didn’t get installed with your solution. It called a GUI, which is worse since is not scriptable.

> was the only way that it seemed to work

But it didn’t work, because it did not finish the installation.

> trying all the things in the CONTRIBUTION docs, under pkg instructions and `installer :script`.

You were using `installer :script` wrongly. It says right there (emphasis added): “introduces a series of key-value pairs describing a command which will **automate completion of the install**”.

> can you point me to which section of the docs would have showed me the right way to do it?

I already linked you to it. It’s `installer :manual`.

PR #14829, opened 2015-10-31
Remove Freelancy.app (not available anymore)

victorpopkov:
It seems that this app is not available anymore.


vitorgalvao:
Yep. Seems like it’s only available as a webapp (and Chrome app).

vitorgalvao:
Merged as e82730f4ba81acf78ebc6ce2eb576455c7095ff4.

PR #14830, opened 2015-10-31
Remove FoldIt.app (not available anymore)

victorpopkov:
It seems that the app is not available anymore. The homepage and download URL don't work.


vitorgalvao:
Merged as 32d9d35c805e85f809e478c64f429e86928c261b.

PR #14837, opened 2015-10-31
Add vimr binary link to VimR.app

ldong:
Users now are able to use `vimr` in the terminal.

### DIFF
--- a/Casks/vimr.rb
+++ b/Casks/vimr.rb
@@ -10,4 +10,5 @@ cask :v1 => 'vimr' do
   license :gpl
 
   app 'VimR.app'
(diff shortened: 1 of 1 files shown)


Casks/vimr.rb
```diff
@@ -10,4 +10,5 @@
   license :gpl
 
   app 'VimR.app'
+  binary "VimR.app/Contents/Resources/vimr"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

ldong:
@vitorgalvao I added another commit to change double to single quote, I hope that its okay.

PR #14899, opened 2015-11-03
Upgrade PyCharm CE.app from 4.5.4 to 5.0

tsparber:
This commit upgrades PyCharm CE.app to latest version 5.0.

JetBrains offers also a version with a bundled JDK, should this cask
be switched to that version and the caveeat removed?


fanquake:
The version with the bundled JDK is available from the versions repo, [here](https://github.com/caskroom/homebrew-versions/blob/master/Casks/pycharm-bundled-jdk.rb).

PR #14903, opened 2015-11-03
A cask for Tor Messenger (beta)

vmax:
Added a cask for new messenger by Tor Project
https://blog.torproject.org/blog/tor-messenger-beta-chat-over-tor-easily


Casks/tormessenger.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'tormessenger' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'tor-messenger' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

When you do, you’ll find the cask already exists.

PR #14920, opened 2015-11-04
Upgrade to DockerToolbox 1.9.0

joschi:
https://github.com/docker/toolbox/releases/tag/v1.9.0

PR #14922, opened 2015-11-04
Upgrade dockertoolbox to v1.9.0

nxtonic:


nxtonic:
Alas, already beat to it! #14920

PR #14944, opened 2015-11-04
Use https instead of http for Webstorm download

ddm:


adityadalal924:
This has the bundled JDK, which belongs in `homebrew versions`

ddm:
OK, my bad. I saw that IntelliJ-CE did the same so I thought it was OK...

See https://github.com/caskroom/homebrew-cask/blob/dfdaea12a0b032bd44f49b7220d94797fd30a8df/Casks/intellij-idea.rb and https://github.com/caskroom/homebrew-cask/blob/dfdaea12a0b032bd44f49b7220d94797fd30a8df/Casks/intellij-idea-ce.rb

ddm:
Can someone more knowledgeable just switch to https?

vitorgalvao:
@adityadalal924 They all have the bundled-jdk now, as I understand it. Versions without it bundled are no longer distributed. Is that correct, @fanquake?

adityadalal924:
@vitorgalvao I see, that is likely to cause some confusion. Maybe those casks should be removed from homebrew-versions, then?

Can confirm that https://download.jetbrains.com/webstorm/WebStorm-11.0.0-custom-jdk-bundled.dmg is the correct HTTPS URL though.

adityadalal924:
Thank you for the contribution. It was merged directly as 30a0994393129bff0b363e908110def53d353f1e to keep commit history cleaner. Your contribution is still credited to you.

adityadalal924:
@ddm You were correct, thanks!

ddm:
@adityadalal924  @vitorgalvao Thank you for being patient. I'm still a bit a bit new to this...

fanquake:
@vitorgalvao There is still a mix between the jetbrains casks. Looks like a few pulls have been merged to remove the bundled versions that are now in cask. I'll keep and eye out for any more switchovers.

PR #14950, opened 2015-11-04
Upgrade MultiBit.app to v0.5.19

carloscheddar:
Multibit now diverged into legacy and HD versions so the url had to be  updated to the new version.

### DIFF
--- a/Casks/multibit.rb
+++ b/Casks/multibit.rb
@@ -1,8 +1,8 @@
 cask :v1 => 'multibit' do
-  version '0.5.18'
-  sha256 '0d2fe6fa68385c1ca964d9588272787dabffbc2061f29ebaab422317d0972257'
+  version '0.5.19'
+  sha256 'f84aefa0b3762e36659ea3e71806f747db4198641d658d88c8772978b23f99dc'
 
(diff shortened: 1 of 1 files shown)


Casks/multibit.rb
```diff
@@ -1,8 +1,7 @@
 cask :v1 => 'multibit' do
-  version '0.5.18'
-  sha256 '0d2fe6fa68385c1ca964d9588272787dabffbc2061f29ebaab422317d0972257'
-
```
vitorgalvao: Do not remove this line.

carloscheddar: I just updated the values but I can amend the commit if it should be structured in another way.

vitorgalvao: You removed an empty line that should remain.

carloscheddar: Ah I see :+1:. I'll update it now.

carloscheddar: Fixed.

vitorgalvao:
Thank you for the contribution. It was merged directly as ff69406b68537289b724e56c34e42ab66d17c100 to keep commit history cleaner. Your contribution is still credited to you.

PR #14960, opened 2015-11-05
Update Litecoin to v-0.10.2.2

ralyodio:
I updated path to dmg but I'm not sure where :key_id comes from -- I kept it the same.

Caskroom/cask/litecoin


fanquake:
Please include the name of the cask being updated in the commit message. You'll also need to squash your commits.

chovy:
Done

PR #14972, opened 2015-11-05
Update Android Studio to version 1.4.1.0

RJHsiao:
Hi all,

I use static url to replaced original url with `#{version}` variable in this change,  
because no only version number, but also build number are in the download url.

Maybe we can add a variable for build number to generate download url dynamically next time?

Release page: http://tools.android.com/download/studio/builds/1-4-1

### DIFF
--- a/Casks/android-studio.rb
+++ b/Casks/android-studio.rb
@@ -1,9 +1,9 @@
 cask :v1 => 'android-studio' do
-  version '1.4.0.10'
-  sha256 '1168faf59ffbe485bfe0e95e80bd9672a3ae5599efd2b720cf4b825bd859ffa0'
+  version '1.4.1.0'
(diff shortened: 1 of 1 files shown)


fanquake:
Please revert the changes to #{version}

vitorgalvao:
Removing interpolation is hardly ever the solution. Being smarter about it.

vitorgalvao:
Removing interpolation is hardly ever the solution. Being smarter about it is. However, in this case, its’s like they use two different versions, making it harder for us to make it better but still understandable. Since the whole `url` needs to change anyway, there’s no hard in removing the interpolation here.

vitorgalvao:
Thank you for the contribution. It was merged directly as 9dcc849f0864cbfa2602215ce6aecaf5282394d5 to keep commit history cleaner. Your contribution is still credited to you.

RJHsiao:
@vitorgalvao 
Thanks for your understanding. ;)

PR #14975, opened 2015-11-05
Update the version of antirsi

michaek:
This is the current (and presumably last?) version that can be downloaded without the Mac App Store.


vitorgalvao:
In that case, we need to remove it.

byronyi:
Why remove it instead of merge this commit?

vitorgalvao:
Because it is only available on the Mac App Store, and we don’t accept those apps. We’re not going to have an app that is perpetually outdated when there is an up-to-date version somewhere else. That is both a bad experience and potentially dangerous.

Mange:
That's too bad. This is the last free version of the app as the version in the App store became a payed app. I also think the free version looks better.

PR #14993, opened 2015-11-06
Update valentina-studio to v6.0.13

MatzFan:


vitorgalvao:

victorpopkov:
It seems that the download URL points to the latest version, so I guess the `latest` should be used here instead.

MatzFan:
Good point. Fixed with PR #14994

adityadalal924:
Duplicate of #14994

PR #15005, opened 2015-11-07
Upgrade trezor-bridge to v1.1.2

johnsockwell:
Updated the version number and sha256 to v1.1.2.
Additionally removed 'protobuf' and 'libmicrohttpd' dependencies since these are already included in the latest package.


vitorgalvao:
Thank you for the contribution. It was merged directly as b069c0f0c9cd1cf7134061464ca23ce667fa9f30 to keep commit history cleaner. Your contribution is still credited to you.

PR #15016, opened 2015-11-07
Fix checksum in anylogic.rb Cask

clearofcant:


victorpopkov:
Thank you for the contribution. It was merged directly as 0ff178b2ef7c039ee02eb04c68ac3bd2149487d6 to keep commit history cleaner. Your contribution is still credited to you.

PR #15018, opened 2015-11-07
Upgrade 'The Hit List' to 1.1.17

hutschen:


Casks/the-hit-list.rb
```diff
@@ -1,11 +1,10 @@
 cask :v1 => 'the-hit-list' do
-  version :latest
-  sha256 :no_check
+  version '1.1.17'
+  sha256 '0a6bd16d26c19a27f2146cdba27fc5c8119da8c382643d285203b41fadaebe98'
 
-  url 'http://www.potionfactory.com/files/thehitlist/TheHitList.zip'
-  appcast 'http://www.potionfactory.com/appcast/thehitlist.php'
+  url 'http://distrib.karelia.com/downloads/TheHitList-302.zip'
```
vitorgalvao: What does the `302` mean? Is it the build number?

Casks/the-hit-list.rb
```diff
@@ -1,11 +1,10 @@
 cask :v1 => 'the-hit-list' do
-  version :latest
-  sha256 :no_check
+  version '1.1.17'
+  sha256 '0a6bd16d26c19a27f2146cdba27fc5c8119da8c382643d285203b41fadaebe98'
 
-  url 'http://www.potionfactory.com/files/thehitlist/TheHitList.zip'
-  appcast 'http://www.potionfactory.com/appcast/thehitlist.php'
```
vitorgalvao: Don’t remove the appcast. Why would you? It’s still working.

Casks/the-hit-list.rb
hutschen: I'm not pretty sure but I think it's the build number. I was not able to find a generic download link.

Casks/the-hit-list.rb
hutschen: Sorry, the download url wasn't working so I supposed the same for the appcasts' url and don't checked it before I removed it. I just found out that the appcasts' url redirects to another one which I added to the file now.

vitorgalvao:
Thank you for the contribution. It was merged directly as f1e00110d58101668f6383a42ed25314318b7142 to keep commit history cleaner. Your contribution is still credited to you.

PR #15048, opened 2015-11-09
Upgrade Netbeans CPP and PHP to v8.1

tsparber:
This commit upgrades netbeans-cpp and netbeans-php from v8.0.2 to v8.1


adityadalal924:
Manually created two separate PRs to update the casks, as generally we like one cask change per commit. 

Thank you anyway!

PR #15063, opened 2015-11-09
speeds@ver: initial cask

jedahan:
Screensaver that shows speedruns :)

### DIFF
--- a/Casks/speeds-ver.rb
+++ b/Casks/speeds-ver.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'speeds-ver' do
+  version :latest
+  sha256 :no_check
+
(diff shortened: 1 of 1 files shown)


Casks/speeds-ver.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'speeds-ver' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'speeds-at-ver' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `speeds-at-ver.rb`.

jedahan:
Resubmitted in a more appropriately named branch

PR #15076, opened 2015-11-10
Update MarkRight to 0.1.5

dvcrn:
Also wondering if there is a better (automated?) solution for pushing a release or if a pull request is necessary every time

### DIFF
--- a/Casks/markright.rb
+++ b/Casks/markright.rb
@@ -1,6 +1,6 @@
 cask :v1 => 'markright' do
-  version '0.1.3'
-  sha256 'b386b2f24294ceafd012cd009da8af0a329d5a1c560ca7a9b18088bdf086965e'
+  version '0.1.5'
+  sha256 '3631f805285293e2653fe5e5d96c975f85b145ad841694661ec9cd572b2907d5'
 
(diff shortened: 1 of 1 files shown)


jawshooah:
Only the maintainers of this repository have permission to push commits, so no.

However, you could easily automate the process of opening a PR as part of a release script.

dvcrn:
I'm just wondering if the human time spent on just merging version bumps could be eliminated somehow. 
Maybe a github bot that just checks if the version (and hash) changed in new PRs and merges automatically.

jawshooah:
Not really. We'll always need at least one set of eyes scanning incoming PRs to ensure that changes make sense, even if they are just version bumps.

For example, say that someone submits a PR that would downgrade a cask to an older version. We wouldn't want a bot auto-merging such a change, as it would obviously demand an explanation. If the bot was simply checking that only the version number and checksum have changed, given the diversity of version number formats between casks, this would be a difficult edge case to catch.

Also, we review and merge PRs at a pretty fast clip around here. There are only ever about 25 open at a time (including long-lived ones that require discussion), and that number is steadily shrinking as we add maintainers to the project.

PR #15085, opened 2015-11-10
Upgrade GeekTool to v3.1.3

Zirro:
Updates version and sha256. Removes appcast which has not been updated since 2013, despite being included in the recent versions of the app. This and the previous version fixes issues that appeared with El Capitan.


victorpopkov:
Thank you for the contribution. It was merged directly as 2751b06177f93b328b9e50c745de5f5faf332d84 to keep commit history cleaner. Your contribution is still credited to you.

PR #15095, opened 2015-11-10
Add Apcupsd 3.14.13

reitermarkus:


adityadalal924:
This relies on a decision being made about https://github.com/caskroom/homebrew-cask/issues/9203 - the "shim script" in this case is an `installer_choices.xml` that allows for unattended install.

This actually doesn't seem like a very uncommon scenario, so the choice might be to either a) implement this as core functionality (extend cask DSL) or b) allow scripts like this (because of the relative stagnation of the main codebase at the moment)

cc @caskroom/maintainers

vitorgalvao:
@reitermarkus What is the `preflight` intending to accomplish, exactly?

reitermarkus:
The preflight unchecks the daemon in the PKG installer, otherwise the installer would fail under El Capitan.

vitorgalvao:
Yeah, this is one of those cases. This is a bad way to fix the issue. The correct way is to contact the developers / fix it upstream. Crippling the cask / building workarounds in homebrew-cask when the issue is an upstream one is **never** the solution.

Please remove the `preflight` and instead have a `depends_on` that’ll prevent it from installing in El Capitan. That is the correct course of action. *Unless* you know this will never be fixed upstream (perhaps there’s a technical reason that can’t be fixed), in which case it’s not worth adding at all.

reitermarkus:
Ok, then I better remove it completely and close this because I have added it to https://github.com/Homebrew/homebrew/pull/45899 anyway.

vitorgalvao:
Even better! In the spirit of https://github.com/caskroom/homebrew-cask/issues/14384, not having duplicates between the projects starts to be desirable.

Issue #14977, opened 2015-11-05
Cask request - Microsoft Remote Desktop Beta


my2ter:
Hi there,

current cask **remote-desktop-connection: 2.1.1**
>has been officially discontinued upstream. It may stop working correctly (or at all) in recent versions of OS X.

but Microsoft released recently a new app:

https://rink.hockeyapp.net/apps/5e0c144289a51fca2d3bfa39ce7f2b06/

It would be nice to update the cask or create a new one.

Cheers

vitorgalvao:
Is there a homepage for that?

adityadalal924:
The Microsoft Remote Desktop app for Mac available in the Mac App Store and updated regularly, is our fully tested and fully featured application. It supports connecting to desktops, remote resources and Azure RemoteApp. This is the app you should use in a live operating environment.

The Microsoft Remote Desktop Beta app for Mac is a **pre-release app** available on HockeyApp (see instructions on how to install the app from HockeyApp in the section in this post on Installation and Checking for updates). The pre-release app showcases the new user interface design and features we would like to receive your feedback on. This app does not have all the features that the release version has, and will not have the release level of product support.

REF: http://blogs.msdn.com/b/rds/archive/2015/08/11/introducing-the-new-microsoft-remote-desktop-preview-app-for-mac.aspx

**tl;dr:** You can add the cask for the new beta app to `homebrew-versions`, and the existing app in homebrew should be removed (latest version is available only through MAS currently)

adityadalal924:
As this is a beta app, the issue should be moved to `homebrew-versions` if there is still any interest.

Thanks!

sj26:
For reference, here'a a link to the Microsoft Remote Desktop client on the Mac App Store:

http://appstore.com/mac/microsoftremotedesktop10

You can automate this using [mas](https://github.com/mas-cli/mas) on command line:

```sh
brew install mas
mas install 1295203466
```

Or in a [Brewfile](https://github.com/Homebrew/homebrew-bundle#readme):

```ruby
brew "mas"
mas "Microsoft Remote Desktop, id: 1295203466
```

PR #15108, opened 2015-11-12
Delete remote-desktop-connection.rb

adidalal:
Last updated 5/5/2011 and replaced by an app with the same name in the Mac App Store (and only available there)

Ref: #14977


adityadalal924:
Merged as 627175a0430f55332320a0f8bc05d6dbf2c05a99.

PR #15110, opened 2015-11-12
Get FileShuttle over HTTPS & update appcast URL.

Zegnat:
The old domain is no longer used, updating the `appcast` URL and its `sha256`. All resources are available over HTTPS.

---

Note on license: the next version of FileShuttle (3.x) is being developed under the MIT license. Older versions do have their source publicly available ([2.x source](https://github.com/FileShuttle/fileshuttle/tree/2.x)) but were never officially released under an open license. `:gratis` might be the best fit, considering [the application’s history](https://github.com/FileShuttle/fileshuttle/issues/9#issuecomment-35065124):

> the app previously was free then went to app store and now opensourced/githubbed

Will `:gratis` be acceptable here? [CASK_LANGUAGE_REFERENCE](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#valid-licenses) defines `:gratis` as closed-source but overall forgets to make a distinction between [open-source and source-available](https://en.wikipedia.org/wiki/Open-source_software#Open-source_vs._source-available).

I have left it as `:unknown` until one of the Cask contributors can clear that up for me.

### DIFF
--- a/Casks/fileshuttle.rb
+++ b/Casks/fileshuttle.rb
@@ -2,11 +2,11 @@ cask :v1 => 'fileshuttle' do
   version '2.1'
   sha256 '431b7cb2161eace5b8d39146bdd1b7cc9a85d335eb29175c1a1879208928565a'
 
-  url "http://fileshuttle.io/fileshuttle-#{version}.zip"
-  appcast 'http://updates.getfileshuttle.com/update.xml',
-          :sha256 => '9c8640556eedc9f07894e1c3d140f86ccdf27c95714aead2937ae562143079e0'
+  url "https://fileshuttle.io/fileshuttle-#{version}.zip"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
This is a great PR. Thank you for all the links.

Since they are working on changing the license, we don’t need to worry too much. You can set it as `:other`.

Zegnat:
Apparently my machine has decided I am not allowed to squash commits at the moment. I’ll sit down later and see if I can get `:other` in on this commit.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in ebbe1a37e22026d68d1d92cff559a1d4cb3ac098. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15118, opened 2015-11-12
Update Manuscripts.app to use :latest/:no_check for non-versioned URL.

jaredsampson:
No version in the download URL.  Converted to use the :latest/:no_check setup.


adityadalal924:
Duplicate of #15121 

Thanks for the effort, anyway!

Issue #2264, opened 2014-01-01
Services (i.e. workflows) installation doesn’t work


niksy:
Similar to #2206, simply symlinking Automator created service (with `.workflow` extension) doesn’t work. Symlinked service is shown inside Services menu, but activating it doesn’t do anything.

It will work only if you explictly confirm that you want to install service (by activating workflow file and pressing "Install").

Maybe create something similar to pull request #2216 (@miguelfrde)?

vitorgalvao:
Linking is a mess, and has proven to not be the best solution. That should be changed, and we’ll be working towards that.

PR #15138, opened 2015-11-13
Update korean-spelling-checker to 1.4.1

buo:


adityadalal924:
The download URL per the homepage is `http://appletree.or.kr/automator/korean-spelling-checker-workflow.zip`

I don't speak Korean - is the github one more updated, or are they the same?

buo:
I just compared those two versions, and they are the same
except the one from the homepage has QuickLook directory which is in .gitignore file.

The github repository is maintained by the same person who maintains the homepage,
so I think they should be always the same.

buo:
As discussed in issue #2264, the service with `.workflow` doesn't work unless users install it manually via Service Installer.
Hardlinking it didn't work for me, so `cp -r` is used instead to copy the `.workflow` to the `~/Library/Services/` directory.

adityadalal924:
It looks like the only two Casks that use the service stanza are this and `markdown-service-tools.rb`, but this seems like a band-aid rather than an actual fix.

As I'm not sure what do do in this case, I'll leave it up to @caskroom/maintainers

vitorgalvao:
Seeing as `service`s aren’t used much and we have https://github.com/caskroom/homebrew-cask/issues/13201 that automatically fixes this, this solution is not great. We also have `app`s that would benefit from such a solution, yet we don’t fix them in this manner. In this case, we’re sure to forget about this once the good solution is introduced.

The cask should be left as is regarding its installation (broken), with [an added `caveat`](https://github.com/caskroom/homebrew-cask/blob/012d03a2e0008b8f49cd7d2dc3bbbd8fcac3fea6/Casks/megasync.rb#L13L20).

buo:
I added caveats and squashed commits.

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in 1aa866c6393adc8269cf2edd13a8a68da3a48276. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15181, opened 2015-11-14
Update xquartz.rb

adidalal:
Add license and update appcast shasum and homepage


adityadalal924:
Merged as f0424544e657c40a441c57d1221814a8d5865932.

mrjcleaver:
Necessary because: http://apple.stackexchange.com/questions/169657/ssh-x-does-not-work-on-yosemite

PR #15196, opened 2015-11-15
Switch mactex download link

adidalal:
This link automatically redirects users to the best download location, but we lose "versioning" as the URL does not have a #{version}

The package is huge, so arguably download speed is important

Based on a comment in the Gitter chat by @fanquake

### DIFF
--- a/Casks/mactex.rb
+++ b/Casks/mactex.rb
@@ -1,9 +1,8 @@
 cask :v1 => 'mactex' do
-  version '20150613'
-  sha256 'c5f5b0fd853a17dab6e844fb5e893804af78d938fa18ee94ec3b257611a95c12'
+  version :latest
+  sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


adityadalal924:
Merged as b280b4e70c51767903371cf0923f3540d28001fa.

vitorgalvao:
Download speed is important, yes, but so are versions, and those are arguably more important. Isn’t there a way to get the best of both worlds?

[Example of a case where this change caused issues](https://github.com/caskroom/homebrew-cask/issues/15301).

adityadalal924:
Unfortunately, it looks like the round-robin redirect doesn't have any version in its URL.

We could "fix" this by going back to having an explicit version and shasum (and keeping the unversioned URL), but this will cause an ungraceful fail when the package is updated. Not sure if that's any better, but it's an option.

PR #15208, opened 2015-11-16
Add 360Controller driver v0.14

vin047:
Same as d235j-xbox360-controller-driver Cask but repository name changed

### DIFF
--- a/Casks/xbox360-controller-driver-improved.rb
+++ b/Casks/xbox360-controller-driver-improved.rb
@@ -1,0 +1,28 @@
+cask :v1 => 'xbox360-controller-driver-improved' do
+  version '0.14'
+  sha256 'a7efe48fae89aa592e904d1dc3ba9a73b2d644362eac88feb5e73a774e3fcf88'
+
+  url "https://github.com/360Controller/360Controller/releases/download/v#{version}-unofficial/360ControllerInstall_#{version}_unofficial.dmg"
+  appcast 'https://github.com/360Controller/360Controller/releases.atom'
+  name 'TattieBogle Xbox 360 Driver with improvements'
+  homepage 'https://github.com/360Controller/360Controller'
+  license :gpl
+
+  pkg 'Install 360 Controller.pkg'
+
+  uninstall :pkgutil   => 'com.mice.pkg.Xbox360controller',
+            :launchctl => 'com.mice.360Daemon',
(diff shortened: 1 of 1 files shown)


vitorgalvao:
The current cask needs to be changed. Not add a new one.

vitorgalvao:
Thank you for the contribution, but it’ll be fixed shortly in a different PR.

vin047:
@vitorgalvao the existing cask name no longer reflects the project name, which makes it confusing for newcomers.

vitorgalvao:
And that’s exactly why the cask was renamed. Keeping it with the old name to something that doesn’t exist makes no sense.

adityadalal924:
@vitorgalvao I think whats being asked is a rename from `xbox360-controller-driver` to `360Controller`, which seems to make sense given that the project itself has changed names (rather, it looks like the old project was discontinued, and this is the most active current fork)

adityadalal924:
and @vin047, we use homebrew-cask for stable versions, but feel free to submit a PR for "v0.15_beta3_unofficial" to homebrew-versions if you are so inclined.

vitorgalvao:
@adityadalal924 Not exactly. @vin047 was adding a cask that is just an existing one with a new repo, and leaving the existing one there (dead).

Either way, [the matter is fixed](https://github.com/caskroom/homebrew-cask/pull/15209). I’m also a user of the driver ever since the original (from Tattiebogle), and it makes sense to have it like this.

adityadalal924:
Ah I see. To me, the rename makes sense (ie. using the github project name as the canonical cask name), but as I don't use the cask, it doesn't make too much of a difference to me.

vin047:
@vitorgalvao apologies, I didn't realise you had already consolidated the existing casks. Makes sense now, thanks.

vitorgalvao:
> apologies, I didn't realise you had already consolidated the existing casks.

No worries.

> To me, the rename makes sense (ie. using the github project name as the canonical cask name)

Since we don’t have clear rules for `pkg` casks, I kept it like this for consistency, but also as a bit of proactivity with possible renames. If/when https://github.com/caskroom/homebrew-cask/issues/14384 goes through, we might or might not lose the ability of starting cask names with numbers, so since this one makes sense with a name before, this way we ensure it won’t need to change.

PR #15209, opened 2015-11-16
xbox360-controller-driver.rb (with changes) is now the only driver

vitorgalvao:
One of them hasn’t worked for many OS X releases, now, and the other changed repos.


vitorgalvao:
Merged as ecfd028b1565f0939ad51c11a5fe37b68d5ca209.

PR #15212, opened 2015-11-16
added explicit sudo and zap for amazon-cloud-drive

adidalal:
Ref: https://github.com/caskroom/homebrew-cask/issues/13608


adityadalal924:
Merged as 20e1b0c68e3c24d20e3de107da1924ac413ff92d.

PR #15216, opened 2015-11-16
Add Bose Soundtouch Controller 9.0.41

inz:


adityadalal924:
Please add in a `:sudo` (true or false, as required), as https://github.com/caskroom/homebrew-cask/issues/13608 will make this needed for all Casks.

inz:
Thanks for the heads-up. Done.

PR #15236, opened 2015-11-17
basictex: remove caveats and update uninstall

cebor:
Since El Capitan, the path is `/Library/TeX/texbin` and it gets added automatically to `/etc/paths.d` and `/etc/manpaths.d`.
Also remove TeX from path and Settings on uninstall.


vitorgalvao:
Thank you for the contribution. It was merged directly as 09be3f9e5b2e6a5ae5ac5b87d4ac780662593f07 to keep commit history cleaner. Your contribution is still credited to you.

PR #15242, opened 2015-11-17
updated macdown (0.5.3)

adidalal:
Switch to versioned URL and add in binary symlink, from https://github.com/caskroom/homebrew-cask/pull/15240


adityadalal924:
Merged as 473d0be51eea4a9494ab554b3add5eeffa18f98e.

PR #15248, opened 2015-11-17
update fitbit-connect (20150619_2.0.0.6583)

adidalal:
adds explicit sudo Ref https://github.com/caskroom/homebrew-cask/issues/13608


adityadalal924:
Merged as 1ebafffe65487452bd22f5302017f148c037b3b1.

PR #15251, opened 2015-11-17
Add Duet license

buo:
Duet turns an iPad into an extra display. The Duet for iPad is not free, but the app for Mac & PC is provided free of charge and not limited in functionality.


adityadalal924:
Merged as 80c15c91a325db759e38f2a44f4256d6ebbb88de.

PR #15266, opened 2015-11-18
Remove Kontroll.app

vinkla:
In the next 48 hours this application will be removed from the download link will be broken. The iOS application is already removed form the App Store.


adityadalal924:
Thank you for the contribution. It was merged directly as 692de69dc359d320111f099714ee6bda418a1bfb to keep commit history cleaner. Your contribution is still credited to you.

vinkla:
Thanks.

PR #15269, opened 2015-11-18
Add Endurance.app latest version

vinkla:
No matter how good your battery is, everyone could use a little more battery power! http://enduranceapp.com/


vitorgalvao:
Thank you for the contribution. It was merged directly as 61ddc60f861d114763a4131fc2872e3f17ee871a to keep commit history cleaner. Your contribution is still credited to you.

PR #15294, opened 2015-11-19
add cask file: 'cross-web-ex.rb'

daehyunpy:


Casks/cross-web-ex.rb
```diff
@@ -0,0 +1,13 @@
+cask :v1 => 'cross-web-ex' do
+  version '1.0.1.752'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

Casks/cross-web-ex.rb
```diff
@@ -0,0 +1,13 @@
+cask :v1 => 'cross-web-ex' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'crosswebex' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `crosswebex.rb`.

Casks/cross-web-ex.rb
```diff
@@ -0,0 +1,13 @@
...
+  version '1.0.1.752'
+  sha256 :no_check
+
+  url "https://open.shinhan.com/initech/extension/down/CrossWebEX.pkg?ver=#{$version}"
```
vitorgalvao: remove the `$`.

adityadalal924:
Please [squash your commits](https://davidwalsh.name/squash-commits-git), thanks.

vitorgalvao:
Thank you for the contribution. It was merged directly as 0240ae9b7d345a019c119110f705986a520f949f to keep commit history cleaner. Your contribution is still credited to you.

PR #15298, opened 2015-11-19
Bump elm-platform version

bltavares:


adityadalal924:
The shasum also needs to be updated when the version gets bumped, I've fixed it in https://github.com/caskroom/homebrew-cask/pull/15299 so I'm closing this PR

bltavares:
I was waiting the file to download, my bad.

Thank you for fixing it.

PR #15304, opened 2015-11-19
added cask for nanobox bundle http://nanobox.io/

liljenstolpe:
Cleaned up, squashed, and finally, I think, ready.

### DIFF
--- a/Casks/nanobox-bundle.rb
+++ b/Casks/nanobox-bundle.rb
@@ -1,0 +1,16 @@
+cask :v1 => 'nanobox-bundle' do
+  version :latest
+  sha256 :no_check
+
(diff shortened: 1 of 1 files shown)


adityadalal924:
`nanobox` itself handles everything now, there's no need for this "bundle" Cask, as all it does is include vagrant and virtualbox. Those are now dependencies in the original cask, so the two are identical.

Check it out by looking at [`nanobox.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/nanobox.rb)

Please let me know if you have any more questions, thanks!

liljenstolpe:
I missed the dependencies. That looks good, thx.

PR #15319, opened 2015-11-20
Updated Mapbox Studio to 0.3.3

cedwardsmedia:
Mapbox Studio is now at version 0.3.3. Upon installing the current version via Cask, I discovered the discrepancy, so I updated the Cask to reflect the latest release.


adityadalal924:
Thank you for the contribution. It was merged directly as 642c7441dabd002342e98743945014e2d3c098bf to keep commit history cleaner. Your contribution is still credited to you.

PR #15358, opened 2015-11-22
Update sencha.rb

rednoah:
Update to Sencha ExtJS 6 and new install4j installer.


adityadalal924:
Why the change to no_jre version? Also, URL should use #{version} as needed. Please also add an explicit `:sudo` true/false as needed for the script stanzas.

rednoah:
Done.

* with JRE
* :sudo => yes

adityadalal924:
Thank you. Could you please also [squash your commits](https://davidwalsh.name/squash-commits-git) and post another comment here when that's done, and I'll go ahead and merge it. Your commit message can be something like: "updated sencha to 6.0.2.14"

rednoah:
https://github.com/caskroom/homebrew-cask/pull/15374

PR #15373, opened 2015-11-23
Add WordPress.com App

cedwardsmedia:
Added the WordPress.com app.

Note: WordPress's website for the app and the download link mark the version as 1.1 while the app itself reports 1.0.0 release.

See https://desktop.wordpress.com/

### DIFF
--- a/Casks/wordpresscom.rb
+++ b/Casks/wordpresscom.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'wordpresscom' do
+  version '1.1'
+  sha256 '2ee6503a319180895f7e75e4ff3ba06a695524e34152bdb5f6d6f39aa6c7c50c'
+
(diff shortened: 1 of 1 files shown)


Casks/wordpress.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'wordpress' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'wordpresscom' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `wordpresscom.rb`.

cedwardsmedia:
Done

adityadalal924:
Thank you for the contribution. It was merged directly as 5354d72113f7349608cca1b15dd09b392eda48bc to keep commit history cleaner. Your contribution is still credited to you.

PR #15387, opened 2015-11-24
Add Trello.App

russoedu:
Trello.app is a Trello web app created with Epichrome.


adityadalal924:
This is not an official app, and as such, can only go in `homebrew-unofficial`, if at all you wanted to add it.

PR #15411, opened 2015-11-25
Use last known good release for chromium instead of trunk

adidalal:
Resolves https://github.com/caskroom/homebrew-cask/issues/15357 - Chromium builds will now use LKGR instead of trunk

Thanks to @paulirish and @beaufortfrancois for the necessary changes on the `download-chromium.appspot.com` site.

### DIFF
--- a/Casks/chromium.rb
+++ b/Casks/chromium.rb
@@ -3,7 +3,7 @@ cask :v1 => 'chromium' do
   sha256 :no_check
 
   # appspot.com is the official download host per the vendor homepage
-  url 'https://download-chromium.appspot.com/dl/Mac'
+  url 'https://download-chromium.appspot.com/?platform=Mac&type=continuous'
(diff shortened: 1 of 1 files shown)


adityadalal924:
Merged as 89cdb64b68c923d72e7f634b71a47385aafb0251.

beaufortfrancois:
Yeah!

wfdd:
@adityadalal924: The URL is incorrect; it should be, I presume, `https://download-chromium.appspot.com/dl/Mac?type=continuous`.

adityadalal924:
@wfdd Thanks for the heads up, this is what happens when you forget to use `brew cask audit chromium.rb --download`....
Fixed in https://github.com/caskroom/homebrew-cask/pull/15413

paulirish:
Nice one. Thanks!

PR #15416, opened 2015-11-25
Attempt to properly arrange some items & fix a typo in the manpages

Amorymeltzer:
Noticed one minor typo, and that a few items were out of order (alphabetical, compared to the help menu).  Those should be fixed here.

I also regenerated the manpages, since that hadn't been done from the previous (admittedly minor) [change](https://github.com/caskroom/homebrew-cask/commit/4a6e2af25d54bb82ffe218cd0c88fa6cfd58c707#diff-7aada97c8658de7cabe6d4692705a994).  Happy to revert that if it'd rather be put off.

These will probably need to be changed in the near future.

### DIFF
--- a/doc/src/brew-cask.1.md
+++ b/doc/src/brew-cask.1.md
@@ -83,6 +83,13 @@ names, and other aspects of this manual are still subject to change.
     If <token> is given, summarize the staged files associated with the
     given Cask.
 
+  * `search` or `-S`:
+    Display all Casks available for install.
+
+  * `search` or `-S` <text> | /<regexp>/:
+    Perform a substring search of known Cask tokens for <text>. If the text
+    is delimited by slashes, it is interpreted as a Ruby regular expression.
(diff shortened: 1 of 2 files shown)


doc/src/brew-cask.1.md
```diff
@@ -98,6 +105,9 @@ names, and other aspects of this manual are still subject to change.
     `uninstall` without `--force` is also imperfect. It may be unable to
     perform an `uninstall` operation if the given Cask has changed since you
     installed it. This issue is being addressed.
+	
+  * `update`:
+    For convenience, `brew cask update` is a synonym for `brew update`.
```
vitorgalvao: Would you mind changing it to “For convenience. `brew cask update` is a synonym for `brew update`.”? The meaning changes just slightly, and becomes more accurate.

vitorgalvao:
Thank you.

PR #15417, opened 2015-11-26
Add Pritunl OpenVPN Client cask

jwillis:
Add cask for the Pritunl OpenVPN Client. http://client.pritunl.com/

### DIFF
--- a/Casks/pritunl.rb
+++ b/Casks/pritunl.rb
@@ -1,0 +1,21 @@
+cask :v1 => 'pritunl' do
+  version '0.10.14'
+  sha256 'afa302d8e95d1635584673a3c206d370c93846f519735d9898e75af43d9bd3a8'
+
+  # github.com is the official download host per the vendor homepage
+  url "https://github.com/pritunl/pritunl-client-electron/releases/download/#{version}/Pritunl.pkg.zip"
+  appcast 'https://github.com/pritunl/pritunl-client-electron/releases.atom'
+  name 'Pritunl OpenVPN Client'
+  homepage 'http://client.pritunl.com'
+  license :gpl
+
+  pkg 'Pritunl.pkg'
+
+  uninstall :pkgutil => 'com.pritunl.pkg.Pritunl'
+
+  zap :delete => [
+                   '~/Library/Application Support/pritunl',
+                   '~/Library/Caches/pritunl',
+                   '~/Library/Preferences/com.electron.pritunl.plist'
+                 ]
+end


Casks/pritunl-client.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'pritunl-client' do
+  version '0.10.14'
+  sha256 'afa302d8e95d1635584673a3c206d370c93846f519735d9898e75af43d9bd3a8'
+
+  url 'https://github.com/pritunl/pritunl-client-electron/releases/download/0.10.14/Pritunl.pkg.zip'
```
vitorgalvao: Please add a `appcast 'https://github.com/pritunl/pritunl-client-electron/releases.atom'` line under.

Casks/pritunl-client.rb
```diff
@@ -0,0 +1,12 @@
...
+  name 'Pritunl OpenVPN Client'
+  homepage 'http://pritunl.com'
+  license :oss
```
vitorgalvao: Can’t you be more specific?

Casks/pritunl-client.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'pritunl-client' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'pritunl' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `pritunl.rb`.

Casks/pritunl-client.rb
vitorgalvao: Write it as `url "https://github.com/pritunl/pritunl-client-electron/releases/download/#{version}/Pritunl.pkg.zip"` instead

Eitot:
There is a dedicated webpage for the client by the way: `http://client.pritunl.com`.

Can you add this zap stanza as well?

```
  zap :delete => [
                  '~/Library/Application Support/pritunl',
                  '~/Library/Caches/pritunl',
                  '~/Library/Preferences/com.electron.pritunl.plist'
                 ]
```

Casks/pritunl.rb
```diff
@@ -0,0 +1,20 @@
+cask :v1 => 'pritunl' do
+  version '0.10.14'
+  sha256 'afa302d8e95d1635584673a3c206d370c93846f519735d9898e75af43d9bd3a8'
+
+  url "https://github.com/pritunl/pritunl-client-electron/releases/download/#{version}/Pritunl.pkg.zip"
```
vitorgalvao: Needs a `# github.com is the official download host per the vendor homepage` comment above this line ([example](https://github.com/caskroom/homebrew-cask/blob/cb487ad405f74cb1f26b93e3f3e8412c82a22e34/Casks/0ad.rb#L5); [reasoning](https://github.com/caskroom/homebrew-cask/blob/7c010b42bd7ac679c1f002a0368b36bb3f495612/doc/CASK_LANGUAGE_REFERENCE.md#when-url-and-homepage-hostnames-differ-add-a-comment)).

vitorgalvao:
> Wouldn't it be better to use `http://client.pritunl.com` as the homepage instead?

On closer inspection, looks like it.

jwillis:
Yes, thank you, that's a better homepage.

vitorgalvao:
Still needs [the comment](https://github.com/caskroom/homebrew-cask/pull/15417/files#r46309914).

Also, could you please [squash your commits](http://davidwalsh.name/squash-commits-git)?

Casks/pritunl.rb
```diff
@@ -0,0 +1,21 @@
...
+  # github.com is the official download host per the vendor homepage
+  url "https://github.com/pritunl/pritunl-client-electron/releases/download/#{version}/Pritunl.pkg.zip"
+  appcast 'https://github.com/pritunl/pritunl-client-electron/releases.atom'
+  name 'Pritunl OpenVPN Client'
+  homepage 'http://client.pritunl.com'
+  license :gpl
+
+  pkg 'Pritunl.pkg'
+
+  uninstall :pkgutil => 'com.pritunl.*'
```
Eitot: Is it safe to use a wildcard here? On my system the package identifier is `com.pritunl.pkg.Pritunl`.

vitorgalvao: If there is no reason to use the wildcard (i.e. it doesn’t install a lot of related packages with similar but different identifiers), then we shouldn’t use it.

You would know best, though, since I do not use the app.

vitorgalvao:
Thank you for the contribution. It was merged directly as b193db1dc6b427722e6152f1e5659838bd742ac5 to keep commit history cleaner. Your contribution is still credited to you.

jwillis:
I agree on the more precise package name during the uninstall. Thanks for your guidance and patience!

vitorgalvao:
Thank you for the contribution, sticking around for the changes, and doing them so fast.

PR #15419, opened 2015-11-26
Update obs

tonini:
Hi,

There is an `0.12.2` patch [release](https://github.com/jp9000/obs-studio/releases/tag/0.12.2) available which fixes some issues.

### DIFF
--- a/Casks/obs.rb
+++ b/Casks/obs.rb
@@ -1,5 +1,5 @@
 cask :v1 => 'obs' do
-  version '0.12.0'
+  version '0.12.2'
   sha256 '701b8d172b914fb574a70a0e4db907474c9c29316b05a2f41b6cfae872ab0c43'
 
   # github.com is the official download host per the vendor homepage


tonini:
Currently there are no binary download for `0.12.2`. I'll get in contact with @jp9000 so we could them.

Casks/obs.rb
```diff
@@ -1,5 +1,5 @@
 cask :v1 => 'obs' do
-  version '0.12.0'
+  version '0.12.2'
   sha256 '701b8d172b914fb574a70a0e4db907474c9c29316b05a2f41b6cfae872ab0c43'
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

adityadalal924:
On hold until binaries for 0.12.2 are released.

vitorgalvao:
This has since been updated to `0.12.4`.

PR #15424, opened 2015-11-26
generic_unar.rb: use long flags

vitorgalvao:
@jawshooah (and other maintainers) We should try to use long flags to commands whenever possible. Short flags are fine when writing them out to run immediately, but have no place in a script, where comprehension is more important.

Do you agree?


vitorgalvao:
Merging it quickly so I can work on https://github.com/caskroom/homebrew-cask/issues/15025, but please voice your concerns, if any.

PR #15427, opened 2015-11-26
svnX 1.3.4: update sha256

reitermarkus:
https://github.com/caskroom/homebrew-versions/pull/1437


vitorgalvao:
Thank you, but no need to update those. The feature to handle `appcasts` isn’t built, yet.

reitermarkus:
I just thought I would update it, because I saw it changed because of the version 2 beta.

PR #15428, opened 2015-11-26
remove qtox

adidalal:
Ref https://github.com/caskroom/homebrew-cask/issues/15420


adityadalal924:
Merged as 43bcf9fa32a4fbb367dcd1b537953c99484567af.

PR #15434, opened 2015-11-27
Set world readable permissions on Blender so that it can be used system wide.

SushiDude:
The directory in the Blender ZIP has its permissions set to 0700. This means that `brew cask install --appdir='/Applications' blender` will not have the desired effect of properly installing Blender for every user on the system.

### DIFF
--- a/Casks/blender.rb
+++ b/Casks/blender.rb
@@ -9,4 +9,8 @@ cask :v1 => 'blender' do
 
   app "blender-#{version}-OSX_10.6-x86_64/Blender.app"
   app "blender-#{version}-OSX_10.6-x86_64/BlenderPlayer.app"
+
+  postflight do
+    set_permissions "#{staged_path}/blender-#{version}-OSX_10.6-x86_64/", '0755'
+  end
 end


vitorgalvao:
Thank you, but this should be addressed with https://github.com/caskroom/homebrew-cask/issues/13201. We should refrain from adding workarounds to issues that will be solved by that (since stop building workarounds is the whole point of it).

SushiDude:
#13201 is not implemented yet, so would it not make more sense to add this workaround now and then remove it later?

vitorgalvao:
No, because no one will remember (no one ever does) to remove the workarounds.

Again, the point of that issue is to eliminate workarounds. Now that there is a plan, every workaround that will be solved by the implementation of that issue is refused.

adityadalal924:
@SushiDude Remember, you can always use this workaround in your own personal tap!

SushiDude:
It seems wrong to leave issues unfixed because of a proposal that has yet to have been implemented and lacks any time line.

Why not tag these types of workarounds with `TODO: Remove after #13201`?

The alternative just leaves things, that have fixes, in a broken state with the promise that they will be fixed.

vitorgalvao:
> Why not tag these types of workarounds with `TODO: Remove after #13201`?

Because those are left forgotten as well.

Has you have been told before, we’re pre-alpha, and should be treated as such. A bunch of stuff is broken, and a bunch more stuff is *severely* disorganised. That disorganisation holds up a bunch of stuff and is very detrimental to moving things forward.

You’ve made your point clear before that everything is important, and I made it clear as well that no, not everything is important, and has varying degrees of importance. You’re again arguing for a minor inconvenience, and this time it is so minor, it’ll affect a really small minority. It is so minor, it is likely there from the start and no one was ever bothered by it. Yes, it should be fixed, but not if it’ll make other things harder to manage *and be removed any way*, and right now, that is exactly what it is.

**Putting temporary band-aids on minor inconveniences is not the way to go right now. We need to fix the core of the problems.**

> The alternative just leaves things, that have fixes, in a broken state with the promise that they will be fixed.

Why does the directory have those permissions, have you even tried to find that out? Is that a deliberate decision by the upstream developers, or was it a mistake that could be fixed there perhaps by asking them, *therefore fixing the issue here and everywhere else, instead of just having a workaround here*?

> It seems wrong to leave issues unfixed because of a proposal that has yet to have been implemented and lacks any time line.

Helps us implement it, then, or help us find people who can do it and help review the progress so far.

Don’t take this the wrong way, but since (once again) you’ve been clearly explained the reasoning behind a point and this is (once more) incredibly minor (even more than last time) and are nonetheless continuing to argue for it even after being explained over and over the position and reasonings, I’m going to lock the issue, since I do not intend to waste a bunch of time (yet again) arguing this, as there is absolutely no point.

Yes, I agree this should be fixed. No, this is not the way to do it. We either fix it at our core (#13201) or at blender’s core (ask them). No workarounds when there is a clear outline solution in sight. Workarounds got us into the current mess we have, in the first place, lets not make it worse.

PR #15441, opened 2015-11-27
Add final three apps in the Laurence Anthony 'ant' suite

Amorymeltzer:
This adds three new casks, antfileconverter, protant, and antpconc, which should complete the suite of mac apps from http://www.laurenceanthony.net/software/ — antconc, antwordprofiler, segmentant, and tagant were already there.


adityadalal924:
Thank you for the contribution. It was merged directly as b7bfa38b6ae983316b093566a90e2abf0060ef15 to keep commit history cleaner. Your contribution is still credited to you.

PR #15451, opened 2015-11-28
Upgrade JDownloader to v2

tsparber:
With regard to https://github.com/caskroom/homebrew-versions/pull/1438 this version does not create dock/desktop icons anymore (Escaping of $ was needed) and also mutes all stdout and stderr.

One thing I noticed is, that even when the system call fails (does not return 0) the app is linked (JDownloader2.app already exists but the installer failed due to no internet connection).
Is there a way to make the `preflight` block fail if the return value is not 0?

Install4j: [Command Line Options For Generated Installers](http://resources.ej-technologies.com/install4j/help/doc/helptopics/installers/options.html)

(Based on https://github.com/winkelsdorf/homebrew-custom/commit/7798474dd1629d6492b8be2afb5738ebd806855f by @winkelsdorf, https://github.com/svenjacobs/homebrew-cask/commit/8d627e5dd8fd75201214f1fe61714093695de1d5 by @svenjacobs and https://github.com/caskroom/homebrew-versions/commit/96e9e191fc128cd27559c8ab7ae4f4029a53ce59)

### DIFF
--- a/Casks/jdownloader.rb
+++ b/Casks/jdownloader.rb
@@ -2,13 +2,26 @@ cask :v1 => 'jdownloader' do
   version :latest
   sha256 :no_check
 
-  url 'http://installer.jdownloader.org/JDownloader09Setup_mac.dmg'
-  name 'JDownloader'
+  if MacOS.release <= :snow_leopard
+    url 'http://installer.jdownloader.org/JD2Setup_10_6orlower.dmg'
+  else
+    url 'http://installer.jdownloader.org/JD2Setup.dmg'
+  end
+
+  name 'JDownloader 2'
   homepage 'http://jdownloader.org/'
   license :gpl
 
-  installer :script => 'JDownloader Installer.app/Contents/MacOS/JavaApplicationStub',
-            :args => [ '-q' ]
+  preflight do
+    system "\"#{staged_path}/JDownloader Installer.app/Contents/MacOS/JavaApplicationStub\" " \
+           "-dir \"#{staged_path}\" " \
+           '-q ' \
+           '-Dinstall4j.suppressStdout=true ' \
+           '-Dinstall4j.debug=false ' \
+           '-VcreateDesktopLinkAction\$Boolean=false ' \
+           '-VaddToDockAction\$Boolean=false ' \
+           '> /dev/null 2>&1'
+  end
 
   caveats <<-EOS.undent
     #{token} requires Java 6+, you can install the latest Java using
@@ -16,5 +29,5 @@ cask :v1 => 'jdownloader' do
       brew cask install java
   EOS
 
-  uninstall :delete => '/Applications/JDownloader.app'
+  app 'JDownloader 2.0/JDownloader2.app'
 end


Casks/jdownloader.rb
```diff
@@ -2,13 +2,26 @@
...
+
+  name 'JDownloader 2'
   homepage 'http://jdownloader.org/'
   license :gpl
 
-  installer :script => 'JDownloader Installer.app/Contents/MacOS/JavaApplicationStub',
-            :args => [ '-q' ]
+  preflight do
+    system "\"#{staged_path}/JDownloader Installer.app/Contents/MacOS/JavaApplicationStub\" " \
+           "-dir \"#{staged_path}\" " \
```
jawshooah: No need to interpolate `staged_path` here, just use `'-dir', staged_path`. `system` should take care of escaping spaces.

tsparber: To redirect the stdout/stderr to /dev/null I needed to pass the whole string as the cmd and let the shell parse the arguments [[1]](http://stackoverflow.com/a/13338316/3433642). Should I use another method [[2]](http://mentalized.net/journal/2010/03/08/5-ways-to-run-commands-from-ruby/)?

jawshooah: Ah, I missed that. Just ran the install without redirecting output, and it is indeed extremely noisy. An alternative to your method, though, might be this:

```ruby
system "\"#{staged_path}/JDownloader Installer.app/Contents/MacOS/JavaApplicationStub\" > /dev/null 2>&1",
       '-dir', staged_path,
       '-q',
       '-Dinstall4j.suppressStdout=true',
       '-Dinstall4j.debug=false',
       '-VcreateDesktopLinkAction$Boolean=false',
       '-VaddToDockAction$Boolean=false'
```

jawshooah: Hmm, that doesn't seem to work actually. Not sure why. In any case, I don't care much either way.

vitorgalvao:
Thank you for the contribution. It was merged directly as 9958c742c42adf6ef0e65fb12f121b2b59a61294 to keep commit history cleaner. Your contribution is still credited to you.

winkelsdorf:
@vitorgalvao Thanks for merging, I can confirm it is working as expected.

@tsparber Great work, Tommy, really! I didn't consider encapsulating the Install4j options, because using the Args was the recommended way. Had several talks about the very noisy Install4j with the manufacturer of JDownloader but they were unable to help.

Now I can purge my custom versions repo :+1: 

Just one thing to consider: The settings of JDownloader are stored together with the Application in the ```/opt/homebrew-cask/Caskroom/jdownloader/latest/``` folder. If the Cask is updated for some reason, the settings will be lost. Do you think there is a way to separate the data from the App?

vitorgalvao:
> Just one thing to consider: The settings of JDownloader are stored together with the Application in the ```/opt/homebrew-cask/Caskroom/jdownloader/latest/``` folder. If the Cask is updated for some reason, the settings will be lost. Do you think there is a way to separate the data from the App?

That’s a pain. Is that the way JDownloader itself operates, or is it a consequence of how we’re installing it here?

winkelsdorf:
> That’s a pain. Is that the way JDownloader itself operates, or is it a consequence of how we’re installing it here?

Not our fault: That's from the way JDownloader works itself. Those are Java Developers trying to make it ultra portable, by breaking all the best practice rules any OS has. Installed without Cask it put's all files, Application and Data, under a ```~/JDownloader 2.0/```folder, which is a mess too as the settings are not shareable for different User sessions.

The whole configuration seems to be stored in this folder:
```/opt/homebrew-cask/Caskroom/jdownloader/latest/JDownloader 2.0/cfg```

Two ideas: 
- Probably there is a way to keep this folder upon Update and move it to the new folder (if target is anything else than ```latest```)?
- Leave it as it is as we don't set a specified Version folder. As long as everything is stored in ```latest```it's working fine as it is now.

Personally I don't think it's worth the work taking care of now. If the Cask Upgrade mentioned in the Roadmap is going productive this just might be something to check again. I'll try to keep an eye on this.

vitorgalvao:
> Those are Java Developers trying to make it ultra portable, by breaking all the best practice rules any OS has.

“Write once, run with caveats everywhere”. I’d expect there were some kind of (perhaps even default) library to handle this. Something where you’d say “save this as a preference” and it’d figure out where that is based on the OS.

> Leave it as it is as we don't set a specified Version folder. As long as everything is stored in `latest` it's working fine as it is now.

Agreed. It’s unlikely that’ll change.

> Personally I don't think it's worth the work taking care of now. If the Cask Upgrade mentioned in the Roadmap is going productive this just might be something to check again.

Agreed again.

tsparber:
One last question: Is it possible to make the `preflight` fail (and thus `brew cask install`) ?
(Raising an exception?, Return value?)

(The installer returns a non-zero exitcode if the installation fails)

winkelsdorf:
> “Write once, run with caveats everywhere”. I’d expect there were some kind of (perhaps even default) library to handle this. Something where you’d say “save this as a preference” and it’d figure out where that is based on the OS.

Right. I started to dislike the non-native, "embedded Java"-using Applications for OSX a long while ago. I am a fan of code reuse and portability (a.k.a. cross-compilation) but those slow Applications, with Swing GUIs, partly with old embedded JREs, not having native controls are just a pain. JDownloader does not even support window minimize by ⌘-M. Makes me remember the days were every Java Application had this "cool" Motif UI.

> One last question: Is it possible to make the preflight fail (and thus brew cask install) ?
> (Raising an exception?, Return value?)
> 
> (The installer returns a non-zero exitcode if the installation fails)

Iirc, if the installer returns a non-zero exitcode on error, it should already fail. But better wait for the official confirmation from @vitorgalvao.

vitorgalvao:
Yes, it’s supposed to fail, in that case.

tsparber:
I checked the installation again. I think it is ok how it works. Sorry for the confusion.

If the installer fails before creating the .app the cask is not installed and cleaned up.
If the .app is already created but the installer fails (returns non-zero) to download the updates (= Rest of the application), the cask installation succeeds. The user is then prompted to download everything on the first run.

So even if the full installation fails there is a functional linked app (which needs downloads on first run).

> Yes, it’s supposed to fail, in that case.

Just as a note: The `preflight` block does not fail if the [system](http://ruby-doc.org/core-2.2.0/Kernel.html#method-i-system) method has a non-zero return value. How should it know?

vitorgalvao:
Perhaps I’m mistaken or there is a bug, but it was supposed to.

PR #15453, opened 2015-11-28
Renamed Mapbox Studio to Mapbox Studio Classic

cedwardsmedia:
The developers of Mapbox now refer to the desktop app as Mapbox Studio Classic and the web-based version as Mapbox Studio. To avoid confusion and to adhere to the Vendor's naming, I have updated this cask to reflect the new changes.

Worth nothing, despite the changes to the name, the Vendor continues to name the app bundle "Mapbox Studio.app", so this remains unchanged in the cask until/unless the Vendor changes it in the future.

### DIFF
--- a/Casks/mapbox-studio-classic.rb
+++ b/Casks/mapbox-studio-classic.rb
@@ -1,12 +1,13 @@
-cask :v1 => 'mapbox-studio' do
+cask :v1 => 'mapbox-studio-classic' do
   version '0.3.3'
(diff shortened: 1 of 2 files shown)


vitorgalvao:
Thank you, but the vendor’s naming is secondary. We adhere to a strict [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) to avoid duplicates. Without it, hombrew-cask becomes a mess (we know, we’ve been there).

PR #15469, opened 2015-11-29
Add alternative AntConc version for snow leopard or lower

Amorymeltzer:
See http://www.laurenceanthony.net/software/antconc/


adityadalal924:
Thank you for the contribution. It was merged directly as a2c90efc128fadc60d8e8da499e052bc72e7cae3 to keep commit history cleaner. Your contribution is still credited to you.

PR #15470, opened 2015-11-29
Upgrade KiCad to 4.0.0-RC2-1

gibsjose:
- Upgrade KiCad to 4.0.0-RC2-1
- Add KiCad Extras (4.0.0-RC2-1)
- Remove old formula `kicad-library`, which is outdated

The `kicad-library` cask was over a year old. Also, it seems KiCad has changed their nomenclature, and now calls it `kicad-extras`.

KiCad was updated to 4.0.0 RC2-1, which is the latest stable version for OS X. The `kicad-extras` cask was also added, and is also version 4.0.0 RC2-1.

Finally, these casks now use the official KiCad website as the host for both the application image and the extras image.


gibsjose:
It may be that the `kicad-library` package is not the same as the `kicad-extras`, but there is no mention of `kicad-library` on the official downloads page, and the `kicad-library` cask is very outdated, regardless.

Casks/kicad.rb
```diff
@@ -1,9 +1,9 @@
 cask :v1 => 'kicad' do
-  version '2014-02-26'
-  sha256 '066520f10e9646e88c2e1d7812adf54a73a331592076f394cb439a97c228410b'
+  version '4.0.0-rc2-1'
+  sha256 '32ff44659380aada7ad74d1d3aba27da964d92479c5fd290bf81f9abd27b18b4'
 
-  # mdx4.org is the official download host per the vendor homepage
-  url "http://www.mdx4.org/uploads/kicad/Kicad-product-#{version}.zip"
+  # kicad-pcb.org, which is the homepage, now hosts the official distributions
```
vitorgalvao: This comment must be removed.

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'kicad-extras' do
+  version '4.0.0-rc2-1'
+  sha256 '3049e936dce4510f3bc85dc57a466742e0de9882a2f3b4a63144f61d33b3cf4b'
+
+  # kicad-pcb.org, which is the homepage, now hosts the official distributions and extras
```
vitorgalvao: Comment must go.

gibsjose:
@vitorgalvao I'm actually going to close this and open a new one because they just released 4.0.0 stable (this one was RC2-1) today haha :smile:

PR #15505, opened 2015-11-30
Update PowerPhotos to 1.1.12

toonetown:


adityadalal924:
It seems that the software is actually on version 1.1.2, and I've updated the Cask at https://github.com/caskroom/homebrew-cask/pull/15507

Thank you for the contribution, regardless!

toonetown:
That is correct - I misread the version number.  I apologize.  Thanks!

PR #15518, opened 2015-12-01
Update Reveal to version 1.6.2

mauricerkelly:
This commit changes the Reveal cask to use a specific version instead
of the `:latest` token.


adityadalal924:
Thank you for the contribution. It was merged directly as f4666dfc95aab04cc9cb700bd85e17d2bf9c9ba3 to keep commit history cleaner. Your contribution is still credited to you.

PR #15530, opened 2015-12-01
Add Pullover 0.3.2

amyreese:
Add new cask definition for Pullover based on contributing docs.

### DIFF
--- a/Casks/pullover.rb
+++ b/Casks/pullover.rb
@@ -1,0 +1,12 @@
+cask :v1 => 'pullover' do
+  version '0.3.2'
+  sha256 'a4208c33c2fbc6bca96c08c9085d90fbd4caa484cc7d2206db0a456f975f0052'
+
+  # sourceforge.net is the official download host per the vendor homepage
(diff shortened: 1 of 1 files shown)


Casks/pullover.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'pullover' do
+  version '0.3.2'
+  sha256 'a4208c33c2fbc6bca96c08c9085d90fbd4caa484cc7d2206db0a456f975f0052'
+
+  url 'https://downloads.sourceforge.net/sourceforge/pullover/Pullover_0.3.2.dmg'
```
adityadalal924: Needs a comment above url -> `# sourceforge.net is the official download host per the vendor homepage` and use string interpolation for URL (use `#{version}` and ")

jreese:
Updated.

adityadalal924:
Thank you for the contribution. It was merged directly as 6cc48598e72655048664c5b609c348b6035528e0 to keep commit history cleaner. Your contribution is still credited to you.

PR #15533, opened 2015-12-02
long update command now uses ';' instead of '&&'

vitorgalvao:
As per https://github.com/caskroom/homebrew-cask/issues/15532.
On hold until https://github.com/caskroom/homebrew-cask/pull/15381.

Decided to go ahead and remove the `upgrade` part as well, since [that will stop being needed](https://github.com/caskroom/homebrew-cask/pull/15381).

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ When reporting bugs, remember that homebrew-cask is an independent project from 
 Before reporting a bug, make sure you have the latest versions of homebrew, homebrew-cask, and all Taps by running the following command:
 
 ```bash
-$ brew update && brew upgrade brew-cask && brew cleanup && brew cask cleanup
+$ brew update ; brew cleanup ; brew cask cleanup
 ```
 
 If the issue persists, please run the problematic command with the `--verbose` flag and post its and `brew cask doctor`’s outputs in distinct [fenced code blocks](https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks).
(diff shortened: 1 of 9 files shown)


Amorymeltzer:
:+1:

jawshooah:
I don't think this is necessary once we remove `brew upgrade brew-cask`. It's probably a good thing to halt early if `brew update` fails for some reason.

vitorgalvao:
Normally I’d agree, but this command is the “fix all” we give users, and every part of it should indeed run (there’s no harm in the later commands running if the earlier commands fail).

For a recent example, [a user had an issue that look related to a cask](https://github.com/caskroom/homebrew-cask/issues/15512), but [later was revealed to be likely related to something wrong with a downloaded/cached version](https://github.com/caskroom/homebrew-cask/pull/15538#issuecomment-161274597).

Had the `&&` command be used to try and resolve this, the problem would have remained (as https://github.com/caskroom/homebrew-cask/issues/15532 reports), but with the `;` command, it was indeed solved.

Apart from `brew update && brew upgrade` do make sense being dependent, but the last ones don’t, since the purpose os this command is to be the “magical” fix for most issues.

---

As a tangential reference a small piece of HBC history, this isn’t the first time we had a [fix-most-issues-command](https://github.com/caskroom/homebrew-cask/issues/47#issuecomment-9360755) we could [call “magical”](https://github.com/caskroom/homebrew-cask/issues/917#issuecomment-22636954)

adityadalal924:
Needs merge conflicts resolved but no longer on hold.

jawshooah:
Merged in 5e79596ef7be5a273ed1d2eea507563cc252d7a8.

vitorgalvao:
Thank you. Was just about to start fixing the conflicts.

PR #15538, opened 2015-12-02
Change folder holder app file from latest to version 1.5.2

ninjahoahong:
I solved the problem by changing the name of the folder holding app file, and I successfully installed the app by using brew-cask. Please check it if there are anything to fix. Thank you.

### DIFF
--- a/Casks/firealpaca.rb
+++ b/Casks/firealpaca.rb
@@ -1,5 +1,5 @@
 cask :v1 => 'firealpaca' do
-  version :latest
+  version '1.5.2'
   sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
This change won’t do anything that could fix your issue with the cask. I’ve just tested it, and the cask worked fine. You probably had some broken/interrupted download.

Try `brew update ; brew upgrade brew cask ; brew cleanup ; brew cask cleanup ; brew cask install firealpaca`, and it should work,

ninjahoahong:
Thank Vítor. Your solution works.

PR #15549, opened 2015-12-02
install correct version of quicksilver on mountain lion

### LINKED ISSUE #15548: quicksilver installs wrong version on mountain lion
From [the homepage](https://qsapp.com/download.php):
> OS X 10.11 El Capitan, OS X 10.10 Yosemite, OS X 10.9 Mavericks
> Download 1.3.4
>
> OS X 10.8 Mountain Lion, OS X 10.7 Lion
> Download 1.2.0
>
> OS X 10.6 Snow Leopard
> Download 1.0.0

But the [logic to determine which version of quicksilver to install on which version of OSX](https://github.com/caskroom/homebrew-cask/blob/722757a038ac1eb03240a70d0c8544a3fcf1aece/Casks/quicksilver.rb#L2-L11) is flawed.

This means that on Mountain Lion, the cask installs the newest version, instead of the correct one for that OSX version (1.2.0). Took me several minutes to realize why a fresh install was crashing on startup.

psibre:
fixes #15548

### DIFF
--- a/Casks/quicksilver.rb
+++ b/Casks/quicksilver.rb
@@ -2,7 +2,7 @@ cask :v1 => 'quicksilver' do
   if MacOS.release <= :snow_leopard
     version '1.0.0-16384'
     sha256 '0afb16445d12d7dd641aa8b2694056e319d23f785910a8c7c7de56219db6853c'
-  elsif MacOS.release <= :lion
+  elsif MacOS.release <= :mountain_lion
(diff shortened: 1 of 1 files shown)


adityadalal924:
Thank you for the contribution. It was merged directly as 3bc101247a312e34d80fa92aa3f10329a2681e4a to keep commit history cleaner. Your contribution is still credited to you.

PR #15550, opened 2015-12-02
update warsow to 20

lourinaldi:


adityadalal924:
Should be version 2.0

You can use `#{version.delete('.')}` to make the URL work

PR #15563, opened 2015-12-03
Fix urls in Deezer.app Cask

craigrosie:
The old url pointed to an outdated https Deezer domain for which the
SSL certificate has expired. This resulted in the following error
when trying to run brew cask install deezer:

`curl: (60) SSL certificate problem: Invalid certificate chain
More details here: http://curl.haxx.se/docs/sslcerts.html`

This commit updates the url to reflect the url used when downloading
via a browser. The homepage url has also been updated as it
doesn't use https.

### DIFF
--- a/Casks/deezer.rb
+++ b/Casks/deezer.rb
@@ -2,9 +2,9 @@ cask :v1 => 'deezer' do
   version '1.1_4191'
   sha256 'f9d491fb8d4b055a60b3d4a13a4e8b19e4b9b8d70dc8740df734afeee5482a34'
 
-  url "https://cdns-content.deezer.com/builds/mac/Deezer_#{version.sub(%r{^[^_]*_(\d+)},'\1')}.dmg"
+  url "http://e-cdn-content.deezer.com/builds/mac/Deezer_#{version.sub(%r{^[^_]*_(\d+)},'\1')}.dmg"
   name 'Deezer'
-  homepage 'https://www.deezer.com/formac'
+  homepage 'http://www.deezer.com/formac'
   license :gratis
 
   app 'Deezer.app'


adityadalal924:
Deezer is unavailable in the US, so I shall wait for one of the other maintainers to check the URL. That regex looks really unnecessarily complicated.
Thanks.

Pinging @vitorgalvao

Casks/deezer.rb
```diff
@@ -2,9 +2,9 @@
   version '1.1_4191'
   sha256 'f9d491fb8d4b055a60b3d4a13a4e8b19e4b9b8d70dc8740df734afeee5482a34'
 
-  url "https://cdns-content.deezer.com/builds/mac/Deezer_#{version.sub(%r{^[^_]*_(\d+)},'\1')}.dmg"
+  url "http://e-cdn-content.deezer.com/builds/mac/Deezer_#{version.sub(%r{^[^_]*_(\d+)},'\1')}.dmg"
   name 'Deezer'
-  homepage 'https://www.deezer.com/formac'
+  homepage 'http://www.deezer.com/formac'
```
vitorgalvao: Should remains `https`.

Casks/deezer.rb
```diff
@@ -2,9 +2,9 @@
   version '1.1_4191'
   sha256 'f9d491fb8d4b055a60b3d4a13a4e8b19e4b9b8d70dc8740df734afeee5482a34'
 
-  url "https://cdns-content.deezer.com/builds/mac/Deezer_#{version.sub(%r{^[^_]*_(\d+)},'\1')}.dmg"
+  url "http://e-cdn-content.deezer.com/builds/mac/Deezer_#{version.sub(%r{^[^_]*_(\d+)},'\1')}.dmg"
```
vitorgalvao: Not sure why they stopped offering it via `https`, but does indeed seem to be the case.

vitorgalvao:
> wait for one of the other maintainers to check the URL.

Same domain as the homepage, so not strictly necessary. I confirmed it is indeed correct, though.

> That regex looks really unnecessarily complicated.

It is a bit, yes. Will simplify.

vitorgalvao:
Also, the homepage does indeed still work via `https` (and trying `http` will actually redirect there).

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 1e7543d99657a980f22993d406f0ee36e5eeaa74. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15564, opened 2015-12-03
Update CONTRIBUTING.md

adidalal:
Change examples to show versioned URLs more prominently and add in further explanation about squashing commits.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ Making a Cask is easy: a Cask is a small Ruby file.
 
 ### Examples
 
-Here’s a Cask for `Shuttle.app` as an example. Note that the `url` stanza uses `#{version}` (string interpolation) to create a Cask that only needs `version` and `sha256` updated with an application update.
+Here’s a Cask for `shuttle` as an example. Note that the `url` stanza uses `#{version}` ([string interpolation](https://en.wikipedia.org/wiki/String_interpolation#Ruby)) to create a Cask that only needs `version` and `sha256` changes when updated.
 
 ```ruby
 cask :v1 => 'shuttle' do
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -28,39 +28,46 @@ Making a Cask is easy: a Cask is a small Ruby file.
 
 ### Examples
 
-Here’s a Cask for `Alfred.app` as an example. Note that you may repeat the `app` stanza as many times as you need, to define multiple apps:
+Here’s a Cask for `Shuttle.app` as an example. Note that the `url` stanza uses `#{version}` (string interpolation) to create a Cask that only needs `version` and `sha256` updated with an application update.
```
vitorgalvao: Suggestion so we don’t repeat “update”: “that only needs `version` and `sha256` changes when updated.”

vitorgalvao: I’d also suggest for “string interpolation” to be a link to https://en.wikipedia.org/wiki/String_interpolation#Ruby.

vitorgalvao: Also, `shuttle`, instead of `Shuttle.app`, to be consistent with the next example.

CONTRIBUTING.md
```diff
@@ -381,7 +388,7 @@ Congratulations! You are done now, and your Cask should be pulled in or otherwis
 
 ### Squashing
 
-If your pull request has multiple commits which revise the same lines of code, it is better to [squash](http://davidwalsh.name/squash-commits-git) those commits together into one logical unit.
+If your pull request has multiple commits which revise the same lines of code, or if you make some changes after comments from one of the maintainers, it is better to [squash](http://davidwalsh.name/squash-commits-git) those commits together into one logical unit.
```
vitorgalvao: Big thumb up to making it explicit here.

adityadalal924:
Noticed that firefox isn't actually unversioned anymore, so I changed it to another example with an appropriate url (and one with a comment, just so that there's a clear example on the page)

CONTRIBUTING.md
```diff
@@ -371,7 +381,7 @@ $ github_user='<my-github-username>'
 $ git push "$github_user" my-new-cask
 ```
 
-If you are using [GitHub two-factor authentication](https://github.com/blog/1614-two-factor-authentication) and set your remote repository as HTTPS you will need to set up a personal access token and use that instead your password. See more on https://help.github.com/articles/https-cloning-errors#provide-access-token-if-2fa-enabled
+If you are using [GitHub two-factor authentication](https://help.github.com/articles/about-two-factor-authentication/) and set your remote repository as HTTPS you will need to set up a personal access token and use that instead your password. See more on [here](https://help.github.com/articles/https-cloning-errors/#provide-access-token-if-2fa-enabled).
```
vitorgalvao: “access token and use that instead **of** your password”.

vitorgalvao: “See more here” (no “on”).

adityadalal924:
Merged as 41aa133fe24fd7705dd71626a7506dab4db2c3bf.

PR #15569, opened 2015-12-04
Update Cask create scripts to emphasize versioned URLs

proinsias:
Fix #15564 (https://github.com/caskroom/homebrew-cask/pull/15564)
changed CONTRIBUTING.md to emphasize the use of versioned
URLs. Updated Cask create scripts to use double-quotes for urls by
default, instead of single-quotes - these are needed for versioned
URLs. Also included a comment recommending the use of versioned URLs.

### DIFF
--- a/lib/hbc/cli/create.rb
+++ b/lib/hbc/cli/create.rb
@@ -29,7 +29,7 @@ class Hbc::CLI::Create < Hbc::CLI::Base
         version ''
         sha256 ''
 
-        url 'https://'
+        url "https://"      # todo: if applicable, use \#{version} rather than hardcoding version number
         name ''
         homepage ''
         license :unknown    # todo: change license and remove this comment; ':unknown' is a machine-generated placeholder
--- a/test/cask/cli/create_test.rb
+++ b/test/cask/cli/create_test.rb
@@ -40,7 +40,7 @@ describe Hbc::CLI::Create do
         version ''
         sha256 ''
 
-        url 'https://'
+        url "https://"      # todo: if applicable, use \#{version} rather than hardcoding version number
         name ''
         homepage ''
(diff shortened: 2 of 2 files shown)


adityadalal924:
This change doesn't seem major, but I'm not sure if pushing for a versioned URL via a comment is the best option, especially as it is already emphasized strongly in CONTRIBUTING.md now

cc: @vitorgalvao

vitorgalvao:
Though the initiative is appreciated, experience tells us this won’t work as expected, for various reasons. All of these are backed by observations and conversations with users in all my time with the project.

Though it is fine to recommend interpolation, adding it like this to the cask creation scripts I see as a barrier. Many users don‘t understand interpolation, and many others are already afraid/shy to submit changes, even though casks are so simple. Adding something some won’t understand makes it even scarier. It’s way better to “get them in the door” and then introduce interpolation.

This is promoting double quotes, and those are discouraged. Double quotes are only to be used in cases of needed string interpolation, and not every url can use them. In fact, our own tests catch that. Many users still change single quotes to double quotes even when those aren’t needed, and we shoudn’t encourage that.

The comment is pretty much useless. See the comment in `license`? The amount of users who submit a new cask or an update to one that changes the license but still leaves the comment is surprising *and the comment tells you to remove it*. Other comments that say you should do things (like “when updating this cask, this other should be updated as well”) are ignore most of the time (not “many”, but “most”).

Comments are useful for us maintainers and for the users who need them when auditing casks, but comments to be read at update/submission time have shown to be close to worthless. With this change, we’ll start getting the same types of submissions, only they’ll now have that comment attached, and fail audit due to having double quotes when they aren’t needed.

Again, thank you for the initiative, but we have enough data to know it won’t work as you expect.

PR #15582, opened 2015-12-04
Add Saleae Logic Analyzer v1.2.5.

rbdixon:
Support software for the Saleae Logic Analyzer.

### DIFF
--- a/Casks/saleae.rb
+++ b/Casks/saleae.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'saleae' do
+  version '1.2.5'
+  sha256 '6699a0da5b14c85da662d2df3b8c6453459a49c1be7e833854bbb000bfce9288'
+
+  url 'http://downloads.saleae.com/betas/1.2.5/Logic-1.2.5-Darwin.dmg'
+  name 'Saleae Logic'
+  homepage 'http://saleae.com'
+  license :closed
+
+  app 'Logic.app'
+end


Casks/saleae.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'saleae' do
+  version '1.2.5'
+  sha256 '6699a0da5b14c85da662d2df3b8c6453459a49c1be7e833854bbb000bfce9288'
+
+  url 'http://downloads.saleae.com/betas/1.2.5/Logic-1.2.5-Darwin.dmg'
```
adityadalal924: Please use `#{version}`. https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#examples

Casks/saleae.rb
```diff
@@ -0,0 +1,11 @@
...
+  name 'Saleae Logic'
+  homepage 'http://saleae.com'
+  license :closed
+
+  app 'Logic.app'
```
adityadalal924: The cask token reference must be followed

vitorgalvao: As @adityadalal924 said, the token reference needs to be followed.

That is so we can avoid duplicates, which is what this would be, if we didn’t have such rules. See [`logic.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/logic.rb).

rbdixon:
Well... its irritating that the app itself is named the ambiguous "Logic"! I'll update the existing formula to the newest version. Thanks.

PR #15596, opened 2015-12-04
Added anaconda3.rb

cclauss:
Anaconda for Python 3.5.0

@adityadalal924 Does it look OK to you?


adityadalal924:
needs to be `anaconda3.rb` (you forgot the extension).

Everything else looks OK (havent verified shasum, as CI will do that once you fix and squash commits)

vitorgalvao:
Actually, should just be `anaconda.rb`, the number isn’t part of the name, on the site.

The one currently here should be moved to [caskroom/versions](https://github.com/caskroom/homebrew-versions).

cclauss:
* Anaconda2 is for Python2.7.10
* Anaconda3 is for Python3.5.0

adityadalal924:
@vitorgalvao They have unique versions for python 2 and 3, so like we have python2 and 3 both in the main repo, same same logic follows for anaconda2 and anaconda3

vitorgalvao:
Yes, I understand that. They’re still different versions of the same app, just like basic and pro, or enterprise and community.

Keeping in line with the new organisation rules we’re trying to introduce, it still makes sense to make this `anaconda` and move the other one to [caskroom/versions](https://github.com/caskroom/homebrew-versions).

Similarly to Python versions, we should remove version 2 from this repo.

vitorgalvao:
The number *isn’t* part of the name. A number that is part of the name is [`4k-youtube-to-mp3`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/4k-youtube-to-mp3.rb), for example. Here the number is an identifier of a version. The version of a different piece of software, but a version nonetheless. They still unequivocally call this “Anaconda” on the website, not Anaconda2 and Anaconda3.

cclauss:
Both files are v2.4.0.  Both are the Forever Free Edition and there are SEPARATE Pro, Workgroup, and Enterprise  versions.  https://www.continuum.io/content/anaconda-subscriptions

vitorgalvao:
Precisely, you just made my point. They’re *alternate* versions of the same piece of software, just like enterprise and commercial editions. As such, one of these belongs in the [caskroom/versions](https://github.com/caskroom/homebrew-versions) repo.

cclauss:
bash Anaconda2-2.4.0-MacOSX-x86_64.sh
bash Anaconda3-2.4.0-MacOSX-x86_64.sh

vitorgalvao:
**Yes, I know that, stop repeating yourself**. I understand how they are structured. do you understand the rules I’m explaining to you?

cclauss:
As a user of the system, I would never have guessed that.  Outside of caskroom they are considered two different things that have two different names.  I do not find it intuitive but it is your repo,  You make the rules.

adityadalal924:
@cclauss I think in this particular case, Anaconda can safely be considered one thing, on their website:

Because Anaconda includes installers for Python 2.7 and 3.5, either is fine. You can create a 3.5 environment with the conda command if you've downloaded 2.7 — and vice versa.

vitorgalvao:
You’re welcome to challenge the rules, they’re all open to discussion. I just ask you to stop repeating the same point over and over when I’ve repeatedly shown you I understood what you’re saying but don’t agree *and explained why*.

Challenge the rules, sure, but argue as to why you think the suggested approach is wrong (which is closer to what you’ve done in your latest reply). Hammering the same point gets us nowhere.

vitorgalvao:
Ideally, we’d be able to have the both in the same cask and pick it at install time with an install flag or something. Unfortunately, that is not something we’re equipped to do, right now.

adityadalal924:
`brew cask anaconda` will now install anaconda(3)

If you need to install anaconda(2), you can always use `brew cask install caskroom/versions/anaconda2`

cclauss:
Nice.  Thanks guys.

vitorgalvao:
Thank you for the changes, @adityadalal924.

Thank you for the submission, @cclauss, and introducing us to the two versions.

Apologies (to both) if I seem to have exploded a bit, before. Reading it again, it sound worse than what I originally intended.

cclauss:
We are all good.

PR #15601, opened 2015-12-04
remove python(2)

adidalal:
Gets rid of python(2)

But AFAIK, we shouldn't even have any python in homebrew cask, as `brew` has forumulas for both (and we don't want duplicate tokens between the two projects).

`brew install python` vs. `brew cask install python` seems weird

@vitorgalvao removing both altogether would be correct but if it breaks workflows we probably shouldn't do this (yet)


vitorgalvao:
Agreed. In the past we had no issue with duplicating software in homebrew, but since we’re planning on merging, we should take that into account.

Rather than deleting it now, though, perhaps we should open an issue where we list every cask we have in common with homebrew. That way we can see the bigger picture, and really understand the ramifications of going all out with a “no duplicates with homebrew” stance. That could be beneficial for both teams.

If we decide it looks good, we can then go ahead and delete everything at once.

Python 2 should still be deleted, though.

adityadalal924:
Closing in favor of the bigger picture discussion. Can always open this back up at a later time, it's literally just a deletion.

PR #15610, opened 2015-12-05
Redis Desktop Manager

otzy007:
http://redisdesktop.com

### DIFF
--- a/Casks/redis-desktop-manager.rb
+++ b/Casks/redis-desktop-manager.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'rdm' do
+  version '0.8.3'
+  sha256 '10132a8fcd4d3dc67ddc862ac5f2ef16ba9d5620f57ec9ef906ed9a7c134fd86'
+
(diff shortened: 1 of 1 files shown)


Casks/redis-desktop-manager.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'rdm' do
```
vitorgalvao: The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

This [already exists](https://github.com/caskroom/homebrew-cask/blob/master/Casks/rdm.rb).

vitorgalvao:
The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

This [already exists](https://github.com/caskroom/homebrew-cask/blob/master/Casks/rdm.rb).

otzy007:
oops. Soory

PR #15623, opened 2015-12-05
Update time-relic.rb

julianrogmans:
looking at the download url on timerelic.com there is a .app before the .zip in the URL


adityadalal924:
Thank you for the contribution. It was merged directly as 9f0ccfe83db5d6a263e877363f8d660dc264e501 to keep commit history cleaner. Your contribution is still credited to you.

PR #15628, opened 2015-12-06
Add Synth1 VST / Audio Unit

joshka:


Casks/synth1-au.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'synth1-au' do
+  version '1.13beta8'
+  sha256 '78660848d78c19cdda2597bd050a2ec95fa915cc1a1ab598bde0192a1cbc9a6f'
+
+  url 'http://www.geocities.jp/daichi1969/softsynth/Synth1macau113beta8.zip'
```
adityadalal924: You can use a `#{version.delete('.')}` here

adityadalal924:
Do these two depend on each other? If so, which one is considered the "main" one

Casks/synth1-au.rb
joshka: You're right. Will fix later tonight.

joshka:
They don't depend on each other, They are two different types of plugin (Audio Unit / VST).
Some DAWs use VSTs (Cubase, Ableton-live, etc.) some use Audio Units (Apple Logic etc.)
In this case, the plugin is released as two separate zip files on the vendor's homepage.
Some plugins are released as both in the same.

If you want to hold off merging this, I'm working on changing to use two new artifacts called vst_plugin and audiounit_plugin, for the purpose of being able to write a bunch of these quickly. In the same vein as the internet_plugin artifact type.

adityadalal924:
That sounds like a more flexible option.

Open up a PR when you're ready and we'll take a look!

Thanks.

(Closing this for the time being, in that case). Feel free to reopen if you end up not making any core changes and we can merge this.

PR #15629, opened 2015-12-06
Kicad 4.0.0

fanquake:
There is another 'extras' DMG that is recommended to be installed. Should we add that to this cask?

See http://kicad-pcb.org/download/osx/

### DIFF
--- a/Casks/kicad.rb
+++ b/Casks/kicad.rb
@@ -1,9 +1,8 @@
 cask :v1 => 'kicad' do
-  version '2014-02-26'
+  version '4.0.0'
   sha256 '066520f10e9646e88c2e1d7812adf54a73a331592076f394cb439a97c228410b'
 
-  # mdx4.org is the official download host per the vendor homepage
-  url "http://www.mdx4.org/uploads/kicad/Kicad-product-#{version}.zip"
+  url "http://downloads.kicad-pcb.org/osx/stable/kicad-#{version}.dmg"
   name 'KiCad'
   homepage 'http://www.kicad-pcb.org/'
   license :gpl


adityadalal924:
there's a kicad-extras cask.

Also this is funky - there's another folder that needs to be dragged into application support from the .dmg; not sure how to handle that with cask language

vitorgalvao:
Regarding `kicad-extras`, @gibsjose [submitted a cask a while ago](https://github.com/caskroom/homebrew-cask/pull/15470). He was supposed to resubmit with an updated version, but ended up never doing so.

gibsjose:
@vitorgalvao Yeah I ran into the same issue as @adityadalal924 in that I was not sure how to elegantly handle the need to drag a folder from the DMG, since it seems like that defeats the purpose of using `homebrew`. Can this be solved by just using the `app` stanza in a certain way?

I was also waiting for the stable 4.0.0 release, which is now available.

adityadalal924:
`artifact 'kicad', :target => Pathname.new(File.expand_path('~')).join('Library/Application Support/kicad')` would likely do the trick (as previously seen in `kicad-library`

I think adding in that line to `kicad`, removing `kicad-library`, and making a new Cask `kicad-extras` with a depends_on => kicad would be the most sane option here.

gibsjose:
Yeah that sounds logical. The `kicad-library` Cask does need to go.

jawshooah:
@adityadalal924 `artifact 'kicad', :target => "#{ENV['HOME']}/Library/Application Support/kicad"` seems a little cleaner to me.

adityadalal924:
Closing in favor of https://github.com/caskroom/homebrew-cask/pull/15692

PR #15630, opened 2015-12-06
update kindlecomicconverter v5.0.1

ghost:


Casks/kindlecomicconverter.rb
```diff
@@ -1,11 +1,11 @@
 cask :v1 => 'kindlecomicconverter' do
-  version '4.6.3'
-  sha256 'e895841842990b4e009d94f2fa125afdae632bfca60953d61b8eb7e167da6bfd'
+  version '5.0.1'
+  sha256 '48bf8bb6d57972652399d99a89d04aee2fcd3ead31898ea07594ea7db0edfed9'
 
-  url "http://kcc.iosphe.re/OSX/KindleComicConverter_osx_#{version}.zip"
+  url "https://kcc.iosphe.re/OSX/"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

Casks/kindlecomicconverter.rb
```diff
@@ -1,11 +1,11 @@
 cask :v1 => 'kindlecomicconverter' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'kindle-comic-converter' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `kindle-comic-converter.rb`.

vitorgalvao:

Casks/kindlecomicconverter.rb
adityadalal924: The url should be `https://kcc.iosphe.re/OSX/KindleComicConverter_osx_#{version}.dmg`

adityadalal924:
@vitorgalvao That is no longer the case. Note the new versioned URL I found.

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in b66f86335ebf8330acff2bb509c77062f2a60ead. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15633, opened 2015-12-06
Add GitKraken (invitation only beta currently)

joshka:


adityadalal924:
This should probably go in https://github.com/caskroom/homebrew-unofficial as the app itself is not usable without an invite code.

It can always be moved into the main caskroom repo after the closed beta is over.

Mind making a PR over in the other repo?

joshka:
Done. I wasn't sure though, given the description of unofficial in the docs didn't quite match:

"Unofficial: An allegedly unmodified compiled binary, by a third-party, of a binary that has no existing build by the owner of the source code."

This is definitely official, but just not open beta yet.

PR #15652, opened 2015-12-07
Add Calq

rinopo:
"Calq is an easy to use, on demand calculator."

I followed [`git-annex.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/git-annex.rb) to force the `.dmg` file extension to the gzipped container. (cf. #3650)

### DIFF
--- a/Casks/calq.rb
+++ b/Casks/calq.rb
@@ -1,0 +1,21 @@
+cask :v1 => 'calq' do
+  version '1.4.6'
+  sha256 '3e12d3fe2aea2df0483e91081ff24558b7382bf7390b07ab9bd0b5912627ef86'
+
+  url "http://www.katoemba.net/download/Calq-#{version}.dmg.gz"
+
+  # This is a horrible hack to force the file extension.  The
+  # backend code should be fixed so that this is not needed.
+  preflight do
+    system '/bin/mv', '--', staged_path.join("Calq-#{version}"), staged_path.join("Calq-#{version}.dmg")
+  end
+  container :nested => "Calq-#{version}.dmg"
+
+  name 'Calq'
+  homepage 'http://www.katoemba.net/makesnosenseatall/calq/'
+  license :gratis
+
+  app 'Calq.app'
+
+  zap :delete => '~/Library/Preferences/com.katoemba.calq.plist'
+end


adityadalal924:
@vitorgalvao This Cask is technically okay, but it uses a "horrible hack" so it's questionable if we should merge it (even though we do already have a cask with the same behavior, git-annex)

vitorgalvao:
Yes, it’s OK to merge, I’d say.

Casks/calq.rb
```diff
@@ -0,0 +1,21 @@
+cask :v1 => 'calq' do
+  version '1.4.6'
+  sha256 '3e12d3fe2aea2df0483e91081ff24558b7382bf7390b07ab9bd0b5912627ef86'
+
+  url "http://www.katoemba.net/download/Calq-#{version}.dmg.gz"
+
+  # This is a horrible hack to force the file extension.  The
+  # backend code should be fixed so that this is not needed.
+  preflight do
+    system '/bin/mv', '--', staged_path.join("calq-#{version}"), staged_path.join("calq-#{version}.dmg")
```
jawshooah: This will cause problems on case-sensitive filesystems. Needs to be `Calq-#{version}`

rinopo:
Hi,

Thanks @jawshooah for pointing out. I've made the fix.

@adityadalal924 
Thank you for your consideration.
I'd argue that this cask shouldn't do any harm at least on the end user's side.
[`fuse.rb`](https://github.com/caskroom/homebrew-cask/blob/3699589a533da2e25bdf0996ea342be089c2abf6/Casks/fuse.rb) which uses (to my eyes) essentially the same technique has been merged 3 days ago, and there seems to be [a few more examples](https://github.com/caskroom/homebrew-cask/search?utf8=%E2%9C%93&q=preflight+system+mv&type=Code).
And I find `calq.app` itself is, old it is, still quite useful ;)

vitorgalvao:
@rinopo To clarify, @adityadalal924 wasn’t exactly worried about any harm caused by the cask, but rather that we frown upon these types of hacks (we prefer to fix the core issue).

rinopo:
@vitorgalvao Thanks. Squashed.

vitorgalvao:
Thank you for the contribution. It was merged directly as c61a4631e437c6a7689b0b7518041bc0429fb31b to keep commit history cleaner. Your contribution is still credited to you.

PR #15679, opened 2015-12-08
Add sha256 check for mactex

dionyziz:


vitorgalvao:
Seeing as `url` doesn’t point to a specific version, this cask needs to use `version :latest` and `sha256 :no_check`, or it’ll break constantly.

dionyziz:
This is a security problem. I'll change the URL to point to a specific
version.

PR #15692, opened 2015-12-08
added kicad and kicad-extras v4.0.0

adidalal:
Handles all changes for kicad v4
- updated kicad to v4
- removed kicad-library
- added new Cask kicad-extras


adityadalal924:
Looks like travis is acting funky.

```
The command "/System/Library/Frameworks/Ruby.framework/Versions/"${CASK_RUBY_TEST_VERSION}"/usr/bin/bundle exec "/System/Library/Frameworks/Ruby.framework/Versions/${CASK_RUBY_TEST_VERSION}/usr/bin/rake" rubocop" exited with 0.
0.00s$ developer/bin/audit_modified_casks "${TRAVIS_BRANCH}..${TRAVIS_COMMIT}"
fatal: Invalid revision range master
The command "developer/bin/audit_modified_casks "${TRAVIS_BRANCH}..${TRAVIS_COMMIT}"" exited with 0.
```


@fanquake When you get a chance can you verify that this Cask (and kicad-extras) works for you locally?

jawshooah:
Something's up with `kicad-extras`:

```
$ brew cask install kicad-extras
==> Satisfying dependencies
==> Installing Cask dependencies: kicad
kicad ...
==> Downloading http://downloads.kicad-pcb.org/osx/stable/kicad-4.0.0.dmg
==> Symlinking App Suite 'Kicad' to '/Applications/Kicad'
==> Symlinking Generic artifact 'kicad' to '/Users/joshuahagins/Library/Application Support/kicad'
🍺  kicad staged at '/opt/homebrew-cask/Caskroom/kicad/4.0.0' (5470 files, 725M)
done
complete
==> Downloading http://downloads.kicad-pcb.org/osx/stable/kicad-extras-4.0.0.dmg
Error: It seems the symlink source is not there: '/opt/homebrew-cask/Caskroom/kicad-extras/4.0.0/kicad'
```

2manyprojects2littletime:
there's a requirement to open kicad at least once before installing the kicad-extras

2manyprojects2littletime:
is it ok to do something like:
`Thread.new {system("#{ENV["Home"]}/Applications/Kicad/kicad.app/Contents/MacOS/kicad")}`
as a postflight on the Kicad install?

vitorgalvao:
Calling a GUI after install is frowned upon. It’s one thing when the installer does it (and even then we don’t like it), but for us to do it on purpose is a different story.

Why would it be a requirement for `kicad-extras` to be installed that `kicad` is open at least once? All we’re doing is moving a directory.

2manyprojects2littletime:
It's acknowledged as a hack in the kicad-extras readme, but the folder it's looking for isn't created until the first run of kicad.  The authors say it will be changed in a future release, but I don't know if it's scheduled yet or what their release cadence is.

vitorgalvao:
> but the folder it's looking for isn't created until the first run of kicad.

Then all we need to do is make sure the directory is there (i.e. create it). No need to launch `kicad`.

adityadalal924:
@vitorgalvao How would I go about doing that? (I'm assuming in a `preflight` stanza with a conditional like "if directory does not exist, create it", but I didn't see any examples of other Casks that have such a requirement).

vitorgalvao:
Well, `:target` actually takes care of creating the needed directories. There’s at least [one cask that creates a directory](https://github.com/caskroom/homebrew-cask/blob/1af44ca2b1469244dd218d31fc8ab30a2fa480d8/Casks/xquartz.rb#L16) but that’s for an unrelated reason

Symlinking a `Application Support` directory isn’t a great idea, but I guess it’s fine for now, untill #13201 fixes that as well.

The issue here is different though. `kicad` inside `kicad-extras` isn’t a directory, but a symlink to `~/Library/Application Support/kicad` (like other `dmg`s usually include a symlink to `/Applications`), so the line actually needs to be `artifact 'kicad', :target => "#{ENV['HOME']}/Library/Application Support/kicad/modules"`.

adityadalal924:
Fixed kicad-extras, things should be OK now.

Will check out the CI build when it's done and merge accordingly, as the packages are huge and take quite a bit of time to download.

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  url "http://downloads.kicad-pcb.org/osx/stable/kicad-extras-#{version}.dmg"
+  name 'KiCad Extras'
+  homepage 'http://www.kicad-pcb.org/'
+  license :gpl
+
+  depends_on :cask => 'kicad'
+
+  artifact 'modules', :target => "#{ENV['HOME']}/Library/Application Support/kicad/modules"
+  artifact 'fp-lib-table', :target => "#{ENV['HOME']}/Library/Preferences/kicad/"
```
jawshooah: Need to append `fp-lib-table` to the target path. Otherwise you get 

```
~/Library/Preferences/kicad -> /opt/homebrew-cask/Caskroom/kicad-extras/4.0.0/fp-lib-table
```

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,14 @@
...
+  sha256 '8606e3a181b89a8234ea87d44342cbfa50d420375bce30167392c03e6c96db69'
+
+  url "http://downloads.kicad-pcb.org/osx/stable/kicad-extras-#{version}.dmg"
+  name 'KiCad Extras'
+  homepage 'http://www.kicad-pcb.org/'
+  license :gpl
+
+  depends_on :cask => 'kicad'
+
+  artifact 'modules', :target => "#{ENV['HOME']}/Library/Application Support/kicad/modules"
```
jawshooah: On case-insensitive filesystems, this results in the following:

```
==> It seems there is already a Generic artifact at '/Users/joshuahagins/Library/Application Support/kicad/modules'; not linking.
```

This is because `KiCad` and `kicad` are considered the same on a case-insensitive filesystem. When staging the `kicad` cask, the `KiCad` and `kicad` folders from the dmg are merged into one, so the two symlinks `/Applications/KiCad` and `/Users/joshuahagins/Library/Application Support/kicad` point to the same directory.

jawshooah: This can't be solved by renaming one of the directories in a `preflight` block, since the cask is already staged by the time it's executed. This may require an upstream fix. Considering that case-insensitive is the default for new Macs, this was incredibly short-sighted of the KiCad team.

adityadalal924:
In this case, we could just merge `kicad` and ignore `kicad-extras`, as it seems that there's nothing more we can do.

jawshooah:
> When staging the `kicad` cask, the `KiCad` and `kicad` folders from the dmg are merged into one, so the two symlinks `/Applications/KiCad` and `/Users/joshuahagins/Library/Application Support/kicad` point to the same directory.

The `kicad` cask is also a problem (for case-insensitive filesystems, which I'd wager is the majority). This is a potential solution:

```rb
cask :v1 => 'kicad' do
  version '4.0.0'
  sha256 '71b47ce37acd1d32eaee898254114203ef2de8b382f1353db7a01caa37a798e5'

  url "http://downloads.kicad-pcb.org/osx/stable/kicad-#{version}.dmg"
  name 'KiCad'
  homepage 'http://www.kicad-pcb.org/'
  license :gpl

  preflight do
    system '/bin/mkdir', '--', "#{staged_path}/Kicad-apps"
    system '/bin/mv', '--', "#{staged_path}/Kicad/*.app", "#{staged_path}/Kicad-apps/"
  end

  suite 'Kicad-apps', :target => 'Kicad'
  artifact 'kicad', :target => "#{ENV['HOME']}/Library/Application Support/kicad"
end
```

adityadalal924:
Updated to incorporate @jawshooah's suggestions. `kicad-extras` will still break, however.
Will look into contacting upstream as time permits.

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in e420576e65ca55084af1ce61851e20cfd5f3d806. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15694, opened 2015-12-08
cdock 0.9.8

dpoggi:


adityadalal924:
Cask needs to be renamed to cdock2.rb. Also should have a depends on, with a minimum of osx 10.9

vitorgalvao:
> Cask needs to be renamed to cdock2.rb.

Not in this case. The `2` is still very much interpreted to mean a version, a continuation. This is evidenced by the fact the app bundle retains the name without the version, and the previous one was discontinued.

adityadalal924:
Thank you for the contribution. It was merged directly as 69cc3d08ee7ab1d7abb5c8179b6ee7a3194b64f0 to keep commit history cleaner. Your contribution is still credited to you.

PR #15697, opened 2015-12-08
Mactex sha256

dionyziz:
Added sha256 check and specific version URL for mactex as requested in #15679

### DIFF
--- a/Casks/mactex.rb
+++ b/Casks/mactex.rb
@@ -2,7 +2,7 @@ cask :v1 => 'mactex' do
   version :latest
   sha256 'c5f5b0fd853a17dab6e844fb5e893804af78d938fa18ee94ec3b257611a95c12'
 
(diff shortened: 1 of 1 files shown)


Casks/mactex.rb
```diff
@@ -1,8 +1,8 @@
 cask :v1 => 'mactex' do
   version :latest
```
vitorgalvao: `version` can’t be left unset.

vitorgalvao:
But this isn’t a good link, though, as a mirror has implications (such as download speed).

See https://github.com/caskroom/homebrew-cask/pull/15196 as to why it was changed. That discussion came from reported cases of when the versioned mirror url caused issues (slow downloads that failed and made it impossible to get), so we cannot accept this.

vitorgalvao:
If you really need a versioned download link here, you have two choices:

1. Make your own tap.
2. Ask the mactex developers to make their official download link versioned.

PR #15708, opened 2015-12-09
Add Mouseposé

rinopo:
"When turned on, it dims the screen and puts a spotlight on the area around the mouse pointer, easily guiding the audience‘s attention to an area of interest."

---

Some notes:
- The download link (`url`) is not obvious on [the vendor's webpage](https://www.boinx.com/mousepose/overview/) - only Mac App Store link - but is [not hidden](https://forum.boinx.com/discussion/4516/can-t-update-due-to-error) for those who have purchased the license elsewhere.
- Ideally the app should be installed as "Mouseposé.app" (as is the case with "normal" installation process), but this cask installs it as "Mousepose.app". Specifying the `app` stanza as `'Mouseposé.app'` yields the following error:
  
  ```
  Error: It seems the symlink source is not there: '/opt/homebrew-cask/Caskroom/mousepose/3.2.7/Mouseposé.app'
  ```
  
  I couldn't find a way to resolve this.

### DIFF
--- a/Casks/mousepose.rb
+++ b/Casks/mousepose.rb
@@ -1,0 +1,22 @@
+cask :v1 => 'mousepose' do
+  version '3.2.7'
+  sha256 '05a3c5175206e971b386a75cd28e51873fdff6c47e24c5673a8116af127d336c'
+
+  url "https://cdn.boinx.com/software/mousepose/Boinx_Mousepose_#{version}-10878.app.zip"
+  name 'Mouseposé'
+  name 'Mousepose'
+  homepage 'https://www.boinx.com/mousepose/'
+  license :gratis
+
+  # Renamed for consistency: app name is different in the Finder and in a shell.
+  # Original discussion: https://github.com/caskroom/homebrew-cask/pull/15708
+  app 'Mousepose.app', :target => "Mousepose\314\201.app"
+
+  zap :delete => [
+                  '~/Library/Application Support/com.boinx.Mousepose',
+                  '~/Library/Caches/com.boinx.Mousepose/',
+                  '~/Library/Cookies/com.boinx.Mousepose.binarycookies',
+                  '~/Library/Preferences/com.boinx.Mousepose.plist',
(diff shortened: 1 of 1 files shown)


vitorgalvao:
As for your first point, we don’t care as long as the download host is the same as the homepage.

As for your second point, try looking at the app in a Terminal, rather than the Finder. Does it show as `Mousepose.app`? Then it likely has a different name in the Finder and in a Terminal (not unheard of), in which case, follow [this solution](https://github.com/caskroom/homebrew-cask/blob/1c802dc1359e09383f8ee1f74c5e1714ae6efb04/Casks/devonthink-pro-office.rb#L14#L16).

adityadalal924:
Thank you for the contribution. It was merged directly as 9cf3550ef90b6bde7c370247cce26a311aa70b6f to keep commit history cleaner. Your contribution is still credited to you.

rinopo:
@vitorgalvao 
Thanks a lot for the suggestion.
Since that was exactly the case, I added the `target` key.

`:target => 'Mouseposé.app'` seemed to work at first, but turned out that upon `brew cask uninstall` it somehow leaves an orphaned icon of Mouseposé in the Launchpad. So I tried `:target => "Mousepose\314\201.app"` and that seems to work. (All this is just empirical, and maybe only me, but I'm afraid I have no guts to dig into this.)

@adityadalal924 
Thank you for merging.

adityadalal924:
@rinopo I verified that things were OK before merging.

```
[~] > brew cask install mousepose
==> Downloading https://cdn.boinx.com/software/mousepose/Boinx_Mousepose_3.2.7-1
==> Symlinking App 'Mousepose.app' to '/Applications/Mouseposé.app'
🍺  mousepose staged at '/opt/homebrew-cask/Caskroom/mousepose/3.2.7' (300 files, 12M)
```

```
[~] > brew cask zap mousepose
==> Implied "brew cask uninstall mousepose"
==> Removing App symlink: '/Applications/Mouseposé.app'
==> Dispatching zap stanza
==> Running zap process for mousepose; your password may be necessary
==> Removing files: ["~/Library/Application Support/com.boinx.Mousepose", "~/Lib
==> Removing all staged versions of Cask 'mousepose'
[~] > 
```

PR #15714, opened 2015-12-09
Add version and checksum to Adobe CC apps

cprecioso:


vitorgalvao:
Thank you, but seeing as `url` doesn’t point to a specific version, these casks need `version :latest` and `sha256 :no_check`, or they’ll break on updates.

cprecioso:
I know, and that's why I submitted them with `:latest` and `:no_check` initially. But having re-downloaded these it still asked me to update all the way to the latest release, which got me investigating.

Even though it doesn't say the version explicitly in the URL, these are canonical URLs that are always alive (Adobe still supports these kind of trial download links back to CS3), and never change (Adobe itself can't change the Dreamweaver download even though it's broken). That's why I thought it was appropriate to version and checksum them.

vitorgalvao:
That’s precisely why we cannot have a version and checksum. Because when Adobe updates the version on the server buts keeps the URL the same, then the shasum won’t march anymore, making it a broken cask.

cprecioso:
> are always alive (...), and **never change** (Adobe itself can't fix the Dreamweaver download even though it's broken)

I meant the downloaded contents themselves don't change.

vitorgalvao:
Ah, I see what you mean. It is fine, then.

adityadalal924:
Thank you for the contribution. It was merged directly as c8d0a9075d17fda7b322433f392f7f89331069c3 to keep commit history cleaner. Your contribution is still credited to you.

galaxy001:
Will you add Adobe Audition ?

cprecioso:
Open a new issue and I'll look at it as soon as I can.

PR #15728, opened 2015-12-09
Fix and update thebrain cask to latest version (8.0.2.1)

oillio:


adityadalal924:
Please add an explicit sudo (ie true or false), as we are trying to enforce that in casks moving forward. Ref: https://github.com/caskroom/homebrew-cask/blob/master/Casks/thinkorswim.rb for an example.

Also if you could [squash your commits](https://davidwalsh.name/squash-commits-git) when done, that would also be helpful.

Thanks for the fix!

oillio:
Done

adityadalal924:
Thank you for the contribution. It was merged directly as 16ee8b76dd5c71a23d6bba93306dd202365febf3 to keep commit history cleaner. Your contribution is still credited to you.

PR #15731, opened 2015-12-09
README.md: Add in command to tap cask

adidalal:
As we no longer get automatically installed by homebrew, let's add in the tap command to the readme again

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ It’s implemented as a `homebrew` [external command](https://github.com/Homebre
 ## Let’s try it!
 
 ```bash
+$ brew tap caskroom/cask
 $ brew cask install google-chrome
 => Downloading https://dl.google.com/chrome/mac/stable/GGRO/googlechrome.dmg
 => Success! google-chrome installed to /opt/homebrew-cask/Caskroom/google-chrome/stable-channel


adityadalal924:
Merged as e48fcbe1218b8d65a9ac7f3fe56f2e4737423f96.

vitorgalvao:
We do still get automatically installed.

If you’re thinking of [the recent submission](https://github.com/Homebrew/homebrew/pull/46845/files) by @jawshooah to homebrew, what he’s doing there is removing the specific `install` command, but just because it is no longer needed since https://github.com/caskroom/homebrew-cask/pull/15381.

adityadalal924:
@vitorgalvao I'm confused then, why is `brew tap caskroom/cask` needed on caskroom.io at all then? (or more generically, when should we be telling people to `brew tap caskroom/cask`)

jawshooah:
Only reason I can think of is to protect against the Homebrew team changing their minds about auto-installing us.

adityadalal924:
Ah I see, this PR can likely be reverted then, though there's probably no harm in keeping it.
Up to you!

vitorgalvao:
I was thinking about possible user confusion. If we don’t have a “how to install” section on the website, I envision users getting confused about how to get HBC. On the other hand, explaining “you don’t need to do anything, as homebrew will take care of everything for you” seems like it may be confusing as well.

Since the website is more user-friendly (“do this”) and the repo is more developer-friendly (“do this, and here’s why”), I was thinking it would make sense and be helpful to have that instructions *just* on the website (to prevent about the mentioned user confusion).

I do believe we should remove the tap instruction from the repo. As for the website, I’m very open to whatever you think best.

adityadalal924:
Reverted in https://github.com/caskroom/homebrew-cask/commit/e8ff79f03e3d0f00c016e7ba29ee06016b347f1a

I'll take a closer look at the website as time permits

PR #15736, opened 2015-12-10
Dbvisualizer upgrade

sparhomenko:
Besides upgrading to the latest available version, this also switches to the default distribution of DbVisualizer which requires Java 8 instead of Java 7. Since 8 is also the default version of the "java" cask, and since Java 7 version of DbVisualizer is not even available on the [download page](https://www.dbvis.com/download/) and may become unsupported, this seems like a logical switch.


vitorgalvao:
Thank you for the contribution. It was merged directly as 3a2f1441a60d89207083b821d8eb64fe52509555 to keep commit history cleaner. Your contribution is still credited to you.

PR #15740, opened 2015-12-10
Add support for uninstalling login items

### LINKED ISSUE #14854: add support for Login Items removal during uninstallation
Majority of GUI clients persist their autostart preference inside users Login Items list.
A few examples of such applications are [dropbox.rb](../blob/master/Casks/dropbox.rb), [spotify.rb](../blob/master/Casks/spotify.rb) and [flux.rb](../blob/master/Casks/flux.rb).

Configuration files are located at `~/Library/Preferences/com.apple.loginitems.plist` and items can be removed from there.
A nicer option would be to use `osascript`:
`osascript -e 'tell application "System Events" to delete every login item whose name is "AppName"'`
This not only removes every matching Login Item, but also it would continue normally if there aren't any matches.

I'd like to see a helper dedicated to deal with Login Items within HBC so that 
```ruby
uninstall_postflight do
  system '/usr/bin/osascript', '-e', %Q{tell application "System Events" to delete every \
    login item whose name is "#{token.gsub(/\w+/, &:capitalize).gsub('-', ' ')}"}
end
```
can be replaced with something like
```ruby
uninstall :login_item => 'Turbo Boost Switcher'
```

---

Incomplete list of apps that would benefit from this:

- [ ] [dropbox.rb](../blob/master/Casks/dropbox.rb)
- [ ] [flux.rb](../blob/master/Casks/flux.rb).
- [ ] [spotify.rb](../blob/master/Casks/spotify.rb)
- [ ] [turbo-boost-switcher.rb](../blob/master/Casks/turbo-boost-switcher.rb)

jawshooah:
Closes #14854.

I've also included some refactorings for `Hbc::Artifact::UninstallBase` which will make it easier to maintain existing directives and add new ones in the future.

This will require changes to the language reference and language delta docs, but I wanted to get some input first.

[Review on Reviewable](https://reviewable.io/reviews/caskroom/homebrew-cask/15740)

### DIFF
--- a/developer/bin/list_login_items_for_app
+++ b/developer/bin/list_login_items_for_app
@@ -1,0 +1,64 @@
+#!/usr/bin/env ruby
+#
+# list_login_items_for_app
+#
+
+###
+### dependencies
+###
+
+require 'open3'
+
+###
+### methods
+###
+
+def usage
+  <<-EOS
+Usage: list_login_items_for_app <path.app>
+
+Given an Application (app) bundle directory on disk, find all
+login items associated with that app, which you can use in a
+Cask uninstall stanza, eg
+
+    uninstall :login_item => 'login item name'
+
+Note that you will likely need to have opened the app at least
+once for any login items to be present.
+
+See CONTRIBUTING.md for more information.
+
+EOS
+end
+
+def process_args
+  if ARGV.first =~ /^-+h(?:elp)?$/
+    puts usage
+    exit 0
+  elsif ARGV.length == 1
+    $app_path = ARGV.first
+  else
+    puts usage
+    exit 1
+  end
+end
+
+def list_login_items_for_app(app_path)
+  out, err, status = Open3.capture3(
+    '/usr/bin/osascript', '-e',
+    "tell application \"System Events\" to get the name of every login item " \
+    "whose path contains \"#{File.basename(app_path)}\""
+  )
+  if status.exitstatus > 0
+    $stderr.puts err
+    exit status.exitstatus
+  end
+  puts out.gsub(', ', "\n")
+end
+
+###
+### main
+###
+
+process_args
+list_login_items_for_app $app_path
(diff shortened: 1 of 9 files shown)


vitorgalvao:
Do we need sudo, or is that just a proactive measure?

Since discussing that issue, it has also come to my attention that some apps, rather than using `Login Items`, add themselves to startup by installing a launch daemon (under `~/Library/LaunchAgents`). I didn’t remember at the time to keep track of them for this issue, but I do remember at least one was an `app`, not a `pkg`.

It might make sense, in the future, to also add a `:login_daemon` which would take the path to the `plist`, unload it and delete it.

reitermarkus:
@vitorgalvao, then wouldn't `:login_daemon` be the same as `:launchctl`?

jawshooah:
@vitorgalvao Adding sudo was a proactive measure. Login items are per-user, so I figured using sudo would remove them for all users. Turns out not to be true; only the login items for the currently logged-in user are removed, regardless of sudo. I'll set it to `false`.

And @reitermarkus is right, what you describe is exactly what `:launchctl` does.

vitorgalvao:
Ah, yes, you’re right. I associate `:launchctl` so close with `:pkgutil`, it didn’t even cross my mind.

Regarding sudo, we wouldn’t want to remove them for all users anyway, since we’re using the app’s name and that could lead to unexpected results. I’ve seen it happen and done it myself, when you use another person’s Mac with your own account temporarily, and so download and keep the apps locally to the Desktop, or something, just because it’s temporary and more convenient. We would’t want the administrator to mess the setup of the the other person by accident.

vitorgalvao:
Will need some fixes, now (unintentional conflicts caused by the merging of https://github.com/caskroom/homebrew-cask/pull/15755).

jawshooah:
Very good point, we don't want to accidentally screw up someone else's account settings. `sudo` is unnecessary here in any case, so I've set it to `false`.

Also rebased to resolve conflicts. I'll update the docs in a bit before merging.

jawshooah:
Added docs and an additional test for `zap :login_item`. Let me know if you guys have any comments, otherwise I'll merge in the morning.

szhu:
This makes it impossible to uninstall apps that use `:login_item` over SSH because the System Events app cannot be started over SSH. Is there a flag that makes it possible to bypass this?

jawshooah:
@szhu Please open a separate issue instead of commenting on this old PR.

PR #15741, opened 2015-12-10
chameleon-ssd-optimizer 'version :latest'

yurikoles:
Upstream now uses permanent link for latest stable version.


vitorgalvao:
Thank you, but the old link still works, and that old link points to the same version, so there’s a good chance versioned links will continue to work (not uncommon). Lets wait for the app to be updated and reevaluate then.

PR #15755, opened 2015-12-10
Remove extra space in update command

jawshooah:
I assume this is the more commonly used form. I've never seen anyone space out their semicolons like `foo ; bar`.

[Review on Reviewable](https://reviewable.io/reviews/caskroom/homebrew-cask/15755)


vitorgalvao:
No objection here.

PR #15760, opened 2015-12-10
Update PHPStorm Version for from 10.0.1 for 10.0.2

josevieirame:


phpmypython:
i just made another pull with the updates. Would you mind accepting mine instead of this one since i did have it pushed just needed to get the commits squashed correctly

vitorgalvao:
Either one is fine, it doesn’t really matter. I was merging the other already. Thank you anyway, @vieirasantosn.

PR #15772, opened 2015-12-11
Compiled version of “Winbox” by MikroTik with Wine

nathanielvarona:
Compiled version of “Winbox” by MikroTik with Wine in order to make it usable on Mac.

### DIFF
--- a/Casks/winbox4mac.rb
+++ b/Casks/winbox4mac.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'winbox4mac' do
+  version '3.0'
+  sha256 '2924d7629524f835ae9b49e125edb44f9e8cda890e41609fab5db66a8d63f595'
+
+  url 'http://joshaven.com/Winbox4Mac_3.0.dmg'
+  name 'Winbox for Mac'
+  homepage 'http://joshaven.com/resources/tools/winbox-for-mac/'
+  license :unknown
+
+  app 'Winbox4Mac.app'
+end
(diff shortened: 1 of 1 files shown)


Casks/winbox4mac.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'winbox4mac' do
+  version '3.0'
+  sha256 '2924d7629524f835ae9b49e125edb44f9e8cda890e41609fab5db66a8d63f595'
+
+  url 'http://joshaven.com/Winbox4Mac_3.0.dmg'
```
fanquake: You can use #{version} in the URL

Casks/winbox4mac.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'winbox4mac' do
+  version '3.0'
+  sha256 '2924d7629524f835ae9b49e125edb44f9e8cda890e41609fab5db66a8d63f595'
+
+  url 'http://joshaven.com/Winbox4Mac_#{version}.dmg'
```
victorpopkov: You should use double quotes here due to string interpolation: `#{version}`.

fanquake:
Please squash your commits.

Casks/winbox4mac.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'winbox4mac' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'winbox' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `winbox.rb`.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 657d3a91d06b800c10786057ee3bc3f2ca995e5c. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15776, opened 2015-12-11
Add soundcast.app latest

danjenkins:
Soundcast allows you to output your OSX audio to a chromecast

### DIFF
--- a/Casks/soundcast.rb
+++ b/Casks/soundcast.rb
@@ -1,0 +1,14 @@
+cask :v1 => 'soundcast' do
+  version :latest
+  sha256 :no_check
+
+  # dropboxusercontent.com is the official download host per the vendor homepage
+  url 'https://dl.dropboxusercontent.com/u/6618408/soundcast.zip'
+  name 'Soundcast'
+  homepage 'https://github.com/andresgottlieb/soundcast'
+  license :mit
+
+  depends_on :cask => 'soundflower'
+
+  app 'soundcast.app'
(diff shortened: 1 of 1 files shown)


Casks/soundcast.rb
```diff
@@ -0,0 +1,13 @@
+cask :v1 => 'soundcast' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://dl.dropboxusercontent.com/u/6618408/soundcast.zip'
```
vitorgalvao: Needs a `# dropboxusercontent.com is the official download host per the vendor homepage` comment above this line ([example](https://github.com/caskroom/homebrew-cask/blob/cb487ad405f74cb1f26b93e3f3e8412c82a22e34/Casks/0ad.rb#L5); [reasoning](https://github.com/caskroom/homebrew-cask/blob/7c010b42bd7ac679c1f002a0368b36bb3f495612/doc/CASK_LANGUAGE_REFERENCE.md#when-url-and-homepage-hostnames-differ-add-a-comment)).

danjenkins: Ah thanks! I read about it needing to be the proper download url but didn't realise it needed to be comment. Will update

vitorgalvao:
Thank you for the contribution. It was merged directly as 13d657f1df12fe51ff130ee0b0adcb5bb443f48f to keep commit history cleaner. Your contribution is still credited to you.

PR #15787, opened 2015-12-12
chromecast app is too old to support new devices

ghost:


julienlavergne:
Google removed chromecast app to the benefit of a Chrome extension

Issue #15782, opened 2015-12-12
Should we deprecate cask DSL versions?


jawshooah:
With #15381 out of the way, casks should no longer be out of sync with the core library. From here on out, no cask should ever be using a DSL that's newer than the one in core (since updating casks requires running `brew update`, which will update core as well).

So what's the point of maintaining a DSL version? Keep in mind that if we do so, every time a cask update includes a new DSL element, we'll need to remember to bump the header as well.

/cc @caskroom/maintainers

jawshooah:
The way I see it, these are the problem scenarios:

- We make breaking changes to the DSL, and:
  - we forget to update a cask in one of our repos
  - an external tap maintainer is slow to catch up

In either case, we would error out regardless of whether we checked the cask's DSL version. So I don't see the point in the added maintenance burden.

@mikemcquaid, I'm curious to hear your take on this. How does Homebrew handle breaking DSL changes?

vitorgalvao:
I’m all in favour of this.

adityadalal924:
Also in favor of this. Would also encourage contributions like https://github.com/caskroom/homebrew-cask/pull/15643 as they purely add functionality - and we should make that contribution process as easy as possible.

mikemcquaid:
@jawshooah We basically never (intentionally) break backwards compatibility and just tell people to `brew update` if they hit forwards compatibility issues. When we're deprecating parts of our API we never actually remove the API method but make them print warnings on use and then eventually fail (which means the `Formula` parses but fails at audit or run time).

jawshooah:
@mikemcquaid That sounds like a good strategy going forward.

Closing this, since it seems we have a consensus. Maintainers, feel free to re-open if you think this requires more discussion.

PR #15793, opened 2015-12-12
Remove login items in 9 casks

vitorgalvao:
Following https://github.com/caskroom/homebrew-cask/issues/14854 and https://github.com/caskroom/homebrew-cask/pull/15740, adding the feature to some casks.

### DIFF
--- a/Casks/alfred.rb
+++ b/Casks/alfred.rb
@@ -1,4 +1,4 @@
-cask :v1 => 'alfred' do
+cask :v1_2 => 'alfred' do
   version '2.8.1_425'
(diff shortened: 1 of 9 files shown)


jawshooah:
#15782 is yet to be decided, so we should probably bump the headers on these casks to v1_2 for now.

vitorgalvao:
I always forget that.

vitorgalvao:
I always forget that. On the other hand, we don’t implement checks yet, so increasing them isn’t actually accomplishing anything.

vitorgalvao:
Merging now, with correct headers.

vitorgalvao:
Merged as 6229c1adb92a8dfe60f38cadf99bd56870908ba9.

PR #15808, opened 2015-12-13
add El Capitan support for Deeper and Maintenance; update sha256 value for R

tinhkyaw:
added El Capitan support for Deeper and Maintenance

### DIFF
--- a/Casks/deeper.rb
+++ b/Casks/deeper.rb
@@ -16,6 +16,8 @@ cask :v1 => 'deeper' do
     url 'http://www.titanium.free.fr/download/109/Deeper.dmg'
   elsif MacOS.release == :yosemite
     url 'http://www.titanium.free.fr/download/1010/Deeper.dmg'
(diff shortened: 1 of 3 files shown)


adityadalal924:
We like to keep PRs independent, so these will need to be split up.
Updating Deeper & Maintenance make sense together, but changing R's checksum is unrelated.

Additionally, above `url`, they MUST have the comment `#pagesperso-orange.fr is the official download host per the vendor homepage`

Please fix & comment, or if it's easier, you can close this PR and open another one.

Will look into the R shasum change.

adityadalal924:
Just a clarification - I'll look into the R shasum (a :no_check may be needed if it updates silently).

A new PR with the changes (and url comment) for Deeper & Maintenance would be appreciated.

Thanks!

tinhkyaw:
Understood.  I will put in a pull request in a few mins.  Thanks!

PR #15814, opened 2015-12-14
seil.rb: Update to 12.0.0

z0rc:
Updated to new version. Seil removed support of 10.9 and 10.10.

I didn't find information about how to calculate appcast sha sum, so this is left untouched.


vitorgalvao:
Thank you for the contribution. It was merged directly as 0924943465cec800698800a6ea477d36545d3ce7 to keep commit history cleaner. Your contribution is still credited to you.

PR #15816, opened 2015-12-14
new version of soundcast released, upped to 1.5

danjenkins:
This involved also changing the url to be from github as a release instead of dropbox

### DIFF
--- a/Casks/soundcast.rb
+++ b/Casks/soundcast.rb
@@ -1,9 +1,8 @@
 cask :v1 => 'soundcast' do
-  version :latest
-  sha256 :no_check
+  version '1.5'
+  sha256 '19c53bac9900904ed3773360036ecc04f0c3cc24d09f908e3692b4ac9c422505'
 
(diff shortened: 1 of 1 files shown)


Casks/soundcast.rb
```diff
@@ -1,9 +1,9 @@
 cask :v1 => 'soundcast' do
-  version :latest
-  sha256 :no_check
+  version '1.5'
+  sha256 '19c53bac9900904ed3773360036ecc04f0c3cc24d09f908e3692b4ac9c422505'
 
-  # dropboxusercontent.com is the official download host per the vendor homepage
-  url 'https://dl.dropboxusercontent.com/u/6618408/soundcast.zip'
+  # github.com is the official download host per the vendor homepage
```
vitorgalvao: Please remove the comment. It’s not needed, since `url` and `homepage` have the same host.

vitorgalvao:
Thank you.

PR #15820, opened 2015-12-14
Update Silverlight Cask 5.1.40728 to 5.1.41105

kesor:
New version of Silverlight.dmg file also changed the .pkg from Silverlight.pkg to silverlight.pkg (lowercase initial).


vitorgalvao:
Thank you for the contribution. It was merged directly as 822361d1464e21768d99b862dbfad8edd9bcd398 to keep commit history cleaner. Your contribution is still credited to you.

PR #15822, opened 2015-12-14
removed sha check, many versions popping out

liljenstolpe:


vitorgalvao:
Please [amend](http://stackoverflow.com/a/179147/1661012) the commit message to include the name of the updated cask.

Also, when doing this, a [specific](https://github.com/caskroom/homebrew-cask/blob/ccaf00a189cdbe50109ba6bf294c2b72aecba784/Casks/adobe-digital-editions.rb#L3) and exact comment is needed.

liljenstolpe:
Ok - will do and sqash

vitorgalvao:
Thank you. Please @mention me after you do, so I get an email to check back.

PR #15823, opened 2015-12-14
removed SHA256 check - lots of revs happening

liljenstolpe:


vitorgalvao:
[Same](https://github.com/caskroom/homebrew-cask/pull/15822#issuecomment-164526425).

liljenstolpe:
ok, will do so and squash

PR #15833, opened 2015-12-14
Add .minerva v3.5.0

Zirro:
Add .minerva, a flash shared object editor. The application has been
discontinued in favour of a version hosted online, but remains
available on the homepage and is desirable for offline use. A caveat
has been added to reflect this.


adityadalal924:
Thank you for the contribution. It was merged directly as c5dd45982656df1cdb166f4fdbd23948d37988e3 to keep commit history cleaner. Your contribution is still credited to you.

PR #15864, opened 2015-12-15
Update eagle.rb

gurghet:
Bumped version from 7.4.0 to 7.5.0, nothing fancy

### DIFF
--- a/Casks/eagle.rb
+++ b/Casks/eagle.rb
@@ -1,5 +1,5 @@
 cask :v1 => 'eagle' do
-  version '7.4.0'
+  version '7.5.0'
   sha256 '11aa7b6fc7bf9fce2e88f0cfca15b821404e42b03e9a8529c9c0b5c2781e9133'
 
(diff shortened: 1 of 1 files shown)


Casks/eagle.rb
```diff
@@ -1,5 +1,5 @@
 cask :v1 => 'eagle' do
-  version '7.4.0'
+  version '7.5.0'
   sha256 '11aa7b6fc7bf9fce2e88f0cfca15b821404e42b03e9a8529c9c0b5c2781e9133'
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

adityadalal924:
@gurghet Fixed in https://github.com/caskroom/homebrew-cask/pull/15869, but thank you for the contribution attempt, we appreciate it!

For future reference, `version` and `shasum` both need to be updated to "update" a Cask.

PR #15865, opened 2015-12-15
Add Sublime cask

sarcx:
Real program name is just `Sublime`. But is very similar
to the popular `Sublime-text`. That's why I renamed the cask to
`Sublime-subtitles`.

This app is for auto search and download subtitles only by dragging
a movie over it.

### DIFF
--- a/Casks/sublime.rb
+++ b/Casks/sublime.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'sublime' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.salukistudios.com/s/Sublime.zip'
+  name 'Sublime'
+  homepage 'http://www.salukistudios.com/sublime/'
+  license :unknown
+
+  app 'Sublime.app'
+end


Casks/sublime-subtitles.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'sublime-subtitles' do
```
vitorgalvao: This has to be changed to `cask :v1 => 'sublime' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `sublime.rb`.

vitorgalvao:

sarcx:
Fixed

Casks/sublime.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'sublime' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.salukistudios.com/s/Sublime.zip'
+  name 'Sublime'
+  homepage 'http://www.salukistudios.com/sublime/'
+  license :unknown
```
vitorgalvao: `:gratis`.

vitorgalvao:
Thank you. Could you please [squash your commits](http://davidwalsh.name/squash-commits-git)?

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in 75a9283f92f1e6f3a7882d3fd958302e8efdb33c. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15876, opened 2015-12-16
PyCharm: Add Command Line Tool

reitermarkus:


vitorgalvao:
This is smart. And it definitely threads [shim script territory](https://github.com/caskroom/homebrew-cask/issues/9203).

It’s my understanding these tools will actually guide you through the process of making such CLI executable on first run, isn’t that so? In that case, I see this as a potential source for conflict (we should assume that if the tool is asking it, the user will assume we didn’t make such executable).

adityadalal924:
See also https://github.com/caskroom/homebrew-cask/issues/12028 for previous discussion.

vitorgalvao:
Thank you for that link, @adityadalal924. Had forgotten about that discussion.

I’m more inclined to follow [this comment](https://github.com/caskroom/homebrew-cask/issues/12028#issuecomment-142387133) and close this issue and that one.

reitermarkus:
The main issue here is of course that you would have to update it manually every time you update the application because the cask directory will have a new version number. Anyway, I don't think this will be an issue anymore once the new installation behaviour is implemented as the path will stay the same.

adityadalal924:
@reitermarkus You're correct on both.

In that case, both this PR and the older issue should be closed, as we'll try to keep working towards the best long-term solution rather than adding in short-term fixes that end up making things difficult in the long run.

PR #15877, opened 2015-12-16
Use latest stable version of Proxifier on El Capitan

realdimas:
Follow-up to #14659
No need to switch to beta on El Capitan as the latest stable release now fully supports it.


adityadalal924:
Thank you for the contribution. It was merged directly as f19752572c995064a069e8606e674239ab4b4bce to keep commit history cleaner. Your contribution is still credited to you.

PR #15890, opened 2015-12-17
Update disk-sensei.rb

adidalal:
Update shasum to no_check
Finishes fix from https://github.com/caskroom/homebrew-cask/pull/15889


adityadalal924:
Merged as faf42d30515f9892d465f181f5f75330b1489e51.

PR #15892, opened 2015-12-17
Add cask for DataGrip 1.0

mirfilip:


adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in 4db3f936c7d1f6a5c90eaf49766ba9473983e801. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

mirfilip:
@adityadalal924 Hi, thanks for including that. Just out of curiosity, was that "audit" done by some tool or you in person? I ask cause I've ran `rubocop` and got no errors.

What was fixed is a removal of a blank newline. If that's the case, https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md is also wrong as examples there do care about blank lines. I didn't spot any more differences. Was that the case?

adityadalal924:
`rubocop` won't throw up any errors, but our convention has been to put a blank line between shasum and url, for readability.

You are correct in that I don't think any of the stylistic conventions are formally written down anywhere (but it's basically empty line after shasum, and empty line between stanzas). However, if you use [`brew cask create`](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#the-brew-cask-create-command), it will use that format automatically.

We'll look into adding that to the contributing document to clarify things for further PRs where people create the Cask file manually.

PR #15900, opened 2015-12-17
download.rb: Don't use cached download when cask uses `:latest`

adidalal:
Fixes https://github.com/caskroom/homebrew-cask/issues/13206


jawshooah:
:+1:

PR #15901, opened 2015-12-17
Do not require a DSL version in cask header

jawshooah:
Following up on #15782. Casks can now be created with no DSL version in the header. For example:

``` rb
cask 'no-dsl-version' do
  version :latest
  ...
end
```

Casks with DSL versions in the header are still valid, but the DSL version will be ignored.

### DIFF
--- a/lib/hbc/source/path_base.rb
+++ b/lib/hbc/source/path_base.rb
@@ -46,18 +46,11 @@ class Hbc::Source::PathBase
 
       # munge text
       cask_contents.sub!(%r{\A(\s*\#[^\n]*\n)+}, '');
-      if %r{\A\s*cask\s+:v([\d_]+)(test)?\s+=>\s+([\'\"])(\S+?)\3(?:\s*,\s*|\s+)do\s*\n}.match(cask_contents)
-        dsl_version_string = $1
-        is_test = ! $2.nil?
+      if %r{\A\s*(test_)?cask\s+(?::v[\d_]+(test)?\s+=>\s+)?([\'\"])(\S+?)\3(?:\s*,\s*|\s+)do\s*\n}.match(cask_contents)
+        is_test = $1 || $2
         header_token = $4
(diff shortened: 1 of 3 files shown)


jawshooah:
@caskroom/maintainers Any comments? If not, I'll merge at 10pm UTC.

vitorgalvao:
Why not make it a symbol as well (`cask :no-dsl-version do`)?

jawshooah:
@vitorgalvao We could, though I'm not sure what the benefit would be.

jawshooah:
Also, symbols can't contain hyphens unless surrounded in quotes. So it would need to be either `:no_dsl_version` or `:'no-dsl-version'`.

vitorgalvao:
Ah, no, ignore it. I was thinking of having it as symbol for consistency with what we had, but just realise I misinterpreted it. On first look it seemed like we needed to be explicit about not having a version, but I see now that is supposed to be the name of the cask.

adityadalal924:
LGTM.

After this merge, existing Casks could be changed to the new header as time permits. That would also be a (silent) indicator that a maintainer has looked at the Cask and ensured that everything checks out.

That would also be a  good time to implement any other breaking changes, so it might be best to avoid a mass update until the core stabilizes a bit more/the backlog of current PRs that deal with core functionality are merged or otherwise addressed.

jawshooah:
> That would also be a (silent) indicator that a maintainer has looked at the Cask and ensured that everything checks out.

To be honest, I'm not too jazzed about combing through ~3000 casks to find bugs. A mass update would be as easy as `s/cask\s+:v[\d_]+\s+=>/cask/g`.

Any PRs that include truly breaking changes should also include fixes for affected casks (in separate commits).

adityadalal924:
Yeah, on second thought, maybe not the best idea. I forgot how many Casks we had!

I would say then that this PR doesn't seem to have any regressions and so a merge, followed by a mass scripted update whenever convenient should work out well.

PR #15903, opened 2015-12-17
Added bitcoin-xt

vitorgalvao:
`curl` is having an awful lot of trouble downloading this, not yet sure why.

### DIFF
--- a/Casks/bitcoin-xt.rb
+++ b/Casks/bitcoin-xt.rb
@@ -1,0 +1,12 @@
+cask 'bitcoin-xt' do
+  version '0.11D'
+  sha256 '1152f84032a71b242f9dde79aadc77f353b9879a506692d4c0fde35ae4182f33'
+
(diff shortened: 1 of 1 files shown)


adityadalal924:
curl does't work for me either, but wget does...odd

vitorgalvao:
It’s just a github releases URL, and not a particularly big app. I wonder what’s special about it.

jawshooah:
It resolves to an Amazon S3 host, not sure why there are so many redirects
though.

vitorgalvao:
> It resolves to an Amazon S3 host

All github releases URLs do that, though (as far as I’m aware).

vitorgalvao:
Tried a bunch of other solutions and it doesn’t seem to want to work. Will close this for now, perhaps future versions will work.

If there’s any interest in the future, the interested user should have a solution/reasoning so we can work on making this function.

vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/20090.

PR #15915, opened 2015-12-18
Add Quickbooks Desktop 2016

sscotth:
I chose to go with no hash check because I'm not sure how often they make minor version bumps. The download link should be consistent.

Otherwise it would be:

``` ruby
version '17.0.1.534 R02'
sha256 'c06518f6d9118275c4126b02b5a69234aee8f91852b5bc5b97aa117f8315b024'
```

### DIFF
--- a/Casks/quickbooks-desktop.rb
+++ b/Casks/quickbooks-desktop.rb
@@ -1,0 +1,12 @@
+cask 'quickbooks-desktop' do
+  version '2016'
+  sha256 :no_check
+
+  url "http://http-download.intuit.com/http.intuit/Downloads/2016/Latest/QuickBooksMac#{version}.dmg"
+
+  name 'QuickBooks Desktop'
+  homepage 'http://quickbooks.intuit.com/mac/'
+  license :commercial
+
+  app "QuickBooks #{version}.app"
+end


adityadalal924:
What's the difference between this and quickbooks.intuit.com/macdownload/ ?

Casks/quickbooks-desktop.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'quickbooks' do
+  version '2016'
+  sha256 :no_check
+
+  url 'http://http-download.intuit.com/http.intuit/Downloads/2016/Latest/QuickBooksMac2016.dmg'
```
adityadalal924: Use double quotes (string interpolation) and #{version} instead of 2016.

Casks/quickbooks-desktop.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  name 'QuickBooks Desktop'
+  homepage 'http://quickbooks.intuit.com/mac/'
+  license :commercial
+
+  app 'QuickBooks 2016.app'
```
adityadalal924: Use double quotes (string interpolation) and #{version} instead of 2016.

sscotth:
http://quickbooks.intuit.com/mac/#compare

Quickbooks has two mac versions. One is a connivence app for their Quickbooks online service. The other is their standalone full edition which they sell every year. This cask is for the "Quickbooks Desktop" edition.

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in a346afa592a177d0f3fcf49fad46a22c78b5ce38. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

sscotth:
Missed the other 2016, thanks.

PR #15923, opened 2015-12-18
Upgrade TextExpander to 5.1.3

sirn:
Since previous version, TextExpander has switched its URL for newer versions to use `dl.smilesoftware.com/bundle_id/version/timestamp/file`. I'm not really sure if there's a way to avoid hard-coding the timestamp, so the URL is being hard-coded now.

### DIFF
--- a/Casks/textexpander.rb
+++ b/Casks/textexpander.rb
@@ -20,11 +20,11 @@ cask 'textexpander' do
     sha256 'ec90d6bd2e76bd14c0ca706d255c9673288f406b772e5ae6022e2dbe27848ee9'
     url "http://cdn.smilesoftware.com/TextExpander_#{version}.zip"
   else
-    version '5.1.1'
-    sha256 '21cc3f6e2145a7bfe5feea24b1afbb52c63ac62337eb639a9f52905840069b9e'
-    url "http://cdn.smilesoftware.com/TextExpander_#{version}.zip"
+    version '513.0-1446607872'
+    sha256 '28873bc18f4d62eda453e31609966d971f8165691ec4b5f3a8a37382cec5145c'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Please [squash your commits](http://davidwalsh.name/squash-commits-git).

adityadalal924:
Version and url will need to be updated to a style similar to https://github.com/caskroom/homebrew-cask/blob/master/Casks/imazing.rb - both use similar sparkle appcast style URLs.

Let us know if you need help.

sirn:
Hi, I've updated the URL as well as squash & rebase the commit.

Casks/textexpander.rb
```diff
@@ -20,11 +20,11 @@
     sha256 'ec90d6bd2e76bd14c0ca706d255c9673288f406b772e5ae6022e2dbe27848ee9'
     url "http://cdn.smilesoftware.com/TextExpander_#{version}.zip"
   else
-    version '5.1.1'
-    sha256 '21cc3f6e2145a7bfe5feea24b1afbb52c63ac62337eb639a9f52905840069b9e'
-    url "http://cdn.smilesoftware.com/TextExpander_#{version}.zip"
+    version '513.0-1446607872'
+    sha256 '28873bc18f4d62eda453e31609966d971f8165691ec4b5f3a8a37382cec5145c'
+    url "https://dl.smilesoftware.com/com.smileonmymac.textexpander/#{version.sub(%r{-.*},'')}/#{version.sub(%r{.*?-},'')}/TextExpander-#{version.sub(%r{-.*},'')}.zip"
```
vitorgalvao: What’s the `?` in `#{version.sub(%r{.*?-},'')}` there for?

sirn: OK, will fix. I copied this from imazing.rb above, but turns out it makes no difference whether the `?` is there or not.

PR #15925, opened 2015-12-18
Add rubocop-cask plugin for analyzing cask files

jawshooah:
I've written a [RuboCop plugin](https://github.com/jawshooah/rubocop-cask) with custom cops for linting casks.

Currently it's at version 0.0.0 with only one cop (`NoDslVersion`), but I figured we may as well take advantage of it.

I've also updated various dependencies, and updated the RuboCop configs to handle recently added cops.


jawshooah:
Here's some example output from the plugin:

```bash
$ vim Casks/010-editor.rb # revert header to legacy version
$ bundle exec rubocop Casks/010-editor.rb
Inspecting 1 file
C

Offenses:

Casks/010-editor.rb:1:6: C: Use cask '010-editor' instead of cask :v1 => '010-editor'
cask :v1 => '010-editor' do
     ^^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected
```

PR #15927, opened 2015-12-18
fix openoffice.rb

daehyunpy:
I fixed sha256.


vitorgalvao:
Thank you for the contribution. It was merged directly as 698ccf5b0a521405bbd887b3358582f7879a1d79 to keep commit history cleaner. Your contribution is still credited to you.

PR #8613, opened 2015-01-03
mono-mdk: Cask is updated in-place, use sha256 :no_check

jawshooah:
Closes #8610.

Issue #8610, opened 2015-01-03
brew cask install mono-mdk gives SHA2 mismatch


mercer:
When attempting to install mono-mdk, I get

```
➜  ~  brew cask install mono-mdk
==> Downloading http://download.mono-project.com/archive/3.10.0/macos-10-x86/MonoFramework-MDK-3.10.0.macos10.xamarin.x86.pkg
######################################################################## 100.0%
==> Note: running "brew update" may fix checksum errors
Error: SHA2 mismatch
Expected: e71b558c2b7857c124ccbb5e73bc622733e1cd5da494a7b5d9d14a90f555c1b8
Actual: dccf2478906689221788d6c1799acb05ee41502c4316339bccd8727e092fd868
Archive: /Library/Caches/Homebrew/mono-mdk-3.10.0.pkg
To retry an incomplete download, remove the file above.
```

PR #15931, opened 2015-12-18
Upgrade Mono-MDK package to 4.2.1.102 and switch to universal/x64

outlyer:
- Upgrade to latest upstream version of mono (4.2.1.102)
- Switch to the Universal Package instead of the 32-bit/x86 version
- pkgutil removes both the MDK and MRE which are installed by this package
- shasum is specific to this version so it should be verified

### DIFF
--- a/Casks/mono-mdk.rb
+++ b/Casks/mono-mdk.rb
@@ -1,13 +1,13 @@
-cask 'mono-mdk' do
+cask :v1 => 'mono-mdk' do
   version '4.2.1.102'
-  sha256 :no_check # required as upstream package is updated in-place
+  sha256 'ce886f20e509544cfc4379d10a0e4c574b06f0f97ae09096a06aee0c769e1f48'
 
-  url "http://download.mono-project.com/archive/#{version.sub(%r{\.[^.]*$},'')}/macos-10-x86/MonoFramework-MDK-#{version}.macos10.xamarin.x86.pkg"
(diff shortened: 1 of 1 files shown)


Casks/mono-mdk.rb
```diff
@@ -1,13 +1,13 @@
-cask 'mono-mdk' do
+cask :v1 => 'mono-mdk' do
```
vitorgalvao: Please remove the `:v1 => ` part. It’s no longer needed.

Casks/mono-mdk.rb
```diff
@@ -1,13 +1,13 @@
...
   version '4.2.1.102'
-  sha256 :no_check # required as upstream package is updated in-place
+  sha256 'ce886f20e509544cfc4379d10a0e4c574b06f0f97ae09096a06aee0c769e1f48'
```
vitorgalvao: And are you sure it won’t break? Because everything suggests it will.

outlyer:
I wasn't clear on why they would. The link is a direct link to the relevant archive, and the sum would only change if the version changed.

Each release on Github has it's own checksum from what I can tell. 

See here: https://github.com/mono/mono/releases

jawshooah:
We changed it to `:no_check` in #8613 because they *did* update the file in-place, as indicated by #8610. Whether or not they've done it again since, I'm not sure.

I'm fine with reinstating the shasum. If there are problems later, we can always switch back.

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in 95a6e27f3c09e25cd56e24861010e51604ed485b. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

reitermarkus:
Just tried updating, the shasum changed for me.

Issue #15753, opened 2015-12-10
wch-ch34x-usb-serial-driver: Mirror; Yosemite & El Capitan compatibility


tjanson:
The [WCH CH34x USB Serial driver cask](https://github.com/caskroom/homebrew-cask/blob/master/Casks/wch-ch34x-usb-serial-driver.rb) has two problems:

1. The website from which the pkg is downloaded is often slow or entirely unavailable, and has been for months. I suppose this is one of those cases where a mirror would be appropriate (like [this one](http://kiguino.moos.io/downloads/CH341SER_MAC.ZIP) on @kigster's blog), right?

2. The driver does not work "out of the cask" on either Yosemite or El Capitan, apparently because it is unsigned. The typical workaround is [disabling the signature check (or using an alternative, signed driver)](http://kiguino.moos.io/2014/12/31/how-to-use-arduino-nano-mini-pro-with-CH340G-on-mac-osx-yosemite.html) (again @kigster's blog).
That doesn’t sound like something this cask should do, but neither should the install essentially fail silently. Maybe we should add a notice to the `caveats`?

adityadalal924:
We can pull this cask from the main repo and put it into `caskroom/unofficial` with the changes you have suggested. Then, you could `brew cask install caskroom/unofficial/wch-ch34x-usb-serial-driver` and have things work with no further issues.

Ideally, all Casks in this repo should work out of the box on the latest OS X, so it might make sense to tombstone this cask, anyway.

adityadalal924:
@xyb I see you originally created the Cask, thoughts on the above discussion?

tjanson:
To be honest, I’m not sure that "Please disable kernel extension signature check" should *ever* be part of a cask. (I have no idea what the security implications are.) But I’ve added `caveats` to the cask that at least mention that this is the problem (see PR above).

As for the signed driver that’s making the rounds: It works fine, but there is no official download. (Apparently is was obtained via email from one of the involved companies but never officially published.) It might be a fit for `caskroom/unofficial`, though it still makes me uncomfortable considering that we’re talking about a kext from an unknown source.

Also, I don’t know how to write a formula. [Here’s an attempt, but I’m 90% sure the uninstall stanza is wrong.](https://gist.github.com/tjanson/e1b52d2218782998589c)

xyb:
[...]

PR #15954, opened 2015-12-19
Add compatibility note to wch-ch34x-usb-serial-driver

tjanson:
see also #15753 (does _not_ fix that issue)


adityadalal924:
We actually don't want Casks that don't work on the latest OS X - it makes maintaining things difficult.
Will comment in the main issue about the replacement cask that can be put in `unofficial`.

Cask removed in https://github.com/caskroom/homebrew-cask/pull/15958

PR #15955, opened 2015-12-19
Add Kivuto Secure Download Manager

tjanson:
This is a cask for the so-called Secure Download Manager that Microsoft requires people to use to download software via Dreamspark, their education program (formerly MSDNAA); though the program is actually from a company called Kivuto.

Since this is my first contribution, I’m not sure whether I got this right (even though it seems to work fine for me). In particular, I’m unsure about the uninstall stanza. Please check and test, I’d be glad to correct any problems.

Also, somewhat unrelated: Why is there no description (or even just keyword) field – was this a deliberate decision? Seems like it would be helpful to describe the program in a sentence or two (like I did above); this would also improve searchability. E.g., in this case, I’d like to mention "Dreamspark downloader" or something.

### DIFF
--- a/Casks/sdm.rb
+++ b/Casks/sdm.rb
@@ -1,0 +1,15 @@
+cask 'sdm' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://e5.onthehub.com/Static/Installers/SDM.pkg'
+  name 'Secure Download Manager'
+  name 'SDM'
+  name 'Kivuto Secure Download Manager'
+  license :gratis
+  homepage 'https://e5.onthehub.com/WebStore/Account/SdmDownloadFaq.aspx?ws=2cbc6a33-25f4-e011-8e6c-f04da23e67f6&secure=1'
+
+  pkg 'SDM.pkg'
+
+  uninstall :pkgutil => 'e-academy.com.secureDownloadManager.SecureDownloadManager.pkg'
+end


adityadalal924:
Re: No keywords, [we're not a discoverability service](https://github.com/caskroom/homebrew-cask/issues/4323#issuecomment-67257513).

I'll comment about the Cask-related chages when I get a chance to look through it.

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
+cask 'secure-download-manager' do
+  # Kivuto Secure Download Manager (SDM) is the download manager used by
+  # Microsoft Dreamspark (formerly MSDNAA).
+  # FIXME: Isn't there a `description` field?
+  version :latest
+  sha256 :no_check
+
+  # OnTheHub.com belongs to Kivuto Solutions Inc.
+  url 'https://e5.onthehub.com/Static/Installers/SDM.pkg'
+  name ['Secure Download Manager (SDM)', 'Kivuto Secure Download Manager']
```
vitorgalvao: Instead of a single array, provide this as multiple lines:

```
name 'Secure Download Manager'
name 'SDM'
name 'Kivuto Secure Download Manager'
```

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
+cask 'secure-download-manager' do
+  # Kivuto Secure Download Manager (SDM) is the download manager used by
```
vitorgalvao: Remove all of this comments

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
...
+  # Microsoft Dreamspark (formerly MSDNAA).
+  # FIXME: Isn't there a `description` field?
+  version :latest
+  sha256 :no_check
+
+  # OnTheHub.com belongs to Kivuto Solutions Inc.
```
vitorgalvao: And this one.

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
...
+  url 'https://e5.onthehub.com/Static/Installers/SDM.pkg'
+  name ['Secure Download Manager (SDM)', 'Kivuto Secure Download Manager']
+  license :closed
+  # OnTheHub SDM FAQ page in lieu of an actual website
```
vitorgalvao: And this one.

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
...
+  # Kivuto Secure Download Manager (SDM) is the download manager used by
+  # Microsoft Dreamspark (formerly MSDNAA).
+  # FIXME: Isn't there a `description` field?
+  version :latest
+  sha256 :no_check
+
+  # OnTheHub.com belongs to Kivuto Solutions Inc.
+  url 'https://e5.onthehub.com/Static/Installers/SDM.pkg'
+  name ['Secure Download Manager (SDM)', 'Kivuto Secure Download Manager']
+  license :closed
```
vitorgalvao: Be more specific. Is it free to use without limitations (no trial, not freemium)? The it is `:gratis`.

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
...
+  # OnTheHub SDM FAQ page in lieu of an actual website
+  homepage 'https://e5.onthehub.com/WebStore/Account/SdmDownloadFaq.aspx?ws=2cbc6a33-25f4-e011-8e6c-f04da23e67f6&secure=1'
+
+  pkg 'SDM.pkg'
+
+  uninstall :pkgutil => ['com.E-ACADEMY.SecureDownloadManager', 'SecureDownloadManager.app']
```
vitorgalvao: `SecureDownloadManager.app` isn’t an appropriate value. ave you seen the [scripts](https://github.com/caskroom/homebrew-cask/tree/master/developer/bin) and [documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#uninstall-stanza-details) geared for `uninstall`?

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
...
+  # FIXME: Isn't there a `description` field?
+  version :latest
+  sha256 :no_check
+
+  # OnTheHub.com belongs to Kivuto Solutions Inc.
+  url 'https://e5.onthehub.com/Static/Installers/SDM.pkg'
+  name ['Secure Download Manager (SDM)', 'Kivuto Secure Download Manager']
+  license :closed
+  # OnTheHub SDM FAQ page in lieu of an actual website
+  homepage 'https://e5.onthehub.com/WebStore/Account/SdmDownloadFaq.aspx?ws=2cbc6a33-25f4-e011-8e6c-f04da23e67f6&secure=1'
```
vitorgalvao: Isn’t there a better homepage? Isn’t https://imss.caltech.edu/node/862 one?

Casks/secure-download-manager.rb
```diff
@@ -0,0 +1,18 @@
+cask 'secure-download-manager' do
```
vitorgalvao: This should really be changed to `cask 'sdm' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `sdm.rb`.

vitorgalvao:
We always review before merging. No worries there.

vitorgalvao:
Could you expand a bit more on the app? Looking at it, it seems like these downloads aren’t exactly official. It seems like Kivuto provides the app for various companies/services, and they redistribute it according to their needs. Is that about right?

Casks/secure-download-manager.rb
adityadalal924: Only, an FAQ page, unfortunately.

Casks/secure-download-manager.rb
adityadalal924: From UninstallPKG:
```
Packagename: e-academy.com.secureDownloadManager.Applications.pkg
Packagefilename: applications.pkg
```
and

```
Packagename: e-academy.com.secureDownloadManager.SecureDownloadManager.pkg
Packagefilename: securedownloadmanager.pkg
```

Casks/secure-download-manager.rb
reitermarkus: I used this just yesterday to download a Windows 7 ISO from Dreamspark. Once it's installed, it's called `SecureDownloadManager.app`, so it should be called `securedownloadmanager`, right?

tjanson: I changed it to `sdm.rb`, but you’re right, @reitermarkus, `generate_cask_token` proposes `securedownloadmanager`. What’s your opinion on that, @vitorgalvao? Too verbose?

Casks/secure-download-manager.rb
tjanson: Unfortunately there really isn’t. Kivuto has [a PDF manual on their website](http://kivuto.com/dreamspark/SecureDownloadManagerReferenceGuide_en.pdf), but that’s it.

I want to clarify that OnTheHub.com *is* "from Kivuto", it’s one of their products.

Casks/secure-download-manager.rb
tjanson: Thanks, that helped, but I’m still unsure: The .app has ID `com.E-ACADEMY.SecureDownloadManager`, the .pkg includes `e-academy.com.secureDownloadManager.Applications.pkg` and `e-academy.com.secureDownloadManager.SecureDownloadManager.pkg`, and it’s the last of the three that actually works (i.e., removes the .app). So that’s all I should include, correct?

tjanson:
Thanks for the help. I corrected the cask as you suggested. Let me know if there still anything I need to change.

-----
Re: what this software actually does:

OnTheHub.com is Kivuto’s educational software portal. They are commisioned by Microsoft, IBM & others to distribute software (e.g., Windows, MS Office) to members of their academic partners (e.g., university students). Microsoft’s *Dreamspark* (MSDNAA) is probably the most well-known program (and the reason I’m interested in the whole thing). The only way to download any software is via this proprietary download manager, SDM.

I can’t tell you what exactly the economic deals are behind the scenes or whether SDM is also used elsewhere, but SDM is provided free of charge and is usable without restrictions (within the limits of the license) on OnTheHub, which is directly run by Kivuto. So this is not a third-party – the download comes directly from the vendor.

Casks/secure-download-manager.rb
reitermarkus: I think the `applications.pkg` makes sure that the directory `~/Applications` is created. But `brew cask` installs `PKGs` to `/`, so this won't do anything beside removing the installer receipt because it can't remove `/Applications`.

adityadalal924: @reitermarkus is correct. You can add in a `pkgutil` for `e-academy.com.secureDownloadManager.Applications.pkg` under a zap stanza if you want, for completeness' sake.

adityadalal924:
@vitorgalvao This looks ready to merge, if you don't have any further concerns.

reitermarkus:
@adityadalal924, what about the token name? (https://github.com/caskroom/homebrew-cask/pull/15955#discussion_r48184633)

adityadalal924:
Ah, missed that because the diffs were collapsed. Yeah, should be `securedownloadmanager` IMO, as that's what the installed app is called.

reitermarkus:
And as you said, maybe add `e-academy.com.secureDownloadManager.Applications.pkg` for completeness' sake.

adityadalal924:
@reitermarkus updated my earlier comment just as you posted :wink:

reitermarkus:
:smile:

vitorgalvao:
Everything looks correct except for stanza order. We could also now collapse one of the `name`s as per the conversation in https://github.com/caskroom/homebrew-cask/issues/15971.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in b9c40bef50033b7f7f08c7cc13ea8e8314c04879. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #15958, opened 2015-12-19
Delete wch-ch34x-usb-serial-driver.rb

adidalal:
Ref: https://github.com/caskroom/homebrew-cask/issues/15753

Does not work on the latest OS X


adityadalal924:
Merged as 0f34e1e37bdc9e337afdd1b04d0319be7a72c2d3.

PR #15959, opened 2015-12-19
Remove plainview, can't find on website/appcast dead

Amorymeltzer:
I can't find any reference to it on their site, appcast link is dead (and has been since its inception here). Still, pinging @victorpopkov in case you know anything I don't.  Pending tweet [for more info](https://twitter.com/Amorymeltzer/status/678103139501805568).


victorpopkov:
I also think this should be removed. I remember was already checking this cask before few months ago and the situation was the same.

victorpopkov:
Thank you for the contribution. It was merged directly as f1f0e5c34ae4d18e7e09434b3c79f7ce0ecfef94 to keep commit history cleaner. Your contribution is still credited to you.

PR #15960, opened 2015-12-19
Fix empty appcast shasums

Amorymeltzer:
With #15959 this should remove all casks with currently invalid (empty hash) `appcast` shas.  Refs proposed audits in #15831.
- A few had their appcast urls changed (two raw.github -> releases.atom; 1 redirect)
- `robofont` wasn't happy with curl?  wget worked

### DIFF
--- a/Casks/ubersicht.rb
+++ b/Casks/ubersicht.rb
@@ -5,7 +5,7 @@ cask 'ubersicht' do
   url "http://tracesof.net/uebersicht/releases/Uebersicht-#{version}.app.zip"
   name 'Übersicht'
   appcast 'http://tracesof.net/uebersicht/updates.xml.rss',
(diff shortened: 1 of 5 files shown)


Casks/ubersicht.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
vitorgalvao: Do we need this? It shouldn’t be needed, particularly since we require ruby 2.0 as the minimum.

Amorymeltzer: No that's just emacs' `ruby-mode` being a smartypants upon `save-and-quit`.  Fix incoming...

vitorgalvao:
Merged as 86548ed285bb80787d8e78ed76fc5df05ddc8ba2.

PR #15969, opened 2015-12-19
Update appcast.rb: remove :format

adidalal:
Remove format from `appcast`

https://github.com/caskroom/homebrew-cask/issues/15908


adityadalal924:
Closing because documentation and tests not included in PR.

PR #15973, opened 2015-12-19
Added install instructions to README

bcomnes:
Please excuse me if this was omitted for a reason.  It seemed strange to me that the latest install instructions were not present in the README, so I added them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ It’s implemented as a `homebrew` [external command](https://github.com/Homebre
 [![Coverage Status](https://img.shields.io/coveralls/caskroom/homebrew-cask.svg)](https://coveralls.io/r/caskroom/homebrew-cask)
 [![Join the chat at https://gitter.im/caskroom/homebrew-cask](https://img.shields.io/badge/gitter-join%20chat-blue.svg)](https://gitter.im/caskroom/homebrew-cask)
 
+## Get Cask
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Install instructions were purposefully removed because they are no longer needed. Currently, as long as you have homebrew installed, any `brew cask` command you run will automatically tap and install `homebrew-cask`.

bcomnes:
Cool!  Did not understand that.  Thanks for all the great work :)

jawshooah:
@vitorgalvao I think this is a good indicator that we should say *something* about installation in the README, even if it's just that no explicit installation is necessary.

bcomnes:
It just occured to me that these are the instructions at the bottom of the web page too.  Maybe that should be change?

vitorgalvao:
> It just occured to me that these are the instructions at the bottom of the web page too. Maybe that should be change?

It was a [deliberate decision](https://github.com/caskroom/homebrew-cask/pull/15731#issuecomment-163430358).

In light of this issue, though, it seems like we might indeed benefit from it. In that case, we might benefit from having the same thing in the website and the README. Any suggestions? Personally, I’m torn on the best solution. `brew tap caskroom/cask` seems clearer and shorter than any explanation of “you don’t need to do anything”, but on the other hand making users think they need it bothers me a little.

PR #15977, opened 2015-12-20
ncar-ncl.rb: make it pass audit

vitorgalvao:
Not too happy with the solution, but it currently fails audit on El Capitan and this is the only way to make it pass. There is no version for `10.11`. Development is still going, so it should be a matter of time.


Amorymeltzer:
This is more for my own purposes, but given the `depends_on`, isn't your edit what it should look like, in an ideal world?  What's the issue with it?

vitorgalvao:
Yes, since we have the `depends_on`, this will work correctly. However, if you’re just skimming the cask, the `else` will make it seem like it works on anything above Mavericks, which isn’t (or is it? haven’t actually tried to do it) the case. Having it explicit in the conditional seems better somehow.

Amorymeltzer:
Gotcha, thanks :+1:

PR #15985, opened 2015-12-20
Update virtualbox and qqbrowser

ywfwj2008:


vitorgalvao:
In the future, please submit separate pull requests for each cask, as more often than not this type of pull request brings problems with one or more of them, and it halts the inclusion of the other ones.

PR #15994, opened 2015-12-20
import.io: latest

reitermarkus:
Closes https://github.com/caskroom/homebrew-cask/issues/15704.

### DIFF
--- a/Casks/importio.rb
+++ b/Casks/importio.rb
@@ -1,0 +1,18 @@
+cask 'importio' do
+  version :latest
+  sha256 :no_check
+
+  # cloudfront.net is the official download host per the vendor homepage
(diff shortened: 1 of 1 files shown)


reitermarkus:
The workaround is neede because the `PKG` installer would install the app to `~/Desktop`, or – in case case of `brew cask` – to `/Desktop`.

vitorgalvao:
They use a `pkg` just to move an `app`? I really dislike that trend some companies are following. Thank you for making this.

vitorgalvao:
Merged as 5eb5d82fd55d9bd6d17aef64dc1ce6e0dc5e3a2e.

reitermarkus:
Yeah, sadly. And c'mon, *who* moves an app to the desktop?

PR #16013, opened 2015-12-21
updated freac (20151122)

adidalal:


adityadalal924:
Merged as beb428507b6dd93d1951bf4527e0f392dc5137f9.

adityadalal924:
cask-repair ended up with a broken download and thus a bad shasum, fixed in https://github.com/caskroom/homebrew-cask/pull/16015

PR #16017, opened 2015-12-21
Upgrade Subtitle Master to v2.01

scribblemaniac:
The homepage redirects to their Github project, and the old download link was broken.


adityadalal924:
Merged as 2683aabce11f0608160eca22ac6d1a04e370b2d8.

PR #16019, opened 2015-12-21
Simple Sync 3.0.1

reitermarkus:
formerly Roomie Agent


adityadalal924:
As the appcast refers to the version as `Version 3.0.1 (232)`, it should be safe to put `3.0.1-232` in `:version` and update the download link accordingly.

reitermarkus:
Ok, should be good now.

adityadalal924:
Merged as ab8ef944f3cd9a43f041bc13ff9bf6e42aa30e96.

PR #16021, opened 2015-12-21
`installer` and `uninstall` make their `:script` targets executable

### LINKED ISSUE #13607: installer :script should make its target executable in advance
Just came across this while working on [anaconda](https://github.com/caskroom/homebrew-cask/pull/13362). The `installer :script` target wouldn’t run due to it not having permission to execute. [`set_permissions`](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#flight-mini-dsl) doesn’t work to fix this because the soonest it’s available is `postflight`, and this needed to be `preflight`. [The old way had to be used](https://github.com/caskroom/homebrew-cask/blob/a5e7b4345a2c947a198db21448c4e600a7c0a567/Casks/anaconda.rb#L13#L15).

Either way, it should be the responsibility of `installer :script` to make sure its target can be run.

vitorgalvao:
Closes https://github.com/caskroom/homebrew-cask/issues/13607.

`installer` and `uninstall` had a few syntax fixes and now make their `:script` targets executable before trying to execute them.

Also fixed two casks that previously required workarounds.

### DIFF
--- a/lib/hbc/artifact/installer.rb
+++ b/lib/hbc/artifact/installer.rb
@@ -26,10 +26,12 @@ class Hbc::Artifact::Installer < Hbc::Artifact::Base
                                                                         "#{self.class.artifact_dsl_key}",
                                                                         {:must_succeed => true, :sudo => true},
                                                                         {:print_stdout => true}
-                                                                        )
+                                                                       )
         ohai "Running #{self.class.artifact_dsl_key} script #{executable}"
         raise Hbc::CaskInvalidError.new(@cask, "#{self.class.artifact_dsl_key} missing executable") if executable.nil?
-        @command.run(@cask.staged_path.join(executable), script_arguments)
+        executable_path = @cask.staged_path.join(executable)
+        @command.run('/bin/chmod', :args => ['+x', executable_path]) if File.exists?(executable_path)
+        @command.run(executable_path, script_arguments)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/hbc/artifact/installer.rb
```diff
@@ -26,10 +26,12 @@ def install_phase
                                                                         "#{self.class.artifact_dsl_key}",
                                                                         {:must_succeed => true, :sudo => true},
                                                                         {:print_stdout => true}
-                                                                        )
+                                                                       )
         ohai "Running #{self.class.artifact_dsl_key} script #{executable}"
         raise Hbc::CaskInvalidError.new(@cask, "#{self.class.artifact_dsl_key} missing executable") if executable.nil?
-        @command.run(@cask.staged_path.join(executable), script_arguments)
+        executable_path = @cask.staged_path.join(executable)
+        @command.run('chmod', :args => ['+x', executable_path]) if File.exists?(executable_path)
```
jawshooah: Best to use `/bin/chmod` so there are no surprises.

lib/hbc/artifact/uninstall_base.rb
```diff
@@ -225,15 +225,18 @@ def warn_for_unknown_directives(directives)
...
                                                                     {:print_stdout => true},
-                                                                    :early_script)
-
+                                                                    :early_script
+                                                                   )
     ohai "Running uninstall script #{executable}"
     raise Hbc::CaskInvalidError.new(@cask, "#{stanza} :early_script without :executable") if executable.nil?
-    @command.run(@cask.staged_path.join(executable), script_arguments)
+    executable_path = @cask.staged_path.join(executable)
+    @command.run('chmod', :args => ['+x', executable_path]) if File.exists?(executable_path)
```
jawshooah: Same here.

lib/hbc/artifact/installer.rb
vitorgalvao: Good point. Done.

jawshooah:
LGTM :+1:

PR #16035, opened 2015-12-22
New mysql-utilities 1.5.6 cask

hanxue:


vitorgalvao:
Thank you for the contribution. It was merged directly as 8377e9d71f9679d60c6d9d824086a8216a06712a to keep commit history cleaner. Your contribution is still credited to you.

hanxue:
Thanks for the merge @vitorgalvao . Asking so that I can contribute more effectively: how can I submit a pull request that you (comitters) can merge directly? For example: will running a pull + rebase right before I commit, push and send in a pull-request help?

adityadalal924:
@hanxue You're doing things the right way and don't need to change. We use a [script](https://github.com/vitorgalvao/tiny-scripts/blob/master/fastmerge) to merge PRs generally, in order to keep the homebrew-cask commit history cleaner.

Really the only time where different behavior is needed is when we suggest changes and you make them - squashing commits (`git rebase -i origin/master`) lets us merge a bit faster.

PR #16042, opened 2015-12-22
verify.rb: Warning on sha :no_check

adidalal:
Fixes https://github.com/caskroom/homebrew-cask/issues/7062

Wasn't sure if I should toss up an "error" instead, as we're not _doing_ anything about a shasum of `:no_check` (yet).

Currently not working but I find discussing on a PR (with code) more productive than with an issue, anyway.

### DIFF
--- a/lib/hbc/verify.rb
+++ b/lib/hbc/verify.rb
@@ -10,7 +10,10 @@ module Hbc::Verify
       expected = cask.sha256
     rescue RuntimeError => e
     end
-    return if expected == :no_check
+    if expected == :no_check
+      ohai `Checksum not available`
+      return
+    end
     computed = Digest::SHA2.file(path).hexdigest
     if expected.nil? or expected.empty?
       raise Hbc::CaskSha256MissingError.new("sha256 required: sha256 '#{computed}'")


lib/hbc/verify.rb
```diff
@@ -10,7 +10,10 @@ def self.checksum(path, cask)
       expected = cask.sha256
     rescue RuntimeError => e
     end
-    return if expected == :no_check
+    if expected == :no_check
+      ohai `Checksum not available`
```
jawshooah: These should be single-quotes, not backticks.

jawshooah:
I'm fine with this once the errors are fixed.

adityadalal924:
@jawshooah Is there an error type I should be raising instead? Will fix the backticks before merge.

jawshooah:
I don't think so. Why should we error out? It's expected that some casks will have `sha256 :no_check`.

adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in 9c6b40808e5a35898d572c20ce0465cc29583911. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #16053, opened 2015-12-22
Update MacID to 1.3.1.3

stqism:
The worst part is the version is technically 1.3.1.3.0 but in downloads they strip the last .0


vitorgalvao:
Thank you for the contribution. It was merged directly as c6a10e327e7e1401306496e8583b9af997a994bb to keep commit history cleaner. Your contribution is still credited to you.

PR #16059, opened 2015-12-23
Tower 2.3.3

reitermarkus:


adityadalal924:
Merged as e52b3a476bb0e00b57c66f62e8af667f8c1cbf53.

adityadalal924:
FYI, not an immediate concern, but we're going to try to keep appcast shasums up-to-date from now on as well, in line with https://github.com/caskroom/homebrew-cask/issues/15908.

The new cask-repair script handles these for you as well (if you use it).

reitermarkus:
Thanks, I'll keep that in mind for next time.

PR #16060, opened 2015-12-23
added cask for dymo-stamps

liljenstolpe:


Casks/dymo-stamps.rb
```diff
@@ -0,0 +1,13 @@
+cask 'dymo-stamps' do
+  version :latest  # Upstream application upgraded in place
```
adityadalal924: No need for this comment.

Casks/dymo-stamps.rb
```diff
@@ -0,0 +1,13 @@
...
+  sha256 :no_check
+
+  url "http://download.endicia.com/dymostamps/dymostamps.dmg"
```
adityadalal924: Casks/dymo-stamps.rb:5:7: C: Prefer single-quoted strings when you don't need string interpolation or special symbols.

Casks/dymo-stamps.rb
liljenstolpe: I did that, as I was asked to by another reviewer on another cask.  Sorry.

liljenstolpe:
Will clean-up and re-submit

PR #16063, opened 2015-12-23
Add CloudApp.app v3.7.0

pstauffer:
Adding a new cask for CloudApp in version 3.7.0

### DIFF
--- a/Casks/cloudapp.rb
+++ b/Casks/cloudapp.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'cloudapp' do
+  version '3.7.0'
+  sha256 '29b6ad9938c76e9007a66f156922777fd5ec3deb6ac6d3b3a6d147142d07cb9a'
+
+  url "https://s3.amazonaws.com/downloads.getcloudapp.com/mac/CloudApp-#{version}.dmg"
+  name 'CloudApp'
+  homepage 'http://www.getcloudapp.com/'
+  license :commercial
+
+  app 'CloudApp.app'
+end


vitorgalvao:
Read the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md). It must be followed, to avoid duplicates.

[This already exists](https://github.com/caskroom/homebrew-cask/blob/master/Casks/cloud.rb).

Casks/cloudapp.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'cloudapp' do
+  version '3.7.0'
+  sha256 '29b6ad9938c76e9007a66f156922777fd5ec3deb6ac6d3b3a6d147142d07cb9a'
+
+  url "https://s3.amazonaws.com/downloads.getcloudapp.com/mac/CloudApp-#{version}.dmg"
+  name 'CloudApp'
+  homepage 'http://www.getcloudapp.com/'
+  license :commercial
```
vitorgalvao: It’s not commercial, as you do not pay for the app.

Casks/cloudapp.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'cloudapp' do
```
vitorgalvao: This header is wrong. We no longer use `:v1 =>` (it is removed).

pstauffer:
Hi there,

thanks for the feedback and the link to the already existing cask. In my oppinion is the already existing cask token wrong, because the app is named "CloudApp", so the cask token should be "cloudapp" and not "cloud".
But anyway, I will use your feedback, if I create another cask.

Thanks and Greets
Pascal

vitorgalvao:
> In my oppinion is the already existing cask token wrong, because the app is named "CloudApp", so the cask token should be "cloudapp" and not "cloud".

This exact case has been discussed before. We need consistency and strict rules, because we know what a repo filled with duplication looks like, and it is not good. “App” is a common suffix that doesn’t matter most of the time and makes more sense to be removed. CloudApp isn’t special (and has an arguably weak name) and is the only app where this rule causes some occasional doubts. Having duplicates isn’t worth the exception.

PR #16064, opened 2015-12-23
Update Ableton Live, add Ableton Live Lite

epsimatic:


Casks/ableton-live-lite.rb
```diff
@@ -0,0 +1,26 @@
...
+    url "http://cdn2-downloads.ableton.com/channels/#{version}/ableton_live_lite_#{version}_32.dmg"
+  else
+    sha256 '054895bf3593bf5849921d3f83791602accc19d6157e8408f369352e3bb2cf56'
+    url "http://cdn2-downloads.ableton.com/channels/#{version}/ableton_live_lite_#{version}_64.dmg"
+  end
+
+  name 'Ableton Live Lite'
+  homepage 'https://www.ableton.com/en/products/live-lite/'
+  license :commercial
+  caveats <<-EOT
```
vitorgalvao: Remove the caveat since it is not related to installation.

vitorgalvao:
Please submit separate pull requests for each cask, as more often than not this type of pull request brings problems with one or more of them, and it halts the inclusion of the other ones.

In this case, the lite version belongs in [caskroom/versions](https://github.com/caskroom/homebrew-versions).

Casks/ableton-live-lite.rb
```diff
@@ -0,0 +1,26 @@
...
+This app is not sold separately.
+It comes bundled with hardware and needs a product key from that hardware's product card.
+General public should use ableton-live instead.
+EOT
+
+  app "Ableton Live #{version.to_i} Lite.app"
+
+  zap :delete => '~/Library/*/*[Aa]bleton*',
+      :rmdir => '~/Music/Ableton/Factory Packs'
+      #:trash => '~/Music/Ableton/User Library'
```
vitorgalvao: This doesn’t really make sense, here. `zap` is to throw away configurations, not work.

epsimatic:
> In this case, the lite version belongs in caskroom/versions.

it's more a edition, not a version. And there are 4 editions of Ableton Live:

— ableton-live-lite (the cask in question) (bundled with hardware only)
— ableton-live 
— ableton-live-standard
— ableton-live-suite

Should any of the others go to homebrew-versions too?

> This doesn’t really make sense, here. zap is to throw away configurations, not work.

I've copied it from the ableton-live cask. Guess that one needs to be fixed too.

> Remove the caveat since it is not related to installation.

App won't work without a license key. There's a free_license(web_page) method in Caveats Mini-DSL related to licensing but unrelated to installation as well, so I guessed it's the right way to tell about licensing caveats.

Anyway, will fix and re-submit things later today.

vitorgalvao:
> it's more a edition, not a version

Same thing as per [the rules](https://github.com/caskroom/homebrew-versions#acceptable-casks) (fifth point in the bulleted list). Actually, “edition” is really a better name for it, will change it to that. Thank you.

> Should any of the others go to homebrew-versions too?

All of them except the main one (`ableton-live`).

> I've copied it from the ableton-live cask. Guess that one needs to be fixed too.

Yes. Would you mind doing it, please?

> App won't work without a license key.

And you should know that before installing it with homebrew-cask. We’re not a discoverability service.

> There's a free_license(web_page) method in Caveats Mini-DSL related to licensing but unrelated to installation as well

There’s a reason that’s a method: it’s common enough. It is also *always* used with `discontinued`, which does pertain to installation (it warns you it may or may not work correctly).

epsimatic:
> Would you mind doing it, please?

Yep, will do everything we discussed in several hours.

There will be several atomic pull requests in both repos.

vitorgalvao:
Thank you.

joshka:
@vitorgalvao this rule about versions / editions feels a little restrictive (IMO) and it would seem from the [history of people attempting to add Ableton Suite](https://github.com/caskroom/homebrew-cask/search?q=ableton&type=Issues&utf8=✓) to this repo, rather than checking the versions tap, that others agree.

In situations where there's no obvious 'primary stable' version (or edition), I wonder whether it makes sense to relax this rule a little and allow multiple versions to hit the main repo. I know I wasted time creating the same ableton-live-suite casks that others have already done before realizing that it was already in versions.

I may be beating a dead horse on this issue. Apologies if so. I won't be offended if you say 'it's been decided, let's just leave it'. Alternatively, if you want to start a new issue for for comment, happy to go that direction also.

vitorgalvao:
Behind every strict rule is a prior time where the repo was a serious mess. Strict rules are reactionary, and they have been working. That means it’s unlikely we’ll reverse this one.

However, your point is perfectly valid and we’re aware of the problem. The solution is to make so that when you `brew cask search`, it’ll also check `caskroom/versions`, even if you don’t have it tapped. If you `brew search` they actually check our repos, but we (`cask`) still don’t.

There’s an open issue about this, but I’m now on a phone, so can’t easily check.

joshka:
I've seen the issue you mention - no need to add it here. Will go with the flow on it based on what you're saying here.

PR #16067, opened 2015-12-23
Add Sublime Text 3 Build 3083

cilq:
It is basically a copy of the Sublime Text 2 cask but adapted to the current stable Beta release.


vitorgalvao:
Betas belong in [caskroom/versions](https://github.com/caskroom/homebrew-versions). This cask is already there.

PR #16069, opened 2015-12-23
Remove CX.app (not available)

victorpopkov:
The homepage and download URL is down for few months already.


vitorgalvao:
Merged as 390ec14e3425a4b67ed3e0c0761cb6f54e61b0a2.

PR #16070, opened 2015-12-23
Remove Hall.pkg (not available)

victorpopkov:
The homepage and download URL are down for few months already. Hall has joined HipChat, so won't be available anymore: https://blog.hipchat.com/2015/05/06/hall-joins-hipchat/


vitorgalvao:
Merged as 46434d86205d1d70978e9dc9e9bd0c1337e0ab23.

PR #16071, opened 2015-12-23
Add discontinued caveat for Latexian.app

victorpopkov:
Taco Software is closed as it says on their [homepage](http://tacosw.com/). The download URL is still available, though.


vitorgalvao:
I think we should remove this.

It’s one thing when the app’s homepage says “we’re discontinuing this, but here, you can still download it”, and quite another when they say “this is no longer available” but the download still works. The former shows some commitment, while the latter indicates the url will stop working at any time. Particularly here, where the whole company has gone under, this should really be removed.

victorpopkov:
I also agree. Will prepare the new PR.

PR #16072, opened 2015-12-23
add virtualbox binaries

adidalal:
Noticed they were missing from a related cask over in homebrew-versions

### DIFF
--- a/Casks/virtualbox.rb
+++ b/Casks/virtualbox.rb
@@ -14,6 +14,8 @@ cask 'virtualbox' do
   tags :vendor => 'Oracle'
 
   pkg 'VirtualBox.pkg'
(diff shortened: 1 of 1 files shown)


adityadalal924:
Merged as 06cd14f141ae380ba5b4cfe8f4f04612c254f15a.

jawshooah:
Does `pkgutil` no longer [install these for us?](https://github.com/caskroom/homebrew-cask/commit/7acae2f145a7b6322d6cdb4c60384039d66e1679)

adityadalal924:
@jawshooah Nice catch - didn't realize virtualbox did that, but can confirm after taking a look with pacifist that it is indeed the case.

Reverted.

PR #16075, opened 2015-12-23
Remove MacTubes.app (discontinued)

victorpopkov:
Not available as it doesn't work anymore. The author has stopped releasing this application and download URLs were removed from the homepage: http://macapps.sakura.ne.jp/mactubes/index_en.html


adityadalal924:
Merged as 32969cae285610e1c0071ae28ee9f9ef436cdd99.

PR #16078, opened 2015-12-23
Remove younited.app (discontinued)

victorpopkov:
It's safe to finally remove this cask as all the younited account and their content were deleted permanently since October 30, 2015.


adityadalal924:
Merged as b10763d753d460e00007e2245abda6b5d81508d3.

PR #16079, opened 2015-12-23
add fonttong.rb

daehyunpy:
I added fonttong.rb for fonttong app

### DIFF
--- a/Casks/fonttong.rb
+++ b/Casks/fonttong.rb
@@ -1,0 +1,11 @@
+cask 'fonttong' do
+  version '2.0'
+  sha256 '70974dcdc39bb6f49a3440a265c3f6f87df0d5cf2b843e22fbd3120b9ce428a1'
+
+  url 'http://www.asiafont.com/FontTong.dmg'
+  name '폰트통'
+  homepage 'http://www.asiafont.com'
+  license :gratis
+
+  app "폰트통v#{version}.app"
+end


Casks/fonttong.rb
```diff
@@ -0,0 +1,11 @@
+cask 'fonttong' do
+  version '2.0'
+  sha256 '70974dcdc39bb6f49a3440a265c3f6f87df0d5cf2b843e22fbd3120b9ce428a1'
+
+  url 'http://www.asiafont.com/FontTong.dmg'
+  name 'FontTong'
```
vitorgalvao: The original name should also be added.

Casks/fonttong.rb
```diff
@@ -0,0 +1,11 @@
...
+  homepage 'http://www.asiafont.com'
+  license :gratis
+
+  app "폰트통v#{version}.app", :target => 'FontTong.app'
```
vitorgalvao: Why are you renaming it? This doesn’t seem to be [a case for it](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#target-should-only-be-used-in-select-cases).

hyuna917:
I added original name. and I renamed it and now it is easy to run with Spotlight and shell.

vitorgalvao:
That’s not an acceptable case, though. The developer named the app in such a manner for a reason, and there’s no reason to change it. You’re making it easier for you, but perhaps harder for their target audience. Especially whith #13201, where this would rename the app itself, this makes no sense.

Please remove `:target`. If you need another name, you’ll need to do it in your setup, or ask the developer.

hyuna917:
I change the name to original.

vitorgalvao:
Thank you.

PR #16081, opened 2015-12-23
Provide caveats for Amazon Music

swrobel:
App sometimes won't launch after install and a zap fixes it per https://github.com/caskroom/homebrew-cask/issues/12173#issuecomment-161810926

### DIFF
--- a/Casks/amazon-music.rb
+++ b/Casks/amazon-music.rb
@@ -24,4 +24,14 @@ cask 'amazon-music' do
                   '~/Library/Preferences/com.amazon.music.plist',
                   '~/Library/Application Support/Amazon Music/'
                 ]
+
+  caveats <<-EOS.undent
+    If the app won't launch after installation, try
+
+      brew cask zap #{token}
+      brew cask install #{token}
+
+    then re-launch the app. You can read more about the issue on Amazon's customer forums
+    http://www.amazon.com/App-wont-open-OS-Yosemite/forum/FxZLHSK3AW6KZU/Tx1EJYW65OQ5TZS
+  EOS
 end


vitorgalvao:
Did you have this problem as well? Because this doesn’t seem like a common issue. Most apps can be “fixed” by getting rid of corrupt preferences. If this is a common occurrence, then the correct fix is to report the issue to Amazon so they can fix it.

vitorgalvao:
I’ll add, though, that we can indeed have this as a caveat, with an important modification. If you can report the issue to Amazon *publicly* in one of their forums or something where they can respond/mark as solved, somewhere where they can acknowledge this issue and the workaround and report they’re working on it, then it would be a valid caveat.

swrobel:
@victorgalvao I did have this issue as well, hence my desire to prevent others from running into the same thing. Unfortunately, it doesn't seem that Amazon addresses such issues in any sort of public form. I've reported the issue to them privately and it has been raised on their own [customer forums](http://www.amazon.com/wont-Yosemite-might-going-here/forum/FxZLHSK3AW6KZU/Tx1EJYW65OQ5TZS) and [stackoverflow](https://apple.stackexchange.com/questions/194558/amazon-music-app-wont-launch)

vitorgalvao:
That is perfect. Just please add that customer forums link to the caveat, so people have a frame of reference.

After you do, please also [squash your commits](http://davidwalsh.name/squash-commits-git).

jawshooah:
The caveat should probably also mention that you'll have to reinstall after zapping.

swrobel:
@jawshooah are you certain that a reinstall is needed? I don't believe it was in my case.

jawshooah:
Well, `zap` implicitly runs `uninstall`, so if the app remained after zapping, something is wrong.

swrobel:
@jawshooah of course ... ha! OK, will mention.

swrobel:
Updated!

adityadalal924:
Thank you for the contribution. It was merged directly as d06c60b690f7e41236b226637eb26d8d68ad624a to keep commit history cleaner. Your contribution is still credited to you.

PR #8749, opened 2015-01-08
Activate GPG verification upon install

tapeinosyne:
Refer to #5971.

This patchset lands GPG verification as a working feature, with a minimum of tests. GPG verification is automatically enabled if a working `gpg` binary is found on the system.

Most notably, two decisions are open for debate.
##### Failure handling

Handling of failed GPG verification is delegated from the `Hbc::GpgCheck` class to the caller. (As of this PR, the only caller is intended to be `Hbc::Verify.gpg_signature`.)

This is largely a matter of internal design, with no functional consequences.
##### CLI output

Standard `gpg` output is printed unmodified. Security features may benefit from a transparent UI, but I worry that the galore of lines and information is noise to most users, or otherwise irrelevant.

As an example, consider the output of `brew cask install libre-office`, sans our own GPG messages:

```
==> Downloading https://download.documentfoundation.org/libreoffice/stable/4.3.5
######################################################################## 100.0%
######################################################################## 100.0%
==> gpg: requesting key AFEEAEA3 from hkps server hkps.pool.sks-keyservers.net
==> gpg: key AFEEAEA3: "LibreOffice Build Team (CODE SIGNING KEY) <build@documen
==> gpg: Total number processed: 1
==> gpg:              unchanged: 1
==> gpg: Signature made Fri Dec 12 17:52:48 2014 GMT using RSA key ID AFEEAEA3
==> gpg: Good signature from "LibreOffice Build Team (CODE SIGNING KEY) <build@d
==> gpg: WARNING: This key is not certified with a trusted signature!
==> gpg:          There is no indication that the signature belongs to the owner
==> Primary key fingerprint: C283 9ECA D940 8FBE 9531  C3E9 F434 A1EF AFEE AEA3
==> Symlinking App 'LibreOffice.app' to '/Applications/LibreOffice.app'
```

(For the purpose of this discussion, ignore truncation as an unrelated issue.)

It is arguably desirable to silence `gpg` commands, as commonly done by other package managers, and favor more concise notifications for each step: key retrieval, signature download, and final check.

PR #16090, opened 2015-12-24
[WIP] Activate GPG verification upon install

jawshooah:
Picking up where @ndr-qef left off with #8749.

This is still a work-in-progress, so please feel free to throw your feedback at me.


fanquake:
Nice work. Glad this is getting some love again.

ndr-qef:
Thanks for taking up the patchset, @jawshooah.

Beyond refactoring, it might be best to improve the CLI output —that is, hide the rather verbose one of GPG— in this PR, rather than postponing UX changes to a later date.

adityadalal924:
A (rather simple) idea would just be to dump the output to `dev/null` altogether (unless the `--verbose` flag is used) and have a single line, similar to the shasum check, i.e. "GPG verification of Cask succeded/failed"

ndr-qef:
We can simply set `:print_stdout => false` when invoking system gpg with our `system_command` wrapper.

jawshooah:
@ndr-qef I'm not very familiar with GPG, but should we be using `gnupg2` instead of `gnupg`? Is the legacy version still supported?

jawshooah:
Also should we just add `depends_on :formula => 'gpg'` to casks with a `gpg` stanza?

adityadalal924:
I don't like adding that dependency to Casks, as gpg verification should be "best effort" - the notification that gpg verification was skipped (and to install gpg to activate it) should be enough

ndr-qef:
> should we be using gnupg2 instead of gnupg? Is the legacy version still supported?

GnuPG 1 is stable and maintained, as well as widely deployed — notably in the Debian package manager, where it is the standard version to this day.

In any case, the `gpg` command is meant to call the standard system GPG, whichever that may be. (Much like `git`, `ruby`, and so on.) As it happens, many OS X users install GPG through GPGTools, which ships version 2.

jawshooah:
Gotcha. Just wanted to make sure we weren't dating ourselves. I'll work on this more tonight.

dconnolly:
What's the plan for getting this merged? I'm looking forward to checking signatures without local patches on top.

jawshooah:
There's still a bit more work to be done before merging. I've been swamped for a while, but I'll try to get this out within the next few days.

dconnolly:
Awesome, thank you!

fanquake:
@jawshooah Any update on this?

fanquake:
@jawshooah Ping

jawshooah:
Sorry for the huge delay on this. Life has been dropping a hammer on me the last few months. I should have some time this evening to work on this, though.

fanquake:
@caskroom/maintainers Does anyone want to give any feedback here?

vitorgalvao:
Hadn’t noticed there were commits after the last comment. Is this ready @jawshooah?

jawshooah:
Should be functional, I've just been meaning to add more tests but haven't gotten around to it.

lib/hbc/verify/gpg.rb
```diff
@@ -3,12 +3,13 @@ def self.me?(cask)
     cask.gpg
   end
 
-  attr_reader :cask, :downloaded_path
+  attr_reader :cask, :downloaded_path, :force_fetch
 
-  def initialize(cask, downloaded_path, command=Hbc::SystemCommand)
+  def initialize(cask, downloaded_path, force_fetch=false, command=Hbc::SystemCommand)
```
claui: I think we should change the `force_fetch` and `command` parameters inside the signature into an options hash. Options hashes clean up signatures and result in code that is easier to read.

That way, we would also stay consistent to existing, similar initializers (example [1](https://github.com/caskroom/homebrew-cask/blob/0cac301be4faa63f8f2b5b4841eda1074425c458/lib/hbc/artifact/base.rb#L81-L85) [2](https://github.com/caskroom/homebrew-cask/blob/0cac301be4faa63f8f2b5b4841eda1074425c458/lib/hbc/installer.rb#L21-L26) [3](https://github.com/caskroom/homebrew-cask/blob/0cac301be4faa63f8f2b5b4841eda1074425c458/lib/hbc/system_command.rb#L34-L38)).

jawshooah: Since we now assume users are running Ruby >=2.0, why not use [keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments)?

claui: @jawshooah Thanks for pointing that out. Those would be an even better choice!

claui:
One more issue that needs to be discussed before we merge is error handling.

For example, as of right now

> https://get.videolan.org/vlc/2.2.4/macosx/vlc-2.2.4.dmg

is up, while at the same time

> https://get.videolan.org/vlc/2.2.4/macosx/vlc-2.2.4.dmg.asc

returns a HTTP 500 status:

> We're very sorry but...
> none of our mirrors are able to handle your request, please try again later

When I rebase this PR and then do `brew cask install vlc`, the following happens:

image

I think it’s perfectly acceptable for the whole installation to fail when the signature is unavailable. However, this should result in a clear error message indicating what happened, rather than just bailing out with a stack trace.

fanquake:
This needs a rebase.

fanquake:
ping @jawshooah Any chance you could rebase and update the error handling?

jawshooah:
@fanquake Hopefully I can get to this later tonight.

jawshooah:
Leaving this open since there is still interest, I just haven't gotten around to cleaning it up yet. Will try to migrate to Homebrew/brew later this evening.

jawshooah:
Continued in https://github.com/Homebrew/brew/pull/1335.

PR #16092, opened 2015-12-24
Install adobe-air before running tests in Travis

jawshooah:
This way we can run [integration tests](https://github.com/caskroom/homebrew-cask/blob/3407a19def04a81fe8c9d9ace6a8c2e25384927f/test/cask/installer_test.rb#L60-L70) for the Adobe AIR container type.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,13 @@ before_install:
   - echo ls_ruby_bindir; ls "/System/Library/Frameworks/Ruby.framework/Versions/${CASK_RUBY_TEST_VERSION}/usr/bin"
 
 # install steps
-# * brew Formulae without which some tests will be skipped
+# * brew Formulae and Casks without which some tests will be skipped
 # * bundler gem
 # * Ruby gems required for brew-cask
 install:
   - brew install cabextract
   - brew install unar
+  - cmd/brew-cask.rb install Casks/adobe-air.rb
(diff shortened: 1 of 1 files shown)


jawshooah:
The alternative would be to add `depends_on :cask => 'adobe-air'` to `test/support/Casks/adobe-air-container.rb`. Since all mainline casks using Adobe AIR containers have this dependency, it might be best to do the same in the test cask.

adityadalal924:
I'd be for the second option - the tests Casks should mirror actual usage as much as possible.
The runtime should be the same either way, I believe?

vitorgalvao:
I also agree the second option sounds best.

jawshooah:
Just had a thought - if we add the dependency to the cask, we're adding it for everyone who runs the tests locally as well. Leaving it in `.travis.yml` isolates it to the CI environment. What do you think?

adityadalal924:
The amount of people that run tests locally (and are not developers) are probably pretty slim anyways.

That being said if you think one way is better than the other, go for it. It doesn't sound like a huge problem one way or another.

vitorgalvao:
That’s a good point. I for one don’t want Adobe AIR anywhere near my system, so I consider it being isolated in Travis a plus.

Looking at it further, even, we also have `cabextract` and `unar`, there, so having those all together also makes sense.

PR #16093, opened 2015-12-24
Add constant Hbc::Container::Air::INSTALLER_PATHNAME

jawshooah:
Because DRY.

### DIFF
--- a/Casks/scratch.rb
+++ b/Casks/scratch.rb
@@ -12,7 +12,7 @@ cask 'scratch' do
             :sudo   => true
 
   uninstall :script => {
-              :executable => '/Applications/Utilities/Adobe AIR Application Installer.app/Contents/MacOS/Adobe AIR Application Installer',
+              :executable => Hbc::Container::Air::INSTALLER_PATHNAME,
               :args => %w[-uninstall -silent /Applications/Scratch\ 2.app]
             }
 
--- a/lib/hbc/container/air.rb
+++ b/lib/hbc/container/air.rb
@@ -1,11 +1,15 @@
 class Hbc::Container::Air < Hbc::Container::Base
+  INSTALLER_PATHNAME =
+    Pathname('/Applications/Utilities/Adobe AIR Application Installer.app' \
+             '/Contents/MacOS/Adobe AIR Application Installer')
+
   def self.me?(criteria)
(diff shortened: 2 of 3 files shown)


Amorymeltzer:
Will this make it easier for someone installing `scratch` to understand or follow?

jawshooah:
It won't make any difference in the output they see; the only people who will ever see it are those who make changes to the cask. The point is to remove unnecessary code duplication.

adityadalal924:
> we shouldn't be using interpolation, `gsub` with regexes, or other language features in Casks either

As much as I [love regular expressions](https://xkcd.com/208/), one of these days, I'll sit down and learn Ruby, for the sole purpose of fixing the version string and turning it into a more robust object. Some of those substitutions are horrifically unreadable :grinning:

jawshooah:
> Some of those substitutions are horrifically unreadable :grinning:

They really are. Wrapping the version string in a class is also on my to-do list. We should probably open an issue for it.

adityadalal924:
Just noticed https://github.com/caskroom/homebrew-cask/issues/12827 is exactly that

jawshooah:
@Amorymeltzer do you have any issues with this being merged?

Amorymeltzer:
Well, I think it's a bad idea for usability.  It'd be the first time ruby code is used for something other than a switch, completely obscuring the natural path, which is not, IMO, a great road to go down.

That being said, this is a weird, special case where we already have a built-in workaround, and it's not likely to make a huge difference to anyone's life, so I don't really mind.  I wouldn't say no to a single comment offering some modicum of explanation, but it's not a big deal.

vitorgalvao:
I certainly understand @Amorymeltzer’s reservations, but also think in this case it doesn’t really matter. There’s at least [one other cask that will benefit from this](https://github.com/caskroom/homebrew-cask/pull/16109/files), and it actually made me think if instead of this solution, having something like `uninstall :adobe-air-app => '/Applications/Scratch 2.app'` wouldn’t be both clearer and more useful, but to be honest I don’t think we should care in this case, and merge this as is. Adobe AIR will just get less and less popular over time, and apps made in it will start to die out, so it isn’t worth the time.

vitorgalvao:
Seems like we’re all fine with this, so I’ll go ahead and merge it.

PR #9255, opened 2015-01-26
Added caveats for discontinued apps and apps that provide free licenses

vitorgalvao:
Should make it easier to detect discontinued apps in the future, as well.

Refs #8638.

Issue #8638, opened 2015-01-04
List of officially abandoned apps


vitorgalvao:
This list is likely incomplete.

These apps are all officially abandoned, and may or may not work on newer versions of OS X. They should be checked, and possibly moved to the appropriate repos.

- [ ] [candybar.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/candybar.rb)
- [ ] [google-notifier.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/google-notifier.rb)
- [ ] [handbrakebatch.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/handbrakebatch.rb)
- [ ] [flow.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/flow.rb)
- [x] ~~[meerkat.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/meerkat.rb)~~ (still works in Mavericks)
- [ ] [milkmaid.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/milkmaid.rb)
- [ ] [panic-unison.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/panic-unison.rb)
- [ ] [perian.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/perian.rb)
- [ ] [rightzoom.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/rightzoom.rb)
- [x] ~~[scrup.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/scrup.rb)~~ (still works in Mavericks)
- [ ] [sensiolabsdesktop.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/sensiolabsdesktop.rb)
- [x] [sparrow.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/sparrow.rb)
- [ ] [temperature-monitor.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/temperature-monitor.rb)
- [ ] [yasu.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/yasu.rb)

iggyvolz:
Would depends_on :macos => '< 10.10' or something of the like be appropriate for these casks?

vitorgalvao:
[...]

PR #16097, opened 2015-12-24
Add discontinued caveat to kobito

Amorymeltzer:
Last version available for direct download; I always feel weird using 'discontinued' for these.
https://twitter.com/kobitoapp/status/679183539028426752

### DIFF
--- a/Casks/kobito.rb
+++ b/Casks/kobito.rb
@@ -10,4 +10,8 @@ cask 'kobito' do
   license :gratis
 
   app 'Kobito.app'
+
+  caveats do
+    discontinued
+  end
 end


vitorgalvao:
> I always feel weird using 'discontinued' for these.

And they shouldn’t use. An app is considered discontinued if the developer stopped working on it. If they simply decided to stop providing direct download versions and are making them MAS-only, then they are considered MAS-only, and the cask should be removed.

adityadalal924:
@vitorgalvao In the past, these have been moved to ``homebrew-versions`
Ex. https://github.com/caskroom/homebrew-versions/blob/master/Casks/ccmenu18.rb

vitorgalvao:
> In the past, these have been moved to `homebrew-versions`

Crap. Do you know of any more? I think it’s a really bad idea to keep these around. Just look at the amount of people who still ask for MAS-apps and those that don’t understand why we cannot have those. Those people will likely see these apps here and use them, without realising they’re outdated (which can be potentially dangerous, depending on the nature of the software).

We should really get rid of them.

vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/16101.

Amorymeltzer:
I'm good with that @vitorgalvao, but if that's policy it needs to be stated somewhere.  [The only mention of 'discontinued' is](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#caveats-stanza-details):

 >software has been officially discontinued upstream

That needs to be narrowed-down.  Might need to look over #8638 and #9255 as well.

vitorgalvao:
> Might need to look over #8638 and #9255 as well.

I worked on those ones and I always had this rule in mind, so hey should be correct.

> The only mention of 'discontinued' is

But that’s the wrong place to look at it, I’d say. The rule seems clear to me, if the software has simply changed distribution methods, it has not been discontinued, and [MAS-only apps are not allowed](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#trial-and-freemium-versions).

As such, any suggestions for improving clarity are appreciated. I still don’t think having a note in `discontinued` is the right place, though. Perhaps a more prominent description regarding MAS apps?

Amorymeltzer:
>But that’s the wrong place to look at it, I’d say. [snip] Perhaps a more prominent description regarding MAS apps?

My point exactly.  I might have some ideas after vacation.  But :+1: for the distinction (and for tossing `homebrew-versions`, one rung away from `boneyard`).

Amorymeltzer:
I've been thinking over this, and FWIW I think I'll rescind my above agreement.  I simply don't see a large enough gulf between "apps with no development whatsoever but an old download is still provided" and "apps with MAS-only development but an old download is still provided" that makes the former acceptable with `discontinued` but the latter worth deleting from here and `versions`.  The latter is indeed more likely to be confused with the up-to-date version from the MAS, but that's what `caveats` are for (as stated elsewhere, I'm probably for more `caveats` than most current maintainers) (alternatively, why shouldn't the former be considered "legacy" and moved out?).

No need to reply; just giving my $0.02 following a change of :heart:.

vitorgalvao:
I think it’s clear which points we disagree on (you specifically touched on them) so no need for me to reiterate them.

I’ll add, though, that I’ve seen enough user confusion to make me believe this is the right solution. We should never knowingly promote a legacy version in the main repo.

Regarding `caveats`, I strongly believe they are incredibly inefficient. People don’t read messages. They don’t even read comments inside casks when updating them, let alone messages that just pass by when installing (exacerbated by when you install a bunch at once). You could make a similar argument for `discontinued` in general, and I agree. However, going back to “not a discoverability service”, it’s more likely you know and understand the app is no longer being maintained than understand you are not getting the latest version when making an install (just look at the amount of people who still ask for MAS-only apps *when they know apps are MAS-only*).

As for

> alternatively, why shouldn't the former be considered "legacy" and moved out?

Because with [the rules to keep that repo clean](https://github.com/caskroom/homebrew-versions#acceptable-casks), there isn’t a point.

PR #16099, opened 2015-12-24
Remove tags

vitorgalvao:
Closes https://github.com/caskroom/homebrew-cask/issues/15971.


adityadalal924:
Vendor tags need to be prepended to name, otherwise we lose that data!

vitorgalvao:
Ah, right, had a note to do that and forgot.

vitorgalvao:
Fixed `name`s.

vitorgalvao:
Merging now as I’d like to make other changes that might conflict.

PR #16102, opened 2015-12-24
Add :auto_updates stanza for Casks

vitorgalvao:
@jawshooah Addressed your comment, fixed the header for the test, and made some small adjustments to the documentation. Should be ready to go.


adityadalal924:
Sorry, must have missed the discussion for this. What was the use case?

Is is still:

auto_updates. Similarly to stage_only, it should be used when a value of true is needed. It means “the artifact this cask uses auto-updates, so even if the cask file updates, leave this alone when upgrading others”. This does not mean upgrade in general is no longer a desirable feature — it is, but for casks who do not do it themselves.

vitorgalvao:
@jawshooah Addressed your comment, fixed the header for the test, and made some small adjustments to the documentation. Should be ready to go.

vitorgalvao:
@adityadalal924 Yes.

adityadalal924:
So this would likely be `true` for any casks which are a) versioned b) have an appcast (sparkle, most likely) and c) update themselves (again, likely Sparkle). 

This shouldn't need to be set as `true` when the Cask itself is unversioned, (ie, spotify)?

vitorgalvao:
This should only be set if the app itself can auto-update, every other part of the cask is irrelevant. It doesn’t matter if the url is versioned or unversioned, or if the app has an appcast we can access (likely it will). The only thing that matters is “does this app auto-update by itself”.

Having an appcast isn’t a guarantee the app can auto-update (every github appcast).

adityadalal924:
Gotcha, so anything with an $app_name > Check for Updates.. menu item (Sparkle) as well as a few apps with their own updaters (MS Office, Adobe products, Steam, Dropbox come to mind, but I'm sure there are many more).

LGTM then - the stanza can be progressively filled in as we encounter Casks with auto-update.

vitorgalvao:
Precisely.

jawshooah:
Still not sure the new error type is necessary, but it may be more useful later on. I'm fine with merging this as is.

vitorgalvao:
I had actually removed it, but then had to redo the work and realised only after pushing that I forgot to re-remove it. Looking at it closer, though, I agree it may indeed come in handy, and also see no harm in keeping it for now.

PR #16122, opened 2015-12-25
fotomagico.rb: delete appcast

vitorgalvao:
Appcast does not work.


vitorgalvao:
Merged as b87fc793c7ed38a37284b49aab30851de17b5365.

PR #16123, opened 2015-12-25
ionu.rb: delete appcast

vitorgalvao:
Appcast does not work.


vitorgalvao:
Merged as f86860dad9b8b06b5ad6d084bbb60e176a490ee1.

PR #16124, opened 2015-12-25
panic-unison.rb: delete appcast

vitorgalvao:
Appcast does not work. Irrelevant either way, since the app was discontinued.


vitorgalvao:
Merged as 9d3a31e09b4c36b47e3516aaf6132f40494417e1.

PR #16127, opened 2015-12-25
Delete safemonk.rb

vitorgalvao:
App is dead.


adityadalal924:
What happened here?

"vitorgalvao merged commit c0aef7d into proxpn-appcast 10 hours ago"

vitorgalvao:
I did this at the same time I was adding `:sha256` to all appcasts, so something appears to have been confused regarding branches. All commits appear to be correct, though.

PR #16128, opened 2015-12-25
shimo.rb: removed appcast

vitorgalvao:
It does not work.


adityadalal924:
Also same problem?

"vitorgalvao merged commit 3dad1bf into delete-safemonk "

PR #16152, opened 2015-12-27
Add Unity Editor

cybertk:


Casks/unity-editor.rb
```diff
@@ -0,0 +1,17 @@
+# Manual installing of unity
```
vitorgalvao: Remove these comments.

Casks/unity-editor.rb
```diff
@@ -0,0 +1,17 @@
...
+# Followed http://docs.unity3d.com/Manual/InstallingUnity.html
+cask 'unity-editor' do
+  version '5.3.1f1'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

Casks/unity-editor.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  # Download url can be found at ~/Downloads/UnityPackages/install.ini after opening 'Unity Download Assistant.app'
```
vitorgalvao: So are you saying this cask completely replaces the need for `Unity Download Assistant.app`?

vitorgalvao: Either way, comment is useless and must go.

cybertk:
This Formula is a replacement of existing one `unity`. And this Formula will install Unity without `Unity Download Assistant.app`.

I just want to add some comments for future update guides, where should they go?

vitorgalvao:
> This Formula is a replacement of existing one `unity`. And this Formula will install Unity without `Unity Download Assistant.app`.

Those are good news. Then please delete the current `unity.rb` cask as well, and rename this one to `unity.rb` (with header `cask 'unity' do`). As per the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) that is the correct name, anyway.

Are you sure it’s a replacement, though? Because it looks like it can do fewer things. Either way, the current one should be removed or at least renamed to `unity-download-assistant`.

> I just want to add some comments for future update guides, where should they go?

They shouldn’t go anywhere. This app isn’t doing anything others aren’t doing, there’s no need for comments.

cybertk:
After Unity v5.3, unity is a collection of components. Each component can be installed standalone. This formula installs the core component called **Unity Editor**

vitorgalvao:
Got it.

Comments still need to be removed and `sha256` updated, but everything else is fine, then.

vitorgalvao:
> After Unity v5.3, unity is a collection of components. Each component can be installed standalone. This formula installs the core component called **Unity Editor**

Read the token reference I’ve now linked to multiple times. What the developer calls it in a particular page does not matter. What matters is the name of the bundle, so we have someway to make sense of everything and not have duplicates. `name` is where the correct name matters, not the header.

This should still be renamed to `unity`. Comments also still need to be removed and `sha256` updated, but everything else is fine.

vitorgalvao:
Thank you for the contribution. It was merged directly as d7248c8e5640db8322d1f04fa7fb55612a8483e4 to keep commit history cleaner. Your contribution is still credited to you.

PR #16153, opened 2015-12-27
Add Unity iOS Build Support

cybertk:


Casks/unity-ios-build-support.rb
```diff
@@ -0,0 +1,18 @@
+# Manual installing of unity
```
vitorgalvao: Remove comments.

Casks/unity-ios-build-support.rb
```diff
@@ -0,0 +1,18 @@
...
+# Followed http://docs.unity3d.com/Manual/InstallingUnity.html
+cask 'unity-ios-build-support' do
```
vitorgalvao: This has to be changed to `cask 'unity-ios-support-for-editor' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `unity-ios-support-for-editor.rb`.

Casks/unity-ios-build-support.rb
```diff
@@ -0,0 +1,18 @@
...
+  version '5.3.1f1'
+  sha256 :no_check
+
+  depends_on :cask => 'unity-editor'
+
+  # Download url can be found at ~/Downloads/UnityPackages/install.ini after opening 'Unity Download Assistant.app'
```
vitorgalvao: Comment is useless. Please remove it.

Casks/unity-ios-build-support.rb
```diff
@@ -0,0 +1,18 @@
+# Manual installing of unity
+# Followed http://docs.unity3d.com/Manual/InstallingUnity.html
+cask 'unity-ios-build-support' do
+  version '5.3.1f1'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

Casks/unity-ios-build-support.rb
cybertk: The official name from Unity is called **Unity iOS Build Support**. So do you mean it should be `cask 'unity-ios-build-support-for-editor'. And the deps of `unity-editor` is not required as user can Install unity-editor without using brew

vitorgalvao: > The official name from Unity is called **Unity iOS Build Support**

The “official name” of most apps is open to interpretation. Which is why we need those rules. Without them, we’ll be filled with duplicates.

> So do you mean it should be `cask 'unity-ios-build-support-for-editor'`.

Yes.

> And the deps of `unity-editor` is not required as user can Install `unity-editor` without using brew.

That is true of any dependency. It should stay.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 863e08a7edd65d105b29aa1e083103b3d80e6de0. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #16158, opened 2015-12-27
Add auto_updates to a few casks

vitorgalvao:
Edit (@jawshooah): on hold until #16161 is resolved

### DIFF
--- a/Casks/alfred.rb
+++ b/Casks/alfred.rb
@@ -7,6 +7,8 @@ cask 'alfred' do
   homepage 'https://www.alfredapp.com/'
   license :freemium
 
+  auto_updates true
+
   app 'Alfred 2.app'
 
   accessibility_access true
--- a/Casks/alfred.rb
+++ b/Casks/alfred.rb
@@ -8,10 +8,9 @@ cask 'alfred' do
   license :freemium
 
   auto_updates true
+  accessibility_access true
 
   app 'Alfred 2.app'
-
-  accessibility_access true
 
   postflight do
     suppress_move_to_applications :key => 'suppressMoveToApplications'
--- a/Casks/atom.rb
+++ b/Casks/atom.rb
@@ -11,7 +11,6 @@ cask 'atom' do
   license :mit
 
   auto_updates true
-
   depends_on :macos => '>= :mountain_lion'
 
   app 'Atom.app'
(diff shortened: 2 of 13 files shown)


jawshooah:
Are we sure that the `auto_updates` stanza should go at the bottom of the cask? I don't think we've discussed where it fits in the stanza order.

That said, there's not much point in worrying about it until we can check (and auto-correct) the stanza order with [rubocop-cask](https://github.com/jawshooah/rubocop-cask).

vitorgalvao:
That’s a good point. The second “group” (`url`, `name`, …) seems like a good option. What about directly under `license`?

Pinging @adityadalal924 and @Amorymeltzer since they’re likely to also have some opinions for this.

adityadalal924:
Under `:license` should be good. It could also go above name.

 I was thinking something like the below could work:
```
version
shasum

url
appcast, shasum
name
homepage
license
auto_updates

pre/post flights

artifact stanzas

uninstall

zap
```

jawshooah:
@adityadalal924 Don't forget `depends_on`, `conflicts_with`, and `stage_only` (though the latter could be treated as an artifact stanza, I suppose).

adityadalal924:
@jawshooah Updated.

The name/homepage/license block could also come before url/appcast, shasum/auto_updates - it might make more sense semantically that way.

jawshooah:
Some others that are missing: `gpg`, `container`, `caveats`, `accessibility_access`

jawshooah:
Let's move the discussion of stanza order to #16161.

vitorgalvao:
> The name/homepage/license block could also come before url/appcast, shasum/auto_updates - it might make more sense semantically that way.

That order has (to a point) some established logic. It isn’t about semantics, but about what contributors should be concerned with. `version` and `shasum` are the *only* stanzas guaranteed to change with a version update, hence why they are separate and at the top. Next are `url` and `appcast` which can change but many times won’t (interpolation)<sup>1</sup>. Next are all the others that (in theory) won’t change over the lifetime of the app: `name`, `homepage`, `license`.

Then come the next blocks, that change widely across casks. Keeping at least those two blocks consistent always made sense to me, and I believe their order is good.

---

<sup>1</sup> Yes, `:sha256` in `appcast` will change, but those need to be coupled together, and `appcast` itself won’t change much.

vitorgalvao:
Updated for new stanza order (also fixed other stanzas).

jawshooah:
Should be good to go after a rebase.

PR #16178, opened 2015-12-28
Update homepage of breach

tsparber:
Original homepage has a dummy site, Github project points to github.io.
The appcast does not work.


Casks/breach.rb
```diff
@@ -4,10 +4,8 @@
 
   # githubusercontent.com is the official download host per the vendor homepage
```
vitorgalvao: Comment should be removed.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 35119ff751ce51a14f8fb3ba31d3631194b89056. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #16183, opened 2015-12-28
Update and rename unity.rb to unity-download-assistant.rb

vitorgalvao:


fanquake:
Was there meant to be an update included here, or just the renaming?

vitorgalvao:
Just the rename. The “Update” part is due to having changed the header (github’s web interface default message).

PR #16187, opened 2015-12-28
Upgrade Inform.app v7.6M62

axodys:
Also set license to oss based on information provided at http://inform7.com/sources/
GUI is GPL v2 and Inform natural language parser component is Artistic License.


adityadalal924:
You had unrelated commits, an extraneous comment, and version string interpolation needed to be updated - fixed in https://github.com/caskroom/homebrew-cask/pull/16188

Thank you for the contribution!

PR #16203, opened 2015-12-28
removed mysql-utilities

adidalal:
Does not work with OSX El Capitan
As reported in https://github.com/caskroom/homebrew-cask/issues/16202


adityadalal924:
Merged as 3154ed87d5c6e04856d5e1be79188cdb24a18062.

vitorgalvao:
Instead of removing it, we should instead add a `depends_on`, since it is likely this will soon start working again. No reason for people not on El Capitan to not be able to use it.

Issue #12827, opened 2015-07-25
Read the version stanza into a version object instead of a string


vitorgalvao:
As per https://github.com/caskroom/homebrew-cask/issues/6347:

> The purpose of the version object would be to provide more readable methods for manipulating the version number.  This would allow us to write eg `"#{version.underscorify}"` and `"#{version.major}"`.  String manipulation with `sub`/`gsub` works well enough, but is hard to read, and surely presents a barrier for first-time Cask authors and people without Ruby knowledge.  We could still use `sub`/`gsub` for corner cases.

---

Table in progress (feel free to update<sup>1</sup> at will as we agree on things). Initial version taken from the next comment, from @adityadalal924.

| Method Name         | Sample Input | Sample Output | Original Function                                        |
|---------------------|--------------|---------------|----------------------------------------------------------|
| dots_to_underscores | 4.3.2        | 4_3_2         | `#{version.gsub('.','_')}`                               |
| dots_to_hyphens     | 4.3.2        | 4-3-2         | `#{version.gsub('.', '-')}`                              |
| no_dots             | 4.3.2        | 432           | `#{version.gsub('.', '')}` and  `#{version.delete('.')}` |
| major               | 4.3.2        | 4             | `#{version.to_i}`                                        |
| major_minor         | 4.3.2        | 4.3           | `#{version.to_f}`                                        |
|                     | 4.3.2-1337   | 4.3.2         | `#{version.sub(%r{-.*$}, '')}`                           |
|                     | 4.3.2-1337   | 1337          | `#{version.sub(%r{.*?-}, '')}`                           |

----------------------

<sup>1</sup> (Editing is easiest with [Markdown Tables generator](http://www.tablesgenerator.com/markdown_tables) - use File > Paste table data...)

adityadalal924:
Here are some common ones I've seen. Update at will - coming up with intuitive names is hard.

| Method Name  	| Sample Input 	| Sample Output 	| Original Function                                       	|
|--------------	|--------------	|---------------	|---------------------------------------------------------	|
| underscorify 	| 4.3.2        	| 4_3_2         	| `#{version.gsub('.','_')}`                              	|
|              	| 4.3.2        	| 4-3-2         	| `#{version.gsub('.', '-')}`                             	|
|              	| 4.3.2        	| 432           	| `#{version.gsub('.','')}` and  `#{version.delete('.')}` 	|
[...]

PR #16205, opened 2015-12-28
Read version stanza into Hbc::DSL::Version object

jawshooah:
I've implemented a few of the methods outlined in #12827. I'll implement more as we decide on them.

Right now we delegate a lot of methods to `to_s` to avoid having to update a ton of casks in one go. We can remove the ones we don't want over time.


jawshooah:
Alternatives to `def_delegators` include:
  - explicitly define all methods (redundant IMHO)
  - subclass `String`
  - delegate to `to_s` in `method_missing`

vitorgalvao:
[First pass at documentation](https://github.com/caskroom/homebrew-cask/pull/16255).

PR #16214, opened 2015-12-29
Upgrade to latest mysqlworkbench version

derekperkins:


Casks/mysqlworkbench.rb
```diff
@@ -1,5 +1,5 @@
 cask 'mysqlworkbench' do
-  version '6.3.5'
+  version '6.3.6'
   sha256 '7b92e7dc0909f5225bb1d940addd2384d2f1b3d60c3b46a03014bdfc1e95f1ae'
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

derekperkins:
Thanks for the pointer. That's done now.

vitorgalvao:

derekperkins:
Done

vitorgalvao:
Thank you for the contribution. It was merged directly as c49b08b9fbfd2889a8e518f44893b6e86d61db49 to keep commit history cleaner. Your contribution is still credited to you.

derekperkins:
If you don't mind me asking, what was your workflow to have it merged like that, saying "with @vitorgalvao"? I'm a contributor to github.com/nnnick/chart.js and we often have squashing issues where we lose contribution, typically when the PR needs to be rebased on master.

vitorgalvao:
I use some custom-built scripts. In this case, I used [`fastmerge`](https://github.com/vitorgalvao/tiny-scripts/blob/master/fastmerge), but if I wanted to make changes to your PR before merging (and I would still keep the contribution credited to you) I would’ve used [`prfixmaster`](https://github.com/vitorgalvao/tiny-scripts/blob/master/prfixmaster)<sup>1</sup>.

You can install any of them by first tapping their homebrew repo: `brew tap vitorgalvao/homebrew-tinyscripts` and then installing in the usual way. All of them will give you instructions if you run them without arguments or with `-h`/`--help`. Also, if there’s anything you need to configure first, they’ll also tell you.

---

<sup>1</sup> There is also [`prfixbranch`](https://github.com/vitorgalvao/tiny-scripts/blob/master/prfixbranch) that creates a branch, instead of merging to master directly.

vitorgalvao:
To note that I don’t do anything special to get that particular message, github just interprets the way the result of those scripts in such a way, so if you don’t want to use them, you could try to see what is different there from the way you do it at the other repo.

PR #16215, opened 2015-12-29
Revert "Delete Subversion support"

vitorgalvao:
Reverts caskroom/homebrew-cask#16118.

Looks like [there was a case for this](https://github.com/caskroom/homebrew-fonts/issues/428) after all.

This necessity makes me dislike subversion support even more, though. The reason we’re doing it this way is `svn`, unlike `git`, allows us to checkout a single directory from the repo, making for a smaller download.

Download size in itself isn’t that big a deal, I’d say, since worst case scenario the download of a full fonts repo should be ≈250MBs (Google Fonts) and there are bigger apps. With caching, and since users likely don’t install fonts all the time, that wouldn’t be a huge deal, except that since we need to separate apps by directory before linking (moving, in the future) we’d leave users with a bunch of copies of those uncompressed repos, and that is not good.

However, using subversion in this manner is in itself bad for casks, I’d argue, since it requires [an important additional value](https://github.com/caskroom/homebrew-fonts/blob/ab1e0df2005278a6b132b1b8a5d4dc55cd0ff233/Casks/font-sauce-code-powerline.rb#L8) to `url`s. This [subversion revision](https://github.com/caskroom/homebrew-cask/pull/16118/files#diff-eb6c53cc0d6eb7ed0964481ed17f8624L300) value is (seems like) an obtuse parameter to get in the context of a github repo, and I’m not even sure there’s an easy way for contributors to know when to updated it and to what value. It also makes the cask somewhat more difficult to understand in the sense that [its `url` value](https://github.com/caskroom/homebrew-fonts/blob/ab1e0df2005278a6b132b1b8a5d4dc55cd0ff233/Casks/font-sauce-code-powerline.rb#L6) will seem to lead nowhere.

In light of all this, I’d rather have yet a different solution. I briefly considered a sibling project, like [our sister](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#unofficial-vendorless-and-walled-builds) [alehouse](https://github.com/alehouse) to host these multiple-fonts-per-repo fonts. I was envisioning a script that regularly would download those repos, package fonts as individual zip files, and push them to a github repo. Corresponding cask font files would then be set to `:latest` (fonts is one case where you absolutely always want the latest) and download from there. Those wouldn’t really be official links, though, and there may be other things I’m also not predicting there.

Any ideas?

### DIFF
--- a/doc/CASK_LANGUAGE_REFERENCE.md
+++ b/doc/CASK_LANGUAGE_REFERENCE.md
@@ -291,6 +291,16 @@ Web browsers may obscure the direct `url` download location for a variety of rea
 $ ./developer/bin/list_url_attributes_on_file <file>
 ```
 
+### Subversion URLs
+
+In rare cases, a distribution may not be available over ordinary HTTP/S. Subversion URLs are also supported, and can be specified by appending the following key/value pairs to `url`:
+
+| key                | value       |
+| ------------------ | ----------- |
+| `:using`           | the symbol `:svn` is the only legal value
+| `:revision`        | a string identifying the subversion revision to download
+| `:trust_cert`      | set to `true` to automatically trust the certificate presented by the server (avoiding an interactive prompt)
+
 ## Appcast Stanza Details
 
 The value of the `appcast` stanza is a string, holding the URL for an appcast which provides information on future updates.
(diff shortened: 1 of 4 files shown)


adityadalal924:
@vitorgalvao We should revert this merge ASAP and open up a discussion about a solution - the fonts repository is currently unusable.

Issue #16204, opened 2015-12-28
Enable Travis caching


adityadalal924:
As seen on https://docs.travis-ci.com/user/caching#Caching-directories-(Bundler%2C-dependencies)

Looks like the new container-based infrastructure has to be used, which means no sudo.

Not sure if it's feasible, but if so, would speed up Travis checks considerably.

jawshooah:
It isn't feasible unfortunately, as we need sudo in order to run our tests using system Ruby.

PR #16226, opened 2015-12-30
Travis: test multiple versions of OSX

jawshooah:
Significantly clean up the Travis config, and test on OSX 10.11, 10.10, and 10.9.

Caching is not available for OSX environments at the moment, but in the hopes that this will change, I have preemptively set sudo to false and indicated directories to cache.

Refs #16204.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,70 +1,90 @@
 language: objective-c
 
-# todo
-#
-# test with Ruby 2.1, but allow failure
-#
+sudo: false
 
-# "Current" is dependent on OS X release.  At the time of writing, it
-# is always either 1.8 or 2.0
 env:
-  matrix:
-    - CASK_RUBY_TEST_VERSION="2.0"
-    - CASK_RUBY_TEST_VERSION="Current"
+  global:
+    - HOMEBREW_RUBY=2.0.0
+    - LANG=en_US.UTF-8
+    - LANGUAGE=en_US.UTF-8
+    - LC_ALL=en_US.UTF-8
+    - CASK_TAP_DIR=/usr/local/Library/Taps/caskroom/homebrew-cask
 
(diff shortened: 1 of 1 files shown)


jawshooah:
@adityadalal924 - Turns out we don't need sudo after all, but we still can't cache :disappointed:

adityadalal924:
Travis issue for tracking:

https://github.com/travis-ci/travis-ci/issues/4011

adityadalal924:
Why is this done?
"uninstall pre-installed brew-cask and untap caskroom/cask"

jawshooah:
@adityadalal924 Because otherwise, when we `brew cask install Casks/adobe-air.rb`, Homebrew finds `/usr/local/bin/brew-cask` before `cmd/brew-cask.rb` due to [the order in which it checks the `PATH` for subcommands](https://github.com/Homebrew/homebrew/blob/4cf94797f7c94184ffaebb585382e1d8aac3e41c/Library/brew.rb#L144-L149).

jawshooah:
There are two alternatives:

```bash
ruby cmd/brew-cask.rb install Casks/adobe-air.rb
```

and:

```bash
export CASK_TAP_DIR="$(brew --repository)/Library/Taps/caskroom/cask"
brew uninstall --force brew-cask
rsync -az --delete "${TRAVIS_BUILD_DIR}/" "${CASK_TAP_DIR}/"
export TRAVIS_BUILD_DIR="${CASK_TAP_DIR}"
cd "${CASK_TAP_DIR}"
brew cask install adobe-air
```

I'm partial to the latter, actually. It more closely replicates an actual user's environment.

adityadalal924:
I'll leave the actual choice up to you - I was just curious about the rationale. Thanks for explaining!

PR #16237, opened 2015-12-30
Attempt somewhat more clear language upon skipping checksum verification

Amorymeltzer:
The language here was bugging me and [at least one other person](https://github.com/caskroom/homebrew-cask/pull/16084#issuecomment-167047851), so this is an attempt at being somewhat more clear while retaining a concise message.

Better suggestions welcome. Pinging @adityadalal924

### DIFF
--- a/lib/hbc/verify/checksum.rb
+++ b/lib/hbc/verify/checksum.rb
@@ -3,7 +3,7 @@ require 'digest'
 class Hbc::Verify::Checksum
   def self.me?(cask)
     return true unless cask.sha256 == :no_check
-    ohai "Skipping checksum for Cask #{cask}"
+    ohai "No checksum defined for Cask #{cask}, skipping verification"
     false
   end
 


lib/hbc/verify/checksum.rb
```diff
@@ -3,7 +3,7 @@
 class Hbc::Verify::Checksum
   def self.me?(cask)
     return true unless cask.sha256 == :no_check
-    ohai "Skipping checksum for Cask #{cask}"
+    ohai "No checksum provided for Cask #{cask}, skipping verification"
```
vitorgalvao: “Provided” suggests simply that no one gave it, instead of the actual case that we can’t have one. Perhaps “available”.

Amorymeltzer: Yeah, but don't you think "available" feels like we couldn't figure it out?  Both felt awkward to me.  I had meant to use "viable" but as a biologist I was worried I was being too jargony.

Other thoughts I had, of varying quality:
- Acceptable
- Procurable
- Applicable
- Feasible
- Convenient
- Attainable
- Obtainable
- Employable

jawshooah: How about "defined"? That or "available" seem fine to me.

vitorgalvao: I’m fine with “defined”.

adityadalal924: Defined sounds good.

Amorymeltzer:
Defined FTW :clap:

PR #16239, opened 2015-12-31
Fix Spotify OS requirement and stanza order

iniquities:


adityadalal924:
Thank you for the contribution. It was merged directly as 4713c5b866e870910ed502a3da86e87f382537db to keep commit history cleaner. Your contribution is still credited to you.

Casks/spotify.rb
```diff
@@ -6,7 +6,8 @@
   name 'Spotify'
   homepage 'https://www.spotify.com'
   license :gratis
-  depends_on :macos => '>= :snow_leopard'
+
```
vitorgalvao: There’s trailing whitespace here, which will make the build fail.

vitorgalvao:
Thank you for this. The stanza order rule is pretty new, so many casks will have errors like this. Any fix is appreciated.

Casks/spotify.rb
vitorgalvao: No seat, fixed in another commit.

PR #16240, opened 2015-12-31
fix spotify whitespace

adidalal:
Fixes https://github.com/caskroom/homebrew-cask/pull/16239


adityadalal924:
Whoops, fixed by @vitorgalvao already.

PR #16245, opened 2015-12-31
Extract travis build logic into scripts

jawshooah:
I like this setup much better. Build scripts should be just that, scripts. We were doing way too much inside `.travis.yml`, and I think splitting out each build phase into its own script gives us a much better separation of concerns, and a more maintainable build flow overall.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,23 +4,21 @@ sudo: false
 
 env:
   global:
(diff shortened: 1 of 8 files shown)


vitorgalvao:
> I like this setup much better

So do I. Big fan of this change.

PR #16247, opened 2015-12-31
sourceforge.jp is now osdn / more generic checking for bad url formats

vitorgalvao:


jawshooah:
This will require changes to `lib/hbc/audit.rb` and `spec/cask/audit_spec.rb` as well.

vitorgalvao:
Right, I always forget those when making changed to the sourceforge documentation.

Not exactly pretty, but I’ve kept them bundled together there as well, since they are the only formats we check for, and for all intents and purposes, `sourceforge.jp` URLs are still equivalent to `osdn.jp` and function.

spec/cask/audit_spec.rb
```diff
@@ -74,20 +74,20 @@
     end
 
     describe "preferred download URL formats" do
-      let(:warning_msg) { /SourceForge URL format incorrect/ }
+      let(:warning_msg) { /SourceForge\/OSDN URL format incorrect/ }
 
-      context "incorrect SourceForge URL format" do
-        let(:cask_token) { 'sourceforge-incorrect-url-format' }
+      context "incorrect SourceForge/OSDN URL format" do
+        let(:cask_token) { 'sourceforge-osdn-incorrect-url-format' }
```
jawshooah: These are actual test casks present in `spec/support/Casks`. You'll have to update those too :wink:

jawshooah: Better to just add separate fixtures and tests for the OSDN URLs, correct and otherwise.

vitorgalvao: Thank you for the audit. Should be good now.

vitorgalvao:
Went ahead and made further changes. Checking for bad url formats is now done in a more generic way, so we can easily add more in the future if needed.

spec/support/Casks/osdn-correct-url-format.rb
```diff
@@ -0,0 +1,5 @@
+test_cask 'sourceforge-correct-url-format' do
```
jawshooah: Almost there! Forgot to update the token :smile:

spec/support/Casks/osdn-incorrect-url-format.rb
```diff
@@ -0,0 +1,5 @@
+test_cask 'sourceforge-incorrect-url-format' do
```
jawshooah: This one as well.

spec/support/Casks/osdn-correct-url-format.rb
vitorgalvao: Fixed.

vitorgalvao:
Crap, there are conflicts, now.

vitorgalvao:
Have to leave now, so this will have to be fixed later. Some of the logic was changed in https://github.com/caskroom/homebrew-cask/commit/49bddda783b32e79bec6f05fb3be98ae7a2951a9, so will have to remake changes around that.

jawshooah:
@vitorgalvao No worries, I'm taking care of it. `git rerere` isn't helping here for some reason, and I don't want to resolve conflicts per-commit, so we'll just grind our teeth and live with a merge bubble.

vitorgalvao:
Thank you for handling it, I appreciate it.

PR #16249, opened 2015-12-31
added shortcuts

vitorgalvao:
Closes https://github.com/caskroom/homebrew-cask/issues/16151.

Alternatively, we could call this `lock-logout-restart-shutdown-sleep`. The number of utilities should only increase, though, so that may not be the best idea. Since this is a collection of apps, it might warrant this exception.

### DIFF
--- a/Casks/shortcuts.rb
+++ b/Casks/shortcuts.rb
@@ -1,0 +1,13 @@
+cask 'shortcuts' do
+  version '1.0.0'
+  sha256 '1881958abf9078656de1dac1ccc50777f363f5602183260abe2a8b30dc76c4c6'
+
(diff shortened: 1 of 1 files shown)


reitermarkus:
Wouldn't `suite 'system', :target => 'Shortcuts'` be more appropriate in this case?

vitorgalvao:
That sounds about right. I always considered it (even when originally discussing the rule) as the case to use only when necessary (i.e. it wouldn’t work without it), but it makes sense to do it for a group of related applications as well (and looking at it, we even considered that case in the documentation).

vitorgalvao:
Merged as 397b3ce5bba476c535e79bead4ea47c045fdaaf5.

PR #16255, opened 2015-12-31
Documentation on new version methods

vitorgalvao:


CONTRIBUTING.md
```diff
@@ -92,6 +92,51 @@ cask 'gateblu' do
 end
 ```
 
+#### `version` methods
+
+In the examples above, when possible the `url` stanza uses `#{version}` ([string interpolation](https://en.wikipedia.org/wiki/String_interpolation#Ruby)) to create Cask that only needs `version` and `sha256` changes when updated. This can be taken further, when needed, with [ruby String methods](http://ruby-doc.org/core/String.html). For example:
```
Amorymeltzer: Should read: "create **a** Cask"

vitorgalvao: Fixed.

Amorymeltzer:
Once the changes in #16205 et al are made, I imagine some enterprising young contributor will go through and mass-replace the obvious ones; after that, a link to an example using one (or two!) of the methods would be useful.

CONTRIBUTING.md
```diff
@@ -92,6 +92,51 @@ cask 'gateblu' do
...
+| Method              | Input                  | Output             |
+|---------------------|------------------------|--------------------|
+| `major`             | `1.2.3-a45_b67/ccdd88` | `1`                |
+| `minor`             | `1.2.3-a45_b67/ccdd88` | `2`                |
+| `patch`             | `1.2.3-a45_b67/ccdd88` | `3`                |
+| `major_minor`       | `1.2.3-a45_b67/ccdd88` | `1.2`              |
+| `major_minor_patch` | `1.2.3-a45_b67/ccdd88` | `1.2.3`            |
+| `before_slash`      | `1.2.3-a45_b67/ccdd88` | `1.2.3-a45_b67`    |
+| `after_slash`       | `1.2.3-a45_b67/ccdd88` | `ccdd88`           |
+| `no_dividers`       | `1.2.3-a45_b67/ccdd88` | `123a45b67/ccdd88` |
```
jawshooah: Should be `123a45b67ccdd88` now.

We should probably also mention the conversion methods (`dots_to_underscores`, etc).

vitorgalvao: Oh, right. Fixed.

CONTRIBUTING.md
```diff
@@ -92,6 +92,60 @@ cask 'gateblu' do
...
+| `dots_to_hyphens`        | `1.2.3-a45_b67/ccdd88` | `1-2-3-a45_b67/ccdd88` |
+| `dots_to_underscores`    | `1.2.3-a45_b67/ccdd88` | `1_2_3-a45_b67/ccdd88` |
+| `hyphens_to_dots`        | `1.2.3-a45_b67/ccdd88` | `1.2.3.a45_b67/ccdd88` |
+| `hyphens_to_underscores` | `1.2.3-a45_b67/ccdd88` | `1.2.3_a45_b67/ccdd88` |
+| `underscores_to_dots`    | `1.2.3-a45_b67/ccdd88` | `1.2.3-a45.b67/ccdd88` |
+| `underscores_to_hyphens` | `1.2.3-a45_b67/ccdd88` | `1.2.3-a45-b67/ccdd88` |
+| `no_dots`                | `1.2.3-a45_b67/ccdd88` | `123-a45_b67/ccdd88`   |
+| `no_hyphens`             | `1.2.3-a45_b67/ccdd88` | `1.2.3a45_b67/ccdd88`  |
+| `no_underscores`         | `1.2.3-a45_b67/ccdd88` | `1.2.3-a45b67/ccdd88`  |
+| `no_dividers`            | `1.2.3-a45_b67/ccdd88` | `123a45b67ccdd88`      |
```
jawshooah: We also have `no_slashes`, `{dots,hyphens,underscores}_to_slashes`, and `slashes_to_{dots,hyphens,underscores}`.

jawshooah: The deletion and conversion methods are generated by the `define_divider_deletion_method` and `define_divider_conversion_methods` macros, which is why they aren't explicitly in the source.

vitorgalvao: Making a note. Will have to change later.

vitorgalvao: Done.

CONTRIBUTING.md
```diff
@@ -92,6 +92,67 @@ cask 'gateblu' do
...
+```
+
+We can
+
+```ruby
+version '1.2.3build4'
+url "http://example.com/#{version.sub(%r{build\d+}, '')}/file-version-#{version}.dmg"
+```
+
+That can become hard to read, however. But since many of these changes are common, we provide a few helpers to clearly interpret otherwise obtuse cases:
```
Amorymeltzer: I think at this point we might as well be honest and say "a number of helpers" rather than "a few".  I'd also toss the "But", the two sentences aren't in opposition ("Since" is fine, but maybe "As many of these..."?).

vitorgalvao: Agreed. Changed.

PR #16257, opened 2015-12-31
Add Mouse Locator v1.1

NaanProphet:
Created new cask for Mouse Locator tool. Ordinarily requires manual
installation of vendor's app file, but it installs just a preference
pane. Automated install and uninstall instead. Tested on OS X 10.11.1
El Capitan.

### DIFF
--- a/Casks/mouse-locator.rb
+++ b/Casks/mouse-locator.rb
@@ -1,0 +1,18 @@
+cask 'mouse-locator' do
+  version '1.1'
+  sha256 '1809760210e5afb80f9be34dc930c0c6fb84efee91747640d2d9717561149645'
+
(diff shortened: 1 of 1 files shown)


SirDap:
Cancelling pull request, more changes needed

adityadalal924:
FYI you don't need a zap stanza if you use the artifact `prefpane`, an uninstall will remove it for you.
Also, remember you can use string interpolation with double quotes so you don't need to hardcode in the `1.1`

SirDap:
Thanks, will do!

jawshooah:
@SirDap No need to close the pull request. You can make new commits on the same branch and push them up, and we'll see them. Once it's ready, we'll just need you to [squash your commits](https://davidwalsh.name/squash-commits-git) before we merge.

PR #16258, opened 2015-12-31
Updates Keycastr to v0.9.2

akitchen:


adityadalal924:
appcast shasum also needs to be updated, and version string should not have a v.

Fixed in https://github.com/caskroom/homebrew-cask/pull/16259

PR #16260, opened 2015-12-31
Add Mouse Locator 1.1 and PlistEdit Pro 1.8.2

NaanProphet:
Created new cask for Mouse Locator tool. Ordinarily requires manual installation of vendor's app file, but that just installs a preference pane. Tested on OS X 10.11.1 El Capitan.

Note: because of symlinking, preference pane's activate button cannot be toggled! Need to wait for issue #13201 to be resolved to move files instead of symlinking them. (Maybe it's a permissions thing?) Still committing though, in case it works as is on other versions on OS X.

Created new cask for PlistEdit Pro. Freemium software with zap command 
to remove preferences. Tested on OS X 10.11.1 El Capitan.

### DIFF
--- a/Casks/mouse-locator.rb
+++ b/Casks/mouse-locator.rb
@@ -1,0 +1,14 @@
+cask 'mouse-locator' do
+  version '1.1'
+  sha256 '1809760210e5afb80f9be34dc930c0c6fb84efee91747640d2d9717561149645'
+
+  url 'http://www.2point5fish.com/files/MouseLocator.dmg'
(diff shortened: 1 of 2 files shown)


SirDap:
Not sure if best practice to combine pull requests like this or create separate ones. Please advise.

jawshooah:
We prefer one PR per-cask if you don't mind.

SirDap:
@jawshooah Got it! No problem. I'll close this request and open two new ones. Sorry for any inconvenience.

PR #16261, opened 2015-12-31
Add Mouse Locator 1.1

NaanProphet:
Created new cask for Mouse Locator tool. Ordinarily requires manual installation of vendor's app file, but that just installs a preference pane. Tested on OS X 10.11.1 El Capitan.

Note: because of symlinking, preference pane's activate button cannot be toggled! Need to wait for issue #13201 to be resolved to move files instead of symlinking them. Still committing though, in case it works as is on other versions on OS X.

### DIFF
--- a/Casks/mouse-locator.rb
+++ b/Casks/mouse-locator.rb
@@ -1,0 +1,14 @@
+cask 'mouse-locator' do
+  version '1.1'
+  sha256 '1809760210e5afb80f9be34dc930c0c6fb84efee91747640d2d9717561149645'
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 9f5ffc0bda71722b68dd0ec62ede3b0767cbaa10. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #16263, opened 2015-12-31
Travis: add helpers for brew install and upgrade

jawshooah:
`brew install` a formula if it is not installed. If it is, `brew upgrade` if it's outdated.

### DIFF
--- a/ci/travis/helpers.sh
+++ b/ci/travis/helpers.sh
@@ -29,6 +29,22 @@ header () {
   echo
 }
 
+brew_install () {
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
Line 44, is there any issue if/when `brew upgrade` returns an error?

jawshooah:
True, it's not necessary to the build so we could add `|| true` just to be safe. Figured it couldn't hurt to be up-to-date, since I imagine most Homebrew users brew update/upgrade fairly frequently.

PR #16264, opened 2016-01-01
Travis: use first parent of last commit as diff base if no PR

jawshooah:
When we are not building a PR, `TRAVIS_BRANCH` and `TRAVIS_COMMIT` point to the same git revision. Use the first parent of the last commit as a base instead.


jawshooah:
Pinging @Amorymeltzer since he brought this up in a comment on jawshooah/homebrew-cask@93873eeb767bba73adfe760f183c9e90d31b8be0.

Amorymeltzer:
:+1: 71e3cdd201d59f5c4b82b42d7b7cba571bdf4f07

PR #16274, opened 2016-01-01
Upgrade MediaElch.app to v2.4

boneskull:
that's about the size of it.


adityadalal924:
Thank you for the contribution. It was merged directly as c363caac04d8458b8edf385d412d8b972c4354a9 to keep commit history cleaner. Your contribution is still credited to you.

PR #16276, opened 2016-01-01
Upgrade PowerMate.app to v1.2.4

boneskull:
This seems to be the latest version, [as per the support page](https://support.griffintechnology.com/product/powermate/).

I hardcoded the version number in the URL because it's unlikely to have a matching base path for any subsequent version.

### DIFF
--- a/Casks/powermate.rb
+++ b/Casks/powermate.rb
@@ -1,8 +1,8 @@
 cask 'powermate' do
-  version '1.2.4_0'
+  version '1.2.4'
   sha256 '6fd755961bb5a486cb3ab9d47217f7877d4bb021134f32a3c32ceede3fd3cf3d'
 
-  url "http://support.griffintechnology.com/sites/default/files/PowerMate_v#{version}.zip"
+  url 'https://support.griffintechnology.com/wp-content/uploads/sites/3/2015/08/PowerMate_v1.2.4.zip'
   name 'Griffin Powermate'
   homepage 'https://support.griffintechnology.com/support/powermate/'
   license :gratis


adityadalal924:
@vitorgalvao Any clever way to put the version with string interpolation? The benefits are probably minimal, but just wondering, as this is a case where before and after / wouldn't work.

Casks/powermate.rb
```diff
@@ -1,8 +1,8 @@
 cask 'powermate' do
-  version '1.2.4_0'
+  version '1.2.4'
   sha256 '6fd755961bb5a486cb3ab9d47217f7877d4bb021134f32a3c32ceede3fd3cf3d'
 
-  url "http://support.griffintechnology.com/sites/default/files/PowerMate_v#{version}.zip"
+  url "https://support.griffintechnology.com/wp-content/uploads/sites/3/2015/08/PowerMate_v1.2.4.zip"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

vitorgalvao:
> this is a case where before and after / wouldn't work

It could with slight trickery:

```ruby
version '1.2.4/3.2015.08'
url "https://support.griffintechnology.com/wp-content/uploads/sites/#{version.after_slash.dots_to_slashes}/PowerMate_v#{version.before_slash}.zip"
```

boneskull:
used single quotes

vitorgalvao:
Will make some changes, in light of https://github.com/caskroom/homebrew-cask/pull/16314.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in f7d244f52adabf5f7c7d486244885e71dee9f9c0. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

boneskull:
tyvm

boneskull:
@vitorgalvao Semantically, the version number `1.2.4,3/2015/08` is not representative of the actual version, which is `1.2.4`.  Coming from a place of ignorance, I would put more importance on the version being accurate.  So, out of curiosity, why was the change necessary?

vitorgalvao:
> Semantically, the version number

Doesn’t matter. Being able to interpolate as much as possible (while still making sense) takes priority. This is useful because updating as little as possible from casks is highly desirable (for various reasons, including maintainability).

Being “correct” about versions is impossible in some cases, and brings no advantage in any case.

PR #16279, opened 2016-01-02
Use /usr/bin/env bash instead of /bin/bash in a number of developer scripts

Amorymeltzer:
I left alone anything I thought a contributor might reasonably be expected to run in the course of creating a cask, which basically means `list_*` and `generate_cask_token` (See #8023 and #8021).<sup>1</sup>

That being said, I'm a firm believer in [using bash as defined in the user's environment](https://github.com/caskroom/homebrew-cask/pull/9180) so if there are no objections...

---

1: Although since ruby 2.0 is required anyway, this might as well fail imo

### DIFF
--- a/developer/bin/cask-switch-https
+++ b/developer/bin/cask-switch-https
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -o pipefail
 
--- a/developer/bin/develop_brew_cask
+++ b/developer/bin/develop_brew_cask
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # develop_brew_cask
 #
--- a/developer/bin/generate_man_pages
+++ b/developer/bin/generate_man_pages
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # generate_man_pages
 #
--- a/developer/bin/irregular_cask_whitespace
+++ b/developer/bin/irregular_cask_whitespace
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # irregular_cask_whitespace
 #
--- a/developer/bin/project_stats
+++ b/developer/bin/project_stats
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # project_stats
 #
--- a/developer/examples/brewcask-showargs
+++ b/developer/examples/brewcask-showargs
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # brewcask-showargs
 #


vitorgalvao:
I don’t exactly oppose, as it likely doesn’t make a difference, but also don’t see why we should. While the user’s default `bash` may change, the OS’s won’t (since Apple can’t move to a later `bash` thanks to the GPL<sup>1</sup>), and it’s always better to write your script and be sure it’ll run exactly the same for every user.

Seems like we gain nothing by the change, but may lose something<sup>2</sup>.

---

<sup>1</sup> Funny (read, not funny at all) how the supposedly “free” licenses give you so little freedom (leeway) in how to use what they give you.

<sup>2</sup> Highly unlikely, since `bash` must be a hole of bad legacy decisions that can’t be changed for compatibility.

jawshooah:
I don't care much either way, though I understand @vitorgalvao's concerns about compatibility. Using `/bin/bash` ensures that we know exactly which `bash` is running our scripts, all of the time (unless a dev explicitly runs `bash cask-switch-https` or something).

> Funny (read, not funny at all) how the supposedly “free” licenses give you so little freedom (leeway) in how to use what they give you.

GNU greybeards hate commercial software. Like, **hate** it. It's like a religion or something. The thought of anyone making any money from their work without completely open-sourcing their software is just appalling to them. I don't get it.

Amorymeltzer:
I won't disagree with either of you, but that's why I kept this to only things that shouldn't be frequently used project-wide, where I don't feel "consistency" is a strong enough argument.

With that in mind, though, `update_issue_template_urls` uses `env bash`, and `audit_modified_casks`,<sup>1</sup> `generate_issue_template_urls`, `list_login_items_for_app`, `list_running_app_ids`, and `the_long_tail` all use `env ruby`, which is certainly inconsistent.  Would you both be in favor of changing those, then?  Some of those are more important day-to-day.

-----
If it's not PD it's not free.  Although I'm curious to see the future; there have been enough people wanting OSS on the app store that something could be inbound.

------
1: Although I think maybe travis needs this

vitorgalvao:
My concern here wasn’t related to consistency, but to making sure things don’t break. I don’t mind `env ruby` because different OS X releases ship different versions of `ruby`, which means using the system `ruby` as opposed to the user’s `ruby` doesn’t offer any guarantees, anyway. However, using the system `bash` offers us the guarantee of what version it is, and since `bash` is a finickier language, sticking to what we are sure will work seems like a good bet.

In this case, I simply do not see a good argument to using `env bash`.

> Would you both be in favor of changing those, then?

Yes, I would be in favour of making them all consistent *within their own languages* (so using `env` in one `ruby` script means they all should use it as well, but that doesn’t mean all `bash` scripts would need to follow the rule as well, and those could have their own different rule). Typically, in my own scripts, I `#!/bin/bash` and `#!/usr/bin/env ruby`.

---

> GNU greybeards hate commercial software. Like, **hate** it.

One of the worst attitudes I’ve seen (read) from Stallman was a student at a college he gave a talk writting him a really polite and considerate letter arguing how he believed that his (Stallman’s) approach of telling people to blindly abandon all commercial software they use at once and start using “free” software was unrealistic. He made a really compelling case of suggesting an alternative attitude, teaching how to make the switch, which software to use for common cases that create hurdles, and the like. It was clear from the letter (and he indeed mentioned it) he was supportive of Stallman’s stance but was mostly worried that his approach was ineffective due to how he presented it, since without clear alternatives, people won’t know how to make the switch, and the whole attitude as a whole made people dismiss his message. Stallman’s reply was only (to the effect of) “I won’t discuss anything with people who don’t share my views”.

He literally believes his choices are the best possible, without knowing what the others are. [Case in point](https://twitter.com/vhgalvao/status/599990666182205442).

> If it's not PD it's not free.

Agreed. All my github repos are under The Unlicense for this reason.

Amorymeltzer:
@vitorgalvao I had basically your same response recently reading his 15-year old "[Because I'm a bastard, and proud of it](https://lwn.net/2000/0914/a/lt-debugger.php3).  Just the worst possible attitude.

vitorgalvao:
> Because I'm a bastard, and proud of it

I wasn’t expecting he had written those exact words. It’s even more jarring to see that he did.

jawshooah:
> "Because I'm a bastard, and proud of it."

That's Linus Torvalds btw, not Richard Stallman. Still a bastard, though :wink:

Amorymeltzer:
Hah, thanks @jawshooah.  Didn't even notice...

PR #16295, opened 2016-01-02
Use documentation format for test reporters

jawshooah:
I don't know about you guys, but I find this format much more useful than the `.............` output by the default reporter.


jawshooah:
Seems Travis doesn't like verbose output :disappointed:

Damn it, Travis

Issue #12991, opened 2015-08-02
Support glob (or regex) expansion in `zap :delete` and `uninstall :delete`


vitorgalvao:
Currently not supported, but many casks would benefit from it.

jawshooah:
Would this apply to `uninstall :delete` as well? If so, I believe this overrides #12864.

vitorgalvao:
Well, yes, it could also apply to `uninstall :delete`. Actually, looking at it, [it is even mentioned in the docs as a desirable feature](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#uninstall-key-delete) (last bullet).

Why would it override #12864, though? Isn’t that pertaining to paths in `~`?

Those same docs need to be revised when #12864 is taken care of, since some of the rules there will be changed. They contradict themselves, actually: you can’t have both “static, absolute paths” and allow globs.

jawshooah:
You're right, I was thinking that glob expansion entailed tilde expansion.

mxl:
@jawshooah Do you plan to implement this feature?

vitorgalvao:
@mxl If you’re asking so you can go ahead and submit a PR, I’d say go for it. @jawshooah has been pretty busy lately, and I doubt he’d object.

jawshooah:
@mxl Not in the near future, no. Please go ahead and submit a PR at your leisure :+1:

mxl:
@jawshooah @vitorgalvao https://github.com/Homebrew/brew/pull/2042

Issue #9921, opened 2015-03-06
Linter/style checker integration


jawshooah:
What do you all think about integrating a static analysis tool like [RuboCop](https://github.com/bbatsov/rubocop) into the project? It's exceedingly useful for enforcing style conventions, checks for a number of logical errors, and is completely configurable. It could be added to the Travis build so PR submitters know to fix something without maintainer intervention.

I've already started working on this in [my own branch](https://github.com/jawshooah/homebrew-cask/tree/feature/rubocop-integration), and am currently using RuboCop in conjunction with [Overcommit](https://github.com/causes/overcommit) and other tools for all of my personal projects.

Does this sound like something you guys would like to pursue?

phinze:
I personally think this would be _AMAZING_ because we could toss it into Travis and automate a ton of the common maintainer tasks that are manual today. :+1:

jawshooah:
I thought so :smile: I'm happy to do a lot of the legwork here if everyone agrees it's a good idea.

phinze:
...though I'm curious as to what you're thinking RuboCop would be useful for that we couldn't just implement as a normal Rspec test? It seems like RuboCop and Overcommit are sort of big libraries to pull in out of the gate - perhaps there's a smaller set of goals we can flesh out here before we decide on tooling?

jawshooah:
RuboCop is useful in addition to RSpec for the same reason that [JSHint](https://github.com/jshint/jshint) is useful in addition to [Nodeunit](https://github.com/caolan/nodeunit), or [Checkstyle](https://github.com/checkstyle/checkstyle) in addition to [JUnit](http://junit.org/), or [Scalastyle](http://www.scalastyle.org) in addition to [ScalaTest](http://scalatest.org), or any other combination of linters and testing frameworks: style matters.

RSpec is awesome for verifying correctness, but it tells you nothing about the quality of the implementation. [Style guides](https://github.com/bbatsov/ruby-style-guide) exist for many reasons, and tools like RuboCop exist to enforce adherence to those style guides, or whatever modified version thereof you have decided is right for your repository.

Just to give you an idea, running RuboCop with the default configuration against the current `master` returns [over 200 offenses](https://gist.github.com/jawshooah/c1321612f3c11e080fd4) with just the linter errors enabled. Counting style errors, it's [over 13,000](https://gist.github.com/jawshooah/ade7c50ce2684df70ee1).

[...]

Issue #8202, opened 2014-12-17
List of all per-Cask todo comments


iggyvolz:
Below is a list of all todos inside of Homebrew Cask, sorted by todo name and file.  For those curious, here's how I generated it: https://gist.github.com/iggyvolz/93f17a54ad2e22077fc0

"Improve this machine-generated value" has thousands of uses so it is simply the number that is listed.
[`bittorrent-sync.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/bittorrent-sync.rb)

- [x] "response was not XML" in [`bittorrent-sync.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/bittorrent-sync.rb#L6)
- [ ] "glob/expand needed here" in [`clamxav.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/clamxav.rb#L20)
- [x] "the URL indicates there is an appcast" in [`colorpicker-propicker.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/colorpicker-propicker.rb#L5)
- [ ] "the nested_container hack must be revised when container :nested stops being an alias around 0.50.0" in [`decitime.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/decitime.rb#L15)
- [x] "should this be" in [`dwarf-fortress.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/dwarf-fortress.rb#L9)
- [ ] "verify that this does not contain user-generate content" in [`filezilla.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/filezilla.rb#L10)
- [ ] "is this user-created content?" in [`fontexplorer-x-pro.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/fontexplorer-x-pro.rb#L12)
- [ ] "This is a horrible hack […]. The backend code should be fixed" in [`git-annex.rb`](../blob/1d60a07c4bdf8ea54659ccd553090964ea1ef857/Casks/git-annex.rb#L8-L9)
- [ ] "remove all ENV variables" in [`gpgtools.rb`](../blob/db9b3de003b20a4489d3c73700ff0e468c7dbc02/Casks/gpgtools.rb#L12)
[...]

PR #16298, opened 2016-01-03
Enforce a few more rubocop guidelines, minor stylistic fixups

Amorymeltzer:
1. Only two casks (`java` and `applepi-baker`) had `:cookies` on three lines, moved 'em to one line for the sake of convenience and matching. (Technically not rubocop, but felt silly doing a PR for this by it's lonesome)<sup>1</sup>
2. Rubocop: Removed whitespace following the cask header in 29 casks, should enforce `no_empty_lines` in `Style/EmptyLinesAroundBlockBody`;   @jawshooah You were originally unsure if this was worthwhile, I rather think it's a good idea, if only for regex/parsing.
3. Rubocop: Tidied up casks based on Style/AlignHash and Style/AlignParameters (also one each from Style/CommentIndentation<sup>2,3</sup> and Style/EmptyLineBetweenDefs)
4. Rubocop: Removed spaces found inside square brackets, should enforce `Style/SpaceInsideBrackets`
5. ~~Rubocop: Enforce `Style/Tab` and `Style/TrailingWhitespace` (no current errors, intended from the start in #9921, they are indeed "gross")~~

Ping @jawshooah via #9921.

---

1: I couldn't get `applepi-baker` to fail without it, though, so perhaps it's no longer necessary?
2: Comments in `soulver` and `gpgtools` are waiting on glob support (#12991; see also #8202).
3: Comments in `ableton-*` are user-generated content being kept until `:trash` is properly implemented per https://github.com/caskroom/homebrew-cask/pull/11587#discussion_r31761341


jawshooah:
I was actually waiting for them to release `0.36.0` before doing this, since there will be a couple of upcoming [additions](https://github.com/bbatsov/rubocop/pull/2529) and [fixes](https://github.com/bbatsov/rubocop/issues/2565) that will make things much nicer for us.

.rubocop.yml
```diff
@@ -33,3 +36,12 @@ Style/FileName:
 
 Style/HashSyntax:
   EnforcedStyle: hash_rockets
+
+Style/SpaceInsideBrackets:
+  Enabled: true
+
+Style/Tab:
+  Enabled: true
```
jawshooah: This is [enabled by default](https://github.com/bbatsov/rubocop/blob/v0.35.1/config/enabled.yml#L726-L729), no need to make it explicit.

.rubocop.yml
```diff
@@ -33,3 +36,12 @@ Style/FileName:
 
 Style/HashSyntax:
   EnforcedStyle: hash_rockets
+
+Style/SpaceInsideBrackets:
+  Enabled: true
```
jawshooah: [Same here.](https://github.com/bbatsov/rubocop/blob/v0.35.1/config/enabled.yml#L667-L670)

.rubocop.yml
```diff
@@ -23,6 +23,9 @@ Performance/StringReplacement:
 Style/BarePercentLiterals:
   EnforcedStyle: percent_q
 
+Style/EmptyLinesAroundBlockBody:
+  EnforcedStyle: no_empty_lines
```
jawshooah: [Same here.](https://github.com/bbatsov/rubocop/blob/v0.35.1/config/default.yml#L378-L379)

.rubocop.yml
```diff
@@ -33,3 +36,12 @@ Style/FileName:
...
   EnforcedStyle: hash_rockets
+
+Style/SpaceInsideBrackets:
+  Enabled: true
+
+Style/Tab:
+  Enabled: true
+
+Style/TrailingWhitespace:
+  Enabled: true
```
jawshooah: [Same here.](https://github.com/bbatsov/rubocop/blob/v0.35.1/config/enabled.yml#L741-L744)

jawshooah:
All that aside, the changes look fine (and won't prevent us from doing more once we get a new RuboCop release), except for one thing: `AlignHash`. I'm partial to the `table` style (align keys with keys, hashrockets with hashrockets, and values with values). Not a big deal either way, but we should come to a consensus on that before merging.

@vitorgalvao, you're a graphic designer (sorry, I stalked your website), so you probably have a better eye for this than me. What would you suggest? (examples of the different styles are [here](https://github.com/bbatsov/rubocop/blob/v0.35.1/config/default.yml#L75-L88)).

Amorymeltzer:
@jawshooah My thinking with the `AlignHash` changes was to just get those nine casks in line with the rest of them; the issue for them all was key alignment being way out of whack.

Personally I prefer the `key` design — less mucking about with spaces and aligning — but I was saving that conversation for later, to be had alongside `Style/SingleSpaceBeforeFirstArg` (per your comments at #9921)

vitorgalvao:
> (sorry, I stalked your website)

Well, it is public after all, but it is in dire need of an update in every respect (the website itself and the works in it). It was the first html/css (as opposed to flash) website I’ve made, though, so with that mind I guess it isn’t *that* bad (well, a bit, I should really get to it eventually).

As for layout, I’d say `separator` is the only “wrong” one. When we have a list of things, having them start at different points (since we’re reading left-to-right and they’re aligned right-to-left) makes it harder to scan with good comprehension.

Personally (as in a matter of taste) I *like* `key` better, but I think `table` *works* better for reading/skimming/comprehension and that is more important. Users seem to be somewhat partial to it when submitting, as well. [Autocorrect support](https://github.com/caskroom/rubocop-cask/issues/1#issue-124013830) should take care of @Amorymeltzer’s concerns (i.e. no need to manually muck about with spacing) and popular text editors have plugins for this type of alignment. I say we go with `table`.

On a tangential note, I remember you (@jawshooah) asking about (months ago) why the heck do we only leave one space when indenting certain things:

```ruby
:symbol => [
            'one',
            'two'
           ]
```

That is indeed a weird pattern, and I’d be in favour of changing that to two spaces.

Amorymeltzer:
I meant it more for contributors, manually spacing to make travis happy, but that is sort of the point of this whole thing.

One space is weird, weird, and so weird (speaking of manually spacing).

jawshooah:
@vitorgalvao that particular oddity can't be fully rectified until a new release including [this PR](https://github.com/bbatsov/rubocop/pull/2529) is pushed.

In any case, I say we wait on this for now. @bbatsov has alluded to a release coming out in the next week or so.

Amorymeltzer:
Mooted via #16317

PR #16302, opened 2016-01-03
Add Rest Timer

niksy:


Casks/rest.rb
```diff
@@ -0,0 +1,16 @@
+cask 'rest' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://0.dangelov.com/rest-trial-mac'
```
vitorgalvao: I have no idea where you got this URL. It’s not at all what I see on the website.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in a1187db893e43758d996a002a3b8baa151727411. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

Casks/rest.rb
niksy: When you go to the homepage and click on a [link to download a trial](https://resttimer.com/en/download-trial/mac), there is a timeout of maybe 2-3 seconds which will trigger automatic download. If it doesn’t do that, developer has giver the user option to download the app manually: "Your download should start shortly. If it doesn't, please click here to start it manually." That "click here" contains the link I provided in `url`.

vitorgalvao: I see it now. The automatic download is from the same domain as the webpage, though, so it is preferred.

PR #16304, opened 2016-01-03
A Happy New Year!

sachin21:
Updated to 2016 :dancers:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright © 2013-2015, Paul Hinze & Contributors
+Copyright © 2013-2016, Paul Hinze & Contributors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,


CONTRIBUTING.md
```diff
@@ -415,7 +415,7 @@ We maintain separate Taps for different types of binaries. Our nomenclature is:
 
 ### Stable Versions
 
-Stable versions live in the main repository at [caskroom/homebrew-cask](https://github.com/caskroom/homebrew-cask). They should run on the latest release of OS X or the previous point release (in 2015, for example, that meant El Capitan and Yosemite).
+Stable versions live in the main repository at [caskroom/homebrew-cask](https://github.com/caskroom/homebrew-cask). They should run on the latest release of OS X or the previous point release (in 2016, for example, that meant El Capitan and Yosemite).
```
vitorgalvao: This case shouldn’t change. In 2016 it won’t mean El Capitan and Yosemite, but Whatever Comes Next and El Capitan. The reason we specify a year here is precisely so we don’t have to keep changing it.

sachin21: Thanks! fixed :smile_cat:

vitorgalvao: Thank you.

PR #16314, opened 2016-01-03
core: slash comma colon

vitorgalvao:
I haven’t included `comma` and `colon` as dividers (and no extra methods to replace/delete them) since these are special characters (shouldn’t appear organically in `url` and the like) and I think we should treat them as such, i.e. say “these are special and to be used sparingly in particularly difficult cases; there should be no more than one of each per `version`”.

Will fix the documentation after discussion.


jawshooah:
LGTM, I'm fine to merge as soon as Travis goes green.

vitorgalvao:
In that case, [documentation](https://github.com/caskroom/homebrew-cask/pull/16315).

PR #16315, opened 2016-01-03
documentation: extend (comma and colon) and simplify version methods

vitorgalvao:


adityadalal924:
Documentation LGTM. Should be good to merge both the documentation and the actual changes.

jawshooah:
:+1:

CONTRIBUTING.md
```diff
@@ -126,32 +126,21 @@ url "http://example.com/#{version.sub(%r{build\d+}, '')}/file-version-#{version}
...
-| `no_slashes`             | `1.2.3-a45_b67/ccdd88` | `1.2.3-a45_b67ccdd88`  |
-| `no_dividers`            | `1.2.3-a45_b67/ccdd88` | `123a45b67ccdd88`      |
+| Method                   | Input              | Output             |
+|--------------------------|--------------------|--------------------|
+| `major`                  | `1.2.3-a45,ccdd88` | `1`                |
+| `minor`                  | `1.2.3-a45,ccdd88` | `2`                |
+| `patch`                  | `1.2.3-a45,ccdd88` | `3`                |
+| `major_minor`            | `1.2.3-a45,ccdd88` | `1.2`              |
+| `major_minor_patch`      | `1.2.3-a45,ccdd88` | `1.2.3`            |
+| `before_comma`           | `1.2.3-a45,ccdd88` | `1.2.3-a45_b67`    |
```
tsparber: _b67 should maybe removed :)

vitorgalvao: Nice catch. Thank you.

adityadalal924:
All of this should also be added into CASK_LANGUAGE_REFERENCE *somewhere*, as that should be the definitive guide - right now information is split between the two (CONTRIBUTING is more up to date) - thoughts for another PR, though.

vitorgalvao:
> All of this should also be added into CASK_LANGUAGE_REFERENCE

The reason I haven’t done so is this doesn’t actually concern the cask language itself, it’s a feature provided by using ruby (a custom built feature, but still). I see *version methods* akin to the *token reference*: yes, they are specific to casks, but they exist not to *enable* the language of casks, and instead to *support* it.

They define rules (token reference) and provide enhancements (version methods) but aren’t needed to define what a cask is.

All that said, I’ll reiterate that is why I haven’t done so, not that I think it shouldn’t be done.

PR #16320, opened 2016-01-03
updated blackvue-viewer (latest)

adidalal:


adityadalal924:
@vitorgalvao `cask-repair` shouldn't override a shasum manually set as `:no_check` (ie. this example)

vitorgalvao:
> `cask-repair` shouldn't override a shasum manually set as `:no_check`

For some reason, I’m not sure I’m comfortable with changing that (can’t really put my finger on why). Either one, it is done.

adityadalal924:
Latest version is actually `1.34_20151215` which doesn't seem to play nice with the existing URL (and I'm not sure where this URL even came from)

vitorgalvao:
To be clear (not sure it was, now that I reread it), I have indeed made the change to `cask-repair`.

adityadalal924:
Ah sounded like you didn't. Thanks for the clarification.

PR #16324, opened 2016-01-03
Delete padre.rb

adidalal:
Link on http://padre.perlide.org/download/osx.html leads to nonexistent page


adityadalal924:
Merged as 83fe7566a7d53ae6a73973a90a434c571b5fbf7c.

PR #16334, opened 2016-01-03
Apply RuboCop to test casks

jawshooah:
Also be more explicit about what we are excluding in `.rubocop.yml`. This allows us to consistently lint test casks, and also will make future updates to the list of files we lint a bit easier.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,7 @@ AllCops:
     - '**/Gemfile'
     - '**/Rakefile'
     - '**/brew-cask.rb'
+    - '**/invalid-header-format.rb'
     - '**/spec/*.rb'
     - '**/spec/support/*.rb'
     - '**/test/*.rb'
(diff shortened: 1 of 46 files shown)


adityadalal924:
:+1: 

Makes sense and is a fairly small (and obvious) jump from https://github.com/caskroom/homebrew-cask/pull/16317 to this.

adityadalal924:
Travis was backed up due to a bunch of commits on my part, but finally went through:

```ruby
Offenses:
test/support/Casks/invalid/invalid-header-format.rb:1:11: E: unexpected token tASSOC
test_cask => 'invalid-header-format' do
          ^^
test/support/Casks/invalid/invalid-header-format.rb:9:1: E: unexpected token kEND
end
^^^
test/support/Casks/with-conditional-caveats.rb:11:42: W: Literal false appeared in a condition.
    puts 'This caveat is conditional' if false
                                         ^^^^^
3059 files inspected, 3 offenses detected
RuboCop failed!
```

jawshooah:
Ah, didn't catch that in the output locally.

All we can do for `invalid-header-format` is exclude it, since it's intended to be unparseable. For `with-conditional-caveats`, we could either throw in some dummy logic or exclude it from that particular cop.

adityadalal924:
Excluding both of them seems like the easiest option, IMO.

PR #16341, opened 2016-01-04
Add Fuse for OS X (a ZX Spectrum emulator), 1.1.1.1, stable

yaroslav:
http://fuse-for-macosx.sourceforge.net/Site/Fuse_for_Mac_OS_X/Fuse_for_Mac_OS_X.html

### DIFF
--- a/Casks/fredm-fuse.rb
+++ b/Casks/fredm-fuse.rb
@@ -1,0 +1,11 @@
+cask 'fredm-fuse' do
+  version '1.1.1.1'
+  sha256 '28533a6b1c31d1ca76b5509a408fb587f232b24392a0ffe1e45dac8e0a3e2880'
+
+  url "http://downloads.sourceforge.net/sourceforge/fuse-for-macosx/fuse-for-macosx/#{version}/FuseForMacOSX-#{version}.zip"
+  name 'Fuse for Mac OS X'
+  homepage 'http://fuse-for-macosx.sourceforge.net/'
+  license :gpl
+
+  app 'Fuse for Mac OS X/Fuse.app'
+end


Casks/fuse-for-macosx.rb
```diff
@@ -0,0 +1,11 @@
+cask 'fuse-for-macosx' do
+  version '1.1.1.1'
+  sha256 '28533a6b1c31d1ca76b5509a408fb587f232b24392a0ffe1e45dac8e0a3e2880'
+
+  url "http://downloads.sourceforge.net/sourceforge/fuse-for-macosx/fuse-for-macosx/#{version}/FuseForMacOSX-#{version}.zip"
+  name 'Fuse for Mac OS X, a ZX Spectrum emulator'
```
vitorgalvao: `, a ZX Spectrum emulator` needs to go. That’s a description, not a name.

yaroslav: @vitorgalvao It's there so that people won't mix it with "actual" fuse (below), which is way more popular. Seemed reasonable, according to https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#name-stanza-details.

```
❯ brew cask search fuse
==> Exact match
fuse
==> Partial matches
fender-fuse		osxfuse
```

Casks/fuse-for-macosx.rb
```diff
@@ -0,0 +1,11 @@
+cask 'fuse-for-macosx' do
```
vitorgalvao: This has to be changed to `cask 'fredm-fuse' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `fredm-fuse.rb`.

`fredm` is based on the creator’s email. Unless we get a better vendor name, that will have to do.

Casks/fuse-for-macosx.rb
vitorgalvao: `fuse-for-macosx` is easily confused with `osxfuse`.

`name` is a stanza. The token (whose reference I linked you to) is what matters here. We need strict rules for tokens, and making an exception for such a minor app is not a good idea. We have [clear rules for when there are conflicts](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#forks-and-apps-with-conflicting-names).

vitorgalvao: Nowhere in that documentation says descriptions should be included. Again, please remove it.

Casks/fuse-for-macosx.rb
yaroslav: @vitorgalvao sure, I will update the PR, thanks for the clarification.

yaroslav:
@vitorgalvao done!

vitorgalvao:
Thank you. Merging as soon as Travis goes green.

adityadalal924:
Thank you for the contribution. It was merged directly as 338e2358115db89776201e86fded6c627efd1b51 to keep commit history cleaner. Your contribution is still credited to you.

PR #16351, opened 2016-01-04
fuse.rb: improve name

vitorgalvao:


adityadalal924:
Shouldn't this be fuse according to the token reference, though? I realize there's a conflict - so is that the criteria for an exception?

vitorgalvao:
This is a `pkg`, so rules aren’t as clear. I’ve reverting the file name change now, though. Looking at it further, `fuse` still makes the most sense.

adityadalal924:
Merged as bf3745be59d0d8098dbcecad5fc77e4ac7c6e81f.

PR #8871, opened 2015-01-13
increment seed for Travis tests

rolandwalker:
hack to ward off spurious test failures

PR #16367, opened 2016-01-04
Updates to Rake tasks

jawshooah:
- Use `pry` in `console` task for nicer formatting and debugging tools
- Move man page generation to `man` rake task
- Reorganize test tasks
  - Move `rspec` and `minitest` tasks under the `test` namespace
  - Create separate `coverage` task for generating coverage reports from tests


adityadalal924:
Don't know enough about Ruby testing to have an opinion either way, but LGTM.

Why is the `seed` needed?

jawshooah:
> Why is the seed needed?

There's some temporal dependency in our integration tests that requires certain tests be run before others, and it's been there for a long time (#5037, #5064, #5613, #6611, #8871). Nobody has put in the time to root it out yet.

Issue #16344, opened 2016-01-04
Error: Cask 'ionic-lab' definition is invalid: Bad header line: parse failed


pkoppstein:
```
$ brew search iconic-lab
...
This formula was found in a tap:
Caskroom/cask/ionic-lab
To install it, run:
  brew install Caskroom/cask/ionic-lab

$ brew install Caskroom/cask/ionic-lab
==> brew cask install Caskroom/cask/ionic-lab
Error: Cask 'ionic-lab' definition is invalid: Bad header line: parse failed

$ brew --version
Homebrew 0.9.5 (git revision 077e0d; last commit 2016-01-04)
```

jawshooah:
Please *always* follow the [reporting bugs](https://github.com/caskroom/homebrew-cask#reporting-bugs) guide. To do so, edit the top post and then comment after this one so I can check back.

Closing in the meantime, because a number of people never report back and it’s useless to keep the issue open in that case and without all the required information.

pkoppstein:
OK, running `brew uninstall --force brew-cask; brew update` did the trick; there was no clue that there has been a switch to a "new system".  Is that a bug of some kind?

vitorgalvao:
> there was no clue that there has been a switch to a "new system". Is that a bug of some kind?

[Huge bolded message at the top of the README](https://github.com/caskroom/homebrew-cask#homebrew-cask). Also, `brew-cask` itself outputted a message when you updated to the latest version.

pkoppstein:
@vitorgalvao - You might try to see things from a homebrew user's point of view. In short, I ran `brew update` (which gave no message or hint about having to uninstall brew-cask), and then the following commands, shown here with their output so you can see that there was again no message or hint about having to uninstall brew-cask:

```
$ brew install ionic-lab
Error: No available formula with the name "ionic-lab" 
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
This formula was found in a tap:
Caskroom/cask/ionic-lab
To install it, run:
  brew install Caskroom/cask/ionic-lab

$ brew install Caskroom/cask/ionic-lab
==> brew cask install Caskroom/cask/ionic-lab
Error: Cask 'ionic-lab' definition is invalid: Bad header line: parse failed

```

vitorgalvao:
> You might try to see things from a homebrew user's point of view.

[...]

PR #16368, opened 2016-01-04
Improve error message suggestions

jawshooah:
- Print error message with suggestions if header parse failed
- Add `brew uninstall --force brew-cask` to error suggestions

Refs #16344

### DIFF
--- a/developer/bin/generate_cask_token
+++ b/developer/bin/generate_cask_token
@@ -384,9 +384,9 @@ def report
   puts "Proposed file name:           #{cask_file_name}"
   puts "Cask Header Line:             cask '#{cask_token}' do"
   if warnings.length > 0
-    STDERR.puts "\n"
-    STDERR.puts warnings
-    STDERR.puts "\n"
+    $stderr.puts "\n"
+    $stderr.puts warnings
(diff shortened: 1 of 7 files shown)


adityadalal924:
Should prevent some user confusion, which is always nice (through I guess the people with such errors might still be tapped to `phinze` anyways)...

jawshooah:
> I guess the people with such errors might still be tapped to phinze anyways

The funny thing is, since the [phinze repo](https://github.com/phinze/homebrew-cask) just redirects to us now, if you `brew tap phinze/cask` you get the same contents as if you had tapped `caskroom/cask`. With that in mind, you'd think `brew update` would be all that's needed to clean up, but apparently not.

jawshooah:
Added `brew untap phinze/cask` to the suggestions in e429666 just to be safe. It's a bit long, though.

adityadalal924:
It is long - though hopefully seeing such a descriptive error message will help them fix the issue by themselves. How about merging it for the time being, and seeing if that prevents most of the (same) issues being opened up? We can always remove the extra two commands, say, six months down the line.

PR #16374, opened 2016-01-05
Add sync-messenger provided by Wantedly

izumin5210:
Add a new cask for Sync.app, group-chat service client

### DIFF
--- a/Casks/sync-messenger.rb
+++ b/Casks/sync-messenger.rb
@@ -1,0 +1,11 @@
+cask 'sync-messenger' do
+  version '1.3.0'
+  sha256 'b1cc06995d222083b2f6cce7b3a96307e52cc26d4aadd066cbdfaf7e555740e8'
+
+  url 'https://wtd-sync-update-channel.herokuapp.com/download/latest?platform=darwin&channel=production'
+  name 'sync-messenger'
+  homepage 'https://www.wantedly.com/sync'
+  license :unknown # todo: change license and remove this comment; ':unknown' is a machine-generated placeholder
+
+  app 'Sync.app'
+end
--- a/Casks/sync-messenger.rb
+++ b/Casks/sync-messenger.rb
@@ -5,7 +5,7 @@ cask 'sync-messenger' do
   url 'https://wtd-sync-update-channel.herokuapp.com/download/latest?platform=darwin&channel=production'
   name 'sync-messenger'
   homepage 'https://www.wantedly.com/sync'
-  license :unknown # todo: change license and remove this comment; ':unknown' is a machine-generated placeholder
+  license :unknown # TODO: change license and remove this comment; ':unknown' is a machine-generated placeholder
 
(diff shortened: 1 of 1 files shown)


izumin5210:
Fix comment for pass the broken test

Casks/sync-messenger.rb
```diff
@@ -0,0 +1,11 @@
+cask 'sync-messenger' do
+  version '1.3.0'
+  sha256 'b1cc06995d222083b2f6cce7b3a96307e52cc26d4aadd066cbdfaf7e555740e8'
+
+  url 'https://wtd-sync-update-channel.herokuapp.com/download/latest?platform=darwin&channel=production'
+  name 'sync-messenger'
```
vitorgalvao: This is supposed to be the name as it appears on the website, not simply a repeat of the token. See [documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#name-stanza-details).

Casks/sync-messenger.rb
```diff
@@ -0,0 +1,11 @@
+cask 'sync-messenger' do
```
vitorgalvao: This has to be changed to `cask 'sync' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `sync.rb`.

Since there is [a token conflict](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#forks-and-apps-with-conflicting-names), this needs to be `wantedly-sync`.

Casks/sync-messenger.rb
```diff
@@ -0,0 +1,11 @@
...
+  version '1.3.0'
+  sha256 'b1cc06995d222083b2f6cce7b3a96307e52cc26d4aadd066cbdfaf7e555740e8'
+
+  url 'https://wtd-sync-update-channel.herokuapp.com/download/latest?platform=darwin&channel=production'
```
vitorgalvao: I can’t verify this download url, it seems like it is behind a registration wall. [If that is the case we cannot accept this](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#unofficial-vendorless-and-walled-builds) and you must submit it to [caskroom/unofficial](https://github.com/caskroom/homebrew-unofficial) instead.

adityadalal924: Looks like it needs to go to `unofficial` (and a comment added about the hostname being different from url also)

vitorgalvao: Please submit it there, then, @izumin5210. All the other requests still apply. Closing.

@adityadalal924 

> and a comment added about the hostname being different from url also

In this case it won’t apply, since we maintainers couldn’t check it anyway and users won’t be able to confirm it.

PR #16376, opened 2016-01-05
Add Google Chrome Beta v48.0.2564.48

clawfire:


vitorgalvao:
Thank you, but betas belong in [caskroom/versions](https://github.com/caskroom/homebrew-versions) (this is already there).

Issue #16318, opened 2016-01-03
Revert `rubocop` to latest tagged release in Gemfile


adidalal:
Go back to tracking latest release instead of `bbatsov/rubocop/master`

To be done after v0.36 is released

Ref: https://github.com/caskroom/homebrew-cask/pull/16317 (specifically https://github.com/jawshooah/homebrew-cask/commit/b03bfcdb1105cfedac4b468a446739e5f82fc2e7)

adityadalal924:
ping @jawshooah It's out!
https://github.com/bbatsov/rubocop/releases

jawshooah:
It is! Just need to clean up some stuff in rubocop-cask before we update
here.

jawshooah:
Done in #16377.

PR #16377, opened 2016-01-05
Add `style` subcommand to run RuboCop against casks

### LINKED ISSUE #14736: Integrate RuboCop into `brew cask audit`
As discussed in [#14713 (comment)](https://github.com/caskroom/homebrew-cask/issues/14713#issuecomment-151677660), we should integrate the [RuboCop](https://github.com/bbatsov/rubocop) style check into `brew cask audit`.

This would obviate the need to run RuboCop directly in the Travis build, and reduce the amount of effort required of cask contributors. We could even install the `rubocop` gem if it doesn't exist, *à la* [`brew style`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/style.rb) (or `brew audit --strict`).

jawshooah:
Closes #14736

This is on hold until a new version of RuboCop has been released ([soon!](https://github.com/bbatsov/rubocop/issues/2551#issuecomment-169826865)) and #16318 closed.

~~Considering that we will run RuboCop against all submissions anyway, it's debatable whether we should keep `style` and `audit` separate as Homebrew does.~~ `style` and `audit` will be kept separate.

~~Still needs tests and documentation, but I figured I'd throw it up here for review.~~ Tests and documentation have been added. Now we're just waiting on #16318.


Amorymeltzer:
>Considering that we will run RuboCop against all submissions anyway, it's debatable whether we should keep style and audit separate as Homebrew does.

See https://github.com/caskroom/homebrew-cask/issues/10997#issuecomment-166519416.  The `audit --strict` addition might be blurring the lines.

jawshooah:
> The `audit --strict` addition might be blurring the lines.

Yeah, I figured it would be familiar to Homebrew users, but I agree. Their CI only applies RuboCop to new formulas, which comprise a very small minority of PRs, so it makes sense that you have to opt-in with `--strict`.

We apply RuboCop to *all* casks in incoming PRs, so it might make sense to fold RuboCop into `audit`. There are a couple of issues with that, though:
  * If the contributor does not already have the `rubocop` and `rubocop-cask` gems installed, they will need to have an internet connection to download them<sup>1</sup>
  * We explicitly run RuboCop in `ci/travis/script.sh`, so running it again for modified casks in `developer/bin/audit_modified_casks` seems like overkill<sup>2</sup>

--

1: This is already the case when running `audit --download`, and probably a safe assumption anyway. If the `gem install` fails, we could just print an error message and continue with the rest of the audit.
2: We could add an `audit --no-rubocop` flag or similar to opt out in this case.

adityadalal924:
`brew cask style` makes sense (although it feels like an alias for `bundle exec rubocop cask_name.rb`)

Not a fan of the `--strict` flag, but am impartial to folding style into audit - it might make sense to keep them separate if only because they do different things, and people who actually read CONTRIBUTING (which will hopefully get much shorter soon - https://github.com/caskroom/homebrew-cask/issues/16373) will know to run both (or not, and Travis will do it anyway...)

Amorymeltzer:
@adityadalal924 Basically said what I was trying to get at.  I still think `audit` should test anything that will break `brew cask`, and should be available to a user with minimal requirements: it should work out of the box, without `rubocop`.  `style` can/should incorporate `rubocop` style choices, and we can/should encourage users to take advantage of it before submitting, but I like the idea of keeping the tasks separated.  To wit, `brew cask audit --strict` seems unnecessary in the presence of `style`

vitorgalvao:
In agreement with the latest comments.

jawshooah:
Sounds good.

> `brew cask style` makes sense (although it feels like an alias for `bundle exec rubocop cask_name.rb`)

It basically is, except that it doesn't require the contributor to install *all* of our development dependencies, only `rubocop` and `rubocop-cask`.

jawshooah:
Expecting Travis to fail with e2ab09c until we get a new RuboCop release.

jawshooah:
Had an idea:

* separate `brew cask style` from our internal RuboCop check (once we have one)
  * move configuration from `Casks/.rubocop.yml` to `rubocop-cask` gem and [use inheritance](https://github.com/bbatsov/rubocop#inheriting-configuration-from-a-dependency-gem)
  * remove internal dev dependency on `rubocop-cask` entirely

What do you guys think?

adityadalal924:
So, in my mind, rubocop-cask *is* `brew cask style`, so I'm not sure what the above change would accomplish (other than possibly a more cleanly separated-out config). Am I missing something?

That being said, I personally have no problem with either way (I just bundle install and go on my merry way :p)

jawshooah:
@adityadalal924 Sorry, should have been more clear.

I'd like for us to use RuboCop on our actual codebase, not just casks. But the configs for each will likely be very different, so rather than dancing around with overrides and such, I'm thinking we just make a clean separation. `bundle exec rubocop` will operate exclusively on library and spec files, and `brew cask style` will handle the casks.

adityadalal924:
Got it, that makes a lot more sense.

`bundle exec rubocop` = RuboCop on core library code
`brew cask style` = RuboCop (more specifically rubocop-cask with our custom rules) running on Casks

That makes sense, and with that, I'm all for the above-mentioned separation.

jawshooah:
Here's another question: should the flag for `brew cask style` to auto-correct casks be `--fix` as it is now, or `--auto-correct` like in RuboCop?

adityadalal924:
It can be either - I'm updating CONTRIBUTING right now so anything goes. 

`--fix` seems to be shorter, which would be the only reason "for" it, but if `--auto-correct` works out of the box I don't think it's necessary to add more code just to alias it.

jawshooah:
[How about both?](https://github.com/jawshooah/homebrew-cask/commit/98e69ac4d165b18587d549ad3697580f77f63a1c)

adityadalal924:
Ah yes, good old 
enhanced-25942-1427414324-7

Casks/cisco-spark.rb
```diff
@@ -12,7 +12,7 @@
   app 'Spark.app'
 
   uninstall :signal => [
-                         ['TERM', 'Cisco-Systems.Spark']
+                         ['TERM', 'Cisco-Systems.Spark'],
```
jawshooah: @vitorgalvao @adidalal What do you guys think about the trailing comma after each line in a multiline array or hash literal? I think it makes updating and reordering elements easier, but it doesn't exactly look pretty, especially for single-element literals spread over multiple lines like the ones in this PR.

jawshooah: It's also likely not intuitive for new contributors submitting fresh casks.

PR #16385, opened 2016-01-05
Java caveat

vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/issues/16383.

If merged, we can start fixing:
- [ ] https://github.com/caskroom/homebrew-cask/issues/4885
- [ ] https://github.com/caskroom/homebrew-cask/search?utf8=%E2%9C%93&q=brew+cask+install+java

And others as we notice them.

### DIFF
--- a/doc/CASK_LANGUAGE_REFERENCE.md
+++ b/doc/CASK_LANGUAGE_REFERENCE.md
@@ -219,6 +219,7 @@ The following methods may be called to generate standard warning messages:
 | --------------------------------- | ----------- |
 | `path_environment_variable(path)` | users should make sure `path` is in their `$PATH` environment variable
 | `zsh_path_helper(path)`           | zsh users must take additional steps to make sure `path` is in their `$PATH` environment variable
+| `depends_on_java(version)`        | users should make sure they have the specified version of java installed (`version` should be one of `6`, `7`, or empty for `8`)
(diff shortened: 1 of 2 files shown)


adityadalal924:
As a short term fix - this is good. We can use this until a real depends_on happens. Switching over should be trivial after the fact also, which is nice.

doc/CASK_LANGUAGE_REFERENCE.md
```diff
@@ -219,6 +219,7 @@ The following methods may be called to generate standard warning messages:
 | --------------------------------- | ----------- |
 | `path_environment_variable(path)` | users should make sure `path` is in their `$PATH` environment variable
 | `zsh_path_helper(path)`           | zsh users must take additional steps to make sure `path` is in their `$PATH` environment variable
+| `depends_on_java(version)`        | users should make sure they have the specified version of java installed (defaults to `latest`)
```
jawshooah: Should probably specify that `version` must be an integer, not a full version spec (e.g. `6` or `7`)

vitorgalvao: How about like this?

jawshooah: :+1:

PR #16387, opened 2016-01-05
Delete pycharm-ce.rb

vitorgalvao:
To be added to [caskroom/versions](https://github.com/caskroom/homebrew-versions).


Amorymeltzer:
Should probably update https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#name-stanza-details.  Not sure of a good candidate replacement, maybe [`rro.rb`](https://github.com/caskroom/homebrew-cask/blob/fc05c0353aebb28e40db72faba04b82ca832d11a/Casks/rro.rb#L6#L8) after a cleanup?

vitorgalvao:
We can but we don’t need to. That is still a good example, and it links to a specific revision, so it’ll keep working (also, the cask still exists, even if in a different repo).

vitorgalvao:
Merged as cb68d903ba0dfab27e69d7de645bdccf8af068b6.

PR #16389, opened 2016-01-05
Attempt clearer language for fetch and resources

Amorymeltzer:
It was never particularly clear to me what "resources" meant, in part because I think it was used over-broadly.  Maybe I'm the outlier, but [two years later](https://github.com/caskroom/homebrew-cask/pull/2970) I'm still confused.

Happy to entertain alternative language choices here, I'm not wed to these.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -67,12 +67,12 @@ Note that `uninstall --force` is currently imperfect. See the man page for more 
 
 * `info` — displays information about the given Cask
 * `list` — with no args, lists installed Casks; given installed Casks, lists staged files
-* `fetch` — downloads Cask resources to local cache (with `--force`, re-download even if already cached)
+* `fetch` — downloads remote application files for the given Cask to the local cache (with `--force`, re-download even if already cached)
 * `doctor` — checks for configuration issues
 * `cleanup` — cleans up cached downloads (with `--outdated`, only cleans old downloads)
 * `home` — opens the homepage of the given Cask; or with no arguments, the homebrew-cask project page
 * `update` — a synonym for `brew update`
-* `zap` — try to remove *all* files associated with a Cask (including resources which may be shared with other applications)
+* `zap` — try to remove *all* files associated with a Cask (may include resources shared with other applications)
 
 The following commands are for Cask authors:
 
--- a/doc/src/brew-cask.1.md
+++ b/doc/src/brew-cask.1.md
@@ -57,8 +57,9 @@ names, and other aspects of this manual are still subject to change.
     Open the given Cask definition file for editing.
 
   * `fetch` [--force] <token> [ <token> ... ]:
-    Fetch remote resources for the given Cask to the local cache. With
-    `--force`, force re-download even if the resources are already cached.
+    Download remote application files for the given Cask to the local
+    cache. With `--force`, force re-download even if the files are already
+    cached.
 
   * `home` or `homepage` [<token>]:
     Display the homepage associated with a given Cask in a browser.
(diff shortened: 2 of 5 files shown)


USAGE.md
```diff
@@ -67,12 +67,12 @@ Note that `uninstall --force` is currently imperfect. See the man page for more
 
 * `info` — displays information about the given Cask
 * `list` — with no args, lists installed Casks; given installed Casks, lists staged files
-* `fetch` — downloads Cask resources to local cache (with `--force`, re-download even if already cached)
+* `fetch` — downloads files containing Cask applications to local cache (with `--force`, re-download even if already cached)
```
vitorgalvao: “files containing Cask applications” is confusing to me.

“downloads files need to install applications to local cache (with `--force`, re-download even if already cached)”?

doc/src/brew-cask.1.md
```diff
@@ -123,7 +124,7 @@ names, and other aspects of this manual are still subject to change.
     files. `zap` actions are variable, depending on the level of detail
     defined by the Cask author.
 
-    **`zap` may remove resources which are shared between applications.**
+    **`zap` may remove files which are shared between applications.**
```
vitorgalvao: It’s not just files, though, as `zap` can do everything `uninstall` can.

USAGE.md
vitorgalvao: Better yet, why not just steal directly from the manpage? “download remote application files for the given Cask to the local cache (with `--force`, re-download even if already cached)” seems pretty clear to me.

Amorymeltzer: Hmm, I like that latter suggestion a lot.

USAGE.md
```diff
@@ -67,12 +67,12 @@ Note that `uninstall --force` is currently imperfect. See the man page for more
...
 * `doctor` — checks for configuration issues
 * `cleanup` — cleans up cached downloads (with `--outdated`, only cleans old downloads)
 * `home` — opens the homepage of the given Cask; or with no arguments, the homebrew-cask project page
 * `update` — a synonym for `brew update`
-* `zap` — try to remove *all* files associated with a Cask (including resources which may be shared with other applications)
+* `zap` — try to remove *all* files associated with a Cask (may include any resources shared with other applications)
```
vitorgalvao: “any” seems unnecessary.

lib/hbc/cli/fetch.rb
```diff
@@ -14,6 +14,6 @@ def self.run(*args)
   end
 
   def self.help
-    "downloads Cask resources to local cache"
+    "downloads Cask application files to local cache"
```
vitorgalvao: “downloads remote application files to local cache”

doc/src/brew-cask.1.md
Amorymeltzer: Meant it in an "everything is files" sort of way; in the end, `uninstall` is just removing files...

lib/hbc/cli/fetch.rb
Amorymeltzer: :+1:

doc/src/brew-cask.1.md
vitorgalvao: But in that case, “files and resources” (your change in `CASK_LANGUAGE_REFERENCE.md`) stops making sense. I usually prefer “paths” to mean both files and directories, but I’m fine with just “files”, as long as we’re consistent.

Amorymeltzer: Okay I see your point there, that's good to me.

PR #16398, opened 2016-01-06
Add EditReady 1.3.8

NaanProphet:
Add EditReady 1.3.8

Added cask for EditReady, a video editing app for transcoding
inter-frame delivery/prosumer codec footage (AVCHD, H.264, etc.) into
intra-frame editing formats (e.g. ProRes).

Note: version specific URL deduced. The original is simply
http://www.divergentmedia.com/filedownload/editready

### DIFF
--- a/Casks/editready.rb
+++ b/Casks/editready.rb
@@ -1,0 +1,17 @@
+cask 'editready' do
+  version '1.3.8'
+  sha256 '74eda021699b29c7922ed1ddfa61556e2433df271fce1c7bc619a69b99a0d5db'
+
+  url "http://www.divergentmedia.com/filedownload/editready%20#{version}.dmg"
+  name 'EditReady'
+  homepage 'http://www.divergentmedia.com/editready'
+  license :commercial
+
+  app 'EditReady.app'
+
+  zap :delete =>
+                 [
+                   '~/Library/Application Support/EditReady',
+                   '~/Library/Preferences/com.divergentmedia.EditReady.plist',
+                 ]
+end


Casks/editready.rb
```diff
@@ -0,0 +1,17 @@
...
+  sha256 '74eda021699b29c7922ed1ddfa61556e2433df271fce1c7bc619a69b99a0d5db'
+
+  url "http://www.divergentmedia.com/filedownload/editready%20#{version}.dmg"
+  name 'EditReady'
+  homepage 'http://www.divergentmedia.com/editready'
+  license :commercial
+
+  app 'EditReady.app'
+
+  zap :delete =>
```
vitorgalvao: @jawshooah The `[` is on the wrong line, and yet, it passes. That shouldn’t be the case, should it?

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 56754239f10eed30577de389dce43941d00337e1. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

Casks/editready.rb
jawshooah: HOMEBREW_NO_EMOII

jawshooah: Probably not, but we don't have a cop for that right now. What would the rule be? First stanza argument must always be on the same line (except for block stanzas)?

SirDap:
Thanks! i'll also make a note to commit with the `[` on the same line as the `zap` command next time.

PR #16401, opened 2016-01-06
mcs783x.rb: use our own version methods for interpolation

jawshooah:


jawshooah:
Merged as 4cc511c995acff945506ae76063a572e0152b61f.

PR #16402, opened 2016-01-06
Updated BetterTouchTool Cask with new release location

t3dotgg:
BetterTouchTool's releases are now at a new address. This only applies for versions 1.5 and greater, which have the same compatibility as the 1.49 release that 1.5 replaces

### DIFF
--- a/Casks/bettertouchtool.rb
+++ b/Casks/bettertouchtool.rb
@@ -6,10 +6,10 @@ cask 'bettertouchtool' do
     # bettertouchtool.com is the official download host per the vendor homepage
     url "http://bettertouchtool.net/btt#{version}.zip"
   else
-    version '1.49'
+    version '1.5'
     sha256 'a0ea79136322c634545c0b9da6001944fad5ec734b891fbc63c5485ba0cfd516'
 
-    url "http://boastr.net/releases/btt#{version}.zip"
+    url "https://bettertouchtool.com/releases/BetterTouchTool.zip"
     appcast 'http://appcast.boastr.net',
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
You'll need to update the `sha256`, I'm getting `f6408c0ba48588cd547bef4abdcfd723af5133af141f9f7a146d9ee7721f86e6`.  Also, from the appcast this appears to be version `1.5b`, although it appears previous versions may not be kept online; this may need to move to `:latest`?

@jawshooah Did travis run `brew cask audit`?  I don't think I see it and this should've died on the `--download`.

TheoBr:
My mistake, pushed the wrong commit. Will fix when I get home

jawshooah:
@Amorymeltzer because RuboCop failed the build (unnecessary double-quotes) before we got to the audit.

Amorymeltzer:
@jawshooah Ahh right, duh.  I think I thought the recent configuration put the cask-breaking checks would first, dunno why.

jawshooah:
That's probably a good idea, though. PR incoming.

TheoBr:
@Amorymeltzer 's conjectures about older versions is true, they are no longer kept online and using :latest for this Cask makes more sense

vitorgalvao:
Actually, since this has an appcast, it makes no sense to have it as `:latest`. We still prefer versioned urls since we can `shasum` them, and the appcast will make updating easier in the future, and that’s no longer a “someday down the line, who knows when”, since we now have [a plan](https://github.com/caskroom/homebrew-cask/issues/15908) (see the script/bot section, I plan to start making something soon).

[PR for update](https://github.com/caskroom/homebrew-cask/pull/16405).

PR #16405, opened 2016-01-06
updated bettertouchtool (1.5b)

vitorgalvao:


TheoBr:
Is it preferred to use the official BetterTouchTool.com download source, as it is https?

vitorgalvao:
We prefer it to be versioned (there’s still some modicum of security in checking its checksum). But more importantly, it is available from the appcast.

vitorgalvao:
Thank you for the contribution. It was merged directly as d989a85ca4077ec99f546db62c502771a9d952a2 to keep commit history cleaner. Your contribution is still credited to you.

TheoBr:
Thank you for the clarification @vitorgalvao

PR #16412, opened 2016-01-06
updated acorn (5.2.1)

vitorgalvao:
Yes, the url is versionless, but it has an appcast, so we can still know when it is updated.

### DIFF
--- a/Casks/acorn.rb
+++ b/Casks/acorn.rb
@@ -1,13 +1,15 @@
 cask 'acorn' do
-  version :latest
-  sha256 :no_check
+  version '5.2.1'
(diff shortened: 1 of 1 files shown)


adityadalal924:
Defining a version even though there's none in the url?

vitorgalvao:
> Defining a version even though there's none in the url?

[Top comment](https://github.com/caskroom/homebrew-cask/pull/16412#issue-125250912) (edited right after submission, which is why you must’ve missed it).

adityadalal924:
Yep, commented before the edit. Makes sense, especially with `auto_updates`

adityadalal924:
Merged as 313d4e6ae893394df3203577957b86231c75842e.

PR #16416, opened 2016-01-06
gdal-framework.rb: fix no_check comment

vitorgalvao:
Might seem trivial, but these comments following a pattern is important (will be soon in `cask-repair`, for example).


adityadalal924:
Merged as d7587310fd8cf5c724dce0937a13f9a9b4870cd6.

PR #16431, opened 2016-01-06
updated breeze (1.5.11)

vitorgalvao:


adityadalal924:
Merged as e1a0eee733c51b499374fe47ed02dd75e20846be.

Casks/breeze.rb
```diff
@@ -3,8 +3,6 @@
   sha256 'eaa54f9ccc6539fb548ffa274a7317913840ee3144ed781acd4da8b0b2d0d5ba'
 
   url "https://autumnapps.com/downloads/Breeze#{version}.dmg"
-  appcast 'https://www.autumnapps.com/updates/profiles/breeze.php',
```
tsparber: Maybe the mysql error is only temporary...
(So removing the appcast might loose the ability for automatic checks)

vitorgalvao: It has been there for a while (either that or I always bump against it when it is down).

tsparber: Ok :smiley:

PR #16455, opened 2016-01-07
Delete dockmod.rb

vitorgalvao:
Cask is completely broken.

URLs don’t work, appcasts don’t work. App has switched installation format. Cask seems abandoned and is likely not in use. I propose we remove it and if there’s interest in it, it’ll be added with fixes (as it is, it is useless, anyway).


adityadalal924:
Merged as 06466e596e1d96e7ac70bb2a22a427f5eea6b485.

PR #16460, opened 2016-01-07
updated dupeguru-pe (2.10.1)

vitorgalvao:


adityadalal924:
@jawshooah Looks like audit_modified_casks might be using the wrong commit range here - it's testing more than the Cask modified?

For now, looking at the Travis log manually and merging is the specific Cask is OK.

adityadalal924:
Merged as 11bef6bb9d83998a657025921e520b534cb2c233.

PR #16490, opened 2016-01-07
bbc-iplayer-downloads.rb: delete appcast (changes on each request)

vitorgalvao:
This `appcast` is useless to us. It changes (an `<epoch>` value) on every request, which means `:sha256` will never be correct, and we’ll always get false positives.


adityadalal924:
Merged as 804705e428f3655111e6ce07851d393674d1199b.

adityadalal924:
Merged as a59c494b6f115b2c2f6664b208040c8bec07ca9b.

PR #16503, opened 2016-01-07
Delete fabric.rb

vitorgalvao:
Missing the `official download host per the vendor homepage` comment, and seems to be behind a registration wall.


vitorgalvao:
Merged as c9e89036c0b552e02ab6b17473c69160b23d77f4.

PR #16507, opened 2016-01-07
updated Postgres.app (9.5.0.0)

vcabbage:


Casks/postgres.rb
```diff
@@ -1,6 +1,6 @@
 cask 'postgres' do
-  version '9.4.5.0'
-  sha256 'ce3018430d7a783b2211c153aa9d7ab6197a39356668b516b9086fd0b5084e6a'
+  version '9.5.0.0'
+  sha256 '2c8abeeac2a2569d70b394afe76aca4714bc021aa6ac2af2b58cd9ac20db4eb1'
 
   # github.com is the official download host per the vendor homepage
   url "https://github.com/PostgresApp/PostgresApp/releases/download/#{version}/Postgres-#{version}.zip"
```
vitorgalvao: Updating this value is desired. You can typically find it with `curl --silent "{{appcast_url_here}}" | shasum --algorithm 256 | awk '{ print $1 }'`.

You can easily install it with `brew install vitorgalvao/tinyscripts/caskrepair`. `cask-repair --help` will show you some options, but as a whole the script will guide you through the update process.

vitorgalvao:
Updating the `:sha256` value in `appcast` is now required. You can typically find it with `curl --silent "{{appcast_url_here}}" | shasum --algorithm 256 | awk '{ print $1 }'`.

You can easily install it with `brew install vitorgalvao/tinyscripts/caskrepair`. `cask-repair --help` will show you some options, but as a whole the script will guide you through the update process.

PR #16526, opened 2016-01-07
Updated MacVim

jonashaag:


vitorgalvao:
Updating the `:sha256` value in `appcast` is now required. You can typically find it with `curl --silent "{{appcast_url_here}}" | shasum --algorithm 256 | awk '{ print $1 }'`.

You can easily install it with `brew install vitorgalvao/tinyscripts/caskrepair`. `cask-repair --help` will show you some options, but as a whole the script will guide you through the update process.

adityadalal924:
Closing in favor of https://github.com/caskroom/homebrew-cask/pull/16561/ - version interpolation changed to use with our new helpers.

Thanks for the contribution.

PR #16541, opened 2016-01-07
updated inssider (4.2.2)

vitorgalvao:


adityadalal924:
This probably needs to be shasum :no_check

vitorgalvao:
> This probably needs to be shasum :no_check

It’s another of those cases. Since we have an `appcast` (and we’ll automate checking those), we can actually keep a checksum here and reap its benefits without downsides (unless the user tries to install it in the exact interval between upstream upgrading the app and us updating the cask.

adityadalal924:
Merged as 2d02a1f69a842708823d75792bdaff976370ca09.

PR #16558, opened 2016-01-08
Tell the user how much disk space was freed after `cleanup`

jawshooah:
Also do a bit of refactoring.

### DIFF
--- a/lib/hbc/utils.rb
+++ b/lib/hbc/utils.rb
@@ -237,4 +237,39 @@ module Hbc::Utils
       container_path.join(timestamp)
     end
   end
+
+  def self.size_in_bytes(files)
+    Array(files).reduce(0) { |t, f| t + (File.size?(f) || 0) }
+  end
+
+  # originally from Homebrew
+  def self.disk_usage_readable(size_in_bytes)
+    if size_in_bytes >= 1_073_741_824
+      size = size_in_bytes.to_f / 1_073_741_824
+      unit = "G"
+    elsif size_in_bytes >= 1_048_576
+      size = size_in_bytes.to_f / 1_048_576
+      unit = "M"
+    elsif size_in_bytes >= 1_024
+      size = size_in_bytes.to_f / 1_024
+      unit = "K"
+    else
+      size = size_in_bytes
+      unit = "B"
+    end
+
+    # avoid trailing zero after decimal point
+    if (size * 10).to_i % 10 == 0
+      "#{size.to_i}#{unit}"
+    else
+      "#{"%.1f" % size}#{unit}"
+    end
+  end
+
+  # originally from Homebrew
+  def self.number_readable(number)
+    numstr = number.to_i.to_s
+    (numstr.size - 3).step(1, -3) { |i| numstr.insert(i, ",") }
+    numstr
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/hbc/utils.rb
```diff
@@ -237,4 +237,13 @@ def self.nowstamp_metadata_path(container_path)
...
     end
   end
+
+  def self.space_in_megs(files)
+    bytes = Array(files).reduce(0) { |t, f| t + (File.size?(f) || 0) }
+    bytes / (1024.0 * 1024.0)
+  end
+
+  def self.pretty_space_in_megs(files)
+    sprintf('%0.2f MB', space_in_megs(files))
```
tsparber: Hmm, feels wrong to use MB and a base 2. OS X in general (since [10.6](https://support.apple.com/en-us/HT201402)) uses a base 10 system for MB (and base 2 would be [MiB](https://en.wikipedia.org/wiki/Mebibyte)).
Homebrew uses only [M](https://github.com/Homebrew/homebrew/blob/fb3bda683361b608dcfa62f4a3939d32e7420fe4/Library/Homebrew/utils.rb#L614) and a base 2.

jawshooah: Ah, nice catch, thanks! I always assumed MB meant 2<sup>20</sup> bytes. That means the original logic was broken as well, assuming "megs" == MB.

@vitorgalvao, what do you think? Should we consistently display the cache size in MB (or MiB), or change the units depending on the size?

vitorgalvao: Though I don’t have a too strong opinion on the matter, I’m more inclined towards changing units depending on the size, it seems more practical/natural<sup>1</sup>.

If users wanted to use the result as part of a script, consistency would be important, but that is extremely unlikely. Being this a “nice to have”, having what’s more immediate seems to make more sense.

---

<sup>1</sup> When using [`du`](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/du.1.html), I always use the `-h` (`"Human-readable" output`, their words).

tsparber: MiB is not really used on OS X, I would stick with the same convention that homebrew uses. There was no discussion about it (https://github.com/Homebrew/homebrew/pull/45642), but I still like their use of only M (atleast it does not use MB with base 2 :wink:).
Would it possible to use their [`disk_usage_readable`](https://github.com/Homebrew/homebrew/blob/fb3bda683361b608dcfa62f4a3939d32e7420fe4/Library/Homebrew/utils.rb#L614) directly?

jawshooah: @tsparber we do plan to integrate more closely with Homebrew in the future, but for now we're trying to keep our library independent from theirs, bringing in helpers as they are needed.

Amorymeltzer:
Just getting around to looking at this, thanks! :trophy:   Was just noticing brew had this but hbc didn't.  Is there anyway to only output if something is cleaned up?  As of now, `cask cleanup` with nothing to cleanup outputs 5 very lines, three of which are essentially `you had nothing to cleanup`.  This new line looks especially silly after two lines of nada.

jawshooah:
> As of now, cask cleanup with nothing to cleanup outputs 5 very lines, three of which are essentially you had nothing to cleanup.

Nice catch, and a very quick fix. PR incoming.

PR #16562, opened 2016-01-08
Update .travis.yml

adidalal:
Add `fast_finish: true` (from https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing)

Pinging @jawshooah - This may help speed up failures - it seems it was added once, long ago, but then removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ matrix:
       os: osx
       osx_image: beta-xcode6.2
       rvm: system
+  fast_finish: true
 
 cache:
   directories:


jawshooah:
> it seems it was added once, long ago, but then removed.

Yeah, that was intentional. I was concerned that if we fail early in one OS environment, we lose information about whether or not we would fail in the other environments.

Given how long our builds typically take, though, I'm willing to sacrifice that information or the sake of speeding things up. If we encounter a failure that seems related to the build environment, we can just disable `fast_finish` until we get it sorted out.

PR #16563, opened 2016-01-08
Add Adguard.app v1.1.0

miccal:


Casks/adguard.rb
```diff
@@ -0,0 +1,11 @@
+cask 'adguard' do
+  version '1.1.0'
+  sha256 '4d48a7c048e36d002b371e22765e68fef31dd011db18635bef35063e456c27f8'
+
+  url 'https://static.adguard.com/mac/Adguard.release.dmg'
+  name 'Adguard for Mac'
+  homepage 'https://adguard.com/'
+  license :closed
```
vitorgalvao: `:gratis`.

vitorgalvao:

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 72f4876feb179db16f27c7362d11c811e35268ed. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

miccal:
Thanks for that. When I initially created the cask I set it up as you suggested, `<version: latest>` and `<sha256 :no_check>`, but then received a sha check error when I tested it - I am not really sure why?

Also, I have realised that this is a 14 day free trial and requires a licence after, so should the licence then be `<commercial>`?

vitorgalvao:
> When I initially created the cask I set it up as you suggested, `<version: latest>` and `<sha256 :no_check>`, but then received a sha check error when I tested it - I am not really sure why?

I’d need to know the exact steps and the exact error to answer.

> Also, I have realised that this is a 14 day free trial and requires a licence after, so should the licence then be `<commercial>`?

Yes. Fixing.

PR #16565, opened 2016-01-08
QQBrowser update to version 3.9.3952.400

ghost:
QQBrowser update to version 3.9.3952.400, and the download url has changed to  "http://dldir1.qq.com/invc/tt/QQBrowser_for_Mac.dmg"

### DIFF
--- a/Casks/qqbrowser.rb
+++ b/Casks/qqbrowser.rb
@@ -1,8 +1,8 @@
 cask 'qqbrowser' do
-  version '3.9.3943.400'
-  sha256 '4770ad6d40db55c765a8937424b228778e53c960e41db1e6112b788557ed52e1'
+  version '3.9.3952.400'
+  sha256 '24ee717d4f74cdf3239d0f9898749ff0ee1fa2004707f698c2d86ddcac82c819'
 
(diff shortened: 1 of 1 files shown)


Casks/qqbrowser.rb
```diff
@@ -1,8 +1,8 @@
 cask 'qqbrowser' do
-  version '3.9.3943.400'
-  sha256 '4770ad6d40db55c765a8937424b228778e53c960e41db1e6112b788557ed52e1'
+  version '3.9.3952.400'
+  sha256 '24ee717d4f74cdf3239d0f9898749ff0ee1fa2004707f698c2d86ddcac82c819'
 
-  url "http://dldir1.qq.com/invc/tt/QQBrowser.#{version.sub(%r{^.*\.}, '')}.dmg"
+  url "http://dldir1.qq.com/invc/tt/QQBrowser_for_Mac.dmg"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

vitorgalvao:

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 186fc5ae0a317f278e9443db23c8bc9ad1446a71. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #16587, opened 2016-01-08
Delete mail-plugin-manager.rb

vitorgalvao:
Looks like the company was bought. Even though the link still works, there’s no page to the app to be found.


adityadalal924:
All the `delete`s will cause Travis to fail - the commits are okay though.

adityadalal924:
Merged as 8f732ea164a9b4f903c5912b8bd55367df9bd72a.

vitorgalvao:
> All the `delete`s will cause Travis to fail - the commits are okay though.

Ah, alright ([mentioned that just now](https://github.com/caskroom/homebrew-cask/pull/16589#issuecomment-170092181)). Related to the recent `rubocop` commits and whatnot?

PR #16589, opened 2016-01-08
mailplane.rb: removed appcast

vitorgalvao:
Outdated over one major version.

### DIFF
--- a/Casks/mailplane.rb
+++ b/Casks/mailplane.rb
@@ -3,8 +3,6 @@ cask 'mailplane' do
   sha256 :no_check
 
   url 'http://update.mailplaneapp.com/mailplane_3.php'
-  appcast 'http://update.mailplaneapp.com/appcast.php?appName=Mailplane',
-          :sha256 => 'adae9ab9b89b12ce661a2475ff8a9d7b8220bb228e01dc38727ff28e2cd855e9'
   name 'Mailplane'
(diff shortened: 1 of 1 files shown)


adityadalal924:
Thank you for the contribution. It needed some fixes, so they were made in e253cd5f6ff5e8ed530fcb2e729283b0f066a8a5. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

vitorgalvao:
@jawshooah Travis failed multiple times on this PR (and siblings). It seems like it failed on any PR that removed parts of a cask (specifically `appcast` and `:sha256`) or the cask as a whole. Does that make sense?

jawshooah:
@vitorgalvao Notice the PR build succeeded, but the push (branch) build failed.

Travis set `TRAVIS_COMMIT_RANGE` to `92365edf8023...6809a0ecd31a`. Looking at my local history, 92365edf8023 and 6809a0ecd31a appear to be the exact same commit, just based on different parents, likely due to a rebase.

Branch builds track the last known position of the branch ref, which includes rebases and force-pushes. There's nothing wrong with the build script; the problem is that we shouldn't be building branches other than `master` at all.

jawshooah:
Fixed in 8db87affce12cf44f5e65a950ec692ee3d4b3219.

PR #16654, opened 2016-01-09
Delete youtrack-workflow.rb

adidalal:
Latest version does not have any `.app` bundle at all - see https://confluence.jetbrains.com/display/YTD65/Installing+YouTrack+with+ZIP+Distribution

It needs to be manually configured with `./youtrack.sh start`


vitorgalvao:
Merged as f7e46ae18741c95105b29f1eff2eb61e91d6aa0e.

PR #16667, opened 2016-01-09
Update cleanmymac versions and strings

Amorymeltzer:
These commits[1]
1. Update versions, add a sha, and add an appcast+sha
2. Use the new string methods

Assuming I got the latter correct, this should be straightforward.

---

This [cask](https://github.com/caskroom/homebrew-cask/blob/77b529d0011fcc687852b04a5c4bb1c1edf10433/Casks/cleanmymac.rb), though, gave me a headache as I have been trying to square away some recent decisions, namely what belongs in caskroom/homebrew-versions[3] and the use of discontinued[3].  Basically, I'm not sure what those decisions mean for our use of `MacOS.release` in the main repo.
- CleanMyMac Classic (version 1) is old, legacy, no longer under development, and should be moved to caskroom/homebrew-versions and caveat-ed
- CleanMyMac2 was last updated alongside version 3 in late December, to be El Capitan-able, which means it's an "alternate" version and should go to caskroom/homebrew-versions
  - Is version 3 more similar to a "pro-version" and should be moved while version 2 stays in the same repo?  No, but both should be available to a user on 10.8-10.11 since they are different products.
- Either way, version 1 and 2 don't have their own homepages (even though 2 appears to be actively developed), and would need to exist under the same cask as version 3?
  - But version 2 and version 3 are different products.
- Does the lack of a homepage or the lack of updates signify `discontinued`, or is a note from the developer required?

Broadly speaking, if something has a different `#{version.major}` for a below-Yosemite `MacOS.release`, does it belong in the main repo?  Is it not, by definition, a "legacy" version?  Do we keep it here to avoid multiple casks, or is it dishonest to offer users a version of software that is discontinued without saying so?  I can't figure it out.

Pinging @vitorgalvao for obvious reasons.

---

1: Kept them separate since they are different tasks, but also to make it easy in case the string methods didn't work (difficult to check old operating systems); happy to squash once travis is happy if preferred.
2: See caskroom/homebrew-versions#1450 and caskroom/homebrew-versions#1504
3: See #16097 and #16225

### DIFF
--- a/Casks/cleanmymac.rb
+++ b/Casks/cleanmymac.rb
@@ -10,11 +10,12 @@ cask 'cleanmymac' do
     app 'CleanMyMac.app'
     # TODO: add uninstall and zap stanzas for legacy app
   elsif MacOS.release <= :lion
-    version '2.3.5-1427986644'
-    sha256 '16e192edcf58f25c6763349ef0e5194268bec4d000912b64b34f5897b4784097'
+    version '2.4-1443544143'
+    sha256 '0d08f4d9b36493359f6ca3ff2f96a9b769a8eed4ea017ecbb2d5644f75aafad0'
     # devmate.com is the official download host per the appcast feed
     url "https://dl.devmate.com/com.macpaw.CleanMyMac2/#{version.sub(%r{-.*$}, '')}/#{version.sub(%r{.*?-}, '')}/CleanMyMac#{version.to_i}-#{version.sub(%r{-.*$}, '')}.zip"
-    appcast "http://updates.devmate.com/com.macpaw.CleanMyMac#{version.to_i}.xml"
+    appcast "http://updates.devmate.com/com.macpaw.CleanMyMac#{version.to_i}.xml",
+            :sha256 => '612be3e443f49b6b1d5539611693714b2897b9b774896ac15c048daec76a34cb'
     app "CleanMyMac #{version.to_i}.app"
 
     uninstall :launchctl => "com.macpaw.CleanMyMac#{version.to_i}.Agent"
@@ -32,11 +33,13 @@ cask 'cleanmymac' do
                      "~/Library/Preferences/com.macpaw.CleanMyMac#{version.to_i}.plist",
                    ]
   else
-    version '3.2.1-1448556223'
-    sha256 'a5e7587d7edcece90c0d5b0b6aed8d008b6a1a93cb6911f0cf2e9e0c0453e6c4'
+    version '3.3.0-1451475053'
+    sha256 '14038d3554f771e32840dbd0936235d62b95713638b4b1ca94daf32847da389f'
 
     # devmate.com is the official download host per the appcast feed
     url "https://dl.devmate.com/com.macpaw.CleanMyMac#{version.to_i}/#{version.sub(%r{-.*$}, '')}/#{version.sub(%r{.*?-}, '')}/CleanMyMac3-#{version.sub(%r{-.*$}, '')}.zip"
+    appcast "http://updates.devmate.com/com.macpaw.CleanMyMac#{version.to_i}.xml",
+            :sha256 => 'e62a4026c55f52a3c31a0070e8fb9d49c63244697dff9332a3ab37f74d02c5aa'
     app "CleanMyMac #{version.to_i}.app"
 
     postflight do
--- a/Casks/cleanmymac.rb
+++ b/Casks/cleanmymac.rb
@@ -1,66 +1,67 @@
 cask 'cleanmymac' do
   if MacOS.release <= :snow_leopard
-    version '1.11-1417522595'
+    version '1.11,1417522595'
     sha256 'ac5d4bf36882dd34bdb0a68eb384a6b3aba355be896d03dfa40a120c6bef4a0d'
 
     # devmate.com is the official download host per the appcast feed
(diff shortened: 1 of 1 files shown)


vitorgalvao:
I’ll start by saying I think anything that is `if MacOS.release <= :snow_leopard` should be removed from every cask (including the ones in caskroom/versions). Same for everything that is 32-bit or PowerPC. Heck, I even think support for those should be removed from the core<sup>1</sup>. Simply put, those are old and I don’t think they’re even in use. They’re overhead for nothing.

I’ll also set an important point: many of the current cases are wrong according to our rules, because they existed before the rules were in place. Hence, they should be corrected.

On to this specific case.

CleanMyMac2 existed (before the recent cleanse) in caskroom/versions because it’s a commercial app (and you may not *want* to use/buy the latest version). However, it also existed here because you might be running an old OS (so you may not be *able* to run the latest version). That’s the difference. The former says “I want this particular version”, while the latter says “I want the latest version I can run”. Old versions exist in caskroom/versions so you can *choose* to use them, but inside conditionals in the main repo for when you’re *forced* to use them. That is the important distinction.

You cannot buy version 2 today (from the official vendor, at least). This is not a case of a standard and pro or regular and eap where the company says “pick whichever you prefer”. They are unapologetically saying “pick this one, it’s superior and the future”. The fact they made an update to an older version (essentially a bug fix, if all it consisted of was support for a later OS) doesn’t change that fact. They still see (and present) version 3 as a *continuation* of version 2, not as an alternative. This is also why not having a specific homepage for version 2 in this cask isn’t a big deal, it’s still the same app, after all.

As for `discontinued`, think of it as always applying to the cask as a whole. It might help to [look at the actual outputted message](https://github.com/caskroom/homebrew-cask/blob/59b77f1b80829432e8f986d07d28f2d11dd05500/lib/hbc/caveats.rb#L122L123). It says the cask (not a version of it) has been officially discontinued, and “it may stop working correctly (or at all) in recent versions of OS X”. “May”. If it were inside a conditional, you’d know for sure (that’s the whole point). Also remember another important point: `caveats` aren’t merely informative and must always be relevant to the installation of the app. `discontinued` doesn’t tell you a cask has been discontinued as a curiosity, but specifically so you know why the heck it isn’t working when you open it (in case it doesn’t).

It can be tangentially informative in a very specific case, however. After we know a discontinued cask no longer works in a certain OS version, we could `depends_on macos => '<= :whatever'`, but doing *only* that might still make a user think “it doesn’t work on this version *yet*”. `discontinued` signals “go look for an alternative instead”, and has the added bonus of marking those for us maintainers, for future removals.

I hope everything is now clear. If not, please fire away.

---

<sup>1</sup> I simply haven’t yet opened issues with those proposals.

Amorymeltzer:
:100: I think a lot of this is significantly reducing the scope of hbc, but that's not inherently bad.  I really do appreciate the steps you're going through to make the future actually happen.

Just to clarify one of your excellent, per usual, points, are you saying `cleanmymac2` should go in `versions`?  I think you're saying that, not sure.

And, this is more for myself when the issue arises, but I think `snow_leopard` and `lion` need to have their fates tied:  SL is probably used more than [either lion or ML](https://www.intego.com/mac-security-blog/os-x-market-share-statistics-1-in-5-macs-still-unsupported/), is possibly behind only [El-Capitan and Yosemite in usage](http://lowendmac.com/2015/the-rise-and-fall-of-mac-os-x-versions-2009-to-2015/), and is likely [here to stay](http://www.computerworld.com/article/2476004/mac-os-x/why-mac-users-still-use-os-x-snow-leopard.html).

vitorgalvao:
> I think a lot of this is significantly reducing the scope of hbc, but that's not inherently bad.

I think so as well. I’ve always thought an issue with this project was how quickly it grew and how much excitement there was. Everything was said yes to, and that left some scars. I firmly believe stepping back, even if a bit much at first, will allow us to not only make better decisions, but build a better foundation for what we support, including future additions.

Tangentially, I’m pro-tap and do think those should be encouraged. Not being tied to the main repos while still reaping the benefits of functionality is a great feature, [and in some cases](https://github.com/caskroom/homebrew-cask/issues/16687) it makes more sense than official support, anyway.

> I really do appreciate the steps you're going through to make the future actually happen (…) to clarify one of your excellent, per usual, points

Thank you. I truly value that feedback and the kind words. It is rewarding to see something come out of all the work. Your continuous questions do also help in making solid decisions. Even when I immediately know the answer to something asked, having to clarify it in writing does help cement the reasoning.

> are you saying `cleanmymac2` should go in `versions`? I think you're saying that, not sure.

Yes, I am saying it belongs there. However, I’m not saying it should be added as of now. It was there previously, [and rightly so](https://github.com/caskroom/homebrew-versions/pull/1235), but I’ve decided to remove it [in a recent purge](https://github.com/caskroom/homebrew-versions/pull/1542) in an effort to bring that repo to a manageable state before starting to add stuff again. Reasoning being, if someone needs it they’ll readd it and we can start applying the rules from there.

> And, this is more for myself when the issue arises, but I think `snow_leopard` and `lion` need to have their fates tied:  SL is probably used more than [either lion or ML](https://www.intego.com/mac-security-blog/os-x-market-share-statistics-1-in-5-macs-still-unsupported/), is possibly behind only [El-Capitan and Yosemite in usage](http://lowendmac.com/2015/the-rise-and-fall-of-mac-os-x-versions-2009-to-2015/), and is likely [here to stay](http://www.computerworld.com/article/2476004/mac-os-x/why-mac-users-still-use-os-x-snow-leopard.html).

Good enough for me. If we’re keeping it, it’s good to have that kind of data. Thank you.

PR #16683, opened 2016-01-10
Add EiskaltDC++.app v2.3.0 Built on Qt 5.5.0

katoquro:
based on v2.2.10
alternate version based on newest Qt framework

I didn't get how should appcast stanza work and how can I evaluate it's sha256 so I've skipped this stanza

### DIFF
--- a/Casks/eiskaltdcpp-qt5.rb
+++ b/Casks/eiskaltdcpp-qt5.rb
@@ -1,0 +1,13 @@
+cask 'eiskaltdcpp-qt5' do
+  #EiskaltDC++ from git master for Mac OS X 10.9 64-bit
+  #Built on Mac OS X 10.9 with Qt 5.5.0
+  version '2.3.0-v2.2.10-62-ge42c04e'
+  sha256 '1463fd447ffd16d9f3af96b0d2306a728d580e59bc7b9a45fcbe895c99c5432c'
+
+  url "http://downloads.sourceforge.net/sourceforge/eiskaltdcpp/EiskaltDC++-#{version}-x86_64-qt5.dmg"
+  name 'EiskaltDC++'
+  homepage 'http://sourceforge.net/projects/eiskaltdcpp/'
+  license :gpl
+
+  app 'EiskaltDC++.app'
+end
--- a/Casks/eiskaltdcpp-qt5.rb
+++ b/Casks/eiskaltdcpp-qt5.rb
@@ -1,6 +1,6 @@
 cask 'eiskaltdcpp-qt5' do
-  #EiskaltDC++ from git master for Mac OS X 10.9 64-bit
-  #Built on Mac OS X 10.9 with Qt 5.5.0
+  # EiskaltDC++ from git master for Mac OS X 10.9 64-bit
+  # Built on Mac OS X 10.9 with Qt 5.5.0
   version '2.3.0-v2.2.10-62-ge42c04e'
   sha256 '1463fd447ffd16d9f3af96b0d2306a728d580e59bc7b9a45fcbe895c99c5432c'
 


Casks/eiskaltdcpp-qt5.rb
```diff
@@ -0,0 +1,13 @@
+cask 'eiskaltdcpp-qt5' do
+  # EiskaltDC++ from git master for Mac OS X 10.9 64-bit
```
vitorgalvao: Please remove the comments.

vitorgalvao:
Is there an appcast for this? I.e. a file online that it uses to check for updates? If, so, post the link and we’ll help you make the stanza.

Either way, this belongs in [caskroom/versions](https://github.com/caskroom/homebrew-versions).

vitorgalvao:
What exactly are the differences between this and [`eiskaltdcpp`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/eiskaltdcpp.rb)? One uses `qt` and the other not? This seems to be a newer version.

Casks/eiskaltdcpp-qt5.rb
katoquro: So about comment. Is there more convenient way to provide additional info about package? 
The output of `cask info` is quite poor. F.e. how can I understand the difference between `eiskaltdcpp` and `eiskaltdcpp-qt5` packages via `cask info output` (or some thing like `apt-cache search` result page)

katoquro:
> I.e. a file online that it uses to check for updates?

there is a usual sourceforge rss for project. I think it can be used as appcast http://sourceforge.net/projects/eiskaltdcpp/rss?path=/Mac%20OS%20X

>What exactly are the differences between this and eiskaltdcpp? One uses qt and the other not? This seems to be a newer version.

I't s based on the same tag version. But it contains small changes (because of moving to github infrastructure I think) and new major version of Qt, there is a version with Qt4 (this one `EiskaltDC++-2.3.0-v2.2.10-62-ge42c04e-x86_64-qt4.dmg` is tight candidate for `eiskaltdcpp` formuls)

katoquro:
After closing PR new commits don't be added. Should I create new one with additional new commit?

vitorgalvao:
> there is a usual sourceforge rss for project. I think it can be used as appcast http://sourceforge.net/projects/eiskaltdcpp/rss?path=/Mac%20OS%20X

In theory it could, yes. However, it unfortunately seems sourceforge casks won’t do.

This is how this one could be:

```ruby
appcast 'http://sourceforge.net/projects/eiskaltdcpp/rss?path=/Mac%20OS%20X',
        :sha256 => 'c8f46da22da07f8ec68159b6f9a5a996e89669dcaf8dd0e5cdc3a2461506e4fe'
```

To get the `:sha256` value, simply `curl 'http://sourceforge.net/projects/eiskaltdcpp/rss?path=/Mac%20OS%20X' | shasum -a 256`. However, sourceforge appcast have a `<pubDate>` value that changes depending on when the request was made, so we can’t actually track them.

> After closing PR new commits don't be added. Should I create new one with additional new commit?

Looking further into it, it seems the version we now have is old compared to the one you submitted. Submit a new PR, yes, but instead of adding a new cask, simply submit that new url to the existing `eiskaltdcpp` cask.

katoquro:
>Submit a new PR, yes, but instead of adding a new cask, simply submit that new url to the existing eiskaltdcpp cask.

You may miss my comment about two build versions, default - qt4 and my cask with qt5

```
File: EiskaltDC++-2.3.0-v2.2.10-62-ge42c04e-x86_64-qt5.dmg
Description: EiskaltDC++ from git master for Mac OS X 10.9 64-bit
Comment: Built on Mac OS X 10.9 with Qt 5.5.0

File: EiskaltDC++-2.3.0-v2.2.10-62-ge42c04e-x86_64-qt4.dmg
Description: EiskaltDC++ from git master for Mac OS X 10.9 64-bit
Comment: Built on Mac OS X 10.9 with Qt 4.8.7
```

Should I leave only one?

vitorgalvao:
Ah, apologies, misinterpreted that. Yes, leave only the `qt-5` one. Is there any reason a user would pick the other one?

PR #16699, opened 2016-01-10
brew-cask: upgrade to 0.60.1

DomT4:
Over at Homebrew we've accidentally caused an edge-case problem for some Caskroom users. We recently merged https://github.com/Homebrew/homebrew/commit/915ac06d which is much stricter on empty installations, failing during install rather than just shouting about it in the audit.

We'd prefer not to revert the change there because it's quite a positive change that's much more obvious about installation problems, given installations shouldn't be empty. There's also the likelihood that if we revert, wait 3 months to add the check back, someone out there will wait 4 months before upgrade and we'll run into this again.

However, this causes some issues with Cask because when Mike rewrote the formula he left `def install` an empty hole, which was fine at the time. Anyone who hasn't updated between December 9th 2015 and January 9th 2016 will now run into a fatal error on upgrading the Cask formula, as seen in: https://github.com/Homebrew/homebrew/issues/47929

I was wondering if the Cask would be kind enough to tag one very last bug fix release, or even retag the existing tag with a similar fix to this. You may want to do this yourself rather than using this PR; this was an easy way to both explain the problem and propose a fix without sticking a `diff` blob in the middle of the text.

Obviously, if tested by Travis here this build will likely fail because there is no 0.60.1 tag. Apologies for the accidental mess here. It's one of those things we'll probably get better at if/when we can bring the two projects closer together and do more unified testing.

CC @mikemcquaid for thoughts.

### DIFF
--- a/brew-cask.rb
+++ b/brew-cask.rb
@@ -1,15 +1,20 @@
 class BrewCask < Formula
   homepage "https://github.com/caskroom/homebrew-cask/"
-  url "https://github.com/caskroom/homebrew-cask.git", :tag => "v0.60.0"
+  url "https://github.com/caskroom/homebrew-cask.git", :tag => "v0.60.1"
 
   depends_on :ruby => "2.0"
 
   def install
+    (buildpath/"UPGRADE").write <<-EOS.undent
+      You must uninstall this formula. It is no longer needed to stay up to date,
+      as Homebrew now takes care of that automatically.
+    EOS
+    prefix.install "UPGRADE"
   end
 
   def caveats
     <<-EOS.undent
-      You can uninstall this formula as `brew tap Caskroom/cask` is now all that's
+      You must uninstall this formula as `brew tap Caskroom/cask` is now all that's
       needed to install Homebrew Cask and keep it up to date.
     EOS
   end


mikemcquaid:
:+1: from me.

adityadalal924:
Pinging @caskroom/maintainers

brew-cask.rb
```diff
@@ -1,10 +1,15 @@
 class BrewCask < Formula
   homepage "https://github.com/caskroom/homebrew-cask/"
-  url "https://github.com/caskroom/homebrew-cask.git", :tag => "v0.60.0"
+  url "https://github.com/caskroom/homebrew-cask.git", :tag => "v0.60.1"
 
   depends_on :ruby => "2.0"
 
   def install
+    (buildpath/"UPGRADE").write <<-EOS.undent
+      You should uninstall this formula. Only `brew tap Caskroom/cask`
```
vitorgalvao: “should” → “must”? We want people to really do it, so I wouldn’t object to the more forceful language.

In addition, “Only `brew tap Caskroom/cask` is required now to stay up-to-date.” could be something like “It is no longer needed to stay up to date, as homebrew takes care of that automatically”. By saying “only `brew tap Caskroom/cask` is required now to stay up-to-date”, I wonder if some people will regularly try that command to update homebrew-cask. Making clear they must uninstall the formula and need do nothing more is likely a plus.

This because we still receive so many bug reports regarding this latest change (despite plastering information about it all over the place), having a tone of “you must uninstall this because it really isn’t needed, and here’s why” will likely be beneficial.

vitorgalvao:
@jawshooah You commented (and I agree) about having a round number for the last release (`0.60.0`) and I concur it makes sense. Shall we make this `0.70.0`?

mikemcquaid:
@vitorgalvao Up to you but given this is just a patch around a Homebrew bug in the interests of semver I'd leave it as a `.1`.

brew-cask.rb
mikemcquaid: :+1: to `must`.

mikemcquaid: But I'd feel free to pull as-is and edit locally ( @DomT4 definitely won't mind) if it lets you get it in quicker and save you work.

DomT4: Yeah. This was more of a suggested fix than a "PR" per se. You can use it if you want to and I'm happy to push changes as desired, but am entirely understanding of the fact it's probably easier for you to push the necessary tweaks on your end and close this PR afterwards and not going to object if you prefer that path.

jawshooah:
I don't care much about the version tag; not like we've been sticking to semver anyway (plenty of breaking changes in the project's history, not a single major version bump). @vitorgalvao, any objection to me pulling this down, tagging and pushing?

DomT4:
Let me know if you want me to make any changes.

vitorgalvao:
> @vitorgalvao, any objection to me pulling this down, tagging and pushing?

No objection at all.

adityadalal924:
Other than the suggested wording changes, it LGTM

DomT4:
Made some wording tweaks.

jawshooah:
I'm going to keep the wording the same between `install` and `caveats` if it's all the same to you guys.

DomT4:
Feel free to make whatever changes you want to to anything I've done here.

jawshooah:
Closed in b4c92f3658d47712aa6114fb29be50a763cf78ee.

DomT4:
Thanks!

PR #16701, opened 2016-01-10
developer scripts: add find_sparkle_appcast

vitorgalvao:


developer/bin/find_sparkle_appcast
```diff
@@ -0,0 +1,77 @@
...
+# get appcast
+appcast_url="$(defaults read "${path_to_plist}" 'SUFeedURL' 2>/dev/null)"
+
+if [[ -z "${appcast_url}" ]]; then
+  echo 'It appears this app does not have a Sparkle appcast.'
+  exit 0
+fi
+
+# vaidate appcast
+appcast_http_response=$(curl --silent --head "${appcast_url}" | grep '^HTTP' | perl -pe 's|.* (\d{3}) .*|\1|')
```
tsparber: There are some servers (mostly cdn so more widley used apps) which do not allow HEAD (only GET) and so fail with non 200. I've encountered this when experimenting with an extened version of `cask-switch-https` (Still problem there, will fix soon). 
I would suggest to not use the HEAD request (but still check the status code).

vitorgalvao: But if we don’t pass `--head`, won’t we be downloading the file? Though they are typically small, that does not seem like the best solution

In addition, although your point is a valid concern, I doubt it’ll actually become a problem. As an effort to bring `appcast`s up to date, I’ve been updating them all and check first for the status code in this manner. Going alphabetically, now on `s`, and have yet to find an issue.

tsparber: Yes, without the `--head` the file would be downloaded. I would also need to find a solution for this for `cask-switch-https`. Maybe there are no appcasts which link to such a server, but there where  some homepages and urls.

* `curl --head --location -v https://developer.android.com/sdk/` Returns 405
* `curl --head --location -v https://install.bankid.com/FileDownloader?fileId=Mac` Returns 500
* `curl --head --location -v https://cronnix.googlecode.com/files/CronniX_3.0.2.app.zip` Returns 404
(Every googlecode download yields a 404 with `--head` but works with GET)
* `curl --head --location -v https://bitbucket.org/blackosx/darwindumper/downloads/DarwinDumper_v2.9.9.2.zip` Redirects to amazonaws with status code 403 when using head.
(All amazonaws links seem not to allow HEAD)

Atleast googlecode (going away) and amazonaws are widley used.

tsparber: This might also be the reason for the workaround in [cask-repair#L235](https://github.com/vitorgalvao/tiny-scripts/blob/master/cask-repair#L235).

vitorgalvao: Sounds like a good candidate for a Stack Overflow question.

tsparber: Build a solution myself :smiley: https://github.com/caskroom/homebrew-cask/pull/16787

PR #16702, opened 2016-01-10
Delete scrup.rb

vitorgalvao:
Dead.


vitorgalvao:
Merged as eb3641132284f3d75c10f0ae0cd42fde4c5ee63f.

Amorymeltzer:
There's a slippery slope to looking for alternate, valid links to a download, but if the link on the readme were changed to point to the GH releases page, this would still be valid, no?

vitorgalvao:
> if the link on the readme were changed to point to the GH releases page, this would still be valid, no?

Yes. I don’t think we’d even need to get that far. The homepage is the github repo, so outright using the releases download would be appropriate. I looked into it before removal, though, and in this case it doesn’t apply because the releases page only contains the source, not a usable binary we can get.

This has been discontinued, though, so it’s unlikely they’d want to add it anyway.

PR #16703, opened 2016-01-10
cask-switch-https: improved getting cask_status

vitorgalvao:
Pinging @tsparber. Just made this change to `cask-repair`.


tsparber:
:yes:

PR #16719, opened 2016-01-11
Add Lima 1.0.0

joaovrb:


vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 6b52c461d9846fc74ad6585d4df42998ee9a717e. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

joaovrb:
Hi, I tried to include the version in the URL via interpolation but I was geting an error, so I reverted to use the explicit version. I see now that I should've used double quotes.

PR #16727, opened 2016-01-11
Upgrade td-agent 2.3.0

n0ts:


adityadalal924:
Fixed issues with https://github.com/caskroom/homebrew-cask/pull/16771 - was missing comment, and url needed to be changed.

PR #16731, opened 2016-01-11
hacking.md: remove mention of separating from homebrew

vitorgalvao:


adityadalal924:
hacking.md should be reformed into "Editing HBC core" or something of the sort.

PR #16780, opened 2016-01-12
Update brew-cask.1

miccal:


vitorgalvao:
Closing since it’ll need to be regenerated after [this fix](https://github.com/caskroom/homebrew-cask/pull/16779/files#r49451979).

PR #16787, opened 2016-01-12
Use curl-check-url to check for HTTP 200 using GET

tsparber:
Some servers do not support HEAD requests, this script uses now the [curl-check-url](https://github.com/tsparber/homebrew-tiny-scripts/blob/master/bin/curl-check-url) script to issue a GET request but abort after the headers have been received.

### DIFF
--- a/developer/bin/cask-switch-https
+++ b/developer/bin/cask-switch-https
@@ -3,7 +3,7 @@
 set -o pipefail
 
 readonly program="$(basename "$0")"
-skip_curl_head=0
+skip_curl_verify=0
 verbose=0
 
 syntax_error() {
@@ -12,11 +12,26 @@ syntax_error() {
   exit 1
 }
 
+depends_on() {
+  formula="$1"
+  [[ "$#" -eq 2 ]] && cmd="$2" || cmd=$(basename "${formula}")
+
+  if [[ ! $(which ${cmd}) ]]; then
+    echo -e "$(tput setaf 1)
+      This script depends on '${cmd}'.
+      If you have [Homebrew](http://brew.sh), you can install it with 'brew install ${formula}'.
+    $(tput sgr0)" | sed -E 's/ {6}//' >&2
+    exit 1
+  fi
+}
+
+depends_on 'tsparber/tiny-scripts/curl-check-url'
+
 usage() {
   echo "
     This script changes the url, appcast and homepage stanzas to https
 
-    After changing to https a http head request is performed to verify if the url is reachable.
+    After changing to https a HTTP GET request is performed to verify if the url is reachable.
     If the https url is not reachable it is reverted to the previous version.
 
     Known Issues: If multiple url/appcast stanzas are present, all urls are changed but only
@@ -26,7 +41,7 @@ usage() {
 
     usage: $program [options] [<cask_name>]
     options:
-      -s, --skip-verify  Skip checking for a HTTP 200 Status Code using curl head.
+      -s, --skip-verify  Skip checking for a HTTP 200 Status Code using curl.
       --verbose          Show more verbose output.
       -h, --help         Show this help.
 
@@ -42,7 +57,7 @@ while [[ "$1" ]]; do
       exit 0
       ;;
     -s | --skip-verify)
-      skip_curl_head=1
+      skip_curl_verify=1
       ;;
     --verbose)
       verbose=1
@@ -65,23 +80,12 @@ check_url_for_https() {
   [[ ${verbose} -ne 0 ]] && verbose_option="-v "
 
   # check if the URL sends a 200 HTTP code, else abort
-  browser_headers="User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/33.0.1750.152 Safari/537.36"
-  cask_status=$(curl --silent ${verbose_option} --head --max-time 10 --proto =https --location --header "${browser_headers}" --write-out '%{http_code}' "${cask_url}" -o '/dev/null')
+  curl-check-url ${verbose_option} "${cask_url}" > /dev/null
   exit_code=$?
 
   if [[ exit_code -ne 0 ]]; then
     echo "curl returned ${exit_code}: FAIL for ${cask_url}"
     return 1
-  fi
-
-  if [[ "${cask_status}" != '200' ]]; then
-    if [[ -z "${cask_status}" ]]; then
-      echo "URL ${cask_url} is not reachable!"
-      return 2
-    else
-      echo "Download URL returned ${cask_status}"
-      return 3
-    fi
   fi
 
   return 0
@@ -124,7 +128,7 @@ fi
 # exit if no argument was given: Run in current directory
 if [[ -z "$1" ]]; then
   options=""
-  [[ ${skip_curl_head} -ne 0 ]] && options+=" --skip-verify"
+  [[ ${skip_curl_verify} -ne 0 ]] && options+=" --skip-verify"
   [[ ${verbose} -ne 0 ]] && options+=" --verbose"
 
   for file in *.rb;
@@ -165,14 +169,14 @@ for stanza in url appcast homepage; do
 
   replace_protocol_of_stanza ${cask_file} ${stanza} "http" "https"
 
-  if [[ ${skip_curl_head} -eq 0 ]]; then
+  if [[ ${skip_curl_verify} -eq 0 ]]; then
     check_url_for_https $(brew cask _stanza ${stanza} "${cask_name}")
   else
     true
   fi
 
   if [[ $? -ne 0 ]]; then
-    echo "Restored original value for stanza ${stanza} as curl head failed"
+    echo "Restored original value for stanza ${stanza} as curl check failed"
     replace_protocol_of_stanza ${cask_file} ${stanza} "https" "http"
   else
     updated_stanzas=$((updated_stanzas+1))


developer/bin/find_sparkle_appcast
```diff
@@ -60,7 +60,7 @@ if [[ -z "${appcast_url}" ]]; then
 fi
 
 # vaidate appcast
-appcast_http_response="$(curl --silent --head --write-out '%{http_code}' "${appcast_url}" -o /dev/null)"
+appcast_http_response="$(curl-check-url "${appcast_url}")"
```
tsparber: @vitorgalvao was there a reason to not follow redirects?

vitorgalvao: Yes. While useful in downloads, redirects are not useful in finding `appcast`s, as it is just a (typically xml) file. It’s much better to just have the final URL.

In `cask-repair` I [warn of the resulting status code](https://github.com/vitorgalvao/tiny-scripts/blob/e04896c874b6051d9847f46c17a5f9f7edde02b8/cask-repair#L132L134) so users can act accordingly. I forgot to also add the code here, though.

developer/bin/cask-switch-https
```diff
@@ -65,8 +65,7 @@ check_url_for_https() {
   [[ ${verbose} -ne 0 ]] && verbose_option="-v "
 
   # check if the URL sends a 200 HTTP code, else abort
-  browser_headers="User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/33.0.1750.152 Safari/537.36"
-  cask_status=$(curl --silent ${verbose_option} --head --max-time 10 --proto =https --location --header "${browser_headers}" --write-out '%{http_code}' "${cask_url}" -o '/dev/null')
+  curl-check-url ${verbose_option} "${cask_url}" > /dev/null
```
vitorgalvao: This won’t work because `curl-check-url` is not in `$PATH`. Something like `$(dirname "$0")/curl-check-url` is needed.

But then we also need something to check if `curl-check-url` even exists and is on hand, and abort if not.

developer/bin/curl-check-url
```diff
@@ -0,0 +1,116 @@
...
+$result = curl_exec($ch);
+$http_code = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+curl_close($ch);
+
+print("$http_code\n");
+
+if ($http_code === 200) {
+  exit(0);
+} else {
+  exit(1);
```
vitorgalvao: Exiting with the status code would be a nice touch.

developer/bin/find_sparkle_appcast
tsparber: Hmm, I think sticking to the URL provided by the app author (eq sparkle in info.plist) should less likely break. As this is the only url the author needs to care. When the author changes the redirect location the app update still works but the cask update check might fail / miss updates.

developer/bin/curl-check-url
tsparber: Exit codes are only 8 bit (0-255), so this does not work. The code is printed to stdout (code "0" if the server is not reachable).

vitorgalvao:
Ingenious solution, though I’m not sure having a dependency on an external script is wise ([some reasoning](https://github.com/caskroom/homebrew-cask/pull/16787/files#r49448107)).

I was also searching for something amongst the same lines, and found that for the most part, simply adding `--range 0-0` to the command works just fine. `--range` tells `curl` to download that byte range and then stop (so in this case, it actually downloads nothing). It works, but the server will likely respond with `206`(Partial Content) rather than `200`. This is actually fine, since the whole point here is not to get `200`, but to know if the thing actually exists, which `206` confirms.

I think, then, going the `--range` solution is better for both `cask-repair` and `find_sparkle_appcast`. [Having said that](http://youtu.be/ENHVQ2gslp8), your script does something else: it only follows https redirects, which is irrelevant for the aforementioned scripts, but extremely relevant for `cask-switch-https` (it’s its whole point). So if you can’t make it work with `--range` (or any other in-line solution) and fix the [the current issue]([some reasoning](https://github.com/caskroom/homebrew-cask/pull/16787/files#r49448107)), I’m fine with merging it.

developer/bin/find_sparkle_appcast
vitorgalvao: In light of the rest of the comments on the issue, I ask you revert this line.

developer/bin/cask-switch-https
tsparber: Whoops, will fix. I think checking for existence should not be needed (as these scripts will exist both if merged :smiley:)

vitorgalvao: I mean something simple, [like this](https://github.com/vitorgalvao/tiny-scripts/blob/9dc9afe93d3866176529186602b879553e751f9d/gifmaker#L13L20).

Point being that if users move the script somewhere else, they should know why it isn’t working.

developer/bin/find_sparkle_appcast
vitorgalvao: > Hmm, I think sticking to the URL provided by the app author (eq sparkle in info.plist) should less likely break. As this is the only url the author needs to care. When the author changes the redirect location the app update still works but the cask update check might fail / miss updates.

That’s a good point, though in practice (as you might remember, I’m currently updating all `appcast`s) I do see some `301` (haven’t kept track, so not sure how they compare to the amount of `302`), and when there’s a redirect, it does so to a URL that is likely to be the new permanent home (mostly `appcast`s moved from a generic host to the same host as the homepage, or corrected to new host when app changed hosts entirely) so it’s probably safer to keep it like this for now, and see how it goes.

tsparber:
While thinking about it, maybe `curl-check-url` should not be merged into homebrew-cask at all. As it is not directly related to homebrew-cask and may even be useful in other scripts/situations.
I really like the idea of your [tiny-scripts](https://github.com/vitorgalvao/tiny-scripts/) and its homebrew tap. Maybe include it there (or I create something like that myself). What do you think?

Then a notice to `brew install ...` would be sufficient and other scripts could depend on it.

Using range was not a good solution for me. I was trying that, and many servers do not support it (returning the whole document or an error). 

> it only follows https redirects, which is irrelevant for the aforementioned scripts

Before this discussion I saw no point in not following redirects, will add a option similar to curl.

developer/bin/find_sparkle_appcast
tsparber: :+1:

developer/bin/curl-check-url
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function usage() {
+  global $program;
+
+  $usage = "
+    This script checks if a given URL returns a HTTP status code 200 using GET
+
+    Some servers do not support HEAD requests, this script uses GET but aborts
+    after the headers have been received. This minimized the overall traffic.
```
Amorymeltzer: Should maybe read: "have been received**; this minimizes** the overall traffic." to match tense.

vitorgalvao:
> As it is not directly related to homebrew-cask and may even be useful in other scripts/situations.

That’s the case with many of the scripts in `developer/bin` (all that start with `list`), though I can see a broader appeal for `curl-check-url`, yes.

> I really like the idea of your [tiny-scripts](https://github.com/vitorgalvao/tiny-scripts/) and its homebrew tap. Maybe include it there

It’d be a bit weird to include there something I haven’t authored from scratch (since that’s really the only thing connecting those; I see it as a `dotfiles` for personal scripts), though more importantly is I don’t write php, so I wouldn’t be able to support it correctly.

> (or I create something like that myself)

If you plan on making more, a repo has been working well for me. If you only plan on having that one, I’d recommend a gist.

> Then a notice to `brew install ...` would be sufficient and other scripts could depend on it.

Being able to make the main script pull the dependencies directly (by having a formula for it) instead of just showing the message is one of the reasons that makes me have `cask-repair` out of the `homebrew-cask` repo. That said, `cask-repair` is something you run all the time and makes sense to be readily available (in your `PATH`), while `cask-switch-https` is something specific, to be run only occasionally, and not by regular users, so I’d say it’s OK.

> Using range was not a good solution for me. I was trying that, and many servers do not support it (returning the whole document or an error).

Do you have some examples? (So I can test in the context of `cask-repair` and see if it’s worth to add).

tsparber:
Updated to use the script from an external repo.

> Do you have some examples? (So I can test in the context of cask-repair and see if it’s worth to add).

Sorry not at the moment, will try to find them again.

--
Not sure why travis fails...

developer/bin/curl-check-url
tsparber: Thank you :smiley:

developer/bin/cask-switch-https
```diff
@@ -12,11 +12,23 @@ syntax_error() {
   exit 1
 }
 
+depends_on() {
+  if [[ ! $(which "${1}") ]]; then
+    echo -e "$(tput setaf 1)
+      This script depends on '${1}'.
+      If you have [Homebrew](http://brew.sh), you can install it with 'brew install ${2}'.
```
vitorgalvao: This is a nice touch, adding the repo as an argument. I like that improvement.

vitorgalvao:
> Not sure why travis fails...

Neither am I. Restarted it, and will take another look in the morning.

developer/bin/cask-switch-https
tsparber: Now with even less typing :smiley: . The command is derived from the repo, or can be added as a second option (if formula / bin names differ)

tsparber:
>  it only follows https redirects, which is irrelevant for the aforementioned scripts

There is a [`--no-location`](https://github.com/tsparber/homebrew-tiny-scripts/blob/master/bin/curl-check-url#L34) option now. The script prints 302/301 and exits with 1.

PR #16948, opened 2016-01-12
Appcasts without pubDate

vitorgalvao:
Pinging @caskroom/maintainers (also @Amorymeltzer).

This PR changes the `:sha256` of every `appcast` in the main repo to the wrong one. That seems like a strange thing to do, especially since I just spent days making sure every single one was correct, so naturally there’s an explanation.

There are currently a few (around ten) appcasts that are impossible to use. Reason being they have a `<pubDate>` tag for the whole feed, that gets rebuilt periodically (sometimes with every request) regardless of it the app was updated or not. That tag seems to be the regular culprit, so instead of removing `appcast` from those few casks, I instead did `curl "{{appcast_url}}" | sed '/<pubDate>/d' | shasum --algorithm 256` on every cask, to get new, reliable appcast shasums.

This might seem overkill just to fix a handful of casks (think of the user confusion!) but in reality [it could win us over 150 new appcasts](https://github.com/caskroom/homebrew-cask/issues/16685).

So that’s the decision to make. Either we keep `:sha256` representing an accurate value, or we get flexibility in a lot more casks. As this PR shows, I’m more inclined for the second option. As long as we’re clear [on the documentation](https://github.com/caskroom/homebrew-cask/issues/16689) how it is we get this value (I would also change the behaviour in `cask-repair`), we should (I hope) be OK. [Auditing](https://github.com/caskroom/homebrew-cask/issues/16529) would also need to take this into account.

Something else we could do is deprecate `:sha256` in favour of a new value. Call it `:checkpoint` or something like that. That way there’s no way we can get reasonable complaints the value isn’t representative of the true shasum, because it isn’t and we’re making that clear. It’s a checkpoint value we use for ourselves, and users encountering it would need to consult the documentation to understand what it means (and hence learn how to correctly set it).

So in the end the idea ends up being simply about:
- Deprecate `:sha256` in `appcast`.
- Add `:checkpoint` (or another name) to `appcast`.

But I wanted to make this PR as soon as possible for discussion, now that we know every appcast is up to date.

### DIFF
--- a/Casks/pixlr.rb
+++ b/Casks/pixlr.rb
@@ -4,7 +4,7 @@ cask 'pixlr' do
 
   url "https://cdn2.pixlr.com/mac/Autodesk_Pixlr_v#{version}_Mac_Installer.dmg"
   appcast 'https://updatefeed.pixlr.com/pixlr-for-mac-appfeed.xml',
-          :sha256 => '46d01575b9fd926b62093e9d5614544428f49a5f1013217b92c29b2dbed78966'
+          :sha256 => '82cada86fe1b784868a714ea6085f295c45deb7c11299adc6acb4cf2372547de'
   name 'Autodesk Pixlr'
   homepage 'https://pixlr.com'
   license :unknown # TODO: change license and remove this comment; ':unknown' is a machine-generated placeholder
--- a/Casks/platypus.rb
+++ b/Casks/platypus.rb
@@ -4,7 +4,7 @@ cask 'platypus' do
 
   url "http://sveinbjorn.org/files/software/platypus/platypus#{version}.zip"
   appcast 'http://www.sveinbjorn.org/files/appcasts/PlatypusAppcast.xml',
-          :sha256 => '2eb402188ef9006152796d65342be4ef75fd5fbc3c8a516f7f3a294509a9a5b9'
+          :sha256 => 'b88c3e9846267f04d5f2fb053e9eeb158ef3072f2e53a472803fb5778a63656d'
   name 'Platypus'
   homepage 'http://sveinbjorn.org/platypus'
   license :bsd
--- a/Casks/playlist-assist.rb
+++ b/Casks/playlist-assist.rb
@@ -4,7 +4,7 @@ cask 'playlist-assist' do
 
   url "http://dougscripts.com/itunes/scrx/playlistassistv#{version.delete('.')}.zip"
   appcast 'http://dougscripts.com/itunes/itinfo/playlistassist_appcast.xml',
-          :sha256 => '28442470f6d3f8dd413dedef0888e051f594799a2861f83134d8ebe21f4a014f'
+          :sha256 => 'd9d3250922561c189de4c3746ba5d36c9f2b03590e7787ff5155128d9b6e5151'
   name 'Playlist Assist'
   homepage 'http://dougscripts.com/apps/playlistassistapp.php'
   license :commercial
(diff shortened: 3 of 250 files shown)


adityadalal924:
Cancelled the Travis job.

I'd be more comfortable if https://github.com/caskroom/homebrew-cask/issues/16529 was implemented before this change, honestly.

vitorgalvao:
> I'd be more comfortable if https://github.com/caskroom/homebrew-cask/issues/16529 was implemented before this change, honestly.

Having the audit won’t change this. This change was made using the system the audit will use, so we can be sure it is correct now, because all appcasts are. That won’t stay like that for long, though (It took me a few days to get them all up to date, and after finishing I still had to red a bunch of them due to updates).

If we agree on this new system, we can merge the change as it is, and then change all occurrences of `:sha256` to `:checkpoint` (the name) and rewire one to the other, and extend audit, all at the same time.

adityadalal924:
In that case, LGTM.

Amorymeltzer:
I am... surprisingly okay with this!

- Not wild about `:checkpoint`.  Maybe `:hash` or `:caskhash`?  Not coming up with any other ideas but `:checkpoint` feels actively misleading, like a last-accessed date.
- I'll repeat this in #16949 but I think there must be a simple script to compute this value for a wannabe-contributor

vitorgalvao:
With `:hash`, `:checksum`, and variants, I worry users take the hexadecimal value too literally (i.e. they wrongly deduce what it is/how to get it). For that reason I think a name that will make you wonder “what exactly does that mean” might be somewhat beneficial, to force contributors to read the documentation on it.

`:value` is something that crossed my mind as well. A bit generic, but not too bad. Like you too, I’m also not a huge fan of `:checkpoint`, but it is specific.

How about `:control`? As in “control point”, or “control value”. That seems accurate.

adityadalal924:
@Amorymeltzer The documentation would look something like https://github.com/caskroom/homebrew-cask/issues/16689#issuecomment-170583701

(See also the WIP CONTRIBUTING rewrite and documentation split-up issues)

Amorymeltzer:
Yeah, I like vague; `:value` helps keep that "hmm," as you say.  Would it be crazy to do `:version` or `:update`?  Correct message, clear it is connected to the stanzas above, and obvious it needs to be updated.

vitorgalvao:
`:identifier`? It uniquely identifies a specific appcast at a specific point in time.

adityadalal924:
`:update_id`?

vitorgalvao:
I’d be fine with `:id`. The name is so insignificant/meaningless in this matter, I’d rather not need two words for it. Lets move the name discussion to https://github.com/caskroom/homebrew-cask/issues/16949, though.

Already there are conflicts cropping up, so we just need to decide on if doing this is a good idea at all. We three seem to agree it is. @jawshooah any objection?

jawshooah:
I'm fine with this. Not sure about the naming, but as you said we can move that discussion to #16949.

adityadalal924:
ID is fine with me, fwiw

vitorgalvao:
Will fix conflicts and merge, then.

vitorgalvao:
As a heads-up, the command to get `:checkpoint` will require a small tweak. Hadn’t noticed it in my initial tests, but caught it now. Automating it and will include the updated values in https://github.com/caskroom/homebrew-cask/issues/16951.

Instead of `sed '/<pubDate>/d'`, it’ll be `sed 's|<pubDate>[^<]*</pubDate>||g'`. The edge-case (30 casks) are `appcast`s that are a single-line xml, in which case the previous code would delete the whole line, making the file essentially empty. This way, we’re only deleting the problematic part.

PR #16951, opened 2016-01-13
Implement `appcast` `checkpoint`

vitorgalvao:
Implements ideas in https://github.com/caskroom/homebrew-cask/pull/16948 and https://github.com/caskroom/homebrew-cask/issues/16949.

Uses `:checkpoint`, but that can be easily changed. This was to get a feel for what will need to be changed, core-wise, and be able to merge as soon as possible, if agreed upon.

Tests fail, but that’s expected. After changing all `:sha256` to `:checkpoint`, they should pass.


adityadalal924:
LGTM after Travis passes (it'll time out, but as long as `brew cask audit` didn't throw any errors for you locally it should be fine) - waiting too long will end up causing lots of merge errors, unfortunately.

vitorgalvao:
Finally passing. Any objections with merging?

Pretty straightforward changes, but I’ll point out two details to who checks:

+ [`:sha256` is still there](https://github.com/vitorgalvao/homebrew-cask/commit/700a024c334c8b7282eb79d43963aa4e44a8aa4e), with a `#DEPRECATED` comment. Though I don’t think it’s needed, really, and we could likely get rid of it entirely (what user tap would care about `appcast`?).
+ [Checking for an invalid `:checkpoint`](https://github.com/vitorgalvao/homebrew-cask/commit/c87a9579439fa1b2af262a2d973c0d70c16686f5) still piggybacks on checking for an invalid sha256 and will output the same `sha256 string must be of 64 hexadecimal characters` generic message if it’s wrong. Adding a repeated test to change that message in this regard, when it is still correct like this, seemed to not make much sense.

Pinging @jawshooah @adityadalal924 (feel free to merge, if everything looks fine).

adityadalal924:
Both points look fine to me - we can remove `:sha256` from the core code after switching over all the repos

adityadalal924:
Merged - Locally, you may want to `git gc; git repack -Ad; git prune` to keep repo size down after this PR.

PR #16966, opened 2016-01-13
Add Dusty (0.6.5)

thieman:
This PR adds a cask recipe for [Dusty](http://dusty.gc.com), a Docker-based development environment manager. https://github.com/gamechanger/dusty

### DIFF
--- a/Casks/dusty.rb
+++ b/Casks/dusty.rb
@@ -1,0 +1,24 @@
+cask 'dusty' do
+  version '0.6.5'
+  sha256 '6e88894c7062f24bb8b4b7b2faa149b6b638c7d3fa2be86a56266e2d1789ea60'
+
+  url "https://github.com/gamechanger/dusty/releases/download/#{version}/dusty.tar.gz"
+  appcast 'https://github.com/gamechanger/dusty/releases.atom',
+          :checkpoint => '14dd655e5ed3a55495436c0fc9fb426bedb461a600083d10f72f330c66181d7a'
+  name 'Dusty'
+  homepage 'https://github.com/gamechanger/dusty'
+  license :mit
+
+  depends_on :cask => 'dockertoolbox'
+  container :type => :tar
+
+  installer :script       => 'brew-install.sh',
+            :args         => %W[#{staged_path}],
+            :must_succeed => true,
+            :sudo         => true
+  binary 'dusty'
+
+  uninstall :launchctl => 'com.gamechanger.dusty'
+
+  zap :delete => '/etc/dusty'
+end


Casks/dusty.rb
```diff
@@ -0,0 +1,38 @@
...
+  plist_name = 'com.gamechanger.dusty.plist'
+
+  url "https://github.com/gamechanger/dusty/releases/download/#{version}/dusty.tar.gz"
+  appcast 'https://github.com/gamechanger/dusty/releases.atom',
+          :sha256 => 'c38f94f0be6d2b29a458994ca3f415c66557f4329de0b5926d66624264618b56'
+  name 'Dusty'
+  homepage 'https://github.com/gamechanger/dusty'
+  license :mit
+
+  depends_on :formula => 'git' if MacOS.release < :mavericks
```
vitorgalvao: Not needed. You already need `git` to use homebrew-cask in the first place.

vitorgalvao:
This cask is atypical, and in need of multiple fixes. For one, we don’t set variables inside casks, and in the rare times we do, we use other methods. Here, it’s not really justified.

And you’re misunderstanding [`installer :script`](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#installer-script). Not only should there be at least one, it doesn’t mean “write a script for me to execute”, it means “point me to an existing script to execute”. What you’re looking for is [`preflight`](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#flight-stanzas-details).

However, you’re not supposed to run commands arbitrarily unless they’re absolutely needed for the installation. Why are you trying to run these, and what exactly do you wish to accomplish?

thieman:
Attempted to address stylistic concerns. Still trying to grok the script stuff. Is there a different set of docs explaining what `preflight` actually does? That seems more like a command reference and I can't find anything more conceptual.

vitorgalvao:
See [casks that use it](https://github.com/caskroom/homebrew-cask/search?p=2&q=preflight&utf8=%E2%9C%93). It lets you run system commands, like you’re trying to do, and does so after download and before other install operations.

thieman:
So would you recommend moving the four commands I need into a script that's included in the tarball or using preflight?

I'm mostly just restating the script here but this is what I'm trying to do:

1. Copy the plist into a standard location `/Library/LaunchDaemons`
2. Unload any previous plists (returns 0 even if none defined)
3. Run the app's preflight check, which mostly ensures that the requirements it needs are installed and of supported versions
4. Load the plist so the app runs on subsequent system starts

adityadalal924:
Looks like you're trying to avoid using `install.sh`, which really isn't what Homebrew Cask is - anaconda is another cask that uses a shell script, but notice it's provided by the vendor (which is what we want)

thieman:
Cool, I'll do that then.

adityadalal924:
Yep, that ends up being much simpler for everyone - installer logic is handled upstream (as it should be), and we just provide a nice(r) interface for installing.

thieman:
I think this is close now but I've got a slight problem in that the last line of my script:

```
launchctl load /Library/LaunchDaemons/com.gamechanger.dusty.plist
```

Shouldn't be run until after the binary has been moved into its final location. It eventually works as-is now because launchctl tries after a few seconds to restart the process, and then it's where it expects to be, but it's sub-optimal. Would this be a legitimate use of something like postflight?

adityadalal924:
So there are two ways you could go here

 - Use `postflight` in HBC
- Create a `.pkg` and release that - it seems to be the standard for "more complicated" installs. You're almost already there, imo, because the `postflight` is also something you can do in a installer (http://stackoverflow.com/questions/17396518/os-x-how-do-you-create-an-installer-pkg-that-sets-up-one-launchd-plist-as-root). It also closes https://github.com/gamechanger/dusty/issues/610 and simplifies your install directions - they are inherently auditable (and can be signed as well).

We (Homebrew-Cask) are fine with either option - I will say though, that we have innumerably more Casks using option (2) than (1)

adityadalal924:
https://github.com/nanobox-io/nanobox 's Cask and nanobox-installer may also be of interest

adityadalal924:
Upon further looking through things (commenting again so you get a notification):

Wouldn't want you to rebuild your entire build system just to repackage in a "better" format - The Cask actually seems fine as-is. 

Lots of programs do (much worse) than calling launchctl a tad early, so I don't think you even need postflight in this case.

Technically, since everything is in the tarball, it is auditable.

Casks/dusty.rb
```diff
@@ -0,0 +1,22 @@
+cask 'dusty' do
+  version '0.6.5'
+  sha256 '6e88894c7062f24bb8b4b7b2faa149b6b638c7d3fa2be86a56266e2d1789ea60'
+
+  url "https://github.com/gamechanger/dusty/releases/download/#{version}/dusty.tar.gz"
+  appcast 'https://github.com/gamechanger/dusty/releases.atom',
+          :sha256 => '14dd655e5ed3a55495436c0fc9fb426bedb461a600083d10f72f330c66181d7a'
```
vitorgalvao: Should be `:checkpoint` instead of `:sha256`.

Code for it not yet merged, but should be pretty soon.

thieman:
That all sounds good to me, thanks.

thieman:
Test failure is the `checkpoint` change, can re-run tests once that's merged

adityadalal924:
Yep, will re-run tests and merge after https://github.com/caskroom/homebrew-cask/pull/16951.

Casks/dusty.rb
```diff
@@ -0,0 +1,22 @@
...
+  depends_on :cask => 'dockertoolbox'
+  container :type => :tar
+
+  installer :script       => 'brew-install.sh',
+            :args         => %W[#{staged_path}],
+            :must_succeed => true,
+            :sudo         => true
+  binary 'dusty'
+
+  uninstall :launchctl => 'com.gamechanger.dusty'
```
jawshooah: Ah, didn't notice the `binary 'dusty'` above. Is there anything apart from that binary and the launch daemon? Configs, cache files, etc. can go in a `zap` stanza, though it's not required.

thieman: Dusty does create some config dirs for itself during operation which we could potentially clean up as part of a brew uninstall. That should all be within `/etc/dusty`, not sure what the best way to get rid of that is. Will look into zap and uninstall commands.

thieman:
Added a `zap` directive to delete Dusty's config directory

vitorgalvao:
There should be an empty line between `uninstall` and `zap` (or Travis will complain). Other than that, it looks fine. Could you please [squash your commits](http://davidwalsh.name/squash-commits-git)?

thieman:
Done.

adityadalal924:
Thank you for the contribution. It was merged directly as e6e2b0dc057aa72564bafcb77ff3a07d09d82a6f to keep commit history cleaner. Your contribution is still credited to you.

PR #16974, opened 2016-01-13
find_sparkle_appcast: Change to checkpoint

adidalal:
Ref: https://github.com/caskroom/homebrew-cask/issues/16949

Not tested locally, but _should_ work


vitorgalvao:
Looks good.

PR #16979, opened 2016-01-13
Delete things.rb

vitorgalvao:
MAS-only, it seems.


adityadalal924:
Can be bought from http://sites.fastspring.com/culturedcode/product/main

vitorgalvao:
Thank you. Missed that.

PR #16981, opened 2016-01-13
find_sparkle_appcast: point out what address was tried when there is an error

vitorgalvao:


adityadalal924:
Makes sense :+1:

vitorgalvao:
Especially when there are redirects, I’m really missing this, since I’m then forced to look for it manually.

PR #17006, opened 2016-01-14
appcast: remove latest_version

vitorgalvao:
I don’t even know/remember what the experiment was about, and how this could possibly be useful.


tsparber:
Added in #4847 (not really helpful)

vitorgalvao:
Also https://github.com/caskroom/homebrew-cask/pull/6680.

No clue about its purpose.

adityadalal924:
I guess we could jut kill it - there's no documentation anywhere about it that I can see, either.

PR #17037, opened 2016-01-14
MacVim: Update to 7.4.90

simono:


Casks/macvim.rb
```diff
@@ -1,6 +1,6 @@
 cask 'macvim' do
-  version '7.4.89'
-  sha256 '779928de48eabc3f90a830e441affb0f28e370881737d2b3f4bf173c41a1798f'
+  version '7.4.90'
+  sha256 'ea480f1f69e524bb9302bdebb66ca0e854e0ab6ec951f7f58e16be46d86bac00'
 
   url "https://github.com/macvim-dev/macvim/releases/download/snapshot-#{version.patch}/MacVim-#{version.major_minor}.dmg"
   appcast 'https://github.com/macvim-dev/macvim/releases.atom',
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can typically find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

You can easily install it with `brew install vitorgalvao/tiny-scripts/caskrepair`. `cask-repair --help` will show you some options, but as a whole the script will guide you through the update process.

simono:
Fixed with cask-repair. See #17038

Casks/macvim.rb
simono: The correct command to install `cask-repair` is
```shell
brew install vitorgalvao/tiny-scripts/cask-repair
```
(Missing hyphen).

Thanks for the great tool :-)

vitorgalvao: Thank you for the kind words and the correction (fixed in my TextExpander snippet).

PR #17055, opened 2016-01-15
Updated sqlitestudio (add zap)

tsparber:
Is the `:rmdir => '~/.config'` needed or would `:delete/:trash` remove the empty parent dir automatically?

Would http://sqlitestudio.pl/rss.rvt be suitable as an appcast? It does not only contain app updates.

### DIFF
--- a/Casks/sqlitestudio.rb
+++ b/Casks/sqlitestudio.rb
@@ -3,9 +3,14 @@ cask 'sqlitestudio' do
   sha256 '9f93d31dfcb6746fb613aed31920b14e34932ae0055c811e74b1aa01ef060a7c'
 
   url "http://sqlitestudio.pl/files/sqlitestudio3/complete/macosx/sqlitestudio-#{version}.dmg"
+  appcast 'http://sqlitestudio.pl/rss.rvt',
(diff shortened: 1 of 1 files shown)


Casks/sqlitestudio.rb
```diff
@@ -8,4 +8,8 @@
   license :gpl
 
   app 'SQLiteStudio.app'
+
+  zap :delete => '~/Library/Saved Application State/com.yourcompany.SQLiteStudio.savedState',
+      :trash => '~/.config/sqlitestudio',
+      :rmdir => '~/.config'
```
vitorgalvao: This is a standard directory, so it shouldn’t be removed, even if empty. Other tools may depend on it already being there.

vitorgalvao:
> Would http://sqlitestudio.pl/rss.rvt be suitable as an appcast?

Yes.

> It does not only contain app updates.

Doesn’t matter, since we can’t parse those automatically anyway. Any online file that changes whenever there’s a new app version is suitable, since we can query it and confirm it changed, meaning a new version is out (which we’ll then need to manually update).

adityadalal924:
Thank you for the contribution. It was merged directly as aed897bc03de681f93429530235c9a30c05bacfc to keep commit history cleaner. Your contribution is still credited to you.

PR #17058, opened 2016-01-15
Delete FAQ.md

vitorgalvao:
As part of revamping docs.

Currently useless. All the information on it is already covered in other places.


adityadalal924:
Another minor point to add about the revamp: https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/update_issue_template_urls should be updated accordingly

vitorgalvao:
Fixing that now.

PR #17061, opened 2016-01-15
Proposal: Script to create appcast stanza from url

tsparber:
This script comes handy when you already have a appcast url (example rss feed) and want add it to the cask.
To reduce duplication (but adding a dependency) this script could also be called in `find_sparkle_appcast`.

### DIFF
--- a/developer/bin/appcast_from_url
+++ b/developer/bin/appcast_from_url
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+set -o pipefail
+
+readonly program="$(basename "$0")"
+verbose=0
+
+syntax_error() {
+  echo "$program: $1" >&2
+  echo "Try \`$program --help\` for more information." >&2
+  exit 1
+}
+
+usage() {
+  echo "
+    This script generates a appcast stanza for the given url
+
+    usage: $program [options] <appcast_url>
+    options:
+      -v, --verbose                 Show more verbose output.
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
Appears to be one rung away from relevant @vitorgalvao comment: https://github.com/caskroom/homebrew-cask/issues/16949#issuecomment-171127783

vitorgalvao:
Thank you for the effort, but the answer is pretty much what @Amorymeltzer linked to.

`find_sparkle_appcast` works well on its own and is useful even if you just want the appcast (and don’t care about homebrew-cask). Adding dependencies in general is already bad, but to do it to such a small and simple script, without even an easy way to solve them all at once (e.g. with a formula) or without forcing homebrew in general to users, is a pain.

Like I said in that issue, adding `get_checkpoint() { curl --compressed "$1" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256; }` to your shell’s startup file already works perfectly, no need for a full blown script and making another one worse. Heck, we can even `get_checkpoint() { curl --silent --compressed "$1" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{print $1}' | pbcopy; }` and all you can immediately paste it to the cask.

tsparber:
I see your point. I grew a bit too much.

PR #17063, opened 2016-01-15
find_sparkle_appcast: Suggest stanza location

tsparber:
~~Suggest the correct order of stanzas to the user.~~
Add indentation to allow easy copy&paste.

### DIFF
--- a/developer/bin/find_sparkle_appcast
+++ b/developer/bin/find_sparkle_appcast
@@ -59,7 +59,7 @@ if [[ -z "${appcast_url}" ]]; then
   exit 0
 fi
 
-# vaidate appcast
+# validate appcast
 appcast_http_response="$(curl --silent --head --write-out '%{http_code}' "${appcast_url}" -o /dev/null)"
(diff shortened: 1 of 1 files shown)


developer/bin/find_sparkle_appcast
```diff
@@ -67,10 +67,12 @@ appcast_checkpoint=$(curl --silent "${appcast_url}" | sed 's|<pubDate>[^<]*</pub
...
+echo "A Sparkle appcast was found. You should add it to your cask between 'url' and 'name':
 
-appcast '${appcast_url}',
-        :checkpoint => '${appcast_checkpoint}'
+  url ...
+  appcast '${appcast_url}',
+          :checkpoint => '${appcast_checkpoint}'
+  name ...
 
 You should likely also add 'auto_updates true'"
```
vitorgalvao: But then this line should also get positional instructions. To add positions at all, I’d rather link to the documented stanza order, in case we ever change it.

tsparber: Yes, linking to the documented stanza order seems to be better and more general. Maybe after the doc revamp?

vitorgalvao: Yes, it’s probably better to wait for that. I’ve started working on it last night and will continue today.

PR #17064, opened 2016-01-15
freesmug-chromium 47.0.2526.111

rhendric:


vitorgalvao:

rhendric:
Thanks for sharing cask-repair!

But it looks like you removed the appcast directive from freesmug-chromium in abc7cbb? That's the cause of this merge conflict, anyway. How would you like to proceed?

vitorgalvao:
Yes, I removed it due to [this](https://github.com/caskroom/homebrew-cask/issues/16685), but it no longer applies.

To fix the merge conflict, you can add it back.

rhendric:
Done!

PR #17067, opened 2016-01-15
Update zap stanza android-file-transfer.rb

miccal:
The directory `'~/Library/Application Support/Google/` is used my other Google programs, and should not be removed if someone is only uninstalling Android File Transfer.

### DIFF
--- a/Casks/android-file-transfer.rb
+++ b/Casks/android-file-transfer.rb
@@ -10,5 +10,5 @@ cask 'android-file-transfer' do
 
   app 'Android File Transfer.app'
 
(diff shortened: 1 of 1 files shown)


miccal:
I tried to squash, but something went wrong - sorry!

vitorgalvao:
> The directory `'~/Library/Application Support/Google/` is used my other Google programs, and should not be removed if someone is only uninstalling Android File Transfer.

It isn’t. `:rmdir`, [like the command](https://en.wikipedia.org/wiki/Rmdir) only removed empty directories. If you have anything else in there, it won’t be erased.

miccal:
Of course, my apologies.

PR #17093, opened 2016-01-16
Update linear.rb

miccal:


miccal:
The spaces in the zap stanza are correct in Vim?

Casks/linear.rb
```diff
@@ -1,8 +1,8 @@
 cask 'linear' do
-  version '1.1.0'
-  sha256 '2f725f9fd377d8c4ca22c6c59a53bdc6cfc7dc90328345ba7ec29166e8d6ec90'
+  version '1.2.0'
+  sha256 '2bb17d6c529d4a8b8a8443803e44d976271ca298bef3f22b43ca9d5b41dc8327'
 
-  url "https://github.com/mikaa123/linear/releases/download/#{version}/linear.zip"
+  url "https://github.com/mikaa123/linear/releases/download/v#{version}/linear.zip"
   appcast 'https://github.com/mikaa123/linear/releases.atom',
           :checkpoint => '604b2fdad9d513e82c80ae1efccab182aa36335a7488b14eb753158969a6b571'
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

vitorgalvao:
> The spaces in the zap stanza are correct in Vim?

That’s because you’re using tabs, and they need to be spaces.

Casks/linear.rb
miccal: I have done that - the command `curl --silent --compressed "https://github.com/mikaa123/linear/releases.atom" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'` returns the sha `604b2fdad9d513e82c80ae1efccab182aa36335a7488b14eb753158969a6b571`.

miccal: I did not use `cask-repair` because I also added the `zap`.

vitorgalvao: `cask-repair` doesn’t prevent you from adding to the cask. When it asks if everything is OK, press <kbd>e</kbd> and you can edit the cask further.

miccal: I'll remember that in the future, thanks.

So why is the command `curl --silent --compressed "https://github.com/mikaa123/linear/releases.atom" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'` returning the same sha?

vitorgalvao: Probably was updated during the last mass rename. Not to worry.

miccal: OK, I will fix the spacing.

miccal:
Fixed and squashed.

vitorgalvao:
Thank you for the contribution. It was merged directly as 065ea428826ae0b8049e4cc650c9c077a1c60c66 to keep commit history cleaner. Your contribution is still credited to you.

miccal:
Thanks again for your help @vitorgalvao.

PR #17157, opened 2016-01-17
Add Kollaborate Plugin Pack 1.0.6.0

NaanProphet:
Upload directly to the cloud from Final Cut Pro X and Adobe Premiere
Pro to Digital Rebellion's Kollaborate video editing servers. Also
can integrate with OS X's Today window to see project activity at a
glance.

Note: manual execution of the package installer specified in order for
users to select which plugin(s) they need.

### DIFF
--- a/Casks/kollaborate-plugin-pack.rb
+++ b/Casks/kollaborate-plugin-pack.rb
@@ -1,0 +1,19 @@
+cask 'kollaborate-plugin-pack' do
+  version '1.0.6.0'
+  sha256 '74d11ff0b93650fc67e48959e377fa9ea2d7cb029eb6e79dc07b4808d8c1ab9d'
+
+  # digitalrebellion.com is the official download host per the vendor homepage
+  url "http://www.digitalrebellion.com/download/kollabplugins?version=#{version.no_dots}"
+  name 'Kollaborate Plugin Pack'
+  homepage 'https://www.kollaborate.tv/resources'
+  license :gratis
+
+  installer :manual => 'Install Kollaborate Plugin Pack.pkg'
+
+  uninstall :pkgutil => [
+                          'com.digitalrebellion.KollabPluginPack',
+                          'com.digitalrebellion.pkg.KollabPluginPack.FCPX',
+                          'com.digitalrebellion.pkg.KollabPluginPack.Premiere',
+                          'com.digitalrebellion.pkg.KollabPluginPack.SystemPlugins',
+                        ]
+end


SirDap:
Anyone think i should change it to an automatic install to install all plugins? Not sure what the convention is.

jawshooah:
I'd suggest just installing all plugins automatically with the `pkg` stanza.

SirDap:
Got it, will recommit momentarily!

Casks/kollaborate-plugin-pack.rb
```diff
@@ -0,0 +1,19 @@
+cask 'kollaborate-plugin-pack' do
+  version '1.0.6.0'
+  sha256 '74d11ff0b93650fc67e48959e377fa9ea2d7cb029eb6e79dc07b4808d8c1ab9d'
+
+  # digitalrebellion.com is the official download host per the vendor homepage
+  url "http://www.digitalrebellion.com/download/kollabplugins?version=#{version.no_dots}"
+  name 'Kollaborate Plugin Pack'
+  homepage 'https://www.kollaborate.tv/resources'
+  license :commercial
```
jawshooah: It's listed as free on the homepage. Why `:commercial`?

SirDap:
So it looks like these plugins don't uninstall properly with the `pkgutil` command for some reason.

For example, if i install the package and open FCP X, the new Kollaborate export destination will be there. However after the `brew cask uninstall` command, the export destination is till in FCP X.

If however one doesn't open FCP X after installing the package, then after uninstalling the Kollaborate destination will not appear. Similar behavior with the Today notification widget.

Sounds like it needs manual add zap stanzas ... right?

SirDap:
Update: checking the file system, cask is definitely deleting the plugin files it creates. FCP X is just caching the preset from `/Library/ProApps/Share Destinations/` to someplace, somehow. Thinking we can let it be. Should i add a caveat instead to clarify?

jawshooah:
> Should i add a caveat instead to clarify?

No, `caveats` should only be used for information pertaining to installation. Might be worth filing a bug report upstream, though.

SirDap:
Awesome. Just changed license to `gratis` and no `caveat` block. Will follow up with Digital Rebellion support!

adidalal:
Thank you for the contribution. It was merged directly as 8fffffde444a7ed122309736e45ecbed421e838b to keep commit history cleaner. Your contribution is still credited to you.

PR #17159, opened 2016-01-17
Add ScopeBox 3.4.3

NaanProphet:
Monitor and capture everything from DV to 4K with top of the line
waveform and vectorscopes. Built-in ScopeLink allows use of scopes
directly in Premiere Pro, Final Cut Pro X and more.


SirDap:
Dang, requested from the wrong branch. Please ignore.

PR #17192, opened 2016-01-17
updated querious (1.1.4)

vitorgalvao:


Casks/querious.rb
```diff
@@ -1,10 +1,10 @@
 cask 'querious' do
-  version :latest
-  sha256 :no_check
+  version '1.1.4'
+  sha256 'c8553c7a81c777201cd4c155329ae0682edcac2d6fa8e21c2949da7bd7023ad2'
 
   url 'http://www.araelium.com/querious/downloads/Querious.dmg'
   appcast 'https://store.araelium.com/updates/querious',
-          :checkpoint => 'd80fc2dda3e645bb622abd2bf72e02dce2e7266859e2cc11c54812f30a5f0c54'
+          :checkpoint => 'appcast is probably incorrect, as a non-200 (OK) HTTP response code was returned (302)'
```
jawshooah: This line causes the audit to fail. Wouldn't it be easier to just remove the appcast?

vitorgalvao: There is an appcast, it’s just the current URL that is wrong. The message is auto-inserted by `cask-repair`, I just didn’t pay enough attention and let this one slide.

I should add a feature to it that won’t let you submit if audit doesn’t pass.

PR #17206, opened 2016-01-17
Add Adobe Bridge CC

NaanProphet:
Created cask for Adobe Bridge, a file explorer capable of enriching
media with XMP metadata. Cross platform and free for life.

### DIFF
--- a/Casks/adobe-bridge-cc.rb
+++ b/Casks/adobe-bridge-cc.rb
@@ -1,0 +1,21 @@
+cask 'adobe-bridge-cc' do
+  version '2015'
+  sha256 'd8f27894419077792b3a01c63150c7bfe43599cf3ebeb240770046c56bc79c94'
+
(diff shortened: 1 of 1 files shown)


SirDap:
Not entirely sure if the `version` is correct. It's technically named *Adobe Bridge CC* whereas the Creative Cloud apps are named *Adobe Illustrator CC 2015*, etc. See full list [here](http://prodesigntools.com/adobe-cc-2015-direct-download-links.html)

Casks/adobe-bridge.rb
```diff
@@ -0,0 +1,21 @@
+cask 'adobe-bridge' do
```
vitorgalvao: Should be `adobe-bridge-cc`.

SirDap:
Great, thanks! Btw, the cookie caching is *really, really* neat! Copied that from the Adobe Illustrator CC cask.

PR #17266, opened 2016-01-18
Use Ruby 1.9 hash syntax in casks and docs

### LINKED ISSUE #17161: RuboCop: Ruby 1.9 vs 1.8 hash syntax (colons vs hashrockets)
Currently we use hashrockets (`:foo => 'bar'`) everywhere in casks. This syntax has been superseded by the more concise colon syntax (`foo: 'bar'`) since Ruby 1.9, which is already so old that it's been EOL'd as of February 2015.

We don't even support 1.8 anymore, so I see no reason to continue to use the old hashrocket syntax. What do you guys think about converting hashrockets to colons where possible?

jawshooah:
Closes #17161


jawshooah:
Likely need to wait until the current batch of PRs from @vitorgalvao is merged to resolve conflicts.

vitorgalvao:
All done!

We no longer have any casks with an `appcast` and `version :latest`.

jawshooah:
:tada:

Issue #17075, opened 2016-01-15
Remove Casks hosted on Google Code


adidalal:
As Google Code is [dead in 10 days](http://google-opensource.blogspot.com/2015/03/farewell-to-google-code.html). List from https://github.com/caskroom/homebrew-cask/issues/17075#issuecomment-172418313.

Move to alehouse?
- [x]  [`cronnix.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/cronnix.rb)
- [x]  [`epub-to-pdf.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/epub-to-pdf.rb)
- [x]  [`maratis.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/maratis.rb)
- [x]  [`pencil.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/pencil.rb)
- [x]  [`quick-search-box.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/quick-search-box.rb)
- [x]  [`skreenics.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/skreenics.rb)

Asked to move hosting:
- [x]  [`duplicati.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/duplicati.rb) https://github.com/duplicati/duplicati/issues/1545
- [x]  [ `fotowall.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/fotowall.rb) https://github.com/enricoros/fotowall/issues/9
- [x]  [ `nethackcocoa.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/nethackcocoa.rb) https://github.com/dirkz/NetHack-Cocoa/issues/1
- [x]  [ `proximity.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/proximity.rb) https://github.com/revned/proximity/issues/11
- [x]  [ `quicklook-pfm`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/quicklook-pfm.rb) https://github.com/lnxbil/quicklook-pfm/issues/3
- [x]  [ `spek.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/spek.rb) https://github.com/alexkay/spek/issues/88
[...]

PR #17275, opened 2016-01-18
updated qlcolorcode (2.0.4)

tsparber:
This seems to be the most active fork. Original project seems to be abandoned. 

Tick #17075

### DIFF
--- a/Casks/qlcolorcode.rb
+++ b/Casks/qlcolorcode.rb
@@ -1,11 +1,13 @@
 cask 'qlcolorcode' do
-  version '2.0.2'
-  sha256 '317eda251ea5af8412401562395d2fbedb2dd915a7d927479cf09ac7251c4074'
+  version '2.0.4'
+  sha256 '6ae0b42c761181e493c94dcfabf22844859820055e77c780d92e56fc476fd9cb'
 
-  url "https://qlcolorcode.googlecode.com/files/QLColorCode-#{version}.tgz"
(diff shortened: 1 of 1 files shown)


Casks/qlcolorcode.rb
```diff
@@ -1,11 +1,13 @@
 cask 'qlcolorcode' do
-  version '2.0.2'
-  sha256 '317eda251ea5af8412401562395d2fbedb2dd915a7d927479cf09ac7251c4074'
+  version '2.0.4'
+  sha256 '6ae0b42c761181e493c94dcfabf22844859820055e77c780d92e56fc476fd9cb'
 
-  url "https://qlcolorcode.googlecode.com/files/QLColorCode-#{version}.tgz"
+  url 'https://github.com/anthonygelibert/QLColorCode/blob/master/bin/QLColorCode.qlgenerator.zip?raw=true'
```
jawshooah: Downloading directly from `master` seems like a really bad idea.

tsparber: I think so too. But this needs to be fixed upstream (I found no other download).

jawshooah: I see. Then at least use an actual commit sha (`7cfb048` is the latest) instead of `master`, which can change at any time.

tsparber: Good point. Amended

vitorgalvao:
Thank you for the contribution. It was merged directly as 939c6d0d75e024d00a8fb4283e3c806e43fa4019 to keep commit history cleaner. Your contribution is still credited to you.

PR #17284, opened 2016-01-18
Updated shelf-leveler

tsparber:


vitorgalvao:
Thank you, but fixing in another commit (it’s faster, since I’m doing all at once, and uses the previously mentioned `releasenotes` URL which is more correct.

PR #17288, opened 2016-01-18
Fixed WordPress Version

cedwardsmedia:
The official download link does not properly identify the app version in the URL. As such,as soon as a new version is available, the URL will point to it, causing a validation error when attempting to validate the checksum.

I changed the version to 'latest' and removed the sha256 check to correct this issue and prevent future breakage.

### DIFF
--- a/Casks/wordpresscom.rb
+++ b/Casks/wordpresscom.rb
@@ -1,6 +1,6 @@
 cask 'wordpresscom' do
-  version '1.2.1'
-  sha256 '1a10ba45287e6e3a6cc2c128aed4358e7e6efb1a624f80f6b9fd0cc71528489b'
+  version :latest
+  sha256 :no_check
 
   url 'https://public-api.wordpress.com/rest/v1.1/desktop/osx/download?type=dmg'
   appcast 'https://github.com/Automattic/wp-desktop/releases.atom',


Casks/wordpresscom.rb
```diff
@@ -1,6 +1,6 @@
 cask 'wordpresscom' do
-  version '1.2.1'
-  sha256 '1a10ba45287e6e3a6cc2c128aed4358e7e6efb1a624f80f6b9fd0cc71528489b'
+  version :latest
+  sha256 :no_check
 
   url 'https://public-api.wordpress.com/rest/v1.1/desktop/osx/download?type=dmg'
   appcast 'https://github.com/Automattic/wp-desktop/releases.atom',
```
tsparber: appcast could be changed to https://public-api.wordpress.com/rest/v1.1/desktop/osx/version?compare=0.1.0&channel=stable (From [updater/index.js#L33](https://github.com/Automattic/wp-desktop/blob/7ffef77d6ce41cf8bfa53fa4e9b5bac1144d0cb4/desktop/app-handlers/updater/index.js#L33)).
With the correct appcast, the `version` could then be '1.2.5' with a `sha256`.

tsparber: There is even a versioned url in the appcast :smiley:

vitorgalvao: @tsparber Care to submit that fix?

vitorgalvao:
@cedwardsmedia But it has an `appcast`, so we can take advantage of having a version, which is the whole point.

This will mitigates when we have automatic checks in place.

Casks/wordpresscom.rb
adidalal: @vitorgalvao `cask-repair` error in these cases:

```
 [.../homebrew-cask/Casks] > cask-repair wordpresscom.rb 
Type the new version (or leave blank to abort)
> 1.2.5
Paste the new URL (or leave blank to use the current one)
> https://public-api.wordpress.com/rest/v1.1/desktop/osx/download?type=app&ref=update&version=#{version}

Download URL returned 405
```

vitorgalvao: @adidalal Yes, noticed. On it.

vitorgalvao: @adidalal Fixed. It handles many cases way better now, showing clear warnings instead of outright exiting.

vitorgalvao: @tsparber You might also be interested in the future changes, as I’ll make it follow redirects for appcasts if the HTTP status code is `302`.

PR #17292, opened 2016-01-18
Added Instasizer

winkelsdorf:
Instasizer converts vector images to iOS and Android assets

### DIFF
--- a/Casks/instasizer.rb
+++ b/Casks/instasizer.rb
@@ -1,0 +1,13 @@
+cask 'instasizer' do
+  version :latest
+  sha256 :no_check
+  
+  url "http://www.tapgods.com/instasizer/Instasizer.zip"
+  name 'Instasizer'
+  homepage 'http://www.tapgods.com/instasizer/'
+  license :gratis
+
+  depends_on macos: '>= 10.7'
+
+  app 'Instasizer.app'
+end


victorpopkov:
Thank you for the contribution. It needed some fixes, so they were made in 6441594a6ff35cbeeabf336ab937cd453d1f5887. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

winkelsdorf:
As I tell all maintainers: You're welcome! 

Ah I see, missed the double quotation for url stanza this time.

By the way: In the contribution FAQ the 2nd example uses single quotation, too. Had this FAQ open during edit for reference and didn't scroll up to 1st example. See 2nd example (**gateblu**) of: https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#examples

Probably best to add a sentence that double quotation is preferred, to prevent a parsing error if ```#{version}``` is added to url in a future cask edit.

Cheers,
Frederik

jawshooah:
@winkelsdorf We actually do prefer single-quotes in cases where interpolation (e.g. `#{version}`) is not needed.

jawshooah:
It's also important to note that using double quotes with interpolated strings has nothing to do with Homebrew-Cask as a project. It's a Ruby language feature.

winkelsdorf:
@jawshooah Yeah, the latter is known and dully noted. 

To solve *why* I posted the above: I mixed up @victorpopkov and my commit - the change for https://github.com/caskroom/homebrew-cask/commit/6441594a6ff35cbeeabf336ab937cd453d1f5887 did not show what the changes to my PR were, so I had to guess by looking at it. 

I rechecked my local repo now: I used the double quotes without an interpolated string, not @victorpopkov.

Probably more a question of taste / best practice, comparable to something like

```
   if abc 
      dosomething()
  else 
      dosomethingelse()
```

while I prefer
```
  if abc {
     domesomething()
  } else {
    dosomethingelse()
  }
```

While in many languages both are correct, the first example is more error-prone if edited later. 

Same here, URL is always likely to be edited to include variables and become an interpolated string at some future point. So I decided intuitively to use double quotes.

Anyway, thanks for the clarification on what's preferred, Josh. Nothing complicated to keep in mind the next time! :)

PR #17321, opened 2016-01-19
Add Arturia software

joshka:
This includes casks for:
- Analog Lab
- Midi Control Center
- Mini V
- Software Center

I wasn't entirely sure whether this required one commit per product, or if you're happy with a single commit here. There are many more Arturia products available here, but as I don't have licenses for any of these, I can't check that they work etc.

Also note. Analog Lab and Mini V both have demo versions available with the following limitations

> Saving/Loading and import/export are disabled. The demo is limited to 20 minutes.

I wasn't sure whether this qualified the product as one where `license :freemium` applies or if what I put (`license: commercial`) was better here.

### DIFF
--- a/Casks/arturia-analog-lab.rb
+++ b/Casks/arturia-analog-lab.rb
@@ -1,0 +1,11 @@
+cask 'arturia-analog-lab' do
+  version '1.2.4.126'
+  sha256 'e66182ba1887d0c1369299c9c7ba6a06d709083b8aa6c8163108973d99b7e220'
+
+  url "https://downloads.arturia.com/products/analoglab/soft/Analog_Lab_#{version.dots_to_underscores}.pkg"
+  name 'Arturia Analog Lab'
+  homepage 'http://www.arturia.com/products/analog-classics/analoglab/overview'
+  license :commercial
+
+  pkg "Analog_Lab_#{version.dots_to_underscores}.pkg"
+end
--- a/Casks/arturia-midi-control-center.rb
+++ b/Casks/arturia-midi-control-center.rb
@@ -1,0 +1,11 @@
+cask 'arturia-midi-control-center' do
+  version '1.2.2.429'
+  sha256 'fba9e3e4835ece3c4a030c98a03c10051a0d8e3f103405843ee113e7f620892e'
+
+  url "https://downloads.arturia.com/extra/mcc/MIDI_Control_Center_#{version.dots_to_underscores}.pkg"
+  name 'Arturia Midi Control Center'
+  homepage 'http://www.arturia.com/support/downloads&manuals'
+  license :commercial
+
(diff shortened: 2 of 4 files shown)


jawshooah:
> I wasn't entirely sure whether this required one commit per product, or if you're happy with a single commit here.

We do prefer one commit per cask.

> I wasn't sure whether this qualified the product as one where license :freemium applies or if what I put (license: commercial) was better here.

The demo expires, so it's not really freemium. `:commercial` is fine.

Casks/arturia-analog-lab.rb
```diff
@@ -0,0 +1,11 @@
+cask 'arturia-analog-lab' do
```
jawshooah: Should just be `analog-lab` in keeping with the [token reference](https://github.com/caskroom/homebrew-cask/blob/73b9083f4d89faaf3815b8991cccafe55b550399/doc/cask_token_reference.md).

Casks/arturia-midi-control-center.rb
```diff
@@ -0,0 +1,11 @@
+cask 'arturia-midi-control-center' do
```
jawshooah: Should just be `midi-control-center` in keeping with the [token reference](https://github.com/caskroom/homebrew-cask/blob/73b9083f4d89faaf3815b8991cccafe55b550399/doc/cask_token_reference.md).

Casks/arturia-mini-v.rb
```diff
@@ -0,0 +1,11 @@
+cask 'arturia-mini-v' do
```
jawshooah: Should just be `mini-v` in keeping with the [token reference](https://github.com/caskroom/homebrew-cask/blob/73b9083f4d89faaf3815b8991cccafe55b550399/doc/cask_token_reference.md).

Casks/arturia-software-center.rb
```diff
@@ -0,0 +1,11 @@
+cask 'arturia-software-center' do
```
jawshooah: This one is fine, since "Arturia" is actually part of the artifact's name.

Casks/arturia-analog-lab.rb
joshka: I wasn't sure, however given that there is a lot of arturia software, I think it makes sense to prefix this. Is this a valid reason. See http://www.arturia.com/products

jawshooah: > Is this a valid reason?

Nope. The token reference is intentionally strict to prevent duplicates, which were a huge problem before it was implemented. These casks will still show up in `brew cask search arturia` though since the vendor is in the `name` stanza.

joshka: I'll fix tomorrow.

PR #17329, opened 2016-01-19
Updated duplicati (1.3.4)

tsparber:
Tick #17075

### DIFF
--- a/Casks/duplicati.rb
+++ b/Casks/duplicati.rb
@@ -2,8 +2,8 @@ cask 'duplicati' do
   version '1.3.4'
   sha256 'd97065d012fb929fa4e37b5efab2d246b6aef667f892a95301a1271f704963f2'
 
-  # googlecode.com is the official download host per the vendor homepage
-  url "https://duplicati.googlecode.com/files/Duplicati%20#{version}.dmg"
+  # storage.googleapis.com is the official download host per the vendor homepage
+  url "https://storage.googleapis.com/google-code-archive-downloads/v1/code.google.com/duplicati/Duplicati%20#{version}.dmg"
   name 'Duplicati'
   homepage 'http://www.duplicati.com/'
   license :oss


Casks/duplicati.rb
```diff
@@ -2,10 +2,10 @@
   version '1.3.4'
   sha256 'd97065d012fb929fa4e37b5efab2d246b6aef667f892a95301a1271f704963f2'
 
-  # googlecode.com is the official download host per the vendor homepage
-  url "https://duplicati.googlecode.com/files/Duplicati%20#{version}.dmg"
+  # storage.googleapis.com is the official download host per the vendor homepage
+  url "https://storage.googleapis.com/google-code-archive-downloads/v1/code.google.com/duplicati/Duplicati%20#{version}.dmg"
```
tsparber: I'm no sure about the comment.

tsparber:
Google maybe has a problem with the spaces. Waiting a bit on https://github.com/duplicati/duplicati/issues/1545#issuecomment-172825070 or will suggest deletion of the cask otherwise.

Casks/duplicati.rb
vitorgalvao: Think of the rule this way: if the link provided on the homepage redirects you to a new domain, it’s still OK to have the comment.

I do know there are cases where the comment can be a bit weirder, so I am thinking of a new substitute comment that’ll make everything clearer and be possible to use in more cases.

Casks/duplicati.rb
```diff
@@ -2,10 +2,10 @@
...
   name 'Duplicati'
-  homepage 'http://www.duplicati.com/'
+  homepage 'http://www.duplicati.com'
```
vitorgalvao: We actually prefer the trailing `/`, as it’s the proper way to present a URL.

tsparber:
Closing in favor of #17338

PR #17340, opened 2016-01-19
Adding cask for Sublime Text 3

cadam11:
There are two other casks for the Sublime text editor, but none of them are set up to use version 3. Version 3 has been in beta over over two years and is separate enough from its older versions to be given its own cask.


adidalal:
Already in /versions

https://github.com/caskroom/homebrew-versions/blob/master/Casks/sublime-text3.rb

(brew cask install caskroom/versions/sublime-text3)

cadam11:
I see, thanks

PR #17351, opened 2016-01-19
initial document breaks

### LINKED ISSUE #16690: Proposal: break up documentation files into directories and files
I occasionally check [electron’s API documentation](https://github.com/atom/electron/tree/master/docs/api), and it’s a breeze. Not only can I go to that particular URL and immediately find a file that has *only* what I need (which also means <kbd>⌘</kbd><kbd>F</kbd> is much more useful), I can actually easily google for it.

Contrast that with our documentation, in which we have humongous files with an index at the top, and we need anchors to navigate.

My proposal: break it up. Specifically (and at least) our most important pieces of documentation: `CONTRIBUTING` and `CASK_LANGUAGE_REFERENCE`. Look at [electron’s documentation structure as a whole](https://github.com/atom/electron/tree/master/docs), and you’ll see how clear and easy to navigate it is.

Unless there are objections, I intend to start taking a crack at it.

vitorgalvao:
Closes https://github.com/caskroom/homebrew-cask/issues/16690.

Please concentrate at this stage on the structure. What matters now is that we have a structure we agree on so we can merge this and @adidalal is no longer stuck on https://github.com/caskroom/homebrew-cask/pull/16777.

After that, improvements to the documentation itself can be made in subsequent PRs.

After the structure is agreed on, I’ll add an extra commit fixing broken links.

See https://github.com/vitorgalvao/homebrew-cask/tree/docs-break/doc for navigation.


jawshooah:
This is great! My only suggestion related to structure would be to rename `synopsis.md` in `cask_language_reference` to `readme.md` so that it's displayed when viewing the contents of `cask_language_reference`.

adidalal:
Looks good - there are some things that could be changed around but they are minor and can be progressively addressed.

Good to merge, from my end.

vitorgalvao:
Thank you both. Renamed `synopsis.md` and fixed links. Merging now.

PR #17354, opened 2016-01-19
added jgrasp v2.0.1_09

sbdchd:
I added jgrasp version 2.0.0_14, but since the [download](http://spider.eng.auburn.edu/user-cgi/grasp/grasp.pl?;dl=download_jgrasp.html) from the project's website requires a browser, I am using a [mirror](https://github.com/butangmucat/aur-file-mirrors/blob/master/jgrasp200_14.zip).

### DIFF
--- a/Casks/jgrasp.rb
+++ b/Casks/jgrasp.rb
@@ -1,0 +1,13 @@
+cask 'jgrasp' do
+  version '2.0.1_09'
+  sha256 '3b16bcc366a06d0a7765801814e02bd94837ee5e2e141ad0d5855ae4bd5b04ba'
+
+  url "http://www.jgrasp.org/dl4g/jgrasp/jgrasp#{version.no_dots}.pkg"
+  name 'jgrasp'
+  homepage 'http://jgrasp.org/index.html'
+  license :unknown # TODO: change license and remove this comment; ':unknown' is a machine-generated placeholder
+
+  pkg "jgrasp#{version.no_dots}.pkg"
+
+  uninstall pkgutil: 'org.jgrasp.jgrasp.jgrasp.pkg'
+end


Casks/jgrasp.rb
```diff
@@ -0,0 +1,11 @@
+cask 'jgrasp' do
```
vitorgalvao: This has to be changed to `cask 'jcs' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `jcs.rb`.

Casks/jgrasp.rb
```diff
@@ -0,0 +1,11 @@
...
+  version '2.0.0_14'
+  sha256 '3849d871262264736b3e7059650825a8f6ec489e7009adabd090b4df772882d6'
+
+  url 'https://github.com/butangmucat/aur-file-mirrors/raw/master/jgrasp200_14.zip'
+  name 'jgrasp'
+  homepage 'http://jgrasp.org/index.html'
+  license :unknown # TODO: change license and remove this comment; ':unknown' is a machine-generated placeholder
+
+  app 'jgrasp/bin/osx104/jcs.app', target: 'jgrasp.app'
```
vitorgalvao: [Not a good case for renaming](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#target-should-only-be-used-in-select-cases).

vitorgalvao:
> but since the download from the project's website requires a browser

Why?

sbdchd:
I took your advice and I was able to find the download link, which was obscured by a redirect.

I think with this updated formula a rename is no longer necessary.

Casks/jgrasp.rb
```diff
@@ -0,0 +1,11 @@
+cask 'jgrasp' do
+  version '2.0.1_09'
+  sha256 '3b16bcc366a06d0a7765801814e02bd94837ee5e2e141ad0d5855ae4bd5b04ba'
+
+  url "http://www.jgrasp.org/dl4g/jgrasp/jgrasp#{version.delete('.')}.pkg"
+  name 'jgrasp'
+  homepage 'http://jgrasp.org/index.html'
+  license :unknown # TODO: change license and remove this comment; ':unknown' is a machine-generated placeholder
+
+  pkg "jgrasp#{version.delete('.')}.pkg"
```
vitorgalvao: Use `#{version.no_dots}`, instead.

Casks/jgrasp.rb
```diff
@@ -0,0 +1,11 @@
+cask 'jgrasp' do
+  version '2.0.1_09'
+  sha256 '3b16bcc366a06d0a7765801814e02bd94837ee5e2e141ad0d5855ae4bd5b04ba'
+
+  url "http://www.jgrasp.org/dl4g/jgrasp/jgrasp#{version.delete('.')}.pkg"
```
vitorgalvao: Use `#{version.no_dots}`, instead.

vitorgalvao:
It’s missing `uninstall`. There are [scripts](https://github.com/caskroom/homebrew-cask/tree/master/developer/bin) and [documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#uninstall-stanza-details) geared for that.

Also, please [squash your commits](http://davidwalsh.name/squash-commits-git).

vitorgalvao:
Thank you for the contribution. It was merged directly as 7a04097efd60e6cb7a01c828d17b2b76bace8ec4 to keep commit history cleaner. Your contribution is still credited to you.

sbdchd:
Thanks for all the help and guidance!

PR #17361, opened 2016-01-19
Updated qlcolorcode (2.0.4)

tsparber:


Casks/qlcolorcode.rb
```diff
@@ -1,13 +1,21 @@
...
-  url 'https://github.com/anthonygelibert/QLColorCode/blob/7cfb0487a9804474d73a85317846cb8fd1a370de/bin/QLColorCode.qlgenerator.zip?raw=true'
+  url "https://github.com/anthonygelibert/QLColorCode/releases/download/release-#{version}/QLColorCode.qlgenerator.zip"
   appcast 'https://github.com/anthonygelibert/QLColorCode/releases.atom',
-          checkpoint: '143ecf2eb923c9608874fe43382c52d3979a7475378d605482e9ce5b2ab50dbf'
+          checkpoint: 'be9c289ef615ab6006674f775bad857874e1c9c48e35c3e58f302ca2fa13a315'
   name 'QLColorCode'
   homepage 'https://github.com/anthonygelibert/QLColorCode'
   license :oss
 
+  depends_on formula: 'highlight'
```
jawshooah: Could you explain this dependency, along with the postflight block?

anthonygelibert: Maybe I can explain you…

In the original version QLColorCode embedded a pre-compiled version of “highlight” (used to generate the preview, coloring…). It’s no longer the case in my own version. 

“highlight” being an external dependency, it can be installed everywhere, so I gave a variable to specify its path. The default value is /opt/local/bin/highlight.

tsparber: Should i add a comment in the cask?
The `defaults write` sets then the path to the installed `highlight` formula.

jawshooah: Got it, that makes sense. A brief comment explaining the purpose of the `defaults write` would be nice in the `postflight` block.

Casks/qlcolorcode.rb
```diff
@@ -1,13 +1,21 @@
...
+
   qlplugin 'QLColorCode.qlgenerator'
+
+  postflight do
+    system 'defaults write org.n8gray.QLColorCode pathHL $(which highlight)'
```
jawshooah: Also, please enclose `$(which highlight)` in double-quotes to prevent arg-splitting in case `brew --prefix` has spaces.

Casks/qlcolorcode.rb
tsparber: amended

anthonygelibert:
Hi @tsparber and @jawshooah, could you update the cask to the latest version of qlcolorcode please?

PR #17371, opened 2016-01-20
Update HipChat version

timlo:
Version bump for hipchat


vitorgalvao:

timlo:
Thanks, should've re-read the docs. Made the update.

vitorgalvao:
Thank you for the contribution. It was merged directly as e358b11c934c1afaa4a0d1839f612b49b6fccb71 to keep commit history cleaner. Your contribution is still credited to you.

PR #17381, opened 2016-01-20
Update Transporter Desktop to version 4.2.3_20317

mauricerkelly:
This commit updates the version and sha256 stanzas.


vitorgalvao:

mauricerkelly:
I’m kinda confused now - I started into doing this, and I’ve noticed that the Cask is already up to date with the Transporter version in master, but my PR is still showing a diff. I think this one should be closed, but I’ve made note to update the `:checkpoint` value in future! (Been a while since I’ve been updating outside of `homebrew-versions`!)

Issue #16685, opened 2016-01-10
Add appcast to sourceforge casks


vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/pull/16683#issuecomment-170348241.

Sourceforge projects seem to have simple and predictable appcasts, just like github projects: `http://downloads.sourceforge.net/projects/{{project_name}}/rss`. A more specific rss can be used as needed, pointing to a specific directory structure: `http://sourceforge.net/projects/{{project_name}}/rss?path=/{{path_here}}`.

---

List of casks in [caskroom/cask](https://github.com/caskroom/homebrew-cask) missing `appcast`

```bash
for cask in $(grep --files-with-matches "url ['\"].*sourceforge.*" "$(brew --repository)/Library/Taps/caskroom/homebrew-cask/Casks/"*); do
  appcast="$(brew cask _stanza appcast "${cask}")"
  if [[ -z "${appcast}" ]]; then
    cask_filename="$(basename "${cask}")"
    cask_name="${cask_filename%.*}"
    echo "- [ ] [${cask_name}](../blob/master/Casks/${cask_filename})"
  fi
done | pbcopy
```

- [ ] [aquaterm](../blob/master/Casks/aquaterm.rb)
- [ ] [aria-maestosa](../blob/master/Casks/aria-maestosa.rb)
- [ ] [arrsync](../blob/master/Casks/arrsync.rb)
- [ ] [art-of-illusion](../blob/master/Casks/art-of-illusion.rb)
- [ ] [audioslicer](../blob/master/Casks/audioslicer.rb)
- [ ] [avidemux](../blob/master/Casks/avidemux.rb)
- [ ] [avogadro](../blob/master/Casks/avogadro.rb)
- [ ] [bitpim](../blob/master/Casks/bitpim.rb)
- [ ] [brain-workshop](../blob/master/Casks/brain-workshop.rb)
- [ ] [breakaway](../blob/master/Casks/breakaway.rb)
- [ ] [brushviewql](../blob/master/Casks/brushviewql.rb)
- [ ] [burn](../blob/master/Casks/burn.rb)
- [ ] [butt](../blob/master/Casks/butt.rb)
- [ ] [camed](../blob/master/Casks/camed.rb)
- [ ] [celestia](../blob/master/Casks/celestia.rb)
- [ ] [chessx](../blob/master/Casks/chessx.rb)
- [ ] [codeblocks](../blob/master/Casks/codeblocks.rb)
[...]

PR #17394, opened 2016-01-21
Add appcast to grandperspective

tsparber:
Tick #16685

### DIFF
--- a/Casks/grandperspective.rb
+++ b/Casks/grandperspective.rb
@@ -2,12 +2,14 @@ cask 'grandperspective' do
   version '1.5.1'
   sha256 '92204458042a337c1091879e167ea95e45cae33a7be16fa6c11e80572c54d135'
 
-  url "http://downloads.sourceforge.net/project/grandperspectiv/grandperspective/#{version}/GrandPerspective-#{version.gsub('.', '_')}.dmg"
+  url "http://downloads.sourceforge.net/project/grandperspectiv/grandperspective/#{version}/GrandPerspective-#{version.dots_to_underscores}.dmg"
+  appcast 'http://sourceforge.net/projects/grandperspectiv/rss?path=/grandperspective',
+          checkpoint: 'aa3a708f08a2b7f02fd269062bdda983d17c36007bb5fe46b1e11ab3dea07769'
   name 'GrandPerspective'
   homepage 'http://grandperspectiv.sourceforge.net/'
   license :gpl
 
   app 'GrandPerspective.app'
 
-  zap delete: '~/Library/Preferences/net.sourceforge.grandperspectiv.plist'
+  zap trash: '~/Library/Preferences/net.sourceforge.grandperspectiv.plist'
 end


Casks/grandperspective.rb
```diff
@@ -2,12 +2,14 @@
...
+  url "http://downloads.sourceforge.net/project/grandperspectiv/grandperspective/#{version}/GrandPerspective-#{version.dots_to_underscores}.dmg"
+  appcast 'http://sourceforge.net/projects/grandperspectiv/rss?path=/grandperspective',
+          checkpoint: 'aa3a708f08a2b7f02fd269062bdda983d17c36007bb5fe46b1e11ab3dea07769'
   name 'GrandPerspective'
   homepage 'http://grandperspectiv.sourceforge.net/'
   license :gpl
 
   app 'GrandPerspective.app'
 
-  zap delete: '~/Library/Preferences/net.sourceforge.grandperspectiv.plist'
```
jawshooah: Why the change here? `:trash` and `:delete` are synonyms. We don't yet support moving things to the trash bin.

tsparber: I thought there were some discussions about using `trash:` for files which are maybe still important to the user (even when zapping) and `delete:` for files which are automatically created and "not so important". (I'm aware that its an alias). (like `trash` for preferences, delete for `caches...`). I wouldn't mind reverting this change.

adidalal:
`:trash` could be as simple as implementing this bash function `trash () { command mv "$@" ~/.Trash ; }`

Casks/grandperspective.rb
adidalal: `:trash` could be as simple as implementing this bash function `trash () { command mv "$@" ~/.Trash ; }`

tsparber: This could be as simple as that, but upon reading the interwebs, I found that such a simple approach would have some drawbacks. Misses handling of duplicate files, Finders "put back" option is also not available. I think the "put back" option would be one of the most important features (The user most likely didn't know where the file was before).
Using [NSFileManager trashItemAtURL](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSFileManager_Class/#//apple_ref/occ/instm/NSFileManager/trashItemAtURL:resultingItemURL:error:) seems to be a better approach. I don't know how hard it is to call Objective-C functions from Ruby, but this (or some wrapper) seems to be the cleanest approach.

jawshooah: @adidalal [It's not quite that simple.](https://github.com/morgant/tools-osx/blob/master/src/trash)

vitorgalvao:
Why the close?

tsparber:
My bad, was cleaning up my branches from command line. I will double check next time.

Casks/grandperspective.rb
vitorgalvao: Should we merge this, or change it back to `delete`? I’m fine either way.

jawshooah:
Doesn't matter I suppose. Not worth holding up the merge.

bazooka241:
Iｍ sorry . I can’t understand English.
So Im calling trancerater.
Thank you very much.

> H28/01/24 1:48、Josh Hagins <notifications@github.com> のメール：
> 
> oesn't matter I suppose. Not worth holding up the merge

PR #17413, opened 2016-01-22
Removed smlnj

jond3k:
This cask installs the [graphical installer](http://smlnj.cs.uchicago.edu/dist/working/110.79/index.html#macos-installer), which requires root and installs smlnj to /usr/local. The homebrew formula is preferable since it uses the unix command line installation process which installs to Cellar.

This is part of https://github.com/caskroom/homebrew-cask/issues/15603


victorpopkov:
Thank you for the contribution. It was merged directly as 005c8d81d7f329379e5c9f1ef9777dc170894ab6 to keep commit history cleaner. Your contribution is still credited to you.

PR #17419, opened 2016-01-22
Fix download link for Framer Studio.app :latest

CameronGarrett:
Update to static download link and remove unsupported appcast per conversation with Koen Bok of Framer https://twitter.com/koenbok/status/690296760783278080

### DIFF
--- a/Casks/framer-studio.rb
+++ b/Casks/framer-studio.rb
@@ -1,10 +1,9 @@
 cask 'framer-studio' do
-  version '1.11.188'
-  sha256 'b1fff971456a4af5ad88661344dbad079d5ba08e25dfc2c6f252824d1916206b'
+  version '49'
+  sha256 '3325beca4aeb52b48eac89226795e46eb217e2fdae85e157612b3f0bd67a680a'
 
-  url "http://studio.update.framerjs.com/static/downloads/Framer%20Studio-#{version}.zip"
-  appcast 'http://studio.update.framerjs.com/appcast.xml',
-          checkpoint: 'e76c4383947ebe76cd1e16c19f2f87c8d4e60c13ade2a5ae591267c7155632da'
+  # devmate.com is the official download host per the appcast feed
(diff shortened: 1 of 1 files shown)


adidalal:
If there's any updated appcast at all, even if the links are to delta updates, we can use it - that file content changing gives us a basis to detect updates.

Casks/framer-studio.rb
```diff
@@ -1,10 +1,8 @@
 cask 'framer-studio' do
-  version '1.11.188'
-  sha256 'b1fff971456a4af5ad88661344dbad079d5ba08e25dfc2c6f252824d1916206b'
+  version :latest
+  sha256 :no_check
 
-  url "http://studio.update.framerjs.com/static/downloads/Framer%20Studio-#{version}.zip"
-  appcast 'http://studio.update.framerjs.com/appcast.xml',
-          checkpoint: 'e76c4383947ebe76cd1e16c19f2f87c8d4e60c13ade2a5ae591267c7155632da'
+  url "https://dl.devmate.com/com.motif.framer/FramerStudio.zip"
```
adidalal: Needs a url comment

CameronGarrett:
@adidalal the appcast appears to be very old, should it be left even if it isn't currently updated?

```<pubDate>Wed, 25 Dec 2005 12:20:11 +0000</pubDate>```

adidalal:
No, but if there is an up to date changelog page that can be used as well

adidalal:
Comment should be `# devmate.com is the official download host per the appcast feed`

CameronGarrett:
Will https://rink.hockeyapp.net/apps/cf6ed17136c869e3604eb59fc900515f do for the changelog? I can't tell if the address is static

adidalal:
Looks like that identifier is per app, so it should be fine, yes. You should also put in an actual version and shasum, even though the download URL is unversioned.

Issue #17388, opened 2016-01-21
Proposal: Add brew cask install/remove to travis script


tsparber:
What do you think about including a `brew cask install/remove` step after the `audit`/`style` check [ci/travis/script.sh#L19](https://github.com/caskroom/homebrew-cask/blob/master/ci/travis/script.sh#L19) in travis?

(I don't see any security reasons, it is in a VM after all and creating other 'malicious' PRs wouldn't be that hard).

I think this could detect issues with wrong artifacts (example updates packages has now an additional subfolder).
After the install, travis could then also test the removal.

This wouldn't catch all possible issues, but maybe won't hurt either. If using `sudo` is a problem it could be limited to those artifacts which do not require `sudo`.

tsparber:
Closing in favor of #17424 (already has a discussion)

PR #17424, opened 2016-01-22
Add brew cask install/remove to travis

tsparber:
Refs #17388 

Example build with multiple casks:
https://travis-ci.org/tsparber/homebrew-cask/jobs/104196142

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: objective-c
 
-sudo: false
-
 env:
   global:
     - LANG=en_US.UTF-8
--- a/ci/travis/script.sh
+++ b/ci/travis/script.sh
@@ -16,6 +16,8 @@ if any_casks_modified; then
   modified_casks=($(modified_cask_files))
   run developer/bin/audit_modified_casks "${TRAVIS_COMMIT_RANGE}"
   run brew cask style "${modified_casks[@]}"
+  run brew cask install "${modified_casks[@]}"
+  run brew cask zap "${modified_casks[@]}"
 fi
 
 if must_run_tests; then


jawshooah:
We recently spent a considerable amount of effort to speeding up Travis builds for cask PRs. This would negate much of that, especially since we don't even bother downloading casks whose `version`, `sha256`, or `url` haven't changed.

For single-cask PRs, though, it wouldn't be much of an issue. For example:

```bash
if [[ "${#modified_casks[@]}" == 1 ]]; then
  for cask in "${modified_casks[@]}"; do
    run brew cask install "$cask"
    run brew cask zap "$cask" # zap implicitly calls uninstall, so we can test both
  done
fi
```

jawshooah:
We'd also need to remove `sudo: false` from `.travis.yml`, since sudo is required to install casks that use `pkg` or `installer :script`.

At that point, we may as well remove much of `before_install.sh`, since we could just install gems to the system path.

The drawback, of course, is that we could no longer (eventually) take advantage of Travis' cache function, which requires `sudo: false`.

tsparber:
> We recently spent a considerable amount of effort to speeding up Travis builds for cask PRs.

These are really great improvements :+1: and I think these lay a base to add `install/zap`.
The thing I think adding this could be helpful, is because the goal (as far as I understand) of the project is automated install of apps (So testing the installability seems to be important).

> For single-cask PRs, though, it wouldn't be much of an issue. For example:

I encountered sometimes PRs where only one cask was changed but travis tested multiple casks (I'm not certain if this is still the case).

> The drawback, of course, is that we could no longer (eventually) take advantage of Travis' cache function, which requires sudo: false.

I'm not sure if this is a real requirement. As far as I understand travis docs, [caching](https://docs.travis-ci.com/user/caching/) is only supported on their [container-based infrastructure](https://docs.travis-ci.com/user/workers/container-based-infrastructure/). To use this infrastructure a `sudo: false` is needed, as they not not allow `sudo` in a docker image for (security?) [reasons](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/). As far as I understand docker there is (currently) no way OSX [could run](http://serverfault.com/a/607717) in a/as a docker image. Please correct me if I'm wrong :smiley:

vitorgalvao:
Agreed that losing the speedups is undesirable, but also that doing this for single-cask PRs only would mitigate that. However, even if Travis’ cache function is in the future (or us being able to take advantage of it is, to be precise), doing things with that in mind is good. Also, although install errors to modified casks do happen, they’re not all that common. The ones that don’t require sudo are also the most benign and easy to fix (basically a wrong `app`).

So to incorporate this we need to either sacrifice a future desirable feature (Travis’ cache function), or make this one crippled (not work on cases that require sudo). Question is, which feature do we want the most.

In light of @tsparber’s last comment, though, perhaps the question is “will we ever be able to ever take advantage of Travis’ cache function at all?”. I was under the impression I read somewhere that is indeed being worked on, but I cannot recall if I read that from any official source.

vitorgalvao:
Leaving this one to whatever you find best, @jawshooah.

tsparber:
I'm ok with any outcome, it's up to you (the maintainers) anyways :smiley:.

Just some things some things I considered before suggesting the addition (and things mentioned in the discussion):
* This could be reverted if travis implements caching and `sudo: false` is really needed [1] to make it work (the changes are not that big).
* This would remove the need to do a local install after a simple version update (Some authors do not package always the same :frowning:) (This would maybe move some testing work off of the maintainers to the "cloud"/automation)
* Added test runtime: I think this could only be an issue for "mass updatings", for most casks, I think, the benefits would outweigh it.
* The tests for style/audit are also nice and important (to have a clean codebase :+1:), but testing "what the user does" (installing/removing) seems also important (especially that it is not so "expensive" [2]) :smiley: 

--

[1] As stated above, I don't think this will be a requirement for OS X hosts. `sudo: false` is a requirement to use the docker based infrastructure (Linux only), which is the only platform where they implemented the caching functionality yet, but caching is not otherwise related to `sudo: false`. This is what I understand from the provided documentation.

[2] For already fetched casks (due to audit) executing the app symlink (+`*flight`) is quite fast. (`pkg` are slower)

jawshooah:
From [travis-ci/travis-ci#4011 (comment)](https://github.com/travis-ci/travis-ci/issues/4011#issuecomment-120001089):

> caching on the Mac for public repositories is not yet available. We want to enable it, though; so we will keep this issue open.

So it sounds like they *do* plan to support caching for Mac environments in the future. Still unclear whether or not that will require `sudo: false`, but I think it's safer to design our build for that case.

How many issues would we really be catching by adding `brew install/uninstall` to the build anyway? The most common problem users run into is that an artifact has changed its name between versions, causing the symlink step to fail, but even that isn't all that common.

vitorgalvao:
Seems like not merging is the decision for now. Closing.

PR #17425, opened 2016-01-22
updated bettertouchtool (1.52b)

tsparber:


Casks/bettertouchtool.rb
```diff
@@ -5,18 +5,21 @@
 
     # bettertouchtool.com is the official download host per the vendor homepage
     url "https://bettertouchtool.net/btt#{version}.zip"
+
+    license :gratis
```
tsparber: [BTT FAQ](http://www.boastr.net/btt-paid-faq/):
> Can I continue to use old versions of BetterTouchTool for free?
> No. Old versions will soon expire. Only the old BeterTouchTool versions for OS X 10.6, which don’t receive updates anymore, will continue to be free.

vitorgalvao: Nice preemptive explanation. Thank you.

PR #17449, opened 2016-01-23
updated nylas-n1 (0.3.45-3b5880e)

adidalal:
appcast from https://github.com/nylas/N1/blob/master/src/browser/auto-update-manager.coffee

Ref: https://github.com/nylas/N1/issues/1124


adidalal:
Merged as bfcd7cebfb3d07d7bb89e26e8194395ccfbf61ee.

Issue #17454, opened 2016-01-24
Error: strongvpn-client: It seems the symlink source is not there


twang2218:
When I try to run `brew cask install strongvpn-client --verbose`, the command returns following error:

```ruby
==> Downloading https://mirror2.reliablehosting.com/mac/StrongVPN_Mac_1.1.8.zip
Already downloaded: /Library/Caches/Homebrew/strongvpn-client-1.1.8.zip
==> Verifying checksum for Cask strongvpn-client
Error: It seems the symlink source is not there: '/opt/homebrew-cask/Caskroom/strongvpn-client/1.1.8/StrongVPN Client.app'
Error: Kernel.exit
```

Here is the output of `brew cask doctor`:

```ruby
brew cask doctor
==> OS X Release:
10.11
==> OS X Release with Patchlevel:
10.11.2
==> Hardware Architecture:
intel-64
==> Ruby Version:
2.0.0-p645
==> Ruby Path:
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby
==> Homebrew Version:
Homebrew 0.9.5 (git revision 839bb; last commit 2016-01-23)
==> Homebrew Executable Path:
/usr/local/bin/brew
==> Homebrew Cellar Path:
/usr/local/Cellar
==> Homebrew Repository Path:
/usr/local
==> Homebrew Origin:
https://github.com/Homebrew/homebrew
==> Homebrew-cask Version:
0.60.0 (git revision 4034; last commit 8 hours ago)
==> Homebrew-cask Install Location:
<NONE>
==> Homebrew-cask Staging Location:
/opt/homebrew-cask/Caskroom
==> Homebrew-cask Cached Downloads:
/Library/Caches/Homebrew
/Library/Caches/Homebrew/Casks
18 files, 1.8G (warning: run "brew cask cleanup")
==> Homebrew-cask Default Tap Path:
/usr/local/Library/Taps/caskroom/homebrew-cask
==> Homebrew-cask Alternate Cask Taps:
<NONE>
==> Homebrew-cask Default Tap Cask Count:
2945
==> Contents of $LOAD_PATH:
/usr/local/Library/Taps/caskroom/homebrew-cask/lib
/usr/local/Library/Homebrew
/Library/Ruby/Site/2.0.0
/Library/Ruby/Site/2.0.0/x86_64-darwin15
/Library/Ruby/Site/2.0.0/universal-darwin15
/Library/Ruby/Site
[...]

PR #17455, opened 2016-01-24
strongvpn-client.rb: update the app name

twang2218:
As mentioned in #17454 , symbol link will fail in the current version. After check the downloaded .zip file, I found the name of app changed, from `StrongVPN Client.app` to `StrongVPN.app`, I think that is the reason of the bug. This PR fixed that.


victorpopkov:
Thank you for the contribution. It was merged directly as 34bd2d5cc17acb8a088ac8e404e0d35104890f98 to keep commit history cleaner. Your contribution is still credited to you.

PR #17456, opened 2016-01-24
Add scan-build cask build 277

mgrebenets:
Home page: http://clang-analyzer.llvm.org/scan-build.html
Relates to version 0.277, but that's not very clear


vitorgalvao:
Could this be added to homebrew instead? In light of https://github.com/caskroom/homebrew-cask/issues/15603, it seems like the best place.

vitorgalvao:
Closing. Should be submitted to homebrew.

PR #17464, opened 2016-01-24
Add VNC Viewer.app v5.3.0

viniciusbig:
Adding new cask for the Real VNC Client.

### DIFF
--- a/Casks/vnc-viewer.rb
+++ b/Casks/vnc-viewer.rb
@@ -1,0 +1,11 @@
+cask 'vnc-viewer' do
+  version '5.3.0'
+  sha256 '1995208a735bfc186efbf899755c56fc6f2f757832e66601c7d105dae5a2726b'
+
+  url 'https://www.realvnc.com/download/binary/1797/'
+  name 'Real VNC Viewer'
+  homepage 'https://www.realvnc.com'
+  license :freemium
+
+  app 'VNC Viewer.app'
+end


victorpopkov:
Thank you for the contribution. However, this cask already exists: [real-vnc.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/real-vnc.rb).

viniciusbig:
Thanks for the quick reply.

Actually this is a different cask.

real-vnc.rb is the full version of Real VNC witch includes a VNC server.
My cask is just the VNC Viewer. A simpler program, smaller, just to connect in a VNC Server.

real-vnc.rb refers to this link https://www.realvnc.com/download/vnc/latest/
real-vnc-viewer.rb refers to https://www.realvnc.com/download/viewer/

victorpopkov:
Thank you for the explanations. Everything looks good. However, the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) should be followed to avoid duplicates. According to it, the cask should be named as `vnc-viewer.rb`. But in my opinion it would be better to leave the name as it now.

What do you think @vitorgalvao about the cask naming for this case?

vitorgalvao:
`vnc-viewer` is the correct name. We can’t take “it makes sense in this case” as a reason, because that is exactly how duplicates happen. What makes sense for a person is different than what makes sense to another. The token reference is strict for that reason, we need an absolute answer. `name` is the place to be correct and verbose (and `name` for this cask is correct).

viniciusbig:
I don't know if I did correctly, but I fixed the file name to match token requirements.

vitorgalvao:
It is correct, yes, thank you. Could you please [squash your commits](http://davidwalsh.name/squash-commits-git)?

victorpopkov:
> We can’t take “it makes sense in this case” as a reason, because that is exactly how duplicates happen.

Fully agree. Just wanted to make sure. Thank you @vitorgalvao for the quick response.

victorpopkov:
Thank you for the contribution. It was merged directly as 3d31b4510246a452f416a95dde3d95548e538fe9 to keep commit history cleaner. Your contribution is still credited to you.

PR #17485, opened 2016-01-25
pdated robomongo to 0.9.0-RC1

ghost:


adidalal:
This is an RC version - put this in homebrew-versions. Also needs a comment about the url (see documentation)

julienlavergne:
okay ... we will wait for the release.

PR #17488, opened 2016-01-25
vox.rb: update to the build 2710.1, fix installation on case-sensitiv…

samm-git:
…e filesystems


vitorgalvao:
Thank you for the contribution. It was merged directly as 0a3d2a1e692bd377905e5e3a3af2af6e2050f831 to keep commit history cleaner. Your contribution is still credited to you.

PR #17494, opened 2016-01-26
Upgrade Github Atom to v1.4.1

burnsra:


victorpopkov:
Thank you for the contribution. It was merged directly as 0735ce7e82da9ab571df247c5c779c3d649ba3bc to keep commit history cleaner. Your contribution is still credited to you.

vitorgalvao:

For this one, since it was already merged, I’ll fix in a different commit.

Issue #17498, opened 2016-01-26
Verify `appcast`s with hostname different from `homepage`, and add comment


vitorgalvao:
Refs #17499.

Casks that need it:

In [caskroom/cask](https://github.com/caskroom/homebrew-cask):

+ [ ] [actotracker](https://github.com/caskroom/homebrew-cask/tree/master/Casks/actotracker.rb)
+ [ ] [aether](https://github.com/caskroom/homebrew-cask/tree/master/Casks/aether.rb)
+ [x] [air-video-server-hd](https://github.com/caskroom/homebrew-cask/tree/master/Casks/air-video-server-hd.rb)
+ [x] [airfile](https://github.com/caskroom/homebrew-cask/tree/master/Casks/airfile.rb)
+ [ ] [alcatraz](https://github.com/caskroom/homebrew-cask/tree/master/Casks/alcatraz.rb)
+ [x] [aliwangwang](https://github.com/caskroom/homebrew-cask/tree/master/Casks/aliwangwang.rb)
+ [ ] [aluxian-messenger](https://github.com/caskroom/homebrew-cask/tree/master/Casks/aluxian-messenger.rb)
+ [ ] [angry-ip-scanner](https://github.com/caskroom/homebrew-cask/tree/master/Casks/angry-ip-scanner.rb)
+ [x] [antetype](https://github.com/caskroom/homebrew-cask/tree/master/Casks/antetype.rb)
+ [x] [anvil](https://github.com/caskroom/homebrew-cask/tree/master/Casks/anvil.rb)
+ [ ] [apm-planner](https://github.com/caskroom/homebrew-cask/tree/master/Casks/apm-planner.rb)
+ [ ] [appserver](https://github.com/caskroom/homebrew-cask/tree/master/Casks/appserver.rb)
+ [ ] [aptanastudio](https://github.com/caskroom/homebrew-cask/tree/master/Casks/aptanastudio.rb)
+ [ ] [aquamacs](https://github.com/caskroom/homebrew-cask/tree/master/Casks/aquamacs.rb)
+ [x] [atom](https://github.com/caskroom/homebrew-cask/tree/master/Casks/atom.rb)
+ [ ] [audio-hijack](https://github.com/caskroom/homebrew-cask/tree/master/Casks/audio-hijack.rb)
+ [x] [audiomate](https://github.com/caskroom/homebrew-cask/tree/master/Casks/audiomate.rb)
[...]

PR #17499, opened 2016-01-26
Updates to comments of diverging hostnames

vitorgalvao:
@caskroom/maintainers Need some feedback.
- Slightly changed the rule for `url`.
- Added a rule to add this comment to `appcast` (see also #17498).

Reasons for the rule change to `url`:
- Only one comment is needed (instead of two: `# URL_HOSTNAME is the official download host per the [vendor homepage/appcast feed]`).
- Serves to explain related but slightly different reasons (rare but they do occur, and we usually bend the definition a bit).
- Makes it clear it was _verified_ (i.e. someone did indeed check it) and it was done so _only when first introduced_ (so users don’t think it’s verified regularly).
- The same comment can be used for both `appcast` and `url`.

In case there are concerns with the word “introduced”, I’ll point out the final result in a cask will be something like: `# amazonaws.com verified as official when first introduced to the cask` so it should be clear we’re referring to when the hostname itself was introduced (in case it changes, at which point we’ll need to verify it once more, as usual), not the cask.

If agreed upon, will make the following changes:
- [ ] Update all casks across all official repos.
  - [ ]  [caskroom/cask](https://github.com/caskroom/homebrew-cask).
  - [ ]  [caskroom/versions](https://github.com/caskroom/homebrew-versions).
  - [ ]  [caskroom/fonts](https://github.com/caskroom/homebrew-fonts).
- [ ] Add examples to documentation:
  - [ ] [url.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md).
  - [ ]  [appcast.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

### DIFF
--- a/doc/cask_language_reference/stanzas/appcast.md
+++ b/doc/cask_language_reference/stanzas/appcast.md
@@ -17,3 +17,17 @@ There are a few different ways the `appcast` can be determined:
  * The popular update framework [Sparkle](http://sparkle-project.org) generally uses the `SUFeedURL` property in `Contents/Info.plist` inside `.app` bundles. You can use the script [`find_sparkle_appcast`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/find_sparkle_appcast) to add this automatically. (Example Cask: [`glyphs`](https://github.com/caskroom/homebrew-cask/blob/161f85b605e160ff96e7dd11732d85609e13dc51/Casks/glyphs.rb#L6#L7))
 
 * An appcast can be any URL hosted by the app’s developer that changes every time a new release is out (e.g. a changelog HTML page). (Example Cask: [`shortcat`](https://github.com/caskroom/homebrew-cask/blob/161f85b605e160ff96e7dd11732d85609e13dc51/Casks/shortcat.rb#L6#L7))
+
+## When Appcast and Homepage Hostnames Differ, Add a Comment
+
+When the hostnames of `appcast` and `homepage` differ, the discrepancy should be documented with a comment of the form:
+
+```
+# URL_HOSTNAME verified as official when first introduced to the cask
+```
+
+Examples can be seen in [insert examples].
+
+These comments must be added so a user auditing the cask knows the appcast’s URL was verified by the Homebrew-Cask team as the one provided by the vendor, even though it may look unofficial or suspicious. It is our responsibility as Homebrew-Cask maintainers to verify both the `appcast` and `homepage` information when first added (or subsequently modified, apart from versioning). The exception to this rule is a `homepage` of `github.io` with a `appcast` of `github.com`, since we know this pair of hostnames is connected.
+
+The comment doesn’t mean you should trust the source blindly, but we only approve casks in which users can easily verify its authenticity, either by visiting the `homepage` or inspecting the app’s `Info.plist`.
--- a/doc/cask_language_reference/stanzas/url.md
+++ b/doc/cask_language_reference/stanzas/url.md
@@ -25,12 +25,14 @@ Example of using `:referer`: [rrootage.rb](https://github.com/caskroom/homebrew-
 When the hostnames of `url` and `homepage` differ, the discrepancy should be documented with a comment of the form:
 
 ```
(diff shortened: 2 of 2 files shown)


doc/cask_language_reference/stanzas/appcast.md
```diff
@@ -17,3 +17,17 @@ There are a few different ways the `appcast` can be determined:
...
+
+```
+# URL_HOSTNAME verified as official when first introduced to the cask
+```
+
+Examples can be seen in [insert examples].
+
+These comments must be added so a user auditing the cask knows the appcast’s URL was verified by the Homebrew-Cask team as the one provided by the vendor, even though it may look unofficial or suspicious. It is our responsibility as Homebrew-Cask maintainers to verify both the `appcast` and `homepage` information when first added (or subsequently modified, apart from versioning). The exception to this rule is a `homepage` of `github.io` with a `appcast` of `github.com`, since we know this pair of hostnames is connected.
+
+The comment doesn’t mean you should trust the source blindly, however, which is why we only add it to casks in which users can easily verify its authenticity, either by visiting the `homepage` or inspecting the app’s `Info.plist`.
```
adidalal: Wording is a bit odd here - We don't add casks where we can't verify the url, *period*. (ie. we don't (or won't) have any casks with no comment and a differing homepage/appcast that we haven't verified)

Maybe something like, the comment is there, but users can always verify by going to the homepage or digging around in Info.plist?

doc/cask_language_reference/stanzas/url.md
```diff
@@ -86,4 +88,4 @@ Some hosting providers actively block command-line HTTP clients (example: FossHu
 
 ## Vendor URLs Are Preferred
 
-When possible, it is best to use a download URL from the original developer or vendor, rather than an aggregator such as `macupdate.com`.
\ No newline at end of file
+When possible, it is best to use a download URL from the original developer or vendor, rather than an aggregator such as `macupdate.com`.
```
adidalal: I don't see any delta here

vitorgalvao: I don’t either. It’s not that uncommon. Probably some weirdness Atom auto-fixed.

doc/cask_language_reference/stanzas/appcast.md
vitorgalvao: Agreed. I also had some doubts when doing this part. Changed.

doc/cask_language_reference/stanzas/appcast.md
```diff
@@ -17,3 +17,17 @@ There are a few different ways the `appcast` can be determined:
...
+
+```
+# URL_HOSTNAME verified as official when first introduced to the cask
+```
+
+Examples can be seen in [insert examples].
+
+These comments must be added so a user auditing the cask knows the appcast’s URL was verified by the Homebrew-Cask team as the one provided by the vendor, even though it may look unofficial or suspicious. It is our responsibility as Homebrew-Cask maintainers to verify both the `appcast` and `homepage` information when first added (or subsequently modified, apart from versioning). The exception to this rule is a `homepage` of `github.io` with a `appcast` of `github.com`, since we know this pair of hostnames is connected.
+
+The comment doesn’t mean you should trust the source blindly, however, and we only approve casks in which users can easily verify its authenticity, either by visiting the `homepage` or inspecting the app’s `Info.plist`.
```
adidalal: Possible rewording: "blindly. However, we only approve Casks where users can verify the authenticity of these parameters, either"

vitorgalvao: Reworded, albeit a bit differently.

adidalal:
LGTM

doc/cask_language_reference/stanzas/url.md
elyscape: Previously, there was no newline at the end of the file. This commit added one.

vitorgalvao:
Closing in favour of https://github.com/caskroom/homebrew-cask/pull/18051.

adidalal:
(was never actually closed)

PR #17509, opened 2016-01-27
Upgrade borgbackup to 0.30.0

zchee:


adidalal:
Please add an [appcast](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md), as it's hosted on GitHub

zchee:
@adidalal Fixed.

BTW, I just wanted to update, your suggestion that add appcast was a little troubled.
I have experienced add of appcast for other Cask, so I was not cumbersome.
But, shallow of program experience users will be confused.
Is this automation is possible?

I'm very fond of `Homebrew|Caskroom` culture, so I think it is better to lower the threshold of contributing as much as possible.
Not a criticism.

Thanks.

PR #17519, opened 2016-01-27
Remove version specific details from Diskmaker X

jonkoops:
This removes version specific details from Diskmaker X as the URL for specific versions is no longer hosted by the creator.

### DIFF
--- a/Casks/diskmaker-x.rb
+++ b/Casks/diskmaker-x.rb
@@ -1,11 +1,11 @@
 cask 'diskmaker-x' do
-  version '5.0.2'
-  sha256 '9a210eb4db6d860e14d7f3e63d0cc67baeadd3a80e1ef607e3d2a44b85080472'
+  version :latest
+  sha256 :no_check
 
-  url "http://diskmakerx.com/downloads/DiskMaker_X_#{version.to_i}.dmg"
+  url 'http://diskmakerx.com/downloads/DiskMaker_X.dmg'
   name 'DiskMaker X'
   homepage 'http://diskmakerx.com/'
   license :gratis
 
-  app "DiskMaker X #{version.to_i}.app"
+  app 'DiskMaker X.app'
 end


Casks/diskmaker-x.rb
```diff
@@ -1,11 +1,11 @@
...
+  sha256 :no_check
 
-  url "http://diskmakerx.com/downloads/DiskMaker_X_#{version.to_i}.dmg"
+  url "http://diskmakerx.com/downloads/DiskMaker_X.dmg"
   name 'DiskMaker X'
   homepage 'http://diskmakerx.com/'
   license :gratis
 
-  app "DiskMaker X #{version.to_i}.app"
+  app "DiskMaker X.app"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

Casks/diskmaker-x.rb
```diff
@@ -1,11 +1,11 @@
 cask 'diskmaker-x' do
-  version '5.0.2'
-  sha256 '9a210eb4db6d860e14d7f3e63d0cc67baeadd3a80e1ef607e3d2a44b85080472'
+  version :latest
+  sha256 :no_check
 
-  url "http://diskmakerx.com/downloads/DiskMaker_X_#{version.to_i}.dmg"
+  url "http://diskmakerx.com/downloads/DiskMaker_X.dmg"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

jonkoops:
@vitorgalvao Done.

diskmakerx:
Hi Guys,

Why not ask me why the change, and ask me to put back the older versions ? ^^

adidalal:
We can switch back to a versioned Cask (which we prefer) if there is either a versioned URL, or an appcast we can glean update information from.

jonkoops:
@diskmakerx That'd be great.

diskmakerx:
No problem, I will leave the DMG's with the older versions besides the current version.

adidalal:
@diskmakerx We can add in a conditional for versions, but as I understand, that's only useful for putting old versions of OSX on a flash drive, it has nothing to do with the host OS. 

What this change did was remove the version, because we can't detect updates automatically anymore. If there's anything like an appcast, we can use that. See https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md (ie. a plain text/rss feed of the changelog page would work just fine, etc).

jonkoops:
@diskmakerx I assume [the site](http://diskmakerx.com/) is down because of maintenance related to this issue?
screen shot 2016-02-03 at 16 44 06

diskmakerx:
Must have been very temporary, it works fine now…

jonkoops:
@adidalal @diskmakerx There is an [RSS feed](http://diskmakerx.com/feed/) used for blog posts on the site that has the appropriate links to the downloads and also contains SHA checksums. Would that be sufficient?

vitorgalvao:
@jonkoops Absolutely! That’d work fine.

jonkoops:
@diskmakerx Yeah, the downtime was so small that it was so small I deleted my post :P

jonkoops:
@diskmakerx Distributing the releases using GitHub releases could also be an option.

adidalal:
@jonkoops Mind making the updated PR?

jonkoops:
@adidalal No problem, I'll try to get it done this weekend.

diskmakerx:
Hey guys,

I renamed the link on the home and download page to explicitely point to 5.0.3, so now it should be OK to get the proper version. Sorry for the change, I thought it was better this way, but you convinced me otherwise :)

PR #17538, opened 2016-01-28
New Cask mysql-connector-python

hanxue:


Casks/mysql-connector-python.rb
```diff
@@ -0,0 +1,17 @@
+cask 'mysql-connector-python' do
+  version '2.1.3'
+  sha256 '49655f970317383fc188ff48c777ade58be40d05d33be44c1c5e676546e41ba4'
+
+  if MacOS.release <= :mountain_lion
+    url "http://dev.mysql.com/get/Downloads/Connector-Python/mysql-connector-python-#{version}-osx10.8.dmg"
+  else
+    url "http://dev.mysql.com/get/Downloads/Connector-Python/mysql-connector-python-#{version}-osx10.9.dmg"
+  end
+  name 'MySQL Connector for Python'
```
vitorgalvao: Should be an empty line before.

vitorgalvao:
Thank you for the contribution, but it looks like there are similar connectors in homebrew. In the spirit of [deduplication](https://github.com/caskroom/homebrew-cask/issues/15603), this should be added there as well.

hanxue:
Hello @vitorgalvao , thanks for your quick feedback. There are a couple of reasons I submit this Cask

1. The `mysql-connector-python` package is a pre-requisite for the `mysql-utilities` cask to work
2. MySQL has [published connectors for many languages ](https://dev.mysql.com/downloads/connector/python/)(C, C++, Java, Python, .NET, ODBC). Only the C and C++ connectors has Homebrew formula
3. There is no Python MySQL Connector homebrew (source) formula, and MySQL does not officially publish the source for Python connector

There are only (source) formulas for the following connectors
```
$ brew search mysql-connector
homebrew/boneyard/mysql-connector-odbc   mysql-connector-c++ 
mysql-connector-c 
```

`mysql` related Casks
```
$ brew cask search mysql
==> Partial matches
mysql-utilities	    mysqlworkbench	navicat-for-mysql
```

I will update the pull-request with the missing blank space at the end. Unless you have another approach to package MySQL Connector for Python?

Casks/mysql-connector-python.rb
hanxue: Can we add the empty line check into `brew cask audit`? I always audit before submitting a new cask/formula

```
$ brew cask audit ./mysql-connector-python.rb 
audit for mysql-connector-python: passed
```

jawshooah:
You should use `brew cask style --fix` for stylistic concerns. `audit` is for checking the validity of the cask metadata itself.

vitorgalvao:
@hanxue Points **1** and **2**  aren’t really a concern. A cask can depend on a formula just as easily as it can depend on another cask. Also, the fact those connectors don’t exist (yet!) as formulas is reason enough to submit them there.

However, point **3** is very valid. If there is no source, then it has to be added here. Reopening.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 41fcec013db9bb34b03218439d27f45717e44949. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #17562, opened 2016-01-28
maintaining.md: new find_outdated_appcasts label

vitorgalvao:
Pinging @caskroom/maintainers to underline the point: **should never be applied manually**.

Currently working on the script.

### DIFF
--- a/doc/development/maintaining.md
+++ b/doc/development/maintaining.md
@@ -38,6 +38,7 @@ Label | Description | Issues | Pull Requests
 **bug** | Something isn’t working as expected. A modification/addition/removal. Must always be accompanied by **cask** or **core** | &#x2713; | &#x2713;
 **cask** | Relates directly to a cask. Must always be accompanied by **bug** or **enhancement**. | &#x2713; | &#x2713;
 **cask request** | Either a request for a new cask or a call for correction in an existing one. | &#x2713; |
+**outdated appcast** | An automated label, handled by the various scripts geared towards updating casks with outdated appcasts. Should never be applied manually. | &#x2713; | &#x2713;
 **chief bug** | When multiple people open new issues for the same bug, the main issue where its progression is being tracked should have this label. Every other one should be marked **duplicate** and closed. | &#x2713; |
 **core** | Relates directly to the code of the core, homebrew-cask itself. Must always be accompanied by **bug** or **enhancement**. | &#x2713; | &#x2713;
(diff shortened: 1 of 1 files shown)


victorpopkov:
Understood. :+1:

vitorgalvao:
For a bit more context, said label should only be applied by that script because I’ll later make another one that’ll loop through all the issues with said label.

vitorgalvao:
Made some changes to the label. Still should never be applied manually, but it will also be applied to PRs.

jawshooah:
:+1: Where are the scripts hosted? Couldn't find them in your tiny-scripts repo. Want to go ahead and open a PR to host them here? You can continue to refine them, and we can offer feedback :smile:

vitorgalvao:
> Where are the scripts hosted?

Nowhere yet. Have been building/refining all day.

> Want to go ahead and open a PR to host them here in `developer/bin`?

Yes. These ones definitely belong in `developer/bin`, not `tiny-scripts`. I’m still finishing everything up (small tweaks here and there) and building/thinking about latter parts of the process — ideally grab all PRs with the `outdated appcasts` tag, and if they passed the build, auto-merge them; but unfortunately so far haven’t found a way to check via CLI if a Travis build has passed.

Finishing that up is basically the reason I haven’t submitted them yet (making so many changes by the minute and cutting and moving so much between them, they’re not even under version control, yet).

I’ll probably finish them tomorrow (and hopefully will find a way to do the Travis check, for the auto-merge script), and then I’d like to do a full run (only did partial ones today) to make sure everything is OK before submitting.

They’re being surprisingly straightforward and small to build, so review should be simple.

jawshooah:
> unfortunately so far haven’t found a way to check via CLI if a Travis build has passed

`hub ci-status {{commit}}` should do the trick.

When testing the auto-merge, just to be safe, you might want to use your own fork as the source repo until you're confident it works as you expect.

vitorgalvao:
> When testing the auto-merge, just to be safe, you might want to use your own fork as the source repo until you're confident it works as you expect.

Yes, I usually do create temporary repos to test features, specially the most annoying ones (like opening issues) until I get them right.

adidalal:
Could also add a section to contributing, in updating a cask:

If you want to jump right in, outdated Casks are listed as issues with the "outdated appcast" label. Use cask-repair to easily submit a PR to close the issue.

vitorgalvao:
Not really needed. I also have a script that does one better than that (which is why the label is actually useful).

vitorgalvao:
Merging. Will submit the scripts shortly.

vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/18237.

PR #17670, opened 2016-01-29
initial panconvert cask

napaxton:
Getting a error message:

```
$ brew cask install panconvert
==> Downloading http://sourceforge.net/projects/panconvert/files/latest/download
==> No checksum defined for Cask panconvert, skipping verification
Error: It seems the symlink source is not there: '/opt/homebrew-cask/Caskroom/panconvert/latest/PanConvert.app'
```


Casks/panconvert.rb
```diff
@@ -0,0 +1,14 @@
...
+  url "http://sourceforge.net/projects/panconvert/files/latest/download?source=files"
+  appcast 'https://github.com/apaeffgen/PanConvert/blob/master/docs/changelog.md',
+      checkpoint: 'e2333a9cd663739863c8306193960401977aaa88'
+
+  name 'PanConvert'
+  homepage 'http://panconvert.sourceforge.net'
+  license :gpl
+
+  app 'PanConvert.app'
+end
```
vitorgalvao: It has a “no newline at end of file” warning. Those are [actually meaningful](http://unix.stackexchange.com/a/18789).

Casks/panconvert.rb
```diff
@@ -0,0 +1,14 @@
+cask 'panconvert' do
+  version :latest
+  sha256 :no_check
+
+  url "http://sourceforge.net/projects/panconvert/files/latest/download?source=files"
```
vitorgalvao: You haven’t changed it like I said. This won’t work.

napaxton: Oh, shoot. I did, but then I couldn't commit the changes (I'm not a vim person, and the commit message via vim…), and it must have reverted. Will fix later today.

-N

vitorgalvao: `git` uses `$EDITOR`, so whatever you set via `export EDITOR` will be used ([more details](https://unix.stackexchange.com/questions/77089/what-is-the-editor-command-in-bash)).

Casks/panconvert.rb
```diff
@@ -0,0 +1,14 @@
+cask 'panconvert' do
+  version :latest
+  sha256 :no_check
+
+  url "http://downloads.sourceforge.net/sourceforge/panconvert/PanConvert-0.1.5.dmg"
+  appcast 'https://github.com/apaeffgen/PanConvert/blob/master/docs/changelog.md',
+      checkpoint: 'e2333a9cd663739863c8306193960401977aaa88'
+
```
vitorgalvao: This empty line must go.

Casks/panconvert.rb
```diff
@@ -0,0 +1,14 @@
+cask 'panconvert' do
+  version :latest
+  sha256 :no_check
+
+  url "http://downloads.sourceforge.net/sourceforge/panconvert/PanConvert-0.1.5.dmg"
```
vitorgalvao: Do instead `url "http://downloads.sourceforge.net/sourceforge/panconvert/PanConvert-#{version}.dmg"`.

Casks/panconvert.rb
```diff
@@ -0,0 +1,14 @@
+cask 'panconvert' do
+  version :latest
```
vitorgalvao: Should be `version '0.1.5'`.

Casks/panconvert.rb
```diff
@@ -0,0 +1,14 @@
...
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

adidalal:
When you're done, don't merge, rebase and squash your commits.
https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history or https://davidwalsh.name/squash-commits-git if you need help

Casks/panconvert.rb
napaxton: so just add "\n" after "end"?

Casks/panconvert.rb
napaxton: If I do that, why would I still have `version 0.1.5` instead of `version :latest`?

jawshooah: It's just string interpolation. The point is to specify the version exactly once, in the `version` stanza, and reuse that value where needed.

Casks/panconvert.rb
vitorgalvao: Don’t worry anymore, looks like it is fixed.

vitorgalvao:
Thank you. There are still some audit and formatting issues, but I can take care of those.

adidalal:
prfixmaster fails on these kinds of PRs, last I checked. Errors at applying patch or something. @vitorgalvao

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 22b967fc7c738ff507380a6712b5e701f3403e48. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

vitorgalvao:
@adidalal Usually, you just need to keep `git am --skip`ping. That’s one of the reasons I’m wary of reducing verbosity in `prfix(master|branch)`, unlike what I’ve done recently with `fastmerge`. It isn’t really `fastmerge`’s fault, but of the commits in the PR and how git handles it (not that it’s wrong, or anything, just providing context).

napaxton:
Why is the checkpoint different?

	I see the singel quotes I left out, so I know for the future…

vitorgalvao:
> Why is the checkpoint different?

Because yours was pointing to a webpage. This is the raw file, which is not only smaller, but less prone to give false positives.

napaxton:
How do I find that checkpoint? For the github webpage enclosgin the raw, it's obvious. (I"m pretty good with Terminal, but less so with git and have been using a GUI.)
> Why is the checkpoint different?
> 
> Because yours was pointing to a webpage. This is the raw file, which is not only smaller, but less prone to give false positives.
>

vitorgalvao:
> How do I find that checkpoint?

[Docs](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

PR #17677, opened 2016-01-29
Fix symlink source is not there in diskmaker-x cask.

ydesgagn:


Casks/diskmaker-x.rb
```diff
@@ -1,11 +1,11 @@
 cask 'diskmaker-x' do
-  version :latest
-  sha256 :no_check
+  version '5.0.2'
```
victorpopkov: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. There is no point in changing this cask to versioned one.

victorpopkov:
Thank you for the contribution. It needed some fixes, so they were made in 48f767da05b700ec39c630ee82b1945801c2e7fc. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #17848, opened 2016-01-30
Updated firefox (44.0)

tsparber:


Casks/firefox.rb
```diff
@@ -3,6 +3,8 @@
   sha256 '74f7d4041f2582553a460fd8249575dfc501b57bdee37f43083e8d7277d2d722'
 
   url "https://ftp.mozilla.org/pub/firefox/releases/#{version}/mac/en-US/Firefox%20#{version}.dmg"
+  appcast 'https://aus5.mozilla.org/update/3/Firefox/40.0.0/0/Darwin_x86_64-gcc3-u-i386-x86_64/en-US/release/Darwin%2015.3.0/default/default/update.xml?force=1',
+          checkpoint: '91a080855e9ee207992c7f5eebf2c4fe8bb7221b983c77f0385e973d34155cea'
```
tsparber: This url is used by firefox to check for updates. More about it can be found in the documentation [Software_Update:Checking_For_Updates](https://wiki.mozilla.org/Software_Update:Checking_For_Updates).
The version number (`40.0.0`) is the "current version" from which to update (anything above `12` gives the latest version `44`. Below `12` it asks to update to firefox version `12` and then to `44`).

tsparber:
Maybe a language specific appcast could/should be used in all other [firefox](https://github.com/caskroom/homebrew-versions/search?p=3&q=firefox&utf8=%E2%9C%93)/[thunderbird](https://github.com/caskroom/homebrew-versions/search?p=3&q=thunderbird&utf8=%E2%9C%93) casks?

vitorgalvao:
It’d be nice to use language-specific `appcast`s in those, yes, but presumably they’re all update at once, so it shouldn’t be strictly necessary. I you’re thinking about submitting, do what’s most convenient.

Casks/firefox.rb
vitorgalvao: We can interpolate `version` in the `appcast`, then

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 846380b61a35800075b4d785d1a422242931c816. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #17895, opened 2016-01-30
change to reflect change in appname in .dmg

liljenstolpe:


vitorgalvao:
Please [amend](http://stackoverflow.com/a/179147/1661012) the commit message to include the name of the updated cask.

Top line has to be changed to `cask 'coreos' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `coreos.rb`.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in e026dda9053f1bd57e1a57cf885caad7f76632ba. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #17897, opened 2016-01-30
Add Kube-Solo v0.5.1

liljenstolpe:


Casks/kube-solo-osx.rb
```diff
@@ -0,0 +1,13 @@
+cask 'kube-solo-osx' do
```
vitorgalvao: This has to be changed to `cask 'kube-solo' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `kube-solo.rb`.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in b0a4138cfb8e7fd7e5da42ac5b17197a53ab737e. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #17899, opened 2016-01-30
Added Kube-Cluster v0.1.7

liljenstolpe:


Casks/kube-cluster-osx.rb
```diff
@@ -0,0 +1,13 @@
+cask 'kube-cluster-osx' do
```
vitorgalvao: This has to be changed to `cask 'kube-cluster' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `kube-cluster.rb`.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 8f5fafa1aaf1c7fc285401a70a56bc9f80698cb7. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #17924, opened 2016-01-31
Remove depends_on_java caveat from jetbrain products

leipert:
Hey. 

Hopefully it does not offend anybody, that this fixes multiple casks.

All current versions of jetbrains products bring a bundled version of java with them [1], therefore it is not necessary to install a JDK if you want to use them.

If you have any questions, I am happy to answer them :+1: 

[1]: [Jet Brains Support Article](https://intellij-support.jetbrains.com/hc/en-us/articles/206544879-Selecting-the-JDK-version-the-IDE-will-run-under) (have a look at the Mac OSX Section)

### DIFF
--- a/Casks/appcode.rb
+++ b/Casks/appcode.rb
@@ -18,8 +18,4 @@ cask 'appcode' do
                 '~/Library/Caches/AppCode33',
                 '~/Library/Logs/AppCode33',
               ]
-
-  caveats do
-    depends_on_java
-  end
 end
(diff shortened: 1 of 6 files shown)


vitorgalvao:
We do prefer you submit separate pull requests for each cask, as more often than not this type of pull request brings problems with one or more of them, and it halts the inclusion of the other ones. For these simple cases, at the very least, one commit per cask.

This isn’t a big change, tough, so we can merge this one as is.

Thank you.

leipert:
Thanks as well. Will keep your advice in mind.

PR #17975, opened 2016-01-31
cog.rb: added appcast verified comment

vitorgalvao:


vitorgalvao:
Merged as 480deb6fcb4645089718f80068c939941841a722.

PR #18056, opened 2016-02-01
Add new nrf5x-command-line-tools: 8.2.0

adzenith:
nRF5x Command Line Tools were just released for OS X: https://devzone.nordicsemi.com/blogs/840/nrfjprog-pynrfjprog-intro-mac-os-x-linux-now-suppo/

I searched around a bit for how to copy .dylib files; let me know if this is the correct approach.

Thanks!

### DIFF
--- a/Casks/nrf5x-command-line-tools.rb
+++ b/Casks/nrf5x-command-line-tools.rb
@@ -1,0 +1,12 @@
+cask 'nrf5x-command-line-tools' do
+  version '8.2.0'
+  sha256 '22c217ce3ef64a1e85d1ec339692fc9eed73ca8a9011cb0b21ddcd5e5bfc84ad'
+
+  url 'http://www.nordicsemi.com/eng/nordic/download_resource/53406/1/8099423'
+  name 'nRF5x Command Line Tools'
+  homepage 'http://www.nordicsemi.com/eng/nordic/Products/nRF52832/nRF5x-Command-Line-Tools-OSX/53406'
+  license :gratis
+
+  binary 'nrfjprog/nrfjprog'
+  binary 'mergehex/mergehex'
+end


Casks/nrf5x-command-line-tools.rb
```diff
@@ -0,0 +1,21 @@
...
+  sha256 '22c217ce3ef64a1e85d1ec339692fc9eed73ca8a9011cb0b21ddcd5e5bfc84ad'
+
+  url 'http://www.nordicsemi.com/eng/nordic/download_resource/53406/1/8099423'
+  name 'nRF5x Command Line Tools'
+  homepage 'http://www.nordicsemi.com/eng/nordic/Products/nRF52832/nRF5x-Command-Line-Tools-OSX/53406'
+  license :gratis
+
+  binary 'nrfjprog/nrfjprog'
+  binary 'mergehex/mergehex'
+  artifact "nrfjprog/libjlinkarm_nrf51_nrfjprogdll.#{version}.dylib",       target: "/usr/local/bin/libjlinkarm_nrf51_nrfjprogdll.#{version}.dylib"
```
vitorgalvao: Is `/usr/local/bin` mandatory for this to work?

vitorgalvao: Needs the [files_in_usr_local caveat](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md#caveats-mini-dsl) if that is the case.

adzenith: That's supposed to be /usr/local/lib—I don't know why I wrote bin!
Can I copy to /usr/local/lib? They can be anywhere as long as they're on the library path.

adzenith: Based on the fact that the libs were in ./bin instead of ./lib and everything still worked, I tried deleting them. Looks like they're unnecessary for the binary files—perhaps provided for linking against when building using the SDK. Should I just remove them?

vitorgalvao: Yes, please remove them, then. It’s best to err on the side of less stuff. If they turn out to be needed, we’ll notice later.

adzenith:
Pushed a new version without the dylibs. Thanks!

adidalal:
Thank you for the contribution. It was merged directly as 05158199f94eb603e89058db6a8c21b894c321af to keep commit history cleaner. Your contribution is still credited to you.

adzenith:
Thanks!

PR #18061, opened 2016-02-01
Add FirefoxNightly.app v47.0a1

radubogdan:


adidalal:
Please update https://github.com/caskroom/homebrew-versions/blob/master/Casks/firefox-nightly.rb instead.

PR #18100, opened 2016-02-01
Mnemosyne version update/added caveat

laurenkt:
Updated Mnemosyne version to 2.3.5 and added note about complications with getting it to work with a LaTeX installation.

Originally I wanted to do the steps outlined in the caveat for the user automatically, but I feel like maybe that isn't appropriate. However, it seems like a common enough usage for Mnemosyne that it's worth putting in the note (especially for me as I face this problem every time I install Mnemosyne on a new machine).

Is there a way to specify options? I had originally thought of something like `brew cask install mnemosyne --with-latex` which would run the suggested commands and also have a cask dependency on MacTeX, but I couldn't find any precedent for that sort of behaviour.

### DIFF
--- a/Casks/mnemosyne.rb
+++ b/Casks/mnemosyne.rb
@@ -4,6 +4,8 @@ cask 'mnemosyne' do
 
   # sourceforge.net is the official download host per the vendor homepage
   url "http://downloads.sourceforge.net/project/mnemosyne-proj/mnemosyne/mnemosyne-#{version}/Mnemosyne-#{version}.dmg"
+  appcast 'http://sourceforge.net/p/mnemosyne-proj/activity/feed?source=project_activity',
+          checkpoint: 'a8f3a1d3ec76cfbc00c91c8957abb728e742e61860bc694c355421e3f8f0d8bc'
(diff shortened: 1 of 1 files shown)


adidalal:
We don't support options. Still undecided about if this is appropriate for caveats in the main repo, maybe @vitorgalvao will have some thoughts.

However, you certainly could do all these things automatically and host the resulting Cask in your own private tap, ie. `mnemosyne-latex` - homebrew cask supports taps as a first class citizen

vitorgalvao:
I’m also a bit torn on these caveats. However, they do not describe things you need to do to get it working, you’re describing something that can be added to the app after installation. In addition, most of the (huge) caveat is teaching users how to set environment variables for specific apps via Launch Services.

Will remove the caveats, then.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in c27827f1d162ddfb1a3d2f47f38e20ac405f14db. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #18102, opened 2016-02-01
Remove Java dependency

pbowyer:
From the [downloads page](http://www.charlesproxy.com/latest-release/download.do):

> Mac OS X users
> 
> The main download of Charles now includes Java 8, so you no longer need to use the alternative download for better SSL support. The alternative download is now strictly for oldies.


adidalal:
Thank you for the contribution. It was merged directly as a99cce37afb6edc8f18bb761d33daefdfd5730c2 to keep commit history cleaner. Your contribution is still credited to you.

PR #18107, opened 2016-02-02
adding Lektor

drewda:


adidalal:
Is this significantly different from the pypy package? We generally don't like to duplicate functionality of other package managers

adidalal:
Ah, I see it's self-contained; should be fine to add in that case

drewda:
There are two packages: a CLI and a GUI version https://www.getlektor.com/downloads/ I assume the former is what's on PyPI. This is the latter.

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 382c22c7181de6fc9af45d96bc35e26e6e5c39d7. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

drewda:
Thanks, @adidalal

PR #18238, opened 2016-02-04
Add miniconda.rb

drzax:
Adds miniconda3 by Continuum Analytics. The full anaconda install is massive and often uneccesary. This is a supported and valid alternative.

### DIFF
--- a/Casks/miniconda.rb
+++ b/Casks/miniconda.rb
@@ -1,0 +1,22 @@
+cask 'miniconda' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh'
+  name 'Continuum Analytics Miniconda'
+  homepage 'https://www.continuum.io/why-anaconda'
+  license :gratis
+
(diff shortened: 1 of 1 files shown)


adidalal:
Use a versions url, check the anaconda cask for an example

drzax:
It appears to be impossible to get the most up to date version from a versioned URL. Indeed, the latest version mentioned in the appcast changelog at the URL suggested (3.19.1) is not available from the distribution repo.

Will this cause problems with the way appcast functions? I also think this will make it impossible for me to create the required checkpoint parameter?

Happy to submit an update if these aren't problems.

adidalal:
Took a closer look; agreed on above points. It's not ideal but AFAIK conda is basically able to update itself with `conda update conda` so it's not a big deal.

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 0154744e1574604f704dde83cea6fdd97d6ba7e3. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

drzax:
Thanks.

PR #18246, opened 2016-02-04
Update sqlworkbenchj.rb

spencerhakim:
Unfortunately, old download URLs are invalidated upon release of a new version, with old versions being moved into the `/archive` subdirectory(eg, `http://www.sql-workbench.net/Workbench-Build#{version}-MacJava7.tgz` moves to `http://www.sql-workbench.net/archive/Workbench-Build#{version}-MacJava7.tgz`). Maybe someone who knows more about Ruby and Homebrew could patch this to check both URLs for a given version # so the recipe doesn't completely break every time they update.


vitorgalvao:
We don’t support that.

vitorgalvao:
Thank you for the contribution. It was merged directly as 8150844b9c219730f4e198855ccc4805f8a03428 to keep commit history cleaner. Your contribution is still credited to you.

PR #18271, opened 2016-02-05
Add FreshBackMac v1.5

aguynamedryan:
Installs latest version of FreshBackMac.app from a zip file.

FreshBackMac updates your desktop background from various sources.

### DIFF
--- a/Casks/freshback.rb
+++ b/Casks/freshback.rb
@@ -1,0 +1,11 @@
+cask 'freshback' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://arkanath.com/FreshBackMac/FreshBackMac.zip'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

adidalal:
Thank you for the contribution. It was merged directly as 112d699be5c302ebf6148d4aa047041d0338adc1 to keep commit history cleaner. Your contribution is still credited to you.

Issue #18247, opened 2016-02-04
Cask request: Craft


erichyland:
### Cask details

(Please fill out as much as possible)

**Name** - 
Craft

**Homepage** -
http://labs.invisionapp.com/craft

**License** -
Freeware

**Download URL** -
https://s3.amazonaws.com/www-assets.invisionapp.com/labs/CraftInstaller-osx.zip

**Description** -
Craft is a plugin suite for Sketch and Photoshop that lets you design with real data in mind.

adidalal:
The installer, if you can call it that, doesn't work particularly well with HBC. 

@sbdchd has opened up a PR that fulfills this request, but it's pending further feedback.

vitorgalvao:
Closing in favour of #18278.

PR #18278, opened 2016-02-05
added Craft Installer cask

sbdchd:
I am not sure if this qualifies to be an official cask since it is an installer, but this should satisfy #18247.

### DIFF
--- a/Casks/craft-installer.rb
+++ b/Casks/craft-installer.rb
@@ -1,0 +1,11 @@
+cask 'craft-installer' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://s3.amazonaws.com/www-assets.invisionapp.com/labs/CraftInstaller-osx.zip'
+  name 'Craft Installer'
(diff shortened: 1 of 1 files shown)


adidalal:
It'd have to be installer :manual, which we don't really prefer. However, it looks like all this is doing is copying files to a given directory.

Leaving it open for opinions.

vitorgalvao:
@sbdchd Like @adidalal mentioned, this needs to use [`installer :manual`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/installer.md#installer-manual) (and hence [`uninstall` as well](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md). However, this is also a [walled build](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask), so it belongs in [caskroom/unofficial](https://github.com/caskroom/homebrew-unofficial), not here.

PR #18295, opened 2016-02-06
Update vectr to account for .zip structure change

dwaite:
Vectr currently fails to install due to the zip now including the Vectr.app in a subdirectory, while this corrects the issue


adidalal:
Thank you for the contribution. It was merged directly as e17b1ec023844776c0f6b887775dbdf479461d1d to keep commit history cleaner. Your contribution is still credited to you.

PR #18297, opened 2016-02-06
Fix App Stanza target in Sassquatch Cask

tysongach:
This fixes the following error, which I was seeing when trying to install Sassquatch:

``` bash
Error: It seems the symlink source is not there: '/opt/homebrew-cask/Caskroom/sassquatch/1.5.0/Sassquatch.app'
```

### DIFF
--- a/Casks/sassquatch.rb
+++ b/Casks/sassquatch.rb
@@ -7,5 +7,5 @@ cask 'sassquatch' do
   homepage 'http://sassquatch.thoughtbot.com'
   license :commercial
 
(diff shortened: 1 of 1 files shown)


adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 1b6c0f423007259dfb4fa6405337b01a6acf219e. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

tysongach:
Thanks!

PR #18298, opened 2016-02-06
Add Azul's Zulu JDK bundle

nathansgreen:
Zulu is a supported release of OpenJDK published by Azul systems.

### DIFF
--- a/Casks/zulu.rb
+++ b/Casks/zulu.rb
@@ -1,0 +1,53 @@
+cask 'zulu' do
+  version '1.8.0_72,8.13.0.5'
+  sha256 '1e6ab8b4660ad50248d5e1c9568566bb2d6206788d4ff913d6539b47520da0ac'
+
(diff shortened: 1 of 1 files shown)


adidalal:
Needs (at least) `brew cask style zulu --fix`
Not too familiar with the java-esque casks so will leave it for another maintainer to review.

vitorgalvao:
Most if not all of the `split`s can be replaced with better [version methods](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md#version-methods).

nathansgreen:
I can't figure out how to use `homepage` as referrer, so that value is duplicated. The version splits came from `java.rb` but it's much cleaner with the version methods.

nathansgreen:
Azul's download page isn't sorted, so the order of releases seems to vary. This now uses the `_72` release.

adidalal:
Thank you for the contribution. It was merged directly as dc26dc02117b5b3f70dd45ecda154a27dcaa9623 to keep commit history cleaner. Your contribution is still credited to you.

PR #18300, opened 2016-02-07
Upgrade utox to 0.5.0

jonafato:
The format of the download link has changed, so that is being updated
along with the version number.


adidalal:
Thank you for the contribution. It was merged directly as 326d1666c3d071fa79cb02f2da8ce24bccb27a5e to keep commit history cleaner. Your contribution is still credited to you.

PR #18433, opened 2016-02-09
Update brew-cask.1

miccal:


vitorgalvao:
The quotes we have are correct.

You should never edit this file directly. It should be generated after editing [the markdown version]( https://github.com/caskroom/homebrew-cask/blob/master/doc/man_page/brew-cask.1.md). Please correct that one instead.

miccal:
Running `man brew-cask` via the terminal, the quotes appear as `ahelpera`?

vitorgalvao:
> Running `man brew-cask` via the terminal, the quotes appear as `ahelpera`?

Ah, you’re right. Still, quotes should be something like `"helper"` and the fixed file should still be the markdown one.

miccal:
OK, but how should the quotes be written so that they appear correct in Terminal?

vitorgalvao:
`"helper"`.

PR #18463, opened 2016-02-09
Fix download URL in zulu.rb

victorpopkov:
Fixes https://github.com/caskroom/homebrew-cask/issues/18456.

Should be a better way, like having a version methods `before_underscore` and `after_underscore`.


vitorgalvao:
We didn’t want to have too many methods. It’s OK for some outliers to use a regex.

PR #18465, opened 2016-02-09
Upgrade consul-template to v0.12.2

mcornick:
Changing the binary target because the zip does not expand into a subdirectory. This is consistent with other Hashicorp tools in the caskroom.


vitorgalvao:
Thank you for the contribution. It was merged directly as 1cedc65d9c93810c70c6d66ba516f440d61bdc34 to keep commit history cleaner. Your contribution is still credited to you.

PR #18471, opened 2016-02-09
Auto updates

toonetown:
These changes add the `auto_updates` flag for those casks (that I have installed on my system) that use Sparkle.framework.

I can split these into separate pull requests, if desired.  I can also squash them together into a single commit, if preferred.


toonetown:
I will fix these style issues...I didn't notice them previously.

toonetown:
There - that's better!  I didn't know that `brew cask audit` and `brew cask style` existed... :)

vitorgalvao:
Thank you!

PR #18484, opened 2016-02-10
Only initialize Homebrew-Cask when necessary

ilovezfs:
Creating directories should be deferred until needed.

Fixes homebrew/homebrew-bundle#139

### DIFF
--- a/lib/hbc/cli.rb
+++ b/lib/hbc/cli.rb
@@ -73,6 +73,10 @@ class Hbc::CLI
     raise unless e.to_s.include? path
   end
 
+  def self.should_init?(command)
+    (command.is_a? Class) && (command < Hbc::CLI::Base) && command.needs_init?
+  end
+
   def self.run_command(command, *rest)
     if command.respond_to?(:run)
       # usual case: built-in command verb
@@ -112,8 +116,8 @@ class Hbc::CLI
   def self.process(arguments)
     command_string, *rest = *arguments
     rest = process_options(rest)
-    Hbc.init
     command = Hbc.help ? 'help' : lookup_command(command_string)
+    Hbc.init if should_init?(command)
     run_command(command, *rest)
   rescue Hbc::CaskError, Hbc::CaskSha256MismatchError => e
     msg = e.message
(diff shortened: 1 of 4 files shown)


ilovezfs:
@adidalal Just a heads up that this is the PR you were interested in.

vitorgalvao:
Also pinging @jawshooah.

lib/hbc/cli.rb
```diff
@@ -280,3 +280,15 @@ def _help_for(klass)
     end
   end
 end
+
+class String
```
jawshooah: I'd rather not add needless extensions to core classes. Why not just add a method `needs_init?(command)` to `Hbc::CLI`?

ilovezfs: You'd need to make help and null real commands.

jawshooah: I mean add the method to `Hbc::CLI`, not `Hbc::CLI::Base` (well, I suppose both). For example:

```rb
# Hbc::CLI

def needs_init?(command)
  command.is_a? Hbc::CLI::Base && command.needs_init?
end
```

ilovezfs: sure, but perhaps call it "should_init" instead?

jawshooah: That's fine with me.

lib/hbc/cli.rb
```diff
@@ -73,6 +73,10 @@ def self.require? path
     raise unless e.to_s.include? path
   end
 
+  def self.should_init?(command)
+    (command.is_a? Class) && (command < Hbc::CLI::Base) && command.needs_init?
```
ilovezfs: @jawshooah better?

also, is_a? won't work but < does. Do you prefer < or ancestors.include?

jawshooah: This is fine, thank you.

adidalal:
:+1: Thanks @ilovezfs

ilovezfs:
@adidalal You're welcome. And https://github.com/Homebrew/homebrew-bundle/issues/139#event-546022808 is now closed. Thanks for merging so quickly.

adidalal:
Looks like this might cause https://github.com/caskroom/homebrew-cask/issues/18606 - any insight on the issue?

PR #18496, opened 2016-02-10
Update pkg.md

miccal:
Missed a colon.


miccal:
And that's all of them I think!

vitorgalvao:
> And that's all of them I think!

PR #18499, opened 2016-02-10
upgrade sketch to version 3.5.2

akrawchyk:
Upgrades sketch from 3.5.1 to 3.5.2, see http://www.sketchapp.com/whats-new/:

**Bug Fix**
- Fixes a vulnerability in the updater by now delivering all updates over https

### DIFF
--- a/Casks/sketch.rb
+++ b/Casks/sketch.rb
@@ -1,6 +1,6 @@
 cask 'sketch' do
-  version '3.5.1'
-  sha256 'f5ee1e689c941bf480fcc7a462347ca45d9ec94dde8395e79a879e12882f643e'
+  version '3.5.2'
+  sha256 'fac9182c50a96cbc144bddb3984c82195980d6a4dc92fa3c2e40a59574f40a81'
 
   url 'http://www.sketchapp.com/static/download/sketch.zip'
   appcast "http://www.sketchapp.com/appcast#{version.major}.xml",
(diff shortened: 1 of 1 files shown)


Casks/sketch.rb
```diff
@@ -1,6 +1,6 @@
 cask 'sketch' do
-  version '3.5.1'
-  sha256 'f5ee1e689c941bf480fcc7a462347ca45d9ec94dde8395e79a879e12882f643e'
+  version '3.5.2'
+  sha256 'fac9182c50a96cbc144bddb3984c82195980d6a4dc92fa3c2e40a59574f40a81'
 
   url 'http://www.sketchapp.com/static/download/sketch.zip'
   appcast "http://www.sketchapp.com/appcast#{version.major}.xml",
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

akrawchyk: @vitorgalvao thanks for the quick feedback and detailed info. I've made the requested update to the `:checkpoint` value for `appcast`.

vitorgalvao:
Thank you for the contribution. It was merged directly as 7f5f5976d0766661b9ecce3e99eaa8b0656562d7 to keep commit history cleaner. Your contribution is still credited to you.

PR #18739, opened 2016-02-11
Delete crashlytics.rb

vitorgalvao:
Seems to be dead (integrated into another app that is behind a registration wall).


vitorgalvao:
Merged as a89dd7884cc7dc981ee9184058a839c631125725.

PR #18746, opened 2016-02-11
Delete darteditor.rb

vitorgalvao:
Seems to no longer be a thing.


vitorgalvao:
Merged as 533ec2e06e6c9acc7ae82e4eed41b2c4f9065251.

PR #18783, opened 2016-02-12
Modified README.md with better wording that should improve readabilit…

rasebo:
Modified README.md with better wording that should improve readability and help people not to skip key information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,13 +71,13 @@ In addition, if you haven’t yet, run the following once to [switch to the new 
 $ brew uninstall --force brew-cask; brew update
 ```
 
-Finally, if it's been a while since you last updated Homebrew-Cask, you may still have the old `phinze/cask` tap installed. Run the following to clean up, then try your command again:
+If it's been a while since you last updated Homebrew-Cask, you may still have the old `phinze/cask` tap installed. Run the following to clean up, then try your command again:
 
 ```bash
 $ brew untap phinze/cask; brew untap caskroom/cask; brew update
 ```
 
-If the issue persists, please use our [bug report template][bug_report_template] to report the output of running the problematic command with the `--verbose` flag, along with the output of `brew cask doctor`.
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you, but this isn’t a common issue.

Also, it changes the meaning to the wrong one. “Finally” refers to the to the list of instructions. Reporting an issue is indeed a separate part.

rasebo:
Welcome, however it might be that it is a common behaviour and nobody thought that "finally" might actually be the issue (which would help with the high number of "fake" bugs raised). I tested it with two colleagues in the office and all three of us missed the last part and stopped at "finally".

The fact that you use the "finally" in there makes it extremely easy to skip the important part at the end, as there is no clear delimitation between the list of instructions and the category as a whole. Finally means just that something has ended, and the expectancy is that nothing after that (besides the obvious code box) is of any importance. We all just skim online text searching for keywords.

vitorgalvao:
> I tested it with two colleagues in the office and all three of us missed the last part and stopped at "finally".

Anecdotal, but made me smile (truly), and I appreciate the effort.

I do see your point, so lets meet halfway: remove the “Finally” and have both sentences start at “If”. Unfortunately, I can’t reopen the PR (perhaps because you deleted the repository?), so if you want to submit a new one with that change, I’ll merge it.

rasebo:
It can be difficult to provide hard evidence when it comes to behaviour. Smiling is good, so I might consider making a video next time I empirically validate something like this (no sarcasm here, really). There's nothing like watching a few definitely not dull people behaving like idiots. :)

That sounds like a good compromise. I did remove the repository, so I will fork it again and make the change. Cheers!

vitorgalvao:
The video idea sounds fun, and good for a few laughs.

As for reforking, not sure if you know, but if you edit the file through the web interface (in this repo), it’ll take care of all that forking, branching, and the like.

rasebo:
Ah, wasn't aware. Thanks!

PR #18786, opened 2016-02-12
Removed the word "Finally"

rasebo:
Removed the word "Finally", which could have caused people to stop reading and not notice the last paragraph.


vitorgalvao:
Merged! Thank you.

PR #18803, opened 2016-02-13
updated macpass to version 0.6-alpha

asbachb:


Casks/macpass.rb
```diff
@@ -1,11 +1,11 @@
...
-  version '0.5.1-alpha'
-  sha256 '6d69ce183d5be1031df344b3cf4941a3d9495dc6dedb2d67611bbd89b754427d'
+  version '0.6-alpha'
+  sha256 '36e603a0ee310ccf341b885948910b0d05a0a29fe2aac19b0e12a47a74159e3a'
 
   # github.com/mstarke/MacPass was verified as official when first introduced to the cask
   url "https://github.com/mstarke/MacPass/releases/download/#{version}/MacPass-#{version}.zip"
   appcast 'https://github.com/mstarke/MacPass/releases.atom',
-          checkpoint: 'c2ff389bb6f739c21ea10de86bead6a6f8fab317754929388e4ce2d1ed5f91c7'
+          checkpoint: 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
```
vitorgalvao: This value is wrong (it’s the value for an empty string). How did you calculate it?

Correct one is `81b5266c94e2bf6910661686962296be947bd1a07de612c36b22d8ba53de1ab4` (`curl --silent --compressed 'https://github.com/mstarke/MacPass/releases.atom' | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }' | tr -d '\n'`).

asbachb:
Good spot you're right. I think I pasted the appcast url wrong. I used the command in the docs. Retried and got the same value.

Did a forced push on my branch.

vitorgalvao:
Thank you for the contribution. It was merged directly as 8768a891c462fd095a1434b711039cdf14d09b14 to keep commit history cleaner. Your contribution is still credited to you.

PR #18807, opened 2016-02-13
Update NZBVortex version

ptb:


Casks/nzbvortex.rb
```diff
@@ -1,5 +1,5 @@
 cask 'nzbvortex' do
-  version '2.11.10'
+  version '3.2.3'
   sha256 'bef3d0a6297bb4fff98361d4108061b7fff0c8d6ab1f2e2d88a56e67d9fa00b8'
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

Leave it, though, as there are other corrections needed on the cask, and will be making them in a different PR. Thank you for the contribution and bringing the updated version to our attention.

ptb:
I think what happened is that the version was missed when it was last updated. The current download for v3.2.3 matches the sha256 listed.

vitorgalvao:
> The current download for v3.2.3 matches the sha256 listed.

[It doesn’t](https://github.com/caskroom/homebrew-cask/pull/18808/files).

vitorgalvao:
Even so, like I said [many other corrections were needed](https://github.com/caskroom/homebrew-cask/pull/18808/files).

PR #18811, opened 2016-02-13
Delete ps3-media-server.rb

vitorgalvao:
Hasn’t been updated in years, is missing `uninstall`, and doesn’t seem popular.

If the cask is still desired, it will need to be readded with `uninstall`.


alswl:
Recommand 'universal-media-server` for instead.

PR #18827, opened 2016-02-13
blender

psibre:


psibre:
on case-sensitive filesystems, the paths to `blender.app` were broken in 4b55e1c2d229fb6d4f23fe7b2a909cc878c94c98, and then not completely fixed in fe00d2b70d41fc37b853beb6d232f3eb9004f359

vitorgalvao:
Thank you.

PR #18829, opened 2016-02-13
installer and pkg must allways be accompanied by uninstall

vitorgalvao:


adidalal:
Thought this was already there, but yeah, makes sense.

vitorgalvao:
Merged as 77f037139b26befceb8f4363df9f5d7701e50a6d.

PR #18878, opened 2016-02-14
geotagalog.rb: Cask added (1.5.6).

janlugt:


adidalal:
Only supports Mac OS X 10.5–10.6 according to the homepage - consider putting this in a user tap, generally speaking, `caskroom` Casks should work on the latest version of OS X.

PR #18884, opened 2016-02-14
(Merge on March 2nd 2016) Delete flash.rb

vitorgalvao:
image

@andyli You’re usually the flash updater. Any reason to keep this?


vitorgalvao:
@andyli You’re usually the flash updater. Any reason to keep this?

adidalal:
Yes, the flash pkg is actually linked elsewhere on adobes site. Will take a look at it once at a computer

adidalal:
https://jamfnation.jamfsoftware.com/viewProduct.html?id=41&view=info

vitorgalvao:
My point is (emphasis mine):

> This page **and the download links** will be decommissioned on March 1st, 2016.

So I don’t see a case to keep this around.

adidalal:
Ah I see, will check out what autopkg is using at some point soon. 

https://github.com/autopkg/recipes/tree/master/AdobeFlashPlayer

Hold off on the deletion until resolved, please, it's a popular Cask AFAIK

andyli:
Although I've been updating the formula, I am just doing that for completeness when updating the `flash-player` formula, which is used by some of my TravisCI builds. However, I do think that there could be many people using the `flash` formula, so it should be updated instead of be removed.

Following the steps in https://get.adobe.com/flashplayer/otherversions/, the download url of the Mac NPAPI flash plug-in is https://admdownload.adobe.com/bin/live/AdobeFlashPlayer_20_a_install.dmg
It should be safe to use since the same domain is used in the `pepper-flash` formula.

adidalal:
Updated installer URL (and even got an appcast)
https://github.com/caskroom/homebrew-cask/pull/18887

PR #18887, opened 2016-02-14
updated flash (20.0.0.306)

adidalal:
@vitorgalvao Played around with the `autopkg` recipies and got this.
Works for me and seems to do the same thing as the old installer.

@andyli Maybe a derivative of this could help with `flash-player` for you also? At least, the appcast should be OK to reuse, assuming the two get updated at the same time.

### DIFF
--- a/Casks/flash.rb
+++ b/Casks/flash.rb
@@ -1,14 +1,16 @@
 cask 'flash' do
   version '20.0.0.306'
-  sha256 'e189a644a15cda617181e0ff2925488ef215ac2d74feb00800688f25bf77e7be'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
@adidalal Stellar work!

@andyli Usually updates them all (don’t forget [`pepper-flash`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/pepper-flash.rb)) at once so I’d say the appcast should be reusable, yes. Also, `pepper-flash` might also have a `pkg` inside its `app`.

adidalal:
Merged as 4214264d263c6d13ec3c36d832af03a1ed266aea.

Issue #18830, opened 2016-02-13
question: how to test a cask binary?


psibre:
Various casks provide `binary` stanzas to symlink some binary or script within a cask-installed app.
Would it be possible or desirable to provide some lightweight functionality to automatically test that the `binary` actually results in a working command?

For example, the `praat` cask creates a symlink
```
/usr/local/bin/praat -> /opt/homebrew-cask/Caskroom/praat/#{installed_version}/Praat.app/Contents/MacOS/Praat
```
It would be nice to automatically check when the cask is tested or installed that running `praat` actually does run that `.../Praat.app/Contents/MacOS/Praat` binary.

Because sometimes paths inside apps are changed upstream, and the symlinks would break on installing an updated cask. Or some other stuff happens upstream, and the binary itself no longer works.

Catching this sort of issue during the test builds, and guarding against it during `postflight` might be a good idea.

vitorgalvao:
I don’t really see this as desirable, because it really isn’t a problem. Also, with https://github.com/caskroom/homebrew-cask/issues/15603 this will only become less of a problem. That said, I also wouldn’t mind.

Lets see other opinions.

psibre:
I've floated a simple test balloon in #18879

vitorgalvao:
That’s not really the same thing as a test. You’re using `postflight` for it, mudding it’s usage. Either we make a proper test block for this, or we don’t do it at all.

psibre:
@vitorgalvao I agree that's not what the `postflight` is for; it was intended as a proof of concept.

```ruby
  binary 'Praat.app/Contents/MacOS/Praat', target: 'praat'

  # verify binary
  postflight do
    cmd = "praat --version"
    raise "Cannot run command `#{cmd}`" unless system cmd
  end
```
does provide the functionality I proposed above, though: it fails the installation if the `praat` command provided by the `binary` cannot be run.

Would it be possible to port the [`test` block](/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula) logic over from the main brew formula code?

vitorgalvao:
> Would it be possible to port the [`test` block](/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula) logic over from the main brew formula code?

[...]

Issue #18809, opened 2016-02-13
Implement methods for shim scripts


vitorgalvao:
@caskroom/maintainers This replaces https://github.com/caskroom/homebrew-cask/issues/9203. Everyone agrees shim scripts are desirable, so lets allow them.

On a first phase, if anyone submits a cask that employs a shim script (usually inside a `preflight` and using ruby methods to write to a file), we should (after determining it is indeed necessary and we have no other way to fix the problem) accept it and **add a comment mentioning it’s a shim** (this will allow us to find them easier to study/replace).

On a second phase (the point of this issue), we should implement something to make those easier to make. So, two points:
- Should we really implement something to make them easier, or should we just accept them in their usual rough form (has the advantage of not encouraging them, which is good since they should be used as a last resort)?
- Any suggestions for implementations/interface?

adidalal:
https://github.com/caskroom/homebrew-cask/blob/master/Casks/minecraft-server.rb - add comment
https://github.com/caskroom/homebrew-cask/pull/15876 - needs new PR w/ comment
https://github.com/caskroom/homebrew-cask/pull/9101 - needs new PR w/ comment

vitorgalvao:
#15876 doesn’t apply, because the shim isn’t necessary (the tool officially creates the necessary script).

psibre:
@vitorgalvao thanks for the heads up and for quickly merging #18827 -- I'll see if I can contribute a test and shim script

psibre:
A thought occurs: is it even possible to add some sort of test to ensure that the shim script works correctly? I'm thinking of something along the lines of the [`test` block](/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula) in homebrew formulas...

vitorgalvao:
@psibre No, we don’t have anything like that. Shim scripts should be discouraged, though (and only used in specific situations). This is not really related to this issue at hand (this issue is about deciding on an interface for shim scripts), so I ask that you open a new issue for further questions (every caskroom mantainer is likely getting an email to each reply to this issue, so we should really stay on topic).

adidalal:
What's the official comment for a shim script?

vitorgalvao:
> What's the official comment for a shim script?

I guess `# shim script (https://github.com/caskroom/homebrew-cask/issues/18809)` is OK.

PR #9101, opened 2015-01-19
blender: add and link wrapper script to enable command-line execution

psibre:
- this allows cask-installed blender to be used in external toolchains
- simply adding `binary 'Blender/blender.app/Contents/MacOS/blender'` is insufficient, since blender needs a custom `PYTHONPATH` pointing to its bundled python library
- a wrapper script is generated during `preflight`, then linked with `binary`

PR #18891, opened 2016-02-14
blender: add shim script for CLI use

psibre:
- allows blender installed with cask to be used in external toolchains
- simply adding `binary 'blender-#{version}-OSX_10.6-x86_64/blender.app/Contents/MacOS/blender'` is insufficient, since blender needs a custom `PYTHONPATH` pointing to its bundled python library
- wrapper shim script is generated during `preflight`, then linked with `binary`

### DIFF
--- a/Casks/blender.rb
+++ b/Casks/blender.rb
@@ -9,4 +9,17 @@ cask 'blender' do
 
   app "blender-#{version}-OSX_10.6-x86_64/blender.app"
   app "blender-#{version}-OSX_10.6-x86_64/blenderplayer.app"
+  # shim script
+  shimscript = "#{staged_path}/blenderwrapper"
+  binary shimscript, target: 'blender'
+
+  preflight do
+    pythonversion = '3.4'
+    File.open(shimscript, 'w') do |f|
+      f.puts '#!/bin/bash'
+      f.puts "export PYTHONHOME=#{staged_path}/blender-#{version}-OSX_10.6-x86_64/blender.app/Contents/Resources/#{version}/python/lib/python#{pythonversion}"
+      f.puts "#{staged_path}/blender-#{version}-OSX_10.6-x86_64/blender.app/Contents/MacOS/blender $@"
+      FileUtils.chmod '+x', f
+    end
+  end
 end


psibre:
supersedes #9101

psibre:
shim script could be tested upon installation, e.g.,
```ruby
  # verify binary
  postflight do
    require 'tempfile'
    tempfile = Tempfile.new('tmp.py')
    tempfile.write 'import bpy\nprint(bpy.app.version_string)'
    cmd = "blender --background --python #{tempfile.path}"
    raise "Cannot run command `#{cmd}`" unless system cmd
  end
```
but did not stage that hunk, pending a proper solution (see #18830)

Casks/blender.rb
```diff
@@ -9,4 +9,18 @@
...
+    pythonversion = "3.4"
+    File.open(shimscript, "w") do |f|
+      f.puts "#!/bin/sh"
+      f.puts "export PYTHONHOME=#{staged_path}/blender-#{version}-OSX_10.6-x86_64/blender.app/Contents/Resources/#{version}/python/lib/python#{pythonversion}"
+      f.puts "#{staged_path}/blender-#{version}-OSX_10.6-x86_64/blender.app/Contents/MacOS/blender $*"
+      FileUtils.chmod "+x", f
+    end
+  end
+
+  binary shimscript, :target => 'blender'
```
vitorgalvao: Should be `target: 'blender'` instead.

Casks/blender.rb
```diff
@@ -9,4 +9,18 @@
 
   app "blender-#{version}-OSX_10.6-x86_64/blender.app"
   app "blender-#{version}-OSX_10.6-x86_64/blenderplayer.app"
+
+  # shim script
+  shimscript = "#{staged_path}/blenderwrapper"
```
vitorgalvao: Why not have this inside the `preflight` block?

Casks/blender.rb
```diff
@@ -9,4 +9,18 @@
...
+  preflight do
+    pythonversion = "3.4"
```
vitorgalvao: Every value here that doesn’t require interpolation should be using single quotes.

Casks/blender.rb
```diff
@@ -9,4 +9,18 @@
...
+    File.open(shimscript, "w") do |f|
+      f.puts "#!/bin/sh"
```
vitorgalvao: On OS X, `/bin/sh` is `/bin/bash` but worse, it should be `/bin/bash`  instead.

Casks/blender.rb
psibre: Because then we Don't Repeat Ourselves in the `binary` stanza... of course I can change it if you prefer.

vitorgalvao: Oh, right, makes sense.

Casks/blender.rb
```diff
@@ -9,4 +9,18 @@
 
   app "blender-#{version}-OSX_10.6-x86_64/blender.app"
   app "blender-#{version}-OSX_10.6-x86_64/blenderplayer.app"
+
+  # shim script
+  shimscript = "#{staged_path}/blenderwrapper"
+  preflight do
+    pythonversion = '3.4'
+    File.open(shimscript, 'w') do |f|
+      f.puts #!/bin/sh'
```
vitorgalvao: Forgot to open the quotes (before the `#`). Still, should be `/bin/bash`.

psibre:
OK, commits are piling up. I'll be happy to squash if and when it's acceptable... =)

Casks/blender.rb
```diff
@@ -9,4 +9,18 @@
...
   app "blender-#{version}-OSX_10.6-x86_64/blenderplayer.app"
+
+  # shim script
+  shimscript = "#{staged_path}/blenderwrapper"
+  preflight do
+    pythonversion = '3.4'
+    File.open(shimscript, 'w') do |f|
+      f.puts '#!/bin/bash'
+      f.puts "export PYTHONHOME=#{staged_path}/blender-#{version}-OSX_10.6-x86_64/blender.app/Contents/Resources/#{version}/python/lib/python#{pythonversion}"
+      f.puts "#{staged_path}/blender-#{version}-OSX_10.6-x86_64/blender.app/Contents/MacOS/blender $*"
```
vitorgalvao: It should be `"$@"` (quoted as it is), here. Using `$*` will break in many cases ([see differences](http://stackoverflow.com/a/12316565/1661012)).

psibre: indeed, thanks for the hint!

vitorgalvao: After this change, guess you can squash.

psibre:
OK, I just replaced all the incremental commits by a single new one.

vitorgalvao:
Just realised something. Travis will break here for sure (wrong [stanza order](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/readme.md#stanza-order)).

Instead of

```ruby
app
app

shimscript

preflight

binary
```

try

```ruby
app
app
shimscript
binary

preflight
```

And run `brew cask style {{your_cask}} after (if there are problems, it’ll find them).

psibre:
Thanks, I did notice all the CI failures and was wondering what was going on!

psibre:
but I don't understand the failures on travis...

vitorgalvao:
I’ve seen this Travis hanging before, in otherwise correct changes (like https://github.com/caskroom/homebrew-cask/pull/18879). Can you shed some light, @jawshooah?

Casks/blender.rb
```diff
@@ -9,4 +9,17 @@
 
   app "blender-#{version}-OSX_10.6-x86_64/blender.app"
   app "blender-#{version}-OSX_10.6-x86_64/blenderplayer.app"
+  # shim script
```
vitorgalvao: Mind [changing it for consistency](https://github.com/caskroom/homebrew-cask/issues/18809#issuecomment-184373999)?

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 40754ecb682d64c8b4d1e8d23610b18234647a8b. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

Casks/blender.rb
psibre: you mean actually inserting the full URL of #18809?

psibre: never mind, already fixed in 40754ecb682d64c8b4d1e8d23610b18234647a8b. Thanks @adidalal !

PR #18894, opened 2016-02-14
Updated carbon-copy-cloner (4.1.7.4285)

ptb:
The previous appcast only had an entry for version 3. I tried to determine what the new version 4 appcast url should be, but since it's https I wasn't able to sniff it.

### DIFF
--- a/Casks/carbon-copy-cloner.rb
+++ b/Casks/carbon-copy-cloner.rb
@@ -1,10 +1,8 @@
 cask 'carbon-copy-cloner' do
-  version '3.4.7'
-  sha256 '4ea19d1c9111fb67b0f7567844a413794bd2d3b6a908a8921cabd10f21101247'
+  version '4.1.7.4285'
+  sha256 'ca8a081d35606c899745bb40a64a461e14a00de7dd6e3bb48da3a3c05f265de8'
 
-  url "http://www.bombich.com/software/download_ccc_update.php?v=#{version}"
-  appcast 'https://bombich.com/software/updates/ccc.php',
-          checkpoint: 'eec027f0d0aa576e292b4fe57441f1821409c6295330787dd82533b0d483cb26'
+  url 'http://bombich.com/software/download_ccc.php?v=latest'
   name 'Carbon Copy Cloner'
   homepage 'https://bombich.com/'
   license :commercial
--- a/Casks/carbon-copy-cloner.rb
+++ b/Casks/carbon-copy-cloner.rb
@@ -1,6 +1,6 @@
 cask 'carbon-copy-cloner' do
-  version '4.1.7.4285'
-  sha256 'ca8a081d35606c899745bb40a64a461e14a00de7dd6e3bb48da3a3c05f265de8'
+  version :latest
+  sha256 :no_check
 
   url 'http://bombich.com/software/download_ccc.php?v=latest'
   name 'Carbon Copy Cloner'


ptb:
According to the Info.plist inside the .app, the previously listed URL "https://bombich.com/software/updates/ccc.php" is the SUFeedURL, which unless it's also sending a cookie or something else I'm not sure how it would be working correctly.

Casks/carbon-copy-cloner.rb
```diff
@@ -1,10 +1,8 @@
 cask 'carbon-copy-cloner' do
-  version '3.4.7'
-  sha256 '4ea19d1c9111fb67b0f7567844a413794bd2d3b6a908a8921cabd10f21101247'
+  version '4.1.7.4285'
+  sha256 'ca8a081d35606c899745bb40a64a461e14a00de7dd6e3bb48da3a3c05f265de8'
 
-  url "http://www.bombich.com/software/download_ccc_update.php?v=#{version}"
-  appcast 'https://bombich.com/software/updates/ccc.php',
```
vitorgalvao: [`find_sparkle_appcast`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/find_sparkle_appcast) is exactly to find this. Just used it, and it returned this same address, though.

Instead of removing it, mind contacting the developer and ask if they plan on updating it, have a new one?

Casks/carbon-copy-cloner.rb
```diff
@@ -1,10 +1,8 @@
...
-          checkpoint: 'eec027f0d0aa576e292b4fe57441f1821409c6295330787dd82533b0d483cb26'
+  url 'http://bombich.com/software/download_ccc.php?v=latest'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version and there’s no `appcast`, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

Casks/carbon-copy-cloner.rb
ptb: I will contact the developer and see what they say.

vitorgalvao: Thank you. In the meantime, after Travis passes we can merge this.

adidalal:
Got a versioned URL: `http://bombich.com/software/files/ccc-4.1.7.4285.zip`

(from http://bombich.com/software/download_ccc.php?v=latest&l=alternate

ptb:
> Hi Peter,
> 
> That's the right URL, but it's expecting some parameters because we customize the result depending on the user's OS and CCC version. For example:
> 
> https://bombich.com/software/updates/ccc.php?os_minor=11&os_bugfix=4&ccc=4285&beta=0&locale=en
> 
> That would deliver the latest version of CCC, 4.1.7.
> 
> Mike
> 
> Mike Bombich
> Bombich Software, Inc.

ptb:
What would be the best way moving forward for this?

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 4dd860ee5e3ad357e717afbf1547c1b5a2768fa9. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

vitorgalvao:
And thank you for contacting them and getting the appcast.

PR #18897, opened 2016-02-14
cask_language_reference, stanza order: appcast has checkpoint, not sha256

vitorgalvao:


adidalal:
Merged as 3e69bcf88d7600df83b514443eb000e6b19303d2.

PR #18904, opened 2016-02-14
updated djview (4.10.4.b)

miccal:


Casks/djview.rb
```diff
@@ -1,8 +1,8 @@
 cask 'djview' do
-  version '4.10.2'
-  sha256 '71023ad00941f7d3a01f273db4dc2e5c9c12b621954d877b9fe34a9ecf4bf149'
+  version '4.10.4.b'
```
vitorgalvao: Is this a beta?

miccal: Not that I am aware of - the [website](http://djvu.sourceforge.net/djview4.html) says the latest version is 4.10.5, but the download [link](https://sourceforge.net/projects/djvu/files/) states that the latest version is `DjVuLibre-3.5.27+DjView-4.10.4.b-intel64.dmg`. It does not mention that it is a beta.

vitorgalvao:
Thank you.

miccal:
Weird - the installed version is indeed 4.10.5.

PR #18907, opened 2016-02-15
Add GameRanger

Eitot:


vitorgalvao:
If it `auto_updates`, then there has to be an `appcast`. Mind contacting the developer about it? Something like (feel free to use this verbatim):

Hello,

I’m currently adding your app details to homebrew-cask (a popular command-line tool to facilitate the installations of GUI apps), and if I can also add the appcast (the URL your app uses to check for updates) to it, the homebrew-cask developers will be able to better keep the details of your app (namely the current version at any given point) up to date more easily, which means users will be sure to get the latest version of your app when installing it in this manner.

Could you then please provide me with the appcast URL your app uses?

If you have further questions, please feel free to ask about it on the repo (https://github.com/caskroom/homebrew-cask) and the team will answer any doubts you might be having.

Thank you for your time. Have a nice day.

adidalal:
I'm vaguely familiar with GameRanger - it downloads pieces (plugins) per app, and doesn't have an appcast per se.

That being said, technically `http://www.gameranger.com/games/mac/` can be used as an appcast, if one is needed for `auto_updates`

vitorgalvao:
> I'm vaguely familiar with GameRanger - it downloads pieces (plugins) per app, and doesn't have an appcast per se.

That’s fine, then. But then `auto_updates` isn’t really applicable, is it? `auto_updates` should consider the core of the app (i.e. what homebrew-cask actually gets). If the app itself then gets plugins, that’s out of our jurisdiction.

> That being said, technically `http://www.gameranger.com/games/mac/` can be used as an appcast, if one is needed for `auto_updates`

It isn’t strictly necessary (the code still works), but it’d be nice. Better not use that page though (remember the other HTML pages we tried).

Eitot:
I just misunderstood the idea of `auto_updates`. The application updates itself and requires a constant Internet connection. There is no appcast or like. I don't think the developer even provides up-to-date initial installs. I'll remove the stanza.

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 332125a2193320b3324d0d9b9cc7e12907001378. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #18908, opened 2016-02-15
switch mactex (20150613) back to versioned download

adidalal:
Found a suitable HTML page for an appcast (Not that the package changes very often, releases are once a year)


adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 8038842f97c60578b72c34a4ef1cec312faf6c32. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #18913, opened 2016-02-15
Nimble 1.0.3

zzamboni:
Nimble: http://madebybright.com


vitorgalvao:
[Already here](https://github.com/caskroom/homebrew-cask/blob/master/Casks/nimble.rb).

PR #18931, opened 2016-02-15
upgrade HotS Replay Uploader to v2.0.2

zhedar:


adidalal:

adidalal:
Did it for you in https://github.com/caskroom/homebrew-cask/pull/18932 as I was at a computer. Info above still valid for future updates, though :)

zhedar:
Thanks, I wondered about that bit before the commit, but than ignored it.
Walked through cask-repair, makes upgrading smoother this way :)

PR #18946, opened 2016-02-16
updated findings (latest)

mskblackbelt:


victorpopkov:
Thank you for your contribution. However, the download URL for the latest version (`http://downloads.findingsapp.com/findings`) redirects exactly to the same versioned URL used in the cask before (`http://downloads.findingsapp.com/Findings_3077_1.3.2.zip`).

The versioned casks are currently more preferred, and `:latest` should be used only if there is no other official versioned URLs available. So there is no point in these changes at the moment.

mskblackbelt:
Sorry, wanted to try out the automated script. The update was triggered by a sha256 mismatch on the latest download, thought that perhaps changing to `:latest` was an ok way to circumvent. In this case, I'll submit a pull request with the updated hash.

victorpopkov:
Thank you for the contribution, @mskblackbelt. You can make all changes mentioned in the same PR, if you want.

victorpopkov:

mskblackbelt:
Ok, should be all set…

victorpopkov:
Thank you for the contribution. It was merged directly as b8558055436f1ca0d3091697e3a371985b5218ef to keep commit history cleaner. Your contribution is still credited to you.

PR #18960, opened 2016-02-16
Update facter version to 2.4.6

moutons:


adidalal:
Fixed in https://github.com/caskroom/homebrew-cask/pull/18990 this time

PR #18961, opened 2016-02-16
Update puppet version to 3.8.6

moutons:


Casks/puppet.rb
```diff
@@ -1,5 +1,5 @@
 cask 'puppet' do
-  version '3.8.2'
+  version '3.8.6'
   sha256 '8ff5746e5d3c121b947f27dabeea81766d902885df0d113af4080085a45aff84'
```
victorpopkov: The sha256 also needs to be updated here. You can get it using:
```bash
shasum -a 256 ~/Downloads/puppet-3.8.6.dmg 
````
For Puppet v3.8.6 it's:
```ruby
  sha256 '25ab90342614b8875e4fac0c1421ea9df8fc63c999f4ef66ad41fdf559f933cf'
```

adidalal:
https://github.com/caskroom/homebrew-cask/pull/18991

PR #18982, opened 2016-02-17
Add support for Fujitsu ScanSnap Manager for S1500.

joliver:


victorpopkov:
Thank you for your contribution. However, this belongs in [caskroom/versions](https://github.com/caskroom/homebrew-versions). You’ll find similar casks there including the one used for S1500 ([scansnap-manager-s1500.rb](https://github.com/caskroom/homebrew-versions/blob/master/Casks/scansnap-manager-s1500.rb)).

PR #19070, opened 2016-02-19
Remove noizio - MAS-only

Amorymeltzer:
Claimed they'd post a direct link on website, but no such luck
https://twitter.com/noizio/status/692121584358789122


vitorgalvao:
> Claimed they'd post a direct link on website, but no such luck

Often the case, I find. I never trust developers when they make that promise. Unless they have a vested interest in providing a direct link<sup>1</sup> and they personally<sup>2</sup> commit to doing so for a specific reason<sup>3</sup>, I won’t trust them to do so and prefer to remove it outright.

---

<sup>1</sup> Often that means their users asking for one for inclusion in homebrew-cask.
<sup>2</sup> And I do mean an actual person you can identify by name, not an account with the name/image of the company/app.
<sup>3</sup> Not relevant if it’s for homebrew-cask or other thing, but we should be able to point to something specific.

PR #19077, opened 2016-02-20
updated omnigraffle (6.5)

liljenstolpe:


liljenstolpe:
I've checked and the url that failed to resolve in Job #37075.3 is valid and resolvable.  It's the same base hostname that succeeded in the other two jobs for the PR.

PR #19084, opened 2016-02-20
updated framer-studio (latest)

vitorgalvao:


adidalal:
There should be an appcast - I'll take a look when I get a chance.

vitorgalvao:
Trouble with the previous `appcast` is it’s no longer being updated (stopped at `48`). `find_sparkle_appcast` doesn’t find one here.

adidalal:
Use `https://updates.devmate.com/com.motif.framer.xml`

vitorgalvao:
Thank you. Is that a common pattern for `devmate`-hosted apps?

adidalal:
Not sure, just sniffed it with Charles.

adidalal:
Nope, not a common pattern - depends entirely on how much the dev wants to use devmate. It's a full service app distribution service, but people still use sparkle, hockeyapp, etc..

PR #19089, opened 2016-02-20
updated framer-studio (52)

vitorgalvao:


adidalal:
`https://dl.devmate.com/com.motif.framer/FramerStudio.zip` will always give the latest - should be OK to use because we have the appcast. The other link might be a delta update (at least, that's my suspicion.

PR #19090, opened 2016-02-20
Updated typora (latest)

hevets:


adidalal:

adidalal:
should be `e3c717ecc8179492764c5f0017451c68f3b4bc8605cac28fd2818de684c8e62f`

Looks like you did use cask-repair though, which is interesting. Wonder why it didn't update

adidalal:
@hevets Please resubmit with a versioned URL.

`cask-repair -v 0.9.8.5 typora`

vitorgalvao:
> Looks like you did use cask-repair though, which is interesting. Wonder why it didn't update

No idea. Just tried it and it worked as usual.

PR #19091, opened 2016-02-20
updated typora (latest)

hevets:


hevets:
no idea why the `appcast` isn't being updated by the `cask-repair` script. I can close and resubmit properly if you'd like

PR #19092, opened 2016-02-20
Update typora (latest)

hevets:


Casks/typora.rb
```diff
@@ -1,10 +1,10 @@
 cask 'typora' do
-  version '0.9.8.1'
```
vitorgalvao: Don’t remove the version, please, it is useful.

vitorgalvao:
Thank you for reporting the outdated cask. Will be updating it in a different PR (was checking `cask-repair`).

Casks/typora.rb
hevets: For sure, out of curiosity what's the best way to detect latest version of app. Only added latest to match the download URL. 

Thanks

PR #19094, opened 2016-02-20
updated framer-studio (52)

adidalal:
@vitorgalvao Original URL is fine, as linked on homepage.

### DIFF
--- a/Casks/framer-studio.rb
+++ b/Casks/framer-studio.rb
@@ -1,9 +1,11 @@
 cask 'framer-studio' do
-  version '50'
-  sha256 '1649c2768d66c0e0c45345a05f43e095852bc26a2c54bc0df7ef6b08d023bdc1'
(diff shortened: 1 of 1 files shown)


adidalal:
Merged as a70992d1b294eec65262d9d59dbb5f7f4ba7c066.

adidalal:
I think here, you won't need to update the URL every time upon update, just version/appcast checkpoint.

Makes more sense to me as it's one less thing to pass into cask-repair, especially since "breaking" casks is less of an issue now (through the new scripts). That being said, I don't even think I've seen a Cask stay broken for more than a day.

vitorgalvao:
> That being said, I don't even think I've seen a Cask stay broken for more than a day.

Can’t argue with that. Not even a full day, likely! We have indeed become very efficient in this regard.

PR #19097, opened 2016-02-20
updated airfoil (5.0.0)

adidalal:


vitorgalvao:
Great work with the appcasts, by the way. Are you sniffing traffic, or just looking hard for them?

adidalal:
Sniffing traffic - figured it's a one-time thing, for basically automating updates in the future.
No real order to apps I'm doing, just as I stumble upon them/have free time.

Adding appcasts would be an excellent "easy" contribution to HBC - I may write up something about it if there's interest.

adidalal:
Merged as d0fcbe2d6da4f35aa29ade61d9e6dd0ce002ea2e.

vitorgalvao:
> Adding appcasts would be an excellent "easy" contribution to HBC

Agreed. Time-consuming but somewhat straightforward tasks fit the bill very well.

> I may write up something about it if there's interest.

I, for one, am interested. I’ve dabbled in it in the past but didn’t get very far then. I quickly abandoned the thought of pursuing it further after [this comment](https://github.com/caskroom/homebrew-cask/issues/260#issuecomment-21370401).

If you write something up, I’ll try it myself a bit more in depth and might even look into CLI (unless you’re already doing it that way, and not with a GUI) ways of doing so and try to write a semi-automatic (`cask-repair`-like) script geared for this.

adidalal:
Unfortunately, little chance of being automated in the current form I'm doing things.
A tl;dr:

- Use Charles, install & trust SSL root certificate, set SSL bypass to *.*
- Download app in question
- Click on check for updates, inspect requests in Charles
- Try out stripped version of URL (minimum parameters), verify by wget & sublime
- Add appcast/checkpoint to Cask with cask-repair


Thoughts on automation:
https://mitmproxy.org is a CLI, free alternative to charles

(could all be done in a VM, to keep things clean)
Type in app name
Downloads/installs cask
Could use applescript to automate "clicking" on check for updates
Spits out a few urls to try
User select URL, tests it
Feeds URL back to script, added as appcast, checkpoint automatically calculated, PR submitted, etc.

PR #19105, opened 2016-02-21
Update BZFlag to latest version

allejo:
We have moved away from SourceForge and on to [GitHub](https://github.com/BZFlag-Dev/bzflag) but our binaries are [self-hosted](http://bzflag.org/downloads).

### DIFF
--- a/Casks/bzflag.rb
+++ b/Casks/bzflag.rb
@@ -1,9 +1,8 @@
 cask 'bzflag' do
-  version '2.4.2'
-  sha256 '4ce7f007cb7c78a5243ced84d310c37637a2aa8fe9569231edd2dbc45286e293'
+  version '2.4.4'
+  sha256 'c832896a65cd3a0bdca62fce8a0fff800c15970dfde75e9f8285e49a01dafc77'
 
-  # sourceforge.net/sourceforge/bzflag was verified as official when first introduced to the cask
-  url "http://downloads.sourceforge.net/sourceforge/bzflag/BZFlag-#{version}.zip"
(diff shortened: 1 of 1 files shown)


adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 5e8a3a9e1c6f8feeace7a46c3dd3c2c8ed1af566. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

allejo:
Thanks! Regarding `appcast`, would edits (e.g. typo fixes, etc.) to releases break things since the SHA256 is for the release file as it is right now?

adidalal:
appcast is purely informational; it lets us (in an automated fashion) detect that there may be an app update available.

So yes, the checkpoint would change, but it's not a bag deal; we have many Casks that tag beta releases (which changes the appcast), but we do manually make sure the release version is the one we use.

That being said, if you have a better appcast source, feel free to change it with another PR!

allejo:
Thanks for the info. GitHub would be the most accurate appcast source we have at the moment so I'll proof read before I make a typo!

adidalal:
Not a problem, but don't let that discourage you from making many new releases/fixing typos! We'll normally pick up appcast changes within a week, or you can submit a new PR with [`cask-repair`](https://github.com/vitorgalvao/tiny-scripts/blob/master/cask-repair).

It can at once update the cask’s version (according to what you give it) taking care of updating both the cask’s `sha256` and the appcast’s `:checkpoint`, while still giving you the chance to perform other manual fixes.

----
Also, I would probably delete your fork (locally and on GH) and re-fork, following the guide in CONTRIBUTING for further PRs, making changes on the master branch of a fork leads to merge errors.

PR #19146, opened 2016-02-23
update Apogee Duet pkg name

yutarody:
Modify .pkg name


victorpopkov:
Thank you for the contribution. It was merged directly as e76c74292d8e208312f4917fc872681bee816e5b to keep commit history cleaner. Your contribution is still credited to you.

PR #19151, opened 2016-02-23
macvim 7.4.97

sjackman:


vitorgalvao:

sjackman:
I've updated the `appcast` SHA. The `sed` didn't seem to be necessary in this case as the `atom` file didn't contain the string `pubDate`. I used:
```sh
curl https://github.com/macvim-dev/macvim/releases.atom |gsha256sum
bcc7df81b062d308bfa87c8b13b89c618ab02c9af288bbaa3382a2a6e71a1628
```

sjackman:
What is the purpose of the `appcast`?

vitorgalvao:
>  The `sed` didn't seem to be necessary in this case as the `atom` file didn't contain the string `pubDate`.

It isn’t, but that command covers all cases.

> What is the purpose of the `appcast`?

It’s used by maintainers (we’re ramping up usage) to quickly gather which casks are out of date. It gives us a more consistent way to update casks.

vitorgalvao:
Thank you for the contribution. It was merged directly as 258f361d1f112aaf70de8056a0101c8c192b00af to keep commit history cleaner. Your contribution is still credited to you.

sjackman:
Thanks, Vítor.

PR #19160, opened 2016-02-24
Install Additional Emacs Utility Binaries

lenary:
When installing Emacs, it's nice to also be able to use `emacsclient` etc too. Now cask knows about them, these utilities get installed in the right place.

There may be other things we should be symlinking that get installed, but I don't know what they are.

### DIFF
--- a/Casks/emacs.rb
+++ b/Casks/emacs.rb
@@ -8,4 +8,9 @@ cask 'emacs' do
   license :oss
 
   app 'Emacs.app'
+  binary 'Emacs.app/Contents/MacOS/bin/emacsclient', target: 'emacsclient'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
`target:` is unnecessary in al of these. Please remove it.

lenary:
@vitorgalvao Done, I wasn't sure if they were needed, but put them in to be on the safe side. Gone now.

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in c06554e352b975aa4577f675234441ef3eaefcc3. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

lenary:
:+1: thanks! :D

PR #19192, opened 2016-02-24
Update vagrant-manager to 2.5.3

shian48263:


adidalal:
Please make PRs from a branch, instead of master, as that normally that leads to issues (ie, 3 commits instead of 1, in this case) - cask-repair can help you out here! Feel free to close this PR/delete/refork/try again, if it's easier than amending.

[`cask-repair`](https://github.com/vitorgalvao/tiny-scripts/blob/master/cask-repair) can at once update the cask’s version (according to what you give it) taking care of updating both the cask’s `sha256` and the appcast’s `:checkpoint`, while still giving you the chance to perform other manual fixes.

shian48263:
Ok, thanks for the information.

I'll try to create another cleaner PR.

PR #19194, opened 2016-02-24
Update minecraft launcher url

Jerzerak:


Casks/minecraft.rb
```diff
@@ -3,7 +3,7 @@
   sha256 :no_check
 
   # amazonaws.com is the official download host per the vendor homepage
```
vitorgalvao: Change to `# mojang.com was verified as official when first introduced to the cask`.

Jerzerak:
Just realized, would you like the commits squashed or are they fine as is?

vitorgalvao:
We always prefer them squashed.

Jerzerak:
Squashed as requested

PR #19205, opened 2016-02-25
Added sublime-text3

vitaliel:


jawshooah:
[Already in the versions repo.](https://github.com/caskroom/homebrew-versions/blob/master/Casks/sublime-text3.rb).

PR #19221, opened 2016-02-26
Add Ironic Leap

yutarody:
http://www.ironicsoftware.com/leap/
- Delete License Comment
- Fix URL

### DIFF
--- a/Casks/Ironic-Leap.rb
+++ b/Casks/Ironic-Leap.rb
@@ -1,0 +1,11 @@
+cask 'Ironic-Leap' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.ironicsoftware.com/downloads/Leap.zip'
+  name 'Leap'
+  homepage 'http://www.ironicsoftware.com/leap/'
+  license :commercial
+
+  app 'Leap.app'
+end


Casks/Ironic-Leap.rb
```diff
@@ -0,0 +1,11 @@
+cask 'Ironic-Leap' do
```
victorpopkov: This has to be changed to `cask 'leap' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `leap.rb`.

Casks/Ironic-Leap.rb
```diff
@@ -0,0 +1,11 @@
...
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.ironicsoftware.com/downloads/Leap.zip'
+  name 'Leap'
```
vitorgalvao: Should be `name 'Ironic Software Leap'`.

Casks/Ironic-Leap.rb
vitorgalvao: As a further note, both this and the file name must always be lowercase.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in fa54267118a57ade51b03ef957c882a4edf31134. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #19223, opened 2016-02-26
Add sqltabs.

athrunsun:
SQL Tabs is an open source cross platform desktop client for Postgresql with:
- Database explorer
- Rich scripts output
- Charts from query result
- Markdown rendering
- Dark and bright themes
- Classic and vim editing modes

and many more useful features.

This tool is based on electron.

### DIFF
--- a/Casks/sql-tabs.rb
+++ b/Casks/sql-tabs.rb
@@ -1,0 +1,13 @@
+cask 'sql-tabs' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.sqltabs.com/download/SQL%20Tabs.app.zip'
+  name 'SQL Tabs'
+  homepage 'http://www.sqltabs.com/'
+  license :gpl
+
+  app 'SQL Tabs.app'
+
+  zap delete: '~/.sqltabs'
+end


athrunsun:
"curl: (6) Could not resolve host: www.sqltabs.com", how can I just trigger build once more? Looks like DNS issue.

Casks/sqltabs.rb
```diff
@@ -0,0 +1,13 @@
+cask 'sqltabs' do
```
victorpopkov: This has to be changed to `cask 'sql-tabs' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `sql-tabs.rb`.

Casks/sqltabs.rb
```diff
@@ -0,0 +1,13 @@
...
+  version '0.12'
```
victorpopkov: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/master/Casks/010-editor.rb).

vitorgalvao:
In addition to the comments, please [squash your commits](http://davidwalsh.name/squash-commits-git).

athrunsun:
@victorpopkov @vitorgalvao Changes are made according to code review comments and commits are squashed. Thanks for all your comments to make it a successful build with correct styles!

vitorgalvao:
Thank you for the contribution. It was merged directly as b4f235b347ab7530c95a8edefc7b901baa034aaa to keep commit history cleaner. Your contribution is still credited to you.

PR #19238, opened 2016-02-26
Add Basecamp 3.app v1.0.2

cameronmalek:
Just released in the last 24 hours. No known versioned download URL.

### DIFF
--- a/Casks/basecamp.rb
+++ b/Casks/basecamp.rb
@@ -1,0 +1,14 @@
+cask 'basecamp' do
+  version :latest
+  sha256 :no_check
+
+  # bc3-desktop.s3.amazonaws.com was verified as official when first introduced to the cask
+  url 'https://bc3-desktop.s3.amazonaws.com/mac/basecamp3.dmg'
+  name 'Basecamp 3'
+  homepage 'https://basecamp.com/help/3/guides/apps/mac'
+  license :gratis
+
+  auto_updates true
+
+  app 'Basecamp 3.app'
+end


adidalal:
There is, because there's a check for updates in the menu. Unfortunately, it seems to be inactive until you sign in to the app.

If you have Charles/mitmproxy, mind sniffing around? https://github.com/caskroom/homebrew-cask/issues/19103 has some pointers. (if not, it's also fine, just add in `auto_updates true` to the Cask).

Casks/basecamp.rb
```diff
@@ -0,0 +1,12 @@
+cask 'basecamp' do
+  version :latest
+  sha256 :no_check
+
+  # amazonaws.com is the official download host per the vendor homepage
```
vitorgalvao: Should be `# bc3-desktop.s3.amazonaws.com was verified as official when first introduced to the cask`, now.

cameronmalek: Thanks @vitorgalvao

cameronmalek:
@adidalal I went through mitmproxy line by line with a fresh first-time install of the app and it found nothing resembling an appcast feed (or even anything ambiguous that *could* be an update feed of some kind) even though the app now knows it's up to date, confirming it actually was able to check somehow.

cameronmalek:
Updated! Thanks for the help.

adidalal:
@cameronmalek Thanks for checking, anyways!

adidalal:
Thank you for the contribution. It was merged directly as 09054a23ab6478ecd51768dc276a0aa76f0a3561 to keep commit history cleaner. Your contribution is still credited to you.

PR #19251, opened 2016-02-27
Update Polymail to v0.75

schneidmaster:
The v0.73 zip no longer exists at the current URL and causes `brew cask install polymail` to error.

### DIFF
--- a/Casks/polymail.rb
+++ b/Casks/polymail.rb
@@ -1,10 +1,10 @@
 cask 'polymail' do
-  version '0.73'
-  sha256 '4e9a9f23d9ecaed00c42dd59203c81cc8e54c374dde6c159499583d313c1a45f'
+  version '0.75'
+  sha256 'b5cd6b4acbfa3f965aaedc65803f75714bb7df2fb3f33b50c1e07419f9cd8c1d'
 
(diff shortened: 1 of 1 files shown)


schneidmaster:
Not sure why the test failed --

> cannot use the sha256 for an empty string: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855

Any insights?

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 92abf3d57aa581da6dfcc56614747971508a4f9d. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

adidalal:
Your appcast checkpoint was wrong, looks like

schneidmaster:
Ah okay. For future reference how is the appcast checkpoint calculated? I thought I was following the docs and calculated it with:

```
curl --compressed "{{https://sparkle-updater.polymail.io/cast.xml}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256
```

vitorgalvao:
You’re not supposed to actually put the `{{}}`, just the url.

schneidmaster:
Ah got it, thanks.

PR #19274, opened 2016-02-27
outdated_appcast scripts: new model

vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/issues/19121.

@caskroom/maintainers The `outdated_appcasts` scripts, for those of you that use/plan on using them, now (after this is merged) use a different model with multiple advantages:
- Requires less of a time commitment.
- You no longer need to warn other maintainers when you plan on running them.
- You no longer need to run the whole stack.

This means you may just, for example, run `find_outdated_appcasts` and let it run, without having to worry about it flooding the repo with new issues. Or you may just run `fix_outdated_appcasts` and fix just one or two. This ideally will make them be run more often by more maintainers, which means faster updates to casks.

How the new model works (from https://github.com/caskroom/homebrew-cask/issues/19121#issuecomment-187457577):

`find_outdated_appcasts`:
- When starting, check if there’s an `outdated appcast`-labeled issue.
  - If yes and it is assigned to someone, try next repo.
  - If yes and it is not assigned to anyone, recommend running `fix_outdated_appcasts`. Try next repo.
  - If no, continue.
- Open a new empty issue labeled `outdated appcast` and assign it to yourself.
- Check all appcasts.
  - If aborted, close the empty issue.
- Edit issue with all outdated appcasts.
- Unassign yourself.

`fix_outdated_appcasts`:
- When starting, check if there’s an `outdated appcast`-labeled issue.
  - If yes and it is assigned to someone, try next repo.
  - If yes and it is not assigned to anyone, assign yourself. Continue.
  - If no, try next repo.
- Get all outdated casks in issue.
- Start fixing them.
  - If aborted, get list of already fixed casks and remove them from issue. Unassign yourself.
- If finished, close issue. Unassign yourself.

As you can see from the diff, this required more additions than edits to the previous scripts, and I worry a bit some parts may not be as clear. `bash` really showed a weaknesses and inadequacy for this new model, as did `ghi` in some parts. Many parts need to be so hacky for now, I’m not sure this is completely worth the change, but testing will tell.


adidalal:
Thanks for doing this, @vitorgalvao! I'm away this weekend, but will provide feedback in a couple of days.

vitorgalvao:
Made a small but important change. Smallest repos are now done before big ones. This means the huge amount of casks in this main repo won’t affect the others.

adidalal:
@vitorgalvao How do you have appcasts opening up in your editor? Getting annoyed of Safari always hijacking the URL and offering to add it as a feed (so much so that I manually edit out the `open-appcast` command)

vitorgalvao:
> How do you have appcasts opening up in your editor?

Chrome displays the appcasts, for me.

vitorgalvao:
The more I use this new method, the less I like it. It’s too fidgety, unfortunately, and I’m not sure yet if that’s on `ghi` or Github’s issues API.

What I do know, for now, is that I’m not liking the unreliability. I’m finding way too many edge cases.

The problem with the previous method was the amount of noise it created, so we should rethink how to fix that.

adidalal:
How about making issues in an empty repo?

vitorgalvao:
That’s the best idea I’ve been able to come up with as well, but it feels inelegant. Apart from that, it might also push us back again to “one developer doing every step”, so we also need a solution for it.

Ideally, some sort of centrally controlled file we can all write to might work, but the only thing I can remember with such a characteristic would involve writing a file to the repo itself. That might not be worse than opening issues, though.

vitorgalvao:
Closing. As bad as it is to throw away the work, I don’t think this is the right solution.

I’ve been playing in my head with a completely different solution for a while, now. I won’t be working on it too soon, but at least I already have some concepts.

PR #19315, opened 2016-02-27
updated kube-cluster (0.2.0)

victorpopkov:


vitorgalvao:
Mind removing these from https://github.com/caskroom/homebrew-cask/issues/19246 as you go, so we don’t have duplicate efforts?

victorpopkov:
Merged as 349b0a54a59f82677af745354e77c0e288a3302a.

PR #19324, opened 2016-02-27
:=sosha_evanS2005.@/.comupdated godot (2.0)

earlevans:


Casks/godot.rb
```diff
@@ -1,9 +1,11 @@
 cask 'godot' do
-  version '1.1'
```
earlevans: sha256

earlevans: 1.1

adidalal:
https://github.com/caskroom/homebrew-cask/commit/bda8d5993400ecffb7ab6024141282ce41029509

vitorgalvao:
Damn bots. It’s even the same commit. It had already made some nonsense comment early on, in another issue.

adidalal:
Thought so - cleaned up, regardless.

PR #19326, opened 2016-02-28
Add TonePrint.app v3.1

drewble:
TonePrint Editor by TC Electronics is an editor effect settings for
their line of TonePrint guitar pedals. Surely there is enough of an
overlap between developers and guitar players that this command has been
tried before, but no one has created a cask. More fun for me!

### DIFF
--- a/Casks/toneprint.rb
+++ b/Casks/toneprint.rb
@@ -1,0 +1,11 @@
+cask 'toneprint' do
+  version '3.1'
+  sha256 :no_check
+
+  url 'http://cdn-downloads.tcelectronic.com/media/5607727/toneprint-3100-r1373.dmg'
+  name 'TonePrint Editor'
+  homepage 'http://www.tcelectronic.com/toneprint-editor/support/'
+  license :unknown # TODO: change license and remove this comment; ':unknown' is a machine-generated placeholder
+
+  app 'TonePrint.app'
+end


adidalal:
@vitorgalvao This one uses a SOAP request to get version information, response in the `		<a:LatestVersion>` key.

cURL-able request below. Thoughts on how to add it in?
```
curl -H "Host: tp.tcelectronic.com" -H "Proxy-Connection: keep-alive" -H "Accept: */*" -H "Content-Type: text/xml; charset=utf-8" -H "Accept-Language: en-us" -H "SOAPAction: "http://tempuri.org/ITonePrintService/GetVersionsAndInfo"" -H "User-Agent: TonePrint/3.1.00 CFNetwork/760.2.6 Darwin/15.3.0 (x86_64)" --data-binary '<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/"><s:Body><GetVersionsAndInfo xmlns = "http://tempuri.org/"></GetVersionsAndInfo></s:Body></s:Envelope>' --compressed https://tp.tcelectronic.com/TonePrintService.svc
```

vitorgalvao:
@adidalal We’re not setup in any way to be able to do that, so we might need to skip this one.

Casks/toneprint.rb
```diff
@@ -0,0 +1,11 @@
+cask 'toneprint' do
+  version '3.1'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 846c356dea96e1071d00484d6ac7ca0e40afc1a4. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #19385, opened 2016-03-01
Geneious 9 only needs legacy Java on older systems

timhowes:
Per the release notes, starting in 9.0:
  Geneious no longer requires legacy Java 6 from Apple, Oracle Java 8 is
  now bundled. A distribution that uses legacy Java 6 is also available
  for older machines, however, this will be discontinued for R10.

http://assets.geneious.com/documentation/geneious/release_notes.html#v9.0

### DIFF
--- a/Casks/geneious.rb
+++ b/Casks/geneious.rb
@@ -1,15 +1,22 @@
 cask 'geneious' do
   version '9.1.2'
-  sha256 '304afaac8251594d03379737e8263f45297da822b63b306d056057a950255ff2'
+  if MacOS.release <= '10.6'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Stick the caveat inside the top conditional instead. It’ll make it more straightforward to remove when we decide to abandon the old version.

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in ec8926e2b3f07643018db242a919ee0847a0981b. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #19398, opened 2016-03-02
Add Sassafras KeyAccess :latest

timhowes:
This is the client that connects to a Sassafras KeyServer, which manages shared licenses for commercial software.

### DIFF
--- a/Casks/keyaccess.rb
+++ b/Casks/keyaccess.rb
@@ -1,0 +1,36 @@
+cask 'keyaccess' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.sassafras.com/links/K2Client.pkg'
+  name 'Sassafras KeyAccess KeyServer Client'
+  name 'K2 Client'
+  homepage 'http://www.sassafras.com/support/'
+  license :gratis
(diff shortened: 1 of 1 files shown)


Casks/keyaccess.rb
```diff
@@ -0,0 +1,37 @@
+cask 'keyaccess' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.sassafras.com/links/K2Client.pkg'
+  name 'Sassafras KeyAccess'
```
vitorgalvao: [`name`s can be simplified](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/name.md):

```ruby
name 'Sassafras KeyAccess KeyServer Client'
name 'K2 CLient'

timhowes:
OK, thanks. I simplified the names as suggested.

vitorgalvao:
Thank you for the contribution. It was merged directly as 106289ada92705c31d588b2e102aeb65e2cb1672 to keep commit history cleaner. Your contribution is still credited to you.

PR #19401, opened 2016-03-02
Update gitkraken cask.

kud:


Casks/gitkraken.rb
```diff
@@ -1,8 +1,8 @@
 cask 'gitkraken' do
-  version '0.8.0'
-  sha256 '68b3c5fab1edc9f6703ca7e9fe54ca88db7e8a1574ecac2e416c86d8fe10d8f1'
+  version :latest
+  sha256 :no_check
 
-  url "http://release.gitkraken.com/darwin/v#{version}.zip"
+  url "https://release.gitkraken.com/darwin/installGitKraken.dmg"
   appcast 'https://release.gitkraken.com/darwin/RELEASES',
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

Casks/gitkraken.rb
```diff
@@ -1,8 +1,8 @@
 cask 'gitkraken' do
-  version '0.8.0'
-  sha256 '68b3c5fab1edc9f6703ca7e9fe54ca88db7e8a1574ecac2e416c86d8fe10d8f1'
+  version :latest
```
vitorgalvao: Don’t remove the version, we prefer to have them. See the new one in the appcast.

kud: Okay but it's not defined in the .dmg...?

It's indeed still the 0.8.0 though but it won't be forever as the url is generic.

Casks/gitkraken.rb
kud: Okay thank you. :)

What's `appcast` by the way?

Casks/gitkraken.rb
vitorgalvao: We know, but there’s still an appcast, so we can keep track of updates.

Casks/gitkraken.rb
vitorgalvao: It’s a URL the app itself checks to see if there are updates.

vitorgalvao:
Wait, just noticed. You’re assuming the app was updated. It wasn’t. Closing.

Casks/gitkraken.rb
vitorgalvao: You can find a link to the versioned dmg inside the appcast.

kud:
This is weird because when I tried, the zip wasn't available anymore. But now it does. :) Thank you anyway for your explanation. Could be helpful for another PR.

PR #19434, opened 2016-03-03
Upgrading to YourKit 2016.02-b29

robsman:


vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/19433

PR #19469, opened 2016-03-04
updated haskell-platform (7.10.3)

vitorgalvao:


adidalal:
Fixed in https://github.com/caskroom/homebrew-cask/pull/19469

PR #19475, opened 2016-03-04
.travis.yml: remove 10.9

vitorgalvao:
@jawshooah Any issue with removing this? It’s the one that most often fails, and for dumb reasons like not being able to get `bundler`.

We mention somewhere in the documentation we try to support the current OS X version and the previous one, so removing this would keep in line with that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,6 @@ matrix:
       os: osx
       osx_image: xcode7.1
       rvm: system
-    - env: OSX=10.9 HOMEBREW_RUBY=2.0.0
-      os: osx
-      osx_image: beta-xcode6.2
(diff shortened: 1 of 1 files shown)


fanquake:
ACK :+1:

adidalal:
Want to go ahead with the removal? I don't see any particular problems.

That being said, I noticed fewer 10.9 failures recently, so maybe Travis fixed issues on their end?

vitorgalvao:
> That being said, I noticed fewer 10.9 failures recently, so maybe Travis fixed issues on their end?

It seems some days are worse than others. There was a bad day recently. Let’s go ahead and merge, yes.

PR #19483, opened 2016-03-05
Add Xamarin Android Player (preview)

yurikoles:


Casks/xamarin-android-player.rb
```diff
@@ -0,0 +1,11 @@
+cask 'xamarin-android-player' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://xamarin-android-player.s3.amazonaws.com/installer/Mac/Xamarin%20Android%20Player.dmg'
+  name 'Xamarin Android Player'
+  homepage 'https://xamarin.com/android-player'
+  license :gratis
+
+  app 'Xamarin Android Player.app'
```
victorpopkov: It seems that the downloaded DMG image doesn't contain an app, but a pkg instead (`XAP-Beta-15.43-v0.6.5.pkg`). So this should be changed to `pkg` instead of an `app` and point to the pkg file.

Furthermore, if `pkg` is used, then it should also contain [uninstall](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md) stanza with correct ID:
```ruby
  pkg 'XAP-Beta-15.43-v0.6.5.pkg'

  uninstall pkgutil: 'com.xamarin.xamarinandroidplayer'
```

Casks/xamarin-android-player.rb
```diff
@@ -0,0 +1,11 @@
+cask 'xamarin-android-player' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://xamarin-android-player.s3.amazonaws.com/installer/Mac/Xamarin%20Android%20Player.dmg'
```
victorpopkov: This needs a comment `# xamarin-android-player.s3.amazonaws.com was verified as official when first introduced to the cask`. In addition, there is also a versioned URL available, which can be found [here](https://developer.xamarin.com/releases/android/android-player/) and more preferable.

This should look like this if we also reuse the version number:
```ruby
  # xamarin-android-player.s3.amazonaws.com was verified as official when first introduced to the cask
  url "http://xamarin-android-player.s3.amazonaws.com/player/mac/#{version.before_comma}/XAP-Beta-#{version.after_comma}-v#{version.before_comma}.dmg"
```

Casks/xamarin-android-player.rb
```diff
@@ -0,0 +1,11 @@
+cask 'xamarin-android-player' do
+  version :latest
```
victorpopkov: As the versioned URL is available, version can be specified with correct sha256 checksum:
```ruby
  version '0.6.5,15.43'
  sha256 'c7b220dfbe1dc876597fcb35e55834d42066fcb41747b52c760c3f0a78943da3'
```

victorpopkov:
Thank you for the contribution. It needed some fixes, so they were made in fdd7f2757da583a36508fade493b99970449d187. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #19502, opened 2016-03-06
updated transmission (2.91)

aint:
Transmission 2.90 contained malware. More information https://forum.transmissionbt.com/viewtopic.php?f=4&t=17834


cameronmalek:
Line 5 needs to be changed to reflect the new official download host (`download.transmissionbt.com`), and while we're at it, it would be best to change the homepage to HTTPS rather than HTTP.

adidalal:
Done https://github.com/caskroom/homebrew-cask/pull/19505

PR #19506, opened 2016-03-06
Use https in transmission download url

avendael:
The previous version of transmission (2.90) [may contain malware](https://forum.transmissionbt.com/viewtopic.php?f=4&t=17834) if not downloaded from https. Using https may help prevent this problem in the future.

https://news.ycombinator.com/item?id=11234589

### DIFF
--- a/Casks/transmission.rb
+++ b/Casks/transmission.rb
@@ -2,7 +2,7 @@ cask 'transmission' do
   version '2.91'
   sha256 '8d0c36540406022a47e9b9bc64190f9e852b2b6f45fe4c0d2853ac2c604b062b'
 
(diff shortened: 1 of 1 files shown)


adidalal:
If you installed and updated via HBC, you're safe. https://github.com/caskroom/homebrew-cask/issues/19504#issuecomment-192992223

Will switch over the download link, anyways.

adidalal:
Thank you for the contribution. It was merged directly as 29108e4002b1a9c73e70fb2289536cb7c965dc27 to keep commit history cleaner. Your contribution is still credited to you.

avendael:
You're welcome!

PR #19515, opened 2016-03-07
Add Cevelop C++ IDE v1.4.0

silvno:
Cevelop extends Eclipse CDT with many additional features: CUTE unit
testing with Test Driven Development support, new refactorings and
quick fixes, and much more. And it's free!

It is developed by the Institute for Software at the University of Applied
Sciences in Rapperswil, Switzerland (HSR) within the context of the
REPARA project, co-funded by the European Commission within the 
Seventh Framework Programme

### DIFF
--- a/Casks/cevelop.rb
+++ b/Casks/cevelop.rb
@@ -1,0 +1,13 @@
+cask 'cevelop' do
+  version '1.4.0'
+  sha256 '9199a6e5978e608e51de95e165f13acfdec49d265fef4bf4df5b93973f6cf2ae'
+
+  url 'https://www.cevelop.com/cevelop/downloads/cevelop-1.4.0-201603071314-macosx.cocoa.x86_64.tar.gz'
+  name 'Cevelop'
+  homepage 'https://www.cevelop.com/'
+  license :gratis
+
+  depends_on arch: :x86_64
+
+  app 'Cevelop.app'
+end


vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 2a865a1c84b6ef7bf1b810b79170524fb8a8ae20. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

silflow:
Thanks for your help! I was impressed with the workflow for adding new casks. The docs and tools were extremely helpful :)

I see that you added the archive-timestamp to the version number. For me this makes sense. Now, to add future releases we only have to adapt the version variable and the sha256 hash. The URL should automatically be correct, provided we keep our folder hierarchy. 
Also, the relationship between the version number and corresponding archive is clearly visible.
I hope my assumptions are correct :)

Cheers

vitorgalvao:
> Thanks for your help! I was impressed with the workflow for adding new casks. The docs and tools were extremely helpful :)

Thank you, great to know.

> I see that you added the archive-timestamp to the version number. For me this makes sense. Now, to add future releases we only have to adapt the version variable and the sha256 hash.

Precisely, that’s even why we visually separate that section from the rest of the cask, with a blank line. 

> The URL should automatically be correct, provided we keep our folder hierarchy.

Yep, and that’s an uncommon change.

> Also, the relationship between the version number and corresponding archive is clearly visible.
> I hope my assumptions are correct :)

They are. Everything is right on the money.

PR #19524, opened 2016-03-08
Fix SHA for iStat Menus

riyad:


victorpopkov:
The current valid checksum for v5.11 is:
```
5931ff8a024ea54f9268e66c650ae4a33b399ae25ca36586b7814d960ed77d19
```
No need to change it, at least at the moment.

riyad:
You're right ... I rechecked ... very suspicious. o.O

PR #19526, opened 2016-03-08
Amidst (new Cask)

JMoVS:
New cask to install Amidst app via homebrew cask

### DIFF
--- a/Casks/amidst.rb
+++ b/Casks/amidst.rb
@@ -1,0 +1,13 @@
+cask 'amidst' do
+  version '4.0'
+  sha256 '921300b83acf3a09f72b9adc1abeea6ce9975438b42814ae40c27e9debf13262'
+
+  url "https://github.com/toolbox4minecraft/amidst/releases/download/v#{version}/amidst-v#{version.dots_to_hyphens}.zip"
+  appcast 'https://github.com/toolbox4minecraft/amidst/releases.atom',
+          checkpoint: '84beb95511a46c011ddd2f72ab75c6cf56624d6baf59b85288ab19830620a864'
+  name 'amidst'
+  homepage 'https://github.com/toolbox4minecraft/amidst'
+  license :gpl
+
+  app 'Amidst.app'
+end


Casks/amidst.rb
```diff
@@ -0,0 +1,13 @@
+cask 'amidst' do
+  version '4.0'
+  sha256 '921300b83acf3a09f72b9adc1abeea6ce9975438b42814ae40c27e9debf13262'
+
+  url 'https://github.com/toolbox4minecraft/amidst/releases/download/v4.0/amidst-v4-0.zip'
```
victorpopkov: You can use string interpolation with [version methods](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md) here. An [appcast](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) also should be specified with appropriate `checkpoint`:
```ruby
  url "https://github.com/toolbox4minecraft/amidst/releases/download/v#{version}/amidst-v#{version.dots_to_hyphens}.zip"
  appcast 'https://github.com/toolbox4minecraft/amidst/releases.atom',
           checkpoint: '84beb95511a46c011ddd2f72ab75c6cf56624d6baf59b85288ab19830620a864'
```

Casks/amidst.rb
```diff
@@ -0,0 +1,13 @@
...
+  name 'amidst'
+  name 'Amidst'
```
victorpopkov: Only one name would be enough. [We don’t intend to become a discoverability service](https://github.com/caskroom/homebrew-cask/issues/4323#issuecomment-67257513). [It is not part of the goals of homebrew-cask](https://github.com/caskroom/homebrew-cask/issues/3817#issuecomment-39868330), so descriptions are currently not supported, especially in the `name` stanza:
```
  name 'Amidst'
```

JMoVS: I didn't know whether finding it is case-sensitive or not ;-)

Casks/amidst.rb
JMoVS: hmm ok. I can change that, but will that mean that every time a new release is published, this cask will need an update?

Also, this app has its own update mechanism (that I haven't tested yet with a cask install), are users supposed to update via homebrew-cask or via in-app?

victorpopkov:
Thank you for the contribution. It was merged directly as 9da1d90c7f2fbb4dd8e931d9a3255e13a8aee41c to keep commit history cleaner. Your contribution is still credited to you.

PR #19533, opened 2016-03-08
Update Postico to latest

miguelfrde:
Postico install was not working. That URL redirects to `https://s3-eu-west-1.amazonaws.com/eggerapps-downloads/postico-1.0.5.zip`, but I guess they will be redirecting it always to the latest version. If you prefer to keep the version, I'll update and squash the commits. Let me know!

### DIFF
--- a/Casks/postico.rb
+++ b/Casks/postico.rb
@@ -1,8 +1,9 @@
 cask 'postico' do
-  version '1.0.4'
-  sha256 '6779f03a7be3203cd2f8d3b5f3623917777998b700e500ded21162e293929dc0'
+  version '1.0.5'
+  sha256 '6b8ee69f95f36198b99482f489bd7c00a94f4b1d6efe7d08d09c41a8587e177d'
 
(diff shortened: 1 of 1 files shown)


victorpopkov:
Thank you for submitting a PR. The versioned casks are more preferred at the moment, so it's better to use the versioned URL if possible. In this case also, a [vendor comment](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) is needed:
```ruby
  # amazonaws.com/eggerapps-downloads was verified as official when first introduced to the cask
  url "https://s3-eu-west-1.amazonaws.com/eggerapps-downloads/postico-#{version}.zip"
```

miguelfrde:
@victorpopkov updated and squashed!

victorpopkov:
Thank you for the contribution. It was merged directly as daedbca943f0cf53a767f5152481103b34c1cd62 to keep commit history cleaner. Your contribution is still credited to you.

Issue #17398, opened 2016-01-21
Bug with chmod + Uninstall Script Stanza?


winkelsdorf:
Hi,

I am currently working on the Adobe Casks.

**Installation and Uninstallation works fine, but there is a message coming from ```chmod``` that a file or directory is missing**

This is the current draft for testing the recent changes to the Uninstall Stanza (including staged_path now available):
```shell
cask 'adobe-design-web-premium-cs6-de-test' do
  version :latest
  sha256 :no_check

  # This Cask only works for Adobe dmgs containing the 'deploy' folder,
  # i.e. the Adobe Collections, not the single product installers.

  # For correct download URL see links provided at
  # https://helpx.adobe.com/x-productkb/policy-pricing/cs6-product-downloads.html

  url 'http://trials2.adobe.com/AdobeProducts/DSST/CS6/osx10/DesignWebPremium_CS6_LS4.dmg',
      user_agent: :fake,
      cookies:    { 'MM_TRIALS' => '1234' }
  name 'Adobe CS6 Design and Web Premium' # name must match directory in dmg for later reference
  homepage 'https://www.adobe.com/mena_en/products/creativesuite.html'
  license :commercial

  installer script: "#{name.join}/Install.app/Contents/MacOS/Install",
            args: ['--mode=silent', "--deploymentFile=#{staged_path.join}/#{name.join}/deploy/install-de_DE.xml"],
            sudo: true

  uninstall signal: [['KILL', 'SafariNotificationAgent']],
            script: {
                      executable: "#{name.join}/Install.app/Contents/MacOS/Install",
                      args: ['--mode=silent', "--deploymentFile=#{staged_path.join}/#{name.join}/deploy/uninstall-de_DE.xml"],
                    }

  caveats 'Installation or uninstallation may fail with Exit Code 19 (conflicting processes running) if Browsers, Safari Notification Service or SIMBL Services are running, or Adobe Creative Cloud or any other Adobe Product is already installed. See the logs in /Library/Logs/Adobe/Installers if installation or uninstallation fails in order to identifify the conflicting processes.'
end
```

This is the debug output of the Uninstall Stanza:
```console
Hbc::Artifact::Uninstall
==> Un-installing artifact of class Hbc::Artifact::Installer
==> Nothing to do. The installer artifact has no uninstall phase.
==> Un-installing artifact of class Hbc::Artifact::Uninstall
[...]

PR #19535, opened 2016-03-08
Fix chmod and lastpass uninstall

onlynone:
I was getting errors uninstalling lastpass. I ran into three issues while diagnosing and fixing the cause:
- A problem with the formula for lastpass itself
- Confusing formatting of the commands being executed during uninstall
- Incorrect arguments being passed to `chmod`.

This pull request includes 3 independent commits addressing each of these issues. The `chmod` issue would seem to affect all formulas, and seems to be the cause of issue #17398.

### DIFF
--- a/Casks/lastpass.rb
+++ b/Casks/lastpass.rb
@@ -9,6 +9,8 @@ cask 'lastpass' do
 
   installer manual: 'LastPass Installer.app'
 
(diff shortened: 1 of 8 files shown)


vitorgalvao:
Looks good. Pinging @jawshooah.

vitorgalvao:
Merging now. Thank you for the detailed messages in commits.

PR #19562, opened 2016-03-09
updated password-gorilla (1.5.3.7.3)

mlrobinson:
Spoke with the app maintainer, and this change occurred to address file mode flags needing to be fixed, no other changes: https://github.com/zdia/gorilla/issues/147.


vitorgalvao:
Thank you for going the extra step to clarify this change with the developer. That’s truly appreciated.

PR #19602, opened 2016-03-10
megasync.rb: update url and license

scribblemaniac:
The homepage url was updated because https://mega.co.nz redirects to https://mega.nz. The license was changed because the source code can be found [here](https://mega.nz/#sourcecode) and is under a custom license.


vitorgalvao:
Thank you.

scribblemaniac:
No problem :smile:

PR #19672, opened 2016-03-10
fix unity-ios-support-for-editor pkg file not fount

twhaolee:


adidalal:
Thank you for the contribution. It was merged directly as b2c198f642b7b3d13ab14eae91ef4d8fa45fa841 to keep commit history cleaner. Your contribution is still credited to you.

PR #19673, opened 2016-03-10
fix unity-android-support-for-editor pkg file not fount

twhaolee:


adidalal:
Thank you for the contribution. It was merged directly as 0b9af4fb4b80380ba33f634cae6a4054a3b40b06 to keep commit history cleaner. Your contribution is still credited to you.

PR #19779, opened 2016-03-12
updated macupdate-desktop (6.1)

shonjir:
Current hash is incorrect; download fails.


vitorgalvao:
Travis disagrees, but I get the same as you. Restarting.

fanquake:
This is still failing, I've kicked it again.

PR #19802, opened 2016-03-14
add ableton-live-suite

6uclz1:
https://www.ableton.com/en/live/


vitorgalvao:
[Alternative version](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#beta-unstable-development-nightly-legacy-or-alternative-versions), so [it’s in caskroom/versions](https://github.com/caskroom/homebrew-versions/blob/master/Casks/ableton-live-suite.rb).

PR #19809, opened 2016-03-14
Update iTerm2: depends_on macos and arch

kuteken:
Version 2.1.4 needs `depends_on` setting :)

>  iTerm2 2.1.4 (OS 10.7+, Intel-only)

https://www.iterm2.com/downloads.html


vitorgalvao:
Thank you for the contribution. It was merged directly as 247283aef8c62894a531ed4c48eb53fc37d39694 to keep commit history cleaner. Your contribution is still credited to you.

PR #19825, opened 2016-03-14
Add Ceceree

ishu3101:


Casks/ceceree.rb
```diff
@@ -0,0 +1,17 @@
+cask 'ceceree' do
+  if MacOS.release >= :yosemite
+    version '1.2.1'
+    sha256 'b2038654c48fe5a6a4a4792471853ac05dead5964c0dd44ba7e786b569a920da'
+  else
+    # Unusual case: there is no fall-through.  The software will stop
```
vitorgalvao: Again, this is wrong. You’re pasting the message without actually reading or understanding it.

Casks/ceceree.rb
```diff
@@ -0,0 +1,17 @@
...
+    # working, or is dangerous to run, on the next OS X release.
+  end
+
+  # dl.devmate.com was verified as official when first introduced to the cask
```
vitorgalvao: Comment is incomplete, and doesn’t identify the app enough.

Casks/ceceree.rb
```diff
@@ -0,0 +1,17 @@
...
+  url "https://dl.devmate.com/com.appiculous.Ceceree/Ceceree.zip"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

Casks/ceceree.rb
```diff
@@ -0,0 +1,17 @@
+cask 'ceceree' do
+  if MacOS.release >= :yosemite
+    version '1.2.1'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 5d6a05bf35cb70e735265863d2b16af05499bceb. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

Casks/ceceree.rb
ishu3101: Why is this wrong?

vitorgalvao: Because there’s nothing in the app that suggests it might stop working with the next version of OS X, which is what the message implies.

These conditionals should *always* be `<=` with an `else` for the newest versions.

Here, the correct approach is not a conditional, but [a dependency check](https://github.com/caskroom/homebrew-cask/blob/5d6a05bf35cb70e735265863d2b16af05499bceb/Casks/ceceree.rb#L11).

ishu3101: Thanks. I get it now

PR #19826, opened 2016-03-14

timothybasanov:
Pkg name has only major revision name: 4.2, minor revision is not present.
As a result `version` is set to `:latest` and not `4.2` or `4.2.3`.

Without this change installation failed with pkg not found exception.

### DIFF
--- a/Casks/autopano-pro.rb
+++ b/Casks/autopano-pro.rb
@@ -7,7 +7,7 @@ cask 'autopano-pro' do
   homepage 'http://www.kolor.com/panorama-software-autopano-pro.html'
   license :commercial
 
-  pkg "Autopano Pro #{version}.pkg"
+  pkg "Autopano Pro 4.2.pkg"
 
   uninstall pkgutil: [
                        'com.kolor.pkg.AutopanoPro.*',


vitorgalvao:
Actually, the cask needs to be changed to being versioned. If the version is in the name, there’s no reason not to.

timothybasanov:
Hmm, looks like I misunderstand something. I thought that version is used for caching downloads in brew/cask. But there is a single url and single pkg for different minor versions.

Or should I just switch version to '4.2'?

vitorgalvao:
You can change the version to include the patch part as well, and then change the stanza to be `pkg "Autopano Pro #{version.major_minor}.pkg"`

timothybasanov:
Ok. I switched to version 4.2.3.
Installation verified.

This still bothers me. When I try to install it:
    ==> Downloading http://download.kolor.com/app/stable/macleopard
    Already downloaded: /Library/Caches/Homebrew/autopano-pro-4.2.3
When 4.2.4 would be released (w/o cask update) brew would install it successfully (both url and pkg would still work) and report 4.2.3 installed. This feels confusing to me.

vitorgalvao:
> When 4.2.4 would be released (w/o cask update) brew would install it successfully (both url and pkg would still work)

It won’t, because the `sha256` will differ.

PR #19831, opened 2016-03-15

timothybasanov:
Pkg name has only major revision name: 4.2, minor revision is not 

First try: #19826


Casks/autopano-pro.rb
```diff
@@ -1,5 +1,5 @@
 cask 'autopano-pro' do
-  version :latest
+  version '4.2.3'
   sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

PR #19832, opened 2016-03-15

timothybasanov:
Pkg name has only major revision name: 4.2, minor revision is not present.
#19831 #19826


vitorgalvao:
Thank you for the contribution. It was merged directly as a32a458cf6e1358c11f47d5f6989c78b81dff7cf to keep commit history cleaner. Your contribution is still credited to you.

PR #19848, opened 2016-03-15
Fix linphone (app to pkg)

victorpopkov:
Fixes https://github.com/caskroom/homebrew-cask/issues/19847.


victorpopkov:
Merged as 0070d5401ea77df396503555fc127df05fd86474.

PR #19852, opened 2016-03-15
Upgrade Macpaw Gemini to v1.5.15

cameronmalek:


adidalal:
Thank you for the contribution. It was merged directly as 3b53b0fd26aadff170c5335fc4d8e71637d90df8 to keep commit history cleaner. Your contribution is still credited to you.

adidalal:

cameronmalek:
@adidalal I had an issue where `cask-repair` would return an empty string for the SHA256 value of the app download every time. Pressing `e` after to edit and pasting it in manually would result in the same behavior over and over. Had to do it manually.

adidalal:
If things still don't work, feel free to open up a new issue from the template and we can try figuring things out! `cask-repair` is supported (and encouraged), as it generally saves time for both contributors and maintainers

PR #19863, opened 2016-03-16
chrome-devtools: app renamed to Chrome DevTools App

jawshooah:
Token remains the same per the token reference.


vitorgalvao:
Merged as d8c414f8555ed0f4135b4f50e041e48e6a0a4634.

PR #19889, opened 2016-03-17
Updated CLion(2016.1)

Dekker1:
Updates package to newest version and updates its folders to the match the newer versions (the folder structures changed a few versions ago)

### DIFF
--- a/Casks/clion.rb
+++ b/Casks/clion.rb
@@ -10,7 +10,6 @@ cask 'clion' do
   app 'CLion.app'
 
   zap delete: [
-                # "~/.CLion#{version.major_minor.no_dots}", Deprecated?
                 '~/Library/Preferences/com.jetbrains.CLion.plist',
                 "~/Library/Preferences/CLion#{version}",
                 "~/Library/Application Support/CLion#{version}",
--- a/Casks/clion.rb
+++ b/Casks/clion.rb
@@ -1,8 +1,8 @@
 cask 'clion' do
-  version '1.2.4'
-  sha256 'e476eb093fc4624090f941d5767dc20edbcd12aca98bc388d8482606691452b9'
+  version '2016.1'
+  sha256 'c791636f23c273dd07bbcab4f2495a3532501ec2ec3975c7c80cc3b915686be4'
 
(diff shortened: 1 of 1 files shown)


Casks/clion.rb
```diff
@@ -10,11 +10,11 @@
   app 'CLion.app'
 
   zap delete: [
-                "~/.CLion#{version.major_minor.no_dots}",
+                # "~/.CLion#{version.major_minor.no_dots}", Deprecated?
```
adidalal: Anyone who uses the app can confirm?
If it's deprecated, we can just remove that line entirely.

jjdekker: So when using the app this folder is not automatically created as it was in previous editions. However, I can not find any change log stating that this folder is not used any more. In the EAP, this folder was still in use, thus I decided to keep it as a comment awaiting further evaluation.

peterjosling: From looking in the directories for IntelliJ and WebStorm, I don't think it's supposed to be used on OS X at all. The help document [here](https://www.jetbrains.com/help/idea/15.0/project-and-ide-settings.html]) says the hidden directory is for Linux, and OS X makes no mention. The only thing in the directory for either product is a 'restart' directory, which I guess is a plugin of some kind? I have the directory created for the IntelliJ EAP, but not for WebStorm, so my guess is it could be created by the app at any point.

tan9: https://intellij-support.jetbrains.com/hc/en-us/articles/206544519-Directories-used-by-the-IDE-to-store-settings-caches-plugins-and-logs didn't mention to `~/.<PRODUCT><VERSION>` on OS X.

jjdekker: As the official documentation doesn't report this folder as existing and it is not automatically created any more I would suggest we will remove it completely.
I'll send in a commit changing this.

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 81688b19ed9ac6f1b1e6be9721f317b5dc6fc5ec. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #19895, opened 2016-03-17
Viscosity 1.6.2

mistydemeo:


vitorgalvao:
Done in https://github.com/caskroom/homebrew-cask/pull/19902 (I open my notifications from top to bottom, so I end up always seeing the latest ones first).

For future reference, you should still look at that PR, as it did a few things missing from this one. The first is it changed the URL to a versioned one. We prefer those, if they are available (as you do in homebrew<sup>1</sup>). That one could be easily missed, though.

The important missing part was the `:checkpoint` value in `appcast`, as it is a requirement to update it (currently audit doesn’t check for it). You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

Either way, thank you for the PR, and welcome!

---

<sup>1</sup> Actually, have you left homebrew and are just working on tigerbrew, now?

mistydemeo:
> 1 Actually, have you left homebrew and are just working on tigerbrew, now?

Not correct - I'm still a Homebrew maintainer.

Thanks for the pointers; I'll make sure I do these in the future. If you have the chance, it'd be helpful to get these documented in the CONTRIBUTING.md!

adidalal:
@mistydemeo - should all be documented in CONTRIBUTING AFAIK

PR #19919, opened 2016-03-19
updated qlcolorcode (2.0.6b4)

saagarjha:


adidalal:
We prefer to use stable releases, in this case, `2.0.5`

saagarjha:
2.0.5 should be fine, then?

adidalal:
2.0.5 has no binary release, looks like?
If there's no release within a few days after last commit, just open up a new issue, looks like the autor is pretty responsive.

https://github.com/anthonygelibert/QLColorCode/issues/2, for example

PR #19926, opened 2016-03-19
Fix Splice installation instructions

joshka:
Splice is distributed as both a version specific zip file and an installer. We
use the former, but the instructions were written for the latter.


vitorgalvao:
Thank you for the contribution. It was merged directly as 29ccbe5dfcd1ebfefa1df0d21fd8f30b36c1fdd9 to keep commit history cleaner. Your contribution is still credited to you.

PR #19931, opened 2016-03-19
Add Max for Live name to cycling74-max

joshka:


vitorgalvao:
You have an extraneous commit for Splice.

Casks/cycling74-max.rb
```diff
@@ -5,6 +5,7 @@
   # akiaj5esl75o5wbdcv2a-maxmspjitter.s3.amazonaws.com was verified as official when first introduced to the cask
   url "https://akiaj5esl75o5wbdcv2a-maxmspjitter.s3.amazonaws.com/Max#{version.no_dots}.dmg"
   name 'Cycling ‘74 Max'
```
vitorgalvao: `Cycling ‘74 Max for Live`. See [name](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/name.md).

joshka: This is kind of a weird one, that I'm not sure makes sense to me. Cycling '74 are the creators of Max, but the Max for Live product is only available to purchase from Ableton for owners of the Standard edition of of Ableton Live. See https://cycling74.com/products/maxforlive/ which points you over to https://www.ableton.com/en/live/max-for-live/. It's free with Ableton Suite, but still needs to be installed using this cask. For the full product, the website explicitly calls the product name out as Cycling '74 Max, but for the Max for Live, it leaves the Cycling '74 part off.

Happy to fix this if you'd want it still to have the Cycling '74 prior to Max for Live, but really the only reason for putting this second name here is to catch those searching for the product to be used in Ableton. It also prevents people needlessly creating a new cask (e.g. https://github.com/caskroom/homebrew-cask/pull/19803). I almost did this until I worked out that the download links were the same for Max and Max for Live.

adidalal: Might be a good use case for the multiple name stanzas, if at all, imo

joshka: @adidalal that's the intent of this change - it adds an extra name stanza. See the line below this one.

More succinctly, this cask installs two products: "Cycling '74 Max" and "Max for Live". It's unlikely someone would search for "Cycling 74 Max for Live", as the company name is referenced just about nowhere when talking about Max for Live.

adidalal: whoops, didn't see the diff before commenting.
`name` is just a pattern matching for now, so just putting all the keywords in one *technically* works - I was under the impression we decided against that.

"Ableton Max for Live" sounds like the best option, actually. Will change and push shortly.



So, your approach

adidalal:
@joshka Looks like `prfixmaster` fails for me locally, mind changing the name to `Ableton Max for Live` and squashing/force pushing so I can merge?

joshka:
@adidalal will do and also will drop the splice commit (which should be already in master)

joshka:
@adidalal This is ready for merge based on what we've spoken about

PR #19947, opened 2016-03-20
Add jAlbum.app v13

gfilicetti:
Initial creation of jAlbum.app
First attempt at jAlbum cask
Passed install / uninstall test
Passed audit
Passed style check

### DIFF
--- a/Casks/jalbum.rb
+++ b/Casks/jalbum.rb
@@ -1,0 +1,11 @@
+cask 'jalbum' do
+  version '13'
+  sha256 '1b6466ad6b2272addeaa23b70ac3714b4df3357d0c168ea58ff6060c8cf33e4c'
+
+  url "http://download.jalbum.net/download/#{version}/MacOSX/jAlbum.dmg"
+  name 'jAlbum'
+  homepage 'https://jalbum.net'
+  license :freemium
+
+  app 'jAlbum.app'
+end


Casks/jalbum.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  # amazonaws.com/gateblu was verified as official when first introduced to the cask
+  url "http://download.jalbum.net/download/#{version}/MacOSX/jAlbum.dmg"
+  name 'jAlbum'
+  homepage 'https://jalbum.net'
+  license :freemium
+
+  app 'jAlbum.app'
+
+  uninstall delete:  '/Applications/jAlbum.app'
```
vitorgalvao: Remove this line. Removal is done automatically, and you can’t assume people will install to this directory.

Casks/jalbum.rb
```diff
@@ -0,0 +1,14 @@
+cask 'jalbum' do
+  version '13'
+  sha256 :no_check
+
+  # amazonaws.com/gateblu was verified as official when first introduced to the cask
```
vitorgalvao: This comment is wrong and needs to be removed.

Casks/jalbum.rb
```diff
@@ -0,0 +1,14 @@
+cask 'jalbum' do
+  version '13'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

gfilicetti: Unfortunately, even though the download URL is versioned, it is versioned via the major release, ie: 13. The actual file you get can be any minor version of 13 as the app adheres to continuous release practices and therefore would have a different checksum each time.

Casks/jalbum.rb
gfilicetti: Ugh... sorry about that.

Casks/jalbum.rb
vitorgalvao: That may or may not be a problem, depending on how often it‘s updated. It’s better to err on the other side. Set the `sha256` and if it becomes unmanageable in the future, we’ll remove it.

vitorgalvao:

gfilicetti:
I believe the squash is ok?

vitorgalvao:
Nope, the goal is to end up with one commit. There are currently three. Don’t worry if it doesn’t go well at first, it’s an obtuse process that trips up most people at first.

gfilicetti:
Alright vitor... that should do it. I was messing up on the rebase, leaving > 1 pick line. All set now I hope?!

vitorgalvao:
Yep. Thank you.

gfilicetti:
Hey Vitor... one last thing re: a contribution on github... should I tie off or delete my fork now that my (current) submission is done? Or do they just build up and up and up?

vitorgalvao:
Whatever you feel like. They do tend to pile up (at least for me), but if you’re planning on contributing more, keeping the for will make it faster (no new forks will be created for this repo).

Deleting is also not a big issue, though.

My advice: don’t delete for now, and once every few months look at your repos and get rid of the cruft.

gfilicetti:
Awesome, thanks for the help.

As I keep reinstalling my applications on this mac (just wiped and
installed el capitan yesterday).

What do you think? Separate pull requests for each cask?

G

vitorgalvao:
Yes, we require one PR per cask.

PR #19986, opened 2016-03-22
terraform 0.6.14

phinze:


vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 60b54c3addd7e0ab1d40831b3b90cb3f719b2980. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

vitorgalvao:
We should add appcasts to the other `hashicorp` apps.

phinze:
Thanks for the fix @vitorgalvao! I actually wasn't aware that our new hc-releases platform supported atom feeds, so good find! :grinning:

radeksimko:
@phinze I think the atom channel comes from Github releases, which aren't directly connected to the artefact in `releases.hashicorp.com`. Such atom channel is fine for detecting whether there's any new release, but not very useful for any verification or parsing out download URLs, or at least until uploading the actual artefacts (i.e. ZIP files with binaries) to Github becomes part of the release process.

vitorgalvao:
@radeksimko We don’t parse appcasts, there’s simply too much variance for it to be reliable.

Instead, appcasts tell us when there are updates, like you mentioned, *and the version it was updated to*. So as long as the download URLs on `releases.hashicorp.com` stay consistent, with only the version number changing, updating it a `cask-repair --cask-version {{new_version}} --blind-submit {{cask_name}}` away.

vitorgalvao:
@samshadwell [did it](https://github.com/caskroom/homebrew-cask/pull/19996).

PR #19992, opened 2016-03-22
Added Openoffice in pt-br lang

josevieirame:
Added OpenOffice in pt-br locale lang


josevieirame:
Forget to change the sha256

PR #20010, opened 2016-03-23
Fix pl2303 v1.6.1_20160309

kTitan:
Missed some changes in the pkg structure of the source package.
- Package Name is now without „v“ as prefix of the version
- Kext Location was changed


vitorgalvao:
Thank you for the contribution. It was merged directly as cd85ff2ce245ef1f5d97a5b038393397d08bf5aa to keep commit history cleaner. Your contribution is still credited to you.

PR #20016, opened 2016-03-23
Update USAGE.md

jaredkoontz:
Make it so a user can simply copy and paste the command. If the ticks are in there the command will not run


vitorgalvao:
Thank you.

PR #20027, opened 2016-03-24
Update ChefDK to 0.12.0

daften:
Update ChefDK to the latest version.


victorpopkov:
Done in https://github.com/caskroom/homebrew-cask/pull/20025. But still thank you for your contribution.

PR #20032, opened 2016-03-24
Remove invalid argument separator passed to chmod

wirelyre:
Apple chmod stops parsing options after reading `-N` (which removes ACL) and expects a file name immediately following. The command `chmod -N -- file` attempts to clear the ACL on the file `--`, rather
than `file`.

A recent commit changed all chmod invocations to use the argument separator for consistency. However, there is no way to use the argument separator in a chmod invocation involving `-N`. See line 163 of [the chmod source](http://www.opensource.apple.com/source/file_cmds/file_cmds-251/chmod/chmod.c).


vitorgalvao:
Thank you.

PR #20066, opened 2016-03-25
Initialize parallels desktop and fix uninstallation of parallels-desktop

toonetown:
The inittool script will initialize the installation instead of waiting until the first run.  Also, the owner of the symlink needs to be root:wheel in order for prlctl commands to work.  Prior to uninstallation, the ownership also needs to be modified.

### DIFF
--- a/Casks/parallels-desktop.rb
+++ b/Casks/parallels-desktop.rb
@@ -12,6 +12,19 @@ cask 'parallels-desktop' do
   postflight do
     # Set the file to visible, since it was hidden in the dmg
     system '/usr/bin/SetFile', '-a', 'v', staged_path.join('Parallels Desktop.app')
+    
+    # Run the initialization script
+    system '/usr/bin/sudo', '-E', '--',
+            staged_path.join('Parallels Desktop.app/Contents/MacOS/inittool'), 
+            'init', '-b', staged_path.join('Parallels Desktop.app')
+    
+    # Set the correct permissions on the link - remove after #13201 is solved 
+    system '/usr/bin/sudo', '-E', '--',
+            'chown', '-h', 'root:wheel', Hbc.appdir.join('Parallels Desktop.app')
+  end
+
+  uninstall_preflight do
+    set_ownership staged_path.join('Parallels Desktop.app')
   end
 
   uninstall delete: [


Casks/parallels-desktop.rb
```diff
@@ -12,6 +12,19 @@
...
     # Set the file to visible, since it was hidden in the dmg
     system '/usr/bin/SetFile', '-a', 'v', staged_path.join('Parallels Desktop.app')
+    
+    # Run the initialization script
+    system '/usr/bin/sudo', '-E', '--',
+            staged_path.join('Parallels Desktop.app/Contents/MacOS/inittool'), 
+            'init', '-b', staged_path.join('Parallels Desktop.app')
+    
+    # Set the correct permissions on the link   
+    system '/usr/bin/sudo', '-E', '--',
```
vitorgalvao: Why not use `set_ownership` here as well? [Docs](https://github.com/caskroom/homebrew-cask/blob/c86b8bbe776b496977a061f7609ea60aa9d0bff3/doc/cask_language_reference/stanzas/flight.md#flight-mini-dsl):

> `set_ownership(paths)` defaults user ownership to the current user and group ownership to `staff`. These can be changed by passing in extra options: `set_ownership(paths, user: 'user', group: 'group')`.

Casks/parallels-desktop.rb
```diff
@@ -12,6 +12,19 @@
   postflight do
     # Set the file to visible, since it was hidden in the dmg
     system '/usr/bin/SetFile', '-a', 'v', staged_path.join('Parallels Desktop.app')
+    
+    # Run the initialization script
+    system '/usr/bin/sudo', '-E', '--',
```
vitorgalvao: Can’t this be done with `installer script:`?

Casks/parallels-desktop.rb
toonetown: I tried that, but set_ownership doesn't work on symbolic links...to change the owner of a symlink, you need to specify "-h" to the chown command.

vitorgalvao: Does the ownership of the link really need to be changed? Can’t it be the ownership of the app itself?

toonetown: Yeah - it does have to be changed. If it is not, the link is created with owner of the current user. However, the "headless" service that runs does some kind of check to make sure the file is owned by root:wheel. Seems to be a limitation (or problem) of parallels desktop, and the display service, but it does seem to be required. :(

Casks/parallels-desktop.rb
toonetown: I would assume so, but according to the doc, `installer script:` requires an uninstall stanza, which I don't think is needed for this package.

vitorgalvao: Oh, I misread it as “installation script”.

Casks/parallels-desktop.rb
```diff
@@ -12,6 +12,19 @@
...
+            staged_path.join('Parallels Desktop.app/Contents/MacOS/inittool'), 
+            'init', '-b', staged_path.join('Parallels Desktop.app')
+    
+    # Set the correct permissions on the link
```
vitorgalvao: Please add a note to the comment (`remove after #13201 is solved`).

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 9cfce2b585858e672536062be7be16f38607716f. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #20078, opened 2016-03-25
Switch to Neo11's ChitChat and update to v2.2

DouweM:
@stonesam92's original ChitChat is not actively being maintained: https://github.com/stonesam92/ChitChat/graphs/contributors

@Neo11's fork is: https://github.com/Neo11/ChitChat/graphs/contributors

It has some added functionality like:
- Quick reply from the OS X notification (the original version has that as well as of 2 days ago, the first activity since October, but the fork has had it for a while)
- User avatar in notification
- Automatic conversion of textual smileys like `:)` to emoji 😊
- A lot of bug fixes

I have switched to the fork, and suspect most ChitChat users would choose to do that if they knew the fork existed.

I'm curious what the Caskroom policy is about "updating" to a fork of a project when the original project is no longer actively maintained.

### DIFF
--- a/Casks/chitchat.rb
+++ b/Casks/chitchat.rb
@@ -1,12 +1,12 @@
 cask 'chitchat' do
-  version '1.5'
-  sha256 'dbcfe1698d6d4cd9239480f757e821f7a784f7ac89a0bd26b32a0b9014999db0'
+  version '2.2'
+  sha256 '62e3e6a17808b7c5b77b127acf98a04cbabe73770f2aa1be2b984d6f20a2e93f'
 
-  url "https://github.com/stonesam92/ChitChat/releases/download/v#{version}/ChitChat.zip"
-  appcast 'https://github.com/stonesam92/ChitChat/releases.atom',
-          checkpoint: '5aab4cfa3ff23c6b9d0706df5deb9cf0c251a272555c57e3b35cdaeb2ba8796e'
+  url "https://github.com/Neo11/ChitChat/releases/download/v#{version}/ChitChat.zip"
+  appcast 'https://github.com/Neo11/ChitChat/releases.atom',
+          checkpoint: 'cd10198c6fc8df02488556674ecc3f7e76ffe17128c0b5919259b6d2e2029c21'
   name 'ChitChat'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> I have switched to the fork, and suspect most ChitChat users would choose to do that if they knew the fork existed.

I understand your point, but that’s a big leap to make. We can’t just assume that and surprise users. Also, the latest commit on the main one was two days ago. It’s clearly not dead yet. By the number of stars, it’s clearly more popular (even if subjectively worse).

> I'm curious what the Caskroom policy is about "updating" to a fork of a project when the original project is no longer actively maintained.

[It’s well documented](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#forks-and-apps-with-conflicting-names):

> Forks should have the vendor’s name as a prefix on the Cask’s file name and token. For unrelated Apps that share a name, the most popular one (usually the one already present) stays unprefixed. Since this can be subjective, if you disagree with a decision open an issue and make your case to the maintainers.

So the current `chitchat` stays as is, and you add a new `neo11-chichat` cask.

DouweM:
@vitorgalvao Completely understood, I wasn't expecting this to get merged at all, just wanted to start the conversation. Good that you have a clear policy on this, even if I missed it initially.

Anyway, I'll hold out hope that maintenance of the original chitchat will indeed be picked up again, and I don't see a need to create a `neo11-chitchat` at this point, since people who know about that fork would likely be on the project page anyway, and at that point downloading the release directly is easier than going through cask and hoping the definition has been updated.

vitorgalvao:
> I don't see a need to create a `neo11-chitchat` at this point, since people who know about that fork would likely be on the project page anyway, and at that point downloading the release directly is easier than going through cask and hoping the definition has been updated.

That can be said about any app. The point of homebrew-cask is precisely that you don’t have to do that. This is perfectly mergeable with a name change.

DouweM:
@vitorgalvao Fair enough. Thanks for creating the PR!

vitorgalvao:
Thank you fro telling us about it.

PR #20108, opened 2016-03-27
updated mouse locater to correct sha256 and version number

pacav69:
updated mouse locater to correct sha256 and version number


adidalal:
https://github.com/caskroom/homebrew-cask/pull/20109

Your PR had errors - try using `cask-repair` (check CONTRIBUTING)

PR #20113, opened 2016-03-28
Fix linkliar

stqism:
The existing URL 404'd because it used an schema Github no longer supports.

As a bonus I added a sha256 and appcast support for fun.


victorpopkov:
Thank you for the contribution. It needed some fixes, so they were made in db2bc4dcc69348bda712a820df8b3fc5dff15f80. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #20120, opened 2016-03-28
Update zap delete paths of WebStorm

ghost:
Since WebStorm 2016.1, the plist file created under `~/Library/Preferences/` is renamed as `jetbrains.webstorm.cdacf2ec.plist` instead of the old `com.jetbrains.webstorm.plist`, and it looks like that the random string appears in the middle of the name is version specific.

I also add the command line launcher executable, it's usually created when you confirm the config dialog appeared after the registration.

### DIFF
--- a/Casks/webstorm.rb
+++ b/Casks/webstorm.rb
@@ -9,9 +9,11 @@ cask 'webstorm' do
 
   app 'WebStorm.app'
 
+  uninstall delete: '/usr/local/bin/wstorm'
+
   zap delete: [
                 "~/.WebStorm#{version.major_minor}",
-                '~/Library/Preferences/com.jetbrains.webstorm.plist',
+                # TODO: expand/glob for '~/Library/Preferences/jetbrains.webstorm.*.plist',
                 "~/Library/Preferences/WebStorm#{version.major_minor}",
                 "~/Library/Application Support/WebStorm#{version.major_minor}",
                 "~/Library/Caches/WebStorm#{version.major_minor}",


Casks/webstorm.rb
```diff
@@ -10,8 +10,9 @@
   app 'WebStorm.app'
 
   zap delete: [
+                '/usr/local/bin/wstorm',
```
vitorgalvao: Remove this one. This isn’t a preference.

tty1:
I know it's not a preference file, but shouldn't `zap delete` perform a clean uninstall that remove every file created by the cask?

vitorgalvao:
No, zap is to delete user-created files. Since that isn’t one, it should be in `uninstall`, not `zap`.

tty1:
Thanks for your explanation, so would it be valid if I move it to the `uninstall` field instead?

vitorgalvao:
> Thanks for your explanation, so would it be valid if I move it to the `uninstall` field instead?

Yes. We should likely also do it for all other JetBrains casks.

tty1:
If I want to change other JetBrains cask files, should I open a single pull request or open standalone ones for each cask instead?

vitorgalvao:
We prefer one PR per cask. However, if you really want to make a single PR, at least make each cask change a different commit.

tty1:
Rebased, and I will open another pull request to fix related JetBrains casks later on today.

vitorgalvao:
Thank you for the contribution. It was merged directly as 2a728edfbca937f71602e4b889d36d6a03f28f96 to keep commit history cleaner. Your contribution is still credited to you.

PR #20141, opened 2016-03-29
updated to pass tests

pacav69:
passing offline tests


vitorgalvao:
Seeing as `url` doesn’t point to a specific version, this cask needs to use `version :latest` and `sha256 :no_check`. Otherwise, it’ll break every time upstream updates de binary.

pacav69:
Well the app hasn't been updated since 2006 so i think it would be safe for now.
When i installed it as it stood it failed until i updated it.

pacav69:
added 
  caveats do
    discontinued
  end

PR #20149, opened 2016-03-29
Fix IntelliJ Idea 2016.1 and related casks

ghost:


vitorgalvao:
Thank you!

There are also a few other casks from JetBrains in [caskroom/versions](https://github.com/caskroom/homebrew-versions), if you feel like tackling those.

PR #20170, opened 2016-03-30
Casks/docear.rb: Update to version 1.2.0.0, which only provides zip

krisgry:
Casks/docear.rb: Update to version 1.2.0.0, which only provides zip (not dmg)

### DIFF
--- a/Casks/docear.rb
+++ b/Casks/docear.rb
@@ -1,11 +1,11 @@
 cask 'docear' do
-  version :latest
+  version '1.2.0.0_stable_build291'
   sha256 :no_check
 
-  url 'http://docear.org/download/docear.dmg'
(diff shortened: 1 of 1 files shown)


Casks/docear.rb
```diff
@@ -1,11 +1,11 @@
 cask 'docear' do
-  version :latest
+  version '1.2.0.0_stable_build291'
   sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 71cad8236be8fbab17ffc378f4aecce7ecd180e1. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #20247, opened 2016-04-02
Add new Cask: hypercli

carmark:


adidalal:
Where's the download link on the homepage?

I only see a "private beta"

carmark:
@adidalal Thanks for your review.

The download link is in the [page](https://doc.hyper.sh/GettingStarted/install.html).

vitorgalvao:
@carmark Are you in charge of that page? Because `brew upgrade` is absolutely unnecessary, there.

carmark:
@vitorgalvao 

Thanks for your point this, I have updated the source of page, but it may talk a while to show the modification.

adidalal:
For uninstall, shouldn't `opt/bin/static` and all subdirectories also get removed? Please add that / change uninstall accordingly

vitorgalvao:
Also, for s complete `uninstall`, I’m guessing we could at least get a `pkgutil:`. We have [scripts](https://github.com/caskroom/homebrew-cask/tree/master/developer/bin) and [documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md) to make as complete as possible.

adidalal:
Package ID: `sh.hyper.hyper`

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 5dbd1136ebaa4ab9a5c2902f3fae0bbf9760771c. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #20262, opened 2016-04-03
Add file hash and version for Sonos

Fuzion24:


Casks/sonos.rb
```diff
@@ -1,7 +1,6 @@
 cask 'sonos' do
-  version :latest
-  sha256 :no_check
-
```
vitorgalvao: File structure must be kept.

vitorgalvao:
Seeing as `url` doesn’t point to a specific version, this cask must use `version :latest` and `sha256 :no_check`, or it’ll constantly break with upstream updates.

Fuzion24:
I think breaking on upstream updates is better behavior than the alternative: http://researchcenter.paloaltonetworks.com/2016/03/new-os-x-ransomware-keranger-infected-transmission-bittorrent-client-installer/

vitorgalvao:
We know from experience how constantly breaking casks are. We prefer versioned casks when possible, so if you want this to be versioned, you’re asking in the wrong place. Ask the developers of the app to provide versioned downloads.

Also, that case might not have been prevented anyway. Since the `sha256` is calculated by users, if an app is hosed on the server, homebrew-cask won’t protect you if the cask was updated after it.

PR #20277, opened 2016-04-04
gitkraken 1.0.0

sjackman:
#### Editing an existing cask

### DIFF
--- a/Casks/gitkraken.rb
+++ b/Casks/gitkraken.rb
@@ -1,10 +1,10 @@
 cask 'gitkraken' do
-  version '0.9.2'
-  sha256 '810654fcf0813552f4cd7ee6032440f1e7072d3f2bb2442c3d02e484fed2d345'
+  version '1.0.0'
+  sha256 '3f93b84ea57d58396d759f9882be0275fdf370baba457e463edb5c40b060caad'
 
   url "http://release.gitkraken.com/darwin/v#{version}.zip"
   appcast 'https://release.gitkraken.com/darwin/RELEASES',
-          checkpoint: '6490edaa515b32eebe5b0811b185ecea5ccdaa9fee74dea8fe7dcc047a201c24'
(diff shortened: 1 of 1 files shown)


Casks/gitkraken.rb
```diff
@@ -1,6 +1,6 @@
 cask 'gitkraken' do
-  version '0.9.2'
-  sha256 '810654fcf0813552f4cd7ee6032440f1e7072d3f2bb2442c3d02e484fed2d345'
+  version '1.0.0'
+  sha256 '3f93b84ea57d58396d759f9882be0275fdf370baba457e463edb5c40b060caad'
 
   url "http://release.gitkraken.com/darwin/v#{version}.zip"
   appcast 'https://release.gitkraken.com/darwin/RELEASES',
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

sjackman: Done. It would help for this check to be added to `brew cask audit`. Could `cask-repair` be added to `homebrew-cask`?

vitorgalvao:
> It would help for this check to be added to `brew cask audit`

It would. We’re [in dire need of a PR](https://github.com/caskroom/homebrew-cask/issues/16529).

> Could `cask-repair` be added to `homebrew-cask`?

In what sense?

sjackman:
Thanks for merging. Sorry that I'm not able to contribute the time to make the PR. =(

>>Could cask-repair be added to homebrew-cask?

>In what sense?

Add a `brew cask repair` command.

vitorgalvao:
I thought about that before, but `cask-repair` is made in bash, and there’s a lot that needs to be done in homebrew-cask itself. It’s breaking everywhere, and there’s plenty to fix. I’d rather we fix those problems first.

PR #20279, opened 2016-04-05
updated cubicsdr (0.1.27)

josh64x2:


adidalal:
Is there a reason you switched to an alpha release? We prefer stable releases here.

josh64x2:
Hi,

I understand you prefer stable releases. The previous version 0.1.4 was a beta, is over 8 months old, and wasn't particularly stable. Although 0.1.27 is tagged alpha, it is a very large improvement over 0.1.4 so I thought it was appropriate to provide a newer cask.

Thanks!

adidalal:
Took a quick look at the releases page, seems okay to use "alpha" releases as one tagged as stable is pretty far back.

josh64x2:
Ok thanks. I'll keep it in mind next time for other casks. Cheers!

PR #20284, opened 2016-04-05
updated adobe-air (21.0)

matteocng:


adidalal:
@fanquake Use the "Squash and Merge" setting from the web UI - prevents merge commits

PR #20372, opened 2016-04-07
Begin PR for adding Plug for OSX 10.9.5

ninjasort:
#### Editing an existing cask

### DIFF
--- a/Casks/plug.rb
+++ b/Casks/plug.rb
@@ -1,15 +1,16 @@
 cask 'plug' do
   if MacOS.release <= :lion
-    version '?'
-    sha256 '?'
+    version '0.10.5'
+    sha256 'd8eed07bd1f84d6f1daa7a0699617f2e21c01df2e68924945bcb4889d1251d01'
+    url 'http://plugformac.com/files/Plug-0.10.5.dmg'
   else
     version '2.0.9'
     sha256 'bead86a9880ae8eb63296b48d6aef30d80d6057a11ca8491dc85a27b296f5e36'
     appcast 'https://www.plugformac.com/updates/plug2/sparklecast.xml',
 	  checkpoint: '7092bd0eb0a14a478018d36a44addfa4a8dfc871e214f79d66b6a5e1ce216815'
+    url 'https://www.plugformac.com/updates/plug2/Plug-latest.dmg'
   end
 
-  url 'https://www.plugformac.com/updates/plug2/Plug-latest.dmg'
   name 'Plug'
   homepage 'https://www.plugformac.com/'
   license :gratis
--- a/Casks/plug.rb
+++ b/Casks/plug.rb
@@ -1,10 +1,15 @@
 cask 'plug' do
-  version '2.0.9'
-  sha256 'bead86a9880ae8eb63296b48d6aef30d80d6057a11ca8491dc85a27b296f5e36'
+  if MacOS.release <= :lion
+    version '?'
+    sha256 '?'
+  else
+    version '2.0.9'
+    sha256 'bead86a9880ae8eb63296b48d6aef30d80d6057a11ca8491dc85a27b296f5e36'
+    appcast 'https://www.plugformac.com/updates/plug2/sparklecast.xml',
+	  checkpoint: '7092bd0eb0a14a478018d36a44addfa4a8dfc871e214f79d66b6a5e1ce216815'
+  end
 
   url 'https://www.plugformac.com/updates/plug2/Plug-latest.dmg'
-  appcast 'https://www.plugformac.com/updates/plug2/sparklecast.xml',
-          checkpoint: '7092bd0eb0a14a478018d36a44addfa4a8dfc871e214f79d66b6a5e1ce216815'
   name 'Plug'
   homepage 'https://www.plugformac.com/'
   license :gratis
--- a/Casks/plug.rb
+++ b/Casks/plug.rb
@@ -1,8 +1,8 @@
 cask 'plug' do
-  if MacOS.release <= :lion
+  if MacOS.release <= :mavericks
     version '0.10.5'
     sha256 'd8eed07bd1f84d6f1daa7a0699617f2e21c01df2e68924945bcb4889d1251d01'
-    url 'http://plugformac.com/files/Plug-#{version}.dmg'
+    url 'http://plugformac.com/files/Plug-0.10.5.dmg'
   else
     version :latest
     sha256 :no_check
--- a/Casks/plug.rb
+++ b/Casks/plug.rb
@@ -2,12 +2,12 @@ cask 'plug' do
   if MacOS.release <= :lion
     version '0.10.5'
     sha256 'd8eed07bd1f84d6f1daa7a0699617f2e21c01df2e68924945bcb4889d1251d01'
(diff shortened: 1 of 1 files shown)


adidalal:
What are you looking to do here? It's not immediately apparent.

vitorgalvao:
@adidalal Refs https://github.com/caskroom/homebrew-cask/issues/20364.

@cameronroe You’ve erase the template that is very much necessary. It tells you what command you need to run to make sure this is fine (right now it has multiple errors).

cameronroe:
Have the `?` replaced with the correct version of Plug. I think the author needs to add that in.

Casks/plug.rb
```diff
@@ -1,10 +1,15 @@
 cask 'plug' do
-  version '2.0.9'
-  sha256 'bead86a9880ae8eb63296b48d6aef30d80d6057a11ca8491dc85a27b296f5e36'
+  if MacOS.release <= :lion
+    version '?'
```
vitorgalvao: You need to actually fill these with something, i.e. whatever version last worked on this OS X version.

cameronroe: Yep, that was from the other cask you posted. Need to swap in the old release for Plug.

vitorgalvao: The other cask has a version there. If you want to include this, you need to figure out what is the version for this cask.

cameronroe: I was hoping the Plug guys could jump in on that, but I'll look for it.

vitorgalvao:
I’ve edited your top post so you can check it.

Casks/plug.rb
vitorgalvao: There are no “Plug guys” for this. No cask has a specific maintainer. Either you add it, or likely no one will.

vitorgalvao:
> Have the `?` replaced with the correct version of Plug. I think the author needs to add that in.

Again, there are no authors. You need to find it and replace it yourself.

cameronroe:
Okay, I will get the correct version.

cameronroe:
I've updated the version.

Casks/plug.rb
```diff
@@ -1,10 +1,16 @@
 cask 'plug' do
-  version '2.0.9'
-  sha256 'bead86a9880ae8eb63296b48d6aef30d80d6057a11ca8491dc85a27b296f5e36'
+  if MacOS.release <= :lion
+    version '0.10.5'
+    sha256 'd8eed07bd1f84d6f1daa7a0699617f2e21c01df2e68924945bcb4889d1251d01'
+    url 'http://plugformac.com/files/Plug-0.10.5.dmg'
+  else
+    version '2.0.9'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this should be `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

Casks/plug.rb
```diff
@@ -1,10 +1,16 @@
 cask 'plug' do
-  version '2.0.9'
-  sha256 'bead86a9880ae8eb63296b48d6aef30d80d6057a11ca8491dc85a27b296f5e36'
+  if MacOS.release <= :lion
+    version '0.10.5'
+    sha256 'd8eed07bd1f84d6f1daa7a0699617f2e21c01df2e68924945bcb4889d1251d01'
+    url 'http://plugformac.com/files/Plug-0.10.5.dmg'
```
vitorgalvao: `url "http://plugformac.com/files/Plug-#{version}.dmg"`.

Also, that far back (`0.10.5`)? I have my doubts it’s worth adding this, in this case, seems like you’re the only one who’s going to use it.

vitorgalvao:
Cask is still a mess. You haven’t looked at the check list. Please only report back when those are done (or if you have questions).

Casks/plug.rb
cameronroe: This is for Mavericks users.

Casks/plug.rb
```diff
@@ -1,10 +1,16 @@
 cask 'plug' do
-  version '2.0.9'
-  sha256 'bead86a9880ae8eb63296b48d6aef30d80d6057a11ca8491dc85a27b296f5e36'
+  if MacOS.release <= :lion
```
vitorgalvao: Wait, just noticed, didn’t you want this to work on Mavericks? This will only activate on Lion and earlier.

adidalal:
There's always the option of putting this in your own tap, especially if the value-add is dubious for the general HBC repository.

Casks/plug.rb
vitorgalvao: > This is for Mavericks users.

Mavericks users who use Plug and homebrew-cask. That may very well be an audience of one.

Casks/plug.rb
cameronroe: Yes, would that just be `<= :mavericks`?

cameronroe:
A lot of people still run Mavericks. I'm not interested in hosting my own tap.

Casks/plug.rb
vitorgalvao: Yes. But are you sure that is the latest version available on Mavericks? Seems very low.

cameronroe: Yes, it is.

cameronroe: This should be Uber simple to merge. Not sure why this is so confusing for everyone..

vitorgalvao:
> A lot of people still run Mavericks.

A minority. An even smaller minority is people running Mavericks, using Plug, and using homebrew-cask.

> I'm not interested in hosting my own tap.

But we also won’t keep supporting old things indefinitely. At some point, you’ll either have to upgrade, or host your own tap. It’s fine in this case for now, but [we’re not your personal tap](https://github.com/caskroom/homebrew-versions#acceptable-casks), we have more users.

cameronroe:
How is this a bad thing? It's covering multiple versions. That should be a good thing.

Casks/plug.rb
vitorgalvao: > This should be Uber simple to merge. Not sure why this is so confusing for everyone..

It’s only being confusing for you. You haven’t filled the *mandatory* checklist until you were asked twice (and you shouldn’t even have deleted it in the first place), **and you still haven’t fixed the conditional, which means this is still wrong for your own use case**. If you leave this as `<= :lion`, it won’t even fire when you try to install it (since you’re on a later version). Unless you set it to `:mavericks` *it won’t even work for you*.

It’s unreasonable to expect you just submit something incorrect and expect us to merge it. After you go away, we’re the ones that are going to have to keep maintaining everything. That’s why there are rules that need to be followed: so we can have a consistent and correct experience.

Yes, this should be simple to merge. Make it correct, and it will be. At that point, we will merge it.

Also, I’ll ask again (you’re making this hard): are you sure `0.10.5` is the latest version supported by Mavericks?

vitorgalvao:
> How is this a bad thing? It's covering multiple versions. That should be a good thing.

Once again, in this case it is fine for now. In many cases, it is not (due to overhead and other issues).

cameronroe:
All those checks pass. Is there anything else I'm missing here.

vitorgalvao:
Everything looks fine, now.

Issue #14952, opened 2015-11-04
Cask installed Duet Display app fails to launch 


timup:
```bash
~
$ brew cask install duet
==> Downloading http://dl.devmate.com/com.kairos.duet/1.4.0/1439081188/duet-1.4.
Already downloaded: /Library/Caches/Homebrew/duet-1.4.0_1439081188.zip
==> Symlinking App 'duet.app' to '/Applications/duet.app'
🍺  duet staged at '/opt/homebrew-cask/Caskroom/duet/1.4.0_1439081188' (355 files, 10M)

~
$ cd /applications

/applications
$ open duet.app
```

And then, nada.

Downloading the app from the duet display site, you are greeted with the following message after opening the app initially:

![Imgur](http://i.imgur.com/kuEIj4a.jpg?1)

Is it possible that this authorization is missed/bypassed somehow by the cask install?

Running El Cap, so maybe a new issue?

adityadalal924:
Are you sure `duet.app` is in /applications?

Please make sure that `/Applications/duet.app` actually exists, but otherwise please give us some more more information, as described in [Reporting Bugs](https://github.com/caskroom/homebrew-cask#reporting-bugs)

reitermarkus:
I just tested this. A symlink in `/Applications` doesn't work, because the app tries to copy itself to `/Applications` without prompting and if a symlink exists, it fails.

vitorgalvao:
Another one for https://github.com/caskroom/homebrew-cask/issues/12858. We’ll need to add a caveat.

timup:
Can confirm @reitermarkus.  Opening from ~/downloads copies the app to /applications before the permission prompt.

vitorgalvao:
Closing in favour of https://github.com/caskroom/homebrew-cask/issues/13201, which will fix this. At the moment we don’t have any other type of workaround that could be used, anyway.

toonetown:
Is there a way to suppress putting in the symlink?  I just got bitten by this one - `brew cask install duet` will automatically place the link in /Applications/duet.app - but then the first time you run it, it tries to copy itself /Applications/duet.app, but cannot because the link already exists.  Then it gets in this cycle of crashing and trying to restart.

This is, however, because I have `HOMEBREW_CASK_OPTS=--appdir=/Applications` set...so it is definitely an edge case in my setup.

[...]

PR #20479, opened 2016-04-11
updated duet (1.5.4.9)

tsparber:
#### Editing an existing cask

### DIFF
--- a/Casks/duet.rb
+++ b/Casks/duet.rb
@@ -1,6 +1,6 @@
 cask 'duet' do
-  version '1.5.2.4'
-  sha256 '8b8e64779bee391e4aed9c7539745c7bdcf2ddcd38d945d108a506e8b3daf169'
+  version '1.5.4.9'
+  sha256 'ee6334cb41f2d0a3d164b5c1de61d40f70d4cd9c829a35beba437ba9f6919d55'
 
   # d2ycb980mbr5lq.cloudfront.net was verified as official when first introduced to the cask
   url "https://d2ycb980mbr5lq.cloudfront.net/#{version.dots_to_underscores}/duet-#{version.dots_to_hyphens}.zip"


tsparber:
Updating the `appcast` checkpoint is hard, as manual changes get overwritten by `cask-repair` and `cask-repair` does not follow the redirect.

adidalal:
Pinging @vitorgalvao about `cask-repair`

adidalal:
Doesn't Duet not work, anyway? https://github.com/caskroom/homebrew-cask/issues/14952

vitorgalvao:
> `cask-repair` does not follow the redirect

It should. What version are you using (get it with `brew info cask-repair`; if you installed manually, there’s no way to know)?

adidalal:
Latest version reports:

```
• appcast is probably incorrect, as a non-200 (OK) HTTP response code was returned (302)
```

But what @tsparber is noting is that even if you manually edit, cask-repair will revert that edit when it rechecks after the manual edit

vitorgalvao:
Reverting the edit happens both with `checkpoint:` and `sha256`. It’s by design, as you’re not supposed to do them by hand (it’s an obtuse value that you need a computer to calculate anyway, so let the computer do all the work). For that reason I don’t intend to change that behaviour, as I believe the intent to be correct.

As for the message, it says “probably” precisely to allow some leverage. @tsparber swayed me in the past (regarding the script to find sparkle appcasts) on following appcast redirects, which is why `cask-repair` also (should be, anyway) follows them. It’s a warning, not an error, as it is just meant to ask you if you’re sure, in a way. It does not stop you submitting (not passing audit, for example, does).

So my question is if the latest version of `cask-repair` is really not following redirects, and hence not getting the correct `checkpoint:`. Making it get the correct `checkpoint:` is the solution, not allowing manually editing it.

vitorgalvao:
Just looked at it, and `cask-repair` seems to be working well. The problem, though, is that the `appcast` url is redirecting to an old version `1.5.1.1`, so no wonder no check is detected.

Which leads to the question: to what `checkpoint:` value are you trying to change it, @tsparber?

tsparber:
Sorry, I was a bit sloppy on this one. I'm using the latest version of `cask-repair` (0.23.1 as reported by `brew info`).
I was confused by the warning about the non-200 status code and run the command to get the checksum manually (from [doc/.../appcast](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md)), which indeed does not follow the redirect and got me a different (wrong) checkpoint.
So `cask-repair` works correctly (apart from the a little bit misleading warning), the [`appcast.md`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) might need an update.

> Doesn't Duet not work, anyway? #14952

I does work, because it moves itself to `/Application` and needs to started from there. What does not work is, when a user has the `--appdir` set to `/Application` (and so a symlink exists, which does not get overwritten by the app itself). The https://github.com/caskroom/homebrew-cask/issues/14952#issuecomment-182162172 suggests just to wait.

PR #20482, opened 2016-04-11
Delete rethinkdb.rb

ianwalter:
#### Editing an existing cask

These don't really apply. Simply removing rethinkdb cask since it conflicts with the brew version (not to mention it is out of date and the uninstaller doesn't work). Part of #15603.


vitorgalvao:
Fine by me.

@adidalal Any objections?

PR #20491, opened 2016-04-12
Fixes/addition for etcher.rb

thibmaek:


Casks/etcher.rb
```diff
@@ -5,7 +5,7 @@
   # resin-production-downloads.s3.amazonaws.com was verified as official when first introduced to the cask
   url "https://resin-production-downloads.s3.amazonaws.com/etcher/#{version}/Etcher-darwin-x64.dmg"
   name 'Etcher'
-  homepage 'https://etcher.io/'
```
vitorgalvao: The `/` at the end should remain.

PR #20529, opened 2016-04-14
Update Sketch to 3.7

jgarber623:
:point_up: Like it says on the tin.
#### Editing an existing cask
- [x] `brew cask audit --download sketch` is error-free.
- [x] `brew cask style --fix sketch` left no offenses.

### DIFF
--- a/Casks/sketch.rb
+++ b/Casks/sketch.rb
@@ -4,7 +4,7 @@ cask 'sketch' do
 
   url 'http://www.sketchapp.com/static/download/sketch.zip'
   appcast "http://www.sketchapp.com/appcast#{version.major}.xml",
-          checkpoint: '599f928ba500b61f006719136d8fc0fea15a4be4aab2378f06ee7e66706027f3'
+          checkpoint: '34257bf90fd33349e9acc7e7b1ebeef96299bd8e627c9e786bb2ad934b843c5b'
   name 'Sketch'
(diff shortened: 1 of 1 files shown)


Casks/sketch.rb
```diff
@@ -1,6 +1,6 @@
 cask 'sketch' do
-  version '3.6.1'
-  sha256 '13c449c0e951178ee8d15b6ea9dd41871f3338b3035b3b628861f084eaa41e9c'
+  version '3.7'
+  sha256 '8879703703babf42e33682377e10199368d8a66e53af0af1743c78a0e0b6a8cf'
 
   url 'http://www.sketchapp.com/static/download/sketch.zip'
   appcast "http://www.sketchapp.com/appcast#{version.major}.xml",
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

jgarber623:
@vitorgalvao Done, using the command:

```sh
curl --compressed --location "http://www.sketchapp.com/appcast3.xml" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256
```

PR #20535, opened 2016-04-14
Updated Vectr (latest)

pm5:


adidalal:
No, keep the versioning.
https://github.com/caskroom/homebrew-cask/pull/20543

pm5:
Any reason we should keep the versioning? Pardon me if I'm ignorant to previous discussions as I'm a newcomer.

adidalal:
We prefer versioned URLs, mentioned in https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md

------

vitorgalvao:
@pm5 It was decided a while ago. The majority of users prefer versioned downloads because it means we can have a `sha256`. Amongst other things, it provides some extra security.

PR #20536, opened 2016-04-14
fix bilibili sha256 mismatch

exherb:
#### Editing an existing cask
- [x] `brew cask audit --download bilibili` is error-free.
- [x] `brew cask style --fix bilibili` left no offenses.

Fix bilibili sha256 mismatch


adidalal:
Was there a re-tag? shasum shouldn't change without a corresponding version bump, generally

adidalal:

adidalal:
Verified and fixed in https://github.com/caskroom/homebrew-cask/pull/20544

exherb:
thanks.

PR #20595, opened 2016-04-17
updated sencha (6.1.2)

droid-Q:


Casks/sencha.rb
```diff
@@ -1,13 +1,13 @@
...
+  version '6.1.2'
+  sha256 '8b8ef41800055722698229561575db5d7c62359c87556f3b0a8b77a9b426dc89'
 
   url "https://cdn.sencha.com/cmd/#{version}/jre/SenchaCmd-#{version}-osx.app.zip"
   name 'Sencha Cmd'
   homepage 'https://www.sencha.com/products/sencha-cmd/'
   license :freemium
 
-  installer script: "SenchaCmd-#{version}-osx.app/Contents/MacOS/JavaApplicationStub",
+  installer script: "SenchaCmd-#{version}.15-osx.app/Contents/MacOS/JavaApplicationStub",
```
vitorgalvao: Specifically adding the hardcoded `15` here is bound to lead to bad results. We’ll need to be more creative with [`version` methods](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md).

PR #20596, opened 2016-04-17
Intel power gadget 303

ksato9700:
### Changes to a cask

Upgrade Intel Power Gadget to v3.0.3
#### Editing an existing cask
- [x] `brew cask audit --download intel-power-gadget` is error-free.
- [x] `brew cask style --fix intel-power-gadget` left no offenses.

### DIFF
--- a/Casks/intel-power-gadget.rb
+++ b/Casks/intel-power-gadget.rb
@@ -3,7 +3,7 @@ cask 'intel-power-gadget' do
   version '3.0.3'
   sha256 '93f052f5c1306272fceab7af5740d2837656242ab879436a6ce9e573ed9a274e'
 
-  url "https://software.intel.com/file/501089/download"
+  url 'https://software.intel.com/file/501089/download'
   name 'Intel Power Gadget'
   homepage 'https://software.intel.com/en-us/articles/intel-power-gadget-20'
   license :gratis
--- a/Casks/intel-power-gadget.rb
+++ b/Casks/intel-power-gadget.rb
@@ -1,14 +1,15 @@
+# coding: utf-8
 cask 'intel-power-gadget' do
-  version '3.0.1'
-  sha256 '538a792721604e2155b3a48caa4084db751a91b170e5fa62bf0331d3147f2239'
+  version '3.0.3'
+  sha256 '93f052f5c1306272fceab7af5740d2837656242ab879436a6ce9e573ed9a274e'
 
-  url "https://software.intel.com/sites/default/files/managed/59/39/IntelPowerGadgetMac#{version}.zip"
+  url "https://software.intel.com/file/501089/download"
   name 'Intel Power Gadget'
   homepage 'https://software.intel.com/en-us/articles/intel-power-gadget-20'
   license :gratis
 
   # this bogus-looking character accurately reflects an upstream error
-  container nested: 'IntelÆ Power Gadget.dmg'
+  container nested: "Intel® Power Gadget v#{version}.dmg"
 
   pkg 'Install Intel Power Gadget.pkg'
 
--- a/Casks/intel-power-gadget.rb
+++ b/Casks/intel-power-gadget.rb
@@ -1,4 +1,3 @@
-# coding: utf-8
 cask 'intel-power-gadget' do
   version '3.0.3'
   sha256 '93f052f5c1306272fceab7af5740d2837656242ab879436a6ce9e573ed9a274e'


Casks/intel-power-gadget.rb
```diff
@@ -1,14 +1,15 @@
+# coding: utf-8
```
vitorgalvao: Unecessary. Please remove.

Casks/intel-power-gadget.rb
```diff
@@ -1,14 +1,15 @@
...
 cask 'intel-power-gadget' do
-  version '3.0.1'
-  sha256 '538a792721604e2155b3a48caa4084db751a91b170e5fa62bf0331d3147f2239'
+  version '3.0.3'
+  sha256 '93f052f5c1306272fceab7af5740d2837656242ab879436a6ce9e573ed9a274e'
 
-  url "https://software.intel.com/sites/default/files/managed/59/39/IntelPowerGadgetMac#{version}.zip"
```
vitorgalvao: Are you sure there are no more versioned downloads? We prefer those.

ksato9700: Yes, I wanted to keep the version number but the link on the download page (https://software.intel.com/en-us/articles/intel-power-gadget-20) shows this link. I  hoped it was a redirect to another location with version number, but the header I get is as follows.

----
Server: nginx
Content-Type: application/zip
Content-Length: 2082827
X-Drupal-Cache: MISS
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
Content-Disposition: attachment; filename="=?UTF-8?B?SW50ZWzCriBQb3dlciBHYWRnZXQgdjMuMC4zLnppcA==?="
Content-Transfer-Encoding: binary
Pragma: no-cache
Cache-Control: must-revalidate, max-age=0, post-check=0, pre-check=0
Expires: Sun, 17 Apr 2016 02:51:00 GMT
Date: Sun, 17 Apr 2016 02:51:00 GMT
----

I believe "501089" in the path is a version specific number and I guess we won't get a newer version from the same URL.

vitorgalvao: In that case, seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

Casks/intel-power-gadget.rb
ksato9700: I'm sorry, emacs automatically inserted the line. I will fix it.

Casks/intel-power-gadget.rb
ksato9700: Ok, I will fix it and submit a PR again. Thanks!

Casks/intel-power-gadget.rb
vitorgalvao: Interesting. That might explain why we recently had another contribution with that same line (maybe that was you as well? I don’t recall).

ksato9700: I don't think it was me, but emacs (actually ruby-mode of emacs) inserts the line by default. I prevented it now by adding the following lines in my init.el.

```
(custom-set-variables
 '(ruby-insert-encoding-magic-comment nil))
```

You might want to add it in the CONTRIBUTING.md, or should I?

vitorgalvao: This is an extremely rare occurrence, so no need to add to the documentation.

Also, though we don’t need it, there’s nothing to say it isn’t actually useful to you in other situations, so no need to remove it from your config altogether, when we can just remove it from the cask.

Casks/intel-power-gadget.rb
ksato9700: By the way, I can make the change as you suggested, but because the file name of dmg file contained in the downloaded zip file includes a version number (e.g. "Intel® Power Gadget v3.0.3.dmg" in this time), I'm afraid the cask install will fail if Intel replace the download file with a newer version. Do you still want version: latest and sha256: no_check?

vitorgalvao: Oh, right. No, leave it be, then.

ksato9700: Thank you. I added a patch to remove the first line.

vitorgalvao: Thank you.

ksato9700:
Thank you!!

PR #20597, opened 2016-04-17
updated sencha (6.1.2)

droid-Q:


vitorgalvao:
Wait, what about the `15`, then? Isn’t it needed?

droid-Q:
@vitorgalvao  Yes.It isn't needed.

PR #20655, opened 2016-04-20
battle-net: update install filename

mattrobenolt:
#### Editing an existing cask

On my machine, I end up with this:

```
$ ls /opt/homebrew-cask/Caskroom/battle-net/latest
Battle.net-Setup.app
```

Not sure if the installer name has just changed over time?


victorpopkov:
> Not sure if the installer name has just changed over time?

Yes, sometimes this happens. Thank you for noticing and submitting a PR.

PR #20677, opened 2016-04-21
Add Wowza Streaming Engine v4.4.1

mandrean:
#### Adding a new cask
- [X] `brew cask audit --download wowza-streaming-engine` is error-free.
- [X] `brew cask style --fix wowza-streaming-engine` left no offenses.
- [x] `brew cask install wowza-streaming-engine` worked successfully.
- [X] `brew cask uninstall wowza-streaming-engine` worked successfully.

### DIFF
--- a/Casks/wowza-streaming-engine.rb
+++ b/Casks/wowza-streaming-engine.rb
@@ -1,0 +1,25 @@
+cask 'wowza-streaming-engine' do
+  version '4.4.1'
+  sha256 '2d87573f3aa0af2b23bbef9b327774ed83f6dea5de5c70d66a66545b7db693b4'
+
+  url "https://www.wowza.com/downloads/WowzaStreamingEngine-#{version.dots_to_hyphens}/WowzaStreamingEngine-#{version}-osx-installer.dmg"
+  appcast 'https://www.wowza.com/rss/wowzastreamingengineupdates.xml',
+          checkpoint: '8bb015c2f19e8f3748ba1215434cd624f9508043108c45cb1ffe38eac3b1a51d'
+  name 'Wowza Streaming Engine'
+  homepage 'https://www.wowza.com/products/streaming-engine'
+  license :commercial
+
+  auto_updates true
+
+  installer script: {
+                      executable: "WowzaStreamingEngine-#{version}-osx-installer.app/Contents/MacOS/installbuilder.sh",
+                      args:       ['--mode', 'osx'],
+                      sudo:       true,
+                    }
+
+  uninstall script: {
+                      executable: "/Applications/Wowza\ Streaming\ Engine\ #{version}/Wowza\ Streaming\ Engine\ Uninstall.app/Contents/MacOS/installbuilder.sh",
+                      args:       ['--mode', 'text'],
+                      sudo:       true,
+                    }
+end


mandrean:
**Note:** Right now, I'm doing:

mandrean:
**Note:** Right now, I'm passing the args `['--mode', 'osx']` in `installer script`.

I'd love to do `['--mode', 'text']` to keep it all in the console (and not opening the installer GUI at all), and it works beautifully if I run it directly in Terminal:

`$ WowzaStreamingEngine-4.4.1-osx-installer.app/Contents/MacOS/installbuilder.sh --mode text`

...but together with brew cask it doesn't seem to work. It scrolls past the license lightning-fast (instead of having to press ENTER manually a bunch of times), and then seems to get stuck in a loop or something when asking `[y/N]` if I accept the license.

As if it auto-inputs the ENTER key on repeat or something. Any ideas how I can get around this and get text-mode working?

vitorgalvao:
On a phone, so I can’t take a look at it right now, but the only installer that can call a GUI is `installer manual:`. We try not to call GUIs automatically, so the current solution can’t be accepted.

mandrean:
Are we allowed to prompt the user for some inputs before installing? Like a custom preflight script.

Because then I could run `['--mode', 'unattended']` and send in the serial/username/password as headless parameters.

Otherwise I need help with figuring out why `brew cask` gets stuck in some kind of loop when asking `[y/N]` for the license agreement.

vitorgalvao:
> Are we allowed to prompt the user for some inputs before installing? Like a custom preflight script.

No. Installation can’t block other operations.

vitorgalvao:
When using `['--mode', 'text']`, if you do the `install` command with the `--verbose` flag, what do you get?

mandrean:
I get a million of these, looping on forever:
```
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
==> Do you accept this agreement? [y/n]:
```

mandrean:
I experimented with prompting the user for some inputs (serial key, username), just for the lulz, and noticed that `brew cask` will just automatically continue without waiting for input + newline character

So perhaps this is a core issue in `brew cask`? That license agreement `[y/n]`:s when installing commercial software isn't supported?

vitorgalvao:
Doesn’t the install script have a mode to auto-accept the agreement? They usually do, when they can be installed in this manner.

mandrean:
That was my first thought as well! So I googled. This is the official answer:

> It is not possible since the user has to explicitly accept the license. It is done like that by design, based on the feedback of multiple customers through the years.

http://answers.bitrock.com/questions/2491/automatically-accept-license-agreement

vitorgalvao:
In that case, and unfortunately, it needs to use the aforementioned [`installer manual:`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/installer.md#installer-manual).

Casks/wowza-streaming-engine.rb
```diff
@@ -0,0 +1,25 @@
...
+  auto_updates true
+
+  installer script: {
+                      executable: "WowzaStreamingEngine-#{version}-osx-installer.app/Contents/MacOS/installbuilder.sh",
+                      args:       ['--mode', 'osx'],
+                      sudo:       true,
+                    }
+
+  uninstall script: {
+                      executable: "/Applications/Wowza\ Streaming\ Engine\ #{version}/Wowza\ Streaming\ Engine\ Uninstall.app/Contents/MacOS/installbuilder.sh",
```
vitorgalvao: No escaping spaces and quoting strings at the same time.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 70afae3cb84b1ee1e636591daf44c86e8b3ba199. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #20684, opened 2016-04-21
Adding new cask for DeckHub 1.0.1

chrismetcalf:


vitorgalvao:
There was a PR template you need to fix.

vitorgalvao:
There was a PR template you submitted. You need to fill it. It is not optional.

Casks/deckhub.rb
```diff
@@ -0,0 +1,11 @@
+cask 'deckhub' do
+  version '1.0.1'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

chrismetcalf:
I'll create a new PR with the template filled and the suggested change made.

PR #20702, opened 2016-04-22
Add otx v1.7 b566

mandrean:
#### Adding a new cask
- [X] `brew cask audit --download otx` is error-free.
- [X] `brew cask style --fix otx` left no offenses.
- [X] `brew cask install otx` worked successfully.
- [X] `brew cask uninstall otx` worked successfully.

### DIFF
--- a/Casks/otx.rb
+++ b/Casks/otx.rb
@@ -1,19 +1,16 @@
 cask 'otx' do
-  version '1.7'
+  version '1.7,b566'
   sha256 '116a9441cfed31c28e0f9b3aa26b82f2a7186d3c8ec4afd2173c2ad460e51ab8'
 
-  url "https://github.com/x43x61x69/otx/releases/download/v#{version}/otx_b566.zip"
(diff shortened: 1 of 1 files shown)


mandrean:
Question: Do we have to add the following caveats:

> If you would like to install the #{token} CLI tool as well, please run the following command:
> 
> ln -s '#{staged_path}/#{token}' '/usr/local/bin/#{token}

Or can we just add that as a postflight script and a zapper?

adidalal:
Add it in as a binary - I think the sublime cask should have an example. 

No need to add it as an uninstall item if you define it as an artifact (like sublime), we'll take care of it

Casks/otx.rb
```diff
@@ -0,0 +1,19 @@
+cask 'otx' do
+  version '1.7'
```
adidalal: You can make this 1.7,b66 and use version.before and after_comma methods in url

mandrean:
Awesome @adidalal! Thanks. Pushed fixes. Tested and works nicely with install/uninstall.

PR #20789, opened 2016-04-25
updated clementine (1.3.1)

sanyer:


vitorgalvao:
You’re using an old version of `cask-repair`.

sanyer:
@vitorgalvao oops, just did today `brew update`

vitorgalvao:
It’s not a big deal, but it is necessary to update.

The reason I know your version is old is that the PR template wasn’t submitted (which newer versions do).

adidalal:
`brew reinstall cask-repair` does the trick

PR #20845, opened 2016-04-28
updated typora cask to reflect the latest distro style

tinhkyaw:


vitorgalvao:
There’s nothing wrong with the current method, and it is preferred. [Here](https://github.com/caskroom/homebrew-cask/pull/20853)’s how to correctly fix the cask.

PR #20849, opened 2016-04-28
Update redcine-x-pro to 39.40030

timsutton:
#### Editing an existing cask

Is there a special token I can use to just refer to the file that was just downloaded in the `pkg` stanza? (i.e. `pkg :downloaded_file`)

Thanks!

### DIFF
--- a/Casks/redcine-x-pro.rb
+++ b/Casks/redcine-x-pro.rb
@@ -1,14 +1,13 @@
 cask 'redcine-x-pro' do
-  version '31.0.0'
-  sha256 '9393f84d839214ef4b692434ced8f4f9ecd12f8f49d8230d84c1bef9fe44d4db'
+  version '39.40030'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> Is there a special token I can use to just refer to the file that was just downloaded in the `pkg` stanza? (i.e. `pkg :downloaded_file`)

Not sure. Need to understand your use case. Could you elaborate?

timsutton:
Sorry if that was unclear. I mean that since brew-cask is downloading a file to a location on disk, I thought there might be a special symbol or variable (which would resolve to that item's basename) that I can use rather than giving a full string for the `pkg` filename. The cask works fine as it is now, just saw the duplication and was wondering if there was such a convenience available.

vitorgalvao:
Ah, I see. There’s nothing like that, no. Most casks do not download a naked `pkg`, so this duplication is a rare occurrence.

PR #20867, opened 2016-04-29
Add Datadog Agent.app v 5.7.4-1

Drachemann:
#### Adding a new cask
- [X] `brew cask audit --download datadog-agent` is error-free.
- [X] `brew cask style --fix datadog-agent` left no offenses.
- [X] `brew cask install datadog-agent` worked successfully.
- [X] `brew cask uninstall datadog-agent` worked successfully.

### DIFF
--- a/Casks/datadog-agent.rb
+++ b/Casks/datadog-agent.rb
@@ -1,5 +1,5 @@
 cask 'datadog-agent' do
-  version :5.7.4-1
+  version '5.7.4-1'
   sha256 :no_check
 
   # amazonaws.com/dd-agent is the official download host as per vendor documentation/install script (https://github.com/DataDog/dd-agent/blob/master/packaging/osx/install.sh)
--- a/Casks/datadog-agent.rb
+++ b/Casks/datadog-agent.rb
@@ -1,8 +1,8 @@
 cask 'datadog-agent' do
-  version :latest
+  version :5.7.4-1
   sha256 :no_check
 
-  # amazonaws.com is the official download host as per vendor install scripts/documentation
+  # amazonaws.com/dd-agent is the official download host as per vendor documentation/install script (https://github.com/DataDog/dd-agent/blob/master/packaging/osx/install.sh)
   url 'https://s3.amazonaws.com/dd-agent/datadogagent.dmg'
   name 'Datadog Agent'
   homepage 'https://www.datadoghq.com/'
@@ -19,5 +19,13 @@ cask 'datadog-agent' do
 
   zap delete: '/opt/datadog-agent'
 
-  caveats 'You will need to update /opt/datadog-agent/etc/datadog.conf and replace APIKEY with your api key.'
+  caveats <<-EOS.undent
+    You will need to update /opt/datadog-agent/etc/datadog.conf and replace APIKEY with your api key
+
+    If you ever want to start/stop the Agent, please use the Datadog Agent App or datadog-agent command.
+    It will start automatically at login, if you want to enable it at startup, run these commands:
+
+    sudo cp '/opt/datadog-agent/etc/com.datadoghq.agent.plist' /Library/LaunchDaemons
+    sudo launchctl load -w /Library/LaunchDaemons/com.datadoghq.agent.plist
+  EOS
 end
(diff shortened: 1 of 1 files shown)


Casks/datadog-agent.rb
```diff
@@ -0,0 +1,23 @@
+cask 'datadog-agent' do
+  version :latest
+  sha256 :no_check
+
+  # amazonaws.com is the official download host as per vendor install scripts/documentation
```
vitorgalvao: Could you clarify? Where exactly did you find the URL?

Casks/datadog-agent.rb
```diff
@@ -0,0 +1,23 @@
+cask 'datadog-agent' do
+  version :latest
```
vitorgalvao: Even if the `url` is unversioned, the `pkg` is versioned, so this should still be set.

Casks/datadog-agent.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  # amazonaws.com is the official download host as per vendor install scripts/documentation
+  url 'https://s3.amazonaws.com/dd-agent/datadogagent.dmg'
+  name 'Datadog Agent'
+  homepage 'https://www.datadoghq.com/'
+  license :bsd
+
+  pkg 'datadogagent-5.7.4-1.pkg'
+
+  preflight do
```
vitorgalvao: Why? Doesn’t the `pkg` take care of this?

Drachemann: Unfortunately no.. The installer won't complete successfully without the file /tmp/datadog-install-user containing a valid username for the install. You can also see this being used in their official install script: https://github.com/DataDog/dd-agent/blob/master/packaging/osx/install.sh

Casks/datadog-agent.rb
```diff
@@ -1,8 +1,8 @@
 cask 'datadog-agent' do
-  version :latest
+  version '5.7.4-1'
   sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

Casks/datadog-agent.rb
```diff
@@ -0,0 +1,31 @@
+cask 'datadog-agent' do
+  version '5.7.4-1'
+  sha256 :no_check
+
+  # amazonaws.com/dd-agent is the official download host as per vendor documentation/install script (https://github.com/DataDog/dd-agent/blob/master/packaging/osx/install.sh)
```
vitorgalvao: I meant clarify to me, not in the comment.

Casks/datadog-agent.rb
vitorgalvao: Why not use the script directly instead of the `pkg`, then? [We support it](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/installer.md#installer-script).

Drachemann: The install script in question is meant to be downloaded directly and run from a shell to download the dmg and run the os x installer. It also requires an api key environment variable. i.e. it's meant to be a "one line install". I assume this would break convention?
My understanding of the installer script stanza is that it should be used to "complete" an install, with a script downloaded with the original download archive.

Casks/datadog-agent.rb
Drachemann: Ahh, apologies. I'll clean up the comment but for clarification, the URL is contained within the Datadog documentation, for some reason the page describing this seems to require a login: https://app.datadoghq.com/account/settings#agent/mac

Also, their "one-line" install script contains the url: https://github.com/DataDog/dd-agent/blob/master/packaging/osx/install.sh

vitorgalvao: Good enough. Please make the comment `# s3.amazonaws.com/dd-agent was verified as official when first introduced to the cask`.

Casks/datadog-agent.rb
```diff
@@ -0,0 +1,31 @@
+cask 'datadog-agent' do
+  version '5.7.4-1'
+  sha256 'fa6f2b50da3c97696e1e72098fe7feaa802654f592e5f6954b1bf96332b78819'
+
+  # s3.amazonaws.com/dd-agent was verified as official when first introduced to the cask
+  url 'https://s3.amazonaws.com/dd-agent/datadogagent.dmg'
```
vitorgalvao: Can you point me to where the download is? I do not recall if I verified it.

LeoCavaille:
💯  sweet, thanks!

PR #20871, opened 2016-04-29
openzfs 1.5.2

ilovezfs:


adidalal:
Needs the PR template filled out, next time.

ilovezfs:
@adidalal For simple bumps like this, you may want to adopt a cask-ified version of https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb :)

adidalal:
Cask-repair already does that, afaik

vitorgalvao:
Yep. `cask-repair` also only needs the version (it calculates the `sha256` for you), takes care of `appcast` as well, and fills out the template automatically.

ilovezfs:
@vitorgalvao does it work if no `appcast` is available?

vitorgalvao:
Yes.

PR #20942, opened 2016-05-03
Add ProductHunt.app (1.0)

thii:
#### Adding a new cask
- [x] `brew cask audit --download producthunt` is error-free.
- [x] `brew cask style --fix producthunt` left no offenses.
- [x] `brew cask install producthunt` worked successfully.
- [x] `brew cask uninstall producthunt` worked successfully.

### DIFF
--- a/Casks/product-hunt.rb
+++ b/Casks/product-hunt.rb
@@ -1,0 +1,11 @@
+cask 'product-hunt' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://s3.producthunt.com/mac/ProductHunt.dmg'
+  name 'Product Hunt'
+  homepage 'https://www.producthunt.com/apps/mac'
+  license :mit
+
+  app 'Product Hunt.app'
+end


Casks/producthunt.rb
```diff
@@ -0,0 +1,11 @@
+cask 'producthunt' do
```
vitorgalvao: The token reference was not followed. This needs to be `product-hunt`.

Casks/producthunt.rb
```diff
@@ -0,0 +1,11 @@
...
+  version :latest
+  sha256 :no_check
+
+  url 'https://s3.producthunt.com/mac/ProductHunt.dmg'
+  name 'Product Hunt'
+  homepage 'https://www.producthunt.com'
```
vitorgalvao: Isn’t there a specific page for the app?

thii:
Fixed as comments. I thought no dash would make more sense since it matches their domain name.

vitorgalvao:
> I thought no dash would make more sense since it matches their domain name.

The token reference is very explicit why opinions can’t be taken into account and instead we have strict rules. It’s the way to prevent duplicates, which used to be a huge issue.

PR #20952, opened 2016-05-03
Updated cask Quitter.app - added auto_updates stanza

farmerchris:
#### Editing an existing cask


vitorgalvao:
If there’s an appcast, then we also need a `version` and `sha256`.

farmerchris:
I see - added version and checksum to the cask. The developer isn't providing a versioned download URL. Will this be OK?

vitorgalvao:
> The developer isn't providing a versioned download URL. Will this be OK?

Yes, that is fine. Thank you.

PR #20969, opened 2016-05-04
Upgrade BetterTouchTool to v1.69

IUSR:
Also changed download URL as hinted in the appcast XML.

### DIFF
--- a/Casks/bettertouchtool.rb
+++ b/Casks/bettertouchtool.rb
@@ -8,12 +8,12 @@ cask 'bettertouchtool' do
 
     license :gratis
   else
-    version '1.674'
-    sha256 '20b6cc1f9636d7d7c081c99b20b133dfd6d81049ae1800254c04b34b3fadf09e'
+    version '1.69'
+    sha256 'c85e4de353fbfd86a59949d977d961b1559b0bcd1e5c6d78ee1b645dfbf39afa'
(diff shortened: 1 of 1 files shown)


Casks/bettertouchtool.rb
```diff
@@ -8,12 +8,12 @@
 
     license :gratis
   else
-    version '1.674'
-    sha256 '20b6cc1f9636d7d7c081c99b20b133dfd6d81049ae1800254c04b34b3fadf09e'
+    version '1.69'
+    sha256 'c85e4de353fbfd86a59949d977d961b1559b0bcd1e5c6d78ee1b645dfbf39afa'
 
-    url "https://boastr.net/releases/btt#{version}.zip"
```
vitorgalvao: We prefer this one because it has the same domain as the homepage. Please revert it.

IUSR: @vitorgalvao ok, reverted. I was a bit concerened [https://boastr.net](https://boastr.net) might be less preferred by the author, because on the landing page BetterTouchTool's download link is [http://bettertouchtool.net/BetterTouchTool.zip](http://bettertouchtool.net/BetterTouchTool.zip), in its own domain; also versioned download links seem only on appcast XML. Maybe I overlooked.

PR #21066, opened 2016-05-09
Update SHA256 for Querious

garand:
Just updating the SHA256 for Querious.


adidalal:
There are multiple issues with this submission, please follow CONTRIBUTING and fill out the PR template.

You can also check out `cask-repair`, it makes Cask updates like this quite easy (details in [CONTRIBUTING](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md))

adidalal:
Will update it, as this Cask actually has multiple erorrs. WIll tag you in the PR so you can take a look, as well

PR #21092, opened 2016-05-10
Free Download Manager 5.1

yurikoles:


vitorgalvao:
You need to fill the PR template. You haven’t been doing so.

Casks/fdm.rb
```diff
@@ -0,0 +1,11 @@
+cask 'fdm' do
```
vitorgalvao: This has to be changed to `cask 'free-download-manager' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `free-download-manager.rb`.

Casks/fdm.rb
```diff
@@ -0,0 +1,11 @@
...
+  version :latest
```
vitorgalvao: There’s no reason for this to be `:latest`, when there’s a version in `url`.

yurikoles:
up & rebase

Casks/free-download-manager.rb
```diff
@@ -0,0 +1,11 @@
+cask 'free-download-manager' do
+  version '5.1'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

yurikoles:
up & rebase

Casks/free-download-manager.rb
```diff
@@ -0,0 +1,11 @@
+cask 'free-download-manager' do
+  version '5.1'
+  sha256 '99adbb76e7fded28e4ffdb82b3dc5a26d82aad0fec326c8f8164963d7c461bfd'
+
+  url "http://files2.freedownloadmanager.org/#{version.major}/#{version.major_minor}-latest/fdm.dmg"
```
vitorgalvao: Second one should be just `#{version}`.

yurikoles: done

PR #21114, opened 2016-05-11
Use more accurate license for official WhatsApp

monchote:
WhatsApp.app is free-to-use and closed source so :gratis seems more accurate as per https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/license.md


adidalal:
Thanks

PR #21154, opened 2016-05-12
Update opera to 37.0.2178.43

sudosun:
Update opera version


fanquake:
You also need to update the shasum

sudosun:
Complete

PR #21157, opened 2016-05-12
flash, flash-player, and flash-player-debugger 21.0.0.242

andyli:


fanquake:
You need to update the checkpoint in the Flash cask.

andyli:
The funny thing is that Adobe hasn't updated the appcast yet, so the checkpoint is still valid...

PR #21209, opened 2016-05-14
electron 1.1.0

Jerzerak:
#### Editing an existing cask

### DIFF
--- a/Casks/electron.rb
+++ b/Casks/electron.rb
@@ -1,6 +1,6 @@
 cask 'electron' do
-  version '1.0.2'
-  sha256 'eed3bc4c56ed25bb2b57be9c3c4f0eca3565b4c9a8ffff70c77d7058cf4e320e'
+  version '1.1.0'
+  sha256 '1f9a6a93d20b156ef9011b42fbc3b38c054271578312858ef30b7b6db3b60846'
 
   url "https://github.com/atom/electron/releases/download/v#{version}/electron-v#{version}-darwin-x64.zip"
   appcast 'https://github.com/atom/electron/releases.atom',


Casks/electron.rb
```diff
@@ -1,6 +1,6 @@
 cask 'electron' do
-  version '1.0.2'
-  sha256 'eed3bc4c56ed25bb2b57be9c3c4f0eca3565b4c9a8ffff70c77d7058cf4e320e'
+  version '1.1.0'
+  sha256 '1f9a6a93d20b156ef9011b42fbc3b38c054271578312858ef30b7b6db3b60846'
 
   url "https://github.com/atom/electron/releases/download/v#{version}/electron-v#{version}-darwin-x64.zip"
   appcast 'https://github.com/atom/electron/releases.atom',
```
victorpopkov: Updating the `:checkpoint` value in [appcast](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md#appcast) is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

Jerzerak: @victorpopkov appcast isn't reporting a new checkpoint, that's why I didn't update it

Jerzerak: I assume it's because 1.0.2 became the stable release after 1.1.0 was already tagged so the checkpoint doesn't need to be updated? Correct me if I'm wrong

victorpopkov: Yes, you are absolutely right. Thank you.

victorpopkov: Actually, it seems that there is a permanent redirect, so that's why the `curl` command mentioned earlier should also have a `--location` option since by default it doesn't do any redirects.

The correct checksum for this `appcast` at the moment should be:
```
535667359d6bd878fef8f16cf3e613a52cb9884604e149b869c7cd012919e243
```
Fixed in https://github.com/caskroom/homebrew-cask/pull/21223. Sorry about the hussle.

Jerzerak: @victorpopkov Just noticed this, thanks for the FYI

PR #21226, opened 2016-05-15
Cache Homebrew kegs and casks along with Ruby gems

jawshooah:


adidalal:
According to Travis support `/usr/local` should definitely not be cached - looks like build times jumped from 2 minutes -> 15 after this change as well, as the cacher times out.

Not sure of the impact of caching `/opt/homebrew-cask/Caskroom`

jawshooah:
What about `/usr/local/Cellar`?

adidalal:
Worth a shot, but it seemed to be okay without caching any of it - not sure exactly what needs to be cached but just going the gems should be enough

jawshooah:
Was thinking of caching the Cellar and Caskroom since we install some kegs
and casks during the build, but that doesn't take very long so it's
probably not worth it.

PR #21273, opened 2016-05-17
Update the sha256 for the-unarchiver

michaelgmcd:
I received an sh256 error when installing the-unarchiver. I've checked with the download from their webpage and the new checksum that I've added is correct.


adidalal:
version updated, will fix in another PR

adidalal:
https://github.com/caskroom/homebrew-cask/pull/21274

Issue #20783, opened 2016-04-25
Cask request: Dwarf Fortress Lazy Newb Pack


timvisher:
### Cask details

**Name** - Dwarf Fortress Lazy Newb Pack

**Homepage** - http://lazynewbpack.com/ or http://www.bay12forums.com/smf/index.php?topic=128960.0

**License** - n/a?

**Download URL** - Current https://dl.orangedox.com/szqcolfuOZSNoHr1YI/Macnewbie_0.9.18.dmg or http://dffd.wimbli.com/file.php?id=7922

**Description** - This is an easy way to get started playing Dwarf Fortress on Macs. It includes a bunch of plugins/fixes/utilities.

As a caveat, there are many utilities included that all are marked as _unsafe_ for opening. It would be nice if they all got marked as safe without user intervention. Most of them can be found in `LNP/utilities/**/*` although one of them is top level called `dfhack`

adidalal:
FYI to fix the quarantine: `xattr -dr com.apple.quarantine /opt/homebrew-cask/Caskroom/LNP/utilities/*` should do the trick

SirWindfield:
I can make a cask for it. Got some spare time.

brewingcode:
@timvisher give `brew update && brew cask install dwarf-fortress-lnp` a try. I was able to immediately fire up the game without any complaints about unsafe binaries.

PR #21288, opened 2016-05-18
get dwarf-fortress runnable: just install and type `dwarf-fortress`

brewingcode:

### DIFF
--- a/Casks/dwarf-fortress.rb
+++ b/Casks/dwarf-fortress.rb
@@ -35,4 +35,7 @@ cask 'dwarf-fortress' do
     end
   end
 
+  uninstall_preflight do
+    system 'cp', '-r', "#{staged_path}/df_osx/data/save", '/tmp/dwarf-fortress-save/'
+  end
 end
--- a/Casks/dwarf-fortress.rb
+++ b/Casks/dwarf-fortress.rb
@@ -1,4 +1,15 @@
 cask 'dwarf-fortress' do
+  module Utils
+    def self.update_sdl(name, url, path)
+      ohai "Replacing #{name}.framework from libsdl.org"
+      system 'curl', '-ksL', "https://www.libsdl.org/#{url}", '-o', "/tmp/#{name}.dmg"
+      system 'hdiutil', 'attach', "/tmp/#{name}.dmg"
+      system 'rm', '-rf', "#{path}/df_osx/libs/#{name}.framework"
+      system 'cp', '-r', "/Volumes/#{name}/#{name}.framework", "#{path}/df_osx/libs"
+      system 'hdiutil', 'detach', "/Volumes/#{name}"
+    end
+  end
+
   version '0.43.01'
   sha256 '7fe378b7aeee67f10a1f88a2341b8724edbf91795fd928506f01dfb403304d43'
 
@@ -7,5 +18,20 @@ cask 'dwarf-fortress' do
   homepage 'http://www.bay12games.com/dwarves/'
   license :gratis
 
-  suite 'df_osx', target: 'Dwarf Fortress'
+  shimscript = "#{staged_path}/df_wrapper"
+  binary shimscript, target: Hbc.homebrew_prefix.join('bin/df_osx')
+
+  preflight do
+    File.open(shimscript, 'w') do |f|
+      f.puts '#!/bin/sh'
+      f.puts "cd '#{staged_path}/df_osx' && ./df \"$@\""
+      FileUtils.chmod '+x', f
+    end
+  end
+
+  postflight do
+    # http://dwarffortresswiki.org/index.php/DF2014:Installation#Mac
+    Utils.update_sdl 'SDL_ttf', 'projects/SDL_ttf/release/SDL_ttf-2.0.11.dmg', staged_path
+    Utils.update_sdl 'SDL', 'release/SDL-1.2.15.dmg', staged_path
+  end
 end
--- a/Casks/dwarf-fortress.rb
+++ b/Casks/dwarf-fortress.rb
@@ -9,9 +9,10 @@ cask 'dwarf-fortress' do
 
   # shim script (https://github.com/caskroom/homebrew-cask/issues/18809)
   shimscript = "#{staged_path}/df_wrapper"
-  binary shimscript, target: Hbc.homebrew_prefix.join('bin/df_osx')
   depends_on cask: 'sdl-framework'
   depends_on cask: 'sdl-ttf-framework'
+
+  binary shimscript, target: Hbc.homebrew_prefix.join('bin/dwarf-fortress')
 
   preflight do
     File.open(shimscript, 'w') do |f|
--- a/Casks/dwarf-fortress.rb
+++ b/Casks/dwarf-fortress.rb
@@ -39,6 +39,7 @@ hdiutil detach /Volumes/#{name}
     # http://dwarffortresswiki.org/index.php/DF2014:Installation#Mac
     Utils.update_sdl 'SDL_ttf', 'projects/SDL_ttf/release/SDL_ttf-2.0.11.dmg', staged_path
     Utils.update_sdl 'SDL', 'release/SDL-1.2.15.dmg', staged_path
-    ohai "Run #{staged_path}/update_sdl to fetch and replace DF's SDL libraries"
   end
+
+  caveats "Run #{staged_path}/update_sdl to fetch and replace DF's SDL libraries"
 end
--- a/Casks/dwarf-fortress.rb
+++ b/Casks/dwarf-fortress.rb
@@ -1,20 +1,4 @@
 cask 'dwarf-fortress' do
-  module Utils
-    def self.update_sdl(name, url, path)
-      File.open("#{path}/update_sdl", 'a') do |f|
-        f.puts(<<-EOF)
-#!/bin/sh
-curl -ksL 'https://www.libsdl.org/#{url}' -o /tmp/#{name}.dmg
-hdiutil attach /tmp/#{name}.dmg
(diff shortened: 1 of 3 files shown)


vitorgalvao:
Downloading extra stuff via `curl` is a big no. Why is this better, and why is so much non-standard stuff needed for it?

brewingcode:
Darn, I figured extra curling was a big no, but figured I'd send the PR anyway. DF is ancient (14 years old) and only developed by a team of two, so the release format is...particular.

Is there some accepted method in `homebrew-cask` to supply a script that the user can optionally run after doing a `cask install`?

brewingcode:
image

adidalal:
Might https://github.com/caskroom/homebrew-cask/issues/20783 be of interest?

brewingcode:
I guess while I'm running around playing with casks involving DF, I can do #20783 as well, see #21290. :-)

I still want to get this change to `dwarf-fortress` in though, because it is *just* the base game, and it doesn't work as-is right now. In addition, the `dwarf-fortress-lnp` cask will move a little slower in terms of the version of DF that it includes. In fact, it's already behind:

cask | DF version | released
-------|----------------|-------------
dwarf-fortress | 43.01 | 2016-05-09
dwarf-fortress-lnp | 42.04 | 2015-12-26

Casks/dwarf-fortress.rb
```diff
@@ -7,5 +23,21 @@
...
+      f.puts "cd '#{staged_path}/df_osx' && ./df \"$@\""
+      FileUtils.chmod '+x', f
+    end
+  end
+
+  postflight do
+    # http://dwarffortresswiki.org/index.php/DF2014:Installation#Mac
+    Utils.update_sdl 'SDL_ttf', 'projects/SDL_ttf/release/SDL_ttf-2.0.11.dmg', staged_path
+    Utils.update_sdl 'SDL', 'release/SDL-1.2.15.dmg', staged_path
+    ohai "Run #{staged_path}/update_sdl to fetch and replace DF's SDL libraries"
```
adidalal: This should be in a caveats block instead of ohai

Casks/dwarf-fortress.rb
```diff
@@ -7,5 +23,21 @@
   homepage 'http://www.bay12games.com/dwarves/'
   license :gratis
 
-  suite 'df_osx', target: 'Dwarf Fortress'
+  shimscript = "#{staged_path}/df_wrapper"
+  binary shimscript, target: Hbc.homebrew_prefix.join('bin/df_osx')
+
+  preflight do
+    File.open(shimscript, 'w') do |f|
```
adidalal: Add the standard shim script comment, here

`# shim script (https://github.com/caskroom/homebrew-cask/issues/18809)`

Casks/dwarf-fortress.rb
```diff
@@ -1,4 +1,20 @@
 cask 'dwarf-fortress' do
+  module Utils
```
adidalal: I've never actually seen this used in a Cask so I'll defer to one of the other maintainers, @vitorgalvao

vitorgalvao:
I still don’t like that you’re downloading and mountig external things (essentially doing what a cask is for inside this cask. Can’t you push that to another cask and make this depend on the other?

Like we did [Lightroom 6.3 depend on Lightroom 6.0](https://github.com/caskroom/homebrew-cask/pull/15496).

brewingcode:
> Can’t you push that to another cask and make this depend on the other?
>
> Like we did Lightroom 6.3 depend on Lightroom 6.0.

Yep, I can do that. I was trying to avoid adding more casks, but if there's a precedent for it, I'm happy to follow it. Thanks for the link the Lightroom cask.

There will be two new casks I'll need, should each be their own PR, or can I lump them in here?

adidalal:
You can lump them - they are related and so one PR makes sense

adidalal:
@brewingcode Not actively checking this PR but please comment here when changes are done so we can review and merge ASAP, thanks!

brewingcode:
I think the changes are done....? Let me know!

adidalal:
Sure, I can take a look tonight 

@vitorgalvao if you could look through the shim script. Thanks!

Casks/dwarf-fortress.rb
```diff
@@ -2,10 +2,40 @@
...
   homepage 'http://www.bay12games.com/dwarves/'
   license :gratis
 
-  suite 'df_osx', target: 'Dwarf Fortress'
+  # shim script (https://github.com/caskroom/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/df_wrapper"
+  depends_on cask: 'sdl-framework'
+  depends_on cask: 'sdl-ttf-framework'
+
+  binary shimscript, target: Hbc.homebrew_prefix.join('bin/dwarf-fortress')
```
vitorgalvao: `binary shimscript, target: 'dwarf-fortress'` should be enough.

Casks/dwarf-fortress.rb
```diff
@@ -2,10 +2,40 @@
...
+    File.open(shimscript, 'w') do |f|
+      f.puts '#!/bin/sh'
+      f.puts "cd '#{staged_path}/df_osx' && ./df \"$@\""
+      FileUtils.chmod '+x', f
+    end
+  end
+
+  postflight do
+    Dir.chdir("#{staged_path}/df_osx/libs") do
+      [
```
vitorgalvao: Seems like a good case for a hash. Why an array of arrays?

Casks/dwarf-fortress.rb
```diff
@@ -2,10 +2,40 @@
...
+      ].each do |args|
+        name = "#{args[0]}.framework"
+        File.rename(name, "#{name}.orig")
+        File.symlink(Hbc.load(args[1]).staged_path.join(name), name)
+      end
+    end
+  end
+
+  uninstall_preflight do
+    system 'cp', '-r', "#{staged_path}/df_osx/data/save", '/tmp/dwarf-fortress-save/'
```
vitorgalvao: This is good. Smart way to not trash the data. There should be a `caveat` informing the user this will happen, though, otherwise you’ll only know if you inspect the cask, and this is important.

Casks/dwarf-fortress.rb
brewingcode: I tried that, but then the `dwarf-fortress` symlink ends up in `/usr/local/bin`, which doesn't work with Homebrews that are installed outside that default location.

Is there a reason `homebrew-cask` completely disregards that setting...? It seems like a bug: clearly `homebrew-cask` knows about custom prefixes because it already has the `Hbc.homebrew_prefix` property, so why doesn't it respect that property when symlinking binaries?

Casks/dwarf-fortress.rb
brewingcode: Just cleaner, IMO, both for typing and for reading. If there were more than just a 2x2 of parameters, or more than a few lines of code that used the parameters, I would have gone with a hash.

But I can change this from an array of arrays to an array of hashes, too.

Casks/dwarf-fortress.rb
brewingcode: Alright, I'll add the `caveat` shortly.

Casks/dwarf-fortress.rb
vitorgalvao: > but then the `dwarf-fortress` symlink ends up in `/usr/local/bin`

Not exactly. It ends up in your [`binarydir`](https://github.com/caskroom/homebrew-cask/blob/58fef4f3e63488418a09ab1bb4a201b2fa01b528/USAGE.md#options), which by default is `/usr/local/bin`, but may not be.

That is a feature, not a bug, and makes sense in a world where HBC and HB are decoupled (currently the case).

We’re [discussing changing that](https://github.com/caskroom/homebrew-cask/issues/21372).

Currently, though, `binary shimscript, target: 'dwarf-fortress'` is then the correct option.

Casks/dwarf-fortress.rb
vitorgalvao: Why an array of hashes? Why not just a hash?:

```ruby
{ 'SDL' => 'sdl-framework', 'SDL_ttf' => 'sdl-ttf-framework' }
```

vitorgalvao: Why an array of hashes? Why not just a hash?:

```ruby
{ 'SDL' => 'sdl-framework', 'SDL_ttf' => 'sdl-ttf-framework' }
```

Wouldn’t it work to do:

```ruby
Dir.chdir("#{staged_path}/df_osx/libs") do
  {
    'SDL' => 'sdl-framework',
    'SDL_ttf' => 'sdl-ttf-framework'
  }.each do |key,value|
    name = "#{key}.framework"
    File.rename(name, "#{name}.orig")
    File.symlink(Hbc.load(value).staged_path.join(name), name)
  end
end
```

brewingcode: Very true. Will change to that.

Casks/dwarf-fortress.rb
brewingcode: Yeah, I saw that `binarydir` command line option, and was surprised that it didn't take its default value from Homebrew's prefix.

OK, I'll change the cask, and maybe add a caveat about passing in `--binarydir "$(brew --prefix)"` or such during a `cask install`.

vitorgalvao: No need about adding that `caveat`, as this is not wrong behaviour, it is completely expected and documented.

brewingcode:
OK, changes pushed. Sorry for the delay, long week. :-)

vitorgalvao:
Thank you for your continued work on this.

brewingcode:
My pleasure, thanks for pointing me in the right direction, and for merging.

PR #21292, opened 2016-05-19
Add JANSI v10.4

ayatoy:
#### Adding a new cask

### DIFF
--- a/Casks/jansi.rb
+++ b/Casks/jansi.rb
@@ -1,0 +1,16 @@
+cask 'jansi' do
+  version '10.4'
+  sha256 '7e2e3b1979f87f0aa50ff37f8ff8d1d0f62ef3771d5308bdd160e6c5d57a17b2'
+
+  # dl.dropbox.com/u/43721500 was verified as official when first introduced to the cask
+  url "http://dl.dropbox.com/u/43721500/JANSI-MacOSX#{version}.zip"
+  name 'Japanese ANSI'
+  name 'JANSI'
+  homepage 'http://kenie33-jansi.blogspot.jp/'
+  license :unknown
+
+  depends_on macos: '>= :tiger'
+
+  artifact "JANSI-MacOSX#{version}/JANSI.icns", target: "#{ENV['HOME']}/Library/Keyboard Layouts/JANSI.icns"
+  artifact "JANSI-MacOSX#{version}/JANSI.keylayout", target: "#{ENV['HOME']}/Library/Keyboard Layouts/JANSI.keylayout"
+end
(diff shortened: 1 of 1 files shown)


adidalal:
@vitorgalvao What were we doing with keyboard layouts?

Casks/jansi.rb
```diff
@@ -0,0 +1,16 @@
+cask 'jansi' do
+  version '10.4'
+  sha256 '7e2e3b1979f87f0aa50ff37f8ff8d1d0f62ef3771d5308bdd160e6c5d57a17b2'
+
+  # dl.dropbox.com/u/43721500 was verified as official when first introduced to the cask
+  url "http://dl.dropbox.com/u/43721500/JANSI-MacOSX#{version}.zip"
```
vitorgalvao: This seems to be OS X’s version, not the layout’s version.

Casks/jansi.rb
```diff
@@ -0,0 +1,16 @@
+cask 'jansi' do
+  version '10.4'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

vitorgalvao:
I think we’ve been accepting them.

Casks/jansi.rb
ayatoy: Actually, I was lost what to do with the version. This is because, JANSI don't have specific version numbers despite layout exists for each version of OSX.
To be precise, the layouts exists as follows:
* JANSI-MacOSX10.4.zip is for OSX10.4 to 10.11
* JANSI-MacOSX10.2.zip is for OSX10.2 to 10.3
* In OSX10.0 to 10.1, only procedure of manual

`version '10.4'` is what I gave selfishly while I think that might be inappropriate.

ayatoy:
Actually, I was lost what to do with the version. This is because, JANSI don't have specific version numbers despite layout exists for each version of OSX.
To be precise, the layouts exists as follows:
* JANSI-MacOSX10.4.zip is for OSX10.4 to 10.11
* JANSI-MacOSX10.2.zip is for OSX10.2 to 10.3
* In OSX10.0 to 10.1, only procedure of manual

`version '10.4'` is what I gave selfishly while I think that might be inappropriate.

vitorgalvao:
Anything lower than `10.6` shouldn’t be bothered with, so don’t worry about that other one.

ayatoy:
I understand. I will fix this cask to `version :latest`, `sha256 :no_check` and relevant part.

PR #21297, opened 2016-05-19
knock2: Add install for Knock to Unlock

jasmas:


adidalal:
This is already added. https://github.com/caskroom/homebrew-cask/blob/master/Casks/knock.rb

PR #21303, opened 2016-05-19
updated alfred (3.0_652)

rajiv:
#### Editing an existing cask

### DIFF
--- a/Casks/alfred.rb
+++ b/Casks/alfred.rb
@@ -10,26 +10,20 @@ cask 'alfred' do
   auto_updates true
   accessibility_access true
 
-  app 'Alfred 3.app'
+  app "Alfred #{version.major}.app"
 
   postflight do
     suppress_move_to_applications key: 'suppressMoveToApplications'
   end
 
-  uninstall quit:       'com.runningwithcrayons.Alfred-2',
-            login_item: 'Alfred 2'
+  uninstall quit:       'com.runningwithcrayons.Alfred-3',
+            login_item: 'Alfred 3'
 
   zap delete: [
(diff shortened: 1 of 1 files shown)


Casks/alfred.rb
```diff
@@ -20,11 +20,16 @@
             login_item: 'Alfred 2'
 
   zap delete: [
+                '~/Library/Application Support/Alfred 3',
```
vitorgalvao: Remove the references to Alfred 2.

rajiv: i was following the existing example in the the cask as it has references to alfred 1 ("alfred" without the version number). if you are removing, zapping this cask, shouldn't it remove files from all the previous versions?

vitorgalvao: That’s still something we’re working on. We’ll save a snapshot of the cask you actually installed so we can remove that.

Casks/alfred.rb
```diff
@@ -20,11 +20,16 @@
             login_item: 'Alfred 2'
```
adidalal: Should this also be changed? Don't use Alfred, so not sure

vitorgalvao: Yes. And the `quit:` as well.

Casks/alfred.rb
```diff
@@ -10,7 +10,7 @@
   auto_updates true
   accessibility_access true
 
-  app 'Alfred 2.app'
+  app 'Alfred 3.app'
```
vitorgalvao: `app "Alfred #{version.major}.app"`

PR #21311, opened 2016-05-19
updated teamviewer (11.0.59131)

enavarrocu:
#### Editing an existing cask

### DIFF
--- a/Casks/teamviewer.rb
+++ b/Casks/teamviewer.rb
@@ -1,6 +1,6 @@
 cask 'teamviewer' do
-  version :latest
-  sha256 :no_check
+  version '11.0.59131'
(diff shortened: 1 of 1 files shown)


adidalal:
Unless you can find an appcast url, this needs to stay unversioned

enavarrocu:
What exactly that means?

adidalal:
version and shasum will break when teamviewer updates, because there's no way to detect that. Thus, with an unversioned url, we use latest and don't check the checksum

enavarrocu:
I see. The problem is when there is a new version you should use the --force flag to be able to install it.

adidalal:
Yes, that is correct

PR #21318, opened 2016-05-20
Add wireshark-chmodbpf

jasmas:
For some time now the homebrew wireshark formula has contained the following caveat:

> If your list of available capture interfaces is empty
> (default OS X behavior), try the following commands:
> 
>   curl https://bugs.wireshark.org/bugzilla/attachment.cgi?id=3373 -o ChmodBPF.tar.gz
>   tar zxvf ChmodBPF.tar.gz
>   open ChmodBPF/Install\ ChmodBPF.app
> 
> This adds a launch daemon that changes the permissions of your BPF
> devices so that all users in the 'admin' group - all users with
> 'Allow user to administer this computer' turned on - have both read
> and write access to those devices.
> 
> See bug report:
>   https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=3760

This seems like a good place to reference a cask. However, the file referenced in the caveat is an aging static attachment to an old bug id and not really suitable as the base of a cask.

This cask would install only 'org.wireshark.ChmodBPF.pkg' from the official Wireshark release pkg, but for it to work I need to write a temporary file containing:

```
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<array>
 <dict>
  <key>attributeSetting</key>
  <integer>0</integer>
  <key>choiceAttribute</key>
  <string>selected</string>
  <key>choiceIdentifier</key>
  <string>wireshark</string>
 </dict>
 <dict>
  <key>attributeSetting</key>
  <integer>0</integer>
  <key>choiceAttribute</key>
  <string>selected</string>
  <key>choiceIdentifier</key>
  <string>cli</string>
 </dict>
</array>
</plist>
```

The temporary file would then be read in by the installer script. I attempted doing this with some sort of here document or stdin, but none of those attempts worked or looked like very good form for a cask. 

This pull-request contains a draft cask with everything but the logic to write the temporary file which is required by the installer script.

What would be the best way to accomplish this?

### DIFF
--- a/Casks/wireshark-chmodbpf.rb
+++ b/Casks/wireshark-chmodbpf.rb
@@ -1,0 +1,89 @@
+cask 'wireshark-chmodbpf' do
+  version '2.0.3'
+  sha256 'a64dc77117a4408b48235297215032017e77822885ce3e19cd1065bfbea0ae57'
+
+  url "https://www.wireshark.org/download/osx/Wireshark%20#{version}%20Intel%2064.dmg"
+  name 'Wireshark-ChmodBPF'
+  homepage 'https://www.wireshark.org/'
+  license :gpl
+
+  installer script: '/usr/sbin/installer',
+            args:   [
+                      '-applyChoiceChangesXML',
+                      "#{staged_path}/chmodbpf_only.xml",
+                      '-package',
+                      "#{staged_path}/Wireshark #{version} Intel 64.pkg",
+                      '-target',
+                      '/',
+                    ]
+
+  preflight do
+    # shim script (https://github.com/caskroom/homebrew-cask/pull/21318)
+    FileUtils.touch "#{staged_path}/chmodbpf_only.xml"
+    chmodbpf_only = File.open "#{staged_path}/chmodbpf_only.xml", 'w'
+    chmodbpf_only.puts '<?xml version="1.0" encoding="UTF-8"?>'
+    chmodbpf_only.puts '<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">'
+    chmodbpf_only.puts '<plist version="1.0">'
+    chmodbpf_only.puts '<array>'
+    chmodbpf_only.puts '  <dict>'
+    chmodbpf_only.puts '    <key>attributeSetting</key>'
+    chmodbpf_only.puts '    <integer>0</integer>'
+    chmodbpf_only.puts '    <key>choiceAttribute</key>'
+    chmodbpf_only.puts '    <string>selected</string>'
+    chmodbpf_only.puts '    <key>choiceIdentifier</key>'
+    chmodbpf_only.puts '    <string>wireshark</string>'
+    chmodbpf_only.puts '  </dict>'
+    chmodbpf_only.puts '  <dict>'
+    chmodbpf_only.puts '    <key>attributeSetting</key>'
+    chmodbpf_only.puts '    <integer>0</integer>'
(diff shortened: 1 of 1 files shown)


jasmas:
For some time now the homebrew wireshark formula has contained the following caveat:

> If your list of available capture interfaces is empty
> (default OS X behavior), try the following commands:
> 
>   curl https://bugs.wireshark.org/bugzilla/attachment.cgi?id=3373 -o ChmodBPF.tar.gz
>   tar zxvf ChmodBPF.tar.gz
>   open ChmodBPF/Install\ ChmodBPF.app
> 
> This adds a launch daemon that changes the permissions of your BPF
> devices so that all users in the 'admin' group - all users with
> 'Allow user to administer this computer' turned on - have both read
> and write access to those devices.
> 
> See bug report:
>   https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=3760

This seems like a good place to reference a cask. However, the file referenced in the caveat is an aging static attachment to an old bug id and not really suitable as the base of a cask.

This cask would install only 'org.wireshark.ChmodBPF.pkg' from the official Wireshark release pkg, but for it to work I need to write a temporary file containing:

> <?xml version="1.0" encoding="UTF-8"?>
> <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
> <plist version="1.0">
> <array>
>  <dict>
>   <key>attributeSetting</key>
>   <integer>0</integer>
>   <key>choiceAttribute</key>
>   <string>selected</string>
>   <key>choiceIdentifier</key>
>   <string>wireshark</string>
>  </dict>
>  <dict>
>   <key>attributeSetting</key>
>   <integer>0</integer>
>   <key>choiceAttribute</key>
>   <string>selected</string>
>   <key>choiceIdentifier</key>
>   <string>cli</string>
>  </dict>
> </array>
> </plist>

The temporary file would then be read in by the installer script. I attempted doing this with some sort of here document or stdin, but none of those attempts worked or looked like very good form for a cask. 

What would be the best way to accomplish this?

adidalal:
Check out the shim script for minecraft server cask for an example

jasmas:
Thank-you @adidalal !

I've updated the cask accordingly.

Casks/wireshark-chmodbpf.rb
```diff
@@ -0,0 +1,89 @@
...
+    chmodbpf_only.puts '    <key>choiceIdentifier</key>'
+    chmodbpf_only.puts '    <string>cli</string>'
+    chmodbpf_only.puts '  </dict>'
+    chmodbpf_only.puts '</array>'
+    chmodbpf_only.puts '</plist>'
+    chmodbpf_only.close
+  end
+
+  postflight do
+    if Process.euid == 0
```
vitorgalvao: This shouldn’t be here.

You’re not supposed to run `brew cask` as root. We don’t support or recommend that.

jasmas: Does that mean we shouldn't check if someone did run brew cask as root and take some sane action if it was done? This is just copied from the wireshark cask.

vitorgalvao: It means we should have a better solution and make that check outside casks.

Lets leave it for now, though, since we don’t have a fix prepared, but this is one to remember.

jasmas: If it is not supported, wouldn't it make sense to build this check into homebrew-cask itself and exit with an error when launched as root?

PR #13966, opened 2015-09-23
Change artifact behavior to moving instead of symlinking

mwean:
This is a first step towards change in installation behavior mentioned in #13201.
I had to skip the `two_apps_correct` tests because I don't know the desired functionality (e.g. should we copy the source to both targets and delete the source?)

PR #21341, opened 2016-05-20
[WIP] Force overwrite artifacts on `--force` reinstall

claui:
This PR changes the behavior of `brew cask install --force` such that in case of a `Moved` artifact where a target already exists, `install --force` will remove the existing target before moving the staged artifact. For further details, see the commit messages.

This is intended to be an add-on to @mwean’s PR #13966, and I consider it WIP until it’s clear if anyone except myself even wants this. I have decided to make this add-on a PR on its own in order to keep topics separate and the discussion manageable.

Feedback is welcome. (The WIP tag is just there to make it clear that this PR is not to be merged _directly._)

@mwean In the event that this proposal turn out to be acceptable and useful, feel free to merge it into #13966 if you like!


adidalal:
Added a `on hold` tag so maintainers have another visual reminder to not merge as-is.

I'm in favor of this change, and seems to make sense. At least, I don't have a strong reason for preferring the original behavior, so 👍 from me

vitorgalvao:
Yes, [I’m fine with the behaviour](https://github.com/caskroom/homebrew-cask/pull/13966#issuecomment-220830387). I also agree it should be merged into #13966 directly.

claui:
Thanks for the feedback!  
Prior to merging into #13966, there’s two more things I’d like to fix:

- [ ] Figure out why the Travis build keeps failing (but passes in #13966)
- [ ] On `install --force`, go `sudo` to remove existing artifacts

vitorgalvao:
Thumb up on the second point.

As for the first, it isn’t necessary (not before merging, anyway), as Travis has been acting up a lot even on regular cask contributions. @adidalal has been making some tests to try and fix it. It’s definitely better.

That particular error is one we find a lot and isn’t related to the contents of the contribution itself.

claui:
Done.

[Asciicast of the results on my local machine](https://asciinema.org/a/7bo050vz0fzm8if9g70uig8ym)

@mwean Now would be a good time to merge this into #13966 if you like. I have kept the commits separate so you can reset back to yours anytime in case there are unexpected issues.

adidalal:
@claui Closing this as https://github.com/caskroom/homebrew-cask/pull/13966 now incorporates this PR.

PR #21368, opened 2016-05-22
Update skype.rb

mathbunnyru:
#### Editing an existing cask

Changes version from latest to '7.28.0.316' - gives an ability to update Skype using _brew_ after a new version is released (changing this number will lead to updating Skype).


adidalal:
We don't use this combination without an appcast that we can detect updates from. Use brew install --force to upgrade.

PR #21451, opened 2016-05-26
Re-add indigo cask

smurfless1:
Reverts caskroom/homebrew-cask#9019

This download is now available without registration.


fanquake:
This will need to be updated to the new cask template, see [here](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md).

vitorgalvao:
The homepage is also new (it redirects).

Instead of adding the removed cask, please try making it from scratch, to catch all those little differences.

PR #21516, opened 2016-05-28
Update aerial.rb to 1.2beta5

alistairmcmillan:
#### Editing an existing cask

This is pointing it to a beta version but it is the current version that we regularly point users to that has significant features that aren't in the non-beta version.

### DIFF
--- a/Casks/aerial.rb
+++ b/Casks/aerial.rb
@@ -1,6 +1,6 @@
 cask 'aerial' do
-  version '1.1'
-  sha256 '07365e3d0c8000bc354ab0780ad0b4106f2b9061b8b9da963bae7dd921bf8465'
+  version '1.2beta5'
+  sha256 '5b128fb10a1e088ece7feafc3e58f923036dc73b1efed0b8ac74df33b3e0daff'
 
   url "https://github.com/JohnCoates/Aerial/releases/download/v#{version}/Aerial.zip"
   appcast 'https://github.com/JohnCoates/Aerial/releases.atom',


Casks/aerial.rb
```diff
@@ -1,6 +1,6 @@
 cask 'aerial' do
-  version '1.1'
-  sha256 '07365e3d0c8000bc354ab0780ad0b4106f2b9061b8b9da963bae7dd921bf8465'
+  version '1.2beta5'
+  sha256 '5b128fb10a1e088ece7feafc3e58f923036dc73b1efed0b8ac74df33b3e0daff'
 
   url "https://github.com/JohnCoates/Aerial/releases/download/v#{version}/Aerial.zip"
   appcast 'https://github.com/JohnCoates/Aerial/releases.atom',
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

alistairmcmillan: I tried running `cask-repair` following the instructions under "Updating a Cask" and it worked right up to the submitting point and failed with "Permission denied (publicly)." But when it previewed the diff it only showed changes to the version and sha256 lines I included above.

And I just tried the command you listed above and it results in the same value that is already listed in the file.

I'm guessing because the atom file hasn't actually changed since you committed the original checkpoint value in December.

vitorgalvao: Looks like the case.

PR #21548, opened 2016-05-31
Maltego Classic. Version 4.0.0.8846

DaveCLowe:
New version - Maltego Classic. Version 4.0.0.8846
- [x] `brew cask audit --download maltego` is error-free.
- [x] `brew cask style --fix maltego` left no offenses.

### DIFF
--- a/Casks/maltego-classic.rb
+++ b/Casks/maltego-classic.rb
@@ -1,4 +1,4 @@
-cask 'maltego' do
+cask 'maltego-classic' do
   version '4.0.0.8846'
   sha256 '1fd90f884f40a6b2a7530b8e081b6b9f7d21cdaf53a52a6dc800034a85af84be'
 
--- a/Casks/maltego-classic.rb
+++ b/Casks/maltego-classic.rb
@@ -2,10 +2,11 @@ cask 'maltego-classic' do
   version '4.0.0.8846'
   sha256 '1fd90f884f40a6b2a7530b8e081b6b9f7d21cdaf53a52a6dc800034a85af84be'
 
-  url "https://www.paterva.com/malv#{version.sub(%r{^(\d+)\..*}, '\1\2')}/classic/Maltego.v#{version}.dmg"
+  url "https://www.paterva.com/malv#{version.major}/classic/Maltego.v#{version}.dmg"
   name 'Paterva Maltego'
   homepage 'https://www.paterva.com/web7/buy/maltego-clients/maltego.php'
   license :closed
 
-  app "Maltego Classic v#{version.sub(%r{\.\d+$}, '')}.app"
+  app "Maltego Classic v#{version.major_minor_patch}.app"
+
 end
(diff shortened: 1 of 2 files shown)


fanquake:
Is this the same product as Maltego Chlorine ?
Dose the older version need to be moved to the versions repo?

vitorgalvao:
In either case, this cask needs to be renamed, as it’s current token is wrong.

DaveCLowe:
Yes, this is the same product as Maltego Chlorine. The product has been rebranded a little, and "classic" is a bit misleading. It uses the same licence keys and transforms as the old version. 
I can't think of a reason as to why we would want to keep Chlorine available under the versions repo.

vitorgalvao:
In that case, you need to fill the PR template on your top post.

Casks/maltego.rb
```diff
@@ -1,11 +1,11 @@
 cask 'maltego' do
```
vitorgalvao: This has to be changed to `cask 'maltego-classic' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) must be followed, to avoid duplicates.

The file’s name also needs to be changed to `maltego-classic.rb`.

DaveCLowe: ok - so within the Applications info.plist, there is:
<key>CFBundleName</key>
<string>Maltego v4.0.0</string>

But, within Applications it appears as "Maltego Classic v4.0.0" - so I assume this is why I need to change it to maltego-classic?

Does this mean I need to add it as a whole new cask - rather than an update to maltego.rb?

Thanks :)

vitorgalvao: The info.plist is irrelevant. Read the linked document to understand why. Simply rename the file and change the line as pointed out in the previous comment, and we’re good to go.

Casks/maltego-classic.rb
```diff
@@ -0,0 +1,11 @@
+cask 'maltego-classic' do
+  version '4.0.0.8846'
+  sha256 '1fd90f884f40a6b2a7530b8e081b6b9f7d21cdaf53a52a6dc800034a85af84be'
+
+  url "https://www.paterva.com/malv#{version.sub(%r{^(\d+)\..*}, '\1\2')}/classic/Maltego.v#{version}.dmg"
+  name 'Paterva Maltego'
+  homepage 'https://www.paterva.com/web7/buy/maltego-clients/maltego.php'
+  license :closed
+
+  app "Maltego Classic v#{version.sub(%r{\.\d+$}, '')}.app"
```
vitorgalvao: Use instead `app "Maltego Classic v#{version.major_minor_patch}.app"`

Casks/maltego-classic.rb
```diff
@@ -0,0 +1,11 @@
+cask 'maltego-classic' do
+  version '4.0.0.8846'
+  sha256 '1fd90f884f40a6b2a7530b8e081b6b9f7d21cdaf53a52a6dc800034a85af84be'
+
+  url "https://www.paterva.com/malv#{version.sub(%r{^(\d+)\..*}, '\1\2')}/classic/Maltego.v#{version}.dmg"
```
vitorgalvao: Use instead `url "https://www.paterva.com/malv#{version.major}/classic/Maltego.v#{version}.dmg"`

vitorgalvao:
Please @mention me after you do, so I get an email to check back.

PR #21549, opened 2016-05-31
Require target for generic artifact

jawshooah:


vitorgalvao:
I thought we had an issue for this. Looks like we don’t.

Either way, thumbs up, it is a needed addition.

PR #21552, opened 2016-05-31
Show cached file location if download fails

l5oo00:
### Changes to the core

When I install a cask , I hope the console can show the cache file's path, so  I can download the file by other download tool, and move it to the `brew --cache` directory after rename the filename as same as the cache file path.

My network is not good!!!

### DIFF
--- a/lib/vendor/homebrew-fork/download_strategy.rb
+++ b/lib/vendor/homebrew-fork/download_strategy.rb
@@ -59,6 +59,7 @@ class Hbc::HbCurlDownloadStrategy < Hbc::AbstractDownloadStrategy
 
   def fetch
     ohai "Downloading #{@url}"
+    puts "The cache path for download: #{tarball_path}"
     unless tarball_path.exist?
       had_incomplete_download = temporary_path.exist?
       begin
--- a/lib/vendor/homebrew-fork/download_strategy.rb
+++ b/lib/vendor/homebrew-fork/download_strategy.rb
@@ -76,8 +76,7 @@ class Hbc::HbCurlDownloadStrategy < Hbc::AbstractDownloadStrategy
             msg = "File does not exist: #{@url.sub(%r[^file://], "")}"
           else
             msg = "Download failed: #{@url}"
-            msg << "\nThe desired cache filepath is #{tarball_path}, maybe you can download it by other tool,"
-            msg << "add move to the cache filepath, then run the install command again"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
I disagree. Finding the cache and downloading stuff there via other means is not supported behaviour. It works, but it still not the recommended way to do it.

Clearly you found the workaround for this, and hence this brings no new actual value for you anymore. For other users, this has never been an issue.

If you want to find the cache, you run the well documented `brew --cache` command. Printing the cache’s path every single download is just noise for the vast majority of cases.

l5oo00:
Thanks for your reply, but I think maybe you didn't understand what I meant. I will explain:

image

My focus point is the cached file's name(B in the image), not the brew cache path(`brew --cache`, C in the image). My need is  rename the file that I have downloaded by other tool to the cache path, use the right filename(B).

Why have I to do this? Because I have difficulty to  visit some foreign network **directly**, and most of the people in our country have the same problem.

OK, that's my explain, maybe it's not right.

Also thank you very much!

vitorgalvao:
I see what you mean. Though I still disagree with this particular solution, I believe we may get an acceptable one. I’m reopening to gather feedback from other maintainers.

Outputting the path at every install is noisy information in most cases. However, what we could do instead is *show the path if a download is interrupted*. That way users can get that information when a download files (the case where it might be useful to know that) and you can get it when you <kbd>⌃</kbd><kbd>C</kbd> (which you’d have to do anyway).

l5oo00:
Yeah! I agree with you. Thank you!

vitorgalvao:
> Yeah! I agree with you. Thank you!

In that case, care to update this PR to that?

adidalal:
Phrasing is awkward, maybe:  

`"\n Please retry the download or place the downloaded file at #{tarball_path}, and attempt the install again"`

vitorgalvao:
Since replacing the incomplete download and continuing the process is not something we want to encourage (again: “It works, but it still not the recommended way to do it”), we shouldn’t mention that possibility.

Something like: `The incomplete download is cached at `#{tarball_path}`.

adidalal:
LGTM

vitorgalvao:
Thank you for the initial suggestion, the discussion, and making the changes, @l5oo00. Merged.

l5oo00:
It is my pleasure.  Thank you very much!

PR #21569, opened 2016-05-31
Remove caveats from google-chrome and firefox

ptb:
With the change to moving artifacts to `/Applications`, it seems that the previously necessary caveats should removed, so as to not cause confusion.


vitorgalvao:
Nice catch. Thank you.

PR #21576, opened 2016-05-31
Updated gnubg (1.04.000-1010-x86_64)

tsparber:
Should this caveat be kept? (In case the user uses something else as `--appdir`)


vitorgalvao:
[Still required](https://github.com/caskroom/homebrew-cask/issues/12858#issuecomment-222825769).

PR #21604, opened 2016-06-01
CONTRIBUTING: MAS-only apps tip

adidalal:
Remembered this handy little project.

Might be useful for users, like @claui's [anecdote](https://github.com/caskroom/homebrew-cask/issues/21603#issuecomment-223072877)


claui:
Makes sense! It is a bit like a sister project.

Issue #21583, opened 2016-06-01
`brew cask zap _foo_` errors with: `Error: undefined method `fetch' for Hbc::SystemCommand:Class`


Arcank:
### Description of issue
`brew cask zap _foo_` errors with: `Error: undefined method `fetch' for Hbc::SystemCommand:Class`

### Output of `brew cask zap _foo_ --verbose`

```
% brew cask install gitbox
==> Downloading https://d1oa71y4zxyi0a.cloudfront.net/gitbox-1.6.2-ml.zip
Already downloaded: ~/Library/Caches/Homebrew/gitbox-1.6.2.zip
==> Verifying checksum for Cask gitbox
==> Moving App 'Gitbox.app' to '/Applications/Gitbox.app'
🍺  gitbox staged at '~/caskroom/gitbox/1.6.2' (0B)
% brew cask zap --verbose gitbox
==> Implied "brew cask uninstall gitbox"
==> Removing App: '/Applications/Gitbox.app'
==> Dispatching zap stanza
Error: undefined method `fetch' for Hbc::SystemCommand:Class
  Most likely, this means you have an outdated version of homebrew-cask. Please run:

      brew uninstall --force brew-cask; brew untap phinze/cask; brew update; brew cleanup; brew cask cleanup

  If this doesn’t fix the problem, please report this bug:

      https://github.com/caskroom/homebrew-cask#reporting-bugs

~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/artifact/base.rb:83:in `initialize'
~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/installer.rb:307:in `new'
~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/installer.rb:307:in `zap'
~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/cli/zap.rb:8:in `block in run'
~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/cli/zap.rb:5:in `each'
~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/cli/zap.rb:5:in `run'
~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/cli.rb:83:in `run_command'
~/homebrew/Library/Taps/caskroom/homebrew-cask/lib/hbc/cli.rb:121:in `process'
[...]

PR #21607, opened 2016-06-01
Fix `zap` regression introduced in 6500fa1; ref #21583

claui:
This PR fixes an issue with 6500fa1, in which the (internal) installer API was refactored in preparation for the big change. That refactoring missed one case, causing `zap` to break; see #21583.

@caskroom/maintainers This is a one-line fix but please have a look nonetheless. Thanks!

Caveats:
- I did not include tests with this fix. The `zap` command seems to have test coverage issues. We do have a `cli/zap_test` case but [its main path is commented out](https://github.com/caskroom/homebrew-cask/commit/00921dd5137bcaba95ca816f8fb7a185d4bd220e#diff-7fe581fc0f85395ebffe41217d5bdc83R36). When I uncomment that test, it keeps failing and I can’t figure out why. Shall I open an issue for this on its own?
- # 13966 has exposed a few more (unrelated) issues with `uninstall` and `zap` in edge cases when the user has insufficient permissions. This is easily fixable (using methods from 9664b9f); I am preparing a PR right now.

### DIFF
--- a/lib/hbc/installer.rb
+++ b/lib/hbc/installer.rb
@@ -304,7 +304,7 @@ class Hbc::Installer
     uninstall_artifacts
     if Hbc::Artifact::Zap.me?(@cask)
       ohai "Dispatching zap stanza"
-      Hbc::Artifact::Zap.new(@cask, @command).zap_phase
+      Hbc::Artifact::Zap.new(@cask, command: @command).zap_phase
     else
       opoo "No zap stanza present for Cask '#{@cask}'"
     end


jawshooah:
> When I uncomment that test, it keeps failing and I can’t figure out why. Shall I open an issue for this on its own?

Please do! Commented-out blocks of code do no one any good; let's figure out what's wrong and fix it before something else breaks.

adidalal:
LGTM, don't want zap broken for too long :)

claui:
Thanks @jawshooah and @adidalal.
I have opened an investigation PR at #21610. Feel free to help out!

PR #21629, opened 2016-06-02
Update orange.rb to 3.3.5

tomck:
Changes to cask:

Update version, update sha256, update app stanza to include target because oddly they changed from Orange.app to Orange3.app between 3.3.2 and 3.3.5, which would cause two copies of the app to be installed. (This could also be due to using a "dev" version before which may lack a major version number.)

Audit performed, style check passed, install succeeded and replaced old version (when using --force).

### DIFF
--- a/Casks/orange.rb
+++ b/Casks/orange.rb
@@ -1,11 +1,12 @@
 cask 'orange' do
-  version '3-3.2.dev0+2fc556d'
-  sha256 '9811008e071590c66abb2ac65355d55e466a59de2cd392b76fd328114094cf70'
+  version '3-3.3.5'
+  sha256 'e6a839ef6488d8becf979750701c8f395a97155b379ff8b5f7f12ab90a0cdb01'
 
   url "http://orange.biolab.si/download/files/Orange#{version}.dmg"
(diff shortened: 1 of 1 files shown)


Casks/orange.rb
```diff
@@ -1,11 +1,12 @@
...
+  sha256 'e6a839ef6488d8becf979750701c8f395a97155b379ff8b5f7f12ab90a0cdb01'
 
   url "http://orange.biolab.si/download/files/Orange#{version}.dmg"
   name 'Orange'
   homepage 'http://orange.biolab.si/'
   license :gpl
 
-  app 'Orange.app'
+  app 'Orange3.app', target: 'Orange.app'
+  # Target for consistency, old versions did not include "3" in the App name and would not be overwritten.
```
vitorgalvao: Overwriting is not a reason for renaming. Please remove the comment.

Also, it should be on top, if it were to be present.

tomck:
Removed the renaming, couldn't figure out how to squish the commits

vitorgalvao:
> couldn't figure out how to squash the commits

No worries, we can do that on merge.

PR #21658, opened 2016-06-03
[Audit] Warn if http is used without sha256

lukeadams:
### Changes to the core

`brew cask audit`
will show warnings for casks that use unencrypted URLs alongside `sha256 :no_check`

Example:

```
audit for scansnap-manager-s1500: warning
 - http used without sha256
```

### DIFF
--- a/lib/hbc/audit.rb
+++ b/lib/hbc/audit.rb
@@ -17,6 +17,7 @@ class Hbc::Audit
     check_sha256
     check_appcast
     check_url
+    check_https_or_sha
     check_generic_artifacts
     check_download
     self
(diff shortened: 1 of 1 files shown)


lukeadams:
I wasn't aware that audit warnings cause coveralls to fail.
Not sure what you guys think the best way to solve it is (if you even want to merge this). Make warnings non fatal (while making current warnings errors)? Add a "suggestion" notification?

vitorgalvao:
Thank you, but this isn’t an `audit` matter. As per the documentation, `audit` “verifies installability of Casks”, which this doesn’t affect at all.

You might be interested in [implementing `--require-sha`](https://github.com/caskroom/homebrew-cask/issues/18831), though.

lukeadams:
My reasoning was that it would be a check similar to the Sourceforge URL checks – give a warning so the author could fix it before publishing.
Haven't heard of that proposal – looking forward to it. 👍

vitorgalvao:
> My reasoning was that it would be a check similar to the Sourceforge URL checks – give a warning so the author could fix it before publishing.

But in this case, it isn’t always fixable.

> Haven't heard of that proposal – looking forward to it.

To be clear, no one is working on it at the moment, so you’re free to tackle it if you so desire.

PR #20405, opened 2016-04-09
Fix Google Music Manager (music-manager) pref pane

tjanson:
The app will copy the pref pane itself on first run (if it does not already exist), but then `brew uninstall` will not remove it.
By letting brew symlink the pref pane, it uninstalls properly.

PR #21659, opened 2016-06-03
music-manager.rb: do not move prefpane

vitorgalvao:
#### Editing an existing cask

Why would we be taking it? Presumably the app will take care of it.

### DIFF
--- a/Casks/music-manager.rb
+++ b/Casks/music-manager.rb
@@ -10,4 +10,6 @@ cask 'music-manager' do
   # Renamed for consistency: app name is different in the Finder and in a shell.
   # Original discussion: https://github.com/caskroom/homebrew-cask/pull/4282
   app 'MusicManager.app', target: 'Music Manager.app'
+
+  uninstall delete: '~/Library/Preference Panes/MusicManager.prefPane'
 end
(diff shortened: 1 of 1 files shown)


tjanson:
See #20405:

> The app will copy the pref pane itself on first run (if it does not already exist), but then brew uninstall will not remove it.
> By letting brew symlink the pref pane, it uninstalls properly.

tsparber:
@tjanson Would a `uninstall delete: '~/Library/Preference Panes/MusicManager.prefPane'` also work in this case?

vitorgalvao:
@tjanson I was going to suggest what @tsparber just mentioned. As to why I’m removing this, see https://github.com/caskroom/homebrew-cask/issues/21657#issuecomment-223638491. Since we now move instead of linking, this current cask essentially modifies the `app` (since it removes the `prefpane` from inside) which isn’t good at all.

Casks/music-manager.rb
```diff
@@ -10,4 +10,6 @@
   # Renamed for consistency: app name is different in the Finder and in a shell.
   # Original discussion: https://github.com/caskroom/homebrew-cask/pull/4282
   app 'MusicManager.app', target: 'Music Manager.app'
+
+  uninstall delete: '~/Library/Preference Panes/MusicManager.prefPane'
```
tjanson: This should be `~/Library/PreferencePanes/…` (no space), right?

tjanson:
Ah, cool to see that the move to move has been made. :+1:
Removing the pane during uninstall works for me, but the path is `"~/Library/PreferencePanes/MusicManager.prefPane"`.

Casks/music-manager.rb
tsparber: I just used the path from `man brew-cask` without checking. Turns out this needed some fixing (#21670).

Issue #18638, opened 2016-02-11
homebrew cask hangs when running install --force for parallels desktop


rasebo:
When running "brew cask install --force parallels-desktop", it hangs right after "==> Verifying checksum for Cask parallels-desktop":

### Running with the --debug flag:
```
==> Creating directories
==> Loading Cask definitions
==> Translating 'parallels-desktop' into a valid Cask source
==> Testing source class Hbc::Source::URI
==> Testing source class Hbc::Source::PathSlashRequired
==> Testing source class Hbc::Source::TappedQualified
==> Testing source class Hbc::Source::UntappedQualified
==> Testing source class Hbc::Source::Tapped
==> Success! Using source class Hbc::Source::Tapped
==> Resolved Cask URI or file source to '/usr/local/Library/Taps/caskroom/homebrew-cask/Casks/parallels-desktop.rb'
==> Cask instance dumps in YAML:
==> Cask instance toplevel:
--- !ruby/object:Hbc::Cask
token: parallels-desktop
sourcefile_path: !ruby/object:Pathname
  path: /usr/local/Library/Taps/caskroom/homebrew-cask/Casks/parallels-desktop.rb
dsl: !ruby/object:Hbc::DSL
  token: parallels-desktop
  version: !ruby/string:Hbc::DSL::Version
    str: !binary |-
      MTEuMS4yLTMyNDA4
    raw_version: 11.1.2-32408
  sha256: 6de7591cd24d837f109c895d23b56938b720debcb336c49634664e67b49f69e0
  url: !ruby/object:Hbc::URL
    uri: !ruby/object:URI::HTTP
      scheme: http
      user:
      password:
      host: download.parallels.com
      port: 80
      path: /desktop/v11/11.1.2-32408/ParallelsDesktop-11.1.2-32408.dmg
      query:
      opaque:
      registry:
      fragment:
      parser:
    user_agent:
    cookies:
    referer:
    using:
    revision:
    trust_cert:
    data:
  name:
  - Parallels Desktop
  homepage: https://www.parallels.com/products/desktop/
  license: !ruby/object:Hbc::DSL::License
    value: :commercial
  artifacts:
    :app: !ruby/object:Set
      hash:
        ? - Parallels Desktop.app
        : true
    :postflight: !ruby/object:Set
      hash:
        !ruby/object:Proc {}: true
    :uninstall: !ruby/object:Set
      hash:
        ? :delete:
          - /usr/bin/prl_convert
[...]

PR #21665, opened 2016-06-03
Fix deadlock in `Hbc::SystemCommand`

claui:
# Summary

The current implementation of `Hbc::SystemCommand` uses the `Open3` library **in a way that causes deadlocks.** In particular, a deadlock occurs whenever `STDERR` grows beyond a certain size (> 5000 lines) on the default OS&nbsp;X system Ruby.

This PR fixes the issue; it also provides failing tests to make the deadlock reproducible.

@caskroom/maintainers Please review! Thanks in advance :blush:
# Analysis

Apparently, the bug was well concealed; it was brought to my attention only today when user @winkelsdorf filed issue #18638 (thanks again!). The deadlock is triggered when `brew cask install parallels-desktop --force` is run with the app already installed; this triggers a `/bin/chmod -R -- u+rwx '/Applications/Parallels Desktop.app'`, producing an extremely long result on standard error, which in turn triggers the bug.
## Deadlock

The Ruby documentation, section Open3 [summarizes quite well why the deadlock occurs](http://ruby-doc.org/stdlib-2.0.0/libdoc/open3/rdoc/Open3.html#method-c-popen3):

> You should be careful to avoid deadlocks. Since pipes are fixed length buffer, ::popen3(“prog”) {|i, o, e, t| o.read } deadlocks if the program generates many output on stderr. You should be [sic] read stdout and stderr simultaneously (using thread or IO.select). However if you don’t need stderr output, ::popen2 can be used. If merged stdout and stderr output is not a problem, you can use ::popen2e. If you really needs stdout and stderr output as separate strings, you can consider ::capture3.
## Root cause

The offending code snippet from `system_command.rb` is:

``` ruby
raw_stdin, raw_stdout, raw_stderr, raw_wait_thr = Open3.popen3(*to_exec)

…

while line = raw_stdout.gets # <-- deadlock occurs here
  processed_stdout << line
  ohai line.chomp if options[:print_stdout]
end

while line = raw_stderr.gets
  processed_stderr << line
  ohai line.chomp if options[:print_stderr]
end
```

The deadlock occurs during the first `while` loop when `raw_stderr` happens to grow beyond a certain size.

In that particular case, both `raw_stdout` and `raw_stderr` will block until someone reads from `raw_stderr`. In the code above, this condition is not met until the second `while` loop.

In other words: things go south because the first `while` loop waits on `raw_stdout`, which waits on `raw_stderr` to be consumed, which in turn waits for the `while` loop to finish; a deadlock.
## Triggering the failure

When Parallels Desktop.app is installed, the bundle is owned by `root` and read-only to everyone else. When `brew cask install --force` is subsequently called, HBC needs to remove the existing app. It is actually quite smart in handling this; as a first gentle attempt before escalating to `sudo`, it tries to make the app world-writeable:

```
/bin/chmod -R -- u+rwx '/Applications/Parallels Desktop.app'
```

In general, this particular `chmod` is expected to fail; it is only a first attempt to avoid `sudo` whenever possible. In our example however, the failing `chmod` causes a lot of noise on STDERR (> 5000 lines) which triggers the deadlock in `Hbc::SystemCommand`.
# The fix

Following the suggestion from the Ruby documentation, I have replaced the hard-coded order of streams in `Hbc::SystemCommand` by polling `IO.select`, which guarantees deadlock-free stream selection.

To clean up the code a bit, I also extracted a few methods and ivars while at it.
# Testing
## Failing specs

I have included a spec which should detect, and fail on, this particular deadlock when run with the default OS&nbsp;X system Ruby. For details, see `spec/cask/system_command_spec.rb`.

The new specs are **in a separate commit** within this PR. They be run against either the old and new implementation.
## Additional test dependencies

To make this case testable, and allow for the spec to be more expressive, I have added two new test dependencies: `rspec-its` and `rspec-wait`. Both gems are released under the MIT license ([1](https://github.com/rspec/rspec-its/blob/master/LICENSE.txt) [2](https://github.com/laserlemon/rspec-wait/blob/master/LICENSE.txt)). I strongly believe that the former package is going to benefit our specs greatly; the latter is needed for deadlock detecting. In my opinion, either package is well worth being included in our testing toolchain.
## Running the tests

To install the added test dependencies, you might want need to run `bundle install` once.

The quickest way to run the specs is:

``` sh
bundle exec rake test:rspec
```

If you want to **watch the spec fail,** simply `git reset` one commit back in history, and execute the specs there.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,9 +16,11 @@ group :release do
 end
 
 group :test do
-  gem 'coveralls', :require => false
+  gem 'coveralls', require: false
   gem 'minitest', '5.4.1'
   gem 'minitest-reporters'
-  gem 'mocha', '1.1.0', :require => false
+  gem 'mocha', '1.1.0', require: false
   gem 'rspec', '~> 3.0.0'
+  gem 'rspec-its', require: false
+  gem 'rspec-wait', require: false
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -64,9 +64,14 @@ GEM
     rspec-expectations (3.0.4)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.0.0)
+    rspec-its (1.2.0)
+      rspec-core (>= 3.0.0)
+      rspec-expectations (>= 3.0.0)
     rspec-mocks (3.0.4)
       rspec-support (~> 3.0.0)
     rspec-support (3.0.4)
+    rspec-wait (0.0.8)
+      rspec (>= 2.11, < 3.5)
     rubocop (0.40.0)
       parser (>= 2.3.1.0, < 3.0)
       powerpack (~> 0.1)
@@ -104,6 +109,8 @@ DEPENDENCIES
   rake
   ronn (= 0.7.3)
   rspec (~> 3.0.0)
+  rspec-its
+  rspec-wait
   rubocop-cask (~> 0.6.1)
 
 BUNDLED WITH
--- a/lib/hbc/system_command.rb
+++ b/lib/hbc/system_command.rb
@@ -2,64 +2,112 @@ require 'open3'
 require 'shellwords'
 
 class Hbc::SystemCommand
+  attr_reader :command
+
   def self.run(executable, options={})
-    command = _process_options(executable, options)
-    processed_stdout = ''
-    processed_stderr = ''
-
-    command[0] = Shellwords.shellescape(command[0]) if command.size == 1
-
-    to_exec = command.map { |arg| arg.respond_to?(:to_path) ? File.absolute_path(arg) : String(arg) }
-
-    odebug "Executing: #{to_exec.utf8_inspect}"
-
-    raw_stdin, raw_stdout, raw_stderr, raw_wait_thr = Open3.popen3(*to_exec)
-
-    if options[:input]
-      Array(options[:input]).each { |line| raw_stdin.puts line }
-    end
-    raw_stdin.close_write
-
-    while line = raw_stdout.gets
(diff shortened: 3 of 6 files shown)


winkelsdorf:
Great finding, and if I might add - one of the best, quickest and most substantial bug analysis I've seen on GH in a long time!

Great work, thanks for spending your time on this! 💯

jawshooah:
Awesome to have you here, @claui. LGTM :+1:

vitorgalvao:
Amazing detective work, here, and so detailed! Big thumbs up to this.

claui:
Thanks so much @vitorgalvao!

Do we tolerate the MIT thing as well? I’ve seen comments of yours suggesting between the lines that this project is not exactly in love with that license. :smile:

vitorgalvao:
> Do we tolerate the MIT thing as well? I’ve seen comments of yours suggesting between the lines that this project is not exactly in love with that license.

I’m guessing you’re referring to https://github.com/caskroom/homebrew-cask/issues/14877. To clarify, I’m not at all opposed to using or including projects licensed under the MIT or other permissive licenses, I just find them a bit silly since their terms sound specific but are actually vague (“At which point is the code so modified that it is no longer the original?”) and they’re so close to the public domain they could very well just be public domain and benefit everyone on a greater scale.

I’d also like to be clear those views were my own and not necessarily the views of the project as a whole (after all, the issue was open as a question). That said, every other maintainer supported the notion, so in that sense I guess those are indeed the views of the project.

But just to reiterate, no problem at all in using other people’s less permissive stuff, I’d just like for our stuff to be as permissive as possible.

Issue #21672, opened 2016-06-04
filebot symlink source is not there


ajc85:
following instructions at: https://www.filebot.net/forums/viewtopic.php?f=4&t=2049

When I run 'brew cask install filebot'

==> Downloading http://downloads.sourceforge.net/project/filebot/filebot/FileBot
######################################################################## 100.0%
==> Verifying checksum for Cask filebot
==> It seems there is already an App at '/Applications/FileBot.app'; not moving.
Error: It seems the symlink source is not there: '/Applications/FileBot.app/Contents/MacOS/filebot.sh'

How to resolve this error?

vitorgalvao:
> It seems there is already an App at

Run it with `--force`.

vitorgalvao:
> It seems there is already an App at

Run it with `--force`.

> Error: It seems the symlink source is not there: '/Applications/FileBot.app/Contents/MacOS/filebot.sh'

Perhaps the file no longer exists in newer versions. Is that the case, @rednoah?

rednoah:
`/Applications/FileBot.app` is probably a manual install, or a Mac App Store install. Not sure how brew cask should deal with that ideally.

Ideally I'd like to have the "brew app" moved to something like `/Applications/FileBot-brew.app` but that pull request was rejected.

vitorgalvao:
Thank you for the quick reply @rednoah.

If installed with `--force`, the behaviour is to overwrite the existing one (not sure we escalate to `sudo`, but I do remember discussing it).

If this is indeed a case of MAS and not-MAS app, then removing the existing one and installing again will fix it. Closing.

ajc85:
Thanks guys - uninstalling MAS version and installing through brew cask fixed this issue.

PR #21683, opened 2016-06-04
FileBot 4.7: workaround MAS conflict

rednoah:
The brew cask version of FileBot and the MAS version seem to interfere with each other.

See https://github.com/caskroom/homebrew-cask/issues/21672

### DIFF
--- a/Casks/filebot.rb
+++ b/Casks/filebot.rb
@@ -8,8 +8,8 @@ cask 'filebot' do
   homepage 'https://www.filebot.net/'
   license :gpl
 
-  app 'FileBot.app'
-  binary "#{appdir}/FileBot.app/Contents/MacOS/filebot.sh", target: 'filebot'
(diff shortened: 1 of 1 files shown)


jawshooah:
`brew cask install --force` resolves #21672. Not sure I understand why this is necessary. Why would you want to have the Cask and the MAS versions installed side by side?

vitorgalvao:
As @jawshooah said, this issue was resolved. This is not an acceptable use of `target`. Closing.

rednoah:
`--force` is a workaround, not a solution. I want users to be able to install both the MAS as well as the brew version and there is reason to have both of them around.

I prefer this solution, but I can rename/remove the app bundle at it's download source if I can't fix this issue with `target`.

rednoah:
Actually, nevermind, probably better to change the name at the source. I'll update the cask accordingly next release.

PR #21760, opened 2016-06-07
avast.rb: added must_succeed false

vitorgalvao:
#### Editing an existing cask

### DIFF
--- a/Casks/avast.rb
+++ b/Casks/avast.rb
@@ -10,8 +10,9 @@ cask 'avast' do
   pkg 'Avast Mac Security.pkg'
 
   uninstall script: {
-                      executable: '/Library/Application Support/Avast/hub/uninstall.sh',
-                      sudo:       true,
+                      executable:   '/Library/Application Support/Avast/hub/uninstall.sh',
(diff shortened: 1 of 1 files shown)


Casks/avast.rb
```diff
@@ -10,8 +10,9 @@
   pkg 'Avast Mac Security.pkg'
 
   uninstall script: {
-                      executable: '/Library/Application Support/Avast/hub/uninstall.sh',
-                      sudo:       true,
+                      executable:   '/Library/Application Support/Avast/hub/uninstall.sh',
+                      must_succeed: false, # A non-0 exit code may be given even if the uninstall fails (https://github.com/caskroom/homebrew-cask/issues/21740#issuecomment-224094946).
```
vitorgalvao: Noticed the error in the comment. Fixing now.

claui:
I still would like to double-check whether this option maybe needs to be enabled in the core first.

vitorgalvao:
> I still would like to double-check whether this option maybe needs to be enabled in the core first.

It seems like it is, which is why I went ahead with it. [We documented it](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#summary-of-keys) and I retried the install/uninstall of this cask in a clean VM. It previously failed but now works.

PR #21765, opened 2016-06-07
updated jabref (3.4)

joeygravlin:
#### Editing an existing cask

### DIFF
--- a/Casks/jabref.rb
+++ b/Casks/jabref.rb
@@ -1,11 +1,11 @@
 cask 'jabref' do
-  version '3.3'
-  sha256 '949bd9a1a3bb0c71dccfa73e6f9c62627071534250db79bdd051f79f75a3c2a8'
+  version '3.4'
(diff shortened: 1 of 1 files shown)


joeygravlin:
So...I'm a little embarrassed that I didn't catch this before submitting a pull request, but after actually installing the new version, I realized that there was an issue with their 3.4 build on OS X. After all of two seconds of browsing jabref's issues page, I found someone had already reported both the issue and a temporary fix in [jabref issue #1481](https://github.com/JabRef/jabref/issues/1481).

Anyway, I can confirm @kojiyokota's report that running the app using `java -jar path/to/JabRef.app/Contents/java/app/JabRef-3.4.jar` works fine. However, I'll leave it to your discretion, @vitorgalvao, whether this will suffice in the interim, or if it would be better to just roll back this pull-request.

vitorgalvao:
@joeygravlin Should be fine. Seems like they’ll look into a fix soon.

PR #21840, opened 2016-06-10
Moving the .app to /Applications would make the DevCenter crash at st…

gfelisberto:
#### Editing an existing cask


vitorgalvao:
This is not a solution. `binary` is not for this. Does `DevCenter.app` need its enclosing directory? If so, use [`suite`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/suite.md).

PR #21843, opened 2016-06-10
Add pgAdmin4.app v1.0-beta1.

lifepillar:
#### Adding a new cask
- [x] When naming the cask, followed the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md). **Note that pgAdmin4 is a totally different app compared to pgAdmin3, so the 4 in the name must be retained. For the same reason, this app does not make pgAdmin3 obsolete (yet), and both apps may coexist.**
- [x] `brew cask audit --download pgadmin4` is error-free.
- [x] `brew cask style --fix pgadmin4` left no offenses.
- [x] `brew cask install pgadmin4` worked successfully.
- [x] `brew cask uninstall pgadmin4` worked successfully.

### DIFF
--- a/Casks/pgadmin4.rb
+++ b/Casks/pgadmin4.rb
@@ -1,0 +1,22 @@
+cask 'pgadmin4' do
+  # note: "4" is not a version number, but indicates a different app (vs pgadmin3)
+  version '1.0-beta1'
+  sha256 'ba004dcedff2a5d9bcc02ba50b3b0b429cd05a0f2ceba962a8afbea9120448f2'
+
+  # postgresql.org is the official download host per the vendor homepage
+  url "https://ftp.postgresql.org/pub/pgadmin3/pgadmin4/v#{version}/osx/pgadmin4-#{version}.dmg"
+  name 'pgAdmin4'
+  homepage 'http://pgadmin.org'
+  license :oss
+  gpg "#{url}.sig",
+      key_id: 'e0c4ceeb826b1fda4fb468e024adfaaf698f1519'
+
+  app 'pgAdmin 4.app'
+
+  zap delete: [
+                '~/.pgadmin',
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Seems to me like the vendor is exactly the same (same website).

Why shouldn’t this just be in [caskroom/versions](https://github.com/caskroom/homebrew-versions)?

lifepillar:
That was a comment from pgadmin3's cask. I have updated it to reflect the fact that this cask is a different application.

>Why shouldn’t this just be in caskroom/versions?

This is not a major upgrade of pgAdmin3, but a complete rewrite with a totally different UI, written in Python/Javascript rather than C++, and deployable as a desktop or web application (pgAdmin3 is only desktop). I believe that not everybody would want to switch from pgAdmin3 (pgAdmin4 is a ~150MB behemoth, pgAdmin3 is ~30MB and has lighter requirements).

vitorgalvao:
But like you said (emphasis mine)

> this app does not make pgAdmin3 obsolete **(yet)**

So presumably version 3 will eventually stop being updated, and this one will take over. Is that correct?

On the website they don’t make a distinction. They call the app `pgadmin` and that’s that. They refer to the number as a version, not as part of the name. Even if they are different and this is a rewrite, if it is an app that will appear in sequence of the previous one, it belongs in [caskroom/versions](https://github.com/caskroom/homebrew-versions) until such time as it is stable, at which point it’ll be moved to this repo, and version 3 will then belong in [caskroom/versions](https://github.com/caskroom/homebrew-versions).

lifepillar:
I am not involved in that project, so I can't speak for the developers: that sentence was just a guess of mine. Anyway, it is fine with me to open a pull request in caskroom/versions.

vitorgalvao:
Please do. You may also remove the comment at the top.

PR #21847, opened 2016-06-10
little-snitch: update zap

jasmas:
Updated to match developer documentation:
https://www.obdev.at/support/index.html?product=LS&topic=faq&entry=245442241039726

### DIFF
--- a/Casks/little-snitch.rb
+++ b/Casks/little-snitch.rb
@@ -10,9 +10,13 @@ cask 'little-snitch' do
   installer manual: 'Little Snitch Installer.app'
 
   zap delete: [
+                '/Library/Application Support/Objective Development/Little Snitch',
+                '~/Library/Application Support/Little Snitch',
+                '~/Library/Preferences/at.obdev.LittleSnitchConfiguration.plist',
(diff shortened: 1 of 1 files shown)


jasmas:

jasmas:
Uninstall for this cask needs some work. We basically need the equivalent of the 'installer manual:' stanza so that when 'brew uninstall little-snitch' completes the user is notified that '/Library/Little Snitch/Little Snitch Uninstaller.app' will need to be run manually and the computer restarted to complete the install.

Please advise the best way to accomplish this?

vitorgalvao:
Well, that’s the problem, we don’t have it.

[We’re aware this cask is missing `uninstall`](https://github.com/caskroom/homebrew-cask/issues/6872). We don’t allow any more casks without it, but this one was added before the rule and the app is popular.

Presumably whatever the cask is doing with its uninstallation can be replicated with CLI commands. Issue is we don’t know what the uninstall is really doing. Best way to go about it would be to contact the developers and ask how to perform the uninstallation, or better yet, if they would mind providing a script we could run (it has happened before).

I’ll still merge this PR now since you are still making a useful change, and the cask was already present.

PR #21879, opened 2016-06-12
USAGE.md: fix uses of /opt

vitorgalvao:
As [pointed out](https://github.com/caskroom/homebrew-cask/pull/21857#issuecomment-225419450).


claui:
LGTM :+1:

PR #21880, opened 2016-06-12
README.md: added detail about changing Caskroom

vitorgalvao:
As [pointed out](https://github.com/caskroom/homebrew-cask/pull/21857#issuecomment-225419450).


claui:
Looks good. :+1:

PR #21886, opened 2016-06-12
updated microsoft-azure-storage-explorer (0.8.0)

jazzwang:
#### Editing an existing cask

### DIFF
--- a/Casks/microsoft-azure-storage-explorer.rb
+++ b/Casks/microsoft-azure-storage-explorer.rb
@@ -1,9 +1,9 @@
 cask 'microsoft-azure-storage-explorer' do
-  version :latest
-  sha256 :no_check
+  version '0.8.0'
+  sha256 'aeeb6d463367f9d31a31553d676af0382c62f1befdbd02c2a651da0affcfba6b'
 
   # microsoft.com was verified as official when first introduced to the cask
-  url 'https://go.microsoft.com/fwlink/?LinkId=708342'
+  url 'http://go.microsoft.com/fwlink/?LinkId=708342'
   name 'Microsoft Azure Storage Explorer'
   homepage 'http://storageexplorer.com/'
   license :gratis


victorpopkov:
Thank you for your contribution. However, there is no point in this update.

Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check` as it was before.

jazzwang:
thanks @victorpopkov for your review :+1: 

the `url` is the same because the [upstream](http://storageexplorer.com/) make it so.
There is a new version on 06/08/2016 shown in <http://storageexplorer.com/releasenotes.html>
I have two simple questions:
- if I leave `version :latest`, how could user knew there are new version available? `brew cask upgrade` will not upgrade this, right?
- if I would like to upgrade, is the only way via `brew cask install microsoft-azure-storage-explorer`?

Casks/microsoft-azure-storage-explorer.rb
```diff
@@ -1,9 +1,9 @@
 cask 'microsoft-azure-storage-explorer' do
-  version :latest
-  sha256 :no_check
+  version '0.8.0'
+  sha256 'aeeb6d463367f9d31a31553d676af0382c62f1befdbd02c2a651da0affcfba6b'
```
jazzwang: `sha256` is generated by `cask-repair`

victorpopkov:
Thank you for mentioning. Since the `url` points to latest version we can't take advantage of string interpolation to modify it and produce the newer version URL. We simply have to use `:latest`. Unless another download URL appears that includes a pattern that could be followed to use as a version value.

> if I leave version :latest, how could user knew there are new version available? brew cask upgrade will not upgrade this, right?

The user won't know since there is no `brew cask upgrade` functionality yet. Homebrew-Cask doesn't support apps updating at the moment. This is the most desirable feature, but there are a lot of steps required in order to implement this. So it won't appear in near future, but it's planned, though.

> if I would like to upgrade, is the only way via brew cask install microsoft-azure-storage-explorer?

The user has to install the app forcefully again if he knows that there is a new version released. In this case: `brew cask install --force microsoft-azure-storage-explorer` most likely will do the trick.

> sha256 is generated by cask-repair

Yes. But when a new update will be available the cask will become broken since a newer version will have another checksum. This only can be applied to casks that have versioned download URLs.

jazzwang:
Great thanks to @victorpopkov for your detail explanation. 
I already got the point. I'm newbie to maintain this cask formula.
It is so impressive to use `cask-repair`. :100: 
Great thanks to `homebrew-cask` team to keep our life easy :+1: 

case closed!

Issue #21858, opened 2016-06-11
Deprecate `--caskroom` flag


vitorgalvao:
In a similar vein to https://github.com/caskroom/homebrew-cask/issues/21372, in [merging with Homebrew](https://github.com/caskroom/homebrew-cask/issues/14384) and [moving the Caskroom](https://github.com/caskroom/homebrew-cask/issues/21603) inside their directory by default, why would you ever want your Caskroom in a separate location?

We would simplify the code by removing the flag. Even when creating the Caskroom for the first time, [this](https://github.com/caskroom/homebrew-cask/blob/d4264c89af61decae13ba2e8d177ad7ee32344c1/lib/hbc.rb#L52L78) (26 lines with multiple `sudo`s) could become this (9 lines, no `sudo`s):

``` ruby
def self.init
  odebug 'Creating directories'
  HOMEBREW_CACHE.mkpath unless HOMEBREW_CACHE.exist?
  HOMEBREW_CACHE_CASKS.mkpath unless HOMEBREW_CACHE_CASKS.exist?
  unless caskroom.exist?
    ohai "Creating Caskroom at #{caskroom}"
    system '/bin/mkdir', '--', caskroom
  end
end
```

claui:
@vitorgalvao I am very much in favor of this – as long as the [particular minority I’m lobbying for](https://github.com/caskroom/homebrew-cask/issues/21603#issuecomment-223072877) is not affected; and it indeed is not. (I would make for a terrible politician!)

Removing things no one really needs anymore is _exactly_ what keeps a product from becoming a big ball of mud. :+1:

jawshooah:
In light of the feedback we've been getting after #21857 (e.g. #21894), I think we should hold off on this for a while. I feel like we jumped the gun, and should have made sure to notify users of the change and ensure that their old Caskroom would remain intact, at least for a specified transitional period.

I'm away from my computer at the moment, but unless someone beats me to it (@claui has been killing it lately), I'm going to write up a PR this evening that will use the old path if it exists, warn users that the default has changed, and encourage them to add `--caskroom=/opt/homebrew-cask/Caskroom` to `HOMEBREW_CASK_OPTS` to prevent data duplication and installation issues.

vitorgalvao:
@jawshooah Made [an official answer](https://github.com/caskroom/homebrew-cask/issues/21894#issuecomment-225464059) before we start getting bashed by the less friendly users.

[...]

PR #21893, opened 2016-06-12
Handle broken symlinks on `install`, `install --force`

claui:
While testing PR #21857 earlier today, I have found that `brew cask install` fails when it encounters a broken symlink, no matter if with or without `--force`:

image

Users _might_ run into this in the wake of PR #21857, e.&nbsp;g. trying to move their Caskrooms around in an attempt to get `brew cask list` back working, or to merge their Caskrooms. One could argue that users are not supposed to do this, but we all know they will do it anyway. The issue _will_ also bite unsuspecting users once PR #21858 is merged. Any legacy (pre-#13966) symlink pointing inside a custom Caskroom location is likely going to break.

I feel that while `install` needs to fail gracefully, `install --force` should do [exactly what it says on the tin](https://github.com/caskroom/homebrew-cask/pull/13966#issuecomment-220830387) and sweep broken symlinks out of the way.

**This PR fixes** both the `install` and `install --force` case.

Note that the `uninstall` and `zap` cases have similar issues; those remain unfixed for the time being. Those cases have additional issues anyway, require a bit more effort to be fixed, and thus deserve a PR on their own.

### DIFF
--- a/lib/hbc/artifact/moved.rb
+++ b/lib/hbc/artifact/moved.rb
@@ -22,7 +22,7 @@ class Hbc::Artifact::Moved < Hbc::Artifact::Base
     each_artifact do |artifact|
       load_specification(artifact)
       next unless preflight_checks
-      delete if File.exist?(target) && force
+      delete if Hbc::Utils.path_occupied?(target) && force
       move
     end
   end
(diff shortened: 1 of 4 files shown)


jawshooah:
Great work once again, @claui 😄

claui:
Thanks @jawshooah 😊

Issue #21894, opened 2016-06-12
Rhetorical: Why did caskroom just re-install (rather than migrate) every app in Brewfile?


remy-actual:
Yes, I know __why__. But since a main topic of discussion in "the relevant [issue](https://github.com/caskroom/homebrew-cask/issues/21603) and [pull request](https://github.com/caskroom/homebrew-cask/pull/21857)" is the Caskroom UX, I thought I'd share my User Experience this morning.  

Steps to replicate: 
* Run Brewfile
* Encounter questions

1. Why is cask reporting Adium not currently installed? 
1. We're re-installing Adium now? Okay...
1. Now cask is reporting BitBar is not currently installed too? Um... I'm _looking at it_ running in the Menu Bar right now.
1. Did I mess something up? Is Cask misconfigured? 
1. Are we re-installing everything? A - Z ?
1. 😑  We **are** reinstalling everything. Greeeeeaaat. How long is this going to take?
1. "It seems there is already an App at '/Applications/iTerm.app'; not moving." No kidding! Where do you think I'm reading this? 
1. So... does this mean I have two copies of iTerm on my system now? Two Google Chromes? 
1. How much am I going to have to clean up? Am I going to have a bunch of broken links? Working links pointing to multiple installs? This is turning into a real PITA. 
1. How much of this is running `brew cask cleanup` going to fix?
1. 😱  Are my app preferences being preserved? Or are they getting wiped out? Because _that would **suck**_. Guess we'll wait and see! 
1. Well my custom key maps aren't functioning now. Time to go re-setup Karabiner! 😡 
1. Ugh. MacTeX. Seriously? All I wanted was to install colordiff and brew upgrade. I'm gonna go take a shower... 
1. Okay. 30 minutes and 6.6 gigs later and we're done! Glad I have a decent processor and unlimited reliable bandwidth. Feel bad today for anybody who doesn't 😕 ...
1. Ugh. What is this system pop-up? I don't even know what app is sending this... 
![whoareyou](https://cloud.githubusercontent.com/assets/952821/15992860/af1f77b6-308b-11e6-81e0-e468016655b4.png)
1. Is this the thing they were talking about in the README about changing Caskroom behavior? They wouldn't push out a major change like that without including **some** notification on the command line, right? I mean, Cask _is_ a CLI workflow -- _why wouldn't you advise your users of changes in the CLI?_
[...]

PR #21901, opened 2016-06-13
Use legacy Caskroom if it exists, but warn of change

jawshooah:
If a Caskroom exists at the old location of /opt/homebrew-cask/Caskroom,
use it and warn the user that the default location has changed.

Refs #21858, #21857, #21894

### DIFF
--- a/lib/hbc/locations.rb
+++ b/lib/hbc/locations.rb
@@ -4,8 +4,30 @@ module Hbc::Locations
   end
 
   module ClassMethods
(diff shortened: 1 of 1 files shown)


jawshooah:
@caskroom/maintainers: I went ahead with the merge given the urgency of the underlying issue, but please do take a look and make suggestions as necessary.

vitorgalvao:
Thank you, @jawshooah.

PR #21906, opened 2016-06-13
AppCleaner: Add backwards-compatible versions.

### LINKED ISSUE #21755: Suggestion: Add Mac OS X version check to AppCleaner
As you may know, [AppCleaner](https://freemacsoft.net/appcleaner/) has 3 official releases, 1.2.2, 2.3 and 3.3 (up to now).

Ver 1 supports 10.4 to 10.5
Ver 2 supports 10.6 to 10.10
Ver 3 supports 10.10 and so on

Therefore, if you can add OS check to this app to specify which version brew should install, it would be wonderful, since they are not backward compatible.

reitermarkus:

Closes https://github.com/caskroom/homebrew-cask/issues/21755.

### DIFF
--- a/Casks/appcleaner.rb
+++ b/Casks/appcleaner.rb
@@ -1,13 +1,26 @@
 cask 'appcleaner' do
-  version '3.3'
-  sha256 'a856e3e08af2ae972582c3a75403086413295d1a38d86a3e2d5f6abff7505fdd'
+  if MacOS.release <= :leopard
+    version '1.2.2'
+    sha256 '58972cb1b8256d6de1a11ed225c94b240c2923f1c814975ccbcb94806fc439dc'
+    url "https://freemacsoft.net/downloads/AppCleaner#{version}.dmg"
+  else
+    if MacOS.release <= :mavericks
+      version '2.3'
+      sha256 '69da212e2972e23e361c93049e4b4505d7f226aff8652192125f078be7eecf7f'
+    else
+      version '3.3'
+      sha256 'a856e3e08af2ae972582c3a75403086413295d1a38d86a3e2d5f6abff7505fdd'
+      appcast 'https://freemacsoft.net/appcleaner/Updates.xml',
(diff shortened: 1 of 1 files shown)


Casks/appcleaner.rb
```diff
@@ -1,14 +1,27 @@
...
+      sha256 '69da212e2972e23e361c93049e4b4505d7f226aff8652192125f078be7eecf7f'
+    else
+      version '3.3'
+      sha256 'a856e3e08af2ae972582c3a75403086413295d1a38d86a3e2d5f6abff7505fdd'
+    end
+    url "https://www.freemacsoft.net/downloads/AppCleaner_#{version}.zip"
+  end
 
-  url "https://www.freemacsoft.net/downloads/AppCleaner_#{version}.zip"
   appcast 'https://freemacsoft.net/appcleaner/Updates.xml',
```
vitorgalvao: `appcast` should be inside the latest `else` (i.e. the one for version `3.3`), since that is the only one that will be updated.

Casks/appcleaner.rb
```diff
@@ -1,14 +1,27 @@
 cask 'appcleaner' do
-  version '3.3'
-  sha256 'a856e3e08af2ae972582c3a75403086413295d1a38d86a3e2d5f6abff7505fdd'
+  if MacOS.release <= :leopard
+    version '1.2.2'
+    sha256 '58972cb1b8256d6de1a11ed225c94b240c2923f1c814975ccbcb94806fc439dc'
+    url "https://freemacsoft.net/downloads/AppCleaner#{version}.dmg"
+  else
```
Amorymeltzer: Is there a reason why this is a nested `if` rather than `elsif MacOS.release <= :mavericks`?

vitorgalvao: It’s so the later `url "https://www.freemacsoft.net/downloads/AppCleaner_#{version}.zip"` doesn’t need to be repeated. A bit unorthodox, but it works.

reitermarkus: Yeah, I don't really see a *“better”* way for this. =)

PR #21909, opened 2016-06-13
Revert "updated linear (1.4.0)"

miccal:
Reverts caskroom/homebrew-cask#21908


vitorgalvao:
Why revert? What’s the problem?

miccal:
There is something wrong with the new version and the app will not run.

vitorgalvao:
I can reproduce. [Reported the crash](https://github.com/mikaa123/linear/issues/31).

Merging in the meantime.

PR #21928, opened 2016-06-13
OS X → macOS and added support for Sierra

vitorgalvao:
Some instances of `OS X` were kept. These were the cases where we fake a user agent and where we are specifically talking about old versions of the OS that will never include Sierra or the future ones.


claui:
Looks like you caught them all! :+1:

vitorgalvao:
Also changed uses of `Mac` to `macOS`.

PR #21952, opened 2016-06-14
Audacity update url

tobadia:
#### Editing an existing cask

Changed Audacity download URL. 
Currently, the official website points to download URL leading to an HTML page, although seemingly pointing to a file. This HTML page is now parsed using open-uri to extract the temporary (protected) download URL that is embedded within an iframe. 
The previous download URL that was hardcoded (http://app.oldfoss.com:81/download/Audacity/audacity_macosx_ub_2_1_2-1453294898.dmg) seems not to be working anymore, so I'm proposing this fix.

### DIFF
--- a/Casks/audacity.rb
+++ b/Casks/audacity.rb
@@ -1,9 +1,17 @@
+require 'open-uri'
 cask 'audacity' do
   version '2.1.2'
   sha256 '2e4b7d608ecc0d2f79bf16663f085d383075e488f7d50bf7d74c0b69173defe7'
 
   # Current official URL as proposed on http://www.audacityteam.org/download/mac/
-  url "http://www.fosshub.com/Audacity.html/audacity-macosx-ub-#{version}.dmg"
+  # must be parsed to extract temporary url embedded in iframe
+  iframe_src_protected_url = nil
+  open("http://www.fosshub.com/Audacity.html/audacity-macosx-ub-#{version}.dmg") { |io|
+    content = io.read
+    iframe_src_protected_url = /^\<iframe.*src=\"(http.*\.dmg)\".*>/.match(content)[1].to_s
+  }
+  url iframe_src_protected_url
+  
   name 'Audacity'
   homepage 'http://audacityteam.org'
   license :gpl
--- a/Casks/audacity.rb
+++ b/Casks/audacity.rb
@@ -1,4 +1,8 @@
+# Audacity does not provide a fixed URL
+# Their download URL points to a html page that generates a temporary URL embedded within an iframe
+# 'open-uri' is required to open that page and grab the temporary URL
 require 'open-uri'
+
 cask 'audacity' do
   version '2.1.2'
   sha256 '2e4b7d608ecc0d2f79bf16663f085d383075e488f7d50bf7d74c0b69173defe7'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
This fix can’t be accepted, because this will run every time the cask is loaded, which is not only when trying to install it specifically. This means that URL will try to be opened many times, and can hang and make every operation lower.

Members of the Audacity team know fosshub cannot be used with `curl` and that as such we cannot use their download URLs, and must resort to oldfoss. They don’t like it, but also have no solution. We either get a working URL, or we remove the cask.

vitorgalvao:
[Visited the download page](http://www.oldfoss.com/Audacity.html), and the URL is still accurate, it just appears to be down at the moment.

tobadia:
Thanks for the feedback. Indeed it is still accurate, but has been down for at least 36 hours now. 
But you're right, hacking through their redirections in embedded iframes is not a clean way to get a valid URL. That was just an attempted contribution but I understand the downside of this.

vitorgalvao:
Believe me, [I do like your solution](https://github.com/caskroom/homebrew-versions/pull/293#issuecomment-46122949), it just [introduces bad things](https://github.com/caskroom/homebrew-versions/pull/293#issuecomment-48755213).

shakeelmohamed:
Just ran into this issue, went back a version and the download started instantly (`http://www.oldfoss.com/Audacity/download/audacity-macosx-ub-2.1.1.dmg`).

Can we downgrade the version used by cask to `2.1.1`?

vitorgalvao:
@shakeelmohamed Not until someone at least contacts the developers upstream to let them know, to see if they know about the issue.

claui:
@vitorgalvao @shakeelmohamed @tobadia I’m just about to finish an enhancement to the core which allows the URL stanza to support lazy evaluation:

```rb
url do
  base_url = 'http://www.fosshub.com/Audacity.html'
  open("#{base_url}/audacity-macosx-ub-#{version}.dmg") do |io|
    content = io.read
    %r{^\<iframe.*src=\"(http.*\.dmg)\".*>}.match(content)[1].to_s
  end
end
```

As @vitorgalvao pointed out earlier, we have had this issue with other casks in the past. I figure this justifies a DSL enhancement to make us less dependent to upstream doohickeys

As the PR is imminent, I would suggest that this issue be closed. Also, no need to contact upstream.

tobadia:
@shakeelmohamed probably not a good idea since this would not keep up with the current release. 

@claui glad to have contributed to other enhancements, just wondering : how will be handled the fact that open() will be called at each load, as @vitorgalvao mentioned earlier? (naive question, really)

vitorgalvao:
> how will be handled the fact that open() will be called at each load

From @claui’s post:

> an enhancement to the core which allows the URL stanza to support lazy evaluation

“Lazy evaluation” basically means “don’t evaluate this until it is needed”. That should mean that with @claui’s improvement, `url` will no longer be called at each load, just when there are operations that require it.

claui:
> `url` will no longer be called at each load, just when there are operations that require it.

Correct. Writing `url` with the new block syntax means: it will only be called when some code wants the URL, and even then only once.

Anything _outside_ that `url` block is still bound to run at each load.
The fix requires the `open` block to be _inside_ the `url` block, as [the example](https://github.com/caskroom/homebrew-cask/pull/21952#issuecomment-226139471) shows.

Writing `url <url_string>` **will remain the recommended way** to specify an URL for almost every Cask.

tobadia:
Got it, thank you both for detailing this.

claui:
@tobadia FYI – the fix has just been merged, including your contribution.

tobadia:
I was following the threads on the other edits you were both documenting. Thanks for the beads up!

tobadia:
Heads* damn autocorrect

PR #21996, opened 2016-06-15
Fix incorrect google code archive urls

jeanregisser:
Hi,

I found an incorrect google code archive url when trying to install cronnix, leading to a 404.
I went through all casks and also found epub-to-pdf which was affected.

According to https://code.google.com/archive/schema, the v2 url should always be used.


victorpopkov:
Thank you for noticing those. 👍

PR #22001, opened 2016-06-15
Removed sqwiggle

jeanregisser:
Hi,

I've removed sqwiggle since they shut their doors recently and the binary is not available anymore.

See https://www.sqwiggle.com/


vitorgalvao:
Thank you.

PR #22002, opened 2016-06-15
[WIP] Fixed Recognition of Compression Types in macOS Sierra

jellyfishcoder:
### Changes to Detecting Package Compression Method, Fixes Errors in macOS Sierra

In macOS Sierra, recognising the packages compression with `criteria.file.include?`, yields the wonderful message `Error: Uh oh, could not figure out how to unpack '/Users/.../Library/Caches/Homebrew/filezilla-3.18.0.tar.bz2'`. This fix adds the ability to also recognise them by file extensions, which unless the packages extension does not match its type (unlikely), works all the time. The original detection method is still present, `or` statements are used between each detection method.

### DIFF
--- a/lib/hbc/container/bzip2.rb
+++ b/lib/hbc/container/bzip2.rb
@@ -2,7 +2,7 @@ require 'tmpdir'
 
 class Hbc::Container::Bzip2 < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.file.include? 'compressed-encoding=application/x-bzip2;'
+    criteria.file.include?('application/x-bzip2;')
   end
 
   def extract
--- a/lib/hbc/container/gzip.rb
+++ b/lib/hbc/container/gzip.rb
@@ -1,11 +1,8 @@
-
-# for pure gzip only, not tar-gzip (.tgz or .tar.gz)
-
 require 'tmpdir'
 
 class Hbc::Container::Gzip < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.file.include? 'compressed-encoding=application/x-gzip;'
+    criteria.file.include?('application/x-gzip;')
   end
 
   def extract
--- a/lib/hbc/container/tar.rb
+++ b/lib/hbc/container/tar.rb
@@ -2,7 +2,8 @@ require 'tmpdir'
 
 class Hbc::Container::Tar < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.file.include? 'application/x-tar'
+    criteria.file.include?('application/x-tar') ||
+      IO.popen(['/usr/bin/tar', '-tf', criteria.path.to_s], err: open('/dev/null').fileno) { |io| !io.read(1).nil? }
   end
 
   def extract
--- a/lib/hbc/container/zip.rb
+++ b/lib/hbc/container/zip.rb
@@ -1,6 +1,6 @@
 class Hbc::Container::Zip < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.file.include? 'compressed-encoding=application/zip;'
+    criteria.file.include?('application/zip;')
   end
 
   def extract
--- a/lib/hbc/container.rb
+++ b/lib/hbc/container.rb
@@ -14,6 +14,8 @@ require 'hbc/container/pkg'
 require 'hbc/container/seven_zip'
 require 'hbc/container/sit'
 require 'hbc/container/tar'
+require 'hbc/container/tar_bzip2'
+require 'hbc/container/tar_gzip'
 require 'hbc/container/ttf'
 require 'hbc/container/rar'
 require 'hbc/container/xar'
@@ -30,12 +32,14 @@ class Hbc::Container
       Hbc::Container::Dmg,
       Hbc::Container::SevenZip,
       Hbc::Container::Sit,
-      Hbc::Container::Tar,     # or compressed tar
       Hbc::Container::Rar,
       Hbc::Container::Zip,
+      Hbc::Container::TarBzip2,
+      Hbc::Container::TarGzip,
       Hbc::Container::Bzip2,
       Hbc::Container::Gzip,    # pure gzip, not tar/gzip
       Hbc::Container::Xar,
+      Hbc::Container::Tar,     # or compressed tar
     ]
     # for explicit use only (never autodetected):
     # Hbc::Container::Naked
--- a/lib/hbc/container/criteria.rb
+++ b/lib/hbc/container/criteria.rb
@@ -43,10 +43,11 @@ class Hbc::Container::Criteria
     path.extname.sub(%r{\A\.}, '').downcase == test.downcase
   end
 
-  def magic_number(num, test)
-    File.open(path, "rb") do |file|
-      bytes = file.read(num).unpack('C*')
-      bytes == test
-    end
+  def magic_number(position, length, test, stream = nil)
+    stream = File.open(path, "rb") if stream.nil?
+    stream.read(position)
+    bytes = stream.read(length)
+    stream.close
+    bytes == test
   end
 end
--- a/lib/hbc/container/pkg.rb
+++ b/lib/hbc/container/pkg.rb
@@ -1,7 +1,8 @@
 class Hbc::Container::Pkg < Hbc::Container::Naked
   def self.me?(criteria)
-    (criteria.extension 'pkg' or criteria.extension 'mpkg') and
-      (criteria.file.include? 'application/x-directory' or
-       criteria.magic_number(4, 'xar!'.unpack('C*')))
+    (criteria.extension('pkg') ||
+     criteria.extension('mpkg')) &&
+      (criteria.file.include?('application/x-directory') ||
+       criteria.magic_number(0, 4, 'xar!'))
   end
 end
--- a/lib/hbc/container/seven_zip.rb
+++ b/lib/hbc/container/seven_zip.rb
@@ -1,11 +1,11 @@
 class Hbc::Container::SevenZip < Hbc::Container::GenericUnar
   def self.me?(criteria)
     # todo: cover self-extracting archives
-    criteria.extension '7z' and
-      criteria.file.include? 'application/octet-stream;' and
-        criteria.magic_number(2, '7z'.unpack('C*')) and
-          ! criteria.lsar.nil? and
-            criteria.lsar.split("\n").first.split(':').last.include?('7-Zip') and
+    criteria.extension('7z') &&
+      criteria.file.include?('application/octet-stream;') &&
+        criteria.magic_number(0, 2, '7z') &&
+          !criteria.lsar.nil? &&
+            criteria.lsar.split("\n").first.split(':').last.include?('7-Zip') &&
               super
   end
 end
--- a/lib/hbc/container/tar_bzip2.rb
+++ b/lib/hbc/container/tar_bzip2.rb
@@ -1,0 +1,9 @@
+require 'hbc/container/tar'
+require 'hbc/container/bzip2'
+
+class Hbc::Container::TarBzip2 < Hbc::Container::Tar
+  def self.me?(criteria)
+    Hbc::Container::Bzip2.me?(criteria) &&
+      criteria.magic_number(257, 6, "ustar\0", IO.popen(['/usr/bin/bunzip2', '-c', criteria.path.to_s, :err => :out]))
+  end
+end
--- a/lib/hbc/container/tar_gzip.rb
+++ b/lib/hbc/container/tar_gzip.rb
@@ -1,0 +1,10 @@
+require 'hbc/container/tar'
+require 'hbc/container/gzip'
+require 'zlib'
+
+class Hbc::Container::TarGzip < Hbc::Container::Tar
+  def self.me?(criteria)
+    Hbc::Container::Gzip.me?(criteria) &&
+      criteria.magic_number(257, 6, "ustar\0", Zlib::GzipReader.open(criteria.path))
+  end
+end
--- a/lib/hbc/container/ttf.rb
+++ b/lib/hbc/container/ttf.rb
@@ -1,10 +1,9 @@
 class Hbc::Container::Ttf < Hbc::Container::Naked
   def self.me?(criteria)
-    (criteria.extension 'ttf' and
-     (criteria.file.include? 'application/x-font-ttf' or
-      criteria.magic_number(4, 'true'.unpack('C*')))) or
-
-    (criteria.extension 'ttc' and
-      criteria.magic_number(4, 'ttcf'.unpack('C*')))
+    (criteria.extension('ttf') &&
+     (criteria.file.include?('application/x-font-ttf') ||
+      criteria.magic_number(0, 4, 'true'))) ||
+      (criteria.extension('ttc') &&
+       criteria.magic_number(0, 4, 'ttcf'))
   end
 end
--- a/lib/hbc/container/xar.rb
+++ b/lib/hbc/container/xar.rb
@@ -2,7 +2,7 @@ require 'tmpdir'
 
 class Hbc::Container::Xar < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.magic_number(4, 'xar!'.unpack('C*'))
+    criteria.magic_number(0, 4, 'xar!')
   end
 
   def extract


vitorgalvao:
> which unless the packages extension does not match its type (unlikely)

Not at all unlikely. Not an everyday occurrence, but I’ve seen it happen multiple times.

lib/hbc/container/bzip2.rb
```diff
@@ -2,6 +2,7 @@
 
 class Hbc::Container::Bzip2 < Hbc::Container::Base
   def self.me?(criteria)
+    criteria.extension 'tbz' or
     criteria.file.include? 'compressed-encoding=application/x-bzip2;'
```
claui: The continued statement on line #6 needs indentation.

lib/hbc/container/tar.rb
```diff
@@ -2,6 +2,10 @@
 
 class Hbc::Container::Tar < Hbc::Container::Base
   def self.me?(criteria)
+    criteria.extension 'tar' or
+    criteria.extension 'tar.bz2' or
+    criteria.extension 'tar.gz' or
+    criteria.extension 'tgz' or
     criteria.file.include? 'application/x-tar'
```
claui: Here, lines 6–9 need to be indented.

lib/hbc/container/zip.rb
```diff
@@ -1,6 +1,7 @@
 class Hbc::Container::Zip < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.file.include? 'compressed-encoding=application/zip;'
+   criteria.extension 'zip' or
+   criteria.file.include? 'compressed-encoding=application/zip;'
```
claui: The third line needs one more space for proper indentation; the fourth line needs three more spaces ☺️

claui:
@jellyfishcoder Thanks for fixing this! Yesterday I tried all day to make Sierra run in a VM but finally gave up. Now I’m relieved to see the issue addressed :)

Looks good so far – could you please clean up the code a bit before we proceed with merging? It boils down to two minor issues:

1. It would help with readability if your code said `||` instead of `or`. ([Rationale](https://github.com/bbatsov/ruby-style-guide#no-and-or-or))
We do have a lot of `or` and `and` left in the existing code base but those are becoming fewer and fewer.

2. For maintainability, it is crucial that code is properly indented. See line notes for details.

Thank you in advance!

jawshooah:
Here's some more [info](http://devblog.avdi.org/2010/08/02/using-and-and-or-in-ruby/) on the difference between `and/or` and `&&/||`. `and` and `or` are useful for control flow, but too confusing in practice for those not in the know.

jellyfishcoder:
@claui @jawshooah If I use `||` instead of `or`, brew cask does not work. It says something like `syntax error, unexpected tSTRING_BEG, expecting keyword_end
    Criteria.extension 'tar.bz2' ||` with an arrow at the `'tar.bz2' ||` part

jawshooah:
Binding strength, my man. `||` binds more strongly than `or`, and more strongly than a method application with no parentheses. Gotta use parentheses.

> All the other container ruby files use `or` instead of `||`

@claui addressed that above. It's not the only thing about the existing code that needs cleaning up, but we haven't gotten around to it yet.

jellyfishcoder:
@jawshooah @claui I tried the "" instead of '' fix, it worked on all but bzip2, which says Users/...../Documents/homebrew-cask/lib/hbc/container/bzip2.rb:6: syntax error, unexpected tSTRING_BEG, expecting keyword_end
    criteria.file.include? "compressed-encoding=application/x-bzip2;"

claui:
@jellyfishcoder What happens if you try the following variant?

```rb
  def self.me?(criteria)
    criteria.extension('tbz') ||
      criteria.extension('bz2') ||
      criteria.file.include?('compressed-encoding=application/x-bzip2;')
  end
```

Does that work for you?

jellyfishcoder:
Thanks a lot @claui, this works perfectly!

jellyfishcoder:
@claui At least its not just me cause Im clueless as to why the "compressed-encoding=" does not work. Do you think I should report a bug about it with Apple?

lachlanhunt:
I experienced this same issue, so I had a look to see if I could understand why it broke. It looks like hbc/container/criteria.rb runs `/usr/bin/file -Izb -- path-to-file` and returns the resulting string.

It seems that the output from /usr/bin/file has changed with the new version in macOS Sierra.  `file --version` says `file-5.25` on Sierra and `file-5.04` El Capitan.

On OS X El Capitan:

    $ file -Izb -- 0xed-latest.tar.bz2
    application/x-tar; charset=binary compressed-encoding=application/x-bzip2; charset=binary; charset=binary

On macOS Sierra:

    $ file -Izb -- 0xed-latest.tar.bz2
    application/x-bzip2; charset=binary

Then when self.me function runs `criteria.file.include?('compressed-encoding=application/x-bzip2;')`, the string doesn't match.

So, the patch could simply be to check for the inclusion of the MIME type itself, and not also look for 'compressed-encoding='.

vitorgalvao:
I like @lachlanhunt’s solution. Seems more “correct” than to check file extensions.

reitermarkus:
I also ran into this. Have fixed it temporarely using a regex as I wasn't sure if simply checking for `application/zip` is sufficient.

```ruby
def self.me?(criteria)
  criteria.file =~ %r{(compressed-encoding=application/zip;|application/zip; charset=binary)}
end
```

jellyfishcoder:
@vitorgalvao I will try to do change it to that tomorrow, but today I may not have time.

lib/hbc/container/bzip2.rb
```diff
@@ -2,7 +2,7 @@
 
 class Hbc::Container::Bzip2 < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.file.include? 'compressed-encoding=application/x-bzip2;'
+    criteria.file.include?('application/x-bzip2;')
```
reitermarkus: Maybe leave off the parentheses for consistency.

jellyfishcoder:
I will flatten the commits once I get home to a computer and can actually test it.

lib/hbc/container/bzip2.rb
jawshooah: I'd much prefer that the parentheses remain, and the same for the other files. The convention that I follow, and that seems to be common among Rubyists, is to omit parentheses only when you don't care about the return value (e.g. when calling a method purely for its side effects, like `puts`).

reitermarkus: Ok, I will keep that in mind. Either way, it should be consistent.

lib/hbc/container/tar.rb
```diff
@@ -2,7 +2,7 @@
 
 class Hbc::Container::Tar < Hbc::Container::Base
   def self.me?(criteria)
-    criteria.file.include? 'application/x-tar'
+    criteria.file.include?('application/x-tar')
```
lachlanhunt: This might cause a problem for users on El Capitan because the output from `file` for bzip2, gzip, zip and tar files is:

    application/x-tar; charset=binary compressed-encoding=application/x-bzip2; charset=binary; charset=binary
    application/x-tar; charset=binary compressed-encoding=application/x-gzip; charset=binary; charset=binary
    application/x-empty compressed-encoding=application/zip; charset=binary; charset=binary
    application/x-tar; charset=binary

Note that the output for bzip2 and gzip both start with `application/x-tar`.  So, that could mistakenly misidentify bzip2 and gzip as tar files, unless it's guaranteed that this check for a tar file comes after it's already checked for bzip and gzip, which doesn't look to be the case in container.rb.

For comparrison, on Sierra, the output is:

    application/x-bzip2; charset=binary
    application/x-gzip; charset=binary
    application/zip; charset=binary
    application/x-tar; charset=binary

reitermarkus: Well, the only difference here are the parentheses, so this is actually the current code, which *is* working on El Capitan.

jawshooah:
Should be trivial to enforce the order in which the different container types are checked. Not at a computer right now so I can't point to where exactly that order is defined right now

lib/hbc/container/tar.rb
claui: `Hbc::Container::Bzip2` indeed appears to be dead code on El Capitan. The current code works by accident because the `tar` implementation built in to macOS happens to auto-detect bzip2 compression. From the man page:

> Note that, unlike other tar implementations, this implementation recognizes bzip2 compression automatically when reading archives.

In case users have GNU tar installed at `/usr/bin/tar`, it [has done the same auto-detection](https://www.gnu.org/software/tar/manual/tar.html#gzip) for more than a decade.

Still, I think it is a good idea to have predictable logic across macOS versions. We should either rearrange and enforce the order so `Hbc::Container::Bzip2` comes before `Tar`, or alternatively get rid of `Hbc::Container::Bzip2`.

jellyfishcoder:
I fixed the order of checking, so tar is checked last, to prevent any errors. @jawshooah @lachlanhunt

claui:
@jellyfishcoder Glad you did the reordering! It revealed another issue with `Hbc::Container::Gzip` which would have occurred only in Sierra.

The thing is, `Hbc::Container::Gzip` [claims to handle](https://github.com/jellyfishcoder/homebrew-cask/blob/e826ee0d0a4889c4970b39511c6d28143836b751/lib/hbc/container/gzip.rb#L8) `application/x-gzip` but it [actually cannot deliver](https://github.com/jellyfishcoder/homebrew-cask/blob/e826ee0d0a4889c4970b39511c6d28143836b751/lib/hbc/container/gzip.rb#L2); it relies on `/usr/bin/gunzip` which stops half-way, leaving a `.tar` file behind.

A small test case (with output on El Capitan) to illustrate the problem:

```sh
cd "$(mktemp -d)"
cp "$(brew --prefix)/Library/Taps/caskroom/homebrew-cask/test/support/binaries/tarball.tgz" .
/usr/bin/file -Izb tarball.tgz
# => application/x-tar; charset=binary compressed-encoding=application/x-gzip; charset=binary; charset=binary

/usr/bin/gunzip -v tarball.tgz
# => tarball.tgz:	   98.1% -- replaced with tarball.tar

ls
# => tarball.tar
```

**tl;dr** Gzipped tarballs are `application/x-gzip`. `Hbc::Container::Gzip` claims to handle those, `gunzip` actually stops half-way, leaving a `.tar` file behind.

@lachlanhunt What does `file` say on Sierra?

jellyfishcoder:
Hm, I can't figure out why it broke, It worked for me.

lachlanhunt:
I saw the same problem with the bzip2 code, which extracted a tar file and then tried to copy the .app and failed. It worked if I replaced the extraction function with the same code from the tar handler that calls:

     tar -xf file.tar.bz2 -C /some/temp/dir

That will work if we can safely assume that all bzip2 and gzip files downloaded by brew cask will really be compressed tar files. Otherwise, the bzip and gzip code will need to be modified to first decompress, and then check if the result is a tar or something else before unpacking.

@claui for a .tgz file, `file -Izb test.tgz` says the same as a .tar.gz file, `application/x-gzip; charset=binary`.

reitermarkus:
Well, why not simply keep using `tar` for everything as it supports `bzip2`, `gzip`, `lzma` and `xz`?

claui:
@reitermarkus That would fail when there is no tarball involved; example:

```sh
cd "$(mktemp -d)" && touch foo && gzip foo
tar xfv foo.gz
ls
# => foo.gz
```

claui:
Not that it helps much, but I just tried out Sierra and found that `man file` includes a TODO list with the following entry:

> TODO
> […]
> Make `file -ki` work, i.e. give multiple MIME types.

That same note is also included in the [online man page](https://www.freebsd.org/cgi/man.cgi?query=file) for the current version 5.25 of `file`, which according to @lachlanhunt is the version that comes with Sierra.

Looks like `file` has had this particular bug since [almost five years](https://www.freebsd.org/cgi/man.cgi?query=file&apropos=0&sektion=0&manpath=FreeBSD+10.0-RELEASE+and+Ports&arch=default&format=html) (version 5.11) and that the authors seem to be well aware of it.

I highly doubt that the FreeBSD team is going to fix this anytime soon. Even if they did, Apple might not include the fix in Sierra when it ships.

To me, it feels like we might **need another method to figure out whether an archive contains a tarball or not.**

lachlanhunt:
@claui I suspect that Apple have forked the file command because it differs from file-5.25 in FreeBSD, and there are some differences between the man pages. The man page in Sierra says "This manual page documents version 5.04 of the file command", even though the actual binary reports its own version as file-5.25.

In the Apple fork, they have `-I` (uppercase) meaning the same thing as `-i` (lowercase) in the FreeBSD version, because `-i` (lowercase) means something else in the Apple fork.

I also went and got a virtual machine of FreeBSD to see how file 5.25 behaves there, and `file -izb test.tar.gz` behaves the same way `file -Izb test.tar.gz` behaved in El Capitan.

I also noted that `file` in FreeBSD and El Capitan do actually distinguish between foo.gz and foo.tar.gz.

foo.gz (not tar):
    application/x-empty compressed-encoding=application/x-gzip; charset=binary; charset=binary

foo.tar.gz:
    application/x-tar; charset=binary compressed-encoding=application/x-gzip; charset=binary; charset=binary

That allows us to distinguish between gzip (and bzip2) files that are and are not tar, and means that we should actually keep the check for tar before bzip and gzip.

So, I believe this may actually be a bug that Apple should fix after all.

lachlanhunt:
After a little further investigation, it's specifically the -z parameter handling that is broken and Apple needs to fix.

Notably, `file -z foo.tar.gz` and `file foo.tar.gz` output exactly the same result in Sierra, whereas they differed usefully in El Capitan and in FreeBSD.

lachlanhunt:
In case no-one else has already, I have submitted feedback to Apple about this issue.

claui:
Thanks @lachlanhunt for your very thorough investigation, including trying an actual FreeBSD installation. Great findings everybody!

I just filed a [Radar bug](http://www.openradar.me/radar?id=5621832178728960), too. Given how Apple’s internal prioritization allegedly works, multiple reports cannot hurt at all.

(By the way, I wonder whether Apple is aware of Homebrew-Cask, and whether they consider it a partial competitor to the Mac App Store? Anyway, I was cautious not to mention HBC – just in case, lest they deprioritize the issue for strategic reasons.)

reitermarkus:
Ok, also filed a bug report.

And for the moment, I found another way to check if the archive contains/is a tarball:

```ruby
class Hbc::Container::Criteria
  def tarball?
    return @tarball unless @tarball.nil?
    result = @command.run('/usr/bin/tar', :args => ['-tf', path])
    @tarball = result.success? && result.stdout.length > 0 && result.stderr.length == 0
  end
end
```

claui:
@reitermarkus Your proposal seems like a sensible and pragmatic workaround.

I ran a test with an uncompressed tarball containing a 1 GB file. The `tar -tf` check takes about 2 seconds while `file` takes less than 20 ms. A similar test with a tarball containing 1000 files, each of 1 MB length, yields the same result.

This is going to increase download time by less than half a second on average, which I think is tolerable. As long as we don’t find a faster solution (e. g. code that directly uses the magic numbers provided by `/usr/share/file/magic/compress` and `/usr/share/file/magic/archive`), **I’m pretty much in favor of the `tar -tf` proposal.**

vitorgalvao:
> Given how Apple’s internal prioritization allegedly works, multiple reports cannot hurt at all.

I’ve heard that was well. Will duplicate your report.

> I wonder whether Apple is aware of Homebrew-Cask

I’d say there’s a good chance. [Max Howell](https://mxcl.github.io/) (original creator of homebrew) now works at Apple creating Swift’s package manager.

> and whether they consider it a partial competitor to the Mac App Store?

I doubt it. They have a graphical interface that promotes discoverability and directly influence developers (both put restrictions on and give advantages to them), and provide a service to end-users where they can easily buy their apps. We have a CLI interface [fervently against discoverability](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not-a-discoverability-service.md) that has no influence at all on developers (they don’t even need to know we exist), and have no monetary or technological stake in the developer-user (money) transaction.

Also, if a developer sells through the Mac App Store, it’s likely that’ll be the only method to acquire the app. If the app is distributed through us, it’ll likely not be the first method on the list mentioned by the developer (a user that opened a terminal would be confused).

There are many reasons for a developer to leave the Mac App Store, but none of them include ease of installation through homebrew-cask, I’d say.

reitermarkus:
Ok, I found a solution using magic numbers, at least for `gzip`.

```ruby
require 'zlib'

def tar_gz?(file)
  Zlib::GzipReader.open(file) { |gzip|
    gzip.read(257)
    return false unless (bytes = gzip.read(6))
    bytes.unpack('C*') == "ustar\0".unpack('C*')
  }
end
```

jellyfishcoder:
@reitermarkus I am not sure how to add this.

reitermarkus:
This is working for me:

```ruby
require 'tmpdir'
require 'zlib'

class Hbc::Container::Gzip < Hbc::Container::Base
  def self.me?(criteria)
    criteria.file.include?('application/x-gzip;') && !tar_gz?(criteria.path)
  end

  def self.tar_gz?(file)
    Zlib::GzipReader.open(file) do |gzip|
      gzip.read(257)
      return false unless (bytes = gzip.read(6))
      bytes.unpack('C*') == "ustar\0".unpack('C*')
    end
  end

  def extract
    Dir.mktmpdir do |unpack_dir|
      @command.run!('/usr/bin/ditto',  :args => ['--', @path, unpack_dir])
      @command.run!('/usr/bin/gunzip', :args => ['-q', '--', Pathname(unpack_dir).join(@path.basename)])
      @command.run!('/usr/bin/ditto',  :args => ['--', unpack_dir, @cask.staged_path])
    end
  end
end
```

jellyfishcoder:
@reitermarkus Just wondering, what does this do? `bytes.unpack('C*') == "ustar\0".unpack('C*')`, It works, I just am wondering what it does so I could try and fix the other file types too.

reitermarkus:
@jellyfishcoder, `unpack` decodes a string and returns the decoded values as an array. `C*` means that the string is split into [8-bit character codes](http://ruby-doc.org/core-2.3.1/String.html#method-i-unpack). I don't know if this is even necessary here, as the string doesn't come from an external tool, but I wanted to keep it consistent with `Hbc::Container::Criteria.magic_number`.

reitermarkus:
A zipped `tar` won't show as `application/x-tar`, so the check in `Hbc::Container::Tar` should now be:

```ruby
criteria.file.include?('application/x-tar') || criteria.tarball?
```

lib/hbc/container/tar.rb
reitermarkus: A zipped `tar` won't show as `application/x-tar`, so the check in `Hbc::Container::Tar` should now be:

```ruby
criteria.file.include?('application/x-tar') || criteria.tarball?
```

reitermarkus:
@jellyfishcoder, FileZilla doesn't install because you only check for `application/x-tar` in `Hbc::Container::Tar`.

You need to add the `tarball?` method I posted above to `Hbc::Container::Criteria`.

Then, in `Hbc::Container::Tar ` check for:

```ruby
criteria.file.include?('application/x-tar') || criteria.tarball?
```

And in `Hbc::Container::Bzip2`, check for:
```ruby
criteria.file.include?('application/x-bzip2;') && !criteria.tarball?
```

reitermarkus:
I think it would be best to have both these methods in `Criteria` so they can depend on each other, so `tarball?` doesn't get called when `tar_gz?` is already `true`.

```ruby
class Hbc::Container::Criteria
  def tar_gz?
    @tar_gz ||= file.include?('application/x-gzip;') && (
      Zlib::GzipReader.open(@path) do |gzip|
        gzip.read(257)
        bytes = gzip.read(6)
        bytes.unpack('C*') == "ustar\0".unpack('C*') unless bytes.nil?
      end
    )
  end

  def tarball?
    @tarball ||= tar_gz? || (
      result = @command.run('/usr/bin/tar', :args => ['-tf', path])
      result.success? && result.stdout.length > 0 && result.stderr.length == 0
    )
  end
end
```

claui:
@reitermarkus No idea if this helps but you can submit PRs against the `master` branch of [jellyfishcoder/homebrew-cask](https://github.com/jellyfishcoder/homebrew-cask) any time if that would be more convenient for you.

hrubymar10:
Hi, can someone please fix Cask provisionql
It has same error (Uh oh, could not figure out how to unpack...)
Thank you

jellyfishcoder:
@reitermarkus Putting both in criteria produces many strange errors for me, most of which point back to issues with @command.

reitermarkus:
@claui, yes, I think that would be more convenient.
@jellyfishcoder, just sent you a PR: https://github.com/jellyfishcoder/homebrew-cask/pull/1

lib/hbc/container/gzip.rb
```diff
@@ -2,10 +2,11 @@
 # for pure gzip only, not tar-gzip (.tgz or .tar.gz)
 
 require 'tmpdir'
+require 'zlib'
```
jawshooah: This should go in `criteria.rb`.

reitermarkus:
There currently is an infinite loop because of `tar_bz2?`. It is blocking when the `bz2` contains less than `257 + 6` bytes.

reitermarkus:
Ok, `tar_bz2?` now uses `IO.read_nonblock` and a timeout.

Tested this with a `10GB.tar.bz2` file. On my machine, `tar -tf` took 55.83s while `tar_bz2?` only needed 0.09s. Quite nice. 😄

reitermarkus:
@jellyfishcoder, just pushed another commit, this time using `IO.popen`, this should be more stable than the version with the timeout.

jellyfishcoder:
@reitermarkus, Sorry it took so long to merge it, just been pretty busy.

reitermarkus:
@jellyfishcoder, no problem, I know how it is. =)

Just sent you another PR.

reitermarkus:
@jellyfishcoder, this needs a rebase because of my changes to `magic_number`.

jellyfishcoder:
I rebased, so what is the conflict it is saying needs to be resolved?

reitermarkus:
> While this was going on, someone modified the file in a commit changing or to || I think.

Yes, that was me, that's already merged, though. 😄

jellyfishcoder:
Just wondering, is this on hold until OS Sierra is public, or are there more issues, because to me there seem to be no issues.

reitermarkus:
There are still some merge conflicts that need to be resolved in `lib/hbc/container/ttf.rb`, `lib/hbc/container/seven_zip.rb` and `lib/hbc/container/pkg.rb`.

jellyfishcoder:
I believe the merge issues are because it was updated since this fork was made, I know how to fix the merge, and have merged them into a local branch before (there isn't a large conflict, just a tiny change since fork), but not how to get the pull request to do so.

claui:
Aside from the merge conflict, I think this PR also needs a few tests before we proceed to merge.

What I’m thinking about is for each compression format involved, we would compress a small file, stash the resulting archive in `spec/support/binaries`, then write a spec that expects the expanded file to show up.

@everyone Feel free to add some tests; I’ll be glad to help out but not before next week.

jellyfishcoder:
@claui Whats the difference from homebrew-cask/test/support/binaries and homebrew-cask/spec/support/binaries. Because there are a lot of zipped test files in the test/support/binaries folder. (IDK what a spec vs test is, I just noticed that there are already those compressed files in the test folder).

lachlanhunt:
Apple has fixed the issue macOS Seirra beta 2 (16A239j). `file -Izb foo.tar.bz2` now returns the expected result.

reitermarkus:
`compressed-encoding=` still has to be removed from `compressed-encoding=application/x-bzip2` and `compressed-encoding=application/zip`.

`file` on a `gzip` shows as `compressed-encoding=application/x-gzip`.

Maybe this is a bug, too? Consider the following, this would match both `bzip2` and `gzip`:

```
$ file -Izb foo.bz2.gz
application/x-bzip2; charset=binary compressed-encoding=application/x-gzip; charset=binary
```

schwa:
Any progress on this one?

jellyfishcoder:
@schwa, Im not sure what is going on now. Its moved out of my league I think. I don't understand what needs to be done now. But if you want, you can clone this pull and it works on Sierra, just still some stuff that needs to be added.

reitermarkus:
Ok, actually `compressed-encoding=application/x-bzip2` is still working. It only shows `application/x-bzip2`, when the contained file is empty, which would never be the case, and of course I used `touch` to create an empty test file, but apparently an empty file is a special case.

So it seems only `application/zip` has to be changed.

vitorgalvao:
So, a lot of this can be undone, correct? Should we close this and concentrate on a smaller PR that just fixes what needs to be done currently?

nikolay:
Can't this be fixed temporarily before doing this the most perfect way? I have a bunch of packages waiting to be installed the proper way and I estimate thousands of people and most importantly newbies just giving up on Caskroom as they can't do much with it. Is there a workaround we can do while "Waiting for Godot"? How about a note on the readme so people can stop submitting the same bug over and over again?

vitorgalvao:
> Can't this be fixed temporarily before doing this the most perfect way?

Famous programming last words. I’d really rather not. If this is “fixed temporarily”, it will in truth be hacky indefinitely.

If you go through the issue (I only followed it tangentially) you should be able to figure out what’s the fix (since it’s a subset of what has already been done). I’d rather someone bitten by this fixes it correctly. We already have too many “fix this later”s on the codebase.

> I estimate thousands of people and most importantly newbies just giving up on Caskroom as they can't do much with it

That number was plucked out of thin air. I very much doubt there are thousands of people that suddenly discovered HBC right after upgrading to a month old beta on their main machine and are just saying “oh well, guess I won’t use it” after seeing an error that occurs in a subset (this only happens on zips) of casks.

> How about a note on the readme so people can stop submitting the same bug over and over again?

It’s not submitted frequently enough for that. Also, there’s a big overlap between people that don’t search issues before making new ones and people who do not read the README.

nikolay:
@vitorgalvao Let's not blow this out of proportion and rely on cliches in an engineering process - all software is delivered in phases and so are some bug fixes. I have several decades of successful development and, no, these were never my last words. Perfection is a direction, not a destination. Software development is a sequence of incremental improvements, not a single-shot perfection, which is a chimera.

There's nothing wrong with having two parallel PRs about the same issue - one fixing it temporarily in production, and the other - the long-term solution. As we know, macOS Sierra has been in Public Beta for a while already, so, this definitely is a production issue.

vitorgalvao:
@nikolay If I insulted you, I apologise. “Famous programming last words” is, to me, a lighthearted joke (you’re not actually going to die after saying them, [it’s just a trope](http://tvtropes.org/pmwiki/pmwiki.php/Main/FamousLastWords)).

I’ll also say, [with all due respect](https://youtu.be/lQECNWI4RHI) (again, that’s a joke, I keep replies like that in TextExpander snippets for some fun interactions), that claiming “thousands of people giving up on Caskroom” and claiming this is a production issue is itself blowing things out of proportion.

This is absolutely something that needs addressing, but it’s also the responsibility of users to know what “beta” means when they enrol in beta programs.

Ever since this PR was opened, a bug report was opened with Apple **and it was fixed by them**. That means progress has been made *the right way*. [We don’t want more workarounds](https://github.com/caskroom/homebrew-cask/issues/13201). Workarounds left many messes in the code base in our process, and we are slowly fixing those.

If this is so important to you, I urge you submit the correct fix or pay someone to do it. This is not as pressing as you’re making it seem, as (again) we’re talking about an error on a beta OS, that for all we know will fix this on its next release.

Once again, I apologise if my previous reply insulted you in any way, as that was absolutely not my intention. However, this really isn’t so pressing as you’re making it seem, that it cannot possibly wait for a correct fix. The OS has been publicly available for less than a month, and again, things are expected to break when you use a beta.

nikolay:
@vitorgalvao I don't get insulted by strangers, so, no worries. This is not an issue I have personally - I already work around it; it's all about defending the idea of switching as many people as possible to Homebrew Cask as the more secure and sane way to install desktop apps on OS X/macOS. Anything that can turn off newbies needs addressing, and, thanks to Apple, now every newbie can install a beta version of the OS and make our lives more miserable. 😠 I apologize if I insulted you as well - I didn't mean to. I'm grateful to the endless people who contribute sweat equity to the Open Source ecosystems and I try to do the same where I can contribute anything more than bugs.

claui:
To add another point to those already made: Allow me to emphasize that @reitermarkus, and all the other developers involved, have carved valuable hours off their schedule to investigate the issue and craft those commits. This issue is tricky and they are doing an excellent job. Every developer involved **deserves due respect for their ongoing commitment** no matter what. You are truly amazing!

As one of the occasional core developers, I can only speak for myself but I am still determined to contribute whatever needs to be done to get this issue and other issues fixed. Ultimately, it’ll be done when it’s done. Life keeps throwing things at us. Some of them get done right away; other stuff will be put off; [if u dont like it deal w/it](http://knowyourmeme.com/memes/katy-t3h-pengu1n-of-d00m). Taking all the time we need while refusing to cut corners is IMHO the best thing we can possibly do to bring Homebrew-Cask closer to its goal, as @vitorgalvao has [already pointed out earlier](#issuecomment-233153815).

I can sincerely relate with every user who feels frustrated, unsatisfied, angry, or all at once. Any user who chooses to rage quit is free to do so. To everyone else, be invited to stay part of the exciting alpha stage of the project that is Homebrew-Cask, with all its ups and downs; watch it prosper, and take pleasure in every second that you are using it. **Bask in the Cask.**

nikolay:
Thank you guys! Just one article from today on the topic of perfect code/solution: https://www.exceptionnotfound.net/code-is-never-perfect-code-is-only-ever-good-enough/

vitorgalvao:
Closing in favour of https://github.com/caskroom/homebrew-cask/pull/22857 (thank you, @reitermarkus).

PR #22014, opened 2016-06-15
Introduce block syntax for `url` stanza

claui:
### Changes to the core

---
## Issue

Some providers use disposable URLs, which need to be constructed from some sort of landing site ([recent example](https://github.com/caskroom/homebrew-cask/pull/21952#issuecomment-225855234)). Doing such expensive things in the middle of a Cask definition [has proven difficult](https://github.com/caskroom/homebrew-versions/pull/293#issuecomment-48755213) in the past. Loading a Cask would become unacceptably slow if HTTP/S requests were done in Cask definitions.
## Solution

Albeit rarely, such cases do pop up every now and then; so I decided to write this PR, which introduces an **optional block syntax** for the `url` stanza:

``` rb
url do
  # No known stable URL; fetching disposable URL from landing site
  open('https://example.com/app/landing') do |landing_page|
    content = landing_page.read
    parse(content) # => https://example.com/download?23309800482283
  end
end
```

The block is only evaluated when needed, for example on download time or when auditing a Cask. It is never evaluated when loading a Cask.

The return value of the block is either a `String`, or a `String, Hash` tuple; the latter case is for additional options such as `referrer`.
## Implementation note

For `stanza_proxy.rb`, I could have used `SimpleDelegator` but it turned out to be much faster to roll my own logic.

The proxy is designed to be easily reused for other stanzas if needed.
## Credits

Thanks @tobadia for contributing the regex/`open-uri` snippet in `audacity.rb`, which also triggered the idea for this PR.
## Feedback

Pinging @caskroom/maintainers for opinions (and for suggestions to improve my convoluted English in the docs). Thanks!

### DIFF
--- a/Casks/audacity.rb
+++ b/Casks/audacity.rb
@@ -1,20 +1,18 @@
-# Audacity does not provide a fixed URL
-# Their download URL points to a html page that generates a temporary URL embedded within an iframe
-# 'open-uri' is required to open that page and grab the temporary URL
-require 'open-uri'
-
 cask 'audacity' do
   version '2.1.2'
   sha256 '2e4b7d608ecc0d2f79bf16663f085d383075e488f7d50bf7d74c0b69173defe7'
 
-  # Current official URL as proposed on http://www.audacityteam.org/download/mac/
-  # must be parsed to extract temporary url embedded in iframe
-  iframe_src_protected_url = nil
-  open("http://www.fosshub.com/Audacity.html/audacity-macosx-ub-#{version}.dmg") do |io|
-    content = io.read
-    iframe_src_protected_url = %r{^\<iframe.*src=\"(http.*\.dmg)\".*>}.match(content)[1].to_s
+  url do
+    # Audacity does not provide a fixed URL
+    # Their download URL points to a html page that generates a temporary URL embedded within an iframe
+    # 'open-uri' is required to open that page and grab the temporary URL
+    require 'open-uri'
+    # fosshub.com/Audacity.html was verified as official when first introduced to the cask
+    open("http://www.fosshub.com/Audacity.html/audacity-macosx-ub-#{version}.dmg") do |io|
+      content = io.read
+      %r{^\<iframe.*src=\"(http.*\.dmg)\".*>}.match(content)[1].to_s
+    end
   end
-  url iframe_src_protected_url
   name 'Audacity'
   homepage 'http://audacityteam.org'
   license :gpl
--- a/Casks/audacity.rb
+++ b/Casks/audacity.rb
@@ -1,9 +1,20 @@
+# Audacity does not provide a fixed URL
+# Their download URL points to a html page that generates a temporary URL embedded within an iframe
+# 'open-uri' is required to open that page and grab the temporary URL
+require 'open-uri'
+
 cask 'audacity' do
-  version '2.1.2-1453294898'
+  version '2.1.2'
   sha256 '2e4b7d608ecc0d2f79bf16663f085d383075e488f7d50bf7d74c0b69173defe7'
 
-  # oldfoss.com:81/download/Audacity was verified as official when first introduced to the cask
-  url "http://app.oldfoss.com:81/download/Audacity/audacity_macosx_ub_#{version.dots_to_underscores}.dmg"
+  # Current official URL as proposed on http://www.audacityteam.org/download/mac/
+  # must be parsed to extract temporary url embedded in iframe
+  iframe_src_protected_url = nil
+  open("http://www.fosshub.com/Audacity.html/audacity-macosx-ub-#{version}.dmg") do |io|
+    content = io.read
+    iframe_src_protected_url = %r{^\<iframe.*src=\"(http.*\.dmg)\".*>}.match(content)[1].to_s
+  end
+  url iframe_src_protected_url
   name 'Audacity'
   homepage 'http://audacityteam.org'
   license :gpl
--- a/doc/cask_language_reference/all_stanzas.md
+++ b/doc/cask_language_reference/all_stanzas.md
@@ -8,7 +8,7 @@ Each of the following stanzas is required for every Cask.
 | ------------------ |------------------------------ | ----------- |
 | `version`          | no                            | application version; give value of `:latest`  if versioned downloads are not offered
 | `sha256`           | no                            | SHA-256 checksum of the file downloaded from `url`, calculated by the command `shasum -a 256 <file>`. Can be suppressed by using the special value `:no_check` (see also [Checksum Stanza Details](stanzas/sha256.md))
-| `url`              | no                            | URL to the `.dmg`/`.zip`/`.tgz`/`.tbz2` file that contains the application. A [comment](stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) should be added if the hostnames in the `url` and `homepage` stanzas differ (see also [URL Stanza Details](stanzas/url.md))
+| `url`              | no                            | URL to the `.dmg`/`.zip`/`.tgz`/`.tbz2` file that contains the application.<br />A [comment](stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) should be added if the hostnames in the `url` and `homepage` stanzas differ. Block syntax should be used for URLs that change on every visit.<br />See [URL Stanza Details](stanzas/url.md) for more information.
 | `name`             | yes                           | a string providing the full and proper name defined by the vendor (see also [Name Stanza Details](stanzas/name.md))
 | `homepage`         | no                            | application homepage; used for the `brew cask home` command
 | `license`          | no                            | a symbol identifying the license category for the application (see also [License Stanza Details](stanzas/license.md))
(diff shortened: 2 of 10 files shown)


Casks/audacity.rb
```diff
@@ -1,9 +1,22 @@
+# Audacity does not provide a fixed URL
+# Their download URL points to a html page that generates a temporary URL embedded within an iframe
+# 'open-uri' is required to open that page and grab the temporary URL
+require 'open-uri'
```
vitorgalvao: Won’t this happen on every cask load? Since this is exclusively needed for `url`, why not have the `require` inside the `url` block? Any reason why that wouldn’t work?

Casks/audacity.rb
```diff
@@ -1,9 +1,22 @@
...
+  version '2.1.2'
   sha256 '2e4b7d608ecc0d2f79bf16663f085d383075e488f7d50bf7d74c0b69173defe7'
 
-  # oldfoss.com:81/download/Audacity was verified as official when first introduced to the cask
-  url "http://app.oldfoss.com:81/download/Audacity/audacity_macosx_ub_#{version.dots_to_underscores}.dmg"
+  url do
+    # Current official URL as proposed on
+    # http://www.audacityteam.org/download/mac/
+    # must be parsed to extract temporary url embedded in iframe
+    base_url = 'http://www.fosshub.com/Audacity.html'
```
vitorgalvao: I’d rather remove this line and have `open("http://www.fosshub.com/Audacity.html/audacity-macosx-ub-#{version}.dmg") do |io|` directly.

That way we can still add in a clear way the required line `# fosshub.com/Audacity.html was verified as official when first introduced to the cask`.

doc/cask_language_reference/all_stanzas.md
```diff
@@ -8,7 +8,7 @@ Each of the following stanzas is required for every Cask.
...
 | `version`          | no                            | application version; give value of `:latest`  if versioned downloads are not offered
 | `sha256`           | no                            | SHA-256 checksum of the file downloaded from `url`, calculated by the command `shasum -a 256 <file>`. Can be suppressed by using the special value `:no_check` (see also [Checksum Stanza Details](stanzas/sha256.md))
-| `url`              | no                            | URL to the `.dmg`/`.zip`/`.tgz`/`.tbz2` file that contains the application. A [comment](stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) should be added if the hostnames in the `url` and `homepage` stanzas differ (see also [URL Stanza Details](stanzas/url.md))
+| `url`              | no                            | URL to the `.dmg`/`.zip`/`.tgz`/`.tbz2` file that contains the application.<br />A [comment](stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) should be added if the hostnames in the `url` and `homepage` stanzas differ. Block syntax should be used for URLs that change on every visit.<br />See [URL Stanza Details](stanzas/url.md) for more information.
```
vitorgalvao:
```diff
- Block syntax should be used for URLs
+ Block syntax can be used for URLs
```

Or something like that. I’d like to not actively encourage the block syntax.

Casks/audacity.rb
claui: For code, I like to stay within a 68-character-per-line limit.

That said, I feel it’s perfectly acceptable for Cask definitions to go over that limit.
Changing it now, thanks for the pointer.

Casks/audacity.rb
claui: It does work, and makes for a good improvement.
Changing this, thanks!

claui:
@vitorgalvao Changed as per your suggestions, thanks.
I also cut short redundant comments and added the required line:

```
# fosshub.com/Audacity.html was verified as official when first introduced to the cask
```

doc/cask_language_reference/stanzas/url.md
```diff
@@ -84,8 +84,49 @@ Also make sure to give the URL for the binary download itself, rather than a pre
 
 ## Some Providers Block Command-line Downloads
 
-Some hosting providers actively block command-line HTTP clients (example: FossHub). Such URLs cannot be used in Casks.
+Some hosting providers actively block command-line HTTP clients. Such URLs cannot be used in Casks.
```
vitorgalvao: Why remove the FossHub example?

claui: The example has not been removed. I moved it to the next line:

> Some providers do not actively block command-line HTTP clients but use URLs that change periodically, or even on each visit (example: FossHub).

I did this for two reasons:

1. My new text wants to establish a clear difference between hosting providers that _actively block_ CLI clients (e. g. using captchas) vs. those that just _use disposable URLs_ for whatever reason. There are certainly many possible reasons for a provider to go for disposable URLs, even if that choice means breaking CLIs as a side effect. I feel it wouldn’t be unreasonable for us to politely err on the side of the accused, rather than implying that FossHub made their choices for a certain reason without giving actual evidence.

2. To me, the current wording would imply that both #21952 and #22014 are acts of [circumventing a technological barrier](https://en.wikipedia.org/wiki/Anti-circumvention) in a way FossHub does not wish to allow. While I am willing to take responsibility for my actions, I certainly don’t feel good about making assumptions which in the worst case might someday be used against the project and its participants. On the other hand, IANAL; my comment is just an account of how I currently feel about the whole thing.

That all said, I dislike how FossHub’s choices makes our lives harder. I’d be glad if someone would inform them about the issue and make them aware of the hoops through which their technical decision makes us jump.

Regarding the docs, I am completely open for feedback/suggestions how to reword this.

vitorgalvao: Sounds good. Great detail, as always.

doc/cask_language_reference/stanzas/url.md
```diff
@@ -84,8 +84,49 @@ Also make sure to give the URL for the binary download itself, rather than a pre
...
+
+Some providers do not actively block command-line HTTP clients but use URLs that change periodically, or even on each visit (example: FossHub). For those, see section _URLs that Change on Every Visit._
```
vitorgalvao:
```diff
- see section _URLs that Change on Every Visit._
+ see section [URLs that Change on Every Visit](#urls-that-change-on-every-visit).
```

It will link directly, this way.

doc/cask_language_reference/stanzas/url.md
```diff
@@ -84,8 +84,49 @@ Also make sure to give the URL for the binary download itself, rather than a pre
...
+
+## URLs that Change on Every Visit
+
+Some providers use disposable URLs, which a Cask author cannot know in advance. Such URLs may change daily, or on every visit, and sometimes need to be dynamically obtained from a landing site.
+
+### The Problem
+
+In theory, one can write arbitrary Ruby code right in the Cask definition to fetch and construct a disposable URL.
+
+However, this typically involves an HTTP/S round trip to a landing site, which may take a long time. Because of the way Homebrew Cask loads and parses Casks, it is not acceptable that such expensive operations be performed directly in the body of a Cask definition.
```
vitorgalvao: Nitpick:

```diff
- Homebrew Cask
+ Homebrew-Cask
```

doc/cask_language_reference/stanzas/url.md
claui: My verbosity can be both a blessing and a curse. :smile:

doc/cask_language_reference/stanzas/url.md
```diff
@@ -84,8 +84,49 @@ Also make sure to give the URL for the binary download itself, rather than a pre
...
+
+### Using a Block to Defer Code Execution
+
+Similar to the `preflight`, `postflight`, `uninstall_preflight`, and `uninstall_postflight` blocks, the `url` stanza offers an optional _block syntax:_
```
vitorgalvao: Other small nitpick:

```diff
- _block syntax:_
+ _block syntax_:
```

(the `:` should not be part of the emphasis)

doc/cask_language_reference/stanzas/url.md
```diff
@@ -84,8 +84,49 @@ Also make sure to give the URL for the binary download itself, rather than a pre
...
+```
+
+The block is only evaluated when needed, for example on download time or when auditing a Cask.
+Inside a block, you may safely do things such as HTTP/S requests that may take a long time to execute. You may also refer to the `@cask` instance variable, and invoke any method available on `@cask`.
+
+The block will be called immediately before downloading; its result value will be assumed to be a `String` and subsequently used as a download URL.
+
+You can use the `url` stanza with either a direct argument or a block but not with both.
+
+Example for using the block syntax: [audacity.rb](https://github.com/caskroom/homebrew-cask/tree/master/Casks/audacity.rb)
```
vitorgalvao: Note: after we merge this, we need to remember to change this link to point to the specific revision and highlight the specific lines (it’s what we do in the other examples).

vitorgalvao:
Everything looks great.

This could provide immense value in the [caskroom/versions](https://github.com/caskroom/homebrew-versions) repo. We could switch all nightlies to use `:latest` and `:no_check` and make `url` always grab the most up-to-date version. Most nightly should prefer that to permanently outdated casks.

claui:
Committed fixes to `url.md` regarding typography, spelling, and the link anchor.

claui:
Rebased.

Leaving the PR open for another few hours before I proceed to merge.

vitorgalvao:
I wonder in which fun way `cask-repair` will blow up with these casks. I’ll likely just make it abort if it finds a cask with a block `url`.

PR #22030, opened 2016-06-16
Show all versions with `brew cask list --versions`.

reitermarkus:
### Changes to the core

Fixes part 4 of https://github.com/caskroom/homebrew-cask/issues/14058.

### DIFF
--- a/lib/hbc/cask.rb
+++ b/lib/hbc/cask.rb
@@ -64,6 +64,23 @@ class Hbc::Cask
     subdir
   end
 
+  def version_comparator
+    Proc.new do |x, y|
+      begin
+        x_version  =  Gem::Version.new(x)
+        x_version <=> Gem::Version.new(y)
+      rescue ArgumentError => e
+        x <=> y
+      end
+    end
+  end
+
+  def versions
+    Pathname.glob(metadata_master_container_path.join('*'))
+      .map { |p| p.basename.to_s }
+      .sort &version_comparator
+  end
+
   def installed?
     staged_path.exist?
   end
--- a/lib/hbc/cli/list.rb
+++ b/lib/hbc/cli/list.rb
@@ -59,7 +59,7 @@ class Hbc::CLI::List < Hbc::CLI::Base
     if @options[:one]
       puts columns
     elsif @options[:versions]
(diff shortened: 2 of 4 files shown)


vitorgalvao:
Thank you. The submission is appreciated. Even if one day we’ll need to remove this ([since we’ll always be keeping only one version](https://github.com/caskroom/homebrew-cask/issues/13201)), for now it means `list` will work better (as users expect), and that in itself means progress.

lib/hbc/cli/list.rb
```diff
@@ -55,7 +55,7 @@ def self.list_installed
     if @options[:one]
       puts columns
     elsif @options[:versions]
-      installed_casks.each { |cask| puts "#{cask} #{cask.version}" }
+      installed_casks.each { |cask| puts "#{cask} #{cask.versions.reverse.join(', ')}" }
```
reitermarkus: Any opinion on whether this should be ordered `oldest – latest` vs. `lastest – oldest`?

claui: No strong opinion on that one, but a slight preference for `latest – oldest` so the most important information (the current version) is visually most accessible. YMMV though.

vitorgalvao: > slight preference for `latest – oldest` so the most important information (the current version) is visually most accessible.

I second this.

lib/hbc/cask.rb
```diff
@@ -64,6 +64,19 @@ def metadata_subdir(leaf, timestamp=:latest, create=false)
     subdir
   end
 
+  def versions
+    Pathname.glob(metadata_master_container_path.join('*'))
+    .map { |p| p.basename.to_s }
+    .sort do |x, y|
```
claui: Please [indent the chained lines](https://github.com/bbatsov/ruby-style-guide#consistent-multi-line-chains), even if that means those lines will flow along the body of the following block.

Even better, you could make the block into a Proc and move it to a separate constant, method, or even class; either way, this will result in shorter methods, which is a good thing.

reitermarkus:
Ok, I separated the version comparator, should be good to merge now.

claui:
Looks great now! 👍
Waiting a few hours in case other maintainers wish to chime in.

PR #22041, opened 2016-06-17
Core: Replace and/or/not keywords with &&/||/!

reitermarkus:
### Changes to the core

Getting started with https://github.com/caskroom/homebrew-cask/issues/22038.

### DIFF
--- a/lib/hbc/utils.rb
+++ b/lib/hbc/utils.rb
@@ -15,17 +15,9 @@ Tty = Hbc::Utils::Tty
 # monkeypatch Object - not a great idea
 class Object
   def utf8_inspect
-    if not defined?(Encoding)
-      self.inspect
-    else
-      if self.respond_to?(:map)
-        self.map do |sub_elt|
-          sub_elt.utf8_inspect
-        end
-      else
-        self.inspect.force_encoding('UTF-8').sub(%r{\A"(.*)"\Z}, '\1')
-      end
-    end
+    return self.inspect unless defined?(Encoding)
+    return self.map(&:utf8_inspect) if self.respond_to?(:map)
+    self.inspect.force_encoding('UTF-8').sub(%r{\A"(.*)"\Z}, '\1')
   end
 end
 
@@ -60,9 +52,9 @@ def onoe(error)
 end
 
 def odebug(title, *sput)
-  if Hbc.respond_to?(:debug) and Hbc.debug
+  if Hbc.respond_to?(:debug) && Hbc.debug
     width = Tty.width * 4 - 6
-    if $stdout.tty? and title.to_s.length > width
+    if $stdout.tty? && title.to_s.length > width
       title = title.to_s[0, width - 3] + '...'
     end
     puts "#{Tty.magenta.bold}==>#{Tty.reset.bold} #{title}#{Tty.reset}"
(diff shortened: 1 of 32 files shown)


lib/hbc/utils.rb
```diff
@@ -15,7 +15,7 @@ module Hbc::Utils; end
 # monkeypatch Object - not a great idea
 class Object
   def utf8_inspect
-    if not defined?(Encoding)
+    unless defined?(Encoding)
```
jawshooah: I prefer `if !condition ... else` over `unless condition ... else`, as recommended by the [style guide](https://github.com/bbatsov/ruby-style-guide#no-else-with-unless).

jawshooah:
Pleasantly surprised to see Travis go green after just one pass! Once my little nitpick is addressed, I'm fine to merge 👍

claui:
Thanks @reitermarkus for tackling this! LGTM :+1:

lib/hbc/utils.rb
vitorgalvao: Wouldn’t the style guide be against `if !condition … else` as well? That seems just as confusing. My interpretation of it is: “never start negating if you’re going to have an else”.

So no `if !condition … else` or `unless condition … else` but only `if condition … else`.

jawshooah: I think in general, the "expected" case should come first, regardless of whether it's a negation. Guard clauses are an exception, as they allow you to avoid nesting.

In this particular case, however, the exceptional case is handled first, so I agree with you that they should be flipped.

claui: One question I find more important than any standard is, »Will a person be able to glance at the code and immediately know what’s going on?«

With so much nesting already going on, flipping does not help this method much.
Look at the `inspect` thing. It is clearly an edge case; it says »I cannot play with you guys and your encoding stuff, I’ll show myself the door«. I believe that is an excellent candidate for a **guard clause** like @jawshooah suggested.

Let’s look at the method as it stands:

```ruby
def utf8_inspect
  if not defined?(Encoding)
    self.inspect
  else
    if self.respond_to?(:map)
      self.map do |sub_elt|
        sub_elt.utf8_inspect
      end
    else
      self.inspect.force_encoding('UTF-8').sub(%r{\A"(.*)"\Z}, '\1')
    end
  end
end
```

Now put a guard clause in place:

```ruby
def utf8_inspect
  return self.inspect unless defined?(Encoding)
  if self.respond_to?(:map)
    self.map(&:utf8_inspect)
  else
    self.inspect.force_encoding('UTF-8').sub(%r{\A"(.*)"\Z}, '\1')
  end
end
```

The edge case is now blatantly obvious: first line, GTFO.
On top of this, considering the `&` trick I covertly applied, we have just **shedded two levels** of nesting.

jawshooah: And we could clean it up even further with one more guard clause (though whether or not this is cleaner is definitely subjective):

```ruby
def utf8_inspect
  return self.inspect unless defined?(Encoding)
  return self.map(&:utf8_inspect) if self.respond_to?(:map)
  self.inspect.force_encoding('UTF-8').sub(%r{\A"(.*)"\Z}, '\1')
end
```

reitermarkus: Now that's what I call clean code. =)

I will definitely take a second look at everything I have changed to see if there are other such cases. My focus for the first run was to change the keywords without breaking anything.

jawshooah: Let's merge this as-is; we don't want the changes to go beyond the original scope of the PR.

PR #22063, opened 2016-06-17
menumeters doesn't work on el capitan

Amorymeltzer:
#### Editing an existing cask

See http://www.ragingmenace.com/software/menumeters/#requirement — `yujitach-menumeters.rb` is presumably a [suitable replacement](https://github.com/caskroom/homebrew-cask/pull/14145).  I'd be open to a caveat here but I know that's a fairly political issue. 😉

### DIFF
--- a/Casks/menumeters.rb
+++ b/Casks/menumeters.rb
@@ -7,6 +7,8 @@ cask 'menumeters' do
   homepage 'http://www.ragingmenace.com/software/menumeters/'
   license :gpl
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
I’d say it’s a good candidate for the `discontinued` caveat, but it’s not clear if it really is discontinued or not.

Amorymeltzer:
Maybe worth revisiting once Sierra comes out

PR #22067, opened 2016-06-17
Renames Telegram Desktop recipe to telegram-desktop.

mhubig:
This renaming is needed to avoid confusion with the native version of Telegram. I'll provide a new recipe for the nativ client as `telegram.rb` in an separate pull request once this one is merged.

The native version can be found here https://macos.telegram.org compared to the desktop version which can be found here https://desktop.telegram.org ...

### DIFF
--- a/Casks/telegram-desktop.rb
+++ b/Casks/telegram-desktop.rb
@@ -1,4 +1,4 @@
-cask 'telegram' do
+cask 'telegram-desktop' do
   version '0.9.51'
   sha256 '7d898a6ddc2c9a2722d37867d22b422d63cf218f721dbb8942d67696b2f6aadd'
 


vitorgalvao:
We need to follow a strict [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md), to avoid duplicates.

A link to that documentation must have appeared in the pull request template that shows when you submit.

mhubig:
Sorry to follow up again but according to the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) the renamed recipe is **now** correct:

App Name on Disk       | Simplified App Name | Cask Token       | Filename
-----------------------|---------------------|------------------|----------------------
`Telegram Desktop.app` | Telegram Desktop    | telegram-desktop | `telegram-desktop.rb`

vitorgalvao:
Look at the `app` stanza for this cask. This links an app called `Telegram.app`, not `Telegram Desktop.app`.

mhubig:
Shit you're right ... so there are two apps with the same on-disk-name. So how can this be handled? Since `Telegram for macOS` seems to be the default app for macOS I would suggest this schema:

App Name on Homepage | App Name on Disk    | Simplified App Name | Cask Token       | Filename
---------------------|---------------------|---------------------|------------------|---------------------
Telegram Desktop     | `Telegram.app`      | Telegram Desktop    | telegram-desktop | `telegram-desktop.rb`
Telegram for macOS   | `Telegram.app`      | Telegram            | telegram         | `telegram.rb`

An alternative would be:

App Name on Homepage | App Name on Disk    | Simplified App Name | Cask Token       | Filename
---------------------|---------------------|---------------------|------------------|---------------------
Telegram Desktop     | `Telegram.app`      | Telegram Desktop    | telegram-desktop | `telegram-desktop.rb`
Telegram for macOS   | `Telegram.app`      | Telegram MacOS   | telegram-macos         | `telegram-macos.rb`

vitorgalvao:
`for macOS` and variants also needs to be removed. The token reference is comprehensive.

We also only have one `telegram`, this is not an issue. We used to have two, but one had to be removed. Search the issues to learn why.

mhubig:
Please look at #22092 for another try. Would be very sad if we could only have one telegram app, especially since **Telegram for macOS** is definitely the better one ...

vitorgalvao:
Doesn’t matter which one is better. Going through the issues I see we removed one of them because it was no longer available outside the Mac App Store. Nothing we can do about that.

Since they’re both available again, we can readd it. Make the requested fixes as we will merge it.

Thank you for the submission.

PR #22209, opened 2016-06-19
New breadbrumbs approach to reporting bugs

vitorgalvao:
These changes aim to simplify the README (that block of code in `Reporting Bugs` always seemed a bit much and kind of drawned out the information — it still exists, but in a separate location) and introduce a new approach to reporting bugs.

A big issue is we get some of the same reports over and over, and many of those aren’t our fault. If users did a bit of work before opening issues, we’d save a lot of time. Here we should try to document the most common issues, and the way to go about solving them.

Also included a new ISSUE_TEMPLATE that outright tells users to not open new issues without following the available methods first.

Currently the only common issue tackled in this method is a `curl` issue, and even that already required a lot of branching possibilities. Will tackle more of them after some feedback.

If you want to try it out, go [to my branch](https://github.com/vitorgalvao/homebrew-cask/tree/breadcrumbs#reporting-bugs) and pick `A cask fails to install`.

Pinging @caskroom/maintainers and @Amorymeltzer.


doc/reporting_bugs/pre_bug_report.md
```diff
@@ -0,0 +1,15 @@
+Before reporting a bug, lets make sure everything is right with your setup.
+
+When reporting bugs, remember that Homebrew-Cask is an independent project from Homebrew. Do your best to direct bug reports to the appropriate project. If your command started with `brew cask`, bring the bug to us first.
+
+Start by searching for your issue before posting a new one. If you find it and have any new information not reported in the original issue, please add your insights.
```
jawshooah: Probably important to specify that if the issue is already closed, and the solution presented in the issue did not work, the user should open a new one rather than commenting on a stale thread.

vitorgalvao: Done.

vitorgalvao:
Seems like there are no objections. Merging.

PR #22251, opened 2016-06-20
Updated file criteria

donysukardi:
`/usr/bin/file -Izb $FILENAME` doesn't return `compressed-encoding=`
prefix on MacOS 10.12b (16A201w)


donysukardi:
Sorry didn't realize that it's already WIP on #22002

PR #22320, opened 2016-06-22
updated sublime-text (3117)

vitorgalvao:
#### Editing an existing cask


kylefberg:
@fanquake – v3117 isn't the most recent version of ST3, it's the dev build. This version bump should be on the sublime-text-dev cask, right? I'm trying to run sublime-text, downloaded from this cask, but it won't let me run it without being registered since that is a dev build and not the current version (v3114) of the public build.

Please revert this PR!

vitorgalvao:
Yes, already noticed and was going to revert.

h4w5:
Awesome, thanks! 👍

PR #22338, opened 2016-06-23
Copied which_editor and exec_editor from brew utils to address #20074. …

### LINKED ISSUE #20074: Feature request: add-in comment of the use of editor
### Description of feature/enhancement
add in a comment of adding this to users profile as what is done in brew

Add this to your .bash_profile or .zsh
export HOMEBREW_EDITOR="subl"

### Justification
not everyone can use the default editor

### Example use case
easy to use sublime then the emacs or vi editor

abrammer:
- Pulled the exec_editor and which_editor functions out of brew utils.rb so that the same warning message is echoed for brew cask edit when HOMEBREW_EDITOR isn't defined, as raised in #20074.  
- Tested with "bbedit" and "subl -w" ,  argument flags seem to be handled fine.  
- Code straight out of brew so not really my doing, I'm not familiar with the differences between safe_exec vs exec but it works. 

Closes #20074.

### DIFF
--- a/lib/hbc/utils.rb
+++ b/lib/hbc/utils.rb
@@ -97,13 +97,31 @@ module Hbc::Utils
     return cmd_pn
   end
 
-  def self.exec_editor(*args)
-    editor = [
-              *ENV.values_at('HOMEBREW_EDITOR', 'VISUAL', 'EDITOR'),
-              *%w{mate edit vim /usr/bin/vim}.map{ |x| which(x) }
-             ].compact.first.to_s
-    exec(*editor.split.concat(args))
-  end
+def self.which_editor
+  editor = ENV.values_at("HOMEBREW_EDITOR", "VISUAL", "EDITOR").compact.first
+  return editor unless editor.nil?
+
+  # Find Textmate
+  editor = "mate" if which "mate"
+  # Find BBEdit / TextWrangler
+  editor ||= "edit" if which "edit"
+  # Find vim
+  editor ||= "vim" if which "vim"
+  # Default to standard vim
+  editor ||= "/usr/bin/vim"
+
+  opoo <<-EOS.undent
+    Using #{editor} because no editor was set in the environment.
+    This may change in the future, so we recommend setting EDITOR, VISUAL,
+    or HOMEBREW_EDITOR to your preferred text editor.
+  EOS
+
+  editor
+end
+
+def self.exec_editor(*args)
+  safe_exec(which_editor, *args)
+end
 
   # originally from Homebrew puts_columns
   def self.stringify_columns items, star_items=[]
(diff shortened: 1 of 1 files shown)


lib/hbc/utils.rb
```diff
@@ -97,13 +97,31 @@ def self.resolve_executable(cmd)
...
   end
 
-  def self.exec_editor(*args)
-    editor = [
-              *ENV.values_at('HOMEBREW_EDITOR', 'VISUAL', 'EDITOR'),
-              *%w{mate edit vim /usr/bin/vim}.map{ |x| which(x) }
-             ].compact.first.to_s
-    exec(*editor.split.concat(args))
-  end
+def self.which_editor
```
vitorgalvao: You remove indentation on the whole block. It should be kept as was.

vitorgalvao: Add the `# originally from Homebrew` comment [we use on the rest of the file](https://github.com/caskroom/homebrew-cask/pull/22338/files#diff-b1656a67650a15c2e1770e4ac80e1fdeR49).

vitorgalvao:
Pinging developers that work most on the core: @claui @jawshooah @mwean.

No objections here.

lib/hbc/utils.rb
```diff
@@ -97,13 +97,31 @@ def self.resolve_executable(cmd)
...
+  return editor unless editor.nil?
+
+  # Find Textmate
+  editor = "mate" if which "mate"
+  # Find BBEdit / TextWrangler
+  editor ||= "edit" if which "edit"
+  # Find vim
+  editor ||= "vim" if which "vim"
+  # Default to standard vim
+  editor ||= "/usr/bin/vim"
```
mwean: I actually prefer the old implementation at line 103 where there's a list of editors and we just find the first one that exists. On the other hand, it feels strange to do any checking at all, when something like git would just use vim if nothing else was set.

vitorgalvao: I don’t really mind either implementation. The only thing this needs to do to satisfy https://github.com/caskroom/homebrew-cask/issues/20074 is output the message when no editor is set.

I also prefer the shorter check we had, but standardising the check to be whatever homebrew uses is good for when we merge.

abrammer:
Sorry git noob.  Royally f*d up all of those minor edits. Hopefully that last commit (not sure why it was duplicated) indents and adds the comments as desired.

vitorgalvao:
Merging so it already gets included in @anastasiasulyagina’s search for code taken from homebrew.

vitorgalvao:
Thank you for the contribution, @abrammer.

pacav69:
Great work guys

PR #22360, opened 2016-06-24
Updated ssh-tunnel-manager (2.2.1)

RJHsiao:
#### Editing an existing cask

### DIFF
--- a/Casks/ssh-tunnel-manager.rb
+++ b/Casks/ssh-tunnel-manager.rb
@@ -1,8 +1,9 @@
 cask 'ssh-tunnel-manager' do
-  version '2.2'
-  sha256 '0ccd88cb309f9215c53c0cf1a133439710af15c985314d0afa20053495e6d08c'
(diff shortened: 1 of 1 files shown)


RJHsiao:
Hi all,

Sorry for that I can't get the url rule about the new version release, so I just put the static url.

adidalal:
Use format like https://github.com/caskroom/homebrew-cask/blob/master/Casks/vox-preferences-pane.rb and needs URL comment also.

RJHsiao:
Hi @adidalal 
You'll right if there are only two integer inside the url,
but there are three integer inside the url: 
https://dl.devmate.com/org.tynsoe.sshtunnelmanager/221.003/1455813213/SSHTunnelManager-221.003.zip
"221" means the version number "2.2.1", "003" may means "build no.3", 
but how about "1455813213"? I can't sure it's a static number or variable number.

RJHsiao:
@vitorgalvao 
OK, please check it out.

PR #22363, opened 2016-06-24
Simplify "Titanium Software" casks (deeper, maintenance, onyx).

reitermarkus:
#### Editing an existing cask

### DIFF
--- a/Casks/deeper.rb
+++ b/Casks/deeper.rb
@@ -2,32 +2,20 @@ cask 'deeper' do
   version :latest
   sha256 :no_check
 
-  if MacOS.release == :tiger
-    url 'http://www.titanium.free.fr/download/104/Deeper.dmg'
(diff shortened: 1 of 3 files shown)


vitorgalvao:
This is smart, but likely to break. You removed the special case, which means it’ll try to download a possibly (haven’t checked yet) non-existing download for Sierra. The `else` case should remain separate.

reitermarkus:
It doesn't break, as `:sierra` isn't in the `depends_on` array.

vitorgalvao:
Good point. Even so, even with it not breaking, I’d prefer to have the comment there. If anything, it keeps us alert to the fact that the cask needs to have the new URL checked and the `depends_on` altered.

reitermarkus:
Ok, I re-added the comment above the `depends_on` stanza.

PR #22383, opened 2016-06-25
How to deal with apps with malware

vitorgalvao:
Since we’ve been getting a few reports on this lately (https://github.com/caskroom/homebrew-cask/issues/15799, https://github.com/caskroom/homebrew-cask/issues/21399, https://github.com/caskroom/homebrew-cask/issues/22375), it’s sensible to have some way of dealing with it.

Removing casks is out of the question, and maintainers seem to agree on that. Adding a `caveat` always bothered me, but seems to be generally accepted. This PR adds the `caveat` but also makes the process (what I feel is) correct. It cuts the issue at the source, instead of making us a gatekeeper, which is greatly ineffective.

Pinging @caskroom/maintainers. Anyone from the community that wants to comment, please read the linked issues first.


doc/faq/apps_with_malware.md
```diff
@@ -0,0 +1,13 @@
...
+
+Within that context, we would still like for users to enjoy some kind of protection while minimising occurrences of legitimate developers being branded as malware carriers. We feel removing casks is an ineffective band-aid and the issue needs to be tackled earlier in the chain: at the macOS level.
+
+If an app that bundles malware was not signed with an Apple Developer ID and you purposefully disabled or bypassed Gatekeeper, no action will be taken on our part. When you disable security features, you do so at your own risk. If, however, an app that bundles malware is signed, Apple can revoke its permissions and it will no longer run on the computers of users that keep security features on — we all benefit, Homebrew-Cask users or not.
```
Eitot: A question about this. Gatekeeper is not triggered when the quarantine file attribute has not been set. This seems to be the case with `curl`, which Homebrew uses. How does Homebrew interact with Gatekeeper and File Quarantine? Is it actually triggered when Apple updates the malware list?

vitorgalvao: Quarantine is a file attribute for the *downloaded* file (i.e. the `dmg` or equivalent). Gatekeeper is something different. It checks if the app itself (which came inside the `dmg`) has a valid digital certificate.

Quarantine isn’t a great security measure, and it can be easily bypassed with a simple terminal command.

vitorgalvao: More specifically, Quarantine is the thing that says (paraphrasing) “this file was downloaded from the internet, do you want to open it?” and gives you a choice. Gatekeeper says (again, paraphrasing) “this app is from an unidentified developer so it cannot be run” and doesn’t let you run it unless you bypass it (turn it off or right-click and pick `Open`).

Quarantine is simply set at the browser level, and browsers can ignore it and not set it. Gatekeeper is baked into macOS itself.

Eitot: I believe you are not correct. Gatekeeper builds upon File Quarantine and itself, as you said, depends on the extended file attribute. Whether or not Gatekeeper is triggered, hinges upon the existence of that attribute as well. When it is not there, then neither of the two messages will appear and the application opens normally, regardless of the Gatekeeper configuration.

You can easily verify this by downloading the same disk image (with an app bundle inside) twice, once with Safari and once with `curl`. It doesn’t matter whether the app bundle is signed or not (I’ve tried both). Only the first image will give you that Gatekeeper warning, the second opens without any kind of alerts, even with the strictest Gatekeeper configuration. If you remove the attribute manually from the first file before allowing it through Gatekeeper, then the result is the same.

You can read more about this here: http://www.thesafemac.com/what-is-gatekeeper/

My concern is this: Homebrew does not append this file attribute, which means that it is protected neither by Gatekeeper or File Quarantine.

vitorgalvao: Just did the test myself, and you are quite right. Quite frankly, I find that notion ridiculous. This makes the feature way less secure. Thank you for the correction.

I remembered [this situation](https://github.com/caskroom/homebrew-cask/issues/17540) that clouded my understanding of the feature.

Supposedly if Apple revokes a certificate, then the app will indeed stop working completely (as was my understanding of what they did with the recent Transmission breach), but I was expecting it to also work without Quarantine being set.

That does not stop us from merging this, since in that sense it will still do its job. However, we should look into adding the quarantine attribute. See https://github.com/caskroom/homebrew-cask/issues/22388.

mwean: I'm not very familiar with this stuff, but could you use the `spctl` command to manually check the app? It returns a non-zero exit code if there's something wrong with an app. 

```shell
> spctl -a /Applications/Amazon\ Music.app
/Applications/Amazon Music.app: a sealed resource is missing or invalid
```

Might be something worth looking into.

Eitot: @vitorgalvao The marked passage in the documentation ought to be changed so that it makes clear that Cask does not interact with Gatekeeper presently. It currently reads as if users can still count on Gatekeeper and File Quarantine if malware happens to be installed through Cask, even though that is not the case.

fuzzyroddis:
Is it possible to install the application without the adware like Ninite does?

vitorgalvao:
@fuzzyroddis For whatever app you’re asking, this is the wrong place. This issue is to discuss the feature itself, not specific casks.

doc/faq/apps_with_malware.md
vitorgalvao: @Eitot Agreed. Though I don’t want to change the text as is (it’s good for the future) I’ll add a note that we’re working on it and point to the relevant issue (since we need help with it, it’s a good way to get it).

fuzzyroddis:
@vitorgalvao I'm talking about a solution similar to [Ninite](https://ninite.com) where it passes switches to the installer to tell it not to install crap.

vitorgalvao:
@fuzzyroddis That’s not a solution. That’s software dependent, and most software that installs malware won’t have such flags.

Issue #22359, opened 2016-06-24
Still installing to /opt/homebrew-cask/Caskroom


timmolter:
After seeing the warning, I migrated all apps to ` /usr/local`.
```
Warning: The default Caskroom location has moved to /usr/local/Caskroom.

Please migrate your Casks to the new location, or if you would like to keep your
Caskroom at /opt/homebrew-cask/Caskroom, add the following to your HOMEBREW_CASK_OPTS:

  --caskroom=/opt/homebrew-cask/Caskroom
```
But installing new apps is still installing in the old location.

    libreoffice staged at '/opt/homebrew-cask/Caskroom/libreoffice/5.1.4

Either this is a bug or I'm too dumb to update homebrew properly. Any ideas?

vitorgalvao:
That means `/opt/homebrew-cask` still exists. It’s the only check that is made. Delete that directory completely.

timmolter:
You might want to add that to the message about migrating.

Issue #21913, opened 2016-06-13
How to migrate to the new location?


mattfelsen:
Good morning! Since the location change and https://github.com/caskroom/homebrew-cask/pull/21901, I'm now seeing the warning about migrating Casks, as expected. I've searched around a bit but don't understand what I need to do as an end-user. Is there a command for migrating? Is it simply a move of the `/opt/homebrew-cask/Caskroom/` directory to `/usr/local/Caskroom/`? A bit more info would be helpful as I don't want to muck anything up. Thanks!

jawshooah:
Yes, the simplest thing would be to `mv /opt/homebrew-cask/Caskroom /usr/local`. However, any casks installed prior to #13966 were symlinked rather than moved to their destinations, so you'll need to re-install those manually.

If we do wind up [phasing out `--caskroom`](https://github.com/caskroom/homebrew-cask/issues/21858), it won't be for a while, so you have plenty of time to find old casks and `brew cask install --force` them.

vitorgalvao:
Closing the issue as it has been answered, but [making it the default to answer the questions that arise](https://github.com/caskroom/homebrew-cask/pull/21915).

If new questions come up, we should edit @jawshooah’s comment to make it clearer.

tehmaspc:
I just re-installed one of my Caskroom apps and noticed that it was still installed to /opt/homebrew-cask/Caskroom? Doesn't seem like homebrew-cask is properly defaulting? I don't have any HOMEBREW_CASK_OPTS options set.

msridhar:
@tehmaspc same for me; it was the `eclipse-java` app.  Re-installing other apps hasn't had the same problem thus far.

tehmaspc:
Ah. I tested my `sonos` app. So there are individual casks that must be expecting /opt/homebrew-cask/Caskroom then.

jawshooah:
#21901 changed the behavior to default to the old location if it still exists and print a warning about the new default, to mitigate the issues brought about by the original change. No need to change `HOMEBREW_CASK_OPTS` if you delete `/opt/homebrew-cask/Caskroom`.

grootknol:
After receiving the warning, I've moved my Caskroom to `/usr/local/Caskroom`. Now `brew cask list` lists the installed casks but `brew cask uninstall` can't remove them.

```
➜  Casks git:(master) brew cask list
hazel					textmate
iterm2					the-unarchiver
java					transmission
jettison (!)				vagrant
[...]

PR #22387, opened 2016-06-25
Tell users to delete old caskroom if it exists

jawshooah:
### Changes to the core

Refs #22359

### DIFF
--- a/lib/hbc/locations.rb
+++ b/lib/hbc/locations.rb
@@ -18,8 +18,9 @@ module Hbc::Locations
           opoo <<-EOS.undent
             The default Caskroom location has moved to #{default_caskroom}.
 
-            Please migrate your Casks to the new location, or if you would like to keep your
-            Caskroom at #{legacy_caskroom}, add the following to your HOMEBREW_CASK_OPTS:
+            Please migrate your Casks to the new location and delete #{legacy_caskroom},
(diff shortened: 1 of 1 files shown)


adidalal:
LGTM

joewiz:
As the OP states in #21913 says, it would be great to give crystal clear, fool proof directions for migrating to the new location. The discussion in that issue gave some ideas, but even the one that seemed to garner the most praise contained a syntax error. I've summarized the approach and corrected the code in the quoted portion below. Would you consider adding this?

> To move your Casks to the new, recommended location, use this command:
> `mv /opt/homebrew-cask/Caskroom /usr/local`
> Please note that any casks installed prior to #13966 were symlinked rather than moved to their destinations, so you'll need to re-install those manually. To simply reinstall all Casks, use this command:
> `for cask in $(brew cask list); do brew cask install $cask --force; done;`

vitorgalvao:
> To simply reinstall all Casks, use this command:
`for cask in $(brew cask list); do brew cask install $cask --force; done;`

As has been said repeatedly, [`brew cask list` is broken](https://github.com/caskroom/homebrew-cask/issues/14058), you cannot rely on it. There’s no simple foolproof command to this migration. Best we can do is give guidelines, which is what we already do.

Issue #18831, opened 2016-02-13
Implement `--require-sha` flag


vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/issues/10380.

When using `--require-sha` on the command line (or `HOMEBREW_CASK_OPTS`) during an `install`, before starting the download HBC must check if the cask in question uses `sha256 :no_check`. If it does, installation will abort for that specific cask and an explanatory message be given.

Steps for implementation ([from the next post](https://github.com/caskroom/homebrew-cask/issues/18831#issuecomment-184514730)):

- [x] Add test for `--require-sha` to [`test/cask/cli/options_test.rb`](https://github.com/caskroom/homebrew-cask/blob/master/test/cask/cli/options_test.rb)
- [x] Add option `--require-sha` to [`lib/hbc/cli.rb`](https://github.com/caskroom/homebrew-cask/blob/master/lib/hbc/cli.rb#L147)
- [x] Add new error `Hbc::CaskNoShasumError` to [`lib/hbc/exceptions.rb`](https://github.com/caskroom/homebrew-cask/blob/master/lib/hbc/exceptions.rb)
- [x] Check if `sha256 :no_check` in [`lib/hbc/cli/install.rb`](https://github.com/caskroom/homebrew-cask/blob/master/lib/hbc/cli/install.rb), and abort if `--require-sha` is present
- [x] Update documentation ([USAGE.md](https://github.com/caskroom/homebrew-cask/blob/master/USAGE.md#options), at a minimum)
- [x] Update man page

adidalal:
Steps for implementation:

- [ ] Add test for `--require-sha` to [`test/cask/cli/options_test.rb`](https://github.com/caskroom/homebrew-cask/blob/master/test/cask/cli/options_test.rb)
- [ ] Add option `--require-sha` to [`lib/hbc/cli.rb`](https://github.com/caskroom/homebrew-cask/blob/master/lib/hbc/cli.rb#L147)
- [ ] Check if `sha256 :no_check` in [`lib/hbc/cli/install.rb`](https://github.com/caskroom/homebrew-cask/blob/master/lib/hbc/cli/install.rb), and abort if `--require-sha` is present
- [ ] Update documentation ([USAGE.md](https://github.com/caskroom/homebrew-cask/blob/master/USAGE.md#options), at a minimum)

jawshooah:
The man page will also need to be updated. Other than that, I believe that should be sufficient.

[...]

PR #22398, opened 2016-06-25
Add --require-sha option for Issue #18831

joshka:

Implementation checklist from the issue #18831:
- [x] Add test for --require-sha to test/cask/cli/options_test.rb
- [x] Add option --require-sha to lib/hbc/cli.rb
- [x] Add new error Hbc::CaskNoShasumError to lib/hbc/exceptions.rb
- [ ] Check if sha256 :no_check in lib/hbc/cli/install.rb, and abort if --require-sha is present
- [x] Update documentation (USAGE.md, at a minimum)
- [x] Update man page

Regarding item 4, the change was to installer.rb instead of install.rb.
I also added an option to options.rb. I wasn't quite sure if this was the correct implementation.

### DIFF
--- a/lib/hbc/exceptions.rb
+++ b/lib/hbc/exceptions.rb
@@ -153,3 +153,17 @@ class Hbc::CaskSha256MismatchError < RuntimeError
       EOS
   end
 end
+
+class Hbc::CaskNoShasumError < Hbc::CaskError
+  attr_reader :token
+  def initialize(token)
+    @token = token
+  end
+
+  def to_s
+    <<-EOS.undent
+      #{token} does not have a sha256 checksum defined and was not be installed.
+      This means you have the "--require-sha" option set, perhaps in your HOMEBREW_CASK_OPTS.
+      EOS
+  end
+end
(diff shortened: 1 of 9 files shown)


jawshooah:
You need to run `bundle exec rake man` to update the generated man page along with the Markdown source.

joshka:
Man page updated

jawshooah:
LGTM, will wait a few hours for other maintainers to chime in before merging.

lib/hbc/exceptions.rb
```diff
@@ -153,3 +153,14 @@ def to_s
...
 end
+
+class Hbc::CaskNoShasumError < Hbc::CaskError
+  attr_reader :token
+  def initialize(token)
+    @token = token
+  end
+
+  def to_s
+    "Cask '#{token}' does not have a sha256 checksum defined and could not be installed."
```
vitorgalvao: `could not be installed` suggests it was due to us. I’d prefer something like `was not installed`. Also, It’s probably a good idea to inform the users why, lest they forget they added the flag (or blindly copied it from somewhere on the web).

Something like:

```ruby
#{token} does not have a sha256 checksum defined and was not be installed.
This means you have --require-sha set, perhaps in your HOMEBREW_CASK_OPTS.
```

joshka: Fixed

jawshooah:
Thank you for the contribution. It needed some fixes, so they were made in ff81663f5611c9746e51bb4a8e5b86741f926ce0. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #22421, opened 2016-06-27
Updated chefdk (MacOS.release support)

xakraz:
Chef provides various URL for the ChefDK according to the version of
macOS: https://downloads.chef.io/chef-dk/mac/

Notes: For now, it seems to be the same artefact despite the different
URLs (The SHA256 are the same). However it is a step to support various
builds in the future.

### DIFF
--- a/Casks/chefdk.rb
+++ b/Casks/chefdk.rb
@@ -2,7 +2,7 @@ cask 'chefdk' do
   version '0.15.15-1'
   sha256 '44adc9519697a468e2cf12322facf1f766dc832b60c58029696a88c135a3b58e'
 
-  url "https://packages.chef.io/stable/mac_os_x/10.11/chefdk-#{version}.dmg"
+  url "https://packages.chef.io/stable/mac_os_x/#{MacOS.release}/chefdk-#{version}.dmg"
   name 'Chef Development Kit'
   name 'ChefDK'
   homepage 'https://downloads.chef.io/chef-dk/'


vitorgalvao:
If it’s the same even though the URL changes, then there is no advantage to making this change, but there’s a disadvantage to having it, as it might break in the future.

In addition, there was a [pull request template](https://github.com/blog/2111-issue-and-pull-request-templates) when you submitted. It must be filled.

If you were not presented a template, you may have submitted via a CLI tool (likely [`hub`](https://hub.github.com/)). It that was the case, you may wish to see if the tool has a way to let you use the template (`hub` [is working on it](https://github.com/github/hub/issues/1118)).

xakraz:
I do not completely agree with you @vitorgalvao.

It is disturbing for me to download a ChefDK artefact with an URL mentioning 10.11 while I am on 10.10 and that I know that Chef.io is providing various URLs for each version of macOS (Very obvious when you go on https://downloads.chef.io/chef-dk/mac/).

So, OK at the end, the result is the same but that is because Chef gives the same artefact and because I carefully checked the Shasum for each versions. From a logical point of view, since Chef.io is supporting various URL for each verion of macOS, it is better to follow it in the formula.

Your point of  "there’s a disadvantage to having it, as it might break in the future" is not really receivable. If I wanted to be sarcastic I would say that everything might break in the future and even more if we arbitrarily chose to ignore a capability that is provided and fix the version of macOS.

Regarding the pull-request template, my bad. I used indeed `hub` and I will be more than happy to update my PR to be compliant with this template. I would also suggest you to make its existence more obvious in https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md

vitorgalvao:
> Your point of "there’s a disadvantage to having it, as it might break in the future"

Wrong. My point was “there’s a disadvantage **with no advantage**”. Please don’t pick and choose parts of my argument.

> If I wanted to be sarcastic I would say that everything might break in the future

If I wanted to be sarcastic, I might say I’ve been with this project almost every day since close to its inception, and I’m exceedingly aware of the common ways casks break. The argument “everything might break someway someday” isn’t valid; “something is likelier to break because of this” is.

> I would also suggest you to make its existence more obvious in https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md

If people read CONTRIBUTING.md, github wouldn’t have implemented the auto-show-template feature in the first place. It was a direct response to [Dear GitHub](https://github.com/dear-github/dear-github).

> It is disturbing for me to download a ChefDK artefact with an URL mentioning 10.11 while I am on 10.10 

In the end, I do appreciate your point, but it is not strong enough. You’re the only person I ever remember having this issue, *and you already clarified your problem*. Occasionally we do get other users like you trying to fix a problem only they had. If there’s a disadvantage to your method and no tangible advantage (only you would benefit, and even now you wouldn’t because you already no the reason), there’s no point.

xakraz:
Fair enough and sorry for the tone of my last answer.

vitorgalvao:
> Fair enough and sorry for the tone of my last answer.

Thank you.

PR #22464, opened 2016-06-28
Core Code Cleanup

reitermarkus:
I'm continuing with https://github.com/caskroom/homebrew-cask/issues/22038 here.

Started by running `rubocop --auto-gen-config` (default configuration) and step-by-step cleaning up parts of the code.

If there are some specific `rubocop` settings I should adhere to please let me know.


lib/hbc/artifact/uninstall_base.rb
```diff
@@ -1,32 +1,31 @@
...
 
 class Hbc::Artifact::UninstallBase < Hbc::Artifact::Base
-
   # todo: 500 is also hardcoded in cask/pkg.rb, but much of
   #       that logic is probably in the wrong location
 
   PATH_ARG_SLICE_SIZE = 500
 
   ORDERED_DIRECTIVES = [
-                        :early_script,
```
vitorgalvao: This one should remain as was, probably (consistent to how it works with casks themselves).

lib/hbc/dsl/license.rb
```diff
@@ -1,37 +1,36 @@
 class Hbc::DSL::License
-
   # a generic category can always be given as a license, so
   # category names should be given as both key and value
   VALID_LICENSES = {
-                    # license         category
```
vitorgalvao: Same in this file regarding indentation.

lib/hbc/macos/release.rb
```diff
@@ -7,20 +7,20 @@ class Release
     include Comparable
 
     SYMBOLS = {
-               :sierra        => '10.12',
```
vitorgalvao: Also here (and you got the idea).

vitorgalvao:
I really like this. The only thing that worries me is that much of it might be undone / become inconsistent after we merge with HB. For instance, I’m a big fan of single quotes when no interpolation is needed, but last I looked at HB’s code (admittedly a good while ago) they preferred double quotes everywhere.

Pinging @jawshooah and @claui for some opinions, as well as @anastasiasulyagina since she’s doing the merger work. At this point, I want to be out of her way as much as possible and not make big changes like this that might undo her work.

lib/hbc/macos/release.rb
reitermarkus: Yes, got it. 😄

reitermarkus:
> I’m a big fan of single quotes when no interpolation is needed

I am too, but I hadn't thought about the merger before so I would say consistency with the `homebrew` core is a bigger priority.

> I want to be out of her way as much as possible and not make big changes like this that might undo her work.

From what I have seen so far, [AnastasiaSulyagina/brew/tree/migration-logic](https://github.com/AnastasiaSulyagina/brew/tree/migration-logic) is seperate from [AnastasiaSulyagina/brew/tree/cask-merge](https://github.com/AnastasiaSulyagina/brew/tree/cask-merge), which would only need to be “rebased”, at least in its current state. @AnastasiaSulyagina, can you confirm this, or have I overlooked something?

jawshooah:
@reitermarkus Thanks for kicking this off! You should use the [rubocop-cask default config](https://github.com/caskroom/rubocop-cask/blob/a69577afc06055bb3d41344c32d75332d9e01a65/config/default.yml#L21-L68) as a baseline.

@vitorgalvao Regarding indentation of hashes and arrays, the one-space indent is no longer present in any Casks since the introduction of rubocop-cask.

Looking at @AnastasiaSulyagina's `cask-merge` branch, it doesn't seem like any changes have been made to the code since it was copied over, so this shouldn't be an issue if it's merged soon.

AnastasiaSulyagina:
@reitermarkus @jawshooah, I decided to first make changes to cask code itself and after that move it to brew so cask-merge branch on brew is not the one to look at. So for now your changes will not do any harm here as I'm changing little parts now and will be able to merge changes easile if any.

vitorgalvao:
> Regarding indentation of hashes and arrays, the one-space indent is no longer present

Ah, yes, I wasn’t even thinking about the one space indent. I was referring to indenting with the braces. This

```ruby
thing = [
          :other_thing
        ]
```

Instead of

```ruby
thing = [
  :other_thing
]
```

reitermarkus:
> the one-space indent is no longer present

👍 I always found the one-space indent a bit odd.

jawshooah:
Could you include your `.rubocop.yml` and `.rubocop_todo.yml` as well?

jawshooah:
[**@reitermarkus**](https://github.com/reitermarkus) commented on [Jul 1, 2016, 4:40 PM EDT](https://github.com/caskroom/homebrew-cask/pull/22464#commitcomment-18101489 "2016-07-01T20:40:33Z - Replied by Github Reply Comments"):
> Sorry, I overlooked this. 
> 
> Should I remove
> 
> ```yaml
> UseHashRocketsWithSymbolValues: true
> PreferHashRocketsForNonAlnumEndingSymbols: true
> ```
> 
> too, or do you agree that these are more readable?

Yes, please remove those as well.

jawshooah:
Also, please add `run bundle exec rake rubocop` to [this block in `ci/travis/script.sh`](https://github.com/caskroom/homebrew-cask/blob/b9b1a7f8072ebb50e3e236605ea376ceada81a7a/ci/travis/script.sh#L21).

lib/hbc/dsl/base.rb
```diff
@@ -27,4 +27,11 @@ def staged_path
   def appdir
     Hbc::DSL.appdir
   end
+
+  def method_missing(method, *args)
```
jawshooah: You can satisfy the `UnusedMethodArgument` cop by changing `args` to `_args`

jawshooah:
Looks like something in Pt 49 or 50 broke the build.

lib/hbc/locations.rb
```diff
@@ -131,8 +131,7 @@ def path(query)
                          query
                        else
                          all_tokens.detect do |tap_and_token|
-                           user, repo, token = tap_and_token.split('/')
-                           token == query
+                           tap_and_token.split('/')[2] == query
```
reitermarkus: I tried `tap_and_token.end_with?(query)`. It should work, but I don't get why it breaks the build. Any idea?

reitermarkus:
Ok, this should be all for now.

jawshooah:
Thank you for the contribution. It needed some fixes, so they were made in 11eedac59927269a37dd92ddb98831f8be367f0c. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

jawshooah:
Other core PRs created some merge conflicts, so I resolved them myself with `prfixmaster`. Thanks for all your hard work, @reitermarkus 😄

PR #22475, opened 2016-06-29
Fix Download URI

Labutin:
#### Editing an existing cask
- [ ] Fixed download URI for mongochef 4.0.1
- [ ] `brew cask audit --download mongochef` is error-free.
- [ ] `brew cask style --fix mongochef` left no offenses.


vitorgalvao:
In the future, at the very least you need to include the cask name in the commit message

PR #22524, opened 2016-07-01
README.md: add a gif showing a typical install

vitorgalvao:


adidalal:
I suggest https://asciinema.org/

ie. https://asciinema.org/a/e5geldgvynjmf8f6uh2b8n74j is one for brew-cask-upgrade (by buo)

vitorgalvao:
I dislike asciinema for this. It cannot play directly in the README (yes, we can make a gif of it, but at that point, there’s no difference to a screen recording) and we lose the Finder, which I find a really important detail, here, I really want the user to feel how quick GUI changes happen with a CLI command (those have to be imagined with ascinema, but are explicitly showed with a screen recording).

adidalal:
Gotcha. Haven't experimented with, but have heard good things about http://www.cockos.com/licecap/ if you want to compare gif output/filesize.

Else, the original gif LGTM. Was just looking to optimize for mobile/low-bandwidth

vitorgalvao:
> I normally use imageoptim on the final output image

Haven’t done so yet, but I should. Will update shortly.

Has for the gif creation itself, I’ve got that covered, and I must’ve obsessively explored all methods in existence, already. I even made two scripts for it (both in `vitorgalvao/tiny-scripts`) in the past: [`gifmaker`](https://github.com/vitorgalvao/tiny-scripts/blob/master/gifmaker) to make a gif from a selection of images (I use [processing](https://processing.org/) to make an animation exported as a series of frames) and [`gfv`](https://github.com/vitorgalvao/tiny-scripts/blob/master/gfv) (“gif from video”) to make a gif from a video file.

`ffmpeg` can also be used to make a gif from a sequence of images, or to export a video as frames (useful for some manual editing). [Drop to GIF](https://mortenjust.github.io/droptogif/) (`brew cask install drop-to-gif`) is also a simple and good-looking tool to convert videos to gifs, with good results.

I don’t recall exactly why I did not like LICEcap, but I think it was due to a lack of precision when setting up the area to record.

Currently, my preferred method is to record with [ScreenFlow](http://www.telestream.net/screenflow/) (`brew cask edit screenflow`) as it gives a a lot of editing control lacking in most screen recording methods. For example, in the gif in this PR I accelerated the parts that took longer: typing the command is about twice as fast, and the download was accelerated 1300%. I then export the video and convert to gif with something like [`gifify`](https://github.com/jclem/gifify) (`brew install gifify`). At this point it may be worth trying the conversion with different tools. If you obsess about the image quality, you’ll notice the differences between the outputs.

Note that ScreenFlow has recently gained the ability to export to GIF directly, in version 6. Can’t attest to it, as I’m still on version 5. The app is a bit pricey, but I won’t say the price is undeserved. I got it for cheap with a bundle with version 4, and then paid upgrade pricing for version 5. Will likely do it again for version 6. As far as I can tell, they don’t do many promotions (a small discount around Christmas, perhaps) and don’t participate in many bundles. They do seem to offer the same upgrading pricing regardless if you’re one version behind, or have just version 1 (this may be an advantage or a disadvantage, naturally).

Hope that wasn’t too boring, and that there’s some useful information there.

adidalal:
Didn't realize you had done so much research into it already, so I gladly defer to your judgement. The writeup was useful & quite comprehensive, appreciate it!

Haven't needed to make gifs too often, but will keep the above in mind, thanks!

PR #22546, opened 2016-07-03
Update rrootage URL

coverslide:
Old url no longer works


vitorgalvao:
Thank you, but doing it in https://github.com/caskroom/homebrew-cask/pull/22552. I’m also fixing the other games from that developer in the process.

PR #22547, opened 2016-07-03
Update noiz2sa URL

coverslide:
Old url no longer works

### DIFF
--- a/Casks/noiz2sa.rb
+++ b/Casks/noiz2sa.rb
@@ -2,7 +2,7 @@ cask 'noiz2sa' do
   version :latest
   sha256 :no_check
 
-  url 'https://workram.com/downloads.php?f=Noiz2sa', referer: 'https://workram.com/games/noiz2sa'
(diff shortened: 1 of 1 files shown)


Casks/noiz2sa.rb
```diff
@@ -2,7 +2,7 @@
   version :latest
   sha256 :no_check
 
-  url 'https://workram.com/downloads.php?f=Noiz2sa', referer: 'https://workram.com/games/noiz2sa'
+  url 'https://workram.com/downloads/Noiz2sa-for-OS-X-0.51.5.dmg', referer: 'https://workram.com/games/noiz2sa'
```
vitorgalvao: Try without `:referer`. It will likely no longer be needed.

Casks/noiz2sa.rb
```diff
@@ -2,7 +2,7 @@
   version :latest
```
vitorgalvao: `version` and `sha256` need to be set, now.

vitorgalvao:
Same as https://github.com/caskroom/homebrew-cask/pull/22546. Thank you anyway for the report.

PR #22572, opened 2016-07-04
README.md: do not show redirection in gif

vitorgalvao:
The old gif, before starting the download, would jump to 100% before starting the proper download. That is actually what happens with the cask, but it might be confusing to a new user and doesn’t actually add anything to the instructions.


fanquake:
Good idea.

PR #22574, opened 2016-07-04
Delete exceptions from homebrew-fork

AnastasiaSulyagina:
### Changes to the core

Using exceptions from homebrew now as a preparation to core merge([#14384](https://github.com/caskroom/homebrew-cask/issues/14384)).


MikeMcQuaid:
👏 It's also worth noting that when https://github.com/Homebrew/brew/pull/390 is merged we'll be verifying that Homebrew's core changes don't end up breaking `brew cask` which means it'll be safer to rely on Homebrew's internals.

Rakefile
```diff
@@ -3,6 +3,7 @@ require 'rake/testtask'
 require 'rspec/core/rake_task'
 require 'rubocop/rake_task'
 
+$LOAD_PATH.unshift(File.expand_path('../../../../Homebrew', __FILE__))
```
jawshooah: There's no guarantee that tests are being run from the tap repo. For example, I have a separate clone in a common directory for Git projects where I do most of my development, so this wouldn't work for me.

Better, I think, to do this:

```ruby
homebrew_repo = `brew --repository`
$LOAD_PATH.unshift(File.expand_path("#{homebrew_repo}/Library/Homebrew"))
```

spec/spec_helper.rb
```diff
@@ -2,6 +2,9 @@
 require 'rspec/its'
 require 'rspec/wait'
 
+homebrew_library_path = Pathname.new(File.expand_path(__FILE__+'/../../../../../Homebrew'))
```
jawshooah: Same problem here

test/test_helper.rb
```diff
@@ -2,6 +2,9 @@
 require 'bundler/setup'
 require 'pathname'
 
+homebrew_library_path = Pathname.new(File.expand_path(__FILE__+'/../../../../../Homebrew'))
```
jawshooah: And here

jawshooah:
Great work, thanks!

Rakefile
luchiano1: 1

spec/spec_helper.rb
luchiano1: 1

test/test_helper.rb
luchiano1: 1

luchiano1:
1

PR #22585, opened 2016-07-04
Fix `url` stanza's comments.

reitermarkus:


adidalal:
Went through changes, all look great. Thanks for everything!

adidalal:
@reitermarkus Didn't see that you had more changes - would you mind adding them in a new PR? (everything after `Fix `url` stanza comment for epub-2-pdf.`)

reitermarkus:
@adidalal, I don't think I had any more changes after `epub-2-pdf`.

adidalal:
Excellent, then with the 3 fixes I just pushed, https://github.com/caskroom/rubocop-cask/pull/26 should now be good to go

PR #22606, opened 2016-07-05
Update macpaw-gemini  to "gemini2" (2.1.0,1464094756)

toonetown:

Moved the original macpaw-gemini cask to caskroom/homebrew-versions as `macpaw-gemini1` in [PR 2298](https://github.com/caskroom/homebrew-versions/pull/2298).


vitorgalvao:
Everywhere there’s a `2` from `Gemini2` or Gemini 2`, it should instead be `#{version.major}`.

toonetown:
I can make that change - but they are actually calling the app "Gemini2" (with the 2 being an exponent).  I don't know that there is any guarantee that the name of the package (`com.macpaw.site.Gemini2`) is going to follow the major version naming convention.

Though I guess that if it doesn't, that will still require a change - and if they happen to follow that convention, then it won't require a change.

I'll make the change.  😄

toonetown:
Actually, when I change the download url to read:

```
  url "https://dl.devmate.com/com.macpaw.site.Gemini#{version.major}/#{version.before_comma}/#{version.after_comma}/Gemini#{version.major}-#{version.before_comma}.zip"
```
I get an error when running `brew cask style`:

```
C:  5:  3: devmate.com/com.macpaw.site.Gemini2 does not match dl.devmate.com/com.macpaw.site.Gemini
```

toonetown:
Looks like that issue is something to do with the "xxx was verified as official" comment immediately prior.  I can fix the issue by setting the comment to read:
```
# devmate.com/com.macpaw.site.Gemini was verified as official when first introduced to the cask
  url "https://dl.devmate.com/com.macpaw.site.Gemini#{version.major}/#{version.before_comma}/#{version.after_comma}/Gemini#{version.major}-#{version.before_comma}.zip"
```

Or, I can get it to pass (and have an "accurate" comment) by setting it as:
```
  # devmate.com/com.macpaw.site.Gemini2 was verified as official when first introduced to the cask
  url "https://dl.devmate.com/com.macpaw.site.Gemini2/#{version.before_comma}/#{version.after_comma}/Gemini#{version.major}-#{version.before_comma}.zip"
```

Which is the preferred style?

adidalal:
Looks like it's referred to as "Gemini 2" everywhere.
Will actually change this one, as there are a few changes I want to make

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 8a3f9ab9325c865ae7bf7a9201829e6438cc31b2. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

adidalal:
@vitorgalvao In line with other casks that have `# note: "2" is not a version number, but an intrinsic part of the product name`, I am *not* using `#{version.major}` in place of `2` as you suggested above. It just reduces readability for no real gain in update url flexibility.

vitorgalvao:
I agree with your logic, but disagree that this one warrants the `# note: "2" is not a version number, but an intrinsic part of the product name`.

`2` is very much a version number, not part of the name. Just like Alfred refers to itself as Alfred 3 everywhere (even when calling it via AppleScript, you need to include the number), the 3 is a version number. One they proeminently show, but a version number nonetheless.

Why do you feel this is different?

adidalal:
Ah I see, looked at Alfred and that does seem more appropriate for the current situation. Was still thinking about `iterm2` subliminally, probably.

Even `alfred` doesn't use `#version.major` everywhere (which I agree with), but I don't think that's a preference that's actually been written down.

Changed the name back to `gemini`, removed the comment, and updated the app stanza in https://github.com/caskroom/homebrew-cask/commit/caf52e3c6085c05178d7508f1aeddabd4b50a693

PR #22623, opened 2016-07-06
"staged at" message changed to "was successfully installed!"

vitorgalvao:
### Changes to the core

---

A user was just confused by this on Gitter. I agree it’s a bit of a weird message, especially in cases like atom where it says it was staged and then points out `(0B)`. `support files` seems to me generic enough to cover any cases, at the same time making clear you should likely not have to worry about it. It’s also arguable if the message should exist at all.

Submitting this as a suggestion to gather opinions/better solutions. Will only change the gif after the final decision (if we decide to change this).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ complete
 ==> Moving App 'Atom.app' to '/Applications/Atom.app'
 ==> Symlinking Binary 'apm' to '/usr/local/bin/apm'
 ==> Symlinking Binary 'atom.sh' to '/usr/local/bin/atom'
-🍺  atom staged at '/usr/local/Caskroom/atom/1.8.0' (0B)
+🍺  atom was successfully installed!
 ```
 
 And there we have it. Atom installed with one quick command: no clicking, no dragging, no dropping.
--- a/USAGE.md
+++ b/USAGE.md
@@ -42,7 +42,7 @@ The command `brew cask install` accepts a Cask token as returned by `brew cask s
 $ brew cask install google-chrome
 ==> Downloading https://dl.google.com/chrome/mac/stable/GGRO/googlechrome.dmg
 ==> Moving App 'Google Chrome.app' to '/Applications/Google Chrome.app'
-🍺  google-chrome staged at '/usr/local/Caskroom/google-chrome/latest' (3 files, 288K)
+🍺  google-chrome was successfully installed!
 ```
 
 ## Uninstalling Casks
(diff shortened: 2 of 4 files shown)


adidalal:
I'm fine with removing the message altogether, actually.

reitermarkus:
Maybe it would make sense to only output the message if there actually are files in the staging directory.

vitorgalvao:
> Maybe it would make sense to only output the message if there actually are files in the staging directory.

Sure, but most apps only have a README of something really unimportant staged in that directory, so claiming they are simply support files still seems helpful. I’ll also say I’m thinking more and more the message should just be removed outright.

reitermarkus:
> so claiming they are simply support files still seems helpful.

Yes, I should have clarified that I'm not opposed to changing the message text.

vitorgalvao:
@reitermarkus What about keeping vs removing?What are your thoughts on that?

reitermarkus:
I would still like to be informed of any “support files” being left in the staging directory, even if it's just to make the installation seem “complete”.

jawshooah:
Also keep in mind that `binary` Casks are symlinked, so the actual source remains in the staging directory.

vitorgalvao:
I’m not contesting the usefulness of the staging directory, what I’m wondering is if informing the user every single time is of any value. Even if there are any binaries, why would users care? All they care is if everything is installed correctly. If it is, the staging directory is irrelevant *and they’ll never have to touch it anyway*.

The only users I envision the staging directory being relevant to are the tinkerers that can figure it out anyway (by coming here and reading the documentation). In HBC’s normal usage, does the message matter? I’d argue it likely doesn’t. Seems like we’re keeping it because it was always there.

vitorgalvao:
So, just for a quick vote so we can decide what to do with this PR, mind adding a reaction to this post?

👍 = Delete the message completely.
👎 = Do nothing, keep everything as is currently.
😆 = Keep the message, but as `supporting files` (i.e. merge this PR).

jawshooah:
Maybe rather than deleting the message or using `supporting files`, we could replace it with something like `🍺 Cask {{token}} successfully installed!`

vitorgalvao:
@jawshooah I’m not opposed in the slightest. Lets include that in the 👍  as well, then.

adidalal:
Yeah, +1 for that last suggestion by @jawshooah

fanquake:
So this is just waiting on an addition from @vitorgalvao ?

vitorgalvao:
Made the change and the build is passing. If in agreement, it can be merged.

reitermarkus:
Is the code that calculates `(0B)` dead then, or is this also used in `info`?

vitorgalvao:
> Is the code that calculates `(0B)` dead then, or is this also used in `info`?

It’s still used in `info` (seems like a good place to continue to output all the information).

adidalal:
LGTM

PR #22626, opened 2016-07-06
Add conflicts_with intellij-idea-eap stanza to intellij-idea

metyl:
#### Editing an existing cask

RC build of Intellij IDEA 2016.2 and the current version of
IntelliJ IDEA both have the same app name.

See also caskroom/homebrew-versions#2302.

### DIFF
--- a/Casks/intellij-idea.rb
+++ b/Casks/intellij-idea.rb
@@ -8,6 +8,7 @@ cask 'intellij-idea' do
   license :commercial
 
   auto_updates true
+  conflicts_with cask: 'intellij-idea-eap'
 
   app 'IntelliJ IDEA.app'
 


vitorgalvao:
So is there a conflict [or not](https://github.com/caskroom/homebrew-cask/pull/22104/files)?

vitorgalvao:
So in this case, the `conflicts_with` should not be removed again. It makes no sense to keep removing and readding.

metyl:
@vitorgalvao it depends, with release of Intellij IDEA 2016.2 RC, they changed the app name from `IntelliJ IDEA 2016.2 EAP.app` to `IntelliJ IDEA.app` and now we have a conflict.
 Then they start EAP for next version - `2016.3` probably the app name will be changed to `IntelliJ IDEA 2016.3 EAP.app` and there won't be any conflict.

We can:
* keep `conflicts_with` stanza here but it will not always be true (most of the time it won't be true and will be misleading)
* remove `conflicts_with` stanza and get conflict from time to time - short time just before release (user will have to install with `--force` to replace currently installed version)
* maintain it (keep removing and readding `conflicts_with` stanza) based on what are the app names of current and next version.

vitorgalvao:
`conflicts_with` doesn’t even work yet, and this has never been a problem. If the conflict only happens occasionally, then yes, lets remove it.

A conflict happens when apps clash. If these clash only in the sense they share a name, it’s not really a clash.

metyl:
ok, so You can revert this pull request and I will made one in  [caskroom/homebrew-versions](http://github.com/caskroom/homebrew-versions) to also remove `conflicts_with` stanza.

vitorgalvao:
Sounds like the best option. Please do.

metyl:
@vitorgalvao To complete this task will You revert this merge or should I make PR for this?

vitorgalvao:
I can revert.

PR #22664, opened 2016-07-08
Update repo name in lib/hbc/utils.rb

szhu:
### Changes to the core

---

For non-ancient installs of HBC, `brew untap phinze/cask` does not remove casks but `brew untap caskroom/cask` does.

Do we need to include both commands in case someone has a _really_ old installation?

### DIFF
--- a/lib/hbc/utils.rb
+++ b/lib/hbc/utils.rb
@@ -6,7 +6,7 @@ require "stringio"
 
 require "hbc/utils/tty"
 
-UPDATE_CMD = "brew uninstall --force brew-cask; brew untap phinze/cask; brew update; brew cleanup; brew cask cleanup".freeze
+UPDATE_CMD = "brew uninstall --force brew-cask; brew untap phinze/cask; brew untap caskroom/cask; brew update; brew cleanup; brew cask cleanup".freeze
 ISSUES_URL = "https://github.com/caskroom/homebrew-cask#reporting-bugs".freeze
 
 # TODO: temporary


vitorgalvao:
The point of the command is not to remove the current instalation (which your new command would) but to get rid of all the old ones that may be lingering.

szhu:
I find that in some cases (if the cask files or git repo are corrupt), removing the current installation and adding it back is required for fixing issues.

I submitted this PR specifically because I ran into an issue and tried the command given and it did not work fix the problem. Untapping caskroom/cask (and letting the next invocation of `brew cask` re-install it) did the trick.

jawshooah:
How about a compromise: `... brew untap phinze/cask; brew untap caskroom/cask; ...`

This is what we recommend in [pre_bug_report.md](https://github.com/caskroom/homebrew-cask/blob/c7fbf76e60d1f085faf68269a6811eceb5eb7a55/doc/reporting_bugs/pre_bug_report.md#L12).

szhu:
That's what I meant by "include both commands". I'm totally for this!

vitorgalvao:
Didn’t realise we had both in the `pre_bug_report`. Thought we only had `phinze`, and was on a phone, so did not check.

> I submitted this PR specifically because I ran into an issue and tried the command given and it did not work fix the problem. Untapping caskroom/cask (and letting the next invocation of `brew cask` re-install it) did the trick.

I’m all for changes that lead to less bug reports and more user-fixing. If that command can also fix things, then by all means, lets add both.

PR #22669, opened 2016-07-08
Move brew-cask formula to Formula directory

jawshooah:
### Changes to the core

This allows us to enforce hash-rocket syntax for `brew-cask.rb`.


DomT4:
Just a heads-up Homebrew handles formula _moving_ within an existing tap kind of badly. This happens:

```
Updated 1 tap (caskroom/cask).
==> New Formulae
caskroom/cask/brew-cask
```

Not a big deal for single formula, but it can look a bit odd & someone may raise a question about it being tagged "new".

PR #22680, opened 2016-07-09
strategies moved

AnastasiaSulyagina:
### Changes to the core
## 

Moved download-strategies from hb-fork to main utils file. Merged hb's strategies to cask's when needed.
(Could not make them use hb's code as the abstract class is slightly different so decided to just merge like this, correct me if I'm wrong)

cc @MikeMcQuaid


jawshooah:
RuboCop caught some style nits that we ignored in #22464 since the code was vendored. `bundle exec rubocop --auto-correct` should fix most of them for you, and the rest should be straightforward to fix manually.

jawshooah:
Thank you!

MikeMcQuaid:
:metal:

PR #22681, opened 2016-07-09
Puppet Agent: Fix failed test on Sierra.

reitermarkus:
#### Editing an existing cask
  - fails with error: download not possible

### DIFF
--- a/Casks/puppet-agent.rb
+++ b/Casks/puppet-agent.rb
@@ -1,11 +1,11 @@
 cask 'puppet-agent' do
   version '1.5.2-1'
 
-  if MacOS.release == :el_capitan
-    sha256 '67613f23e6e7708fd1f8ddaa8528260cce176c4d46d7c3d666c236ea0edd1c8c'
+  if MacOS.release == :mavericks
+    sha256 'bc75b261b2c1e5ed18599e63848e22fd4fca440242ba31391617a9a59806ef7a'
   elsif MacOS.release == :yosemite
     sha256 'ba9d22b4992f96f6e100e689f92f81e4cabb0459730459d9b53fe54c59b392a4'
-  elsif MacOS.release == :mavericks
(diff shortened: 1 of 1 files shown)


Casks/puppet-agent.rb
```diff
@@ -15,6 +15,12 @@
   homepage 'https://docs.puppet.com/puppet/4.5/reference/about_agent.html'
   license :oss # all Apache 2 except for the vendored OpenSSL + Ruby
 
+  depends_on macos: %w[
```
jawshooah: This won't work as you expect. For an array of symbols, you want `%i`. Or just use a regular array literal.

Casks/puppet-agent.rb
```diff
@@ -1,11 +1,11 @@
 cask 'puppet-agent' do
   version '1.5.2-1'
 
-  if MacOS.release == :el_capitan
+  if MacOS.release == :mavericks
     sha256 '67613f23e6e7708fd1f8ddaa8528260cce176c4d46d7c3d666c236ea0edd1c8c'
```
jawshooah: The `sha256` for Mavericks was updated incorrectly in d906041e850a76a2469f1f34da6f5688a235560d. Should be `bc75b261b2c1e5ed18599e63848e22fd4fca440242ba31391617a9a59806ef7a`.

Casks/puppet-agent.rb
reitermarkus: Well, you started it: https://github.com/caskroom/homebrew-cask/commit/0e21ef176d366682ca2dfa3b6bcef17164823f9b :smile: 

Maybe this is because `MacOs::Release#<=>` expects a string?

reitermarkus: Ok, it works either way, but of course a simple array is cleaner.

PR #22703, opened 2016-07-10
Fix symbol arrays.

reitermarkus:
#### Editing an existing cask

### DIFF
--- a/Casks/deeper.rb
+++ b/Casks/deeper.rb
@@ -16,15 +16,15 @@ cask 'deeper' do
   license :gratis
 
   # Unusual case: The software will stop working, or is dangerous to run, on the next macOS release.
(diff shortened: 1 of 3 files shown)


reitermarkus:
Just noticed that `git-annex` is also available as a Formula. The only difference is that the GUI version offers a GUI launcher for the web app, which can also be launched via `git annex webapp`. Is this enough of a difference to keep it as a Cask?

vitorgalvao:
> Is this enough of a difference to keep it as a Cask?

I’m fine with removing it.

adidalal:
+1 for removal. They're functionally identical

reitermarkus:
Created a seperate PR for the removal.

PR #22709, opened 2016-07-10
Create betterzip-latest.rb

My2ndAngelic:
#### Adding a new cask
- [x] `brew cask audit --download betterzip-latest` is error-free.
- [x] `brew cask style --fix betterzip-latest` left no offenses.
- [x] `brew cask install betterzip-latest` worked successfully.
- [x] `brew cask uninstall betterzip-latest` worked successfully.

Comparing to betterzip cask:
- Remove sha256, remove version check
- Download BetterZip straightaway from the home page
- Add depends: only support mavericks and above
- To avoid conflict with the current betterzip cask, I add a new cask

### DIFF
--- a/Casks/betterzip-latest.rb
+++ b/Casks/betterzip-latest.rb
@@ -1,0 +1,15 @@
+cask 'betterzip-latest' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://macitbetter.com/BetterZip.zip'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
We don’t accept these types of duplicates. We also prefer versioned casks, because most of our users do.

PR #22725, opened 2016-07-11
Refactor Artifacts

reitermarkus:
Noticed that the `load_specification` method is exactly the same for `Hbc:Artifact:Moved` and `Hbc:Artifact:Symlinked`. Also moved the common code of `Hbc:Artifact:Hardlinked` and `Hbc:Artifact:Symlinked` to a `Hbc:Artifact:Linked` parent class, as I think it is more logical for them to be siblings. And I re-added the `add_altname_metadata` method to `Hbc:Artifact:Moved`.


jawshooah:
Nicely done, thanks!

PR #22726, opened 2016-07-11
asepsis: discontinued

vitorgalvao:
#### Editing an existing cask


xenyz:
Without a note such as in my PR, this will not install on El Capitan and later, and will appear broken. IMHO it's a good idea to at least give a heads-up on how to get it going.

vitorgalvao:
The developer specifically discourages people from installing, so this is how it’s going to stay.

[Users are responsible for knowing the software they are installing](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md) and make it work themselves.

PR #22736, opened 2016-07-12
Add Launchpad Manager 1.3.10

My2ndAngelic:
#### Adding a new cask
- [x] `brew cask audit --download launchpad-manager` is error-free.
- [x] `brew cask style --fix launchpad-manager` left no offenses.
- [x] `brew cask install launchpad-manager` worked successfully.
- [x] `brew cask uninstall launchpad-manager` worked successfully.

As you may know there is a cask named launchpad-manager-yosemite. However, it is developed for Yosemite and later users. 

So I will add a new cask to Launchpad Manager for Mavericks and prior.

### DIFF
--- a/Casks/launchpad-manager.rb
+++ b/Casks/launchpad-manager.rb
@@ -1,0 +1,13 @@
+cask 'launchpad-manager' do
+  version '1.3.10'
+  sha256 'f686a9a332663a003e9fabd32a1d44fc98debda15225368f1e8aef181955bc72'
+
+  url 'http://launchpadmanager.com/download.php/LaunchpadManager.dmg'
+  appcast 'http://launchpadmanager.com/appyos/sparkle.rss',
+          checkpoint: '9bd3cfb349d301c2e1e6316c250307d002cde68d1773daf757872f61580f9ff7'
+  name 'Launchpad Manager'
+  homepage 'http://launchpadmanager.com/'
+  license :commercial
+
+  app 'Launchpad Manager.app'
+end


Casks/launchpad-manager.rb
```diff
@@ -0,0 +1,13 @@
+cask 'launchpad-manager' do
+  version '1.3.10'
+  sha256 'f686a9a332663a003e9fabd32a1d44fc98debda15225368f1e8aef181955bc72'
+
+  url 'http://launchpadmanager.com/download.php/LaunchpadManager.dmg'
```
reitermarkus: This `url` is actually only for macOS up to Mavericks. http://launchpadmanager.com/download_yosemite.php is for Yosemite and El Capitan.

Casks/launchpad-manager.rb
```diff
@@ -0,0 +1,13 @@
+cask 'launchpad-manager' do
+  version '1.3.10'
+  sha256 'f686a9a332663a003e9fabd32a1d44fc98debda15225368f1e8aef181955bc72'
```
reitermarkus: The `url` doesn't contain a version, so you should use:

    version :latest
    sha256 :no_check

Casks/launchpad-manager.rb
vitorgalvao: Where did you find this `url`?

Casks/launchpad-manager.rb
vitorgalvao: @reitermarkus When an `appcast` is present, we like to keep the version, since appcasts allow us (will allow better, later on) to detect when it was updated (and hence not break as often).

Casks/launchpad-manager.rb
vitorgalvao: Closing, in light of this. @My2ndAngelic You may wish to add a conditional to the current cask, but we shouldn’t make a new one.

Casks/launchpad-manager.rb
reitermarkus: Ok, thanks for letting me know, I didn't know that.

PR #22749, opened 2016-07-12
ISSUE_TEMPLATE.md: inform issue may be closed if rules are not followed

vitorgalvao:
Pinging @caskroom/maintainers.

I, for one, am getting tired of people completely ignoring the guides (most of the time it’s pretty clear no attempt was made to read even part of it), even with a message asking them to when opening issues. It just goes to show that some human problems can’t be solved with technology.

I tried to be respectful with the message, and not at all forceful (if I failed, please let me know). The goal is to say “please respect our time”. If a user can’t take the time to make a correct report, I don’t see why we should take the time to solve their problem.

If accepted, will copy to the other repos.


mwean:
Sounds good to me 👍

PR #22762, opened 2016-07-13
utils and pathname moved

AnastasiaSulyagina:
### Changes to the core

---

Moved utils from hb-fork, deleted pathname, removed duplicated code from hbc/utils. 
_System methods and curl will be deleted in next PR_

cc @MikeMcQuaid


AnastasiaSulyagina:
Worked on my machine, will see what can be wrong

MikeMcQuaid:
This issue can be worked around but the main problem is the CI is running on Linux but pretending to  be a Mac. This isn't something Homebrew supports and may break at some point in future when our generic OS/Linux support improves.

https://github.com/caskroom/homebrew-cask/blob/master/.travis.yml#L13-L17
For now adding `HOMEBREW_OSX_VERSION=$OSX` to the `env` lines may be enough.

lib/hbc/utils.rb
```diff
@@ -67,6 +83,22 @@ def puts_columns(items, star_items = [])
   puts Hbc::Utils.stringify_columns(items, star_items)
 end
 
+def curl(*args)
```
MikeMcQuaid: This probably also wants a `TODO: delete`

lib/hbc/utils.rb
```diff
@@ -33,22 +33,38 @@ def tty?
 end
 
 # global methods
-
-# originally from Homebrew
-def ohai(title, *sput)
-  title = Tty.truncate(title) if $stdout.tty? && !Hbc.verbose
-  puts "#{Tty.blue.bold}==>#{Tty.reset.bold} #{title}#{Tty.reset}"
-  puts sput unless sput.empty?
+# TODO: delete
```
MikeMcQuaid: Might also want to tweak this to be `TODO: replace with Homebrew's utils version when fixed issue handling non-string args`

AnastasiaSulyagina: It's ok with non-string args already as far as I know. But I'll change it, ok.

MikeMcQuaid: Cool, sorry, that was just an example: whatever the reason is 😉

AnastasiaSulyagina:
> /Users/travis/build.sh: line 109: shell_session_update: command not found

It would be great if somebody could explain why it failed only on 10.11 after it finished with 0.

jawshooah:
See rvm/rvm#3627 and my attempt to address it in #22774.

lib/hbc/utils.rb
```diff
@@ -83,6 +83,7 @@ def puts_columns(items, star_items = [])
   puts Hbc::Utils.stringify_columns(items, star_items)
 end
 
+# TODO: delete
```
MikeMcQuaid: When to delete this one 😀

AnastasiaSulyagina: When I figure out how to reference correct curl_args. Decided that such message is not clear enough to put in comments

MikeMcQuaid: Definitely worth adding something in even if unclear. Unclear trumps empty 😉

AnastasiaSulyagina:
Saw that #22774 is merged, updated to see if checks pass

jawshooah:
Thanks!

toonetown:
For some reason, this change makes all cask `ohai` commands displayed as white for me - instead of bold black.  Reverting this change fixes the problem.

The `ohai` commands issued by homebrew formulas are still the correct color.  Just the cask ones.  You can see this in action by running `brew cask info zulu` (or any other cask that has "Caveats") and notice the incorrect color of the `Caveats` line.

jawshooah:
@toonetown Please create a separate issue. I believe I know the cause, but I'd prefer not to pollute this thread with (mostly) unrelated discussion.

PR #22790, opened 2016-07-14
Update LaTeXiT to 2.8.1

reitermarkus:
#### Editing an existing cask
- [x] Commit message includes cask’s name (and new version, if applicable).
- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.

PR #22806, opened 2016-07-14
Allow appcast to specify :user_agent

jawshooah:
### Changes to the core

Ref #22790


vitorgalvao:
We don’t need this, as in fact `audit` is the only case when this will be needed, and there’s no point in requiring users to figure out when this is needed or not. Calculating `checkpoint:` should *always* fake browser headers anyway, for it to be calculated correctly.

See https://github.com/caskroom/homebrew-cask/pull/22790#issuecomment-232815231. It was I who forgot to update the documentation. As you can see, faking the user agent is always expected, so no need for a flag.

PR #22809, opened 2016-07-14
eclipse-platform: install Platform, not SDK

ghost:
The original version of this formula installed Eclipse Platform, which is a minimal Eclipse binary with very few plugins installed. [A subsequent edit](https://github.com/caskroom/homebrew-cask/commit/44067947b891340c0c68a83d6b903f30d812951d#diff-537e6c4a35b723e8299428a00369ac97) changed this to Eclipse SDK which is Platform + a bunch of plugins, including the JDT.

This would change it back to Eclipse Platform. On two different occasions I've installed this thinking I was getting Platform, not the SDK. This is an easy mistake as "Eclipse Platform" is an actual release and `brew cask search eclipse` includes other candidates like `eclipse-java` and `eclipse-jee`, which presumably satisfy the requirements of someone looking to install Eclipse for Java development.

That said, this is a breaking change for anyone relying on this formula installing a Java IDE. Those users can switch to `eclipse-java` or `eclipse-ide`, or do as I wish to do and install the minimal platform and add only those Java development plugins they wish to have.

### DIFF
--- a/Casks/eclipse-platform.rb
+++ b/Casks/eclipse-platform.rb
@@ -1,8 +1,8 @@
 cask 'eclipse-platform' do
   version '4.6-201606061100'
-  sha256 'd8c0b7b970581bbfab047375c0bf44243f5f4e528e950be2313e275451d684b6'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Well Explained. Thank you.

PR #22855, opened 2016-07-18
Run RuboCop on all core files.

reitermarkus:
### Changes to the core

### DIFF
--- a/spec/cask/download_strategy_spec.rb
+++ b/spec/cask/download_strategy_spec.rb
@@ -1,15 +1,13 @@
-require 'spec_helper'
+require "spec_helper"
 
-describe 'download strategies' do
-  let(:url) { 'http://example.com/cask.dmg' }
+describe "download strategies" do
+  let(:url) { "http://example.com/cask.dmg" }
   let(:url_options) { Hash.new }
-  let(:cask) {
-    instance_double(Hbc::Cask,
-      :token => 'some-cask',
-      :url => Hbc::URL.new(url, url_options),
-      :version => '1.2.3.4'
-    )
-  }
+  let(:cask) do
+    instance_double(Hbc::Cask, token:   "some-cask",
+                               url:     Hbc::URL.new(url, url_options),
+                               version: "1.2.3.4")
+  end
 
   describe Hbc::CurlDownloadStrategy do
     let(:downloader) { Hbc::CurlDownloadStrategy.new(cask) }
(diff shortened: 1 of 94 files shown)


test/support/cleanup.rb
```diff
@@ -2,9 +2,7 @@ module Hbc::CleanupHooks
   def after_teardown
     super
     Hbc.installed.each do |cask|
-      Hbc::Installer.new(cask).tap do |installer|
-        installer.purge_versioned_files
-      end
+      Hbc::Installer.new(cask).tap(&:purge_versioned_files)
```
jawshooah: Pretty sure this could be simplified to `Hbc::Installer.new(cask).purge_versioned_files`

jawshooah:
I forgot about all the `let` and `expect` blocks in the specs. Let's add the following to `.rubocop.yml` and auto-correct again:

```yml
Style/BlockDelimiters:
  EnforcedStyle: semantic
```

test/cask/dsl_test.rb
```diff
@@ -12,11 +12,11 @@
     attempt_unknown_method = nil
 
     before do
-      attempt_unknown_method = lambda do
+      attempt_unknown_method = lambda {
```
reitermarkus: Not sure if this is functional.

jawshooah: It's used purely for its side-effects, so it's procedural.

jawshooah: The instances of `lambda { … }.must_raise` are similarly used purely for their side-effects, namely raising exceptions, but I'd prefer to leave their braces since chaining methods with `do … end` blocks just looks so awkward.

reitermarkus: I thought of it this way: In this case the “return value” of the lambda would be the Exception which is checked for, so it is still *kind of* functional. 😄 

But anyway, `do … end`-chaining looks awkward.

reitermarkus:
@jawshooah, is there anything else you can think of that needs to be done?

spec/cask/download_strategy_spec.rb
```diff
@@ -1,15 +1,13 @@
...
+  let(:url) { "http://example.com/cask.dmg" }
   let(:url_options) { Hash.new }
-  let(:cask) {
-    instance_double(Hbc::Cask,
-      :token => 'some-cask',
-      :url => Hbc::URL.new(url, url_options),
-      :version => '1.2.3.4'
-    )
-  }
+  let(:cask) do
```
jawshooah: Should use braces for `let`

spec/cask/download_strategy_spec.rb
```diff
@@ -33,71 +31,75 @@
...
       end
     end
 
-    context 'with cookies set' do
-      let(:url_options) {{
-        :cookies => {
-          :coo => 'kie',
-          :mon => 'ster'
+    context "with cookies set" do
+      let(:url_options) do
```
jawshooah: Should use braces for `let`

jawshooah:
Thank you for the contribution. It needed some fixes, so they were made in ed5452b7285f7701696002e3f36a315c52daae0a. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

jawshooah:
Thanks @reitermarkus! There were a bunch of leftover `let do … end` blocks, so I fixed them in https://github.com/caskroom/homebrew-cask/commit/ed5452b7285f7701696002e3f36a315c52daae0a.

PR #22857, opened 2016-07-18
Fix and refactor recognition of container types.

reitermarkus:
### Changes to the core

---

Removed the dependency on the `file` command completely, everything is handled using magic numbers now.

Also moved the `imageinfo`, `lsar` and `cabextract` methods to  container classes.

https://github.com/caskroom/homebrew-cask/pull/22002 can be closed in favor of this PR.


lib/hbc/container/generic_unar.rb
```diff
@@ -2,14 +2,19 @@
 
 class Hbc::Container::GenericUnar < Hbc::Container::Base
   def self.me?(criteria)
-    !criteria.lsar.nil? && criteria.lsar.include?("passed, 0 failed")
+    Hbc.homebrew_prefix.join("bin", "lsar").exist? &&
```
jawshooah: This same path is used twice; let's set `lsar = Hbc.homebrew_prefix.join("bin", "lsar")` and use it again on the next line.

jawshooah:
I know we're generally not the best with method decomposition in this codebase, but `Hbc::Container::Dmg#extract` is a bit much. Mind breaking that down into smaller, private methods? Bonus points if you can replace the mutative `reject!` and `map!` calls with their functional counterparts to soothe this purist's soul 😉

test/support/Casks/lzma-asset.rb
```diff
@@ -0,0 +1,11 @@
+test_cask 'lzma-asset' do
+  version '1.2.3'
+  sha256 '9d7edb32d02ab9bd9749a5bde8756595ea4cfcb1da02ca11c30fb591d4c1ed85'
+
+  url TestHelper.local_binary_url('lzma-asset.xz')
```
jawshooah: Travis is failing because this should be `lzma-asset.lzma`

reitermarkus: Oh right, I missed that, thanks.

jawshooah:
Thanks!

PR #22894, opened 2016-07-18
Deleted Groovesquid

cafferata:
#### Deleted an existing cask
- [x] `brew cask audit --download groovesquid` is error-free.
- [x] `brew cask style --fix groovesquid` left no offenses.


vitorgalvao:

PR #22899, opened 2016-07-18
Deleted Nibble.

cafferata:
#### Deleted an existing cask
- [x] `brew cask audit --download nibble` is error-free.
- [x] `brew cask style --fix nibble` left no offenses.


vitorgalvao:

PR #22900, opened 2016-07-18
Updated Code::Blocks homepage.

cafferata:
#### Editing an existing cask
- [x] `brew cask audit --download codeblocks` is error-free.
- [x] `brew cask style --fix codeblocks` left no offenses.


vitorgalvao:
Current URL is fine. Also, you need to run the commands. Checking the boxes without running the commands just wastes time.

PR #22901, opened 2016-07-18
Deleted Gifzo.

cafferata:
#### Deleted an existing cask
- [x] `brew cask audit --download gifzo` is error-free.
- [x] `brew cask style --fix gifzo` left no offenses.


vitorgalvao:

PR #22918, opened 2016-07-18
Deleted Musicplayer

cafferata:
#### Deleted an existing cask
- [x] `brew cask audit --download musicplayer` is error-free.
- [x] `brew cask style --fix musicplayer` left no offenses.


vitorgalvao:
For future reference, it seems to be dead.

PR #22956, opened 2016-07-20
updated polymail (1.00)

schneidmaster:
#### Editing an existing cask


schneidmaster:
Note - technically the version is `1.0` according to the app itself, but this does not correctly map to the sparkle url which has two 0's (https://sparkle-updater.polymail.io/osx/builds/Polymail-v100.zip)

PR #22960, opened 2016-07-20
tap_migrations added

AnastasiaSulyagina:
### Changes to the core

---

Preparation for moving casks (see #15603)


vitorgalvao:
I think the only problem is the files needs to be added [here](https://github.com/caskroom/homebrew-cask/blob/master/test/layout_test.rb#L43).

jawshooah:
Thanks!

PR #23019, opened 2016-07-23
updated sketch (39)

adidalal:
#### Editing an existing cask

### DIFF
--- a/Casks/sketch.rb
+++ b/Casks/sketch.rb
@@ -1,10 +1,11 @@
 cask 'sketch' do
-  version '3.8.3'
-  sha256 '8103e15c81968d65c78409a06e95df34ea57299a56c15b063b5e4cb4abb577e9'
+  version '39'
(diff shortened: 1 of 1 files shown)


Ephemera:
How about like https://github.com/Ephemera/homebrew-cask/commit/6a58967a5dd260c3a275e3c8f8f93447ec6bdd4f

adidalal:
@Ephemera Looks like sketch changed their versioning entirely - see http://www.sketchapp.com/support/updates/

We lose the "version.major" in the appcast, but I've updated the URL to point to the correct one according to the official appcast

Ephemera:
@adidalal 
How about like https://github.com/Ephemera/homebrew-cask/commit/bfc61ef7e5af133dd46267b86144585d7c4112a4
Also bohemiancoding.com is redirect to download.sketchapp.com

adidalal:
Updated URL in https://github.com/caskroom/homebrew-cask/pull/23025, thanks.

Will keep version as is, as it's the *correct* current version (`39`, not `3.9`)

PR #23238, opened 2016-07-28
Deprecate `--binarydir` option.

reitermarkus:
### Changes to the core


reitermarkus:
@caskroom/maintainers, not sure if we should also use `brew --prefix`/`bin` in tests.

Eitot:
@reitermarkus: Technically, the option is now obsolete, not deprecated. :)

reitermarkus:
@Eitot, you're right.

PR #23240, opened 2016-07-28
Fix sqlworkbenchj URL

deanmorin:
#### Editing an existing cask

The download fails because a newer version was released and this version
was moved to the `archive` folder.

I've asked the maintainer for more durable download urls. In the
meantime, I figured it's better to have a working url than the latest
url which will break again when a new version is released.


reitermarkus:
> I figured it's better to have a working url than the latest
url which will break again when a new version is released.

Broken URLs are the worst. 😜

PR #23257, opened 2016-07-29
Kid3.app link binaries

beschoenen:
#### Editing an existing cask

### DIFF
--- a/Casks/kid3.rb
+++ b/Casks/kid3.rb
@@ -9,6 +9,5 @@ cask 'kid3' do
   license :gpl
 
   app 'kid3.app'
-  binary "#{appdir}/kid3.app/Contents/MacOS/kid3"
   binary "#{appdir}/kid3.app/Contents/MacOS/kid3-cli"
 end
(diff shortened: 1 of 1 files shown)


adidalal:
I don't use the app, what do the binaries do?

beschoenen:
The `kid3` binary starts the app, with an optional given folder.
```bash
kid3 ~/Music/
```
And the `kid3-cli` can be used to tag music via the, well, CLI.
http://kid3.sourceforge.net/kid3_en.html#kid3-cli

vitorgalvao:
`kid3` seems to just be something you’re adding yourself, to call the app from the CLI. We don’t do that, and only link if it’s something officially recommended.

Reason is to not pollute `/usr/local/bin` directories of users with things only a single user wants.

beschoenen:
The `kid3-cli` binary would be very useful, though for bash scripts or w/e.

vitorgalvao:
Yes, that one is fine.

PR #23265, opened 2016-07-29
Get rid of cache symlinks.

reitermarkus:
### Changes to the core

---
### Changes

The new cache directory is now at `HOMEBREW_CACHE`/`Cask`.

All symlinks in the old cache, `HOMEBREW_CACHE`/`Casks`, are checked, their sources moved to the new location, and then deleted. After this, `HOMEBREW_CACHE`/`Casks` is removed recursively.

When moving old cache files, they are also renamed according to a new naming convention, `#{token}--#{version}.#{ext}`, to allow for easier extraction of the token. The regular expression used for this should work for most cases, if it doesn't the worst case is that the file has to be re-downloaded.

---
### Todo
- [x] ~~Update Documentation~~ (there is none)

---

Fixes https://github.com/caskroom/homebrew-cask/issues/22727.

### DIFF
--- a/lib/hbc.rb
+++ b/lib/hbc.rb
@@ -1,5 +1,3 @@
-HOMEBREW_CACHE_CASKS = HOMEBREW_CACHE.join("Casks")
-
 module Hbc; end
 
 require "hbc/extend"
@@ -51,8 +49,33 @@ module Hbc
 
   def self.init
     odebug "Creating directories"
-    HOMEBREW_CACHE.mkpath unless HOMEBREW_CACHE.exist?
-    HOMEBREW_CACHE_CASKS.mkpath unless HOMEBREW_CACHE_CASKS.exist?
+
+    # cache
+    Hbc.cache.mkpath unless Hbc.cache.exist?
+    if Hbc.legacy_cache.exist?
+      ohai "Migrating cached files to #{Hbc.cache}..."
+
+      Hbc.legacy_cache.children.select(&:symlink?).each do |symlink|
+        file = symlink.readlink
+
+        new_name = file.basename
+                       .sub(%r{\-((?:(\d|#{Hbc::DSL::Version::DIVIDER_REGEX})*\-\2*)*[^\-]+)$}x,
+                            '--\1')
+
+        renamed_file = Hbc.cache.join(new_name)
+
+        if file.exist?
+          puts "#{file} -> #{renamed_file}"
+          FileUtils.mv(file, renamed_file)
+        end
+
+        FileUtils.rm(symlink)
+      end
+
+      FileUtils.remove_entry_secure(Hbc.legacy_cache)
+    end
+
+    # caskroom
     unless caskroom.exist?
       ohai "Creating Caskroom at #{caskroom}"
       current_user = Hbc::Utils.current_user
--- a/lib/hbc/cli/audit.rb
+++ b/lib/hbc/cli/audit.rb
@@ -44,4 +44,8 @@ class Hbc::CLI::Audit < Hbc::CLI::Base
   def cask_tokens
     @cask_tokens ||= @args.reject { |a| a == "--download" }
   end
+
+  def self.needs_init?
+    true
+  end
 end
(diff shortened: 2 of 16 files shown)


reitermarkus:
@caskroom/maintainers, could you take a look?

vitorgalvao:
What documentation do you envision needs updating? I don’t think we mention this anywhere.

reitermarkus:
Right, I didn't actually check, just wanted to make sure I didn't forget to check.

reitermarkus:
@caskroom/maintainers, anything that needs changing?

*Note:* `cache_completes` and `cache_incompletes` are not actually used anywhere, but I'll keep them in because of https://github.com/caskroom/homebrew-cask/issues/21763#issuecomment-224320234.

lib/hbc.rb
```diff
@@ -51,8 +49,33 @@ module Hbc
...
+    if Hbc.legacy_cache.exist?
+      ohai "Migrating #{HOMEBREW_CACHE} to #{Hbc.cache}..."
+
+      relative_to_old_cache = Hbc.cache.relative_path_from(HOMEBREW_CACHE)
+
+      Hbc.legacy_cache.children.select(&:symlink?).each do |symlink|
+        file = symlink.readlink
+
+        new_name = file.basename
+                       .sub(%r{\-((?:(?:\d|#{Hbc::DSL::Version::DIVIDER_REGEX})+\-(?:\d|#{Hbc::DSL::Version::DIVIDER_REGEX})+)*[^\-]+)$},
```
jawshooah: Any way to simplify this substitution, or split it into several different ones? You could even split it into multiple lines with the `x` option to make it more readable.

jawshooah:
Could you summarize the changes to `HOMEBREW_CACHE`, `Hbc.cache`, etc. in the top post? This PR does a bit more than get rid of cache symlinks.

lib/hbc.rb
reitermarkus: Will split it with the `x` option. Unless you can come up with a simpler regex to rename `#{token}-#{version}.#{ext}` to `#{token}--#{version}.#{ext}`, that is.

jawshooah: Why are we replacing one hyphen with two? If everything is cached in a `Cask` subdirectory there shouldn't be any conflicts with cached kegs.

reitermarkus:
@jawshooah, added some info on the changes made to the top post.

lib/hbc.rb
jawshooah: Ah, just saw your addition to the top post, that makes sense.

reitermarkus: Also, I'd like to implement [what you suggested in this comment](https://github.com/caskroom/homebrew-cask/issues/21763#issuecomment-224320234), which is much easier if we know exactly where the token ends.

lib/hbc.rb
```diff
@@ -51,8 +49,33 @@ module Hbc
 
   def self.init
     odebug "Creating directories"
-    HOMEBREW_CACHE.mkpath unless HOMEBREW_CACHE.exist?
-    HOMEBREW_CACHE_CASKS.mkpath unless HOMEBREW_CACHE_CASKS.exist?
+
+    # cache
+    Hbc.cache.mkpath unless Hbc.cache.exist?
+    if Hbc.legacy_cache.exist?
+      ohai "Migrating #{HOMEBREW_CACHE} to #{Hbc.cache}..."
```
jawshooah: I'd change the wording here to `Migrating cached Casks in #{HOMEBREW_CACHE} to #{Hbc.cache}...` since we're not migrating all of `HOMEBREW_CACHE`.

reitermarkus: Maybe simply `Migrating cached files to #{Hbc.cache}...`?

jawshooah: SGTM 👍

mwean:
Looks good to me, nice cleanup! 👍

reitermarkus:
This pull request was merged in 787593a2291460b837d59a3c1cf7129a9c5618f2...92f160f7160c365d3471ddbf68d2225dd6b7756c.

PR #23308, opened 2016-07-31
Add epubmdimporter 1.8

diimdeep:


reitermarkus:
@caskroom/maintainers, should we consider creating a new artifact type for Spotlight plugins like this?

vitorgalvao:
> should we consider creating a new artifact type for Spotlight plugins like this?

If they become common. Do we have any, besides this one?

Casks/epubmdimporter.rb
```diff
@@ -0,0 +1,15 @@
...
+  url "https://github.com/jaketmp/ePub-quicklook/releases/download/v#{version}/epub.mdimporter.zip"
+  appcast 'https://github.com/jaketmp/ePub-quicklook/releases.atom',
+          checkpoint: 'f4b69e735f425ff4f6d8e4cc4c2f2a9daee6140c0988176f64626d4c7b61eb78'
+  name 'EPUB Spotlight'
+  homepage 'https://github.com/jaketmp/ePub-quicklook'
+  license :bsd
+
+  artifact 'epub.mdimporter', target: "#{ENV['HOME']}/Library/Spotlight/AA_epub.mdimporter"
+
+  caveats 'To make it index your EPUB files, you have to: mdimport -r ~/Library/Spotlight/AA_epub.mdimporter'
```
vitorgalvao: Since this is non-destructive, run the command in a `postflight`, instead of telling the user to do it.

PR #23321, opened 2016-08-01
Remove nested container after extracting.

reitermarkus:
### Changes to the core

---

There is no reason to keep a nested container after extraction, as we also wouldn't keep a non-nested container.


reitermarkus:
@caskroom/maintainers, any objections to this?

vitorgalvao:
None.

PR #23322, opened 2016-08-01
Fix support for double extensions.

reitermarkus:
### Changes to the core

---
### Changes

Added support to auto-detect nested containers. The most common ones are archived `.dmg`s (`.dmg.zip`, `.dmg.gz`, `.dmg.bz2`). If a cask has `container nested:` set, auto-detection is disabled.

---

This depends on https://github.com/caskroom/homebrew-cask/pull/23321, as otherwise the KLayout  `suite` will contain the `dmg`.

### DIFF
--- a/Casks/calq.rb
+++ b/Casks/calq.rb
@@ -7,8 +7,6 @@ cask 'calq' do
   homepage 'http://www.katoemba.net/makesnosenseatall/calq/'
   license :gratis
 
-  container nested: "#{token}-#{version}.dmg"
-
   app 'Calq.app'
 
   zap delete: '~/Library/Preferences/com.katoemba.calq.plist'
--- a/Casks/calq.rb
+++ b/Casks/calq.rb
@@ -7,15 +7,9 @@ cask 'calq' do
   homepage 'http://www.katoemba.net/makesnosenseatall/calq/'
   license :gratis
 
-  container nested: "Calq-#{version}.dmg"
+  container nested: "#{token}-#{version}.dmg"
 
   app 'Calq.app'
-
-  # This is a horrible hack to force the file extension.  The
-  # backend code should be fixed so that this is not needed.
-  preflight do
-    system '/bin/mv', '--', staged_path.join("Calq-#{version}"), staged_path.join("Calq-#{version}.dmg")
-  end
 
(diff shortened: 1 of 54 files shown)


reitermarkus:
@caskroom/maintainers, any feedback on this? If not, I'll merge this.

adidalal:
LGTM.

In general, I don't really touch the core and any improvements are a welcome addition, so you can assume a 👍 for anything tagged `core`

reitermarkus:
> you can assume a 👍 for anything tagged core

Will do. 😄

Casks/calq.rb
```diff
@@ -7,15 +7,9 @@
   homepage 'http://www.katoemba.net/makesnosenseatall/calq/'
   license :gratis
 
-  container nested: "Calq-#{version}.dmg"
+  container nested: "#{token}-#{version}.dmg"
```
jawshooah: This won't work for case-sensitive filesystems.

reitermarkus: Why not?

The download is named `#{token}-#{version}.dmg.gz`.
When you `gunzip` it, it becomes `#{token}-#{version}.dmg`.

jawshooah: The filename begins with `Calq`, not `calq`. [This matters](https://github.com/caskroom/homebrew-cask/pull/22743) on case-sensitive filesystems.

reitermarkus: Yes, but in this case it is actually `calq`, not `Calq`. The filename is not derived from the `url` in this case.

jawshooah: Strange, I manually downloaded and extracted the file and it's capitalized.
Oh well; merge away!

reitermarkus: Well, that's because a `gzip` can only contain one file, so the filename will be whatever name your `.gz` file is.

Try renaming it before extracting.

reitermarkus: Ah, I just found noticed that `gzip` has an option:

```
 -N --name            save or restore original file name and time stamp
```

reitermarkus:
This pull request was merged in 93cc9f6852d49e1847ba8e3d26b2b537d205b2c3...cab8caef4fe861abb77305bbbbf30a8b54eb49ca.

PR #23328, opened 2016-08-01
Fix uninstall not working when version changes.

reitermarkus:
### Changes to the core

---
### Todo
- [x] Add test for message when other versions remain installed.
- [x] Add test for uninstalling multiple versions consecutively.

---

https://github.com/caskroom/homebrew-cask/issues/2988
https://github.com/caskroom/homebrew-cask/issues/14058

### DIFF
--- a/lib/hbc/cask.rb
+++ b/lib/hbc/cask.rb
@@ -20,7 +20,7 @@ class Hbc::Cask
   METADATA_SUBDIR = ".metadata".freeze
 
   def metadata_master_container_path
-    caskroom_path.join(METADATA_SUBDIR)
+    @metadata_master_container_path ||= caskroom_path.join(METADATA_SUBDIR)
   end
 
   def metadata_versioned_container_path
@@ -79,7 +79,14 @@ class Hbc::Cask
   end
 
   def installed?
-    staged_path.exist?
+    return false unless caskroom_path.exist?
+
+    if metadata_master_container_path.exist?
+      !(metadata_master_container_path.children.map(&:basename) &
+        caskroom_path.children.map(&:basename)).empty?
+    else
+      !caskroom_path.children.empty?
+    end
   end
 
   def to_s
--- a/lib/hbc/cask.rb
+++ b/lib/hbc/cask.rb
@@ -79,14 +79,7 @@ class Hbc::Cask
   end
 
   def installed?
-    return false unless caskroom_path.exist?
-
-    if metadata_master_container_path.exist?
-      !(metadata_master_container_path.children.map(&:basename) &
-        caskroom_path.children.map(&:basename)).empty?
-    else
-      !caskroom_path.children.empty?
-    end
+    !versions.empty?
   end
 
   def to_s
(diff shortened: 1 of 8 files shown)


lib/hbc/cask.rb
```diff
@@ -72,11 +72,21 @@ def timestamped_versions
   end
 
   def versions
-    timestamped_versions.map(&:first).uniq
+    timestamped_versions.map(&:first)
+                        .reverse
+                        .uniq
+                        .reverse
```
jawshooah: Why reverse twice? Isn't `.uniq.reverse` sufficient?

reitermarkus: Consider this:

```ruby
timestamped_versions = [['a', '1'], ['b', '2'], ['a', '3']]

timestamped_versions.map(&:first) --> ['a', 'b', 'a']
                    .uniq         --> ['a', 'b']
```

Now, `.last` doesn't actually return the version which was installed last.

jawshooah: Ah, gotcha.

jawshooah:
I think a change of this magnitude warrants some new tests. Once those are added and the currently failing tests are fixed, I'm all for merging this 👍

reitermarkus:
@jawshooah, what exactly is `PERSISTENT_METADATA_SUBDIRS`? Is the `gpg` sub-directory used for anything else than `signature.asc`?

According to `fetch_sig` in `gpg.rb`, this is only used for caching. Wouldn't  `Hbc.cache` be a better place for `signature.asc` files?

jawshooah:
@reitermarkus I'm not sure. Looks like it was added in https://github.com/caskroom/homebrew-cask/pull/6117 "to provide persistent per-distribution locations to for GPG data". I'd be fine with moving the signature file to `Hbc.cache`

reitermarkus:
Ah, I had an error in thinking:

I thought the `gpg` directory would be in`caskroom/cask/.metadata/gpg`, which would result in a conflict when globbing the `.metadata` directory to retreive timestamped versions.

But actually it's stored in `caskroom/cask/.metadata/version/timestamp/gpg`, so there's no conflict.

reitermarkus:
@jawshooah, could you take a look at the test I have added (and the one I have altered)? 

Also, not sure what would be the best way to reuse the code from the `uninstalls one version at a time`. `MiniTest` doesn't have a `before(:each)`, right? 

Or should I simply add the test for the `cask 1.2.3 is still installed` message into the existing one?

reitermarkus:
Ok, apparently [`after` and `before`](http://ruby-doc.org/stdlib-2.0.0/libdoc/minitest/rdoc/MiniTest/Spec.html#method-c-after) are equivalent to `teardown` and `setup`. Strange this is, `after` and `before` are working, `teardown` and `setup` are not.

jawshooah:
@reitermarkus LGTM, merge when ready 👍

reitermarkus:
This pull request was merged in 36a98c5a4639f62cad6ecd82f896cbc080c0969e...1835d6d5bd415da0d32924416e2e6edb8fa585b7.

PR #23342, opened 2016-08-01
Push FileBot 4.7.1

rednoah:
- [✓] Commit message includes cask’s name (and new version, if applicable).
- [✓] `brew cask audit --download {{cask_file}}` is error-free.
- [✓] `brew cask style --fix {{cask_file}}` left no offenses.

### DIFF
--- a/Casks/filebot.rb
+++ b/Casks/filebot.rb
@@ -5,7 +5,7 @@ cask 'filebot' do
   # sourceforge.net/filebot was verified as official when first introduced to the cask
   url "https://downloads.sourceforge.net/filebot/filebot/FileBot_#{version}/FileBot_#{version}-brew.tar.bz2"
   appcast 'https://sourceforge.net/projects/filebot/rss?path=/filebot',
-          checkpoint: '099c11dd4063a56f9dd1ed4ed9cb3e63052e6503e9f14424279bcd497bb18a0b'
+          checkpoint: 'b4e8425a3bd238b782cb08efa8793e17a29e76d0cc7e5d6880db4e628ef4630a'
   name 'FileBot'
   homepage 'https://www.filebot.net/'
   license :gpl
--- a/Casks/filebot.rb
+++ b/Casks/filebot.rb
@@ -1,6 +1,6 @@
 cask 'filebot' do
-  version '4.7'
-  sha256 '4f61524ccbfa1be65e3818ba37d983599af67af32cc2dcfd4c8f45d74b394cb2'
+  version '4.7.1'
+  sha256 'a2bb68355f542ce36e562e5f8f4dbccc0d053c4a5a296e85a09f44ca6967e629'
 
   # sourceforge.net/filebot was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


reitermarkus:
Please delete the parts of the pull request template that are not relevant and fill out the rest.

reitermarkus:
Also, the `appcast` `checkpoint` needs to be updated.

rednoah:
Done.

Casks/filebot.rb
```diff
@@ -1,17 +1,17 @@
...
   appcast 'https://sourceforge.net/projects/filebot/rss?path=/filebot',
-          checkpoint: '099c11dd4063a56f9dd1ed4ed9cb3e63052e6503e9f14424279bcd497bb18a0b'
+          checkpoint: 'b4e8425a3bd238b782cb08efa8793e17a29e76d0cc7e5d6880db4e628ef4630a'
   name 'FileBot'
   homepage 'https://www.filebot.net/'
   license :gpl
 
-  app 'FileBot.app'
-  binary "#{appdir}/FileBot.app/Contents/MacOS/filebot.sh", target: 'filebot'
+  app "FileBot_#{version}-brew.app", target: 'FileBot-brew.app'
```
reitermarkus: Please change the `target` here to `FileBot.app` and add [a comment like this](https://github.com/caskroom/homebrew-cask/blob/42437e5fcd93b1a3c95376794e6d5eee2b096160/Casks/bob.rb#L12-L13).

reitermarkus:
Your pull request was merged directly in 6a5496946f146888252800b484da66bc67b75eb3 to keep commit history cleaner. It needed some changes which were made in 6a5496946f146888252800b484da66bc67b75eb3. Thank you for your contribution!

vitorgalvao:
@reitermarkus This one is a bit of a special case. See https://github.com/caskroom/homebrew-cask/pull/21683.

@rednoah has been wanting this different name for a while. However, it’s still not an acceptable use of `target` to remove versions ([we specifically cover that](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases)).

vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/23350.

rednoah:
Does brew cask uninstall old versions? I was worried the `/Applications` folder would end up with lots of `FileBot_1.2.3-brew.app` folders. So in this case just not using `target` will be fine?

Casks/filebot.rb
rednoah: `/Applications/FileBot.app` is not acceptable because it conflicts with the Mac App Store version of the app. Anything else is fine.

vitorgalvao:
> I was worried the `/Applications` folder would end up with lots of `FileBot_1.2.3-brew.app` folders.

That’s a good point. One we’re aware of and working to make better. We have most of installing pretty much nailed down, but the rest still needs a lot of work.

> Mea culpa for not reading the target/version docs.

No worries. I know sometimes our rules can be frustrating when contributing, but as someone who worked on most of them, believe me we have good reasons to enforce them. The alternative leaves us with a bloody mess, and until we’re more stable, being nitpicky pays off.

PR #23359, opened 2016-08-02
Remove more Homebrew-duplicated code.

MikeMcQuaid:
This may well cause some test failures but I'll debug them if it does and it behaves well enough with a quick `brew cask fetch virtualbox` local smoke test .

CC @AnastasiaSulyagina


MikeMcQuaid:
Hooray, this is finally green. The main thing that warrants mention is that tests are now run with `brew cask-tests`. This is because Homebrew's tests also rely on `brew tests` being run and Homebrew's Bash layer now sets a bunch of environment variables that are later needed by `global.rb` and it doesn't make sense to duplicate them here.

jawshooah:
Awesome, thanks @MikeMcQuaid!

PR #23384, opened 2016-08-03
Make cask `versions` method more robust.

reitermarkus:
### Changes to the core

---

Sorting versions by timestamp is more robust than a `semver` comparison with fallback.

### DIFF
--- a/lib/hbc/cask.rb
+++ b/lib/hbc/cask.rb
@@ -64,21 +64,15 @@ class Hbc::Cask
     subdir
   end
 
-  def version_comparator
-    proc do |x, y|
-      begin
-        x_version  =  Gem::Version.new(x)
-        x_version <=> Gem::Version.new(y)
-      rescue ArgumentError
-        x <=> y
-      end
-    end
+  def timestamped_versions
+    Pathname.glob(metadata_master_container_path.join("*", "*"))
+            .map { |p| p.relative_path_from(metadata_master_container_path) }
+            .sort_by(&:basename) # sort by timestamp
+            .map(&:split)
   end
 
   def versions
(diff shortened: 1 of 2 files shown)


lib/hbc/cli/list.rb
```diff
@@ -55,7 +55,7 @@ def self.list_installed
     if @options[:one]
       puts columns
     elsif @options[:versions]
-      installed_casks.each { |cask| puts "#{cask} #{cask.versions.reverse.join(', ')}" }
+      installed_casks.each { |cask| puts "#{cask} #{cask.versions.join(' ')}" }
```
jawshooah: Isn't the `reverse` still necessary to sort installed versions in descending order?

lib/hbc/cask.rb
```diff
@@ -64,21 +64,15 @@ def metadata_subdir(leaf, timestamp = :latest, create = false)
...
-        x_version <=> Gem::Version.new(y)
-      rescue ArgumentError
-        x <=> y
-      end
-    end
+  def timestamped_versions
+    Pathname.glob(metadata_master_container_path.join("*", "*"))
+            .map { |p| p.relative_path_from(metadata_master_container_path) }
+            .map(&:split)
+            .sort_by { |version, timestamp| timestamp }
```
jawshooah: Rubocop is complaining that `version` is unused; just change it to `_version`

lib/hbc/cli/list.rb
reitermarkus: I decided to change it to ascending order as `brew list --versions` also uses ascending order.

jawshooah: Gotcha, fine by me.

PR #23388, opened 2016-08-03
Add gogs v0.9.48

brewingcode:

### DIFF
--- a/Casks/gogs.rb
+++ b/Casks/gogs.rb
@@ -1,0 +1,22 @@
+cask 'gogs' do
+  version '0.9.48'
+  sha256 '8afbd71fde279e77ed5b16c8a80ef409661ccb3953bf6cc867fbede532686475'
+
+  url "https://cdn.gogs.io/gogs_v#{version}_darwin_amd64.zip"
+  name 'Go Git Service'
+  homepage 'https://gogs.io'
+  license :mit
+
+  # shim script (https://github.com/caskroom/homebrew-cask/issues/18809)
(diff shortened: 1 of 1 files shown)


reitermarkus:
Why would you need a wrapper for this?

jawshooah:
@reitermarkus Haven't tested, but it's possible that the script must be run from its containing directory. Is that the case, @brewingcode?

brewingcode:
@reitermarkus as @jawshooah guessed, the script needs to be run in its containing directory, because it:

- looks for files in `public`, `templates`, etc, and
- writes its data into the containing directory

vitorgalvao:
But `binary` only links. Are you saying it still breaks in that case?

jawshooah:
@vitorgalvao Yes, for the same reason that `jar` executables are usually wrapped in a shim that navigates to their containing directory before executing them. It allows the developer to assume all libraries are immediately available from the working directory.

jawshooah:
@brewingcode All of the backup stuff going on in `uninstall_preflight` and `uninstall_postflight` is outside the scope of `homebrew-cask`. If a user wants to back up an install, they can take care of that themselves. It isn't and shouldn't be our concern.

brewingcode:
@jawshooah fair enough, I had a feeling that might be a little out of scope, so it's been removed, I just chopped it off the HEAD of my branch.

brewingcode:
Thanks!

jawshooah:
Thanks for the submission!

PR #23416, opened 2016-08-03
cvc4: migrate to Homebrew

AnastasiaSulyagina:
Deleting cvc4 from cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


jawshooah:
`cvc4` was boneyarded in https://github.com/Homebrew/homebrew-core/pull/909, so there's no longer any conflict.

PR #23419, opened 2016-08-03
docker: migrate to Homebrew

AnastasiaSulyagina:
Deleting docker from cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


jawshooah:
[Docker for Mac](https://docs.docker.com/docker-for-mac/) seems to entail [a good deal more](https://docs.docker.com/docker-for-mac/docker-toolbox/#/the-docker-for-mac-environment) than the vanilla `docker` CLI. I'd prefer to keep this cask around.

vitorgalvao:
Yes, when we first started the deduplicarion issue, these were the same. Now, they install different things.

MikeMcQuaid:
Glad to open this but agreed these may sense to be separate now 👍

PR #23426, opened 2016-08-04
fish: migrate to Homebrew

AnastasiaSulyagina:
Deleting fish from cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


reitermarkus:
These are actually different. I removed `fish` a while ago and renamed `fishapp` to `fish`.

AnastasiaSulyagina:
Sorry, I see.

PR #23430, opened 2016-08-04
Refactoring.

reitermarkus:
### Changes to the core


vitorgalvao:
Seems like something in this change is making tests break.

reitermarkus:
It's nothing in this PR, it's the audit for `audacity` which is failing. Seems like FossHub now uses JavaScript to generate download links, so the `url do` times out.

PR #23433, opened 2016-08-04
use secure (https://) URLs where supported

vszakats:
##### Instructions
#### Editing an existing cask

---

I've used a script to verify if the secure alternative is available, checked the (few) reported ambiguous cases manually:
    https://gist.github.com/vszakats/7bdb547e367338299d029140fd8208ed

### DIFF
--- a/Casks/0ad.rb
+++ b/Casks/0ad.rb
@@ -3,7 +3,7 @@ cask '0ad' do
   sha256 '5161cc784fc0a4e11bf48648cf09cb6f5b38afd13a6e32fa484497690e253f86'
 
   # wildfiregames.com was verified as official when first introduced to the cask
-  url "http://releases.wildfiregames.com/0ad-#{version}-osx64.dmg"
+  url "https://releases.wildfiregames.com/0ad-#{version}-osx64.dmg"
   name '0 A.D.'
   homepage 'http://www.play0ad.com/'
   license :oss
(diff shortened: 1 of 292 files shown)


vszakats:
Ratio of secure/insecure URLs:

Before:
  Secure:     4678
Insecure:     3453

After:
  Secure:     5072
Insecure:     3059

vitorgalvao:
Thank you, but [we already have a script for this](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/cask-switch-https).

There’s an important question: are you certain all of these are `https` at their end, and do not redirect to `http`? Because if an `https` redirects to `http`, we keep it as `http` to not give a false sense of security.

vszakats:
I have seen a few (around 10) of those for homepages. For downloads, my script didn't consider this at this time. It can be modified to not follow insecure redirects though.

Coming from the homebrew-core project with similar contributions, the solution there was to [display a warning](https://github.com/Homebrew/brew/commit/794b08fc0df9c5327f77d44ad4727086bb563a19#diff-ad9201cdd3b6f948345629ad812cd5bd) in case a secure URL got redirected to an insecure one, plus an option to refuse downloads in such case. Maybe this logic can be implemented here as well. Alternatively I can rerun an updated script which refuses to follow insecure redirects.

vitorgalvao:
Since we currently do not have that security feature (we might get it for free after merging with Homebrew), please rerun the script with that in mind.

vszakats:
Okay, will do. In the meantime I've created PR https://github.com/caskroom/homebrew-cask/pull/23434, with some trivial homepage URL upgrades. It was part of this PR, and it'd help to sort out these before running my script. Would you mind merging it?

PR #23438, opened 2016-08-04
Fixed docs/wiki: Adding a cask: style check

yolk:
Changed `brew cask style Casks/my-new-cask.rb` to working `brew cask style my-new-cask`


vitorgalvao:
They both work. Difference is the correct example makes it clearer it can be run on an arbitrary file, if for example you use an external directory for development.

yolk:
Upps, my fault: I used the wrong path.

Sorry!

PR #23442, opened 2016-08-05
Fallback for audacity URL timeout.

reitermarkus:
#### Editing an existing cask

---

This of course only silences the audit error, but doesn't actually fix the URL. FossHub seems to be using JavaScript to generate the actual download link now.

### DIFF
--- a/Casks/audacity.rb
+++ b/Casks/audacity.rb
@@ -9,8 +9,8 @@ cask 'audacity' do
     require 'open-uri'
     # fosshub.com/Audacity.html was verified as official when first introduced to the cask
     open("http://www.fosshub.com/Audacity.html/audacity-macosx-ub-#{version}.dmg") do |io|
-      content = io.read
-      %r{^\<iframe.*src=\"(http.*\.dmg)\".*>}.match(content)[1].to_s
+      content = io.read || ''
+      content.scan(%r{^\<iframe.*src=\"(http.*\.dmg)\".*>})[0].to_s
     end
   end
   name 'Audacity'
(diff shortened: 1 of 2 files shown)


jawshooah:
Looks like the new behavior is part of their remediation after [this recent hack](http://www.theregister.co.uk/2016/08/04/classicshell_audicity_infection/).

reitermarkus:
@jawshooah, just added an overkill solution to scrape the url. 😄

vitorgalvao:
That is *way* overkill. I’d rather just remove audacity or revert to the old `oldfoss` behaviour than require `phantomjs`.

Even excluding we’re adding such overhead for a single cask, it’s not completely installing `phantomjs` will be a sane choice down the line. Recently there was a long period of time where we couldn’t install `phantomjs` on El Capitan due to a problem with the new version that took a good while to fix.

If an app is that difficult to get, then let it not be gettable. I vote to remove the cask altogether, if we can’t get it in a sane manner.

reitermarkus:
> If an app is that difficult to get, then let it not be gettable. I vote to remove the cask altogether, if we can’t get it in a sane manner.

I agree. This “solution” was more of an exercise than anything else.

rgoldberg:
More importantly, `brew cask audit` should not exit when one cask fails audit.  It should output error, then continue to audit the subsequent casks.

jawshooah:
@rgoldberg The problem is that this was an unexpected and uncaught exception, not a failed audit. Normally we do exactly what you suggest and report errors and warnings after all casks have been audited.

rgoldberg:
So, just catch all exceptions in the cask audit loop so this won't happen again.

jawshooah:
Yup, cooking up a PR to do just that.

Out of curiosity, why are you running `brew cask audit` on all casks? There's not much point unless you're a maintainer.

jawshooah:
Closing, as the existing `url` block seems to be working now. No need to remove the cask unless it starts acting up again.

PR #23445, opened 2016-08-05
exiftool: migrate to Homebrew

tsparber:
The `exiftool` was somehow missing from the list in #15603. It is already available in homebrew/core ([exiftool.rb](https://github.com/Homebrew/homebrew-core/blob/master/Formula/exiftool.rb)). It is cmd line only and can be built from source.
Homebrew does not use the latest version, but only those marked as "production release". But it seems also not to be too old.

@AnastasiaSulyagina what do you think?


AnastasiaSulyagina:
Looks like formula to me, 👍

reitermarkus:
Also in favour. And if need be, the development version can always be added to the Formula.

tsparber:
Create ad PR to add the development version to the Formula (#3636).
It can be installed using `brew install exiftool --devel`.

jawshooah:
> The exiftool was somehow missing from the list in #15603.

Either the formula or the cask must have been added after I generated the list. We should probably update it.

PR #23456, opened 2016-08-05
hapu: use secure url

vszakats:
##### Instructions
#### Editing an existing cask

---

Checks fails every time because the appcast checkpoint returns a different value on each download. Not sure how to fix this. Though strangely the "appcast" seems like a human readable ChangeLog page, so maybe it should be used for homepage purposes instead.

### DIFF
--- a/Casks/hapu.rb
+++ b/Casks/hapu.rb
@@ -1,10 +1,8 @@
 cask 'hapu' do
-  version '2.4'
-  sha256 '5a5a2d627b7862b43d231c00de849954340d00097bbab01c2615efefc0600be8'
+  version :latest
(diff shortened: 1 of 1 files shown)


tsparber:
As the appcast changes everytime it must be removed or another one needs to be found (I couldn't find one).
As the url has no version number (and no appcast is available) the `version` token needs to be set to `:latest` and the `sha256` to `:no_check` ([sha256 token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#special-value-no_check)).

vszakats:
Thank you, done.

PR #23462, opened 2016-08-05
use secure URLs (mostly homepages)

vszakats:
##### Instructions
#### Editing an existing cask

---

Let the script try to upgrade _homepage_ URLs to secure ones that may be redirected to insecure.

### DIFF
--- a/Casks/synergy.rb
+++ b/Casks/synergy.rb
@@ -2,7 +2,7 @@ cask 'synergy' do
   version '1.7.6,bcb9da8'
   sha256 '838c14f253bed25ac6cf4b1887bb0cdae530ea969fbe53112aa4dd480e93887a'
 
-  url "http://symless.com/files/packages/synergy-v#{version.before_comma}-stable-#{version.after_comma}-MacOSX1011-x86_64.dmg"
+  url "https://symless.com/files/packages/synergy-v#{version.before_comma}-stable-#{version.after_comma}-MacOSX1011-x86_64.dmg"
   name 'Synergy'
   homepage 'https://symless.com/synergy/'
   license :commercial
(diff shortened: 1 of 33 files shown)


Casks/synergy.rb
```diff
@@ -4,7 +4,7 @@
 
   url "http://symless.com/files/packages/synergy-v#{version.before_comma}-stable-#{version.after_comma}-MacOSX1011-x86_64.dmg"
```
reitermarkus: This should be HTTPS, too.

vszakats: This revealed a regression in my script. Correcting it, and there'll possibly be more catches after a re-run.

tsparber:
Did you also check, that no redirection to an `HTTP` only url is performed? There was once an requirement, that the Cask should use `HTTPS` (for `appcast`, `url`, `homepage`) only if no intermediate server uses `HTTP`.

Btw there is also another script to "switch to https" ([`cask-switch-https`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/cask-switch-https)).

vszakats:
@tsparber Yes, HTTP redirection is only currently allowed for `homepage`s and URLs in comment, but not for `url`s or `appcast`s, where the redirection would be hidden from users' eyes.

But, I'd argue the current system is foolproof. Redirectors may chose either an HTTPS or HTTP redirection based on their own decision, typical scenario is when multiple mirrors are used, some of the HTTPS, some of the not. There is no sure way to detect such logic. For this reason Homebrew core implemented a different solution whereas HTTPS is _always_ used in the formulae where possible, and there is a warning shown at _download time_, telling the user that an insecure redirect has been made. Also such redirects can be disabled altogether on the user's end.

I've linked to the relevant (fairly shot) commit yesterday under another, similar commit.

vszakats:
Link the above referred comment/commit: https://github.com/caskroom/homebrew-cask/pull/23433#issuecomment-237601601

tsparber:
Thanks for point this out, I also think the solution implemented by `brew` is a better solution.

PR #23492, opened 2016-08-06
Lock files while downloading.

reitermarkus:
### Changes to the core

### DIFF
--- a/lib/hbc/cli/cleanup.rb
+++ b/lib/hbc/cli/cleanup.rb
@@ -70,6 +70,7 @@ class Hbc::CLI::Cleanup < Hbc::CLI::Base
 
   def delete_paths(paths)
     cleanup_size = 0
+    files_removed = 0
     paths.each do |item|
       next unless item.exist?
       next if locked?(item)
@@ -77,9 +78,10 @@ class Hbc::CLI::Cleanup < Hbc::CLI::Base
       item_size = File.size?(item)
       cleanup_size += item_size unless item_size.nil?
       item.unlink
+      files_removed += 1
     end
 
-    if cleanup_size.zero?
+    if files_removed.zero?
       puts "Nothing to do"
     else
       disk_space = disk_usage_readable(cleanup_size)
--- a/lib/hbc/cli/cleanup.rb
+++ b/lib/hbc/cli/cleanup.rb
@@ -62,19 +62,13 @@ class Hbc::CLI::Cleanup < Hbc::CLI::Base
     delete_paths(cache_files)
   end
 
-  def locked?(file)
-    !File.open(file).flock(File::LOCK_EX | File::LOCK_NB)
-  rescue
-    true
-  end
-
   def delete_paths(paths)
     cleanup_size = 0
     processed_files = 0
     paths.each do |item|
       next unless item.exist?
       processed_files += 1
-      if locked?(item)
+      if Hbc::Utils.file_locked?(item)
         puts "skipping: #{item} is locked"
         next
       end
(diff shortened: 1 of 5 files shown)


lib/hbc/cli/cleanup.rb
```diff
@@ -66,11 +62,30 @@ def remove_all_cache_files
...
+    true
+  end
+
   def delete_paths(paths)
-    puts "Nothing to do" if paths.empty?
+    cleanup_size = 0
+    files_removed = 0
     paths.each do |item|
+      next unless item.exist?
+      next if locked?(item)
```
jawshooah: Should probably warn the user that the file is locked before skipping.

lib/hbc/cli/cleanup.rb
```diff
@@ -66,11 +62,30 @@ def remove_all_cache_files
     delete_paths(cache_files)
   end
 
+  def locked?(file)
+    !File.open(file).flock(File::LOCK_EX | File::LOCK_NB)
```
jawshooah: Seems like this should be extracted to `Hbc::Utils` since it's also used in `download_strategy.rb`.

lib/hbc/cli/cleanup.rb
reitermarkus: Good idea.

lib/hbc/cli/cleanup.rb
reitermarkus: We could also extend `Pathname` with a `locked?` method.

jawshooah: I'm extremely wary of monkeypatching core library classes and would prefer that we avoid doing so any more than we already have. `file.locked?` certainly reads better than `Hbc::Utils.locked?(file)`, but it also has the potential to conflict with some future `Pathname#locked?` method from the core library.

reitermarkus: Although, I'm still figuring out if this is the best way to check if a file is locked.

A file should ideally be locked during the whole installation process. I found simply locking the files themselves doesn't work for DMGs, as `hdiutil` cannot open locked files, saying the resource is busy.

jawshooah: > I'm still figuring out if this is the best way to check if a file is locked.

Which is exactly why it should be extracted into a utility method, so we only need to change the logic in one place if we find a better way! 😄

> I found simply locking the files themselves doesn't work for DMGs, as hdiutil cannot open locked files, saying the resource is busy.

Is there some way to release the lock for subprocesses?

spec/cask/cli/cleanup_spec.rb
```diff
@@ -25,6 +25,28 @@
       expect(cached_download.exist?).to eq(false)
     end
 
+    it "does not removed locked files" do
+      cached_download = cache_location.join("SomeDownload.dmg")
+      FileUtils.touch(cached_download)
+
+      expect {
```
jawshooah: Need to replace the curly brackets here with parentheses. The `eq` matcher apparently [doesn't support blocks](https://travis-ci.org/caskroom/homebrew-cask/jobs/150228981#L664-L667).

lib/hbc/cli/cleanup.rb
reitermarkus: > Is there some way to release the lock for subprocesses only?

No, when a file is unlocked, any process that wants it can lock it. Homebrew uses per-formula `.brewing` lock files in `HOMEBREW_LOCK_DIR`. That seems like the simplest solution to lock all files associated with a cask.

PR #23493, opened 2016-08-06
Add OpenXcom v1.0 (Stable)

ghost:
- [x] `brew cask audit --download openxcom` is error-free.
- [x] `brew cask style --fix openxcom` left no offenses.
- [x] `brew cask install openxcom` worked successfully.
- [x] `brew cask uninstall openxcom` worked successfully.

### DIFF
--- a/Casks/openxcom.rb
+++ b/Casks/openxcom.rb
@@ -1,0 +1,11 @@
+cask 'openxcom' do
+  version '1.0'
+  sha256 'f20814de28f71ce791506f0c9116b711aa85d6d0dfaa68ecc5678302000c83e4'
+
+  url 'http://openxcom.org/file/1730/'
+  name 'OpenXcom'
+  homepage 'http://openxcom.org'
+  license :gpl
+
+  app 'OpenXcom.app'
+end


Casks/openxcom.rb
```diff
@@ -0,0 +1,11 @@
+cask 'openxcom' do
+  version '1.0'
```
jawshooah: This cask needs `version :latest` and `sha256 :no_check`, as the `url` does not point to a specific version.

socksdev: I considered that, but the url is [http://openxcom.org/file/1730/](http://openxcom.org/file/1730/), which points to a set file. I presume that if they released another version, the url would be different as it would point to `/file/whatever/`.

Do I still need to change it?

vitorgalvao: > I presume that if they released another version, the url would be different

Unlikely. See the other downloads. Windows downloads are `1720`, `1722`, `1724`, depending on type, not version. This is indeed unversioned.

jawshooah:
Thank you for the contribution. It needed some fixes, so they were made in bd9527c8ff9ad5028008158aa32d3b4231e96816. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #23505, opened 2016-08-06
deleted MacOS::Release

AnastasiaSulyagina:
### Changes to the core
## 

Made cask use homebrew's `MacOS::Version`. Not sure how to fix `hacking.md` where it's said about release env variable. 
(_have one test failing but it's failing for me locally in master as well_)

cc @MikeMcQuaid


lib/hbc/macos.rb
```diff
@@ -2,390 +2,384 @@
...
 
-require "hbc/macos/release"
+require "os/mac/version"
 
-module Hbc::MacOS
-  module_function
+module OS
+  module Mac
+    extend self
+    ::MacOS = self # compatibility
```
jawshooah: You can remove the comment here and instead add `# rubocop:disable Style/ConstantName` to avoid the RuboCop error.

lib/hbc/macos.rb
```diff
@@ -2,390 +2,384 @@
...
 require "set"
 
-require "hbc/macos/release"
+require "os/mac/version"
 
-module Hbc::MacOS
-  module_function
+module OS
+  module Mac
+    extend self
```
jawshooah: Use `module_function` instead

jawshooah: Actually, you can omit this entirely, since Homebrew already does this.

lib/hbc/macos.rb
jawshooah: Homebrew actually takes care of this as well, so you can omit it.

lib/hbc/macos.rb
AnastasiaSulyagina: And remove module_function as well?

jawshooah: Yup. `extend self` and `module_function` accomplish [*almost* the same thing](http://idiosyncratic-ruby.com/8-self-improvement.html), so the choice between them is usually just stylistic. In this case, Homebrew has already called `extend self` from within this module, so we don't need to do it again.

lib/hbc/macos.rb
```diff
@@ -2,390 +2,384 @@
...
-  def release_with_patchlevel
-    @release_with_patchlevel ||=
-      Release.new(ENV.fetch("MACOS_RELEASE_WITH_PATCHLEVEL",
-                            `/usr/bin/sw_vers -productVersion 2>/dev/null`.chomp))
-  end
+    def system_dir?(dir)
+      SYSTEM_DIRS.any? { |u| File.identical?(u, dir) }
+    end
 
-  def release
```
jawshooah: Rather than removing this entirely, let's `alias release version` for backwards compatibility.

reitermarkus: Yes, this would break uninstall even more than it already is.

lib/hbc/cli/doctor.rb
```diff
@@ -1,7 +1,7 @@
 class Hbc::CLI::Doctor < Hbc::CLI::Base
   def self.run
-    ohai "macOS Release:", render_with_none_as_error(MacOS.release)
-    ohai "macOS Release with Patchlevel:", render_with_none_as_error(MacOS.release_with_patchlevel)
+    ohai "macOS Release:", render_with_none_as_error(MacOS.version)
+    ohai "macOS Release with Patchlevel:", render_with_none_as_error(MacOS.full_version)
```
reitermarkus: We could just remove `macOS Release with Patchlevel` here alltogether and use the `full_version` for the normal `macOS Release`.

AnastasiaSulyagina: Why use `full_version` instead of `version` for normal `release`? Homebrew uses `version` in cases where Cask used `release` as I understood

reitermarkus: Yes, but `doctor` is mostly used for debugging anyways, so having `Release` when you also have `Release with Patchlevel` doesn't really add any value.

lib/hbc.rb
```diff
@@ -44,7 +44,7 @@ module Hbc
   include Hbc::Utils
 
   # TODO: restrict visibility of this to the DSL
-  ::MacOS = Hbc::MacOS
+  ::MacOS = MacOS
```
reitermarkus: This is already assigned in Homebrew: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/os/mac.rb#L14

PR #23508, opened 2016-08-06
racket: migrate to Homebrew

AnastasiaSulyagina:
Deleting racket from cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 

cc @MikeMcQuaid


leafac:
Currently, `homebrew-core` [does the minimal Racket installation](https://github.com/Homebrew/homebrew-core/blob/c5b4c9e9c03aa77d40f64bffd57f929cf114b749/Formula/racket.rb#L54-L55) and recommends using Cask for the full installation. Now the Cask is gone, and I still want the full installation. What do you think is the correct approach (in order of my personal preference)?

1. Revert the Cask removal. (The `homebrew-core` formula fetches a different URL—a minimal distribution instead of the `.dmg`—, their meanings are distinct.)
2. Add a command-line option to the `homebrew-core` formula to do the full installation.
3. Change the `homebrew-core` Cask to do the full installation.

jawshooah:
Hmm, not sure about this one. I didn't realize the Homebrew distribution was so minimal. Maybe another option would be to revert this PR, and rename Homebrew's `racket` to `minimal-racket`. Pinging @MikeMcQuaid, @vitorgalvao and @rprimus, who has submitted most of the Racket updates to Homebrew.

jawshooah:
Note that Racket's website actually refers to Homebrew's distribution as "Minimal Racket", so it would be a fitting rename.

rprimus:
Mon Aug  8 02:00:34 BST 2016

Thanks @jawshooah.

@Kinnardian originally requested the rename in issue [#49700](https://github.com/Homebrew/legacy-homebrew/issues/49700).

The name change (for the homebrew formula) would be appropriate as there is alot that is not included.

vitorgalvao:
I’m fine with the solution to reinstate here and rename on HB.

MikeMcQuaid:
> I’m fine with the solution to reinstate here and rename on HB.

Also fine with this 👍

leafac:
Thank you all for the feedback on my comment.

I agree the proposed solution is a good idea: keep the Cask to perform the full installation and rename `homebrew-core` formula from `racket` to `minimal-racket`.

I can help with the proposed tasks, if you want.

jawshooah:
@leafac It would be helpful if you could submit a PR to homebrew-core to rename `racket` to `minimal-racket`. Thanks!

leafac:
Here you are: https://github.com/Homebrew/homebrew-core/pull/3735

leafac:
For the record: This Pull Request [was reverted](https://github.com/caskroom/homebrew-cask/pull/23553). The `racket` cask stays.

PR #23528, opened 2016-08-07
removed cask's cpu

AnastasiaSulyagina:
### Changes to the core
## 

Made cask use homebrew's `Hardware::CPU`

cc @MikeMcQuaid


lib/hbc/installer.rb
```diff
@@ -172,7 +172,7 @@ def arch_dependencies
     return unless @cask.depends_on.arch
     @current_arch ||= [
                         Hardware::CPU.type,
-                        Hardware::CPU.arch,
+                        Hardware::CPU.bits,
```
reitermarkus: `bits` is not the same as `arch`.

AnastasiaSulyagina: I couldn't understand why they are different as arch finally fetches bits. It would be nice if you could explain what the difference is so that I could check if brew has anything like this

reitermarkus: `arch` doesn't fetch `bits`, it uses `bits` to fetch from https://github.com/caskroom/homebrew-cask/blob/9e49bb92f53321295d4380f87e09268a5ca263ba/lib/hbc/hardware/cpu.rb#L6-L15.
So effectively, what should be returned by `arch` is one of: `:ppc_7400`, `:ppc_64`, `:i386`, `:x86_64`.

AnastasiaSulyagina: In [depends_on](https://github.com/AnastasiaSulyagina/homebrew-cask/blob/master/lib/hbc/dsl/depends_on.rb#L13) :ppc is specified as category but in brew :ppc [seems to be treated as](https://github.com/AnastasiaSulyagina/brew/blob/master/Library/Homebrew/extend/os/mac/hardware/cpu.rb#L90) 32 bit version of :ppc. Is it ok to refactor cask to have :ppc as 32bit version of :ppc by default?

reitermarkus: I [refactored `depends_on` to not rely on `Hardware::CPU.arch`](https://github.com/caskroom/homebrew-cask/pull/23543), should be good to remove all of it after it is merged.

reitermarkus: https://github.com/caskroom/homebrew-cask/pull/23543 is already merged, so this just needs a rebase and should be good to go then.

lib/hbc.rb
```diff
@@ -44,7 +44,7 @@ module Hbc
   include Hbc::Utils
 
   # TODO: restrict visibility of this to the DSL
-  ::Hardware = Hbc::Hardware
+  ::Hardware = Hardware
```
reitermarkus: This is also not needed anymore.

https://travis-ci.org/caskroom/homebrew-cask/jobs/150564506#L649-L650

PR #23584, opened 2016-08-10
r: migrate to Homebrew 

AnastasiaSulyagina:
Deleting r from cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 

cc @MikeMcQuaid


jawshooah:
Thanks!

strubleca:
I just caught up with the conversations about the Homebrew cask deduplication efforts after having R migrated. The (previously available) R cask contains R GUI apps that are not installed with the homebrew/science/r formula. Installing the formula also has the side effect of forcing all R packages to be built from source instead of installing pre-compiled packages from CRAN. I think it would be best to maintain both the cask and formula given these differences in behavior.

MikeMcQuaid:
> The (previously available) R cask contains R GUI apps that are not installed with the homebrew/science/r formula. Installing the formula also has the side effect of forcing all R packages to be built from source instead of installing pre-compiled packages from CRAN.

This sounds like things that could be added to the formula.



> I think it would be best to maintain both the cask and formula given these differences in behavior.

There will always be some slight differences between the two but that doesn't necessarily justify having both. It would be good to see if it's possible to introduce these changes to the R formula before reverting this PR.

strubleca:
Having been a long time R user on the Mac, deviating from the official build isn't something to do lightly. I appreciated having the official R distribution managed as a cask to automate the installation, but still having the benefits of "official R on OS X."

It's been a while since I've built the whole Mac distribution, GUI included, from source. I recall it being fairly involved. The GUI source itself is maintained separately from the R interpreter source distribution. That might suggest separating the formula if the formula only route is taken, or making the GUI an optional installation component.

The binary R package situation also needs careful consideration. Binary R packages are built and tested against the official R distribution. Mixing the binaries with a 3rd party R build might introduce unexpected incompatibilities. 

Building R packages from source means having to install build dependencies that are otherwise unnecessary. Tracking these down can be time consuming. Granted, most homebrew users won't be too put off about doing this.

Ultimately, I realize that there are conflicting goals at play. Maybe there will be some solution to allow homebrew users to choose between official builds and source packages after the integration is completed.

adidalal:
I really am not a huge fan of forcing builds from source, as not all users have the time/motivation/ability to do that. Additionally, if bottles are being considered, I will refer back to https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-183089961 and a bit earlier, where the preference (my personal preference, as well as some other vocal users) for vendor-supplied builds over the Homebrew-compiled ones.

I can't offer a better solution other than reverting this change, but the issue is still going to be around and arguably warrants discussion now, that the proposed changes are well under way.

----

For `r` specifically:

Apologies if I'm taking @MikeMcQuaid's [comment](https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-183226864) out of context but:
> I'm agreed here; I think we should be pushing precompiled GUIs into HBC and precompiled command-line tools should be built from source and in Homebrew. Building from source in Homebrew means we're able to support more platforms than upstream necessarily supports.

As the "official R" package does indeed [include a GUI](https://cran.r-project.org/bin/macosx/), it should stay in HBC.
> Contains R 3.3.1 framework, R.app GUI 1.68 in 64-bit for Intel Macs, Tcl/Tk 8.6.0 X11 libraries and Texinfo 5.2.

MikeMcQuaid:
How does one pick between `brew cask install r` and `brew install r` if we have both?

MikeMcQuaid:
That's my main issue with the duplications, really. Not all users will look in both places and the reasons why you'd choose one over the other are pretty opaque to non-expert users. It's also worth noting that nothing is stopping users from setting up third-party repositories for either casks or formulae that are removed.

jawshooah:
We could do the same thing we did for `racket`: rename the formula to `minimal-r` or similar.

strubleca:
I see a few options to disambiguate:
- Use comments in the formula/cask information to help users decide
- Use different naming: `r-vendor` vs. `r` or `r` vs. `r-base`
- Rely on blogs from the R and homebrew communities to help users in the decision making.

I believe, for a non-expert R user on a Mac, that installing the source built R interpreter would be a mistake because the behavior won't align with what's available in the R for Mac community.

While it is possible to create a third-party repository, that still makes things difficult for non-expert users (R and homebrew alike). And now the users have to have faith that the third-party maintainers are as good as the core.

What disappoints me is that I no longer have the R cask option, which had been supported. As a more experienced user of both homebrew and R, my options have been diminished or replaced with new dependencies I didn't need before.

MikeMcQuaid:
> I believe, for a non-expert R user on a Mac, that installing the source built R interpreter would be a mistake because the behavior won't align with what's available in the R for Mac community.

Again: I don't see any real reason the source-built R interpreter could not be modified to match functionality with the Cask. Has anyone tried to do so?

strubleca:
I haven't tried to modify the R interpreter code to match the behavior, but really don't relish it as an ongoing effort. I think we're at an impasse, and things will stand the way they are.

MikeMcQuaid:
> I think we're at an impasse, and things will stand the way they are.

If the community is unwilling to perform any work to change things: yes, I think we are.

adidalal:
Is there any reason to inject Homebrew/additional maintenance in between R and the user? Formulas with a GUI should be in caskroom, anyway, afaik.

I like the minimal-r suggestion if there really is a use for a Homebrew formula

MikeMcQuaid:
> I like the minimal-r suggestion if there really is a use for a Homebrew formula

Check the Google results for "Minimal Racket" compared to "Minimal R". The latter does not appear to be a thing. https://cloud.r-project.org doesn't seem to have a particular name but `r-pkg` or `r-gui` could do. Does it require XQuartz, too?

strubleca:
I've seen `r-base` or `r-core` used in other settings for the R interpreter. For the GUI, `r-app` is probably the best choice because it goes by R.app. 

A description of R.app can be found at: https://cloud.r-project.org/bin/macosx/

One of the mirrored locations for the R.app code is: https://cloud.r-project.org/bin/macosx/Mac-GUI-1.68.tar.gz

R.app does not need XQuartz. It's a native application.

jawshooah:
The problem with `r-app` is that it explicitly contradicts the [token reference](https://github.com/caskroom/homebrew-cask/blob/b50e668bdf64cd4ff3ecfc729bc12d931f3906b9/doc/cask_language_reference/token_reference.md). `r-gui` might work, though.

MikeMcQuaid:
> r-gui might work, though.

👍 from me.

strubleca:
`r-gui` works.

jawshooah:
Cool, mind submitting a PR?

strubleca:
Submitted  #23954

gwerbin:
So is there no way now to obtain the official CRAN binary through Homebrew? Or is the [bottled version](https://bintray.com/homebrew/bottles-science/r) an official CRAN build?

adidalal:
It's 'r-app'

gwerbin:
@adidalal thanks, that wasn't clear from the discussion here.

As for the name, maybe I'd have gone with "cran-r". But I'm not about to encourage _another_ renaming.

adidalal:
No, we're using the `-app` suffix to disambiguate. It's a [rule](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md#with-a-formula).

PR #23585, opened 2016-08-10
terraform: migrate to Homebrew

AnastasiaSulyagina:
Deleting terraform from cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 

cc @MikeMcQuaid


jawshooah:
Good to see you here, @phinze! The convention so far has been to keep open-source CLI tools in Homebrew, but if HashiCorp would prefer to keep the Cask around instead I'm fine with it.

@MikeMcQuaid, any objections?

radeksimko:
I think if there is a way to make a Homebrew formula behave like the original cask - i.e. just download the binary and place it to the default path that would be also 👌  (correct me @phinze if I'm wrong).

While I understand why recompilation from scratch may in many cases be useful, the formula upgrade [is now blocked](https://github.com/Homebrew/homebrew-core/pull/4237) purely because a single test is failing while the official binary is available and works just fine.

MikeMcQuaid:
@jawshooah I vaguely object but not strongly.

> While I understand why recompilation from scratch may in many cases be useful, the formula upgrade is now blocked purely because a single test is failing while the official binary is available and works just fine.

Does it definitely work fine on 10.9? If the installation process is failing there I think that's a valid reason for it to be blocked. It's unclear to me whether that's an indication of a legitimate runtime bug on 10.9.

cblecker:
Hi @MikeMcQuaid,
I did a local bottle build under 10.9 using https://github.com/Homebrew/homebrew-core/pull/4237, and terraform builds just fine and creates the bottle. It looks like https://github.com/hashicorp/terraform/pull/8481 was the needed fix, and is patched in to https://github.com/Homebrew/homebrew-core/pull/4237.

That being said, all that might be moot if terraform was moved into homebrew-cask and the pre-compiled and signed binaries are installed into place (instead of compiling locally or installing via bottle).

MikeMcQuaid:
@cblecker It's not unreasonable for open-source software to be compiled by packagers rather than the software creators. As mentioned, I'm completely OK with upstream maintaining their own tap, too.

cblecker:
@MikeMcQuaid I wasn't suggesting that the current process was unreasonable.. I was only referring to @phinze's [above comment](https://github.com/caskroom/homebrew-cask/pull/23585#issuecomment-243585809) that suggested maintaining the cask version over the main homebrew version.

radeksimko:
FYI: I have just created https://github.com/unofficial-cask/hashicorp-tap in case anyone's interested.

phinze:
> It's not unreasonable for open-source software to be compiled by packagers rather than the software creators.

Totally agreed! The only reason I suggested the binary-download strategy supersede the compile-from-source strategy is because HashiCorp doesn't have the bandwidth to provide consistent support to the community members maintaining the build in the Formula. If there is sufficient community interest to successfully maintain the Formula across Terraform releases, then that's perfectly okay. 👌

PR #23603, opened 2016-08-10
utils cleanup from hb code

AnastasiaSulyagina:
### Changes to the core
## 

_Do not merge_

Methods from homebrew are deleted from utils. 
_Problem:_ `install_gem_setup_path!` method from Homebrew doesn't throw an exception when can't find executable (like cask's one did), it exits immediately with error (SystemExit: exit). Cask specs contain tests checking that exception is thrown by this method but it is not any longer. What is more appropriate to do with these tests? I would remove them as I don't know ways to handle SystemExit properly and don't want to leave the duplicated method in utils either. But maybe there are some other ways.

cc @MikeMcQuaid

### DIFF
--- a/lib/hbc/cli/create.rb
+++ b/lib/hbc/cli/create.rb
@@ -18,7 +18,7 @@ class Hbc::CLI::Create < Hbc::CLI::Base
   # for mocking
   # TODO: add an :exec parameter to SystemCommand
   def self.exec_editor(*args)
-    Hbc::Utils.exec_editor(*args)
+    exec_editor(*args)
   end
 
   def self.template(cask_token)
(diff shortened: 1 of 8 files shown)


jawshooah:
@AnastasiaSulyagina I believe I've solved this problem for you in #23730.

jawshooah:
Thank you for the contribution. It needed some fixes, so they were made in 4983d6057a018328adefb6666d69372d0f909e40. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

jawshooah:
Sorry about that @AnastasiaSulyagina, looks like some untracked files screwed up `prfixmaster`. The commit links back to this PR, so rest assured that the changes are still attributed to you!

PR #23633, opened 2016-08-11
Remove compliance test

jawshooah:
### Changes to the core

---

`compliance_test` was cluttering the test suite, and always felt a bit out of place. Move the audit to a separate call to `brew cask audit`, and also run `brew cask style` on all Casks after running the test suite.


jawshooah:
This was ill-advised; auditing all core casks within the test suite makes sense, as any change to the core (or to `audit`) could potentially cause the audit to fail. Also, `brew cask style Casks/*.rb` adds an unacceptable amount of time to the build for individual cask updates.

PR #23677, opened 2016-08-11
updated yate (3.13.0.2)

reitermarkus:
#### Editing an existing cask

### DIFF
--- a/Casks/yate.rb
+++ b/Casks/yate.rb
@@ -1,11 +1,11 @@
 cask 'yate' do
-  version '3.12'
-  sha256 '6611b12ef7336e1add86e35345aa2605fde79248c23354736a14384482a72eff'
+  version '3.13.0.2'
+  sha256 '28aafeefae414c179d1f22866865f87cfe1ff034079ba8fd8825d74d6c48d01f'
 
   url 'https://2manyrobots.com/Updates/Yate/Yate.zip',
       user_agent: :fake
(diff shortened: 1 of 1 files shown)


adidalal:
Not getting the same as you here.

```
-  sha256 '6611b12ef7336e1add86e35345aa2605fde79248c23354736a14384482a72eff'
+  version '3.13.0.2'
+  sha256 'a877df051f1d93086f1bd28617419dc05bd2a6d9f1c5c7c120a7359c5daa991e'
```

reitermarkus:
The `sha256` is confirmed locally as I have downloaded it manually. But the problem is, that the site it “protected” by SiteLock. It shows a captcha when you first visit the site (only if your computer is considered infected, which apparently is always the case when you're using `curl`).

vitorgalvao:
Might be best to just remove the `appcast`. If we can’t reliably check it, it’s worth nothing to us.

reitermarkus:
@vitorgalvao, the problem is not (just) the appcast, it's the whole “protected by SiteLock” thing. The only thing you get with `curl` is the captcha page.

fanquake:
Any consensus on what should happen here?

reitermarkus:
I'm fine with removing it, as I don't think there's a way to bypass SiteLock.

adidalal:
Yep 💀

decrement:
It seems like replacing `user_agent: :fake` with `using: :post` worked on my side. Please verify and it may be the viable solution!

PR #23726, opened 2016-08-12
brew-cask-tests: CD to repo root to enable testing in a dev repo

jawshooah:
### Changes to the core

---

Prior to this change, `brew cask-tests` would always `cd` to `/usr/local/Library/Taps/caskroom/homebrew-cask` before running the test suite, making it impossible to run tests in a dev repo with `developer/develop_brew_cask` enabled.

### DIFF
--- a/cmd/brew-cask-tests.rb
+++ b/cmd/brew-cask-tests.rb
@@ -15,7 +15,8 @@ end
 
 require "English"
 
-(HOMEBREW_LIBRARY / "Taps/caskroom/homebrew-cask").cd do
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -8,3 +8,7 @@
 .DS_Store
 .ruby-version
 coverage
+
+# Sublime Text
```
adidalal: Might want to stick these in your user `.gitignore`?

fanquake: Agree

jawshooah:
Merged as 5a5228c4888af9ad13f36e59417683df8ea11579.

PR #23728, opened 2016-08-12
Move fonts instead of hardlinking them

### LINKED ISSUE #23723: Fonts should be moved, not hardlinked
Per #13201, fonts should be moved to their destinations, not hardlinked. Seems like we missed this in #13966.

Refs https://github.com/caskroom/homebrew-fonts/issues/624

jawshooah:
### Changes to the core

---

Also get rid of the `Hardlinked` class, and merge `Linked` with `Symlinked`.

Closes #23723


jawshooah:
@reitermarkus Looks like there's a race condition of some kind in `cleanup_spec`; it only failed in [one](https://travis-ci.org/caskroom/homebrew-cask/jobs/151696566#L677-L687) of the [two Travis jobs](https://travis-ci.org/caskroom/homebrew-cask/builds/151696565).

reitermarkus:
@jawshooah, might be because [the file is locked non-blockingly](https://github.com/caskroom/homebrew-cask/blob/master/spec/cask/cli/cleanup_spec.rb#L35).

PR #23763, opened 2016-08-13
wireshark: migrated to Homebrew

AnastasiaSulyagina:
Deleting wireshark from Cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). It was decided not to remove it from Homebrew in [#3341](https://github.com/Homebrew/homebrew-core/pull/3341)

cc @MikeMcQuaid


MikeMcQuaid:
I think we can leave wireshark as-is as the Cask is the GUI and Homebrew version defaults to the CLI.

Issue #3333, opened 2014-03-02
Automate the installation of VirtualBox extension pack with cask


yangm97:
It is as simple as running VBoxManage extpack install name-of-extpack

retrography:
+1

yanana:
+1

delitescere:
What I do in bash after `brew cask install virtualbox`:

```bash
(
  cd /tmp;
  vbox_full_version=$(brew cask info virtualbox | head -1 | cut -d' ' -f2);
  vbox_short_version=${vbox_full_version/-*};
  curl -L -O http://download.virtualbox.org/virtualbox/${vbox_short_version}/Oracle_VM_VirtualBox_Extension_Pack-${vbox_full_version}.vbox-extpack;
  sudo VBoxManage extpack install --replace Oracle_VM_VirtualBox_Extension_Pack-${vbox_full_version}.vbox-extpack;
  rm *.vbox-extpack;
)
```

draistrick-fd:
@delitescere - thanks for that josh, very useful.

now any chance we can get a brew cask install virtualbox-extensionpack that does that? ;-)

keen99:
@delitescere - thanks for that josh, very useful.

now any chance we can get a brew cask install virtualbox-extensionpack that does that? ;-)

vitorgalvao:
Unlikely unless we get a PR, @delitescere. This is really low priority (`0`, I’d say) and no maintainer seem to have the time or inclination to work on it.

ftc2:
yeah?
https://github.com/caskroom/homebrew-cask/pull/12806
done

Ngrd:
Thank you, @ftc2! This cask does the job very well. I guess, this issue can be closed for now. Feel free to open it if some additional problems with the current solution will occur.

PR #23764, opened 2016-08-13
graphviz: migrate to Homebrew

AnastasiaSulyagina:
Deleting graphviz from Cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). Was not removed from Homebrew in [#3333](https://github.com/Homebrew/homebrew-core/pull/3333)

cc @MikeMcQuaid


adidalal:
LGTM, graphviz won't even provide binary builds soon anyway

PR #23808, opened 2016-08-15
Fix `info` for casks that use `staged_path` as `suite`.

reitermarkus:
### Changes to the core

---

Closes https://github.com/caskroom/homebrew-cask/issues/23762.


jawshooah:
Is this a common case? I'd really rather we create a new directory rather than moving `staged_path`.

reitermarkus:
Yeah, it's probably better to simply disallow `suite '.'` as only 4 casks use it.

PR #23809, opened 2016-08-15
Fix `list`.

reitermarkus:
### Changes to the core

---

@caskroom/maintainers, does it even make sense anymore to print staged contents when almost everything is moved?


vitorgalvao:
I never use it, precisely because I don’t see any use for it. I’m fine with removing unless someone that uses it can state their use-case.

reitermarkus:
@caskroom/maintainers, also removed the `-l` option, as this is broken anyways because it doesn't use the `installed?` method.

PR #23854, opened 2016-08-16
Move caveats for dotnet installer to postflight

joshka:

The dotnet crypto library must be able to find the openssl libraries in its
rpath, which by default includes /usr/local. If the 1.0.0 libs are not found,
dotnet does not work.


jawshooah:
I'd rather leave this in the `caveats`, as it has the potential to overwrite existing library files in `/usr/local/lib`. Thanks though!

PR #23889, opened 2016-08-18
Assume Ruby 2.

MikeMcQuaid:
Homebrew now always uses Ruby 2.0 so you no longer need to check in all these places for 2.0 or for 2.1 valid syntax.

Part of fixing issues for https://github.com/Homebrew/brew/pull/725


jawshooah:
Thanks!

PR #23890, opened 2016-08-18
Formula/brew-cask.rb: remove old Cask formula.

MikeMcQuaid:
It's no longer needed as it's been an internal Homebrew command for quite a while at this point.


jawshooah:
Thanks!

PR #23892, opened 2016-08-18
Add Halla.app v0.6.2

cubeguerrero:
About Halla "yet another "native" app for accessing your Slack teams. It tries to be more conservative with resources. Halla is built on nw.js, which takes less resources than Electron (most other wrapper apps) or MacGap (Slack official app)."

See more [here](https://github.com/rstacruz/halla/blob/master/README.md)

### DIFF
--- a/Casks/halla.rb
+++ b/Casks/halla.rb
@@ -1,0 +1,15 @@
+cask 'halla' do
+  version '0.6.2'
+  sha256 :no_check
+
+  url "https://github.com/rstacruz/halla/releases/download/v#{version}/halla-osx.zip"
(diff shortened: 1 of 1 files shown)


Casks/halla.rb
```diff
@@ -0,0 +1,15 @@
+cask 'halla' do
+  version '0.6.2'
+  sha256 :no_check
```
jawshooah: You must include a `sha256`.

jawshooah:
Thank you for the contribution. It needed some fixes, so they were made in 7894567e425d8e919cf5c584b2b3b1f213a06edb. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

Issue #19746, opened 2016-03-12
Adobe Apps


lukeadams:
It looks like there are a handful of Adobe applications already in Cask. What do you guys think about adding the rest of them?
I found a [great list with the most recent download links](http://prodesigntools.com/adobe-cc-2015-direct-download-links.html#ddl), so it would be relatively simple to generate formulae for each. I might contact the author of that post to see how they are scraping the links.
Have:
- Air
- Air Runtime helper ("ARH")
- Bridge CC
- CC Desktop App
- Digital Editions
- DNG Convertor
- Illustrator CC
- Indesign CC
- Photoshop CC
- Lightroom
- Reader

Missing:
- [ ] InCopy CC
- [ ] InCopy CC
- [ ] Dreamweaver CC
- [ ] Dreamweaver CC
- [ ] Animate CC
- [ ] Premiere Pro CC 
- [ ] After Effects CC
- [ ] Audition CC
- [ ] Prelude CC 
- [ ] SpeedGrade CC 
- [ ] Media Encoder CC 
- [ ] Adobe Muse CC 
- [ ] Edge Animate CC 
- [ ] Edge Code CC 
- [ ] Edge Inspect CC	
- [ ] Edge Reflow CC 
- [ ] Edge Reflow CC 
- [ ] Fuse CC 
- [ ] Adobe Scout
- [ ] Adobe Drive
- [ ] Extension Manager CC

vitorgalvao:
I clicked a bunch of those links, and they all gave an *access denied* error.

You’re absolutely welcome to submit PRs for those other casks (we don’t call them formulae, as those are a different thing), but the core team will likely not do it. Apart from those not having any demand (here), they’re anything but easy to add (as expected from such a consumer-hostile company). Just ask @winkelsdorf how hard those are to do right. On the other hand, since he has already done a great job with the main apps, you might have an easier time by following the already existing casks. I doubt it’ll be the same for all of those, though.

lukeadams:
> all gave an access denied error.

The Adobe trials require a cookie to be set: (snippet from [InDesign cask](https://github.com/caskroom/homebrew-cask/blob/master/Casks/adobe-indesign-cc.rb#L5-L7))
```ruby
url 'http://trials3.adobe.com/AdobeProducts/IDSN/11/osx10/InDesign_11_LS20.dmg',
  user_agent: :fake,
  cookies: { 'MM_TRIALS' => '1234' }
```
> Apart from those not having any demand

[...]

PR #23900, opened 2016-08-19
Add Adobe After Effects

danielbayley:


adidalal:
Some of the other Adobe casks may have useful tidbits if you're running into issues, FYI

fanquake:
@danielbayley Are you still working on this?

danielbayley:
@fanquake If by working on this you mean gradually working down a long todo list back towards it then yes 😆 Iv’e been on holiday…

I haven’t quite got it working yet; see questions on #19746

vitorgalvao:
@danielbayley If you need help, @winkelsdorf might be the person.

danielbayley:
As it stands this cask downloads AE successfully, and then runs the installer, but nothing after that point actually happens… nothing actually gets installed despite the `0` exit code. I based this off @lukeadams’ [adobe-animate-cc](https://github.com/caskroom/homebrew-cask/pull/19776/files) cask, and the [adobe-illustrator-cc](https://github.com/caskroom/homebrew-cask/blob/42437e5fcd93b1a3c95376794e6d5eee2b096160/Casks/adobe-illustrator-cc.rb)

Here is the Terminal output…
~~~ sh
$ brew cask install adobe-after-effects-cc
Warning: A Cask for adobe-after-effects-cc is already installed. Add the "--force" option to force re-install.
brew cask install --force adobe-after-effects-cc
==> Downloading http://prodesigntools.com/trials3/AdobeProducts/AEFT/13_5/osx10-
==> Verifying checksum for Cask adobe-after-effects-cc
==> Running installer script /usr/local/opt/Caskroom/adobe-after-effects-cc/2015
Password:
==> Starting Installer... In Silent mode.
==> Installation successful.
==> Exiting Installer with Code: 0
🍺  adobe-after-effects-cc was successfully installed!
~~~

vitorgalvao:
@danielbayley It’s unlikely you’ll get help from the core team on this. Adobe un/installations are a pain as casks and complex to make, and I’m somewhat vocal against Adobe (even being a designer and having used their software in the past, they are an incredibly user-hostile company).

If you can get help from the other users who successfully built casks for them (or can do it yourself) and make the cask work, we can merge it, but otherwise I don’t think (I may very well be wrong) any maintainer is interested in wasting (and I do mean “waste”, in this case) time on these.

danielbayley:
> Adobe un/installations are a pain as casks and complex to make, and I’m somewhat vocal against Adobe (even being a designer and having used their software in the past, they are an incredibly user-hostile company).

You’re preaching to the choir on that note! I _really_ can’t stand them as a company… the result of a virtual monopoly. The UX around their software absolutely _sucks_. But what else is a professional going to seriously use?

That said, I don’t think Cask itself should be opinionated against certain software…

vitorgalvao:
> But what else is a professional going to seriously use?

The [Affinity Suite of apps](https://affinity.serif.com/). They’re the only thing that ever came close to me *and* they’re actually superior in some regards. They’re lacking in some aspects and I’m also vocal with them about what those are, but they are young (the apps, not the people working on them) and know they have a lot to improve. They have equivalents to Illustrator, Photoshop, and (coming) InDesign.

For After Effects the competition is less direct, but I did find some that looked seriously capable, a few months ago. Unlike with the Affinity apps, I can’t attest to AE alternatives since I don’t use it.

On that regard, the mentality of “I won’t use anything else because nothing else comes close” is *exactly why* Adobe has such a monopoly. If you don’t make the effort to use something else, even it does not yet satisfy all your needs, that company has one less user, which means less feedback, less money, less resources, less interest, less people hearing about it, less reason to keep it up.

Adobe will lose its throne when people actually stop using them, not by wishing another company magically appears and immediately does everything they can and more. That is not going to happen.

> I don’t think Cask itself should be opinionated against certain software

And we are not. If we were, no Adobe apps would have been allowed. But not being opinionated does not mean maintainers will plunge their volunteer time equally into each app equally. Never have we discriminated against the inclusion of any Adobe app and if you search this repo and [caskroom/versions](https://github.com/caskroom/homebrew-versions) you’ll find me always commenting and trying to make the Adobe apps work. That is exactly why I know they are a pain and (for me) not worth the time they require.

My commitment to homebrew-cask means I’ll guide you in the right direction, it does not mean I’ll spend my weekend researching and running tests on a VM to make sure a cask I want nothing to do with works. Most casks are volunteer-contributed; it’s up to the contributors to make sure they work according to our requirements (that try to be as lenient as possible).

Like I said, in this case if you can get it to work (alone or with help, hence why I suggested someone that worked extensively to make the existing Adobe apps work at all) we will accept it, but I have already wasted too much time on the apps of that rotten company. Ethically, I can no longer justify that to myself.

vitorgalvao:
Any progress on this cask, @danielbayley?

danielbayley:
@vitorgalvao Nope; I’m waiting on the outcome of #24388, which is currently labelled “on hold”…

danielbayley:
> They’re lacking in some aspects

I do keep a hopeful eye on competitors, and have briefly tried some of the Affinity apps out in the past, but those areas that are lacking to me are _vast_, not to mention the integration _between_ the applications (the workflow between Illustrator and After Effects in particular is key for me, or Ai and Photoshop for example…) and yet to me even those aren't nearly integrated enough; I would like to see some of them merged into a single app in fact, given the huge crossover between them.

The ability to seamlessly work with existing Adobe formats (`.ai`,`.psd`,`.indd`,`.aep` etc…) given the massive backlog of projects/assets in these formats is also a big factor unfortunately.

For casual users, or more niche pros who tend to stick mostly to a single app, current alternatives might be almost there, but when we are talking about the whole suite I just can't see anything viable.

I feel like the real Adobe killer would be a Photoshop/Illustrator/InDesign hybrid app all rolled into one with different layer _types_ (typography layer, pixel layer, vector layer, adjustment layer etc…) as opposed to different formats managed by very similar yet just-slightly-different-enough-to-be-annoying apps, and possibly based on web technologies. I would love to see such a thing realised but have no idea where one would even start to climb the absolute mountain that would be to actually build! But yeah like you say… sadly that isn’t going to magically appear.

I don’t blame you for not working on Adobe casks if you don’t even use them; I wouldn’t either… just looking for some help based on the existing casks that _do_ work.

danielbayley:
@vitorgalvao This is working now. 👍

Casks/adobe-after-effects-cc.rb
```diff
@@ -0,0 +1,78 @@
...
+  sha256 '398f9401f524ac10462b2b8cd91389b2b6aee841bd328b6daeaa8fa232bee41a'
+
+  url "http://trials3.adobe.com/AdobeProducts/AEFT/#{version.dots_to_underscores}/osx10-64/AfterEffects_#{version.dots_to_underscores}_LS20.dmg",
+      user_agent: :fake,
+      cookies:    { 'MM_TRIALS' => '1234' }
+  name 'Adobe After Effects CC 2015'
+  homepage 'https://www.adobe.com/uk/products/aftereffects.html'
+  license :commercial
+
+  app = name[0]
```
vitorgalvao: Don’t do this. `name` is fickle and should never be depended on. Make this explicit.

vitorgalvao (changes requested):

vitorgalvao:
Pinging @winkelsdorf and @reitermarkus since they have dealt more with Adobe casks.

Casks/adobe-after-effects-cc.rb
danielbayley: Fixed.

vitorgalvao (approved):

danielbayley:
@reitermarkus https://travis-ci.org/caskroom/homebrew-cask/jobs/166820114#L522

> download not possible: Download failed on Cask 'adobe-after-effects-cc' with message: No space left on device

¯\_(ツ)_/¯

reitermarkus:
> ¯\_(ツ)_/¯

Yep, most of the Adobe casks are too huge for Travis.

One of them passed though, so merging this.

PR #23941, opened 2016-08-20
pre_bug_report: Modify commands

adidalal:
Wasn't sure why `brew untap caskroom/cask` is a suggestion - that ends up killing the local repo.

Suggestions/feedback on a better order/is there anything we can further remove?

Pinging @reitermarkus @jawshooah for comments

### DIFF
--- a/doc/reporting_bugs/pre_bug_report.md
+++ b/doc/reporting_bugs/pre_bug_report.md
@@ -7,9 +7,9 @@ Start by searching for your issue before posting a new one. If you find an open 
 If you did not find your particular bug, before reporting it make sure you have the latest versions of Homebrew, Homebrew-Cask, and all Taps by running the following commands. These will also fix some other issues:
 
 ```bash
+$ cd $(brew --repo); git fetch; git reset --hard origin/master; brew update
+$ brew untap phinze/cask; brew uninstall --force brew-cask
 $ brew update; brew cleanup; brew cask cleanup
-$ brew uninstall --force brew-cask; brew update
-$ brew untap phinze/cask; brew untap caskroom/cask; brew update
 ```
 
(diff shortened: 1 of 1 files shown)


reitermarkus:
We have `brew update` in there twice. Here's how I would order it:

```
$cd $(brew --repo); git fetch; git reset --hard origin/master
$ brew untap phinze/cask; brew uninstall --force brew-cask
$ brew cleanup; brew cask cleanup; brew update



```

adidalal:
Also, I guess that first line really is a just a `brew update` technically, anyway

jawshooah:
> Wasn't sure why `brew untap caskroom/cask` is a suggestion - that ends up killing the local repo.

Because `brew cask <whatever>` is supposed to auto-tap `caskroom/cask` if it's not already tapped. So untapping is just cleaning up, as it should be retapped the next time you run a `brew cask` command.

Seems like this is no longer the case since the merge, though.

adidalal:
Would re-adding in https://github.com/Homebrew/legacy-homebrew/pull/46845 fix things? not sure if it's the *right* way, though.

reitermarkus:
Submitted a PR to auto-tap `caskroom/cask` again: https://github.com/Homebrew/brew/pull/765

adidalal:
Reordered commands and added back in the untap, now that brew will again autotap - is `brew cask cleanup` enough to trigger an autotap?

reitermarkus:
Yes, I moved the auto-tapping before `Hbc.init`, so even `brew cask` without any argument should trigger it.

reitermarkus:
@adidalal, now that I think about it, we should probably move the `brew update` before the `cleanup`s, as the auto-tapping only works after you run `brew update`.

PR #23971, opened 2016-08-22
cocoapods: migrate to Homebrew

AnastasiaSulyagina:
Deleting cocoapods from Cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 

cc @MikeMcQuaid


MikeMcQuaid:
They don't recommend this installation method anywhere on their website so 👍

tsparber:
I'm not sure if this is really a duplicate. The cask installed the CocoaPods.app (editor with syntax highlighting+standalone installation).

> They don't recommend this installation method anywhere on their website so 👍
Most developers do not recommend cask on their website 😉 .
The link to the app can be found on the top right navigation bar.

https://cocoapods.org/app

MikeMcQuaid:
@tsparber I went looking on the homepage before and couldn't find it specifically but I see it now. This could be a separate Cask (again to @jawshooah's disdain I think `cocoapods-app` would make the most sense) if there's demand and it's used.

merktassel:
[`cocoapods` in the main Homebrew repository](https://github.com/Homebrew/homebrew-core/blob/master/Formula/cocoapods.rb) is the command line interface, whereas [`cocoapods` in Caskroom](https://github.com/caskroom/homebrew-cask/blob/4abb9b6d9ff77722c3b1e6a4710c0a02cb885b13/Casks/cocoapods.rb) is (or was) the graphical user interface. They are two different packages. 

Today I noticed the Cocoapods GUI I had installed using Caskroom had been migrated over to the main Homebrew when I was updating my setup, which I found strange considering one is a CLI (more suited for Homebrew) and one is a GUI (more suited for Caskroom). However, I went along with it and now I'm missing the GUI app and can't reinstall it via Caskroom. 

It seems this was a mistake and shouldn't have happened. The packages are not duplicates and install two different things. Perhaps this patch should be reverted, or a new cask created called `cocoapods-app` as @MikeMcQuaid suggested.

PR #23973, opened 2016-08-22
corectl: migrate to Homebrew

AnastasiaSulyagina:
Deleting corectl from Cask as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 

cc @MikeMcQuaid


ilovezfs:
CC @AntonioMeireles FYI.

AntonioMeireles:
@ilovezfs @AnastasiaSulyagina 
Hi, 

this can't be a straight translation. 

- what's in plain homebrew [here](https://github.com/Homebrew/homebrew-core/blob/master/Formula/corectl.rb) is the cli / engine - https://github.com/TheNewNormal/corectl
- the cask this PR relates is the UI front-end to it - https://github.com/TheNewNormal/corectl.app and was provided by an user

so at migration time this cask *should* renamed - perhaps to corectl.app as that's its actual upstream name on our side (https://github.com/TheNewNormal/)

Also FWIW, i'll be mostly unresponsive (vacations 😄) and with very limited reachability till the beggining of September so  please excuse in advance my latency.

ilovezfs:
@AntonioMeireles thanks for that. So if I use corectl.app, does that mean I also need the brew formula installed or is the app complete on its own without the formula installed too?

AntonioMeireles:
@ilovezfs

right now, corectl.app is completely self contained (it actually tries to update corectl  if a new one is released in the meantime) ence consuming it does not require (nor overlaps) corectl formula. 

both are targeted at different audiences - corectl.app is a one stop shop  for those who just want it running and corectl is for power users.

ilovezfs:
@AnastasiaSulyagina @MikeMcQuaid does keeping both (plus a Cask rename) sound reasonable then?

MikeMcQuaid:
To me it sounds like it's worth just keeping the Homebrew formula. "Power users" is more the target audience of Homebrew Cask; it's generally a subset of existing Homebrew users who want a more CLI-heavy workflow. That doesn't map for me with people who want a GUI to manage their service and I think the duplication (or at least naming duplication) is confusing.

CC @jawshooah @vitorgalvao for another example where `brew cask install X` is actually not installing X but a separate project called X.app. Obviously it's up to you but I personally don't think that makes a lot of sense.

ilovezfs:
@AntonioMeireles does keeping only the formula sound all right with you then?

MikeMcQuaid:
👍 from me to keep just the formula.

adidalal:
@MikeMcQuaid I actually disagree here - why are we removing choices from the user?

The GUI and CLI are *not* the same thing, and shouldn't be treated as such. This has been coming up a few times (and I always side towards keeping both, especially if they are vendor supplied). 

Maybe we should write out a naming convention and formalize things?

AnastasiaSulyagina:
I'm also not in favour of just deleting gui. It was developed for special purposes on top of corectl and deleting it just because it happened to have the same name is not really wise. 

In the perfect world where casks are all gui and stay under `.app` I would probably want them to appear in brew search and be installed with `*.app` suffix because it would be clear to user that here is .app and he is going to install it.

ilovezfs:
I don't think it's reasonable to expect us to offer every variant of every piece of software, but a name collision alone isn't a sufficient reason to get rid of something. In this case, I do think dumping the GUI is warranted because it's attempting to install and manage the upgrade of corectl, so they really are colliding in terms of what's being provided not just the name. If users really need this GUI, they can get it outside our ecosystem. We're not actually adding any value by including it since the correctl app gui wants to handle its own upgrades anyway.

reitermarkus:
> We're not actually adding any value by including it since the correctl app gui wants to handle its own upgrades anyway.

This has nothing to do with it upgrading itself, if this was a reason to remove a cask, we would have to remove pretty much every app that has a built-in updater.

ilovezfs:
The reason is that it's upgrading the main component, which is already provided by, managed by, and upgraded by Homebrew. If you can cite other examples of Casks doing that, then we should remove those too.

MikeMcQuaid:
@adidalal I think my main issue is the naming that (repeatedly) makes two projects that are different seem to be the same project, even when created by different authors.

RandomDSdevel:
Here's a thought:  perhaps some of us could work with CoreCTL.app GUI's maintainer to get a version of it that can use Homebrew's copy of CoreCTL and doesn't upgrade it produced and then provide that as the Cask?

MikeMcQuaid:
I think the Cask updating itself is actively desirable as-is being a standalone tool.

ilovezfs:
Is the corectl.app cask especially popular?

jawshooah:
@ilovezfs We don't collect metrics, so there's really no way to tell other than the file's commit history.

AntonioMeireles:
(Still On vacation) just a quick note - one key difference that will remain
between the (corectl) blobs of the corectl and corectl.app formula  is that
the one from corectl is to be built FROM SOURCE at homebrew level and the
corectl.app is to be built/signed by us upstream. (That's why they are at
different paths btw)

Later all (will be back 24/7 early September 😀)

António

RandomDSdevel:
@MikeMcQuaid:  Ah, darn it!  I was hoping some way to decouple updating the GUI via Homebrew Cask from updating the CLI via Homebrew, but, alas, you're right that some people are perfectly fine with applications installed via Cask coming in and swooping down to update themselves right out from under its nose, thus possibly confusing Cask.  (I, personally, am not and typically update my Homebrew Cask–installed apps exclusively from within Cask even when apps' own update mechanisms notify me of updates, but that may or may not be a typical use case.  Besides, the only reason I've been participating in the discussion surrounding this PR is because I'm watching CoreCTL but haven't installed it yet and noticed a similar issue pop up when attempting to migrate a similar case from Cask to Core was discussed briefly.  In any case, I'll shut up and let you guys do your thing here now.)

MikeMcQuaid:
> swooping down to update themselves right out from under its nose, thus possibly confusing Cask

It's the other way around; Cask doesn't provide an upgrade as the expectation is applications will handle it.

RandomDSdevel:
@MikeMcQuaid:  That's the situation right _now_, yes, but some people, like myself, would prefer it if Cask managed things via a `brew cask upgrade` command, which has been discussed multiple times but never been implemented.  At the moment, some of us Homebrew Cask end-users work around this issue by manually running `brew cask uninstall $CASK` (adding `--force` when necessary,) then `brew cask install $CASK`, where `$CASK` is the cask in question (`brew cask reinstall` would work, too, but I don't believe that _this_ has been implemented yet, _either_; sigh…)

adidalal:
Re: cask updating. I use https://github.com/buo/homebrew-cask-upgrade but there's an open issue to implement reinstall, as a start.

Shall we call this 'corectl-gui'?

MikeMcQuaid:
> Shall we call this 'corectl-gui'?

GUI is not used at all on https://github.com/TheNewNormal/corectl.app. If it's worth waiting for @vitorgalvao to return that's fine but I think if upstream call it `corectl.app` then that's a better name.

RandomDSdevel:
@MikeMcQuaid:  
> From @adidalal:  
>> Shall we call this 'corectl-gui'?  
> 
> GUI is not used at all on https://github.com/TheNewNormal/corectl.app. If it's worth waiting for @vitorgalvao to return that's fine but I think if upstream call it `corectl.app` then that's a better name.  

One last peep from me:  how about `corectl-menu-extra` or something along those lines?

adidalal:
Renaming to `corectl-app`

Thanks for all the input, everyone!

ilovezfs:
Do we have a list of others where .app was removed from the name and the resultant name is actually a distinct piece of software, not the exact software whose name was truncated? It would be nice to get these fixed en masse to avoid more nonsense.

ilovezfs:
Also, what's the situation with cask and renaming? brew has explicit support for mapping old names to new names but I noticed you just renamed the file. Does that mechanism not work out of the box for casks? If not, why not and is anyone working on adapting that functionality so it works?

jawshooah:
> Does that mechanism not work out of the box for casks?

Not yet, no.

> If not, why not and is anyone working on adapting that functionality so it works?

No particular reason as far as I know, except that nobody has implemented it yet.

adidalal:
@ilovezfs Wasn't aware that the functionality was implemented, but yes, there's an ongoing discussion at https://github.com/Homebrew/brew-evolution/pull/10, if you have any thoughts on the matter. There are relatively few renames that I can recall, but it would definitely be useful to have, especially as I see a few more renaming coming soon.

ilovezfs:
@jawshooah cool. I was hoping that was the only reason :)

vitorgalvao:
> That's the situation right *now*, yes, but some people, like myself, would prefer it if Cask managed things via a `brew cask upgrade` command

No, not *now*, but at all. @MikeMcQuaid is correct. `upgrade` is on our plans, yes, but if an app can auto-update itself, `upgrade` will have no effect on it. That has already been discussed and pondered at length, decided as the best course of action, and is already underway and reflected on other features. Your expectation/preference is completely removed from our goal. See #13201 for details.

RandomDSdevel:
@vitorgalvao:  Ah, I seem to have missed that particular part of the discussion.  Thanks for pointing me toward it!

vitorgalvao:
It’s the section pertaining to `auto_updates`. In there, it’s mentioned only briefly, but it has been discussed further in other issues (but that is the only one whose number I know by heart).

RandomDSdevel:
@vitorgalvao:  Ah, thanks!  I'll go look at that now, then.

PR #24040, opened 2016-08-25
Refactor PULL_REQUEST_TEMPLATE.

reitermarkus:
Since there is no core anymore, we obviously don't need that part in the template anymore. Also removed the overlapping steps of **Editing an existing cask** and **Adding a new cask**.


chuanxd:
LGTM

PR #24059, opened 2016-08-26
corectl -> corectl-app

adidalal:
https://github.com/caskroom/homebrew-cask/pull/23973


vitorgalvao:
We need to add this to the token reference. It’s so far removed from the rules there, there should be note on exceptions done due to homebrew-collisions.

Care to make the first draft, @adidalal? If not, I can take a look later.

adidalal:
Started the discussion in https://github.com/caskroom/homebrew-cask/issues/24239

PR #24075, opened 2016-08-26
Fix karabiner uninstall stanza

decrement:
karabiner uninstall step was broken. Using uninstall script from author instead.
#### Editing an existing cask

### DIFF
--- a/Casks/karabiner.rb
+++ b/Casks/karabiner.rb
@@ -12,12 +12,15 @@ cask 'karabiner' do
   auto_updates true
 
   pkg 'Karabiner.sparkle_guided.pkg'
-  binary '/Applications/Karabiner.app/Contents/Library/vendor/bin/blueutil'
-  binary '/Applications/Karabiner.app/Contents/Library/utilities/bin/warp-mouse-cursor-position'
+  binary '/Applications/Karabiner.app/Contents/Library/bin/karabiner'
 
-  uninstall quit:    'org.pqrs.Karabiner',
-            pkgutil: 'org.pqrs.driver.Karabiner',
-            kext:    'org.pqrs.driver.Karabiner'
+  uninstall early_script: '/Library/Application Support/org.pqrs/Karabiner/uninstall.sh',
+            script:       {
+                            executable:   '/usr/bin/killall',
+                            args:         ['Karabiner'],
+                            must_succeed: false,
+                            sudo:         false,
+                          }
 
   zap       delete: [
                       '~/Library/Application Support/Karabiner',


jawshooah:
Could you elaborate on how `uninstall` was broken and how this fixes it?

decrement:
I kept getting:

An error of type -10810 has occurred. (-10810)

and it never got past the quit stanza. I have changed the formula to use the native uninstall script, which both quits Karabiner and uninstalls everything relevant cleanly.

Casks/karabiner.rb
```diff
@@ -12,12 +12,13 @@
   auto_updates true
 
   pkg 'Karabiner.sparkle_guided.pkg'
-  binary '/Applications/Karabiner.app/Contents/Library/vendor/bin/blueutil'
-  binary '/Applications/Karabiner.app/Contents/Library/utilities/bin/warp-mouse-cursor-position'
```
jawshooah: Why remove these CLI utilities?

decrement: They were never explicitly referenced in Karabiner's manual:
https://pqrs.org/osx/karabiner/document.html.en#commandlineinterface

I believe the ones I removed are not meant for end-user; there are other binaries that are bundled but were not explicitly called out in formula either:
`/Applications/Karabiner.app/Contents/Library/utilities/bin/post-hid-event`
`/Applications/Karabiner.app/Contents/Library/bin/read-symbolichotkeys`

As such, I am only exposing end-user to the intended CLI binary.

Casks/karabiner.rb
```diff
@@ -12,12 +12,13 @@
...
+  binary "#{appdir}/Karabiner.app/Contents/Library/bin/karabiner"
 
-  uninstall quit:    'org.pqrs.Karabiner',
-            pkgutil: 'org.pqrs.driver.Karabiner',
-            kext:    'org.pqrs.driver.Karabiner'
+  uninstall script: {
+                      executable: '/usr/bin/osascript',
+                      args:       ['/Library/Application Support/org.pqrs/Karabiner/uninstaller.applescript'],
```
jawshooah: This AppleScript opens a GUI dialog prompting the user to click OK before continuing with the uninstall, so that's a no-go. Conveniently, it's really just a wrapper around `uninstall.sh` in the same directory, so you can use that instead.

Casks/karabiner.rb
```diff
@@ -12,12 +12,13 @@
   auto_updates true
 
   pkg 'Karabiner.sparkle_guided.pkg'
-  binary '/Applications/Karabiner.app/Contents/Library/vendor/bin/blueutil'
-  binary '/Applications/Karabiner.app/Contents/Library/utilities/bin/warp-mouse-cursor-position'
+  binary "#{appdir}/Karabiner.app/Contents/Library/bin/karabiner"
```
jawshooah: It doesn't make sense to use `appdir` here, since the `pkg` installer will drop the application into `/Applications` regardless.

decrement:
I have updated pull request with your requests and recommendations.

Casks/karabiner.rb
```diff
@@ -12,12 +12,15 @@
...
-  binary '/Applications/Karabiner.app/Contents/Library/vendor/bin/blueutil'
-  binary '/Applications/Karabiner.app/Contents/Library/utilities/bin/warp-mouse-cursor-position'
+  binary '/Applications/Karabiner.app/Contents/Library/bin/karabiner'
 
-  uninstall quit:    'org.pqrs.Karabiner',
-            pkgutil: 'org.pqrs.driver.Karabiner',
-            kext:    'org.pqrs.driver.Karabiner'
+  uninstall early_script: '/Library/Application Support/org.pqrs/Karabiner/uninstall.sh',
+            script:       {
+                            executable:   '/usr/bin/killall',
```
jawshooah: This was unnecessary. I wasn't able to reproduce your issue using `quit`, but regardless that's why we have `signal`.

adidalal: @jawshooah Didn't realize this PR still had issues - probably easiest to just push a patch rather than revert, but up to you

PR #24124, opened 2016-08-29
Fix Silverlight uninstall

tangestani:
The bundle id was wrong. Even after fixing it there were some files
leftover that weren't included in the receipt data. Turns out these
leftover "*.ni.dll" files are compiled during the installation process,
so the only way to get rid of the plugin completely is to delete the
entire Silverlight.plugin package.


jawshooah:
Thanks!

PR #24125, opened 2016-08-29
Update Microsoft Office uninstall/zap stanzas

tangestani:
- Updated uninstall to include all the jobs that were left running.
- Updated zap to be more thorough.


jawshooah:
Thanks!

PR #24137, opened 2016-08-29
updated hab (0.9.0-20160816000307) Add Docker dependency

davejlong:
#### Editing an existing cask

### DIFF
--- a/Casks/hab.rb
+++ b/Casks/hab.rb
@@ -8,5 +8,7 @@ cask 'hab' do
   homepage 'www.habitat.sh'
   license :apache
 
+  depends_on formula: 'docker'
+
   binary "hab-#{version}-x86_64-darwin/hab"
 end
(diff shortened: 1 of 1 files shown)


Casks/hab.rb
```diff
@@ -8,5 +8,7 @@
   homepage 'www.habitat.sh'
   license :apache
 
+  depends_on formula: 'docker'
```
jawshooah: The homepage says that it specifically requires Docker for Mac, so this should be `depends_on cask: 'docker'`

davejlong:
Habitat really just requires the Docker CLI. Docker for Mac is the easiest way to get up and running, but it also would work with Docker Machine.

jawshooah:
I see. Let's keep the Cask dependency since that's what they explicitly recommend, and users can always skip that dependency with `--skip-cask-deps`.

jawshooah:
Thanks!

PR #24224, opened 2016-09-02
appcast.md: make sure to use system sed

vitorgalvao:
@adidalal @reitermarkus @jawshooah Pointed out in https://github.com/vitorgalvao/tiny-scripts/pull/38. Perhaps it’s why we were getting different results in a few casks a while back.


jawshooah:
👍

PR #24232, opened 2016-09-03
Add sickbeard 13.05.25

deanmorin:
[Sickbeard was removed for duplicating a homebrew formula](https://github.com/caskroom/homebrew-cask/issues/15603), but it has some major differences from the homebrew version.

While it still runs as a web server, it acts more like a native application. It can be added as a login item, keeps its settings & preference files in the standard OSX locations, and has a menu icon.

image

menu icon

I had to reinstall the application as part of [migrating to the new caskroom location](https://github.com/caskroom/homebrew-cask/issues/21913), and the standard homebrew version does not have any of my settings. Migrating the settings would probably be more work than this PR 😛 

Please consider adding this Cask back.
### Checklist

This reverts commit b8feea7b9056689a081f781a0f0dba523c965eb5.

### DIFF
--- a/Casks/sickbeard.rb
+++ b/Casks/sickbeard.rb
@@ -1,0 +1,11 @@
+cask 'sickbeard' do
+  version '13.05.25'
+  sha256 'c6ce8b987447d074622d5ee0c96ee6a6da121f5a43252a552e0fc6f5d1cc0550'
+
+  url "http://sickbeard.lad1337.de/download.php?f=SickBeard-OSX-master-#{version}.dmg"
+  name 'SickBeard'
+  homepage 'http://sickbeard.lad1337.de/'
+  license :gpl
+
+  app 'SickBeard.app'
+end
--- a/tap_migrations.json
+++ b/tap_migrations.json
@@ -33,7 +33,6 @@
   "redis": "homebrew/core",
   "rust": "homebrew/core",
   "serf": "homebrew/core",
-  "sickbeard": "homebrew/core",
   "srclib": "homebrew/core",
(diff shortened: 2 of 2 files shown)


fanquake:
```
==> 1 modified cask: sickbeard
==> Downloading http://sickbeard.lad1337.de/download.php?f=SickBeard-OSX-master-
==> Verifying checksum for Cask sickbeard
audit for sickbeard: warning
 - possible duplicate, cask token conflicts with Homebrew core formula: https://github.com/Homebrew/homebrew-core/blob/master/Formula/sickbeard.rb
Error: audit failed for 1 cask: sickbeard
```

adidalal:
@vitorgalvao 

Should we formalize this as "If a application is not identical or has additional functionality as compared to the homebrew Formula, the Cask should be named with a `-app` or `-gui` suffix."

So, in this case, it would be `sickbeard-app` or `sickbeard-gui`. Thoughts?

I don't like that Homebrew fomulas get the default namespace, especially in cases where I think users will want the GUI more, but I'm fine with it as long as there's a standard in place for both, so we're not just removing the choice (and making HBC more useless in the process)

vitorgalvao:
> Please consider adding this Cask back.

We don’t tend to make decisions without thinking. Adding that line does not make a difference, all your other points do.

As stated in https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-223381890, I think removing the GUI was the right call, at least for now:

image

See also 

And https://github.com/caskroom/homebrew-cask/issues/15603 and

PR #24274, opened 2016-09-05
adding_a_cask.md: make clear the rejected casks faq should be checked

vitorgalvao:
We still get submissions to CLI-only open-source casks and it’s a shame users waste time on crafting those.


fanquake:
Good idea.

PR #24280, opened 2016-09-05
Update pepper-flash SHA256?

mzhaodev:


vitorgalvao:
There was a [pull request template](https://github.com/blog/2111-issue-and-pull-request-templates) when you submitted. It must be filled.

If you were not presented a template, you may have submitted via a CLI tool (likely [`hub`](https://hub.github.com/)). It that was the case, you may wish to see if the tool has a way to let you use the template (`hub` [is working on it](https://github.com/github/hub/issues/1118)).

In either case, please [refer to the template](https://raw.githubusercontent.com/caskroom/homebrew-cask/master/.github/PULL_REQUEST_TEMPLATE.md) and update your comment.

Also, in this case it is likely the version has been updated *and* all the other flash casks also need to be updated.

vitorgalvao:
@andyli You usually update all the flash casks at once. Mind taking a look?

andyli:
I do not know much about pepper-flash. Currently its dmg installation (not using brew cask) doesn't work with my mac for some reason, so I don't know whether the cask works well or not.

vitorgalvao:
@andyli Thank you for the information. How do you usually find the exact version of these?

andyli:
For flash player and flash player debugger, open the app, find "about" in the menu and there is version info.

vitorgalvao:
Seems like the version kept the same, but only this one has the wrong `sha256`. Thank you both.

PR #24291, opened 2016-09-06
upgrade TeXmacs to 1.99.5

da-liii:
### What the commit does
1. upgrade TeXmacs to 1.99.5
2. provide a symbol link in /usr/local/bin
### Checklist

### DIFF
--- a/Casks/texmacs.rb
+++ b/Casks/texmacs.rb
@@ -1,6 +1,6 @@
 cask 'texmacs' do
-  version '1.99.4'
-  sha256 '7065db07e8190c9d211dd32b6e3e711cc9c49210c51afe6fc273fe50aa9c9e3d'
+  version '1.99.5'
+  sha256 '1e753a68bc0dd3dc6d8b5bfe1a0b4a7346408eeae8c76bf60609026c20b7be75'
 
   url "http://www.texmacs.org/Download/ftp/tmftp/macos/TeXmacs-#{version}.dmg"
   name 'GNU TeXmacs'
@@ -8,6 +8,7 @@ cask 'texmacs' do
   license :gpl
 
   app "TeXmacs-#{version}.app"
+  binary "#{appdir}/TeXmacs-1.99.5.app/Contents/MacOS/TeXmacs", target: 'texmacs'
 
   zap delete: '~/.TeXmacs'
 end
--- a/Casks/texmacs.rb
+++ b/Casks/texmacs.rb
@@ -8,7 +8,7 @@ cask 'texmacs' do
   license :gpl
 
   app "TeXmacs-#{version}.app"
-  binary "#{appdir}/TeXmacs-1.99.5.app/Contents/MacOS/TeXmacs", target: 'texmacs'
+  binary "#{appdir}/TeXmacs-#{version}.app/Contents/MacOS/TeXmacs", target: 'texmacs'
 
   zap delete: '~/.TeXmacs'
 end


Casks/texmacs.rb
```diff
@@ -1,13 +1,14 @@
...
+  version '1.99.5'
+  sha256 '1e753a68bc0dd3dc6d8b5bfe1a0b4a7346408eeae8c76bf60609026c20b7be75'
 
   url "http://www.texmacs.org/Download/ftp/tmftp/macos/TeXmacs-#{version}.dmg"
   name 'GNU TeXmacs'
   homepage 'http://www.texmacs.org/'
   license :gpl
 
   app "TeXmacs-#{version}.app"
+  binary "#{appdir}/TeXmacs-1.99.5.app/Contents/MacOS/TeXmacs", target: 'texmacs'
```
vitorgalvao: These are not accepted, as it makes an assumption about the user. If they were, every cask would have one.

Unless you can find an official source of the developer recommending this link, remove it.

Casks/texmacs.rb
```diff
@@ -1,13 +1,14 @@
...
+  version '1.99.5'
+  sha256 '1e753a68bc0dd3dc6d8b5bfe1a0b4a7346408eeae8c76bf60609026c20b7be75'
 
   url "http://www.texmacs.org/Download/ftp/tmftp/macos/TeXmacs-#{version}.dmg"
   name 'GNU TeXmacs'
   homepage 'http://www.texmacs.org/'
   license :gpl
 
   app "TeXmacs-#{version}.app"
+  binary "#{appdir}/TeXmacs-#{version}.app/Contents/MacOS/TeXmacs", target: 'texmacs'
```
vitorgalvao: These are not accepted, as it makes an assumption about the user. If they were, every cask would have one.

Unless you can find an official source of the developer recommending this link, remove it.

sadhen: > it makes an assumption about the user

I am not sure what exactly is `it`. Do you refer to the path to the executable TeXmacs? If it is, I will confirm it from the developers.

I have replaced the hard-coded version number with the variable `version`. Hope it meets the standard.

vitorgalvao: The assumption it makes is the user wants the CLI link that simply calls the GUI app. Read #13201 and you’ll see we want to be as close to a manual installation as possible.

Again, unless the developer specifically encourages users to make this linking (they usually don’t and provide an external CLI script when they do) we will not accept this line.

There’s no point in contacting the developers. If they recommend it, it needs to be somewhere on the website, something anyone can see and is directly recommended for all.

sadhen: > The assumption it makes is the user wants the CLI link that simply calls the GUI app.
There are occasions that the user have to launch the GUI app by CLI with `options`. For example:
1. launch the app with the font cache deleted `texmacs --delete-font-cache`
2. enable debuggings
3. start with a specific guile-scheme init file (because TeXmacs is some what like Emacs, can be extended by scheme)

...

vitorgalvao: This is not a discussion. It’s starting to get repetitive to say that what matters is mimicking a manual installation. Every point you made could be applied to every other cask, and we’re not going to start polluting every user’s `/usr/local/bin/` with things they do not want.

Either you find a section on the official website that promotes doing this type of linking, or you remove the line. As simple as that.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in fba700fd1b97baf83f7d0a0184f0bbe793b5d14f. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

vitorgalvao:
If you find a section on the official website that promotes doing the type of linking you want, you can link us to it for verification and make a new PR with the addition. Otherwise, case closed.

PR #24292, opened 2016-09-06
[merge Sep 9] delete ax* drivers

vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/issues/21279.

Both these casks are broken and there seems to be no resolution or interest in making one.

To give it a few days, this should not be merged before Friday, September 9, 2016.


fanquake:
It is now the 9th, merging this.

PR #24335, opened 2016-09-08
add sync_github_templates

vitorgalvao:
We need only update the [issue and PR templates](https://github.com/caskroom/homebrew-cask/tree/master/.github) in the main repo and run this script to update it across all repos. The script takes into account repo-specific links, so it also saves from mistakes when copy-pasting.

### DIFF
--- a/developer/bin/sync_templates_and_ci
+++ b/developer/bin/sync_templates_and_ci
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+readonly caskroom_online='https://github.com/caskroom'
+readonly caskroom_repos_dir="$(mktemp -d)"
+readonly main_repo_dir="$(mktemp -d)" # From where files will be copied
+readonly caskroom_repos=(homebrew-versions homebrew-fonts homebrew-eid) # homebrew-cask is absent since it's the repo all the others are compared against'
+
+function message {
+  echo "${1}"
+}
+
+function get_main_repo {
+  curl --silent --location "${caskroom_online}/homebrew-cask/archive/master.zip" | ditto -xk - "${main_repo_dir}"
+  mv "${main_repo_dir}/homebrew-cask-master/"{*,.*} "${main_repo_dir}"
+  rmdir "${main_repo_dir}/homebrew-cask-master"
+}
+
+function get_repo {
+  local repo_name repo_dir
+
+  repo_name="${1}"
+  repo_dir="${caskroom_repos_dir}/${repo_name}"
+
+  cd "${caskroom_repos_dir}" || exit 1
+  message "Cloning ${repo_name}…"
+  git clone "${caskroom_online}/${repo_name}.git" --quiet
+  cd "${repo_dir}" || exit 1
+}
+
+function copy_templates_and_ci {
+  local repo_name
+
+  repo_name="${1}"
+  cp -R "${main_repo_dir}/.github" "${main_repo_dir}/.travis.yml" "${main_repo_dir}/ci" '.'
+  /usr/bin/sed -i '' -E "s:homebrew-cask/(pulls|issues):${repo_name}/\1:" '.github/PULL_REQUEST_TEMPLATE.md' # PULL_REQUEST_TEMPLATE has repo-specific links
+}
+
+function push_changes {
+  local is_repo_changed
+
+  is_repo_changed="$(git status --porcelain --ignore-submodules=dirty 2> /dev/null)"
+  if [[ -n "${is_repo_changed}" ]]; then
+    git add --all # Stage everything so 'git diff' does not miss aything (like new paths)
+    for modified_path in $(git diff --name-only --staged); do
+      echo -n "Detected changes to ${modified_path}. "
+      git commit "${modified_path}" --message "$(basename "${modified_path}"): update to match main repo" --quiet
+    done
+
+    echo 'Updating…'
+    git push origin master --quiet
+  fi
+}
+
+get_main_repo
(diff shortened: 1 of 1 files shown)


reitermarkus:
We could also add `ci/travis` to this (or a similar) script, as this is also the same now across all repos.

developer/bin/sync_github_templates
```diff
@@ -0,0 +1,57 @@
...
+
+  repo_name="${1}"
+
+  echo "${github_issue_template}" > '.github/ISSUE_TEMPLATE.md'
+  echo "${github_pr_template}" | sed -E "s:homebrew-cask/(pulls|issues):${repo_name}/\1:" > '.github/PULL_REQUEST_TEMPLATE.md' # PULL_REQUEST_TEMPLATE has repo-specific links
+
+  is_repo_changed="$(git status --porcelain --ignore-submodules=dirty 2> /dev/null)"
+  if [[ "${is_repo_changed}" ]]; then
+    echo 'Changes detected. Updating github templates…'
+    git commit --all --message 'Update github templates to match main repository' --quiet
```
jawshooah: This will also commit any other changes that may have already existed locally. Not a big deal if we create a PR, but since we're pushing directly to master we should probably stash changes before updating the templates. Or just create a PR instead, which I'd prefer.

vitorgalvao: I did not go with the make-a-pr route due to the extra dependency on `hub`.

Regarding your concern, check the top of the script. It always creates temporary repos to make sure they are up-to-date and bypass this possibility, as you wouldn’t have changes on them. Also, it allows you to run the script from any directory even if you don’t have all the repos locally.

developer/bin/sync_github_templates
```diff
@@ -0,0 +1,57 @@
...
+
+  repo_name="${1}"
+
+  echo "${github_issue_template}" > '.github/ISSUE_TEMPLATE.md'
+  echo "${github_pr_template}" | sed -E "s:homebrew-cask/(pulls|issues):${repo_name}/\1:" > '.github/PULL_REQUEST_TEMPLATE.md' # PULL_REQUEST_TEMPLATE has repo-specific links
+
+  is_repo_changed="$(git status --porcelain --ignore-submodules=dirty 2> /dev/null)"
+  if [[ "${is_repo_changed}" ]]; then
+    echo 'Changes detected. Updating github templates…'
+    git commit '.github/ISSUE_TEMPLATE.md' '.github/PULL_REQUEST_TEMPLATE.md' --message 'Update github templates to match main repository' --quiet
```
vitorgalvao: While likely not necessary, this will add a bit more of protection by specifying the files.

vitorgalvao:
> We could also add `ci/travis` to this (or a similar) script, as this is also the same now across all repos.

They still have a few differences, but it seems like they were just forgotten when updating the last time. It seems that if the ones in the other repos were outright replaced with the ones in the main repo, everything would continue to work fine. Is that the case?

developer/bin/sync_github_templates
```diff
@@ -0,0 +1,58 @@
...
+    git pull --rebase origin master --quiet
+  fi
+}
+
+function update_github_templates {
+  local repo_name is_repo_changed
+
+  repo_name="${1}"
+
+  [[ ! -d '.github' ]] && mkdir '.github'
```
vitorgalvao: Added this check so if we ever have more repos, this script will also make sure the initial files can be created.

reitermarkus:
> It seems that if the ones in the other repos were outright replaced with the ones in the main repo, everything would continue to work fine.

At least it should, as the main repo is now also *only* a tap.

vitorgalvao:
> as the main repo is now also *only* a tap

Good point. Revamped the whole script with that in mind. Running it will also fix the CI scripts in the other repos, and if we add new repos, adding them to the top of this script and running it again will also create the necessary files.

developer/bin/sync_templates_and_ci
```diff
@@ -0,0 +1,59 @@
...
+  git clone "${caskroom_online}/${repo_name}.git" --quiet
+  cd "${repo_dir}" || exit 1
+}
+
+function copy_templates_and_ci {
+  local repo_name
+
+  repo_name="${1}"
+  cp -R "${main_repo_dir}/.github" "${main_repo_dir}/ci" '.'
+  sed -i '' -E "s:homebrew-cask/(pulls|issues):${repo_name}/\1:" '.github/PULL_REQUEST_TEMPLATE.md' # PULL_REQUEST_TEMPLATE has repo-specific links
```
jawshooah: /usr/bin/sed

developer/bin/sync_templates_and_ci
```diff
@@ -0,0 +1,59 @@
...
+  message "Cloning ${repo_name}…"
+  git clone "${caskroom_online}/${repo_name}.git" --quiet
+  cd "${repo_dir}" || exit 1
+}
+
+function copy_templates_and_ci {
+  local repo_name
+
+  repo_name="${1}"
+  cp -R "${main_repo_dir}/.github" "${main_repo_dir}/ci" '.'
```
jawshooah: `.travis.yml` belongs here as well

developer/bin/sync_templates_and_ci
vitorgalvao: Done.

developer/bin/sync_templates_and_ci
vitorgalvao: And done.

jawshooah:
👍 Merge when ready!

PR #24338, opened 2016-09-08
Bump SHA & vers to current Bose SoundTouch 13.0.11.16439

BenjaminHCCarr:
### Checklist

The formula was failing silently as the it was serving the current version with the old filename, so had sha mismatch.


vitorgalvao:
You’re not adding any new cask, it makes no sense to fill the bottom part.

BenjaminHCCarr:
edited

vitorgalvao:
Not that bottom part, now you’ve deleted too much. I had already edited your post to fix the issue.

BenjaminHCCarr:
one of these days I'll get it right, thanks for the edit and merge

PR #24341, opened 2016-09-08
sync_templates_and_ci: use rsync

jawshooah:
`rsync` allows us to delete files present at the destination but not the source.

Also split arguments into multiple lines for readability/maintainability.

### DIFF
--- a/developer/bin/sync_templates_and_ci
+++ b/developer/bin/sync_templates_and_ci
@@ -31,7 +31,11 @@ function copy_templates_and_ci {
   local repo_name
 
   repo_name="${1}"
-  cp -R "${main_repo_dir}/.github" "${main_repo_dir}/.travis.yml" "${main_repo_dir}/ci" '.'
+  rsync --archive --delete \
+    "${main_repo_dir}/.github" \
+    "${main_repo_dir}/.travis.yml" \
+    "${main_repo_dir}/ci" \
+    '.'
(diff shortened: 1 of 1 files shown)


developer/bin/sync_templates_and_ci
```diff
@@ -31,7 +31,11 @@ function copy_templates_and_ci {
   local repo_name
 
   repo_name="${1}"
-  cp -R "${main_repo_dir}/.github" "${main_repo_dir}/.travis.yml" "${main_repo_dir}/ci" '.'
+  rsync --archive --delete \
+    "${main_repo_dir}/.github" \
```
jawshooah: Could also shorten this list of sources to `"${main_repo_dir}"/{.github,.travis.yml,ci}` but figured splitting them up would be more readable.

vitorgalvao:
Was just looking into this. It appears GNU `cp` could delete the files not present on the source but BSD `cp` cannot.

developer/bin/sync_templates_and_ci
vitorgalvao: I’d argue keeping them in a single line is readable. Ditto for `"${main_repo_dir}"/{.github,.travis.yml,ci}`, it’s simple enough and easier to add/remove paths in the future.

jawshooah: Cool, done in 18422a7.

PR #24343, opened 2016-09-08
Convert casks to be multilingual.

reitermarkus:
_Sadly_, Firefox only offers [SHA512 checksums](http://releases.mozilla.org/pub/firefox/releases/48.0.2/SHA512SUMS) for their downloads, so maybe we should consider adding SHA512 support.

---

Closes https://github.com/caskroom/homebrew-cask/issues/24342.

### DIFF
--- a/Casks/firefox.rb
+++ b/Casks/firefox.rb
@@ -1,13 +1,81 @@
 cask 'firefox' do
   version '49.0.1'
-  sha256 '4a1b91c93aa3c0d2e530622c2d4203246446409bcc4d0155c2724bd21886669f'
 
-  url "https://ftp.mozilla.org/pub/firefox/releases/#{version}/mac/en-US/Firefox%20#{version}.dmg"
+  language 'de' do
+    sha256 'f7b1d7cb6c14a53e1857d3e6982ed63a00c97bf44051a67c9370f295b6d749da'
+    'de'
+  end
+
+  language 'en-GB' do
+    sha256 '84a9673fb922edb44d6d800b8c8f470964d939251d476de77c8426360595797c'
+    'en-GB'
+  end
+
+  language 'en', default: true do
+    sha256 '4a1b91c93aa3c0d2e530622c2d4203246446409bcc4d0155c2724bd21886669f'
+    'en-US'
+  end
+
+  language 'fr' do
+    sha256 'b553df6fde061caa7f7ae318455c7c8eca5acd5e119d75971c152d91489a5c62'
+    'fr'
+  end
+
+  language 'gl' do
+    sha256 '984cee045a7aefeb4e4adc4ba2e0056775a39a758a7fdea3597efbafc9eeac22'
+    'gl'
+  end
+
+  language 'it' do
+    sha256 '644b07f956ae169a5ad2bf5a7c1d57c4379ea91cd5089e207cd4348eeca1d3c2'
+    'it'
+  end
+
+  language 'ja' do
+    sha256 'f242598e4caced4d126f36acc1541ec2a9e19e3c82859a77935b12649b382e63'
+    'ja-JP-mac'
+  end
+
+  language 'nl' do
+    sha256 '29db6caf2633c43188306974edafcd2a951d1dcbe08aca37c017e5f95611c1fd'
+    'nl'
+  end
+
+  language 'pl' do
+    sha256 '81712bfaf0e167d1408fe2b23e27374892121616e4fc26c996cf2a56ae52416a'
+    'pl'
+  end
+
+  language 'pt' do
+    sha256 '70922d761d9a89a99686855f98a3b820faee7f75d8f289112061c37a5c3bfcb7'
+    'pt-PT'
+  end
+
+  language 'ru' do
+    sha256 'e77704feb0a60d3420440a2712f9cf4889a3911e86a3f63e65c43b07af4b860e'
+    'ru'
+  end
+
+  language 'uk' do
+    sha256 '7e6daa83a3f3ce5332722bf8024cbc41b40c5690391e1203b386db9a56bc945c'
+    'uk'
+  end
+
+  language 'zh-TW' do
+    sha256 'f7678356d09bad2ec1b85771476b87f0e7df4a67eb947db20319a96b53000ee0'
+    'zh-TW'
+  end
+
+  language 'zh' do
+    sha256 '1f35d8a1359757e210c2b01b2891627a6be7425d99ad4f9a7b0487bba129d971'
+    'zh-CN'
+  end
+
+  url "https://ftp.mozilla.org/pub/firefox/releases/#{version}/mac/#{language}/Firefox%20#{version}.dmg"
   appcast "https://aus5.mozilla.org/update/3/Firefox/#{version}/0/Darwin_x86_64-gcc3-u-i386-x86_64/en-US/release/Darwin%2015.3.0/default/default/update.xml?force=1",
           checkpoint: 'a159161411794fc6bb2e6fce6e57df6b99eb14499dde41c0ee4d0803d9277364'
   name 'Mozilla Firefox'
-  homepage 'https://www.mozilla.org/en-US/firefox/'
-  license :mpl
+  homepage 'https://www.mozilla.org/firefox/'
 
   auto_updates true
 
--- a/doc/development/adding_a_cask.md
+++ b/doc/development/adding_a_cask.md
@@ -208,7 +208,7 @@ When an App has a principal stable version, alternative versions should be submi
 
 ### Regional and Localized
 
-When an App exists in more than one language or has different regional editions, the US English one belongs in the main repo, and all the others in [caskroom/homebrew-versions](https://github.com/caskroom/homebrew-versions). When not already part of the name of the App, a [regional identifier](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) and a [language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) are to be appended to the Cask’s token (both when available, or just the appropriate one when not).
+When an App exists in more than one language or has different regional editions, [the `language` stanza should be used to switch between languages or regions](../../../cask_language_reference/readme.md#switch-between-languages-or-egions).
 
 ### Trial and Freemium Versions
 
--- a/Casks/battle-net.rb
+++ b/Casks/battle-net.rb
@@ -2,12 +2,21 @@ cask 'battle-net' do
   version :latest
   sha256 :no_check
 
-  url 'http://www.battle.net/download/getInstallerForGame?os=mac&locale=enUS&version=LIVE&gameProgram=BATTLENET_APP'
-  name 'Blizzard Battle.net'
-  homepage 'https://us.battle.net/en/'
-  license :commercial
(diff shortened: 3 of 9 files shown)


Casks/firefox.rb
```diff
@@ -1,8 +1,51 @@
 cask 'firefox' do
   version '48.0.2'
-  sha256 '9328b33d62a2e95ffd2cc04ca0271c5cf137bc9a09860c527e8608c6ec715445'
 
-  url "https://ftp.mozilla.org/pub/firefox/releases/#{version}/mac/en-US/Firefox%20#{version}.dmg"
+  system_locale = `defaults read .GlobalPreferences AppleLanguages`.sub(%r{\A\((.*)\)\Z}m, '\1').gsub(%r{\s}, '').split(',').first
```
jawshooah: This is pretty messy. If we're going in this direction, we should add support for locales directly in the DSL to avoid having this kind of code in the Cask files themselves.

reitermarkus: Yeah, I thought about something like `MacOS.language`.

adidalal:
Will likely need a cask-repair rewrite/specialized version to handle cases, otherwise updating such casks will be a huge pain

Nice work otherwise, cuts out many extra casks in -versions

reitermarkus:
> updating such casks will be a huge pain

Yes, but there is one upside: All languages will be at the same version.

fanquake:
Nice work, this will also tidy up the versions repo a lot.

Casks/adobe-illustrator-cc.rb
```diff
@@ -1,16 +1,52 @@
 cask 'adobe-illustrator-cc' do
-  version '2015'
+  version '19'
   sha256 'ca3e059b130e0b9a6c56ae55b89fa4d490650e25a9371072f43285c10c9821d2'
 
-  url 'http://trials3.adobe.com/AdobeProducts/ILST/19/osx10-64/Illustrator_19_LS20.dmg',
+  url "http://trials3.adobe.com/AdobeProducts/ILST/#{version}/osx10-64/Illustrator_#{version.major}_LS20.dmg",
```
vitorgalvao: Why `#{version.major}`? It’s the same value as `#{version}`.

Casks/adobe-indesign-cc.rb
```diff
@@ -1,16 +1,52 @@
...
   preflight do
-    system '/usr/bin/sudo', '-E', '--', "#{staged_path}/Adobe InDesign CC 2015/Install.app/Contents/MacOS/Install", '--mode=silent', "--deploymentFile=#{staged_path}/Adobe\ InDesign\ CC\ 2015/deploy/install.xml"
+    deployment_language = case MacOS.language
+                          when %r{^cs} then 'cs_CZ'
+                          when %r{^da} then 'da_DK'
+                          when %r{^de} then 'de_DE'
+                          when 'en-AE' then 'en_AE'
+                          when 'en-GB' then 'en_GB'
+                          when 'en-IL' then 'en_IL'
+                          when %r{^en} then 'en_US'
```
vitorgalvao: Why not let this one be handled by the `else`?

vitorgalvao:
Note: we’ll also need to update [“finding a home for your cask”](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).

Casks/adobe-illustrator-cc.rb
reitermarkus: There is a new version of Photoshop (which I couldn't get to silently install, because Adobe …) that seems to be using this format, so I thought it couldn't hurt using it:

`.../PHSP/17.0.1/osx10-64/AdobePhotoshop17-mul.dmg`

danielbayley:
👍 I think this is definitely a better approach than clogging up _versions_ further… I’ll revise my other Adobe casks once this goes through.

reitermarkus:
Ok, I think that would be all.

jawshooah:
My concern with this is maintenance. We can't test all of these downloads in Travis, and we can't update all of them with `cask-repair`, so the non-US ones are likely to become stale.

doc/development/adding_a_cask.md
```diff
@@ -213,7 +213,11 @@ When an App has a principal stable version, alternative versions should be submi
 
 ### Regional and Localized
 
-When an App exists in more than one language or has different regional editions, the US English one belongs in the main repo, and all the others in [caskroom/homebrew-versions](https://github.com/caskroom/homebrew-versions). When not already part of the name of the App, a [regional identifier](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) and a [language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) are to be appended to the Cask’s token (both when available, or just the appropriate one when not).
+When an App exists in more than one language or has different regional editions, `MacOS.language` should be used to switch between languages or regions based on the system locale. US English (`en-US`) should be used as a fallback.
```
vitorgalvao: This explanation is too technical and specific. If you see the rest of the section, you’ll see it’s concerned with *defining* different types of apps and *where* to put them. Describing how to accomplish it is for another section, like where we decribe `MacOS.release`, or something like that.

vitorgalvao:
Agreed with the maintenance problems. We should likely think of an effective way of how this will be maintained before rushing into changing the casks.

With the current method, a single `for` loop with `cask-repair` can update everything with little effort. With this method, it can take a lot of manual work. It’s preferable to have a single file for all these languages, yes, but not if they’ll be less updated *which is the opposite of the goal*.

I’m also reticent on forcing a specific language on users, in case they want a different one. I’m not affected by any of these casks, though, so I’m fine with whatever users voice preference for.

reitermarkus:
Ok, so the script I added is able to loop through all languages in a `case` statement. Is there an option to run `cask-repair` without creating a PR, so I can test this further? 

I'm thinking of calling `cask-repair` after each time the language is temoprarily changed as a first try.

vitorgalvao:
> Is there an option to run `cask-repair` without creating a PR, so I can test this further?

`cask-repair` does not submit automatically unless you give it the `-b`|`--blind-submit` flag. After making the changes, it asks you to confirm; pick `n`.

jawshooah:
We should allow the system language to be overridden with an environment variable like `HOMEBREW_CASK_LANGUAGE`, which will make testing and updating these casks much easier.

vitorgalvao:
> We should allow the system language to be overridden with an environment variable like `HOMEBREW_CASK_LANGUAGE`, which will make testing and updating these casks much easier.

It will also allow users to force a specific language, which is good (I always saw not having that possibility as a bit of an issue).

mwean:
It seems like it would be better to have a top-level localization concept in the DSL instead of putting `case` statements in all the casks. That would also provide easier integration with testing, etc.

danielbayley:
> We should allow the system language to be overridden with an environment variable like HOMEBREW_CASK_LANGUAGE, which will make testing and updating these casks much easier.

…and/or maybe `--lang=` in `HOMEBREW_CASK_OPTS`?

reitermarkus:
> It seems like it would be better to have a top-level localization concept in the DSL instead of putting case statements in all the casks. That would also provide easier integration with testing, etc.
 
Yes, it *would* be better, but how would you implement this?

I'm thinking something like:

```ruby
language 'la-NG' do
    # ...
end
```

Any other suggestions on how to handle this?

mwean:
@reitermarkus I think the language block you have is a good possibility. The other would be to have methods like `url` accept a hash like

```ruby
url de: "some-url",
     es: "another-url"
```

The benefit of that would be if the url is different but installer, homepage, etc. are all the same. We'd have to look at the breakdown.

> Also, is it worth it when only ~10 casks are using it?

I think so, especially since I assume more casks will take advantage of it once better support is added.

jawshooah:
@mwean But then you'd have to have a corresponding hash param for any stanza whose value might be different for different locales. I think a `language (string|regex) do` block syntax makes the most sense.

mwean:
@jawshooah yeah, I agree the block form is probably better. Just proposing an alternative :)

reitermarkus:
How do we handle the default case with `language do`? Seems like we need some sort of global variable which keeps track of the detected languages (and their hierarchy level).

jawshooah:
> How do we handle the default case with language do?

Easy: don't put it in a `language` block 😄

> Seems like we need some sort of global variable which keeps track of the detected languages (and their hierarchy level).

Isn't that what `MacOS.language` is for?

reitermarkus:
> Easy: don't put it in a language block 😄

This will raise an exception for stanzas where we only allow one per cask, so not that easy.

> Isn't that what MacOS.language is for?

No, MacOS.language simply returns a string, which we have to match against.

With https://github.com/Homebrew/brew/pull/906, there's `languages`, which returns the language hierarchy as an array. So consider this:

When someone is using this language hierarchy and there is no language matching `de-DE`, should we go straight to `en-US` even when there is a language that matches `it-IT` or `en-GB`?

` ['de-DE', 'it-IT', 'en-GB']`

jawshooah:
> This will raise an exception for stanzas where we only allow one per cask, so not that easy.

True. How about `language :default` then?

> When someone is using this language hierarchy and there is no language matching de-DE, should we go straight to en-US even when there is a language that matches it-IT or en-GB?

We should use the first match in `MacOS.languages`, or the default if there are no matches. So in your example, we'd use `it-IT`.

reitermarkus:
> We should use the first match in MacOS.languages

Yes, but how do we keep track? We would have to first know which `language` blocks exist, loop through them. And only after that do we know which one to call.

```ruby
MacOS.languages = ['de-DE', 'it-IT', 'en-GB']

detected_languages = {}

language %r{^it} do
  detected_languages[%r{^it}] = 2
end

language 'en-GB' do
  detected_languages['en-GB'] = 3
end

language :default do

end

final_language = :default

unless detected_languages.empty?
   final_language = detected_languages.sort_by_value.keys.first
end
```

mwean:
I can take a stab at implementing it this weekend

reitermarkus:
Just added another commit with a rough draft to https://github.com/Homebrew/brew/pull/906/commits/b21d419a3f5ff953363dbadf95fe7dabb0650dd6.

The detection is working so far, but I'm not sure where and how to best call the saved block.

reitermarkus:
@caskroom/maintainers, would be great if someone could take a look again at https://github.com/Homebrew/brew/pull/906.

adidalal:
I'm not a rubyist, so I'll defer judgement on the actual code. Big 👍 for the effort and implementation, though - this should clean up `versions` significantly!

Casks/firefox.rb
Saklad5: Is there a reason we couldn’t implement SHA512 without requiring it yet? We could encourage it for future use, but continue using SHA256.

reitermarkus: > Is there a reason we couldn’t implement SHA512 without requiring it yet?

No, but this is pretty low on the list of priorities. The only reason I even brought this up was because it would simplify getting the checksums for *one* specific cask: Firefox.

reitermarkus:
On hold until https://github.com/Homebrew/brew/pull/1216/files is merged and tagged.

reitermarkus:
Yesss!

PR #24391, opened 2016-09-11
Add Messenger for Desktop v2.0.1

wickles:
### Checklist

- [ ] Checked there are no closed [issues](https://github.com/caskroom/homebrew-cask/issues) where that cask was already refused.

Note unmarked checkbox. This cask existed before under as `aluxian-messenger` but was removed, see https://github.com/caskroom/homebrew-cask/pull/18043. However the reason cited for removal was that it was unmaintained. It seems to have been maintained well enough and this is a very recent release, so I think it's worth having available again. (Also, I think it's worth removing [messenger-native](https://github.com/gaston23/MessengerNative) based on the same reasoning.)

Also note the difference between token name and app name. This is intentional and based on the following proposal for organizing the many FB messenger front-ends cropping up: 
- Rename and add 'messenger' type apps as `messenger-[author/project]` wherever possible. examples: 
  - Rename token [`goofy`](https://github.com/danielbuechele/goofy) to `messenger-goofy`
  - Rename token [`messenger`](https://github.com/rsms/fb-mac-messenger) to `messenger-rsms`
  - Add [Caprine](https://github.com/sindresorhus/caprine) as token `messenger-caprine`
  - Add [Messenger for Desktop](https://github.com/Aluxian/Facebook-Messenger-Desktop) as token `messenger-aluxian` (previously name `aluxian-messenger`)
  - (Remove `messenger-native` as outdated / unmaintained)

Obviously these would be largely exceptions to the token naming conventions. So this PR is only for adding back `messenger-aluxian` for now.

### DIFF
--- a/Casks/messenger-for-desktop.rb
+++ b/Casks/messenger-for-desktop.rb
@@ -1,0 +1,14 @@
+cask 'messenger-for-desktop' do
+  version '2.0.1'
+  sha256 'a1813dc526a1eb39099993e677632808372e82bb556cd694d0c2f043f80f4d0c'
+
+  # github.com/Aluxian/Facebook-Messenger-Desktop was verified as official when first introduced to the cask
+  url "https://github.com/Aluxian/Facebook-Messenger-Desktop/releases/download/v#{version}/messengerfordesktop-#{version}-osx.dmg"
+  appcast 'https://github.com/Aluxian/Facebook-Messenger-Desktop/releases.atom',
+          checkpoint: 'ffad9631c3186cb8bc9bdfa9c3a3a984e9981694ae4e3b72de29d27296088071'
+  name 'Messenger for Desktop'
+  homepage 'https://messengerfordesktop.com/'
+  license :mit
+
+  app 'Messenger for Desktop.app'
+end


Casks/messenger-aluxian.rb
```diff
@@ -0,0 +1,12 @@
+cask 'messenger-aluxian' do
+  version '2.0.1'
+  sha256 'a1813dc526a1eb39099993e677632808372e82bb556cd694d0c2f043f80f4d0c'
+
+  # github.com/Aluxian/Facebook-Messenger-Desktop was verified as official when first introduced to the cask
+  url "https://github.com/Aluxian/Facebook-Messenger-Desktop/releases/download/v#{version}/messengerfordesktop-#{version}-osx.dmg"
```
vitorgalvao: Please add an [`appcast`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

vitorgalvao:
There’s no point in making a specific rule for `messenger` apps, when we already have disambiguation rules. Furthermore, [we are not a discoverability service](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not-a-discoverability-service.md), so this grouping serves no purpose.

This needs to be renamed `messenger-for-desktop`, and the other casks will remain as are.

Thank you for the proposal, though.

wickles:
Understood. What about token `messenger` in particular? That one is especially ambiguous. Would it make sense to rename it to `rsms-messenger`? As in the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md#converting-the-simplified-name-to-a-token): 

> We avoid defining Cask tokens in the repository which differ only by the placement of hyphens. Prepend the vendor name if needed to disambiguate the token. 

Re: this PR, I will rename and add the appcast.

vitorgalvao:
We mean disambiguate only in the sense of not making it confusing *in relation to other casks*, so that one is fine.

If facebook ever releases an official one, or another more popular app with the same name emerges, that one will be “demoted”.

Ideally, they would have authorisation from facebook to use the name. There’s an [`appearin` cask](https://github.com/caskroom/homebrew-cask/blob/master/Casks/appearin.rb) whose name may make it seem like it is an official [appear.in](https://appear.in/) app, but it is not (I’m the developer). However, [I have official authorisation to use the name](https://github.com/vitorgalvao/appear.in/pull/1) and I have directly and specifically contacted them to ask for such permissions (there are some attached conditions, which I’m happy to oblige. They are publicly documented in the repository).

But circling back to the notion [we’re not a discoverability service](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md):

> users are expected to have reasonable knowledge about the apps they’re installing through us before doing so

which means the actual tokens for the apps just need to be clear enough for you to be sure which app you want and not make a mistake (i.e. when the change is small between two casks, like the hyphens example), not necessarily to describe what exactly the app is.

wickles:
Gotcha. Updated PR as requested.

vitorgalvao:
Thank you.

PR #24411, opened 2016-09-12
Add OKI MC860 PS printer support

franklouwers:
### Checklist

### DIFF
--- a/Casks/oki-postscript-drivers.rb
+++ b/Casks/oki-postscript-drivers.rb
@@ -1,4 +1,4 @@
-cask 'okimc860' do
+cask 'oki-postscript-drivers' do
   version '2.0.2'
   sha256 'b8d5058744d93d040433eb841dcd2481848e61b659e84da459785310044857fa'
 
--- a/Casks/oki-postscript-drivers.rb
+++ b/Casks/oki-postscript-drivers.rb
@@ -1,0 +1,13 @@
+cask 'okimc860' do
+  version '2.0.2'
+  sha256 'b8d5058744d93d040433eb841dcd2481848e61b659e84da459785310044857fa'
+
+  url 'http://belgium.oki.com/Includes/Pages/FileDownload.aspx?id=tcm:125-152412'
+  name 'OKI MC860 PS Driver for OSX'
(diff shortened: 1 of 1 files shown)


Casks/oki-postscript-drivers.rb
```diff
@@ -0,0 +1,13 @@
+cask 'oki-postscript-drivers' do
```
vitorgalvao: `oki-postscript-driver` (no final “s”).

Casks/oki-postscript-drivers.rb
```diff
@@ -0,0 +1,13 @@
...
+  version '2.0.2'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

franklouwers: I am pretty sure the url changes for each version. But we can't be sure until OKI releases a new version. The URL is referenced at http://belgium.oki.com/support/printer/printer-drivers/detail.aspx?prodid=tcm:125-4613&driverid=tcm:125-152412-16 which does mention the exact version and filename...

vitorgalvao: > which does mention the exact version and filename

It does? Because the numbers in `url` are also in `homepage`, I see no indication of it being related to a version.

franklouwers: If you go to http://belgium.oki.com/support/printer/printer-drivers/detail.aspx?prodid=tcm:125-4613&driverid=tcm:125-152412-16 it lists the version (Version 2.0.2) twice, the date the file was created, the exact filename etc. I would assume that the Download link will change when they push a new version. But if you feel I should exclude the exact version and sha-hash, I'll do that. Your call :)

vitorgalvao: Seeing as it was last updated on `21/11/2014` and if it ever is updated again, it should be for compatibility reasons (i.e. once a year, maximum), I think it‘s OK to leave as is).

PR #24444, opened 2016-09-13
Add Chronograf v1.0.0

gunnaraasen:
### Checklist

The previous version of Chronograf is currently available on [homebrew-binary](https://github.com/Homebrew/homebrew-binary/blob/master/chronograf.rb) and will be deprecated once this is merged. This is my first time submitting a cask so please let me know if there's anything to fix/improve.

### DIFF
--- a/Casks/chronograf.rb
+++ b/Casks/chronograf.rb
@@ -1,0 +1,11 @@
+cask 'chronograf' do
+  version '1.0.0'
+  sha256 '2b4fe4975acf4d0c0de500d81bcbd1f4ac4ee1109ed6c9717bff571262b1d9bb'
+
+  url "https://dl.influxdata.com/chronograf/releases/chronograf-#{version}-darwin_amd64.tar.gz"
+  name 'Chronograf'
+  homepage 'https://www.influxdata.com/time-series-platform/chronograf/'
+  license :unknown
+
+  binary "chronograf-#{version}/chronograf-#{version}-darwin_amd64", target: 'chronograf'
+end


Casks/chronograf.rb
```diff
@@ -0,0 +1,17 @@
...
+  sha256 '2b4fe4975acf4d0c0de500d81bcbd1f4ac4ee1109ed6c9717bff571262b1d9bb'
+
+  url "https://dl.influxdata.com/chronograf/releases/chronograf-#{version}-darwin_amd64.tar.gz"
+  name 'Chronograf'
+  homepage 'https://www.influxdata.com/time-series-platform/chronograf/'
+  license :unknown
+
+  binary 'chronograf-1.0.0/chronograf-1.0.0-darwin_amd64', target: 'chronograf'
+
+  caveats <<-EOS.undent
```
vitorgalvao: We reserve `caveats` for matters related to the installation itself. Since this is related to usage, it should be removed.

Casks/chronograf.rb
```diff
@@ -0,0 +1,17 @@
+cask 'chronograf' do
+  version '1.0.0'
+  sha256 '2b4fe4975acf4d0c0de500d81bcbd1f4ac4ee1109ed6c9717bff571262b1d9bb'
+
+  url "https://dl.influxdata.com/chronograf/releases/chronograf-#{version}-darwin_amd64.tar.gz"
+  name 'Chronograf'
+  homepage 'https://www.influxdata.com/time-series-platform/chronograf/'
+  license :unknown
+
+  binary 'chronograf-1.0.0/chronograf-1.0.0-darwin_amd64', target: 'chronograf'
```
vitorgalvao: Like you did in `url`, `#{version}` should also be used here.

vitorgalvao:
Thank you! This is a nice first submission. Only a few things to improve, but nothing needs fixing.

Casks/chronograf.rb
gunnaraasen: What's the recommended way to provide usage info in a cask? I wasn't able to find a better place than `caveats`.

Casks/chronograf.rb
gunnaraasen: Thanks for pointing that out!

gunnaraasen:
Thanks for reviewing @vitorgalvao. I just had a question above on the correct place to add usage info.

Also, I have a `plist` definition but it's not packaged in the tarball (and unlikely to be added). Is there an alternate way to add it to a formula? I saw this approach, but it seems [hacky](https://github.com/caskroom/homebrew-cask/blob/5549357edc2459043fd279bcac779e4d12106004/Casks/wireshark-chmodbpf.rb#L23).

vitorgalvao:
As I mentioned, our `caveats` are a bit different and reserved just for installation-related matters, because we try to keep messages to a minimum. It’s tangent to [not being a discoverability service](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md): if you need commands to use the software, you’re supposed to know them from somewhere else, we intentionally do not give them to you.

As for the plist definition, it’s not something we do a lot, and it’s discouraged. Contrary to HB, our goal is to be [as close to the manual result as possible](https://github.com/caskroom/homebrew-cask/issues/13201), which means no applied patched and the like, unless absolutely necessary. Why is is needed in this case?

gunnaraasen:
Thanks for the explanation @vitorgalvao. We don't necessarily need the plist or install caveat, I'm happy shipping without them. They're just things we were previously including in the `homebrew-binary` formula which don't have a cask equivalent.

It would be nice if there was a way to link to the documentation from the formula.

I think this is ready to ship as is. Let me know if there are any other outstanding issues!

vitorgalvao:
Looks great! Thank you.

PR #24520, opened 2016-09-15
Update microsoft-office (15.26.0_160910)

toonetown:
When using `:latest` as the version and `:no_check` as the sum, it causes the file to be downloaded again - even if it is in cache.  Since MS Office in particular is quite a large application, this can be draining on network resources.  This also is the way that other casks are generally set up - specific versions are included in the cask where possible.

Also, the mechanism exists to flag certain applications as auto-updating (via the `auto_updates` stanza) so that was also added.

### DIFF
--- a/Casks/microsoft-office.rb
+++ b/Casks/microsoft-office.rb
@@ -1,13 +1,15 @@
 cask 'microsoft-office' do
-  version :latest
-  sha256 :no_check
+  version '15.26.0_160910'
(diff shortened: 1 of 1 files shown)


toonetown:
If this is not desired, feel free to close - I just found it useful when deploying office to multiple computers (I can share the cache directory and only download it once).

vitorgalvao:
We prefer versioned downloads anyway, so this is a welcome addition. Can be merged when Travis passes.

PR #24611, opened 2016-09-19
Update dingtalk to 3.0.1

ripples-alive:
#### Checklist

### DIFF
--- a/Casks/dingtalk.rb
+++ b/Casks/dingtalk.rb
@@ -3,7 +3,7 @@ cask 'dingtalk' do
   sha256 '770205d6af3e46a9ac9746dcda645d8125cd1b275c7c759bcc11451ecf8828e2'
 
   # taobaocdn.com/dingtalk-desktop was verified as official when first introduced to the cask
-  url "http://download.taobaocdn.com/dingtalk-desktop/Release/install/DingTalk_v#{version}.dmg"
+  url "https://download.alicdn.com/dingtalk-desktop/Release/install/DingTalk_v#{version}.dmg"
   name 'DingTalk'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Please run the commands we ask for in the checklist before marking them. The `style` commands wasn’t run.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 152952a88ae3af204f85d85f355696d1aa83863e. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

ripples-alive:
Sorry for the incorrect modifications. This is the first time I've tried to change the url, because of the unexpected network error happened on the travis test. Both the new and old url works well on my computer.
And thanks for your fixes. But I'm wonder whether the comment "when first introduced to the cask" need to be updated, too.

vitorgalvao:
You can still change the URL via `cask-repair` (give `e` when it asks if everything is correct or `--edit`/`-e` when first giving the command). `cask-repair` will run those commands automatically, and if it finds something it cannot fix, it’ll warn you (and depending on the problem, prevent the submissions from going through).

> But I'm wonder whether the comment "when first introduced to the cask" need to be updated, too.

That’s the beauty of the comment. We had another one previously, and this one is more accurate. It means “when the URL was introduced to the cask”. Since now you changed the URL, it was only introduced now. The goal is to inform users “when we changed to this URL, we confirmed it is correct, but it may not be now, use with caution”.

ripples-alive:
Got it. I've misunderstood the comment before.
Thanks for your detailed explanation.

PR #24705, opened 2016-09-21
Add Logos Bible Software 6.14.0.65

ickc:


Casks/logos-bible-software.rb
```diff
@@ -0,0 +1,11 @@
+cask 'logos-bible-software' do
```
vitorgalvao: The token reference was not followed. This needs to be `logos`.

vitorgalvao (changes requested):

Casks/logos-bible-software.rb
```diff
@@ -0,0 +1,11 @@
...
+  version '6.14.0.65'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

vitorgalvao (changes requested):

Casks/logos-bible-software.rb
ickc: Sorry I misunderstand it.

The one in the dmg is `Logos.app`, but the installed app is called `Logos Bible Software 6.app`. And I thought "logos" is too generic and people would thought it is the plural form of the English word logo, where there's another software named as logo which design logo. Rather, "logos" is a transliteration of "λόγος".

Anyway, I'm changing it to logos if you think that's right.

Casks/logos-bible-software.rb
ickc: Done and thanks

vitorgalvao (approved):

PR #24708, opened 2016-09-21
Upgrade Geekbench to v4.0.1

Saklad5:
- [x] `brew cask style —fix {{cask_file}}` left no offenses.

Fix url in Geekbench Cask
Fix appcast in Geekbench Cask
Fix app in Geekbench Cask
Add OS requirement in Geekbench Cask

### DIFF
--- a/Casks/geekbench.rb
+++ b/Casks/geekbench.rb
@@ -1,13 +1,15 @@
 cask 'geekbench' do
-  version '3.4.1'
-  sha256 '9f2730472bba9fd39554290f465d37c32792debc5b20c9840efd1f79d40ca94c'
+  version '4.0.1'
+  sha256 '7713d5760bcbf5577e076a758693d918abbf08bd7b51caaa9b6793c7063ea8f3'
 
-  url "https://cdn.primatelabs.com/Geekbench-#{version}-Mac.zip"
-  appcast 'https://www.primatelabs.com/appcast/geekbench3.xml',
-          checkpoint: '23877fa30558eada7e7beeefee4fa307c7a7ac3fad79b3c250fa09f4c68081c9'
+  url "https://cdn.primatelabs.com/Geekbench-#{version}-Mac.dmg"
+  appcast "https://www.primatelabs.com/appcast/geekbench#{version.major}.xml",
+          checkpoint: '51133594ed34b6ed7286e3b9c03321d0d7e555927f09ccfc2de7f73125a1cfc1'
   name 'Geekbench'
   homepage 'https://www.primatelabs.com/geekbench/'
   license :commercial
 
-  app 'Geekbench 3.app'
+  depends_on macos: '>= :yosemite'
+
+  app "Geekbench #{version.major}.app"
 end


Casks/geekbench.rb
```diff
@@ -1,13 +1,15 @@
...
-          checkpoint: '23877fa30558eada7e7beeefee4fa307c7a7ac3fad79b3c250fa09f4c68081c9'
+  url "https://cdn.primatelabs.com/Geekbench-#{version}-Mac.dmg"
+  appcast "https://www.primatelabs.com/appcast/geekbench#{version.major}.xml",
+          checkpoint: '51133594ed34b6ed7286e3b9c03321d0d7e555927f09ccfc2de7f73125a1cfc1'
   name 'Geekbench'
   homepage 'https://www.primatelabs.com/geekbench/'
   license :commercial
 
-  app 'Geekbench 3.app'
+  depends_on macos: '>= :yosemite'
```
reitermarkus: Could yu rather use an `if` statement to switch between version 3 and 4? Here's an example: https://github.com/caskroom/homebrew-cask/blob/adc5e08d919f61cbe8c96129c7f1379520711e07/Casks/pacifist.rb

reitermarkus (changes requested):

Casks/geekbench.rb
Saklad5: Why would that be better?

vitorgalvao: > Why would that be better?

Because users on older macOS versions will still be able to get it.

> Isn’t it preferable to future-proof the name?

You’ll still do it, since you’ll have `version` twice (one in each part of the conditional).

Saklad5:
Why not use the Legacy flag, which seems intended for this sort of thing?

vitorgalvao:
> Why not use the Legacy flag, which seems intended for this sort of thing?

Because why deny users on older systems the change to install the cask? Also, we don’t have a legacy flag.

Saklad5:
You're right, flag is the wrong term.

The documentation for adding a Cask does state that there is a "Legacy" type of binary under the project nomenclature, defined as "any stable version that is not the most recent". It states that those should go in caskroom/homebrew-versions.

Doesn't that describe this situation?

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 98544b682104b93ad93f20201dcc977744ae39f0. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

vitorgalvao:
Some situations are different (that’s why sometimes we use conditionals).

Version 4 can only be installed from Yosemite on, so users on it will *have to* use version 4. Users below Yosemite will be forced to use version 3.

If version 4 were available in the same system versions as version 3, then yes, version 3 would be moved to [caskroom/versions](https://github.com/caskroom/homebrew-versions).

Saklad5:
Understood. Perhaps the documentation needs clarification on that point?

vitorgalvao:
@Saklad5 This is not a common case of confusion, so I doubt it’s really needed, but we’re always open to improvements (it’s only important it’s not too long).

PR #24836, opened 2016-09-25
Update sha256sum of TuneInstructor v3.6

spookey:
### Checklist
- [x] The commit message includes the cask’s name and version.
- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.

The checksum changed

PR #24838, opened 2016-09-25
Update URL of TuneInstructor v3.6

spookey:
### Checklist
- [x] `brew cask audit --download tuneinstructor` is error-free.
- [x] `brew cask style --fix tuneinstructor` left no offenses.

I decided to open a new request instead of appending to #24836 
The real issue here was a change in the download URL.

### DIFF
--- a/Casks/tuneinstructor.rb
+++ b/Casks/tuneinstructor.rb
@@ -1,6 +1,6 @@
 cask 'tuneinstructor' do
   version '3.6'
-  sha256 '7354d7423d19d75aa590f338596fc17da8d1c9f341c55bd7da592d5aa1b18a8e'
+  sha256 '9f68e5d1d31a7a7502f84e21f1d26cd8fd3a8eba7708f07386aa641551592d14'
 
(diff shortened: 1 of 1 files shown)


Casks/tuneinstructor.rb
```diff
@@ -1,8 +1,8 @@
 cask 'tuneinstructor' do
   version '3.6'
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/f25b6babcd398abf48e33af3d887b2d00de1d661/Casks/kibako.rb).

vitorgalvao (changes requested):

adidalal:
Got the versioned URL - `https://www.tune-instructor.de/resources/downloads/TuneInstructor3.6.dmg`

spookey:
@adidalal the versioned url you mentioned has thrown me some "This Version only supports OS X 10.10 (Yosemite)." error yesterday.
Very strange.

I am very sorry about the mess, I created here, and very happy to see, the problem is fixed already (in #24865).

Issue #24888, opened 2016-09-26
Documentation about Caskroom


toonetown:
In the documentation, it lists that the default value for `--caskroom` is `$(brew --repository)/Caskroom` - when in reality, it is now `$(brew --prefix)/Caskroom`.

Not sure all the places this needs to be changed - or else I'd just submit a PR...

reitermarkus:
> Not sure all the places this needs to be changed - or else I'd just submit a PR...

Please feel free to open a PR even if you don't catch all the places.

PR #24891, opened 2016-09-26
Fix documentation for --caskroom default

toonetown:
As explained in #24888, the documentation is incorrect around the default value of `--caskroom`


adidalal (approved):

reitermarkus (changes requested):
There is also a line in https://github.com/caskroom/homebrew-cask/blob/master/USAGE.md#options where this is used.

reitermarkus:
I see you also opened https://github.com/Homebrew/brew/pull/1159, so you might as well delete the whole man page here, as this isn't actually used anymore to generate it.

reitermarkus:
Thanks!

PR #24934, opened 2016-09-28
Update NW.js to 0.17.5 and remove 32 bit version

joshka:

NW.js is only available for x64 as of v0.13. 32 bit version removed.

The docs suggest to use the SDK flavor rather than the standard (see
http://docs.nwjs.io/en/latest/For%20Users/Getting%20Started/#get-nwjs).
I'd envisage anyone wanting the standard ver should create a package in
homebrew/versions for this.

Added symlink to nwjc

### DIFF
--- a/Casks/nwjs.rb
+++ b/Casks/nwjs.rb
@@ -1,22 +1,12 @@
 cask 'nwjs' do
-  version '0.17.0'
+  version '0.17.5'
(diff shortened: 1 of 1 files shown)


Numbermaniac:
If an older 32-bit download still exists, would it be possible to keep the if-statement and have a separate version download for 32-bit?

vitorgalvao:
@Numbermaniac Snow Leopard was the last macOS release to support 32-bit. I’m even doubtful HBC still works on that version. There’s no point in keeping it around.

PR #24975, opened 2016-09-28
Add version and checksum to Dropbox.app - 11.4.20

colindunn:
### Checklist

The dropbox cask does not need to be an unspecified latest with no checksum because the version number is readily available at from the OSX download page redirect.

The following curl shows the download location and version as a redirect from Dropbox within the Location header:

```
curl -sI "https://www.dropbox.com/download?plat=mac&full=1"
```

Latest and last version links to show that specific versions are readily available:
- https://clientupdates.dropboxstatic.com/client/Dropbox%2010.4.26.dmg
- https://clientupdates.dropboxstatic.com/client/Dropbox%2011.4.20.dmg

### DIFF
--- a/Casks/dropbox.rb
+++ b/Casks/dropbox.rb
@@ -1,8 +1,9 @@
 cask 'dropbox' do
-  version :latest
-  sha256 :no_check
+  version '11.4.20'
+  sha256 '61267a63616308bf9dd59db8c38298f25b9ea302bcbc57a1451f71c5dee2e905'
 
-  url 'https://www.dropbox.com/download?plat=mac&full=1'
+  # clientupdates.dropboxstatic.com/client was verified as official when first introduced to the cask
+  url "https://clientupdates.dropboxstatic.com/client/Dropbox%20#{version}.dmg"
   name 'Dropbox'
   homepage 'https://www.dropbox.com/'
   license :gratis


miccal:
I am not sure about this change - my installed version of Dropbox is `10.4.26`, and how would one find newer versions to update the Cask, would you have to run `curl -sI "https://www.dropbox.com/download?plat=mac&full=1"` to check?

adidalal:
I use https://github.com/buo/homebrew-cask-upgrade

adidalal:
Ah, you mean for Cask updates? Version numbers are available on the dropbox forum "changelog"

vitorgalvao:
To get just the version number: `curl -sI "https://www.dropbox.com/download?plat=mac&full=1" | grep 'Location' | sed -E 's/.*%20(.+).dmg/\1/'`.

Though `11.4.26` isn’t there, so this method seems to produce outdated results. In addition, I’m ambivalent to this change as well. Dropbox is one app where we want to be sure we’re getting the real deal, so having a `sha256` (and a version) is useful. However, not only is this method of getting the latest version obtuse and in need to be explicitly run, Dropbox (unlike most auto-updating apps) updates silently.

This all means in the end this change might actually be harmful to security. Since we have no easy method to know a new version is available, this cask will probably become dangerously outdated. Seems like an outlier case where `:latest` and `:no_check` is actually more secure.

adidalal:
Wouldn't mind a revert if you feel strongly one way or another. I have no strong opinion on this one.

miccal:
My vote would be for a revert.

vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/25015.

colindunn:
Sorry I meant to reply on this earlier as it was my commit.

I made this because I feel more comfortable with the idea of a checksum on versions but I can see the reasoning here.

Is it worth making a comment within the cask itself to reflect this determination so that no one else makes the same update that I did?

vitorgalvao:
> I made this because I feel more comfortable with the idea of a checksum on versions

As do most users and we do prefer that, but this is indeed a special case.

> Is it worth making a comment within the cask itself to reflect this determination so that no one else makes the same update that I did?

No, because I don’t recall it having happened before in all the time the cask exists.

colindunn:
Thanks Vítor

PR #25013, opened 2016-09-29
Update panoply.rb

wjcapehart:
Updated Panoply to v 4.6.0


adidalal:
Thanks for the PR, but it looks like you have some whitespace issues. Merged https://github.com/caskroom/homebrew-cask/pull/25042 instead, which is essentially the same update.

Check out `cask-repair`, it makes Cask updates like this quite easy (details in [CONTRIBUTING](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md))

PR #25015, opened 2016-09-29
Revert "Add version and checksum to Dropbox.app - 11.4.20"

vitorgalvao:
See https://github.com/caskroom/homebrew-cask/pull/24975#issuecomment-250439429 for context.

Not merging yet, in case someone has a good point to not revert this.


reitermarkus:
👍 , as I think Dropbox is probably the only cask where `auto_updates` actually is true.

PR #25054, opened 2016-09-30
Update pdfexpert to 2.0.1

pat-s:
#### Checklist


vitorgalvao:
This cask has an unversioned URL and no appcast. That means it’ll simply break when an update is released upstream, if we give it a version.

PR #25064, opened 2016-09-30
Fix macOS requirement in Sierra Cache Cleaner.app Cask

Saklad5:
### Checklist

This application is supported on OS X 10.4 and later, but is hard-coded to not work on anything newer than macOS Sierra.

### DIFF
--- a/Casks/sierra-cache-cleaner.rb
+++ b/Casks/sierra-cache-cleaner.rb
@@ -8,7 +8,7 @@ cask 'sierra-cache-cleaner' do
   homepage 'http://www.northernsoftworks.com/sierracachecleaner.html'
   license :commercial
 
-  depends_on macos: :sierra
+  depends_on macos: '<= :sierra'
 
   app 'Sierra Cache Cleaner.app'
 end


Casks/sierra-cache-cleaner.rb
```diff
@@ -8,7 +8,7 @@
   homepage 'http://www.northernsoftworks.com/sierracachecleaner.html'
   license :commercial
 
-  depends_on macos: :sierra
+  depends_on macos: '<= :sierra'
```
vitorgalvao: This makes no sense. If you want to do this, simply delete the line.

vitorgalvao (changes requested):

Casks/sierra-cache-cleaner.rb
reitermarkus: The thing is that it will stop working on `> :sierra`.

reitermarkus: Then again, it is a `:latest` cask, so it will likely get updated in time.

vitorgalvao:
> This application is supported on OS X 10.4 and later, but is hard-coded to not work on anything newer than macOS Sierra.

Yes, because [you pointed out it didn’t work in other versions](https://github.com/caskroom/homebrew-cask/pull/24748#discussion_r80224245). So it does work in older versions? Even with `Sierra` specifically in the name?

Saklad5:
It is incredibly confusing. This is the support page detailing this: http://www.northernsoftworks.com/support.html

As you can see, each version of the app stops working when a new major release of macOS comes out.

reitermarkus:
This has to be the worst naming convention I have seen so far:

> Download Sierra Cache Cleaner 11.0.1 (72.9 Mb) macOS X 10.4-10.12.1
> Download El Capitan Cache Cleaner 10.0.6 (61.2 Mb) OS X 10.4 - 10.11.6
> Download Yosemite Cache Cleaner 9.0.8 (58.2 Mb) OS X 10.4-10.10.5
> Download Mavericks Cache Cleaner 8.0.9 (31.1 Mb) OS X 10.4-10.9.5
> Download Mountain Lion Cache Cleaner 7.0.10 (20.4 Mb) OS X 10.4-10.8
> Download Lion Cache Cleaner 6.0.13 (19.3 Mb) OS X 10.4-10.7
> Download Snow Leopard Cache Cleaner 5.0.22 (18.7 Mb) OS X 10.3.9-10.6
> Download Leopard Cache Cleaner 4.0.27 (14 Mb) OS X 10.3.9 - 10.5
> Download Tiger Cache Cleaner 3.2.19 (9.4 Mb) OS X 10.3.9 - 10.4
> Download Panther Cache Cleaner 2.4.8 (3.6 Mb) OS X 10.1 - 10.3
> Download TCC 3.0.7 for OS X 10.1

Saklad5:
It does a lot of stuff that is practically guaranteed to screw up with each major OS update, so the developers basically rework the whole thing and sell it separately.

As it could fail catastrophically on macOS versions it doesn't support, it is pre-emptively hard-coded to not work on the next major release.

vitorgalvao:
This one will be bothersome to update. Not looking forward to having a bunch of similarly named casks filling the repo, or renaming this one every time. Not sure we should keep this one.

Saklad5:
Really? It shouldn't be difficult to maintain. It's not like this cask will break when macOS is updated again, and we could always move older ones to /versions if clutter becomes an issue.




            Yours truly,
                              Jeremy Saklad

On Sep 30, 2016, at 3:51 PM, Vítor Galvão <notifications@github.com> wrote:

This one will be bothersome to update. Not looking forward to having a bunch of similarly named casks filling the repo, or renaming this one every time. Not sure we should keep this one.

vitorgalvao:
I didn’t say it would be difficult, I said it would be a bother.

Yes, of course, we have the versions repo, but it feels like we’d be treating it as a bit of a dumping ground. However, I just remembered [we have rules to prevent that](https://github.com/caskroom/homebrew-versions#acceptable-casks), so it’s fine.

vitorgalvao (approved):

PR #25077, opened 2016-10-01
Add lightkey 1.6.2.

knpwrs:
http://lightkeyapp.com/en/

Please note that the package also installs `com.FTDI.ftdiusbserialdriverinstaller.FTDIUSBSerialDriver.pkg`, however, this is used by other lighting software so I didn't want to add it to the uninstall clause. Maybe a comment in `lightkey.rb` would be appropriate?
### Checklist
- [x] `brew cask audit --download lightkey` is error-free.
- [x] `brew cask style --fix lightkey` left no offenses.

- [x] `brew cask install lightkey` worked successfully.
- [x] `brew cask uninstall lightkey` worked successfully.

### DIFF
--- a/Casks/lightkey.rb
+++ b/Casks/lightkey.rb
@@ -1,0 +1,19 @@
+cask 'lightkey' do
+  version '1.6.2'
+  sha256 '06cf4a9b9c609a91cff69f1fccbf73df0f81a4efdbf223aa92205aec74f43027'
+
+  url "http://lightkeyapp.com/content/06-download/Lightkey-#{version.dots_to_hyphens}/LightkeyInstaller.zip"
+  name 'Lightkey'
+  homepage 'http://lightkeyapp.com'
+  license :commercial
+
+  pkg 'LightkeyInstaller.pkg'
+
+  uninstall pkgutil: [
+                       'de.monospc.lightkey.pkg.App',
+                       'de.monospc.lightkey.pkg.Eurolite',
+                       'de.monospc.lightkey.pkg.OLA',
+                       'de.monospc.lightkey.pkg.Velleman',
+                       'de.monospc.lightkey.pkg.documentation',
+                     ]
+end


Casks/lightkey.rb
```diff
@@ -0,0 +1,19 @@
+cask 'lightkey' do
+  version '1.6.2'
+  sha256 '06cf4a9b9c609a91cff69f1fccbf73df0f81a4efdbf223aa92205aec74f43027'
+
+  url "http://lightkeyapp.com/content/06-download/Lightkey-#{version.major}-#{version.minor}-#{version.patch}/LightkeyInstaller.zip"
```
reitermarkus: You can use `version.dots_to_hyphens` here.

reitermarkus (changes requested):

Casks/lightkey.rb
knpwrs: All set.

knpwrs:
Would we like to add a note about not uninstalling `com.FTDI.ftdiusbserialdriverinstaller.FTDIUSBSerialDriver.pkg`? That's the standard driver used by most (all?) lighting software.

adidalal:
Not sure what the right place for that would be, maybe @vitorgalvao has an opinion?

I'd be fine with it in the uninstall section, as another entry in the uninstall: pkgutil....

`# com.FTDI.ftdiusbserialdriverinstaller.FTDIUSBSerialDriver is used by other lighting software`

knpwrs:
Looks like that package is actually available standalone as a cask: https://github.com/caskroom/homebrew-cask/blob/6759c4427a303631c52bd401364e93d15f17db31/Casks/ftdi-vcp-driver.rb

vitorgalvao:
[The place for it is the `zap` stanza](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/zap.md):

> `zap` stanzas may remove:
> (…)
> Shared resources such as application updaters. Since shared resources may be removed, other applications may be affected by `brew cask zap`. Understanding that is the responsibility of the end user.

knpwrs:
Side note, when I first saw the `zap` command I thought it was just a regular uninstall. Probably because I was an [App Zapper](https://www.appzapper.com/) user prior to installing most software with Cask.

I'll open a PR to get that package in the zap stanza.

PR #25084, opened 2016-10-01
Repair provided apps and scripts for launching j 804

00dani:
According to the readme.txt shipped with the J release, a J installation should add to PATH four commands: `jbrk`, `jcon`, `jhs`, and `jqt`. However none of these were being added to PATH previously, although `jcon` was available under the alternative name `jconsole`. This pull req adds the four specified binaries, preserving the original `jconsole` symlink for compatibility reasons.

Additionally, J provides four .app bundles which launch the four commands mentioned above. These all worked using relative paths, which means that when homebrew-cask moved them into /Applications they all broke. I have added a postflight section which updates the .app bundles on the fly, so that they use an absolute path and therefore work correctly again.

Finally I have added an installer stanza which invokes the `updatejqt.sh` script provided in the J distribution. This is required for jqt to work and indeed for homebrew-cask to generate a symlink to it properly. As this step takes a substantial amount of additional installation time, and jqt is far from required for working in J, I'd suggest we consider making jqt optional if possible. (If jqt is not required, then the script needn't run and neither the jqt command nor jqt.app need be installed.)
### Checklist
- [x] The commit message includes the cask’s name and version. (Note that these issues were _not_ caused by the upgrade to J804; the J802 cask had all the same problems.)
- [x] `brew cask audit --download j` is error-free.
- [x] `brew cask style --fix j` left no offenses.

### DIFF
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -11,7 +11,9 @@ cask 'j' do
     app "j64-#{version}/#{a}.app"
   end
 
-  %w[jconsole].each do |b|
-    binary "j64-#{version}/bin/#{b}"
+  # We have long provided jconsole on the path. However readme.txt specifies
+  # that jconsole is available under the name jcon. Just provide both names.
+  %w[jcon jconsole].each do |b|
+    binary "j64-#{version}/bin/jconsole", target: b
   end
 end
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -12,14 +12,16 @@ cask 'j' do
     app "j64-#{version}/#{a}.app"
   end
 
+  installer sudo: false, script: "j64-#{version}/updatejqt.sh"
+
   # We have long provided jconsole on the path. However readme.txt specifies
   # that jconsole is available under the name jcon. Just provide both names.
   %w[jcon jconsole].each do |b|
     binary "j64-#{version}/bin/jconsole", target: b
   end
 
-  # Provide jbrk and jhs on the path too, as readme.txt specifies.
-  %w[jbrk jhs].each do |b|
+  # Provide jbrk, jhs, and jqt on the path too, as readme.txt specifies.
+  %w[jbrk jhs jqt].each do |b|
     binary "j64-#{version}/bin/#{b}.command", target: b
   end
 
@@ -35,5 +37,11 @@ cask 'j' do
           #{'open' if cli} "#{@cask.staged_path}/j64-#{version}/bin/#{command}"
       EOF
     end
+
+    # jqt.command has a similar problem: it uses dirname $0 expecting to find
+    # J's binary directory, but will find the Homebrew binary directory instead
+    # if we use a symlink to jqt.command. Instead, use a symlink to the wrapper
+    # we just generated, which uses an absolute path.
+    FileUtils.ln_sf "#{@cask.appdir}/jqt.app/Contents/MacOS/apprun", "#{Hbc.binarydir}/jqt"
   end
 end
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -16,4 +16,9 @@ cask 'j' do
   %w[jcon jconsole].each do |b|
     binary "j64-#{version}/bin/jconsole", target: b
   end
+
+  # Provide jbrk and jhs on the path too, as readme.txt specifies.
+  %w[jbrk jhs].each do |b|
+    binary "j64-#{version}/bin/#{b}.command", target: b
+  end
 end
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -15,12 +15,10 @@ cask 'j' do
   installer script: "j64-#{version}/updatejqt.sh",
             sudo:   false
 
-  # We have long provided jconsole on the path. However readme.txt specifies
-  # that jconsole is available under the name jcon. Just provide both names.
+  # target names according to readme.txt
   %w[jcon jconsole].each do |b|
     binary "j64-#{version}/bin/jconsole", target: b
   end
-  # Provide jbrk, jhs, and jqt on the path too, as readme.txt specifies.
   commands = %w[jbrk jhs jqt]
   commands.each do |b|
     binary "j64-#{version}/bin/#{b}.command", target: b
@@ -30,7 +28,7 @@ cask 'j' do
     # Use `readlink` to get full path of symlinked commands.
     commands.each do |c|
       command = "#{staged_path}/j64-#{version}/bin/#{c}.command"
-      IO.write command, IO.read(command).gsub('$0', '$(readlink "$0" || echo "$0")')
+      IO.write command, IO.read(command).gsub('$0', '$(/usr/bin/readlink "$0" || /bin/echo "$0")')
     end
 
     # Fix relative paths inside App bundles.
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -14,30 +14,29 @@ cask 'j' do
 
   installer script: "j64-#{version}/updatejqt.sh",
             sudo:   false
+
   # We have long provided jconsole on the path. However readme.txt specifies
   # that jconsole is available under the name jcon. Just provide both names.
   %w[jcon jconsole].each do |b|
     binary "j64-#{version}/bin/jconsole", target: b
   end
-  # Provide jbrk and jhs on the path too, as readme.txt specifies.
-  %w[jbrk jhs jqt].each do |b|
+  # Provide jbrk, jhs, and jqt on the path too, as readme.txt specifies.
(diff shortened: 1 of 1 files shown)


adidalal:
First off, thanks for submitting this detailed PR - I personally don't have experience with (or a use case for) J, so you're really helping out other J users.

Unfortunately, Homebrew Cask doesn't have option flags, so having `jqt` be optional is not possible.

However, it looks like J distributes a `jqt` included `.dmg`, as I posted on gitter, so that might be a better starting point rather than shelling out to `updatejqt.sh`

Will leave this unmerged for other maintainers to also take a look

Casks/j.rb
```diff
@@ -7,11 +7,41 @@
...
+          #!/bin/sh
+          #{'open' if cli} "#{@cask.staged_path}/j64-#{version}/bin/#{command}"
+      EOF
+    end
+
+    # jqt.command has a similar problem: it uses dirname $0 expecting to find
+    # J's binary directory, but will find the Homebrew binary directory instead
+    # if we use a symlink to jqt.command. Instead, use a symlink to the wrapper
+    # we just generated, which uses an absolute path.
+    FileUtils.ln_sf "#{@cask.appdir}/jqt.app/Contents/MacOS/apprun", "#{Hbc.binarydir}/jqt"
```
reitermarkus: You can put this to the binaries above like:
```ruby
binary "#{appdir}/jqt.app/Contents/MacOS/apprun", target: 'jqt'`
```

Casks/j.rb
```diff
@@ -7,11 +7,41 @@
...
 
-  %w[jbrk jcon jhs jqt].each do |a|
+  apps = %w[jbrk jcon jhs jqt]
+  apps.each do |a|
     app "j64-#{version}/#{a}.app"
   end
 
-  %w[jconsole].each do |b|
-    binary "j64-#{version}/bin/#{b}"
+  installer sudo: false, script: "j64-#{version}/updatejqt.sh"
```
reitermarkus:
```ruby
installer script: "j64-#{version}/updatejqt.sh",
          sudo:   false
```

Casks/j.rb
```diff
@@ -7,11 +7,41 @@
...
+  end
+
+  postflight do
+    # All four of the Mac apps use a relative path to launch the underlying
+    # binaries, so none of them work after homebrew-cask moves them into
+    # /Applications.
+    apps.each do |a|
+      cli = %w[jcon jhs].include? a
+      command = a == 'jcon' ? 'jconsole' : a + '.command'
+      File.write "#{@cask.appdir}/#{a}.app/Contents/MacOS/apprun", <<-EOF.gsub(%r{^ +}, '')
```
reitermarkus: Use only `appdir`, not `@cask.appdir`.

Casks/j.rb
```diff
@@ -7,11 +7,41 @@
...
+          #!/bin/sh
+          #{'open' if cli} "#{@cask.staged_path}/j64-#{version}/bin/#{command}"
```
reitermarkus: Use only `staged_path `, not `@cask.staged_path` and `#{cli ? 'open ' : ''}`.

reitermarkus (changes requested):

Casks/j.rb
```diff
@@ -7,11 +7,41 @@
...
+    binary "j64-#{version}/bin/#{b}.command", target: b
+  end
+
+  postflight do
+    # All four of the Mac apps use a relative path to launch the underlying
+    # binaries, so none of them work after homebrew-cask moves them into
+    # /Applications.
+    apps.each do |a|
+      cli = %w[jcon jhs].include? a
+      command = a == 'jcon' ? 'jconsole' : a + '.command'
```
reitermarkus: Not sure if this should be

```ruby
a == 'jcon' ? 'jconsole' : (a + '.command')
```

or 

```ruby
(a == 'jcon' ? 'jconsole' : a) + '.command'
```

reitermarkus (changes requested):

Casks/j.rb
00dani: You want to `open` the CLI ones so that MacOS opens a terminal to show them in. Remember, all of these are .app bundles - they'll be launched from the GUI and there won't be a terminal available.

00dani:
Just pushed a commit addressing @reitermarkus's change requests. I'm… not sure how to tell GitHub that I've done that?

reitermarkus:
> I'm… not sure how to tell GitHub that I've done that?

You can't yourself, maintainers have an “Approve changes” button for that.

Casks/j.rb
```diff
@@ -7,11 +7,37 @@
...
+  end
+  # Provide jbrk and jhs on the path too, as readme.txt specifies.
+  %w[jbrk jhs jqt].each do |b|
+    binary "j64-#{version}/bin/#{b}.command", target: b
+  end
+  # The provided jqt.command won't work if run through a symlink, so instead
+  # we'll symlink the script inside jqt.app into the path.
+  binary "#{appdir}/jqt.app/Contents/MacOS/apprun", target: 'jqt'
+
+  postflight do
```
reitermarkus: Ok, I tinkered a bit with this. Replace the whole `postflight` block with:

```ruby
  postflight do
    # use `readlink` to get full path of symlinked commands
    commands.each do |c|
      command = "#{staged_path}/j64-#{version}/bin/#{c}.command"
      IO.write command, IO.read(command).gsub('$0', '$(readlink "$0")')
    end

    # fix relative paths inside App bundles
    apps.each do |a|
      apprun = "#{appdir}/#{a}.app/Contents/MacOS/apprun"
      IO.write apprun, IO.read(apprun).gsub(%r{`dirname "\$0"`.*?/bin}, "#{staged_path}/j64-#{version}/bin")
    end
  end
```

reitermarkus (changes requested):

Casks/j.rb
```diff
@@ -7,11 +7,37 @@
...
+  %w[jcon jconsole].each do |b|
+    binary "j64-#{version}/bin/jconsole", target: b
+  end
+  # Provide jbrk and jhs on the path too, as readme.txt specifies.
+  %w[jbrk jhs jqt].each do |b|
+    binary "j64-#{version}/bin/#{b}.command", target: b
+  end
+  # The provided jqt.command won't work if run through a symlink, so instead
+  # we'll symlink the script inside jqt.app into the path.
+  binary "#{appdir}/jqt.app/Contents/MacOS/apprun", target: 'jqt'
```
reitermarkus: You can also get rid of this line if you replace the `postflight` block.

reitermarkus (changes requested):

Casks/j.rb
00dani: `readlink` prints nothing if its argument is *not* a symbolic link, which means the app bundles don't work since they don't point to the symlink. Using `$(readlink "$0" || echo "$0")` appears to work however, so I'm gonna do that.

reitermarkus: > readlink prints nothing if its argument is not a symbolic link.

Didn't know that, good to know. It seems like `realpath` does what we need, though.

00dani: `realpath` does what we need, yes, but `realpath` is a GNU coreutils command so it might not be available on the system. `readlink` however is available on standard MacOS.

00dani: I'd be surprised if anyone using Homebrew *hasn't* installed GNU coreutils, but it's possible!

Casks/j.rb
```diff
@@ -7,11 +7,36 @@
...
     app "j64-#{version}/#{a}.app"
   end
 
-  %w[jconsole].each do |b|
-    binary "j64-#{version}/bin/#{b}"
+  installer script: "j64-#{version}/updatejqt.sh",
+            sudo:   false
+
+  # We have long provided jconsole on the path. However readme.txt specifies
+  # that jconsole is available under the name jcon. Just provide both names.
```
reitermarkus: You could probably use a single `# target names according to readme.txt` comment for all the `binary` stanzas.

Casks/j.rb
reitermarkus: Hah, right!

```
/usr/local/opt/coreutils/libexec/gnubin/realpath
```

We'll leave it at `$(readlink "$0" || echo "$0")` then.

reitermarkus: Although let's hard-code `/usr/bin/readlink` to always use the system one.

Casks/j.rb
00dani: Sure thing. We do still need two loops though, since `jconsole` doesn't have a `.command` suffix like the others do.

reitermarkus (approved):

reitermarkus:
Thanks!

PR #25174, opened 2016-10-04
Update texpad to 1.7.32__104__a6eb552

chohner:
#### Checklist

### DIFF
--- a/Casks/texpad.rb
+++ b/Casks/texpad.rb
@@ -3,13 +3,13 @@ cask 'texpad' do
     version '1.7.21'
     sha256 'cae312b89d25752917bf66572fd1c6b6722097b651abe9a1a1b399dc36493512'
   else
-    version '1.7.28'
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):

chohner:
I couldn't figure out the hash, maybe there's a cleaner way to do it - but at least it now resolves correctly.

vitorgalvao:
> I couldn't figure out the hash, maybe there's a cleaner way to do it

What do you mean?

chohner:
1.7.32__104__a6eb552 should probably be something like 1.7.32 (104) and 'a6eb552' some sort of hash that is now hard coded so the version string is pretty ugly.

vitorgalvao:
Oh, don’t worry about it. We care about versions more in the sense of how they relate to the URL than just by themselves. The way you did it is fine.

PR #25280, opened 2016-10-07
Collection of crowbar brute-force tools for DMG, KeyChains, and PGP

ncovercash:

### DIFF
--- a/Casks/crowbarpgp.rb
+++ b/Casks/crowbarpgp.rb
@@ -2,13 +2,13 @@ cask 'crowbarpgp' do
   version '1.0.1'
   sha256 :no_check
 
-  url "http://www.georgestarcher.com/crowbarPGP/crowbarPGP-v1.0.1.dmg"
+  url 'http://www.georgestarcher.com/crowbarPGP/crowbarPGP-v1.0.1.dmg'
   name 'crowbarPGP'
   homepage 'http://www.georgestarcher.com/software/'
   license :unknown
 
-  app 'crowbarPGP.app'
-
   # developer says that he doesn't plan to recreate for sierra
   depends_on macos: '<= :el_capitan'
+
+  app 'crowbarPGP.app'
 end
--- a/Casks/crowbarpgp.rb
+++ b/Casks/crowbarpgp.rb
@@ -12,3 +12,4 @@ cask 'crowbarpgp' do
 
   app 'crowbarPGP.app'
 end
+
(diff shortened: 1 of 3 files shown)


Casks/crowbarpgp.rb
```diff
@@ -0,0 +1,15 @@
...
+  name 'crowbarPGP'
+  homepage 'http://www.georgestarcher.com/software/'
+  license :unknown
+
+  # developer says that he doesn't plan to recreate for sierra
+  depends_on macos: '<= :el_capitan'
+
+  app 'crowbarPGP.app'
+end
+
```
vitorgalvao: This does *not* pass `style`. Remove the empty line. You need to run the commands after all the changes.

Casks/crowbarpgp.rb
```diff
@@ -0,0 +1,15 @@
+cask 'crowbarpgp' do
+  version '1.0.1'
+  sha256 :no_check
+
+  url 'http://www.georgestarcher.com/crowbarPGP/crowbarPGP-v1.0.1.dmg'
```
vitorgalvao: `url "http://www.georgestarcher.com/crowbarPGP/crowbarPGP-v#{version}.dmg"`

Casks/crowbarpgp.rb
```diff
@@ -0,0 +1,15 @@
+cask 'crowbarpgp' do
+  version '1.0.1'
+  sha256 :no_check
```
vitorgalvao: This needs to be set. Run `shasum -a 256` on the downloaded file.

Casks/crowbarpgp.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  url 'http://www.georgestarcher.com/crowbarPGP/crowbarPGP-v1.0.1.dmg'
+  name 'crowbarPGP'
+  homepage 'http://www.georgestarcher.com/software/'
+  license :unknown
+
+  # developer says that he doesn't plan to recreate for sierra
```
vitorgalvao: Remove comment.

vitorgalvao:
Please submit separate pull requests for each cask, as more often than not this type of pull request brings problems with one or more of them, and it halts the inclusion of the other ones. In this case, all of them have the same issues (only comment on one).

vitorgalvao (changes requested):

vitorgalvao:
In the future, please submit separate pull requests for each cask, as more often than not this type of pull request brings problems with one or more of them, and it halts the inclusion of the other ones. In this case, all of them have the same issues (only comment on one).

All that said, if this won’t be updated, there’s no point in adding casks that’ll only get more outdated over time. We don’t do that.

If you want these as casks, you’ll have to make your own tap.

PR #25357, opened 2016-10-10
Add mediaconch-gui 16.09

g-maxime:

- [ ] Named the cask according to the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
  -->  mediaconch name usually refers to the CLI package.
- [X] Checked there are no [open pull requests](https://github.com/caskroom/homebrew-cask/pulls) for the same cask.
- [X] Checked that the cask was not already refused in [closed issues](https://github.com/caskroom/homebrew-cask/issues?q=is%3Aissue+is%3Aclosed).


Casks/mediaconch-gui.rb
```diff
@@ -0,0 +1,13 @@
+cask 'mediaconch-gui' do
```
vitorgalvao: Make it `mediaconch-app`. It’ll either be that or have no suffix. We’re still deciding.

Casks/mediaconch-gui.rb
```diff
@@ -0,0 +1,13 @@
...
+  version '16.09'
+  sha256 '585b31fb6ec46bd32800ce2f392d2eec3ac0fba9da878be8f448091a5cc6a192'
+
+  url "https://mediaarea.net/download/binary/mediaconch-gui/#{version}/MediaConch_GUI_#{version}_Mac.dmg"
+  appcast 'https://github.com/MediaArea/MediaConch_SourceCode/releases.atom',
+          checkpoint: 'b336f179af85cac73cee05dff15b206f2bb7c2892414a04f5be45d8aafe11954'
+  name 'MediaConch'
+  homepage 'https://mediaarea.net/MediaConch/'
+  license :gpl
```
vitorgalvao: Remove `license`, it has been deprecated and is no longer in use.

vitorgalvao (changes requested):

Casks/mediaconch-gui.rb
vitorgalvao: > if a CLI application is acceptable in you project it will take the name.

Yes, they are very much acceptable. Is that a `pkg` just with the CLI, or a `pkg` with both the CLI and the `app`? That would be very much acceptable, yes, and preferred over this one.

> We also have

Does that mean you’re one of the developers?

g-maxime: The dmg with the pkg contains only the CLI.

I work for the editor.

g-maxime: This app do not need the CLI.

vitorgalvao: Let me clarify. Does this app have basically the same set of features as the CLI, but as a GUI, or do the GUI and CLI do basically different things, and they complement each other (meaning there’re things you can do with one but not with the other, and vice-versa).

And is there a “main” one between the two, one that drives the development of the other?

g-maxime: Basically the same functionalities, a little more options in the CLI but nothing concerning the main features.

adidalal: This should be fine as -app, if you plan on putting the cli into Homebrew

vitorgalvao (approved):

PR #25368, opened 2016-10-10
remove license from all casks

vitorgalvao:


raindev:
Why do you decided to get rid of the license information?

vitorgalvao:
Because it was useless and only holded up PRs.

raindev:
OK, thanks for your reply

PR #25417, opened 2016-10-12
Fix helm 2.0.0-alpha.5 sha256.

tback:

It seems as if the project updated the binary. This provides the new hash.
Waiting for confirmation that the release was changed.


tback:
The new hash is the one issued by the helm project: http://storage.googleapis.com/kubernetes-helm/helm-v2.0.0-alpha.5-darwin-amd64.tar.gz.sha256. We therefore don't need to wait any longer for confirmation if the release was changed.

PR #25419, opened 2016-10-12
generate_issue_template_urls: don't end in newline

vitorgalvao:
I usually run this script with `| pbocopy`. The copied newline is always annoying as it moves other text more than it should.

### DIFF
--- a/developer/bin/generate_issue_template_urls
+++ b/developer/bin/generate_issue_template_urls
@@ -22,7 +22,7 @@ BASE_URL = "https://github.com/caskroom/homebrew-cask/issues/new".freeze
 def main(args)
   args.each do |file|
     File.read(file).scan(%r{(.*?)\n(.*)}m) do |title, body|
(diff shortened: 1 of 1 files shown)


developer/bin/generate_issue_template_urls
```diff
@@ -22,7 +22,7 @@ BASE_URL = "https://github.com/caskroom/homebrew-cask/issues/new".freeze
 def main(args)
   args.each do |file|
     File.read(file).scan(%r{(.*?)\n(.*)}m) do |title, body|
-      puts generate_url(title, body)
+      print generate_url(title, body)
```
jawshooah: This breaks the output of the script when passed multiple files. However, this script is probably only ever called with one file anyway, so we could just rename it to `generate_issue_template_url` and update it to handle only one arg.

jawshooah (changes requested):

jawshooah:
Done in #25424.

PR #25457, opened 2016-10-13
Update synergy to 1.8.4,a6ff9079

miketheman:


miketheman:
Turns out this is incorrect. Apologies! I was so excited that I was able to get the submitter running.

The cask-repair downloaded an html file and checksummed that instead of the dmg.

PR #25498, opened 2016-10-14
Update cask to use Little Flocker Latest

securitygeneration:
- [√ ] `brew cask audit --download {{cask_file}}` is error-free.
- [√ ] `brew cask style --fix {{cask_file}}` reports no offenses.
- [√ ] The commit message includes the cask’s name and version.

### DIFF
--- a/Casks/little-flocker.rb
+++ b/Casks/little-flocker.rb
@@ -1,13 +1,13 @@
 cask 'little-flocker' do
-  version '0.99.90'
-  sha256 'daea3cabb3b900be9330bb67c34d2bcc3c12e88fd08b3d0c6bdcd264f98475ed'
+  version :latest
+  sha256 :no_check
 
-  # zdziarski.com was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


vitorgalvao:
This branch has conflicts.

Also, we prefer versioned casks, like the one you submitted previously.

securitygeneration:
I'm not sure why it ended up with conflicts. Could sort that out. As with versioning, I'm not a member of the Little Flocker project and may not be able to maintain this cask on a timely basis, hence my preference for a Latest approach. I'd rather that than an outdated cask.

vitorgalvao:
> I'm not a member of the Little Flocker project and may not be able to maintain this cask on a timely basis

Most casks are not maintained by the developers, and no cask has a set maintainer.

> I'd rather that than an outdated cask.

We have policies, and don’t change them based on a single user’s preference. Our users in the past came in overwhelming support of versioned casks for security reasons, and so we stick with those.

securitygeneration:
Understood.

PR #25524, opened 2016-10-15
Fix Snow Leopard installation in Sierra Cache Cleaner.app Cask

Saklad5:

  Version is unchanged.

### DIFF
--- a/Casks/sierra-cache-cleaner.rb
+++ b/Casks/sierra-cache-cleaner.rb
@@ -8,6 +8,9 @@ cask 'sierra-cache-cleaner' do
   homepage 'http://www.northernsoftworks.com/sierracachecleaner.html'
 
   depends_on macos: '<= :sierra'
-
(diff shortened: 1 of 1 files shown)


Saklad5:
This Cask downloads a disk image with both a legacy version and a main version of the application. Should the Cask delete the unused app from Caskroom, or just keep it there taking up 70MB?

Saklad5:
Sierra Cache Cleaner has additional components (ClamAV module, Single User Mode script, etc.) that are only downloaded and installed once you open the program and tell it to do so. Should this be mentioned as a caveat?

adidalal:
Caveats are not used for those kinds of messages. You could potentially add in the deletion in a postflight, but personally I don't think it's a huge deal.

adidalal (approved):

reitermarkus:
If it's not installed during installation, but when using the program, it should be a `zap`.

PR #25557, opened 2016-10-16
Add locateme v0.2.1 formula

beaumartinez:

- [x] Named the cask according to the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] Checked there are no [open pull requests](https://github.com/caskroom/homebrew-cask/pulls) for the same cask.
- [x] Checked that the cask was not already refused in [closed issues](https://github.com/caskroom/homebrew-cask/issues?q=is%3Aissue+is%3Aclosed).

---

RE the failing audit, the `url` is for a SourceForce project which includes many other downloads, not just LocateMe, hence why it's complaining (I think). I've seen discussion in issues that the auditing for SourceForge URLs is quite strict.

I hope this is all OK. Pushing this just before bed! ✌️ 

LocateMe is a nice, succinct command line tool to get your location from the built-in Mac services. It allows you to specify the format you want it in, not like other tools, so is especially handy if you're piping it.

### DIFF
--- a/Casks/locateme.rb
+++ b/Casks/locateme.rb
@@ -1,0 +1,10 @@
+cask 'locateme' do
+  version '0.2.1'
+  sha256 '137016e6c1a847bbe756d8ed294b40f1d26c1cb08869940e30282e933e5aeecd'
+
+  url "http://downloads.sourceforge.net/project/iharder/locateme/LocateMe-v#{version}.zip"
+  name 'LocateMe'
+  homepage 'http://iharder.sourceforge.net/current/macosx/locateme/'
+
+  binary 'LocateMe'
+end


vitorgalvao:
> I've seen discussion in issues that the auditing for SourceForge URLs is quite strict.

And it is so for a reason. Because bad URLs (location-specific) kept being submitted.

Either way, thank you for the submission, but this is a CLI-only open-source tool. Please submit it first to [Homebrew](https://github.com/Homebrew/homebrew-core) and only if they don’t accept it, submit here (linking back to the pull request where it was refused).

> The app is both open-source and CLI-only (i.e. it only uses the `binary` artifact). In that case, and [in the spirit of deduplication](https://github.com/caskroom/homebrew-cask/issues/15603), submit it first to [Homebrew](https://github.com/Homebrew/homebrew). If it is rejected there, you may then try again in Homebrew-Cask (link us to the issue on Homebrew so we can see their reasoning for rejection).

beaumartinez:
Hey @vitorgalvao, thanks for the quick response!

This formula is binary only, which is why I submitted it here. [Homebrew specify they don't accept binary-only formulae](https://github.com/Homebrew/brew/blob/de880f1e870d91dd838228de2987f9bd6d194aef/docs/Acceptable-Formulae.md#we-dont-like-binary-formulae), and instead redirects contributors to Homebrew Cask—that's why I created the formula here in the first place.

I anticipate a rejection if I open a PR there since this formula is indeed binary-only, but I can try if you want me to.

vitorgalvao:
> This formula is binary only, which is why I submitted it here. [Homebrew specify they don't accept binary-only formulae](https://github.com/Homebrew/brew/blob/de880f1e870d91dd838228de2987f9bd6d194aef/docs/Acceptable-Formulae.md#we-dont-like-binary-formulae), and instead redirects contributors to Homebrew Cask—that's why I created the formula here in the first place.

No, what it says is (emphasis mine):

> Our policy is that formulae in the core repository ([homebrew/core](https://github.com/Homebrew/homebrew-core)) **must be open-source and** either **built from source** or produce cross-platform binaries like e.g. Java). Binary-only formulae should go to
[Homebrew Cask](https://github.com/caskroom/homebrew-cask).

Which is precisely the case, here. Only closed-source binaries are to be submitted as casks.

MikeMcQuaid:
@vitorgalvao This is a different case that I think should be allowed: open-source binaries that require code signing to function. This isn't something we can do inside Homebrew and I do think Cask makes more sense.

MikeMcQuaid:
@vitorgalvao I'm confirming if that's the case here but even if not I think it's definitely a case that warrants addition.

MikeMcQuaid:
@vitorgalvao I checked and it doesn't apply in this case but may for others in future (but the submitter will probably be aware of that).

Casks/locateme.rb
```diff
@@ -0,0 +1,10 @@
+cask 'locateme' do
+  version '0.2.1'
+  sha256 '137016e6c1a847bbe756d8ed294b40f1d26c1cb08869940e30282e933e5aeecd'
+
+  url "http://downloads.sourceforge.net/project/iharder/locateme/LocateMe-v#{version}.zip"
```
vitorgalvao: Use `"https://downloads.sourceforge.net/iharder/LocateMe-v#{version}.zip"` and it should work.

jawshooah:
@vitorgalvao This was already accepted into homebrew-core.

vitorgalvao:
> @vitorgalvao This was already accepted into homebrew-core.

Ah, from @MikeMcQuaid’s comment I was under the impression he preferred this particular app in HBC.

MikeMcQuaid:
@vitorgalvao I got this confused with another command-line app I saw recently that needed signing to work.

PR #25567, opened 2016-10-17
Revert "Update libreoffice to 5.2.2" to "Update libreoffice to 5.2.1"

RJHsiao:
Hi all,

It looks like LibreOffice 5.2.2 is disappeared:
http://download.documentfoundation.org/libreoffice/stable/
even we still can found 5.2.2.x [here](http://downloadarchive.documentfoundation.org/libreoffice/old/).

Maybe we should revert this PR: caskroom/homebrew-cask#25011


miccal:
[Version 5.2.2.](http://www.libreoffice.org/download/libreoffice-fresh/?version=5.2.2) is back up.

RJHsiao:
OK, just forget it. :no_mouth:

PR #25625, opened 2016-10-18
Remove `pdftoolkit.rb`.

reitermarkus:
This was erroneously added in https://github.com/caskroom/homebrew-cask/pull/25600.


jawshooah:
Whoops!

PR #25925, opened 2016-10-19
Add new: Baidu Cloud Manager 2.0.0

zshbleaker:

Because the app token uses the Chinese word "Guanjia", which means "Manager" in English.

Baidu Cloud is a leading Dropbox-like service in China.


miccal:
The `sha256` needs to be set - you can find what it is by running `shasum -a 256` on the downloaded file.

zshbleaker:
Added.

fanquake:
Is this different from https://github.com/caskroom/homebrew-cask/blob/master/Casks/baiducloud.rb ?

baiducloud-manager.rb
```diff
@@ -0,0 +1,10 @@
+cask 'baiducloud-manager' do
+  version '2.0.0'
+  sha256 '06a3477139a2c96fc8ceaa3a59c1a65dc0b57ec9a069f2f6356067778ae2014f'
+
+  url 'http://issuecdn.baidupcs.com/issue/netdisk/MACguanjia/BaiduNetdisk_mac_2.0.0.dmg'
```
fanquake: Use #{version} in the url stanza

baiducloud-manager.rb
```diff
@@ -0,0 +1,10 @@
+cask 'baiducloud-manager' do
+  version '2.0.0'
+  sha256 '06a3477139a2c96fc8ceaa3a59c1a65dc0b57ec9a069f2f6356067778ae2014f'
+
+  url 'http://issuecdn.baidupcs.com/issue/netdisk/MACguanjia/BaiduNetdisk_mac_#{version}.dmg'
```
jawshooah: Interpolation only works with double quotes.

baiducloud-manager.rb
```diff
@@ -0,0 +1,10 @@
+cask 'baiducloud-manager' do
```
jawshooah: Per the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md), this needs to be called `baidunetdisk`, and the file must be renamed accordingly.

baiducloud-manager.rb
```diff
@@ -0,0 +1,10 @@
...
+  version '2.0.0'
+  sha256 '06a3477139a2c96fc8ceaa3a59c1a65dc0b57ec9a069f2f6356067778ae2014f'
+
+  url 'http://issuecdn.baidupcs.com/issue/netdisk/MACguanjia/BaiduNetdisk_mac_#{version}.dmg'
+  name '百度网盘'
```
jawshooah: Please add the English name as well, as a separate `name` stanza below this one.

baiducloud-manager.rb
jawshooah: This needs a [comment](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment).

jawshooah (changes requested):

baiducloud-manager.rb
jawshooah: Additionally, you've added this file at the root of the repository. It belongs in the `Casks` directory.

jawshooah (changes requested):
You still need to move `baidunetdisk.rb` to `Casks/baidunetdisk.rb`.

jawshooah (changes requested):
The directory is `Casks`, not `Caskroom`.

jawshooah (approved):

PR #25936, opened 2016-10-19
Remove {find,fix,merge}_outdated_appcasts scripts

jawshooah:
Refs https://github.com/caskroom/homebrew-cask/issues/25926


reitermarkus:
We could also remove the `develop_brew_cask`and `production_brew_cask` scripts.

PR #26012, opened 2016-10-21
Chicken 2.2b3 - changed to my version on github 

boecko:
I made a DMG-File  with upgraded build settings (>OSX 10.7, arch) so the app has native fullscreen support.

see https://github.com/boecko/chicken/releases/tag/v2.2b3

Checks

### DIFF
--- a/Casks/chicken.rb
+++ b/Casks/chicken.rb
@@ -1,12 +1,12 @@
 cask 'chicken' do
-  version '2.2b2'
-  sha256 '20e910b6cbf95c3e5dcf6fe8e120d5a0911f19099128981fb95119cee8d5fc6b'
+  version '2.2b3'
+  sha256 '78f05c20e6584d8eb46a30c0f9eb3e90f4e583a0f2d78ad5810370fadec0f323'
 
-  url "https://downloads.sourceforge.net/chicken/Chicken-#{version}.dmg"
-  appcast 'http://chicken.sourceforge.net/chicken.xml',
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Your version as `0` stars. It’s safe to assume users don’t know about it. Don’t just replace the default cask with your own, please. It’s unexpected to users. Submit your patches upstream and get them merged.

boecko:
My repo is just 2 hours old. Even me didn't know about the repo yesterday ;)

The version from sourceforge is from 2011-11-19 (https://sourceforge.net/projects/chicken/files/) and is still targeted to 10.4
Upstream means to get involved in sourceforge and mess around with svn. No thanks

vitorgalvao:
> My repo is just 2 hours old. Even me didn't know about the repo yesterday ;)

All the more reason. No one installing this cask will be expecting to get your version.

jawshooah:
We have protocol for this, just create a new cask called `boecko-chicken`

vitorgalvao:
> We have protocol for this, just create a new cask called `boecko-chicken`

Reason I didn’t mention it is we don’t want to get the repo (and the main repo, even) full of personal forks for every little change. For all we know, this new version fills @boecko needs and will never be updated again and will never have users.

Though we don’t rigidly follow HB’s policy of outright refusing tools with few stars and self-submitted apps, there is a worthy logic behind it.

In this case, I don’t think it is worthy it, at this point in time, to add this app. When and if it becomes a bit more popular, it might be.

jawshooah:
Fair enough. @boecko, you are free to add this Cask to your own custom tap.

boecko:
ok ..

PR #26024, opened 2016-10-22
.travis.yml: Update osx_image versions

adidalal:
Ref:

In order to give users who explicitly use some of our less common Xcode versions some time to transition, we'll be maintaining all the current images through Monday, October 31st. On Oct. 31 we will be simplifying our image catalog to the following three available options:

osx_image: xcode6.4 (with 10.10)
osx_image: xcode7.3 (with 10.11 and default)
osx_image: xcode8 (currently 10.11, but tentatively planned to be 10.12 once we've done some vSphere compatibility testing)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,8 @@ env:
 
 matrix:
   include:
-    - env: OSX=10.11 HOMEBREW_RUBY=2.0.0
-      os: osx
-      osx_image: xcode8
-      rvm: system
-    - env: OSX=10.11 HOMEBREW_RUBY=2.0.0
-      os: osx
-      osx_image: xcode7.3
-      rvm: system
+    - osx_image: xcode8
+    - osx_image: xcode7.3
   fast_finish: true
 
 branches:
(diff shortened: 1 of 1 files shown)


adidalal:
Also, do we need to change the travis scripts in any other way? I see a few errors in the travis log, but nothing that stops our builds.

.travis.yml
```diff
@@ -12,11 +12,11 @@ matrix:
   include:
     - env: OSX=10.11 HOMEBREW_RUBY=2.0.0
```
reitermarkus: I can't find any reference to `OSX` or `HOMEBREW_RUBY`  in the core code, so we might as well remove them.

jawshooah: The `OSX` variable there was intended to make it easy to identify at a glance which version each job was running on. Not essential, just convenient.

adidalal: They're both 10.11 at the moment. I can add them back in, but xcode8 will eventually be sierra.

reitermarkus:
Well, I missed something. `HOMEBREW_RUBY` is used in `before_script.sh`.

adidalal:
https://github.com/caskroom/homebrew-cask/pull/26030 seemed just fine, and that was after this merged.

PR #26027, opened 2016-10-22
[WIP] Travis simplifications

adidalal:
Since this repo doesn't have core code anymore, we can remove all the ruby/gem references

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+cache: bundler
 
 sudo: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,10 @@ matrix:
   include:
     - os: osx
       osx_image: xcode8
+      rvm: system
     - os: osx
       osx_image: xcode7.3
+      rvm: system
   fast_finish: true
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,29 +2,17 @@ language: ruby
 
 sudo: false
 
-env:
-  global:
-    - LANG=en_US.UTF-8
-    - LANGUAGE=en_US.UTF-8
-    - LC_ALL=en_US.UTF-8
-
 matrix:
   include:
     - os: osx
       osx_image: xcode8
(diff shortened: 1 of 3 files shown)


ci/travis/before_script.sh
```diff
@@ -10,46 +10,13 @@
...
-# unset rvm hook functions
-run unset -f cd gem
-
-# print all travis-defined environment variables
-run 'env | sort'
-
-# print detailed macOS version info
-run sw_vers
-
-# capture system ruby and gem locations
```
adidalal: Travis is still failing, and it's definitely due to this removal, any thoughts on a less convoluted way to do things?
@reitermarkus @jawshooah

.travis.yml
```diff
@@ -22,10 +17,6 @@ branches:
   only:
     - master
 
-cache:
-  directories:
-    - $HOME/.gem
```
reitermarkus: I *think* this is still used because `brew cask style` will install `rubocop-cask` in there.

reitermarkus (changes requested):

.travis.yml
reitermarkus: Unless `cache: bundler` is the equivalent to this.

adidalal: looks like `vendor/bundle` https://docs.travis-ci.com/user/caching/#Enabling-Bundler-caching

reitermarkus: This won't work then, since we're not even using `bundler`.

ci/travis/before_script.sh
```diff
@@ -10,46 +10,30 @@
 
 header 'Running before_script.sh...'
 
-# https://github.com/rvm/rvm/pull/3627
+# work around travis xcode bug
+# https://github.com/direnv/direnv/issues/210
```
reitermarkus: https://github.com/travis-ci/travis-ci/issues/6307#issuecomment-233315824 might be a better link.

adidalal:
Splitting this up into smaller PRs.

PR #26035, opened 2016-10-22
Travis: Simplify before_script

adidalal:
Removes some verbosity and potentially unneeded commands

### DIFF
--- a/ci/travis/before_script.sh
+++ b/ci/travis/before_script.sh
@@ -10,17 +10,9 @@
 
 header 'Running before_script.sh...'
 
-# https://github.com/rvm/rvm/pull/3627
+# work around travis osx bug
+# https://github.com/travis-ci/travis-ci/issues/6307#issuecomment-233315824
 run rvm get head
-
-# unset rvm hook functions
-run unset -f cd gem
-
-# print all travis-defined environment variables
-run 'env | sort'
-
-# print detailed macOS version info
-run sw_vers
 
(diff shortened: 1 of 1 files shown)


ci/travis/before_script.sh
```diff
@@ -10,18 +10,10 @@
 
 header 'Running before_script.sh...'
 
-# https://github.com/rvm/rvm/pull/3627
+# work around travis osx bug
+# https://github.com/travis-ci/travis-ci/issues/6307#issuecomment-233315824
 run rvm get head
 
-# unset rvm hook functions
```
adidalal: Don't see where this was ever an issue

ci/travis/before_script.sh
```diff
@@ -39,17 +31,9 @@ run export PATH="${GEM_BINDIR}:${SYSTEM_GEM_BINDIR}:${SYSTEM_RUBY_BINDIR}:${PATH
...
-run which ruby
-run ruby --version
-
-run which gem
-run gem --version
-
+# update homebrew
 run brew update
 
-# ensure that we are the only brew-cask available
```
adidalal: Should no longer be needed

ci/travis/before_script.sh
jawshooah: I didn't create an issue about it, but rvm wraps some commands in shell functions that execute some kind of hooks that were causing problems. If everything looks fine now I'm fine with removing this

PR #26066, opened 2016-10-23
Update freesmug-chromium to 54.0.2840.71

ngquerol:

### DIFF
--- a/Casks/freesmug-chromium.rb
+++ b/Casks/freesmug-chromium.rb
@@ -1,6 +1,6 @@
 cask 'freesmug-chromium' do
-  version '53.0.2785.143'
-  sha256 '654e36265467cd9c412c89118ebdfbd3f942b6409d8cec8a03ab0c2e20b67cd0'
+  version '54.0.2840.71'
+  sha256 'f64cf586d017d737288e29556ccbf6c69d8002fcfdf96e0d48a3a981111e4f59'
 
   # sourceforge.net/osxportableapps was verified as official when first introduced to the cask
   url "https://downloads.sourceforge.net/osxportableapps/Chromium_OSX_#{version}.dmg"


miccal:
You also need to update the [checkpoint](https://github.com/caskroom/homebrew-cask/blob/c023b005d66d4cdcb3a714f7f803833bdd86eff1/doc/cask_language_reference/stanzas/appcast.md).

vitorgalvao:
Weird that it failed, since it seems to have been created with `cask-repair`, meaning it was automatic. Must be one of those uncommon cases where depending on where you are the feed appears to have some differences.

vitorgalvao:
Did a test with cask repair and it seems to have worked as expected: https://github.com/caskroom/homebrew-cask/pull/26069/

Thank you for the contribution, @ngquerol. This time I’m going to merge the other simply due to convenience.

ngquerol:
Thanks, didn't catch that one (trusted the script a bit too much 😀)

vitorgalvao:
You’re supposed to trust the script. People do it every day. I did the exact same thing as you.

The script itself checks for errors independently. If you had run its commands manually, it’s almost certain you would have gotten the same results.

ngquerol:
Didn't want to imply the script is bad at all! Thank you very much for fixing this pull request @vitorgalvao 😄

vitorgalvao:
> Didn't want to imply the script is bad at all!

Oh, I know, no offence taken. My point was that you shouldn’t have to worry about if your submission is correct when using the script: that’s the point of it.

In this particular case, it was something external that caused the issue, and even if you had double-checked, you would likely have gotten the same result.

ngquerol:
Ah yes, didn't want to imply the script is bad at all! Thank you very much for fixing this anyway. :smile:

PR #26097, opened 2016-10-25
remove Cookie

pegurnee:
Cookie is a paid app that is available on the MAS. What the cask here provides is access to the two week trial. Should be removed.


vitorgalvao:
They also have a [non-MAS-store](https://sweetpproductions.com/store/) where you can buy a license.

PR #26208, opened 2016-10-30
Update YY to 1.1.7

zshbleaker:
Update YY to 1.1.7


miccal:
The ` sha256` you have used is not correct - I find it to be `4f04f1d74f66562e4b119c919000e303ac620a57221aefe9013ff404c1f55531`.

vitorgalvao:
Travis seems to agree with @miccal. I also need to test a report with `cask-repair` (check our `CONTRIBUTING.md` document for details), so I’ll update this one. Thank you for the report.

PR #26294, opened 2016-11-01
Update svgcleaner to 0.7.0

reitermarkus:


reitermarkus:
@caskroom/maintainers, I haven't really been following the Formula/Cask duplicate discussion. Should we rename this according to the token reference to `svg-cleaner` or to `svgcleaner-gui`, since it's basically only a GUI for the `svgcleaner` Formula?

vitorgalvao:
Never `-gui`, but `-app`.

This one is a bit like `transmission`. They both seem to be equally important, so it stays without the suffix.

PR #26308, opened 2016-11-02
Update rstudio to 1.0.44

tomyun:

- [X] `brew cask audit --download rstudio` is error-free.
- [X] `brew cask style --fix rstudio` reports no offenses.

### DIFF
--- a/Casks/rstudio.rb
+++ b/Casks/rstudio.rb
@@ -1,6 +1,6 @@
 cask 'rstudio' do
-  version '0.99.903'
-  sha256 'cb7ac2b9ba2e0d4f4096b2dd2859c72742d12faf1461c68a6bfd3bce532ee1d6'
+  version '1.0.44'
+  sha256 'ec571ee4d1415cc031f8f55873ca15e75949038fe680520799934e115138066a'
 
   # rstudio.org was verified as official when first introduced to the cask
   url "https://download1.rstudio.org/RStudio-#{version}.dmg"


miccal:
Could you please add the `depends_on cask: 'r-app'` [stanza](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/depends_on.md)?

tomyun:
I'm not familiar with how cask dependencies are supposed to be like, but how would we deal with brewed `R` package in this case? For instance, I have `R` package installed by brew and `rstudio` works well with no dependency required for `r-app`.

miccal:
Not sure about that - I was under the impression `r-app` Cask was the only version of R in Homebrew. Pinging @vitorgalvao.

tomyun:
`brew install homebrew/science/r` gives you a brewed version of R.

miccal:
You are correct @tomyun - the website says that [R is required](https://www.rstudio.com/products/rstudio/download/), so I think it needs a `depends_on` stanza, but I am not sure which one to use.

vitorgalvao:
If any `r` is going to be required, it will likely be the `homebrew/science` one. I don’t remember ever seeing a dependency on an app bundle.

But are we sure this does even need the CLI installed?

miccal:
Their [download page](https://www.rstudio.com/products/rstudio/download/) states that "RStudio requires R 2.11.1 (or higher)" and links to the [download website](https://cran.rstudio.com) for R.

vitorgalvao:
@miccal Then it should have a `depends_on formula: 'homebrew/science/r'`.

PR #26319, opened 2016-11-02
Update etcher to 1.0.0-beta.16

qiwenmin:

### DIFF
--- a/Casks/etcher.rb
+++ b/Casks/etcher.rb
@@ -1,9 +1,9 @@
 cask 'etcher' do
-  version '1.0.0-beta.15'
-  sha256 '085775118bee1f34e99e54493ae923ef69887467eadee7ccf05caebac9ac737f'
+  version '1.0.0-beta.16'
+  sha256 '4c6e3ec337e871acbd7913d345e169aabff6b2865fa4cd386e66a0d4e38f060a'
 
   # resin-production-downloads.s3.amazonaws.com was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


jviotti:
Hi @qiwenmin ,

Etcher maintainer here! Thanks so much for adding Etcher to `brew cask`, I was just looking into it and found you've added it already.

Is there something we can do from our side to make updates more automatic? (so you don't have to create a PR changing `version` and `sha256` every time).

Also, Etcher has a built-in auto-update dialog that pops up when there is an update available, redirecting users to the website. How does this play with Cask? Do you think we should disable those auto-updates check if the app was installed as Cask?

qiwenmin:
Hi @jviotti ,

I noticed google-chrome and evernote can upgrade themselves smoothly. And their casks' version is "latest". If Etcher can upgrade itself like them - without manually downloading, mounting .dmg image, dragging and dropping, maybe its cask can use the same way.

Many Apps do the auto-updates checking with the brew-cask upgrading. So auto-updates checking is OK, I think.

I don't know if there is a better way. I want to know also.

jviotti:
Hi @qiwenmin ,

Thank you for your response! We plan to have auto-updates in Etcher at some point (other priorities pushed that aside for now) (see https://github.com/resin-io/etcher/issues/336), but definitely sounds like the way to go.

I guess the best way forward is to keep updating the Cask ourselves for now, and move to `latest` once we have auto-updates up and running.

Thanks a lot!

qiwenmin:
@jviotti You are welcome!

PR #26375, opened 2016-11-03
Move Veertu Desktop to non-versioned download URL

dwaite:

### DIFF
--- a/Casks/veertu-desktop.rb
+++ b/Casks/veertu-desktop.rb
@@ -1,9 +1,9 @@
 cask 'veertu-desktop' do
-  version '1.2.87'
-  sha256 'e5edb743e6ba9f84b98475e140b978679a6fd6b9c2cdb63c29e736780670a6c7'
+  version :latest
+  sha256 :no_check
 
   # d2sje6b9huarvp.cloudfront.net was verified as official when first introduced to the cask
-  url "https://d2sje6b9huarvp.cloudfront.net/VeertuDesktop-#{version}.dmg"
+  url 'https://d2sje6b9huarvp.cloudfront.net/VeertuDesktop.dmg'
   name 'Veertu'
   homepage 'https://veertu.com'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you, but we prefer versioned downloads. Our users overwhelmingly chose it that way.

miccal:
The current download link `ttps://d2sje6b9huarvp.cloudfront.net/VeertuDesktop-#{version}.dmg` no longer works though.

vitorgalvao:
> The current download link `ttps://d2sje6b9huarvp.cloudfront.net/VeertuDesktop-#{version}.dmg` no longer works though.

Ah, that’s a different matter, then.

vitorgalvao (approved):

jawshooah:
Alternatively, we could keep the version number and add an appcast: https://d2sje6b9huarvp.cloudfront.net/vdoaupd.rss

vitorgalvao:
> Alternatively, we could keep the version number and add an appcast: https://d2sje6b9huarvp.cloudfront.net/vdoaupd.rss

Yes, best solution.

@dwaite Mind adding the [`appcast`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md)?

dwaite:
will do.

Casks/veertu-desktop.rb
```diff
@@ -1,9 +1,11 @@
 cask 'veertu-desktop' do
-  version '1.2.87'
-  sha256 'e5edb743e6ba9f84b98475e140b978679a6fd6b9c2cdb63c29e736780670a6c7'
+  version '1.2.83'
```
vitorgalvao: Wait, why did we drop a version? According to the appcast we can even go higher than before, to `1.2.90`.

vitorgalvao (changes requested):

adidalal:
Versioned download seems to work, actually.
https://github.com/caskroom/homebrew-cask/pull/26443

PR #26395, opened 2016-11-04
Fix version interpolation in PolarClock 3.0

thetamind:
Correct version interpolation in screen_saver stanza to fix:

`Error: It seems the Screen Saver source is not there:
'/usr/local/Caskroom/polar-clock/3.0/PolarClock #{version}.saver'`.


fanquake:
Thanks 🍻

PR #26545, opened 2016-11-07
New formula: Lacaille 2.2.2

umireon:

---

[Lacaille] is a keyboard remapper for the Japanese thumb-shift method (親指シフト).

[Lacaille]: http://lacaille.jpn.org

### DIFF
--- a/Casks/lacaille.rb
+++ b/Casks/lacaille.rb
@@ -12,7 +12,5 @@ cask 'lacaille' do
 
   uninstall login_item: 'Lacaille'
 
-  zap delete: [
-                '~/Library/Preferences/org.jpn.lacaille.Lacaille.plist',
-              ]
+  zap delete: '~/Library/Preferences/org.jpn.lacaille.Lacaille.plist'
 end
--- a/Casks/lacaille.rb
+++ b/Casks/lacaille.rb
@@ -1,0 +1,18 @@
+cask 'lacaille' do
+  version '2.2.2'
+  sha256 '3885a55a4fe1b80f9340271972372ab4b4fe234aad37c27f849e6d78fbd6fcb2'
+
+  url "http://lacaille.jpn.org/Lacaille_#{version}.dmg"
+  appcast 'http://lacaille.jpn.org',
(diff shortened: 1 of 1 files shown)


Casks/lacaille.rb
```diff
@@ -0,0 +1,16 @@
+cask 'lacaille' do
+  version '2.2.2'
+  sha256 '3885a55a4fe1b80f9340271972372ab4b4fe234aad37c27f849e6d78fbd6fcb2'
+
+  url "http://lacaille.jpn.org/Lacaille_#{version}.dmg"
+  appcast 'http://lacaille.jpn.org',
```
vitorgalvao: Why do you think this is an effective appcast? I does not appear like this page will change with every version.

Casks/lacaille.rb
```diff
@@ -0,0 +1,16 @@
...
+          checkpoint: '9278d72ed9b5179467c1cf8888fc8d981ba32fc085bad32cf09119cf836c3ec4'
+  name 'Lacaille'
+  homepage 'http://lacaille.jpn.org'
+
+  app 'Lacaille.app'
+
+  uninstall login_item: 'Lacaille'
+
+  zap delete: '~/Library/Preferences/org.jpn.lacaille.Lacaille.plist'
```
vitorgalvao: Made this into a string (no point in an array for a single element).

Casks/lacaille.rb
umireon: I'm sorry for that being unclear.
I suggested http://lacaille.jpn.org is an effective appcast for the following reason.

http://lacaille.jpn.org is the only page the site has.
In the second section of the page which means Downloads and New Features (ダウンロード・新機能), the list of Lacaille releases is presented.
Therefore, this page must change with every release.

umireon:
@vitorgalvao Thank you for reviewing!
I think the appcast is valid, but unfortunately the releases is hidden until clicking the header.

vitorgalvao:
No matter that they’re hidden, since they’re in the HTML. Took another look and it seems like it might work as an appcast, yes. Thank you.

PR #26569, opened 2016-11-08
Remove outdated SSHFS.

reitermarkus:
Prefer https://github.com/Homebrew/homebrew-fuse/blob/master/Formula/sshfs.rb, which is currently at version 2.8.


adidalal:
Needs an entry in migrations json also

reitermarkus:
> Needs an entry in migrations json also

Ah yes, forgot about that.

PR #26636, opened 2016-11-10
Remove dead Casks

miccal:
Using the very handy [`brew cask _stanza`](https://github.com/caskroom/homebrew-cask/issues/17007) command, a simple bash script, the `brew cask fetch` command and then investigating each problem Cask individually, I have determined the following 17 Casks are dead.


fanquake:
Ping @caskroom/maintainers

PR #26695, opened 2016-11-12
Update astah-community to 7.1.0

travismckay:


miccal:
I assume that number `f2c212` is like a build number, so it may be better to change `version` to `7.1.0,f2c212`, then the `url` may be written `url "http://cdn.astah.net/downloads/astah-community-#{version.before_comma.dots_to_underscores}-#{version.after_comma}-MacOs.dmg"` and the `package` to `pkg "astah community ver #{version.before_comma.dots_to_underscores}.pkg"` to make future updates easier.

travismckay:
Yes, this does appear to be the build number.  I've updated the cask per @miccal's suggestion.

PR #26818, opened 2016-11-16
Revert "MetaGrowler 0.5.2 extensible monitor using Growl for OSX"

jawshooah:
Reverts caskroom/homebrew-cask#26791


MikeMcQuaid:
Nice work 👍

pirate:
Why was this reverted?  MetaGrowler is still [available via mirrors](http://en.freedownloadmanager.org/Mac-OS/MetaGrowler-FREE.html), and is functional Mac versions below macOS Sierra.

miccal:
This was reverted because it was merged incorrectly - nothing to do with the Cask itself.

pirate:
Thanks @miccal, I was mistakenly searching for `meta-growler` instead of `metagrowler` and didn't see that Cask.

PR #26819, opened 2016-11-16
travis: fail if any cask files added outside Casks dir

jawshooah:
Refs https://github.com/caskroom/homebrew-cask/pull/26791#issuecomment-260959390

### DIFF
--- a/ci/travis/helpers.sh
+++ b/ci/travis/helpers.sh
@@ -14,6 +14,7 @@ shopt -s extglob
 
 CYAN='\033[0;36m'
 MAGENTA='\033[1;35m'
+RED='\033[1;31m'
 NC='\033[0m' # no color
 
 # log command before running and add a blank line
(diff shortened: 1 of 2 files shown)


BenjaminHCCarr:
Thank you!

As mentioned in #267915#issuecomment-260818620

I broke my own patch tree, and it looks like the patches were coming off `/` instead of `/Casks`.

Sincerely sorry for breaking Cask for everyone.

jawshooah:
Accidents happen 😄

miccal:
Hello @jawshooah, it seems the the builds are failing for all new casks. I _think_ your lines 24 and 26 in your modified file `ci/travis/script.sh` shouldn't have the arithmetic tests `arg -gt 0` but instead should have `-n arg`.

BenjaminHCCarr:
I'm seeing this in #26845 as well if you want an example......
cc: @jawshooah @reitermarkus @vitorgalvao @MikeMcQuaid

jawshooah:
@miccal The arithmetic test isn't the problem; it's just checking if the array `casks_outside_casks_dir` is non-empty. Testing the length with `-gt`, `-lt`, etc is less error-prone than string tests like `-n`, `-z`, etc because they only operate on the first element of the array. See the answers [here](http://serverfault.com/questions/477503/check-if-array-is-empty-in-bash).

The problem is that `git diff *.rb` is pulling in files outside the current working directory. Using `./*.rb` instead doesn't seem to change anything. Will investigate further later, but in the meantime this can be reverted.

PR #26848, opened 2016-11-17
Add Upwork Desktop Latest Version

bashu:

replacement of oDesk Team, ref #26539

### DIFF
--- a/Casks/upwork.rb
+++ b/Casks/upwork.rb
@@ -1,0 +1,10 @@
+cask 'upwork' do
+  version '4_2_88_0_zjqu5h1eh5fvxz5t'
+  sha256 '4c321f521f2f42e0cd001a98274b99ea9b269ad8fc5334403c4065667ce56677'
+
+  url "https://updates-desktopapp.upwork.com/binaries/v#{version}/Upwork.dmg"
+  name 'Upwork'
+  homepage 'https://www.upwork.com/'
+
+  app 'Upwork.app'
+end


miccal:
I cannot see a link to this app from the homepage, I can only see a link to the [mobile app](https://www.upwork.com/i/mobile/)? Their [download page](https://www.upwork.com/downloads) appears to require a login?

bashu:
@miccal suddenly yes! but you can download / install desktop app using this formula and no login required

miccal:
Yes, but this may be classified as a [walled build](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#unofficial-vendorless-and-walled-builds) which are not accepted for security reasons.

bashu:
@miccal should I PR to https://github.com/alehouse/homebrew-unofficial instead?

miccal:
Ping @vitorgalvao.

bashu:
@miccal @vitorgalvao Upwork Desktop is a successor of oDesk Team which was a part of homebrew-cask in a past, ref #26539

Casks/upwork.rb
```diff
@@ -0,0 +1,23 @@
...
+  version :latest
+  sha256 :no_check
+
+  url 'https://updates-desktopapp.upwork.com/binaries/v4_2_88_0_zjqu5h1eh5fvxz5t/Upwork.dmg'
+  name 'Upwork'
+  homepage 'https://www.upwork.com/'
+
+  app 'Upwork.app'
+end
+~
```
vitorgalvao: This definitely is not right.

Casks/upwork.rb
```diff
@@ -0,0 +1,23 @@
+cask 'upwork' do
+  version :latest
```
vitorgalvao: `version '4_2_88_0_zjqu5h1eh5fvxz5t'`

Casks/upwork.rb
```diff
@@ -0,0 +1,23 @@
...
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be set. You can find what it is by running `shasum -a 256` on the downloaded file.

Casks/upwork.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  url 'https://updates-desktopapp.upwork.com/binaries/v4_2_88_0_zjqu5h1eh5fvxz5t/Upwork.dmg'
```
vitorgalvao: `url "https://updates-desktopapp.upwork.com/binaries/v#{version}/Upwork.dmg"`

vitorgalvao (changes requested):

vitorgalvao:
Clarification regarding walled builds — see [the glossary](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) (emphasis added):

> **Walled:** When the download URL is both behind a login/registration form **and from a host that differs from the homepage**.

That’s not the case, here. Issue is if you submit a casks with homepage `http://one.com` and the URL points to `http://two.com`. If the download URL is behind a registration/login wall, a user will have no easy way to know the download URL is trustworthy. If the download URL is from the same domain as the homepage, though, then you can be pretty sure the download is official (unless the website was hacked, but that would be an issue in every case).

miccal:
Thanks for the clarification @vitorgalvao.

Casks/upwork.rb
bashu: @vitorgalvao fixed

Casks/upwork.rb
bashu: @vitorgalvao what an ugly copy-paste! fixed

bashu:
@miccal @vitorgalvao thanks for feedback!

Casks/upwork.rb
```diff
@@ -0,0 +1,10 @@
+cask 'upwork' do
+  version '4_2_88_0_zjqu5h1eh5fvxz5t'
+  sha256 '4c321f521f2f42e0cd001a98274b99ea9b269ad8fc5334403c4065667ce56677'
+
+  url 'https://updates-desktopapp.upwork.com/binaries/v#{version}/Upwork.dmg'
```
miccal: You need double quotes around the URL for the interpolation to work, like so:
```
url "https://updates-desktopapp.upwork.com/binaries/v#{version}/Upwork.dmg"
```

bashu:
@miccal thank you

PR #26851, opened 2016-11-17
fix(dotnet): using the right package version (version after comma)

diogodamiani:
Fixing #26826 to use the right package version. It was my mistake.

The correct is: /usr/local/share/dotnet/shared/Microsoft.NETCore.App/1.1.0


diogodamiani:
Already fixed by #26850. Thanks @forevergenin!

forevergenin:
You are welcome @diogodamiani

PR #26856, opened 2016-11-17
Set auto_updates to true on some Casks

MikeMcQuaid:
These are all the ones I have installed locally that are missing these flags (and I verified them all).

### DIFF
--- a/Casks/appcleaner.rb
+++ b/Casks/appcleaner.rb
@@ -19,6 +19,7 @@ cask 'appcleaner' do
   name 'AppCleaner'
   homepage 'https://www.freemacsoft.net/appcleaner/'
 
+  auto_updates true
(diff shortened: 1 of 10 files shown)


vitorgalvao:
Thank you. On this list, only three are missing `appcast`: `doxie`, `java`, `zoomus`.

None of them seems to have a Sparkle appcast, though, so we’ll have to find out where their appcasts reside using other methods later.

vitorgalvao (approved):
Only waiting for Travis

MikeMcQuaid:
> None of them seems to have a Sparkle appcast, though, so we’ll have to find out where their appcasts reside using other methods later.

Yeh, none of them use Sparkle (as far as I can tell) but their own custom updaters.

PR #26868, opened 2016-11-17
Add fastlane bundle 1.0

milch:

### DIFF
--- a/Casks/fastlane.rb
+++ b/Casks/fastlane.rb
@@ -1,0 +1,13 @@
+cask 'fastlane' do
+  version '1.0'
+  sha256 'f642cdae3d4841f85d4a9260866361d136fad82974cf5905629d48f70d86f294'
+
+  # kits-crashlytics-com.s3.amazonaws.com/fastlane/ was verified as official when first introduced to the cask
+  url "https://kits-crashlytics-com.s3.amazonaws.com/fastlane/standalone/bundle-#{version}.zip"
+  name 'fastlane'
+  homepage 'https://fastlane.tools'
+
+  installer script: "#{staged_path}/install", args: ['-p', '-u', '-b'], sudo: false
+
+  uninstall script: { executable: "#{staged_path}/uninstall", args: ['-y'], sudo: false }
+end


Casks/fastlane.rb
```diff
@@ -0,0 +1,15 @@
+cask 'fastlane' do
+  VERSION_NUMBER = '1.0'.freeze
+
+  version VERSION_NUMBER
```
Numbermaniac: Why not just `version '1.0'`, and then use `#(version}` in the URL? Then `VERSION_NUMBER` isn't needed.

miccal:
This seems to be a CLI-only open-source tool, so it should first be submitted to [Homebrew](https://github.com/Homebrew/homebrew). Please see the [rejected casks document](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/rejected_casks.md) for more information.

milch:
@miccal Thanks for your review. However, I don't think Homebrew would accept this. Basically everything that they state in their [Acceptable Formulae](https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md) docs is not fulfilled by this fastlane bundle, specifically:

- **We try hard to avoid dupes in Homebrew/homebrew**: `fastlane` is a ruby gem
- **We don’t like tools that upgrade themselves**: `fastlane` updates itself
- **We don’t like binary formulae**: The way this bundle is structured means it contains a binary release of ruby, gems, and other tools. The fastlane gem itself that is provided here is also a prebuilt version. 
- **Bindings**: Same here, `fastlane` is a ruby gem.

Casks/fastlane.rb
```diff
@@ -0,0 +1,13 @@
+cask 'fastlane' do
+  version '1.0'
+  sha256 'f642cdae3d4841f85d4a9260866361d136fad82974cf5905629d48f70d86f294'
+
+  # kits-crashlytics-com.s3.amazonaws.com/fastlane/ was verified as official when first introduced to the cask
+  url "https://kits-crashlytics-com.s3.amazonaws.com/fastlane/standalone/bundle-#{version}.zip"
+  name 'fastlane'
+  homepage 'https://fastlane.tools'
+
```
jawshooah: Please use `"#{staged_path}/install"` instead of a relative path.

Casks/fastlane.rb
```diff
@@ -0,0 +1,13 @@
...
+  installer script: './install', args: ['-p', '-u', '-b'], sudo: false
```
jawshooah: Please use `"#{staged_path}/install"` instead of a relative path.

milch:
Thanks for the review, @jawshooah ... Just addressed your feedback!

KrauseFx:
🎉 Thanks everyone 🎉

PR #26937, opened 2016-11-19
Update rubitrack to 4.3.3

berndkrannich:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #26941, opened 2016-11-19
Fix rubitrack 4.3.5

berndkrannich:
Thanks @adidalal for #26938; I missed the fact that rubiTrack was already updated again.

The app name of rubiTrack has '` Pro`' suffix since version 4. Without it, the cask doesn't install properly (which would have been also the case for my PR #26937).

---


berndkrannich:
Thanks @adidalal and @vitorgalvao - that was quick! 👏

PR #26951, opened 2016-11-20
Change https to http in homepage of 64 Casks

miccal:
Update homepage to http for 64 Casks from the list [here](http://caskroom.victorpopkov.com/homebrew-cask/homepages.csv).


vitorgalvao:
Steam redirects to a non-secure page? That is so not cool.

PR #26964, opened 2016-11-20
WIP: Add documentation for new option `pkg choices:`

umireon:
Related to Homebrew/brew#1535
I will work on this after I create an example (osxfuse).

PR #26965, opened 2016-11-20
Fix osxfuse 3.5.3 to use the pkg choices: option

umireon:

Related to Homebrew/brew#1535 and #26964 

This won't work until Homebrew/brew#1535 is merged

### DIFF
--- a/Casks/osxfuse.rb
+++ b/Casks/osxfuse.rb
@@ -9,29 +9,12 @@ cask 'osxfuse' do
   name 'OSXFUSE'
   homepage 'https://osxfuse.github.io/'
 
-  installer script: '/usr/sbin/installer',
-            args:   [
-                      '-pkg', "#{staged_path}/Extras/FUSE for macOS #{version}.pkg",
-                      '-target', '/',
-                      '-applyChoiceChangesXML', "#{staged_path}/Extras/Choices.xml"
-                    ]
-
-  preflight do
-    IO.write "#{staged_path}/Extras/Choices.xml", <<-EOS.undent
-      <plist>
-        <array>
-        	<dict>
-        		<key>attributeSetting</key>
-        		<integer>1</integer>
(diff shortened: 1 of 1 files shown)


umireon:
Currently I found no packages on Homebrew repositories that depends on MacFUSE compatibility layer.
This suggest that the choice changes XML is better to be removed rather than to migrate the new syntax.
Does anybody know if a package depending on the compatibility layer exists?

reitermarkus:
Not sure if `veracrypt` still needs it, since it is now at `1.19`: https://github.com/caskroom/homebrew-cask/commit/dccc678fc3d04844a7a05d4a83e9d9ba1c6c8315#diff-afa78f33c44283fd268ce9e4daf365a2

umireon:
At least in my environment (Sierra), Veracrypt works fine without compat layer.
It seems to be resolved: veracrypt/VeraCrypt#11

fanquake:
#1535 has been merged

reitermarkus:
Ok, since VeraCrypt is the only cask that depended on the compatibility layer, I think we can remove the choices for OSXFUSE.

umireon:
Some Formulas like [fossil](https://github.com/Homebrew/homebrew-core/blob/master/Formula/fossil.rb) and maybe some Formulas on [homebrew/homebrew-fuse](https://github.com/Homebrew/homebrew-fuse) still depend on compatibility layer.

BTW, their dependence on compatibility layer should be removed since users can install OSXFUSE without cask and OSXFUSE does not install the compatibility layer by default.

miccal:
Please note that oxsfuse has now been [updated](https://github.com/caskroom/homebrew-cask/pull/27572) to version 3.5.4.

fanquake:
@umireon Is this fine to merge now? Doesn't need a rebase #27572 .

umireon:
@fanquake Fine, it's working. Thank you.

PR #26978, opened 2016-11-21
Add slash in homepage of 77 Casks

miccal:
Update homepage for 77 Casks from the list [here](http://caskroom.victorpopkov.com/homebrew-cask/homepages.csv).

The only ones remaining from the above list are the ones with the warnings _domain_ and _redirect_ - those are much harder (mainly because I cannot think of a clever way to use `sed`!) and will (possibly) require comments added/edited above the `url` stanza.

Ping @victorpopkov.


victorpopkov:
Thank you, @miccal. The lists are updated.

PR #27054, opened 2016-11-23
Update homepage of 21 Casks

miccal:
Update homepage of 22 Casks.

Ping @victorpopkov

### DIFF
--- a/Casks/namecoin.rb
+++ b/Casks/namecoin.rb
@@ -2,7 +2,7 @@ cask 'namecoin' do
   version '0.3.80-a00c33d'
   sha256 '5e2432e384b46615d34c00fbe2bcb0a9cbc4f1cc1a27cd72fbffceebcac7eb95'
 
-  # namecoin.info/files/ was verified as official when first introduced to the cask
+  # namecoin.info was verified as official when first introduced to the cask
   url "https://namecoin.info/files/Namecoin-Qt.app-#{version}.zip"
   name 'Namecoin'
   homepage 'https://namecoin.org/'
--- a/Casks/namecoin.rb
+++ b/Casks/namecoin.rb
@@ -2,6 +2,7 @@ cask 'namecoin' do
   version '0.3.80-a00c33d'
   sha256 '5e2432e384b46615d34c00fbe2bcb0a9cbc4f1cc1a27cd72fbffceebcac7eb95'
 
+  # namecoin.info/files/ was verified as official when first introduced to the cask
   url "https://namecoin.info/files/Namecoin-Qt.app-#{version}.zip"
   name 'Namecoin'
   homepage 'https://namecoin.org/'
--- a/Casks/namecoin.rb
+++ b/Casks/namecoin.rb
@@ -4,7 +4,7 @@ cask 'namecoin' do
 
   url "https://namecoin.info/files/Namecoin-Qt.app-#{version}.zip"
   name 'Namecoin'
-  homepage 'https://namecoin.info/'
+  homepage 'https://namecoin.org/'
 
   app 'Namecoin-Qt.app'
 
--- a/Casks/nndd.rb
+++ b/Casks/nndd.rb
@@ -2,6 +2,7 @@ cask 'nndd' do
   version '2.4.3,62201'
   sha256 '6a73dcad2e73d877ad1503ed1162cae1a1c84f21d1abaa6aaf9b31bb2fbca531'
 
+  # dl.osdn.jp/nndd/ was verified as official when first introduced to the cask
   url "http://dl.osdn.jp/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
   appcast 'https://osdn.jp/projects/nndd/releases/rss',
           checkpoint: '4f1d23fd9ebe4b3abf7495cf512df2f72f2097b3b4ff51d21e5fa837bb88766b'
--- a/Casks/nndd.rb
+++ b/Casks/nndd.rb
@@ -2,9 +2,8 @@ cask 'nndd' do
   version '2.4.3,62201'
   sha256 '6a73dcad2e73d877ad1503ed1162cae1a1c84f21d1abaa6aaf9b31bb2fbca531'
 
-  # dl.osdn.jp/nndd/ was verified as official when first introduced to the cask
-  url "http://dl.osdn.jp/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
-  appcast 'https://osdn.jp/projects/nndd/releases/rss',
+  url "http://dl.osdn.net/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
+  appcast 'https://dl.osdn.net/projects/nndd/releases/rss',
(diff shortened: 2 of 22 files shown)


Casks/namecoin.rb
```diff
@@ -2,9 +2,10 @@
   version '0.3.80-a00c33d'
   sha256 '5e2432e384b46615d34c00fbe2bcb0a9cbc4f1cc1a27cd72fbffceebcac7eb95'
 
+  # namecoin.info/files/ was verified as official when first introduced to the cask
```
victorpopkov: Just `namecoin.info was verified...`.

Casks/nndd.rb
```diff
@@ -2,11 +2,12 @@
   version '2.4.3,62201'
   sha256 '6a73dcad2e73d877ad1503ed1162cae1a1c84f21d1abaa6aaf9b31bb2fbca531'
 
+  # dl.osdn.jp/nndd/ was verified as official when first introduced to the cask
   url "http://dl.osdn.jp/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
   appcast 'https://osdn.jp/projects/nndd/releases/rss',
```
victorpopkov: Host `dl.osdn.net` can be used here for both download and appcast URLs to avoid "host differs" comment addition.

Casks/paymowidget.rb
```diff
@@ -4,7 +4,7 @@
 
   url "https://app.paymoapp.com/desktop-apps/paymowidget/PaymoWidget_mac_#{version}.dmg"
   name 'Paymo Widget'
-  homepage 'https://app.paymoapp.com/'
+  homepage 'https://app.paymoapp.com/auth/login/'
```
victorpopkov: I would leave this link as is since it just points to the authorization page.

Casks/polar-websync.rb
```diff
@@ -5,7 +5,7 @@
   # polarpersonaltrainer.com was verified as official when first introduced to the cask
   url "https://www.polarpersonaltrainer.com/downloads/websync_#{version}.dmg"
   name 'Polar WebSync Software'
-  homepage 'https://www.polar.com/us-en/support/downloads/Polar_WebSync_Software'
+  homepage 'https://support.polar.com/us-en/support/downloads/Polar_WebSync_Software'
```
victorpopkov: Link without localization can be used here instead: `https://support.polar.com/support/downloads/Polar_WebSync_Software`

Casks/porthole.rb
```diff
@@ -1,12 +1,11 @@
 cask 'porthole' do
-  version '1.7.9'
-  sha256 '805847cfec098073feabccd7f65f448c27c3db14574ac4a94b823458e525ebe3'
+  version :latest
```
victorpopkov: The versioned URL is still available so there is no point in changing this.

victorpopkov (changes requested):

Casks/porthole.rb
miccal: But I thought the rule for a `version`ed Cask was that the `version` must appear in the download `url`, the `addcast` or the `name`, and if not the `version` should be `:latest`?

victorpopkov: If the **appcast exists, the cask should always be versioned**. Even when versioned download URL doesn't exist but the app still has an appcast which points to the same latest download URL on each new release (for example [yate.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/yate.rb) or [yummy-ftp.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/yummy-ftp.rb)) the cask still should be versioned.

The `:latest` should be used only when there is **no versioned download URL available and no appcast**.

Casks/nndd.rb
```diff
@@ -2,11 +2,11 @@
   version '2.4.3,62201'
   sha256 '6a73dcad2e73d877ad1503ed1162cae1a1c84f21d1abaa6aaf9b31bb2fbca531'
 
-  url "http://dl.osdn.jp/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
-  appcast 'https://osdn.jp/projects/nndd/releases/rss',
+  url "http://dl.osdn.net/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
+  appcast 'https://dl.osdn.net/projects/nndd/releases/rss',
```
victorpopkov: Oops, the appcast doesn't use subdomain "dl". Just use `https://osdn.net/projects/nndd/releases/rss`.

victorpopkov (changes requested):

Casks/porthole.rb
```diff
@@ -2,11 +2,10 @@
   version '1.7.9'
   sha256 '805847cfec098073feabccd7f65f448c27c3db14574ac4a94b823458e525ebe3'
 
-  url "https://download.getporthole.com/Porthole-v#{version}.zip"
-  appcast 'https://update.getporthole.com/appcast.rss',
-          checkpoint: 'a3602cb1b9271bafaa09b607760da032f2bedaa6315238a339ae22c607510197'
+  # download.getporthole.com/ was verified as official when first introduced to the cask
+  url 'https://download.getporthole.com/Porthole-latest.zip'
```
victorpopkov: The comment should be `# getporthole.com was verified as official when first introduced to the cask`. The subdomain must be included only when we are dealing with a software distributor who uses a subdomain as the identification part.

In addition, the appcast shouldn't be removed since it's still available. Without an appcast, the `:latest` should be used here instead.

victorpopkov (changes requested):

Casks/porthole.rb
```diff
@@ -2,8 +2,10 @@
   version '1.7.9'
   sha256 '805847cfec098073feabccd7f65f448c27c3db14574ac4a94b823458e525ebe3'
 
-  # download.getporthole.com/ was verified as official when first introduced to the cask
+  # download.getporthole.com was verified as official when first introduced to the cask
```
victorpopkov: Just `# getporthole.com was verified...`.

victorpopkov (changes requested):

Casks/nndd.rb
miccal: The `audit` fails as it expects `osdn.jp`.

Casks/nndd.rb
```diff
@@ -2,11 +2,11 @@
   version '2.4.3,62201'
   sha256 '6a73dcad2e73d877ad1503ed1162cae1a1c84f21d1abaa6aaf9b31bb2fbca531'
 
-  url "http://dl.osdn.jp/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
-  appcast 'https://osdn.jp/projects/nndd/releases/rss',
-          checkpoint: '4f1d23fd9ebe4b3abf7495cf512df2f72f2097b3b4ff51d21e5fa837bb88766b'
+  url "http://dl.osdn.net/nndd/#{version.after_comma}/NNDD_v#{version.before_comma.dots_to_underscores}.dmg"
```
victorpopkov: Well, that's sad. Audit fails if `.net` is used for OSDN in download URL. I suppose it's better to leave `.jp` then everywhere (`url`, `appcast` and `homepage`) and I will add a filter to `cask-homepage` script to ignore them in the future just like SourceForge links.

victorpopkov (changes requested):

victorpopkov (approved):

PR #27175, opened 2016-11-27
Update github-desktop.rb for zap

nuomi1:

Update github-desktop.rb for zap

### DIFF
--- a/Casks/github-desktop.rb
+++ b/Casks/github-desktop.rb
@@ -22,8 +22,8 @@ cask 'github-desktop' do
                          'com.github.GitHub.GHInstallCLI',
                        ]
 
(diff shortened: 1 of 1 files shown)


Casks/github-desktop.rb
```diff
@@ -23,9 +23,11 @@
                        ]
 
   zap delete: [
+                '~/.config/git',
```
fanquake: I'm not sure about this entry.

miccal: There are other programs other than  Github Desktop that may create and/or use `~/.config/git` - I think it would be better to replace this with `zap rmdir: '~/.config/git'`.

PR #27184, opened 2016-11-27
Update homepage for 9 Casks

miccal:
Update homepage for 19 Casks.

Ping @victorpopkov - these are all the ones on your updated list that are **not** _bare_slash_ warnings.

### DIFF
--- a/Casks/cellular.rb
+++ b/Casks/cellular.rb
@@ -5,7 +5,7 @@ cask 'cellular' do
   # github.com/downloads/monofonik was verified as official when first introduced to the cask
   url "https://github.com/downloads/monofonik/cellular/Cellular-#{version}.dmg"
   name 'Cellular'
-  homepage 'http://www.flipt.org/#cellular/'
+  homepage 'http://www.flipt.org/#cellular'
 
   suite 'Cellular'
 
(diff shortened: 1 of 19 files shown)


Casks/cellular.rb
```diff
@@ -5,7 +5,7 @@
   # github.com/downloads/monofonik was verified as official when first introduced to the cask
   url "https://github.com/downloads/monofonik/cellular/Cellular-#{version}.dmg"
   name 'Cellular'
-  homepage 'http://www.flipt.org/#cellular'
+  homepage 'http://www.flipt.org/#cellular/'
```
victorpopkov: There is a small bug in the script at the moment, that adds a trailing slash to fragment part of the URL. All cases where it happens should be ignored at the moment.

victorpopkov (changes requested):

Casks/cellular.rb
miccal: Removed the 10 that had trailing slashes - sorry for the trouble.

victorpopkov (approved):

victorpopkov:
Thanks again for your contribution!

PR #27185, opened 2016-11-27
Update little-snitch to 3.7

miccal:

### DIFF
--- a/Casks/little-snitch.rb
+++ b/Casks/little-snitch.rb
@@ -10,6 +10,8 @@ cask 'little-snitch' do
 
   installer manual: 'Little Snitch Installer.app'
 
+  uninstall script: '/Library/Little Snitch/Little Snitch Uninstaller.app/Contents/MacOS/Little Snitch Uninstaller'
+
   zap delete: [
                 '/Library/Application Support/Objective Development/Little Snitch',
                 '~/Library/Application Support/Little Snitch',


Casks/little-snitch.rb
```diff
@@ -10,6 +10,8 @@
 
   installer manual: 'Little Snitch Installer.app'
 
+  uninstall script: '/Library/Little Snitch/Little Snitch Uninstaller.app/Contents/MacOS/Little Snitch Uninstaller'
```
vitorgalvao: This will call a GUI, so it’s no good.

vitorgalvao (changes requested):

Casks/little-snitch.rb
miccal: Of course, apologies. Any idea how we can add this then? The Little Snitch documentation says to run the app /Library/Little Snitch/Little Snitch Uninstaller.app.

vitorgalvao: A possible solution would be to contact the developer and ask if it would be possible to have a CLI script to perform the same steps the GUI does.

miccal: Should we instead add a caveat to the Cask, something like "To uninstall Little Snitch run the .app located at  /Library/Little Snitch/Little Snitch Uninstaller.app" or similar?

vitorgalvao: No, we need to properly fix it.

miccal: Here is a screenshot of the relevant documentation:

screen shot 2016-11-27 at 23 10 54

The only thing I can think of at the moment is to move the files in the attached image from the `zap` stanza to an `uninstall` stanza.

I know that is not perfect, but I cannot see how to properly fix this.

vitorgalvao: > but I cannot see how to properly fix this.

Like I said, contacting the developer is a good first step. If they don’t reply or say “no”, then it’s time to think of another thing.

miccal: It's worth a try, I will send them an email (I actually use this program and this have a valid licence).

Sorry I could not be of more help.

miccal: @vitorgalvao I received a reply from the developers of Little Snitch - their response was that running the Little Snitch Uninstaller.app is the only way to completely remove the program and they claimed that a `.sh` script would not be as thorough and therefore saw no need to develop one.

Although my comment above about an `uninstaller manual:` was in jest, in all seriousness could it be a possibility as a last resort for an edge case like this one?

PR #27207, opened 2016-11-28
Update p4 to 2016.2.1457671

mdouglass:

### DIFF
--- a/Casks/p4.rb
+++ b/Casks/p4.rb
@@ -1,6 +1,6 @@
 cask 'p4' do
-  version '2016.1.1454320'
-  sha256 '205ee5dafca52b7dd069c9273dd3fa4918134828fa798b4667c89adbd66b7891'
+  version '2016.2.1457671'
+  sha256 'a18b0e0338d6331588afb12c57427ddf169dcf374e5e2ab000001bd0260efc1c'
 
   url "http://filehost.perforce.com/perforce/r#{version.major_minor[2..-1]}/bin.macosx105x86_64/p4"
   name 'Perforce Command-Line Client'


vitorgalvao (approved):

neybis:
@mdouglass Just tried this today and I'm still running into issues. Looks like the sha256 was updated to
`a18b0e0338d6331588afb12c57427ddf169dcf374e5e2ab000001bd0260efc1c`
but should be
`1e6165711902b65612637aa190e7ba09a8559f1e4e057e2b4587e7756c2bef59`


Here is my install output:
```
==> Downloading http://filehost.perforce.com/perforce/r16.2/bin.macosx105x86_64/p4
==> Verifying checksum for Cask p4
==> Note: running "brew update" may fix sha256 checksum errors
Error: sha256 mismatch
Expected: a18b0e0338d6331588afb12c57427ddf169dcf374e5e2ab000001bd0260efc1c
Actual: 1e6165711902b65612637aa190e7ba09a8559f1e4e057e2b4587e7756c2bef59
```

Source ([http://filehost.perforce.com/perforce/r16.2/bin.macosx105x86_64/SHA256SUMS](http://filehost.perforce.com/perforce/r16.2/bin.macosx105x86_64/SHA256SUMS)):
```
6273c2841e8a1241ddee19fe0e04f7a4e3295492085689287ddf6384a88c7cc4 *p4ftpd
1e6165711902b65612637aa190e7ba09a8559f1e4e057e2b4587e7756c2bef59 *p4
ed046dad2478a4f67626adc5bcf3c08fa6571624fc4957b6ee74ccbd2ec771c7 *p4api.tgz
```


I forked the repo and started to prepare a pull request. Ran through the first two tests cases (brew cask audit/style) and it passed. Wasn't sure exactly how to go about including the cask's name and version. I checked the other logs in the repo and I couldn't find a good example?? Seems like most list the cask name and the pull request number (long time git user, new to github, how can I get the pull request number before requesting the pull?), but I could be wrong. What exacty should I put in my commit message?

If you wanted to fix this, that'd be fine. But I can see myself using this project a lot in the future and I always like to contribute where ever I can. Just need some tips on that last item of the task list really.

vitorgalvao:
> long time git user, new to github, how can I get the pull request number before requesting the pull?

You don’t Github does it automatically when maintainers merge as squash commits.

> What exacty should I put in my commit message?

Something like “Updated {{cask_name}} to {{new_version}}”, or “{{cask_name}}: updated to {{new_version}}”.

neybis:
Thank you very much @victorpopkov 
I've submitted the pull request for the sha256 update (#27313).

mdouglass:
So looked into this a bit and what Perforce is doing here is a bit annoying. They use a fixed URL for major/minor version number (http://filehost.perforce.com/perforce/r16.2/bin.macosx105x86_64/p4) but reuse that URL when they do a patch release.

When I made this PR, they were at `Rev. P4/MACOSX105X86_64/2016.2/1457671 (2016/10/26)`. As of today, the same URL now returns a different binary that reports as `Rev. P4/MACOSX105X86_64/2016.2/1468155 (2016/11/28)` (as seen in @neybis PR #27313).

The best thing to do would probably be to contact Perforce and find out if there's a URL that homebrew can use that will not change for every patch release.

neybis:
Ahhh, I didn't notice that with the patch releases, def a good catch although I'm sure it would have been obvious to me sooner or later :P I can see if I can get in touch with Perforce since the company I work for uses it.

If not perhaps the best solution would be to skip the sha check?? Maybe not since I just realized that a `brew upgrade` wouldn't pull down the patch release though since cask doesn't see the version change...

mdouglass:
I sent a support ticket into perforce through my company as well. Received the following:

>I have attached your details to an existing enhancement request for this functionality. Should this be implemented you will be notified by email at the time of release.
> 
> Your request will add weight to this enhancement request.

So it seems like multiple people making the request would be helpful.

neybis:
Definitely, I'll make sure to submit something from my end, too, to try to push this through.

One thing I just thought about (which may not be supported by design), is supporting sha's at runtime. We could always download the file containing the sha, pull that out, and then check against that instead of hard coding a sha.

My biggest concern, if it were even possible to do this, is how brew cask would know an update was available... I'd imagine it would pull down the most recent patch build for the first install and it would pass even if a new patch build is released bc we pull the sha for the patch build out of the sha file. But I would guess that it wouldn't pull down the next patch release. I'd expect it wouldn't realize there is an updated until the minor version is bumped.

PR #27250, opened 2016-11-29
Fix version in Dropbox.app v15.4.22

Saklad5:

- [ ] Checked that the cask was not already refused in [closed issues].

Fix checksum in Dropbox.app Cask
Add appcast in Dropbox.app Cask

### DIFF
--- a/Casks/dropbox.rb
+++ b/Casks/dropbox.rb
@@ -1,8 +1,11 @@
 cask 'dropbox' do
-  version :latest
-  sha256 :no_check
+  version '15.4.22'
+  sha256 'a1569dfdea08b585ecb9f50569f4d26bf86434957d604d7bd18498606c7e6a56'
 
-  url 'https://www.dropbox.com/download?plat=mac&full=1'
+  # clientupdates.dropboxstatic.com was verified as official when first introduced to the cask
+  url "https://clientupdates.dropboxstatic.com/client/Dropbox%20#{version}.dmg"
+  appcast 'https://www.dropbox.com/release_notes/rss.xml',
+          checkpoint: 'e5f348f7f83a2c283a5e8bada332407fc200bbadaedc5f3a5d5ef7be39b25922'
   name 'Dropbox'
   homepage 'https://www.dropbox.com/'
 


Saklad5:
I added the line about the download URL being verified myself, as it was required for the Cask to run. The URL comes directly from both dropbox.com and the extended attributes of the downloaded file, so it is definitely legitimate. Am I allowed to add that line, or should I have let a maintainer do it?

vitorgalvao:
> Am I allowed to add that line, or should I have let a maintainer do it?

You are allowed to add, a maintainer just has to verify it.

vitorgalvao (approved):
We refused such a change in https://github.com/caskroom/homebrew-cask/pull/24975. However, that one did not have an appcast, and I feel like the appcast warrants making this versioned, since the point about it being an obtuse process to get the version number no longer applies.

miccal:
As mentioned in [#24975](https://github.com/caskroom/homebrew-cask/pull/24975) Dropbox (unlike most auto-updating apps) updates silently, and my currently installed version is `14.4.19`. and is claiming to be up-to-date.

I would much prefer this left as `:latest` and `:no_check`.

Saklad5:
I think you’re a bit mistaken, quite understandably. When Dropbox says “Up to date,” it is referring to your files. In other words, that means it has finished syncing.

miccal:
No - I can query Dropbox to check if itself is up-to-date, and it claims it is.

Saklad5:
I’m pretty sure that is still just the auto-update rollout in action. Go over to [](dropbox.com) and download it. Or go to their [official release notes](https://www.dropbox.com/release_notes). Either way, the current version is clearly [15.4.22](https://www.dropboxforum.com/t5/Desktop-client-builds/Stable-Build-15-4-22/m-p/195732#M3502).

miccal:
That is exactly my point - the updates are rolled out automatically in stages; first, business uses get the update, then paying personal users, then personal users who use it for free. Many of us use some kind of Cask update scripts, and this change will cause my installed version of Dropbox to be overwritten by the newer one, only then for Dropbox to auto-update me back to the lower one as that is the official latest version for me.

Saklad5:
You can’t be serious. That isn’t how auto-updates work _anywhere_. Besides, a new Dropbox user would download that version, so obviously it isn’t going to revert them.

vitorgalvao:
> updates are rolled out automatically in stages; first, business uses get the update, then paying personal users, then personal users who use it for free.
> (…)
> my installed version of Dropbox to be overwritten by the newer one, only then for Dropbox to auto-update me back to the lower one as that is the official latest version for me.

That sounds weird to me. Do you have a source for official documentation? I sincerely doubt Dropbox would downgrade users depending on their plan. If anything, it would make more sense to update them in the reverse (like Google supposedly does with Gmail and G Suite users): your free customers get first the thing that might break.

miccal:
Nothing official, just my experience after many years of using Dropbox on many machines and collaborating with people who have either paid personal and business plans.

vitorgalvao:
Even if they do upgrade paying users first, that still does not mean they downgrade other users. To take action in that regard, we’d really need something official (or at least multiple unofficial reports).

This also has `auto_updates`, so it’s not supposed to be updated by HBC once installed. If unofficial scripts ignore `auto_updates`, that’s their bug, not ours.

I’m still for merging this, unless there are other issues I haven’t considered.

Saklad5:
As far as I’m concerned, `auto_updates` is irrelevant. Versioned downloads are necessary for checksums to be used, and that’s reason enough to implement them whenever possible.

vitorgalvao:
@Saklad5 Not when casks are left abandoned and as such users are stuck on old, insecure versions.

Also, `auto_updates` is extremely relevant, **especially** in this case. When `auto_updates` is not respected, the app **will** be downgraded, and that is not only a potential security risk, it’s a bad experience.

Saklad5:
Fair enough.

Saklad5:
Well, are you going to merge this?

vitorgalvao:
> Well, are you going to merge this?

There’s no rush. I‚m giving it a while in case other maintainer has an issue with it.

colindunn:
#24975 was mine and I'd definitely prefer to see this go through. Completely understand the need for an appcast url here so that it doesn't get abandoned.

I will happily become a maintainer for the package as well to make sure it stays updated regularly.

adidalal (approved):

reitermarkus (approved):

Saklad5:
I try to keep all of the Casks I have installed updated for everyone else anyway, but thanks for the offer.

PR #27315, opened 2016-12-02
pokerstars.rb: add region support

vitorgalvao:

---

@reitermarkus Requesting some help here. I feel like this could be more elegant. There are a few things to consider:

+ PokerStars will only work if you have the correct version for your region. There used to be an `EU` version, but since now in Portugal it is `PT`, I’m guessing they have one for each country specifically.
  + Because of that, there really shouldn’t be a “default” version. Would it be possible to make it so that a user *has to* specify a region, else it would fail to install?
+ PokerStars redirects you like crazy. They seriously do not by any means want you visiting the the pages for the other regions, which makes verifying any region but yours a pain (which is why I only added one language).

### DIFF
--- a/Casks/pokerstars.rb
+++ b/Casks/pokerstars.rb
@@ -2,21 +2,25 @@ cask 'pokerstars' do
   version :latest
   sha256 :no_check
 
-  language 'en', default: true do
-    %w[com]
+  language 'US', default: true do
+    %w[.com]
   end
 
+  language 'AU' do
+    %w[.eu/de EU]
+  end
+
+  launguage 'BE' do
+    %w[.be BE]
+  end
+  
   language 'PT' do
-    %w[pt PT]
+    %w[.pt PT]
   end
 
-  language 'eu' do
-    %w[eu EU]
-  end
-
-  url "https://www.pokerstars.#{language[0]}/PokerStars#{language[1]}.app.zip"
+  url "https://www.pokerstars#{language[0]}/PokerStars#{language[1]}.app.zip"
   name 'PokerStars'
-  homepage "https://www.pokerstars.#{language[0]}"
+  homepage "https://www.pokerstars#{language[0]}"
 
   auto_updates true
   container nested: "PokerStars#{language[1]}/PokerStars#{language[1]}.dmg"
--- a/Casks/pokerstars.rb
+++ b/Casks/pokerstars.rb
@@ -6,11 +6,11 @@ cask 'pokerstars' do
     %w[.com]
   end
 
-  language 'AU' do
-    %w[.eu/de EU]
+  language 'AT' do
+    %w[.eu EU]
   end
 
-  launguage 'BE' do
+  language 'BE' do
     %w[.be BE]
   end
   
--- a/Casks/pokerstars.rb
+++ b/Casks/pokerstars.rb
@@ -6,7 +6,7 @@ cask 'pokerstars' do
     %w[com]
   end
 
-  language 'pt' do
+  language 'PT' do
     %w[pt PT]
   end
 
--- a/Casks/pokerstars.rb
+++ b/Casks/pokerstars.rb
@@ -2,17 +2,30 @@ cask 'pokerstars' do
   version :latest
   sha256 :no_check
 
-  url 'https://www.pokerstars.com/PokerStars.app.zip'
+  language 'en', default: true do
+    %w[com]
+  end
+
+  language 'pt' do
+    %w[pt PT]
+  end
+
+  language 'eu' do
+    %w[eu EU]
+  end
+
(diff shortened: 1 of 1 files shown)


Numbermaniac:
Maybe that needs to be done for the homepage too? Looks like https://www.pokerstars.pt/ has its own homepage.

vitorgalvao:
@Numbermaniac I had intentionally not done that to the homepage since it redirects you anyway. But it actually makes more sense since it makes the comment no longer necessary.

I’ve also changes `us` to `en`. Since it’s the default, we can refine it later if necessary.

Also going to add the EU version, seems like that may still be around.

reitermarkus:
Seems like `pokerstars.pt` is an exception, as `pokerstars.com` redirects me to `pokerstars.eu` and there specifically is no option to switch to Portoguese, but a bunch of other languages.

All of these link to `pokerstars.eu/xx/`, except for:

- Latvian: `pokerstars1.eu` (actually links to `pokerstars.eu/lt/` but redirects)
- Magyar `pokerstars-cc.eu`
- Romanian `pokerstars.ro`

vitorgalvao:
@reitermarkus Even if I go to the Romanian one, when trying to download I still get the PT software.

In terms of cask, this is easy to fix: either we scale it back further — `homepage "https://www.pokerstars.#{language[0]}"` becomes `homepage "https://www.pokerstars#{language[0]}"` (the `.` itself is removed and added in `language` — or we go back to `homepage 'https://www.pokerstars.com'` and let the redirect happen on their own.

I don’t have a strong preference. I wouldn’t mind `homepage "https://www.pokerstars.#{language[0]}"` for now and see how it goes. It would be fairly trivial to change to `.com` always if that ends up being too much.

What does matter is the termination in the software’s name. We can only add the ones we’re certain about. Which termination do you have, from which location?

reitermarkus:
Downloading from Austria I get `PokerStarsEU.app` (via `https://www.pokerstars.eu/de/client/download/`, although `https://www.pokerstars.eu/client/download/` is the same).

Casks/pokerstars.rb
```diff
@@ -2,17 +2,30 @@
   version :latest
   sha256 :no_check
 
-  url 'https://www.pokerstars.com/PokerStars.app.zip'
+  language 'en', default: true do
+    %w[com]
+  end
+
+  language 'pt' do
```
reitermarkus: This should probably be `'PT'`, since this is region-specific, not language-specific.

Casks/pokerstars.rb
```diff
@@ -2,17 +2,30 @@
...
+    %w[pt PT]
+  end
+
+  language 'eu' do
```
reitermarkus: Also `eu` is not a valid language code, so this should be similar to https://github.com/caskroom/homebrew-cask/blob/9a91f5ae67979652d72943e76d9f5c368f7058d6/Casks/world-of-tanks.rb#L5.

reitermarkus:
Also I just found this: https://www.pokerstars.eu/languages/

vitorgalvao:
Fixed with the languages we’re sure of.

> Also I just found this: https://www.pokerstars.eu/languages/

Can’t see it, it redirects me to the PT site once more (it also does so with help articles, it’s really annoying and user-hostile).

adidalal:
screen shot 2016-12-04 at 5 12 54 pm

Casks/pokerstars.rb
```diff
@@ -2,17 +2,34 @@
   version :latest
   sha256 :no_check
 
-  url 'https://www.pokerstars.com/PokerStars.app.zip'
+  language 'US', default: true do
+    %w[.com]
+  end
+
+  language 'AU' do
```
reitermarkus: I assume you mean Austria, which is `AT`.

Casks/pokerstars.rb
```diff
@@ -2,17 +2,34 @@
...
+    %w[.eu/de EU]
```
reitermarkus: Just `.eu` is enough. I just rechecked and on the downloads page (`https://www.pokerstars.eu/de/poker/download/mac/`) it is actually linked without `/de/`.

Casks/pokerstars.rb
```diff
@@ -2,17 +2,34 @@
...
+  end
+
+  language 'AT' do
+    %w[.eu EU]
+  end
+
+  language 'BE' do
+    %w[.be BE]
+  end
+
```
Numbermaniac: think you left an accidental whitespace here :)

Numbermaniac:
https://web.archive.org/web/20161205070028/https://www.pokerstars.com/languages/

reitermarkus (approved):

PR #27343, opened 2016-12-03
Update silicon-labs-vcp-driver.rb

davea:
The change to remove `container nested:` in 3348a2b52b2dd348ff2e113069e552e1c8a16ef9 (#23322) seems to have broken this cask, which has a `.pkg` inside the `.dmg`. The autodetection didn't find the `.pkg`, causing "Error: pkg source file not found:"

- [x] `brew cask audit --download silicon-labs-vcp-driver` is error-free.
- [x] The commit message includes the cask’s name but not version because it's an unversioned cask.

### DIFF
--- a/Casks/silicon-labs-vcp-driver.rb
+++ b/Casks/silicon-labs-vcp-driver.rb
@@ -7,6 +7,8 @@ cask 'silicon-labs-vcp-driver' do
   name 'CP210x USB to UART Bridge VCP Driver'
   homepage 'https://www.silabs.com/products/mcu/pages/usbtouartbridgevcpdrivers.aspx'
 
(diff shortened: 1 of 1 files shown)


Casks/silicon-labs-vcp-driver.rb
```diff
@@ -6,6 +6,8 @@
   name 'Silicon Labs VCP Driver'
   name 'CP210x USB to UART Bridge VCP Driver'
   homepage 'https://www.silabs.com/products/mcu/pages/usbtouartbridgevcpdrivers.aspx'
+
```
miccal: Please remove the whitespace in this line.

davea: Yep, have done so.

PR #27371, opened 2016-12-04
Update webstorm.rb for zap

nuomi1:

Update webstorm.rb for zap

### DIFF
--- a/Casks/webstorm.rb
+++ b/Casks/webstorm.rb
@@ -6,6 +6,7 @@ cask 'webstorm' do
   name 'WebStorm'
   homepage 'https://www.jetbrains.com/webstorm/'
 
+  auto_updates true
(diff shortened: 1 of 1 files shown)


vitorgalvao:
If an app auto-updates, that means it has an appcast somewhere. Are you able to find where it is? The [`appcast` documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) tells how to find appcasts for the most common methods.

nuomi1:
I just think it is the same as [intellij-idea-ce](https://github.com/caskroom/homebrew-versions/blob/master/Casks/intellij-idea-ce.rb).

vitorgalvao:
And that one should have an `appcast` as well.

PR #27397, opened 2016-12-05
Added Now

leo:
Just added [Now](https://zeit.co/app) from my company @zeit to the Cask list! 😊 

----

### DIFF
--- a/Casks/now.rb
+++ b/Casks/now.rb
@@ -7,5 +7,7 @@ cask 'now' do
   name 'Now'
   homepage 'https://zeit.co/now'
 
+  auto_updates true
+
   app 'Now.app'
 end
--- a/Casks/now.rb
+++ b/Casks/now.rb
@@ -4,6 +4,8 @@ cask 'now' do
 
   # now-auto-updates.now.sh was verified as official when first introduced to the cask
   url 'https://now-auto-updates.now.sh/download/macos'
+  appcast 'https://github.com/zeit/now-desktop/releases.atom',
+          checkpoint: '2fdd5ed203f96a0d7c81f0c6694d2604dae52baf732dbf2a8f2aa1f2826d7761'
   name 'Now'
   homepage 'https://zeit.co/now'
 
--- a/Casks/now.rb
+++ b/Casks/now.rb
@@ -1,0 +1,11 @@
+cask 'now' do
+  version :latest
+  sha256 :no_check
+
+  # now-auto-updates.now.sh was verified as official when first introduced to the cask
+  url 'https://now-auto-updates.now.sh/download/macos'
+  name 'Now'
+  homepage 'https://zeit.co/now'
+
+  app 'Now.app'
+end


vitorgalvao:
It says on the webiste it auto-updates, so it needs [a line like this](https://github.com/caskroom/homebrew-cask/blob/d145da6bdb212f5e855178723bca93d59a05b469/Casks/github-desktop.rb#L11).

Also, if it auto-updates, that means it has an appcast somewhere. Can you add it? The [`appcast` documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) tells how.

leo:
@vitorgalvao Done! 

Do I need to update the `appcast` sha each time we update the app? Because that's exactly that I was trying to avoid by setting `version` to ":latest".

miccal:
> Do I need to update the appcast sha each time we update the app? Because that's exactly what I was trying to avoid by setting version to ":latest".
Yes, if there is an `appcast` then the `version` and `sha256` also need to be set.

leo:
@miccal Then I should probably remove the `appcast` field again. Because [this](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md) docs states that I don't need `sha256` and `version` if I want it to be unversioned download by cask.

Casks/now.rb
```diff
@@ -0,0 +1,15 @@
+cask 'now' do
+  version :latest
+  sha256 :no_check
+
+  # now-auto-updates.now.sh was verified as official when first introduced to the cask
+  url 'https://now-auto-updates.now.sh/download/macos'
```
vitorgalvao: Please change this to the URL provided in the releases page of github.

Casks/now.rb
```diff
@@ -0,0 +1,15 @@
+cask 'now' do
+  version :latest
```
vitorgalvao: Please add a `version` and `sha256`.

vitorgalvao (changes requested):

vitorgalvao:
> Because [this](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md) docs states that I don't need `sha256` and `version` if I want it to be treated as unversioned by cask.

Nowhere in that document does it state that’s a preference. Our users *overwhelmingly* prefer versioned casks for security reasons. We’re a service for users to easily install apps they want, not for developers to distribute them how they prefer. Even if you added this as unversioned, a user at a later date would change it to be versioned.

vitorgalvao:
If you follow the instructions in [CONTRIBUTING](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md), updating the whole cask with `cask-repair` is literally one command away.

leo:
@vitorgalvao So there's basically no way to avoid having to push a new update to cask each time a new release comes out? 😊

vitorgalvao:
> So there's basically no way to avoid having to push a new update to cask each time a new release comes out?

No. Note that when the discussion first came up of preferring versioned or unversioned casks, I was on the unversioned camp. Users overwhelmingly prefer versioned casks, however.

However, note that `appcast` was made with maintainers in mind. In the future, we’ll use it so you’ll never again have to update casks with an appcast (an automated system can do so).

leo:
@vitorgalvao Ahh, cool. Thanks for the details! 😊 

I guess I'll revisit this once you've automated the system. In my mind, it's not a solution for me to have to update the app over multiple channels.

vitorgalvao:
Keep in mind no one is making you keep it up to date. Submitting it is not a sentence to update it constantly. Most apps are user submitted and we have many great users (like @miccal and @artdevjs) and maintainers that keep the apps up to date, even the ones we don’t use personally.

All that said, if you as the developer don’t even intend on running the one command to automatically keep it up to date, then I’d rather the submission indeed does not come from you, but from a user later on. That way, if there’s no user interested in having this in HBC, no point in the added overhead.

Thank you for the submission, though! Those are always appreciated.

PR #27459, opened 2016-12-07
Fix torguard.rb for new install method

colindunn:
Unfortunately when updating to 0.3.56 I missed that the packages has been moved to pkg instead of app and found it when I installed. Updating for pkg installs in future and fixing removals.

I haven't included the version number in the commit, assuming that this will be the install method for all future versions.

Also worth noting that the package uninstall works fine, but leaves an empty /Applications/TorGuard.app folder, thus the delete.

--- 

### DIFF
--- a/Casks/torguard.rb
+++ b/Casks/torguard.rb
@@ -9,5 +9,13 @@ cask 'torguard' do
   name 'TorGuard'
   homepage 'https://torguard.net/'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):
Thank you for a detailed explanation of the fixes and why they are need (such as deleting the app bundle because it would otherwise not be cleaned).

PR #27465, opened 2016-12-07
Add libreoffice-language-pack 5.2.3

miccal:

- [X] Checked that the cask was not already refused in [closed issues].

Closes [#27193](https://github.com/caskroom/homebrew-cask/pull/27193).

@vitorgalvao I have removed the `sudo` requirement from the `preflight` block and successfully tested the Cask.

Please note that in the `preflight` block you have to use `/Applications` - I don't think you can use `#{appdir}` as it needs the `#{appdir}` of LibreOffice itself. I am not sure if there is an alternate way?

If there is not, I could add a `caveat` like so:
```
caveats do
 Requires LibreOffice.app to be installed in /Applications.
end
```

### DIFF
--- a/Casks/libreoffice-language-pack.rb
+++ b/Casks/libreoffice-language-pack.rb
@@ -557,7 +557,7 @@ cask 'libreoffice-language-pack' do
   stage_only true
 
   preflight do
-    system '/usr/bin/tar', '-C', '/Applications/LibreOffice.app/', '-xjf', "#{staged_path}/LibreOffice Language Pack.app/Contents/tarball.tar.bz2"
-    system '/usr/bin/touch', '/Applications/LibreOffice.app/Contents/Resources/extensions'
+    system_command '/usr/bin/tar', args: ['-C', '/Applications/LibreOffice.app/', '-xjf', "#{staged_path}/LibreOffice Language Pack.app/Contents/tarball.tar.bz2"]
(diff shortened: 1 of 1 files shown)


vitorgalvao:
It’s likely that you only set `--appdir` once, in your shell’s startup files, so using `#{appdir}` is still the best option.

But you do make a good point, so the caveat could be something like:

```
#{token} assumes LibreOffice is installed in #{appdir}. If it is not, you’ll need to run #{staged_path}/LibreOffice Language Pack.app manually.
```

miccal:
If I change `/Applications` to `#{appdir}`, I get the following error:
```
==> Failed command:
/usr/bin/tar -C \#{appdir}/LibreOffice.app/ -xjf /usr/local/Caskroom/libreoffice-language-pack/5.2.3/LibreOffice Language Pack.app/Contents/tarball.tar.bz2
==> Standard Output of failed command:
==> Standard Error of failed command:
tar: could not chdir to '#{appdir}/LibreOffice.app/'
==> Exit status of failed command:
#<Process::Status: pid 18640 exit 1>
```

reitermarkus:
@miccal, you have to use double quotes when interpolating strings.

miccal:
> @miccal, you have to use double quotes when interpolating strings.

Crap!

vitorgalvao:
Thank you as always, @miccal.

miccal:
No problem.

PR #27540, opened 2016-12-09
Add Bitdefender.app v2.3

dcondrey:

### DIFF
--- a/Casks/bitdefender.rb
+++ b/Casks/bitdefender.rb
@@ -1,0 +1,12 @@
+cask 'bitdefender' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://download.bitdefender.com/mac/antivirus/en/bitdefender_antivirus_for_mac.dmg'
+  name 'Bitdefender Virus Scanner'
+  homepage 'http://www.bitdefender.com/solutions/virus-scanner-for-mac.html'
+
+  pkg 'Bitdefender Installer.pkg'
+
+  uninstall pkgutil: 'com.bitdefender.*'
+end
--- a/Casks/bitdefender.rb
+++ b/Casks/bitdefender.rb
@@ -4,7 +4,7 @@ cask 'bitdefender' do
 
   url 'http://download.bitdefender.com/mac/antivirus/en/bitdefender_antivirus_for_mac.dmg'
   name 'Bitdefender Virus Scanner'
-  homepage 'http://www.bitdefender.com/solutions/virus-scanner-for-mac.html'
+  homepage 'http://www.bitdefender.com/'
 
   pkg 'Bitdefender Installer.pkg'
 


dcondrey:
The dmg file contains a secondary uninstaller pkg which is not installed with the installer package, but seems to do a more thorough uninstall than what I wrote in the stanza.  I'm not sure how to reference that secondary uninstall pkg though.

miccal:
> I'm not sure how to reference that secondary uninstall pkg though.

You can't at this stage - see discussion [here](https://github.com/caskroom/homebrew-cask/issues/24377).

dcondrey:
I thought perhaps I could copy over the uninstall package as a post_install step but not sure.

dcondrey:
@fanquake Why the on hold tag?

adidalal:
Does the app fully uninstall? If not, we can't add it - that's why the tag, I believe.

dcondrey:
The app does uninstall successfully as is.

Casks/bitdefender.rb
```diff
@@ -0,0 +1,12 @@
+cask 'bitdefender' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://download.bitdefender.com/mac/antivirus/en/bitdefender_antivirus_for_mac.dmg'
+  name 'Bitdefender Virus Scanner'
+  homepage 'http://www.bitdefender.com/solutions/virus-scanner-for-mac.html'
```
miccal: Can you confirm this is the correct homepage? I get a 404 error, and going to `http://www.bitdefender.com.au/solutions/antivirus-for-mac.html` I need to enter an email address to download the trial, which is not the one downloaded via the `url` you provided.

Numbermaniac: The given homepage works just fine, testing with an American VPN, but as you said the .au version fails for me too. Maybe we should just make this `http://www.bitdefender.com/`?

miccal: But even from `http://www.bitdefender.com/` I cannot find a link to the file in the original `url`?

Numbermaniac: Is this maybe a link to the paid version, then?

miccal: @dcondrey any follow-up on this?

dcondrey: All links on the website direct to the App Store but I found this direct link, crawling around for it with a little help from Google.

dcondrey:
All links on the website direct to the App Store but I found this direct link, crawling around for it with a little help from Google.

miccal:
Thanks @dcondrey - unfortunately, unless there is an easily verifiable link via the `homepage`, I do not think we can accept this.

reitermarkus:
I found it here: http://www.bitdefender.com/Downloads/

miccal:
@reitermarkus those links do not download the `.pkg` referenced in the original PR - the `.dmg contains a `.pkg` called `Bitdefender_aa9d5ce5-7b01-4743-9eec-5480ce596363.pkg`.

reitermarkus:
You're right, seems like the trial really is only a trial.

- `http://download.bitdefender.com/mac/av/en/bitdefender_antivirus_for_mac.dmg` (15.1 MB)
- `http://download.bitdefender.com/mac/antivirus/en/bitdefender_antivirus_for_mac.dmg` (171.4 MB)

miccal:
Precisely.

dcondrey:
Downloaded this new one and it installs an obnoxious app called BitDefender Central which just seems to be an intermediary forcing me to create a user account.

reitermarkus:
The 171.4 MB one seems to be from 2012.

miccal:
Ah, vintage!

dcondrey:
This Bitdefender software kinda seems to be it's own virus.. it takes up way too much space and resources. :/

miccal:
Thank you for your efforts though @dcondrey, they are appreciated.

dcondrey:
np

PR #27575, opened 2016-12-10
Created a cask for iDrive backup

austincloudguru:


vitorgalvao:

Everything seems fine with this submission so I’ll go ahead and merge it, but do please fill the template next time.

MarsDominion:
Thanks. I wasn't aware of the template, but I wasn't presented it via hub (it was installed via home-brew, so not sure why that fix wouldn't be in).  I will make sure it is in the right format next time.

PR #27732, opened 2016-12-13
Update dropshare to 4.5.1,4555

colindunn:

### DIFF
--- a/Casks/dropshare.rb
+++ b/Casks/dropshare.rb
@@ -1,11 +1,11 @@
 cask 'dropshare' do
-  version '4.4.5'
-  sha256 '779122019003f27b78e9792133902f18019d389cb93de45ce8b1c2670c60bd5c'
+  version '4.5.1,4555'
+  sha256 '1ecdea9e38c819d26377049e6d3138d4a1796ceefdf45ca034ce23185a15c6c0'
 
   # d2wvuuix8c9e48.cloudfront.net was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


Casks/dropshare.rb
```diff
@@ -1,11 +1,11 @@
 cask 'dropshare' do
-  version '4.4.5'
-  sha256 '779122019003f27b78e9792133902f18019d389cb93de45ce8b1c2670c60bd5c'
+  version '4.5.1,4555'
+  sha256 '1ecdea9e38c819d26377049e6d3138d4a1796ceefdf45ca034ce23185a15c6c0'
 
   # d2wvuuix8c9e48.cloudfront.net was verified as official when first introduced to the cask
-  url "https://d2wvuuix8c9e48.cloudfront.net/Dropshare#{version.major}-latest.zip"
+  url "https://d2wvuuix8c9e48.cloudfront.net/Dropshare#{version.major}-#{version.after_comma}.app.zip"
```
miccal: The link on the homepage still points to `https://d2wvuuix8c9e48.cloudfront.net/Dropshare4-latest.zip` - where did you find this `url`?

colindunn: The link is from the appcast which shows all the individual version links for the product.

The problem is that the cask is currently broken and will continue to break on each application update due to the checksum with the existing link.

It either needs to change to this model, or needs to move to :latest if we want to retain the Dropshare4-latest.zip link

PR #27734, opened 2016-12-13
Update inkdrop to 2.6.2

artdevjs:

### DIFF
--- a/Casks/inkdrop.rb
+++ b/Casks/inkdrop.rb
@@ -1,6 +1,6 @@
 cask 'inkdrop' do
-  version '0.11.0'
-  sha256 '23b47aa14a6531b3daf1ce96244b88255e43627a9b450b84220e1bfb70f095f5'
+  version '2.6.2'
+  sha256 'b6fb511c488dc8ca4d03489b1198e07445164a6066998af136a926ad8e5fe666'
 
   url "https://www.inkdrop.info/api/update/Inkdrop-#{version}-Mac.zip"
   name 'Inkdrop'
--- a/Casks/inkdrop.rb
+++ b/Casks/inkdrop.rb
@@ -2,9 +2,19 @@ cask 'inkdrop' do
   version '2.6.2'
   sha256 'b6fb511c488dc8ca4d03489b1198e07445164a6066998af136a926ad8e5fe666'
 
-  url "https://www.inkdrop.info/api/update/Inkdrop-#{version}-Mac.zip"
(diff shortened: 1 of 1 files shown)


Casks/inkdrop.rb
```diff
@@ -1,6 +1,6 @@
 cask 'inkdrop' do
-  version '0.11.0'
-  sha256 '23b47aa14a6531b3daf1ce96244b88255e43627a9b450b84220e1bfb70f095f5'
+  version '2.6.2'
+  sha256 'b6fb511c488dc8ca4d03489b1198e07445164a6066998af136a926ad8e5fe666'
 
   url "https://www.inkdrop.info/api/update/Inkdrop-#{version}-Mac.zip"
```
miccal: Would you mind switching the `url` to their [Github page](https://github.com/inkdropapp/releases/releases) so an `appcast` can be added?

miccal: Also, the `app` is just called `Inkdrop.app`.

artdevjs:
@miccal done

miccal:
Thank you very much @artdevjs.

Casks/inkdrop.rb
```diff
@@ -1,10 +1,20 @@
 cask 'inkdrop' do
-  version '0.11.0'
-  sha256 '23b47aa14a6531b3daf1ce96244b88255e43627a9b450b84220e1bfb70f095f5'
+  version '2.6.2'
+  sha256 'b6fb511c488dc8ca4d03489b1198e07445164a6066998af136a926ad8e5fe666'
 
-  url "https://www.inkdrop.info/api/update/Inkdrop-#{version}-Mac.zip"
+  url "https://github.com/inkdropapp/releases/releases/download/v#{version}/Inkdrop-#{version}-Mac.zip"
```
miccal: Since the `url` and `homepage` are different, please add a comment above the `url` like so:
```
# github.com/inkdropapp was verified as official when first introduced to the cask
```

miccal:
Thanks again @artdevjs.

Casks/inkdrop.rb
vitorgalvao: Small note: it’s not *mandatory* for `url` and `appcast` to use the same domain. We use appcasts mainly for their version numbers, not the URLs they point to, so less care is needed with them.

miccal: Thanks @vitorgalvao - I suggested the change as I could only find a link to the demo version from the `homepage`.

PR #27736, opened 2016-12-13
Updated Doxie from 'latest' to 2.9.1

sbibauw:

Flexible URL scheme verified


sbibauw:
Checking previous edits on doxie.rb, it seems to be more robust to use `latest` with the redirecting URL from Doxie website... And, I don't know how, but it was updated anyway with `homebrew-cask-upgrade`

PR #27772, opened 2016-12-14
Update omnipresence to 1.5.1

colindunn:

### DIFF
--- a/Casks/omnipresence.rb
+++ b/Casks/omnipresence.rb
@@ -1,12 +1,23 @@
 cask 'omnipresence' do
-  version '1.4.1'
-  sha256 '409bf272e7c4dc488f68abadb3e2ef15d4accde10f8ee9babd8b23f522bfe323'
+  if MacOS.version == :mavericks
+    version '1.2-r220586'
+    sha256 '59a214e6393c984aead4a7242297979dc553d59af9964b5c645636e395794f0f'
+    osx_version_path = MacOS.version
+  elsif MacOS.version == :yosemite
+    version '1.4.1'
+    sha256 '409bf272e7c4dc488f68abadb3e2ef15d4accde10f8ee9babd8b23f522bfe323'
+    osx_version_path = MacOS.version
+  else # MacOS.version >= :el_capitan
+    version '1.5.1'
+    sha256 'e785b910e7413986be45ab14f6a86dc598999461bf3c2a9d0da3759848342ac7'
+    osx_version_path = '10.11'
+  end
 
-  url "https://downloads.omnigroup.com/software/MacOSX/10.10/OmniPresence-#{version}.dmg"
+  url "https://downloads.omnigroup.com/software/MacOSX/#{osx_version_path}/OmniPresence-#{version}.dmg"
   name 'OmniPresence'
   homepage 'https://www.omnigroup.com/omnipresence'
 
-  depends_on macos: '>= :yosemite'
+  depends_on macos: '>= :mavericks'
 
   app 'OmniPresence.app'
 end
--- a/Casks/omnipresence.rb
+++ b/Casks/omnipresence.rb
@@ -1,19 +1,16 @@
 cask 'omnipresence' do
   if MacOS.version == :mavericks
-    version '1.2-r220586'
+    version '1.2-r220586,10.9'
     sha256 '59a214e6393c984aead4a7242297979dc553d59af9964b5c645636e395794f0f'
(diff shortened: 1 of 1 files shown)


colindunn:
Hoping my definition of versioning in the path makes sense. Could just use MacOS.version except that both 10.11 and 10.12 are usable on the latest version which doesn't match the others. Anyway, let me know if there is a better way here

Casks/omnipresence.rb
```diff
@@ -1,12 +1,23 @@
...
-  sha256 '409bf272e7c4dc488f68abadb3e2ef15d4accde10f8ee9babd8b23f522bfe323'
+  if MacOS.version == :mavericks
+    version '1.2-r220586'
+    sha256 '59a214e6393c984aead4a7242297979dc553d59af9964b5c645636e395794f0f'
+    osx_version_path = MacOS.version
+  elsif MacOS.version == :yosemite
+    version '1.4.1'
+    sha256 '409bf272e7c4dc488f68abadb3e2ef15d4accde10f8ee9babd8b23f522bfe323'
+    osx_version_path = MacOS.version
+  else # MacOS.version >= :el_capitan
```
vitorgalvao: Remove the comment.

Casks/omnipresence.rb
```diff
@@ -1,12 +1,23 @@
...
+    version '1.5.1'
+    sha256 'e785b910e7413986be45ab14f6a86dc598999461bf3c2a9d0da3759848342ac7'
+    osx_version_path = '10.11'
+  end
 
-  url "https://downloads.omnigroup.com/software/MacOSX/10.10/OmniPresence-#{version}.dmg"
+  url "https://downloads.omnigroup.com/software/MacOSX/#{osx_version_path}/OmniPresence-#{version}.dmg"
```
vitorgalvao: Yes, use `MacOS.version` instead.

Casks/omnipresence.rb
```diff
@@ -1,12 +1,23 @@
...
+    sha256 '59a214e6393c984aead4a7242297979dc553d59af9964b5c645636e395794f0f'
+    osx_version_path = MacOS.version
+  elsif MacOS.version == :yosemite
+    version '1.4.1'
+    sha256 '409bf272e7c4dc488f68abadb3e2ef15d4accde10f8ee9babd8b23f522bfe323'
+    osx_version_path = MacOS.version
+  else # MacOS.version >= :el_capitan
+    version '1.5.1'
+    sha256 'e785b910e7413986be45ab14f6a86dc598999461bf3c2a9d0da3759848342ac7'
+    osx_version_path = '10.11'
```
vitorgalvao: If `10.12` also works, no point in freezing this one, then.

vitorgalvao (changes requested):

colindunn:
The links are as follows:

mavericks: https://downloads.omnigroup.com/software/MacOSX/10.9/OmniPresence-1.2-r220586.dmg
yosemite: https://downloads.omnigroup.com/software/MacOSX/10.10/OmniPresence-1.4.1.dmg
el_capitan: https://downloads.omnigroup.com/software/MacOSX/10.11/OmniPresence-1.5.1.dmg
sierra: https://downloads.omnigroup.com/software/MacOSX/10.11/OmniPresence-1.5.1.dmg

Sierra doesn't use 10.12, which was why I didn't use MacOS.version but used a variable instead, trying to avoid using url in each if block.

Would you prefer me to just use url in each instead of doing it this way?

Casks/omnipresence.rb
colindunn: See below

Casks/omnipresence.rb
colindunn: See below

Casks/omnipresence.rb
colindunn: Will do

vitorgalvao:
Using the variable in some and setting one of them to a fixed value seems inelegant. We don’t really gain anything by using `MacOS.version` in the early versions (which will not receive more updates and hence not change) if for the later versions we need a fixed value.

I’d rather use the `*_comma` methods. So we’d have `version '1.2-r220586,10.9'` and so forth. The URL would use `"https://downloads.omnigroup.com/software/MacOSX/#{version.after_comma}/OmniPresence-#{version.before_comma}.dmg"`.

colindunn:
Done

vitorgalvao (approved):

PR #27863, opened 2016-12-16
updated to v0.7.4

rebootd:


Casks/laverna.rb
```diff
@@ -1,5 +1,5 @@
 cask 'laverna' do
-  version '0.7.1'
+  version '0.7.4-RC1'
```
miccal: This is a [pre-release version](https://github.com/Laverna/laverna/releases) - the latest release is version 0.7.3.

rebootd: Switched to 0.7.3, which has the fix users need to integrate dropbox.

Casks/laverna.rb
```diff
@@ -1,5 +1,5 @@
 cask 'laverna' do
-  version '0.7.1'
+  version '0.7.3'
   sha256 '2044a681fa23583510e1a8a6a151c21cbd083f29cc66c79ad1e40c27c5a5828b'
```
miccal: You will also need to update the [`sha256`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md) and the [`checkpoint`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

rebootd: Is that the sha of their commit for 0.7.3?

rebootd: Closing and using the update script.

PR #27911, opened 2016-12-17
Update reference to macOS versions

neutric:


vitorgalvao:
This is written in a way as to not need to be constantly updated. That’s the point, it’s just an example.

neutric:
That is how I understood it, which is why I kept/updated the time reference. However, writing it in a way that ages well does not necessarily mean that it shouldn't be kept current if someone is willing to spend the time to update it. Merging would have cost you no more time than typing your comment and come next year's release it probably needs to be updated anyway so as to not look too outdated.

vitorgalvao:
Agreed.

PR #27983, opened 2016-12-19
Update scrivener to 2.81.2,106

colindunn:

### DIFF
--- a/Casks/scrivener.rb
+++ b/Casks/scrivener.rb
@@ -1,9 +1,9 @@
 cask 'scrivener' do
-  version '2.8.1.2,106'
+  version '2.81.2,106'
   sha256 'b4a09047b4b7f7522d58fdd5736f7de9d7250447f3573fd993964bfb3b6be1e1'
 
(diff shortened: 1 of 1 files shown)


colindunn:
I really don't like how I've done the version interpretation on this one. The current version is 2.8.1.2 but take a look through the version download links on the appcast here: https://www.literatureandlatte.com/downloads/scrivener-2.xml

Can someone else see a better way to do this? I'm trying to nicely remove the middle dot.

miccal:
W would change the `version` to `2.8-1.2,106` and the `url` to `http://scrivener.s3.amazonaws.com/mac_updates/Scrivener_#{version.after_comma}_#{version.before_comma.no_hyphens}.zip`.

colindunn:
Done. I just liked getting the full proper version there, but fair enough does make it a lot simpler.

vitorgalvao:
I don’t see why not just `version '2.81.2_106'` and `#{version}` in `url`. We don’t parse `appcast` automatically, it’s more of a warning for when casks are updated.

PR #28099, opened 2016-12-21
removed useless depends_on (snow_leopard, :arch)

vitorgalvao:
HBC itself has these dependencies (and even if we didn’t, those are simply to old to care for and support), so these stanzas are redundant.

### DIFF
--- a/Casks/adapter.rb
+++ b/Casks/adapter.rb
@@ -8,7 +8,5 @@ cask 'adapter' do
   name 'Adapter'
   homepage 'https://www.macroplant.com/adapter/'
 
-  depends_on macos: '>= :snow_leopard'
-
   app 'Adapter.app'
 end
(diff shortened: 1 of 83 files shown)


claui:
I like how you spared specific cases like e. g. [this one](https://github.com/caskroom/homebrew-cask/blob/60531a2812005dd5f17dc92f3ce7419af3c5d019/Casks/quicksilver.rb#L2-L4).

We have about 40 cases in total involving `:snow_leopard` for control flow/fallback purposes. I can see how people went out of their way to provide legacy compatibility; I feel this deserves to be honored as long as it doesn’t become a burden, maintenance-wise. I expect those to die out naturally anyway as URLs become stale over time. Also, bulk-rewriting and testing those 40 cases would seem kinda pointless to me. I think it’s best to leave those alone (which you obviously did).

PR looks fine. 👍

claui (approved):

PR #28102, opened 2016-12-21
depends_on.md: leave only 64-bit Intel

vitorgalvao:
The other options are still available, but it’s useless to list them/have them. See also https://github.com/Homebrew/brew/pull/1745.

Pinging @caskroom/maintainers.

### DIFF
--- a/doc/cask_language_reference/stanzas/depends_on.md
+++ b/doc/cask_language_reference/stanzas/depends_on.md
@@ -75,27 +75,18 @@ The available symbols for hardware are:
 
 | symbol     | meaning        |
 | ---------- | -------------- |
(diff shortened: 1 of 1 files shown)


claui:
Throwing out unneeded code is almost always a good thing, especially if it makes maintenance easier.

In theory, there could be someone out there who might rely on Homebrew-cask in combination with private taps to manage their legacy Mac; but then again, even _those_ taps can be trivially fixed.

Also, a quick `grep` reveals no regressions (provided #28099 will be merged of course). Thus, LGTM!

miccal (approved):

claui (approved):

PR #28103, opened 2016-12-21
Add KickMount.app v1.0.1

ghost:
KickMount is a simple tool for the Mac, that lets you unmount one or all external volumes
from the menu bar.

### DIFF
--- a/Casks/kickmount.rb
+++ b/Casks/kickmount.rb
@@ -1,0 +1,14 @@
+cask 'kickmount' do
+  version '1.0.1'
+  sha256 '60573a751ab78fcc1e9b3551f3306582074eac916f3d5eab878fbd130de41a95'
+
+  url "https://github.com/m12n/kickmount/releases/download/#{version}/kickmount-#{version}.dmg"
+  name 'KickMount'
+  homepage 'https://github.com/m12n/kickmount'
+
+  app 'KickMount.app'
+
+  caveats do
+    depends_on_java('8')
+  end
+end


Casks/kickmount.rb
```diff
@@ -0,0 +1,14 @@
+cask 'kickmount' do
+  version '1.0.1'
+  sha256 '60573a751ab78fcc1e9b3551f3306582074eac916f3d5eab878fbd130de41a95'
+
+  url "https://github.com/m12n/kickmount/releases/download/#{version}/kickmount-#{version}.dmg"
```
vitorgalvao: Please add an [`appcast`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

vitorgalvao (changes requested):

vitorgalvao:
On second thought, thank you for the interest but I think we should refuse it.

This is your own app and it has no stars. No one will benefit from the cask and it is added overhead. Although we don’t have hard rules on refusals like this, like Homebrew, we have to draw the line somewhere, and I think this pretty clearly qualifies.

Pinging @caskroom/maintainers anyway, in case anyone disagrees, as this wasn’t widely decided on (but we have made similar decisions in the past).

miccal:
I agree with @vitorgalvao's assessment.

claui:
I do honor @m12n’s effort in authoring and submitting the cask. I’m also convinced the app will be of perfectly good value for some.

That said, we do have some consensus that _Homebrew-cask is not about discoverability._ In other words, a Cask is only useful for people who already know about the app.

If an app lacks notability to the extent that there are no signs of an existing user base, then I concur there is no need for a Cask at that time.

adidalal:
I'm not opposed to adding it - we have plenty of unmaintained Casks that are old forks. However, from a maintainer's perspective, I do want to stop the further addition of Casks that will be left unmaintained.

@m12n You could make a private tap or just put the .rb file in your repo itself. You can always brew cask install a url directly, so that should let you script installs/distribute. We can always revisit addition to the main HBC repo. How does that sound?

m12n:
Sure, no worries. Thanks for your feedback, everybody.

PR #28108, opened 2016-12-21
Remove basex.rb

colindunn:
Download link is no longer presented on the [downloads page](http://basex.org/products/download/all-downloads/) and instead recommends installing "Mac OSX via [Homebrew](https://github.com/Homebrew/homebrew-core/blob/master/Formula/basex.rb)"

@rzyns or @bchatard is there a reason this cask should be kept?

---


rzyns:
Probably not. I think the only officially supported way to run it is using
their shell script, and I don't think they even bundle it as Mac App
anymore.

colindunn:
Thanks for coming back quickly @rzyns

reitermarkus:
Also add it to [`tap_migrations.json`](https://github.com/caskroom/homebrew-cask/blob/8f7c03cb52bf6340da34db8c676904bbd2c972a9/tap_migrations.json).

colindunn:
Done

adidalal (approved):

PR #28161, opened 2016-12-22
Update squeak to 5.1-16548

colindunn:


vitorgalvao:
No point in the 32bit version, use the 64 instead (simply change it in the url, and you’ll get it).

colindunn:
Sorry, I meant to add a comment to this one @vitorgalvao.

The app has previously been 32 bit and they now have 64 bit versions, but the [download page](http://squeak.org/downloads/) lists them as "experimental". The 32 bit version is stable so kept it at that for the moment.

PR #28164, opened 2016-12-22
Added appcast to Objective-See apps

vitorgalvao:

---

Yes, these all use the same appcast, which means some will trigger when there isn’t a new release. That said, these are all extremely focused apps with a clear scope, so they shouldn’t get too many updates (but those are important to get, since they deal with security of your system).

Also changed the caveat on Ostiarius. It makes it easier to identify and remove some day.

### DIFF
--- a/Casks/oversight.rb
+++ b/Casks/oversight.rb
@@ -4,8 +4,6 @@ cask 'oversight' do
 
   # bitbucket.org/objective-see was verified as official when first introduced to the cask
   url "https://bitbucket.org/objective-see/deploy/downloads/OverSight_#{version}.zip"
-  appcast 'https://objective-see.com/products.json',
-          checkpoint: 'b2004932186e0eb176a9ac01fe499de292d502d7a17559e900822f9cdbfa74d6'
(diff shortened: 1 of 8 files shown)


Casks/oversight.rb
```diff
@@ -4,6 +4,8 @@
 
   # bitbucket.org/objective-see was verified as official when first introduced to the cask
   url "https://bitbucket.org/objective-see/deploy/downloads/OverSight_#{version}.zip"
+  appcast 'https://objective-see.com/products.json',
+          checkpoint: 'b2004932186e0eb176a9ac01fe499de292d502d7a17559e900822f9cdbfa74d6'
   appcast 'https://objective-see.com/products/versions/oversight.json',
           checkpoint: 'da8cba3af8fa83a106e23be6299fd286fe8377931906270bfdd885a609a1ae8b'
```
miccal: Two `appcasts`?

vitorgalvao: My mistake, as I automatically added them all. Rereading, I did run `style` on them, but forgot `audit`.

miccal: Tisk tisk!

PR #28176, opened 2016-12-22
ioquake3-1.36: Use correct path in advice

yurikoles:

Current cask uses bad advice maybe due to historical reasons.
```
/Applications/ioquake3/ioquake3-1.36.app/Contents/MacOS/ioquake3.ub 
ioq3 1.36 macosx-i386 Apr 22 2009
----- FS_Startup -----
Current search path:
/Users/yurikoles/Library/Application Support/Quake3/baseq3
[...]
/Applications/ioquake3/baseq3
```

### DIFF
--- a/Casks/ioquake3.rb
+++ b/Casks/ioquake3.rb
@@ -12,6 +12,6 @@ cask 'ioquake3' do
   caveats <<-EOS.undent
     To complete the installation of #{token}, you will have to copy the file
     'pak0.pk3' from your Quake 3 Arena installation support directory into
-    ~/Applications/ioquake3/baseq3/.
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Still wrong. It’s not `/Applications`, but `#{appdir}`. Still, why have the caveat if we can just move it ourselves?

yurikoles:
@vitorgalvao It's not a full replacement, just engine. Please reopen.

vitorgalvao:
Ah, I see, you need an original install. Making the fix now.

yurikoles:
I have pushed updated & rebased version. Usage of /Applications is discouraged since `brew reinstall ioquake3` will remove app with resources.

yurikoles:
My previous version does not work. It hangs. So use original fix.

PR #28225, opened 2016-12-24
Add note to yujitach-menumeters

aleksandrs-ledovskis:
To avoid problems akin to https://github.com/caskroom/homebrew-cask/pull/27953#issuecomment-269032440 in the future


vitorgalvao:
Thank you, but an isolated incident does not warrant a comment. Especially since we know by experience that most people won’t read this.

aleksandrs-ledovskis:
@vitorgalvao Ok, your call

PR #28237, opened 2016-12-25
cask: Add Bricklink Stud.io 1.0.0_21

dholm:
There is no appcast for this cask as the checksum for the page containing the changelog (same as the release page) changes every time audit is executed.


vitorgalvao:
Thank you.

PR #28255, opened 2016-12-25
Update slack to 2.3.4

timsutton:

### DIFF
--- a/Casks/slack.rb
+++ b/Casks/slack.rb
@@ -1,6 +1,6 @@
 cask 'slack' do
-  version '2.3.3'
-  sha256 '3dd917164a0a4410b3f10f9c1ea2f1cdd899daa8c33e60fc7ba17095ed2c9b86'
+  version '2.3.4'
+  sha256 'f2ed659a8990330fbdd3564bc45cc2bba0ddaae784a1fdff83f68f6d0a8bc336'
 
   # downloads.slack-edge.com was verified as official when first introduced to the cask
   url "https://downloads.slack-edge.com/mac_releases/Slack-#{version}-macOS.zip"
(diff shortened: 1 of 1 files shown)


timsutton:
One other point: Slack hasn't used the Hockeyapp feed since they moved to the 2.3.x releases (a major refactor that moved to Electron). I'm not actually sure whether the app even contains references to the appcast anymore, as they've moved to Squirrel for their auto-updating.

timsutton:
I just checked and there's no references to Hockey, Sparkle or the appcast feed in the current Slack app. I just removed the appcast stanza from this cask.

miccal:
Perhaps `https://slack-files.com/T12KS1G65-F2R3TN3FS-89e821ddb7` can be used as an `appcast`?

Ping @vitorgalvao.

timsutton:
@miccal This seems to just be a public release notes listing, not a Sparkle appcast feed - unless I'm missing something..

miccal:
True, but we use this sort of thing in other Casks.

timsutton:
Oh, ok. I didn't realize that `appcast` is also used for such things when it's not actually a Sparkle feed.

miccal:
I am not 100 % sure about what can and cannot be used for an `appcast` though, hence why I pinged the expert!

adidalal:
@miccal Technically any HTML file will work, however that specific example is probably not a good appcast source (HTML styling/caches files will change the app cast pretty much every time its queried). Think more along the lines of a plain .txt for .xml file. See https://github.com/caskroom/homebrew-cask/issues/19103 for my general process.

Slack is probably OK without an appcast if we can't find one. It's online-only and auto_updates.

miccal:
Thanks for the information @adidalal.

timsutton:
Thanks @adidalal!

PR #28258, opened 2016-12-25
rstudio 1.0.136

sjackman:


vitorgalvao:

Do please fill the template next time.

sjackman:
The released version of `hub` 2.2.9, also the most recent available in Homebrew, does not provide this feature. You may wish to instruct people to install `hub` 2.3.0-pre8 using `brew upgrade --devel hub`, which does include this feature (implemented in 2.3.0-pre3).

Additionally, `hub` looks for the file `$PWD/PULL_REQUEST_TEMPLATE.md` and fails to find it when run from the `Casks` directory, which is probably common for this repo.

Thanks for merging, Vitor!

vitorgalvao:
The merger was @miccal.

For these simple changes, you might [consider `cask-repair` instead](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask). Its simpler and will take care not only of any tests, but of filling the template correctly.

PR #28321, opened 2016-12-28
Make auto_updates declarations consistent (3 casks changes)

alyssais:
There are 3 casks that have an `auto_updates` declaration that is inconsistent with all other casks, and the documentation (which states that `auto_updates true` is the only correct value).

This PR updates those three casks to have an `auto_updates true` declaration, consistent with every other auto-updating cask.

***

- [ ] `brew cask audit --download {{cask_file}}` is error-free. <ins>I'm on a very slow Internet connection right now, and will be leaving soon, so I'm creating this PR now and I'll update if this fails, since I can't imagine it would.</ins>

### DIFF
--- a/Casks/airtame.rb
+++ b/Casks/airtame.rb
@@ -6,7 +6,7 @@ cask 'airtame' do
   name 'Airtame'
   homepage 'https://airtame.com/'
 
-  auto_updates 'True'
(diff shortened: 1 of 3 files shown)


miccal:
Thank you.

PR #28386, opened 2016-12-30
Update stockbarjp.rb

miccal:
Update `homepage`.

### DIFF
--- a/Casks/stockbarjp.rb
+++ b/Casks/stockbarjp.rb
@@ -4,7 +4,7 @@ cask 'stockbarjp' do
 
   url 'http://midnightsuyama.org/download/StockBarJP.zip'
   name 'StockBarJP'
-  homepage 'http://midnightsuyama.org/blog/articles/stockbarjp/'
+  homepage 'http://midnightsuyama.org/'
 
   app 'StockBarJP.app'
 end


reitermarkus:
Not sure if we should keep this if you can't even find it on the homepage.

miccal:
I thought that myself but `brew cask fetch` is working, and the name of the app is at least on the homepage (although the link is not).

Happy to remove, though.

reitermarkus:
Seems like it was initially added by the app author. @midnightSuyama, is this application still supported?

midnightSuyama:
StockBarJP is still supported. However, I changed the article.
Please change to `homepage 'http://midnightsuyama.org/'`.

Thanks.

miccal:
@midnightSuyama that is what I have already done - the issue here is there is no link from there to actually download the `zip`?

midnightSuyama:
@miccal I got it. Please wait for renovation in progress.

vitorgalvao:
@midnightSuyama You’re linking the app to https://caskroom.github.io/, which makes no sense.

midnightSuyama:
Just improved. I changed the download link to `zip`.

vitorgalvao:
Thank you @midnightSuyama.

PR #28424, opened 2016-12-31
Update ezeep.rb to latest

### LINKED ISSUE #28401: Cask ezeep.rb needs modification
The Cask for [`ezeep.rb`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/ezeep.rb) currently fails as the `url` and `app` are no longer valid. The `url` is now `https://www.ezeep.com/wp-content/uploads/packages/osx/stable/ezeepInstaller.dmg` which downloads a `.dmg` containing an `.app` called `ezeep Installer.app`.

Running this `.app` calls a GUI requiring a user's password, which downloads _something_ and then installs an `.app` called `ezeep.app`. There is an `uninstall.sh` included with the `ezeep Installer.app` which is a little weird, in particular it requires `sudo` and to be called with a parameter `debug` or `production`, with no explanation given as to what these parameters actually mean. 

Even stranger, there is no similar uninstall script included with the `ezeep.app` that is actually installed.

I am at a loss as to what to do with this Cask, and need some advice please.

miccal:

Additionally, 

Closes https://github.com/caskroom/homebrew-cask/issues/28401

### DIFF
--- a/Casks/ezeep.rb
+++ b/Casks/ezeep.rb
@@ -2,9 +2,12 @@ cask 'ezeep' do
   version :latest
   sha256 :no_check
 
-  url 'https://packages.ezeep.com/osx/ezeep-latest.dmg'
+  url 'https://www.ezeep.com/wp-content/uploads/packages/osx/stable/ezeepInstaller.dmg'
   name 'ezeep'
   homepage 'https://www.ezeep.com/'
 
-  app 'ezeep.app'
+  installer manual: 'ezeep Installer.app'
+
+  uninstall script:  { executable: "#{staged_path}/ezeep Installer.app/Contents/Resources/uninstall.sh", args: ["production"], 
+                       sudo: true }
 end
--- a/Casks/ezeep.rb
+++ b/Casks/ezeep.rb
@@ -8,6 +8,8 @@ cask 'ezeep' do
 
   installer manual: 'ezeep Installer.app'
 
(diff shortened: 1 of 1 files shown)


miccal:
Thanks @vitorgalvao - the formatting of this is very inconsistent among Casks.

Casks/ezeep.rb
```diff
@@ -2,9 +2,15 @@
...
+  url 'https://www.ezeep.com/wp-content/uploads/packages/osx/stable/ezeepInstaller.dmg'
   name 'ezeep'
   homepage 'https://www.ezeep.com/'
 
-  app 'ezeep.app'
+  installer manual: 'ezeep Installer.app'
+
+  uninstall script: {
+                      executable: "#{staged_path}/ezeep Installer.app/Contents/Resources/uninstall.sh",
+                      args:       ['production'],
```
vitorgalvao: Does this need to be an array? Does it not work as a string?

miccal: I honestly do not know - see for example [zendserver.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/zendserver.rb) and [intel-haxm.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/intel-haxm.rb), which do the same thing.

miccal: @reitermarkus can you enlighten us on this?

reitermarkus: This is passed to `SystemCommand`, which only takes an array.

vitorgalvao (approved):

PR #22043, opened 2016-06-17
Make appdir method available inside *flight blocks

mwean:
### Changes to the core
- [x] Followed [hacking.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/hacking.md).

This makes the `appdir` helper method available inside pre-flight, post-flight, etc. blocks in addition to the regular DSL stanzas.

Closes #22024.

PR #28487, opened 2017-01-02
Updating squirrelsql-3.7.1 to use #{appdir}

dzeigler:
The squirrelsql cask has "/Applications" hardcoded in the paths. This PR changes the paths to use "#{appdir}" so that the install directory can be changed from the command line.

### DIFF
--- a/Casks/squirrelsql.rb
+++ b/Casks/squirrelsql.rb
@@ -76,7 +76,7 @@ EOS
   end
 
   uninstall_postflight do
-    system_command 'java', args: ['-jar', '#{appdir}/SQuirreLSQL.app/Uninstaller/uninstaller.jar', '-f', '-c']
+    system_command 'java', args: ['-jar', "#{appdir}/SQuirreLSQL.app/Uninstaller/uninstaller.jar", '-f', '-c']
   end
 
   zap delete: '~/.squirrel-sql'
--- a/Casks/squirrelsql.rb
+++ b/Casks/squirrelsql.rb
@@ -22,7 +22,7 @@ cask 'squirrelsql' do
         <com.izforge.izpack.panels.HelloPanel id="UNKNOWN (com.izforge.izpack.panels.HelloPanel)"/>
         <com.izforge.izpack.panels.HTMLInfoPanel id="UNKNOWN (com.izforge.izpack.panels.HTMLInfoPanel)"/>
         <com.izforge.izpack.panels.TargetPanel id="UNKNOWN (com.izforge.izpack.panels.TargetPanel)">
-        <installpath>/Applications/SQuirreLSQL.app</installpath>
+        <installpath>#{appdir}/SQuirreLSQL.app</installpath>
         </com.izforge.izpack.panels.TargetPanel>
         <com.izforge.izpack.panels.PacksPanel id="UNKNOWN (com.izforge.izpack.panels.PacksPanel)">
         <pack index="0" name="Base" selected="true"/>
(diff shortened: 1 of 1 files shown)


Casks/squirrelsql.rb
```diff
@@ -76,7 +76,7 @@
   end
 
   uninstall_postflight do
-    system_command 'java', args: ['-jar', '/Applications/SQuirreLSQL.app/Uninstaller/uninstaller.jar', '-f', '-c']
+    system_command 'java', args: ['-jar', '#{appdir}/SQuirreLSQL.app/Uninstaller/uninstaller.jar', '-f', '-c']
```
reitermarkus: You need to use double quotes here.

miccal: @reitermarkus does `#{appdir}` actually work for this Cask? I though it only worked if the `artifact` stanza was an `app`?

dzeigler: Thanks I added the double quotes. The #{appdir} reference on line 25 seems to work as I am able to install this Cask to the directory specified by appdir. However, when I try to uninstall it, which I had not tried before, it fails with 
> ==> Standard Error of failed command:
> Error: Unable to access jarfile #{appdir}/SQuirreLSQL.app/Uninstaller/uninstaller.jar

I'm new to Cask in general so I'm not sure if this is a limitation of this particular Cask or another problem with my change.

vitorgalvao: @dzeigler Do you still get that error? Because that seems like it is caused by the single quotes.

miccal: Just tested the original version of this Cask (that is, without the changes in this PR) via `brew cask install` then `brew cask zap`, and it uninstalled successfully:
```
==> Implied "brew cask uninstall squirrelsql"
==> Dispatching zap stanza
==> Running zap process for squirrelsql; your password may be necessary
==> Removing files: ["~/.squirrel-sql"]
==> Removing all staged versions of Cask 'squirrelsql'
```
Is the error caused by `#{appdir}` not being respected?

dzeigler: @vitorgalvao Yes, I get get the error with the double quote version, unfortunately.

dzeigler: @miccal yes, I suspect `#{appdir}` is not being respected in `uninstall_postflight`, and that's causing the error.

vitorgalvao:
Moving the conversation here, outside the collapsed comment. From https://github.com/caskroom/homebrew-cask/pull/28487#discussion_r94966854:

> I suspect `#{appdir}` is not being respected in `uninstall_postflight`

It should be, since https://github.com/caskroom/homebrew-cask/pull/22043. @mwean @reitermarkus Are you able to shed some light?

reitermarkus:
> It should be, since #22043. @mwean @reitermarkus Are you able to shed some light?

It is.

@dzeigler, how exactly are you specifying `--appdir`? If you are using it manually, you also need to add it to the `uninstall` command.

This is working fine for me:

```
brew cask install squirrelsql --appdir=~/Applications
brew cask uninstall squirrelsql --appdir=~/Applications
```

miccal:
Also, I just modified my local copy of `squirrelsql.rb` to match that of this PR, and it installed and uninstalled successfully, so it seems there is nothing wrong with adding `#{appdir}`.

dzeigler:
You're right. I started over from scratch and install and uninstall both work now. Sorry for the confusion.

miccal (approved):

PR #28527, opened 2017-01-03
Update bassjump to 3.0.1

colindunn:

### DIFF
--- a/Casks/bassjump.rb
+++ b/Casks/bassjump.rb
@@ -1,34 +1,21 @@
 cask 'bassjump' do
-  if MacOS.version <= :mountain_lion
-    if MacOS.version == :snow_leopard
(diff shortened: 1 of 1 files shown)


colindunn:
Hey @vitorgalvao what's the general policy here?
I intentionally backfilled the old versions because snow leopard and up is supported, but would like to gain some general guidance for future updates.

vitorgalvao:
We officially support the current macOS version and the one before that. That does not mean we cannot support versions older than that, but when they make the casks more complex (as was the case here), it’s not worth it.

You’ll notice I’ve left all cases that do not complicate the cask further than what is needed to work on the current version.

colindunn:
Thanks for that. Perfect.

PR #28531, opened 2017-01-04
Added nix 1.11.5

zimbatm:
Hi, this is my first cask package. Let me know if I missed anything.

### DIFF
--- a/Casks/nix.rb
+++ b/Casks/nix.rb
@@ -9,12 +9,12 @@ cask 'nix' do
   depends_on macos: '>= :yosemite'
 
   installer script: "nix-#{version}-x86_64-darwin/install",
-            sudo: false
+            sudo:   false
 
   uninstall delete: [
                       '/nix',
                       '~/.nix-channels',
                       '~/.nix-defexpr',
-                      '~/.nix-profile'.
+                      '~/.nix-profile',
                     ]
 end
--- a/Casks/nix.rb
+++ b/Casks/nix.rb
@@ -8,7 +8,8 @@ cask 'nix' do
 
   depends_on macos: '>= :yosemite'
 
-  installer script: "nix-#{version}-x86_64-darwin/install", sudo: false
+  installer script: "nix-#{version}-x86_64-darwin/install",
+            sudo: false
 
   uninstall delete: [
                       '/nix',
--- a/Casks/nix.rb
+++ b/Casks/nix.rb
@@ -17,4 +17,21 @@ cask 'nix' do
                       '~/.nix-defexpr',
                       '~/.nix-profile',
                     ]
+
+  uninstall_postflight do
+    ['~/.bash_profile', '~/.bash_login', '~/.profile'].each do |profile_path|
+      profile_path = File.expand_path(profile_path)
+
+      if File.exist?(profile_path)
+        system_command '/usr/bin/sed',
+                       args: [
+                               '-E',
+                               '-i', '.nix_uninstall_bak',
+                               '-e', '/# added by Nix installer/d',
+                                profile_path
+                             ],
+                       sudo: false
+      end
+    end
+  end
 end
--- a/Casks/nix.rb
+++ b/Casks/nix.rb
@@ -11,24 +11,23 @@ cask 'nix' do
   installer script: "nix-#{version}-x86_64-darwin/install",
             sudo:   false
 
-  uninstall delete: '/nix'
-
   uninstall_postflight do
     ['~/.bash_profile', '~/.bash_login', '~/.profile'].each do |profile_path|
       profile_path = File.expand_path(profile_path)
 
-      if File.exist?(profile_path)
-        system_command '/usr/bin/sed',
-                       args: [
-                               '-E',
-                               '-i', '.nix_uninstall_bak',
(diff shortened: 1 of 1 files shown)


Casks/nix.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  url "https://nixos.org/releases/nix/nix-#{version}/nix-#{version}-x86_64-darwin.tar.bz2"
+  name 'nix'
+  homepage 'https://nixos.org/nix/'
+
+  depends_on macos: '>= :yosemite'
+
+  installer script: "nix-#{version}-x86_64-darwin/install", sudo: false
+
+  uninstall delete: [
```
vitorgalvao: Did this pass `style` as it was? I find that weird.

vitorgalvao (changes requested):

vitorgalvao:
This seems to be a CLI-only open-source tool. Please submit it first to [Homebrew](https://github.com/Homebrew/homebrew-core) and only if they don’t accept it, submit here (linking back to the pull request where it was refused).

> The app is both open-source and CLI-only (i.e. it only uses the `binary` artifact). In that case, and [in the spirit of deduplication](https://github.com/caskroom/homebrew-cask/issues/15603), submit it first to [Homebrew](https://github.com/Homebrew/homebrew). If it is rejected there, you may then try again in Homebrew-Cask (link us to the issue on Homebrew so we can see their reasoning for rejection).

Casks/nix.rb
zimbatm: Yes it did

zimbatm:
Thanks a lot for reviewing and fixing my code.

nix is a package manager itself, just like macports which is also in this repo. It's quite unlikely that I will be able to build it from source and propose a recipe to the main homebrew repo as it's build process depends on nix itself. Right now it's bootstrapped from another system that has nix installed already.

miccal:
> nix is a package manager itself, just like macports

Yes but the Cask for [MacPorts](https://github.com/caskroom/homebrew-cask/blob/master/Casks/macports.rb) installs a `.pkg` as its `artifact`, which is totally different to how Nix is installed.

zimbatm:
The Macports CLI could also probably be compiled using a Homebrew formula. After the installation I think both have the same set of constraints that don't make them good candidates for Homebrew.

They both install under a non-standard prefix. Macports uses `/opt/local` and Nix uses `/nix`. I believe that Homebrew wants to keep everything under `/usr/local`.

They both self-upgrade and want to manage software themselves. Having Homebrew hold one part of the software under `/usr/local` would probably cause issues. http://docs.brew.sh/Acceptable-Formulae.html also says "We don’t like tools that upgrade themselves"

In that sense I see Homebrew Cask most like a bootstrap in both cases. Basically if users already have Homebrew installed but need extra dependencies from macports or nix then it makes it easier to extend the package set. They could also run some `curl | sh` script but having this managed is nice to have and also helps by establishing the trust chain.

vitorgalvao:
> They could also run some `curl | sh` script but having this managed is nice to have and also helps **by establishing the trust chain**.

[emphasis added]

**I really hope not**, with the utmost sincerity. For one, we’re not actually verifying the script, just running it, so there’s absolutely no reason users should trust this more simply because it comes from us.

In addition, we’re adamant users do not trust us blindly: see https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment and https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md.

The rest of the argument makes sense, though.

Casks/nix.rb
```diff
@@ -0,0 +1,20 @@
...
+  url "https://nixos.org/releases/nix/nix-#{version}/nix-#{version}-x86_64-darwin.tar.bz2"
+  name 'nix'
+  homepage 'https://nixos.org/nix/'
+
+  depends_on macos: '>= :yosemite'
+
+  installer script: "nix-#{version}-x86_64-darwin/install",
+            sudo:   false
+
+  uninstall delete: [
```
vitorgalvao: Is this really enough to completely remove it?

vitorgalvao: We don’t have upgrade functionality.

vitorgalvao (approved):

Casks/nix.rb
zimbatm: Okay.

I added the cleanup of it in 75658532d570df6375aa6a464c0b2629d7474702

zimbatm:
Thanks @vitorgalvao. It's the first time I'm cooperating on a PR with the reviewer pushing code to it and it's quite nice actually, it feels much more collaborative like that.

Do you want me to squash and rebase the PR?

miccal:
There are a few `style` errors:
```
>>> brew cask style Casks/nix.rb
== Casks/nix.rb ==
C: 14:  3: uninstall stanza out of order
C: 16:  3: uninstall_postflight stanza out of order
C: 20:  7: Use next to skip iteration.
C: 26: 33: Align the elements of an array literal if they span more than one line.
```

zimbatm:
All fixed now

adidalal (approved):

PR #28580, opened 2017-01-04
appcast.md: improve documentation consistency

neutric:
To be more consistent with rest of documentation

- add .rb to cask names and 
- update example casks where commit linked to 
  -  either still shows license stanza that is no longer being used,
  - or is missing a "# URL_SECTION was verified as official when first introduced to the cask" comment,
  - or both.


miccal:
Thank you.

PR #28643, opened 2017-01-05
Update elektriktrickquicklook to 1.1.0

colindunn:

### DIFF
--- a/Casks/elektriktrick.rb
+++ b/Casks/elektriktrick.rb
@@ -1,4 +1,4 @@
-cask 'elektriktrickquicklook' do
+cask 'elektriktrick' do
   version '1.1.0'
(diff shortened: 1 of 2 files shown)


miccal:
This Cask should be renamed to just `elektriktrick.rb` now.

colindunn:
I did think about that, but according to the homepage it's still called quicklook despite the file name change. No problem with the change.

vitorgalvao:
> but according to the homepage it's still called quicklook

The token (name of the file and first line) conforms to the rules of the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).

colindunn:
That's a fair point. Will keep it in mind @vitorgalvao

PR #28657, opened 2017-01-06
Update real-vnc to 6.0.1

colindunn:

### DIFF
--- a/Casks/real-vnc.rb
+++ b/Casks/real-vnc.rb
@@ -1,6 +1,6 @@
 cask 'real-vnc' do
-  version '5.3.2'
-  sha256 'ed075ad1bec0d3d6ae9c446fbdec211798853d79f14c7a98520819f22fe991d3'
+  version '6.0.1'
+  sha256 '404d3b1ffe4b576736a1802d8f44270e26af390f1b6c379bfe95b0803de3e5d2'
 
   url "https://www.realvnc.com/download/file/vnc.files/VNC-#{version}-MacOSX-x86_64.pkg"
   name 'Real VNC'
@@ -8,5 +8,10 @@ cask 'real-vnc' do
 
   pkg "VNC-#{version}-MacOSX-x86_64.pkg"
 
-  uninstall script: '/Applications/RealVNC/Advanced.localized/Uninstall VNC Viewer.app/Contents/Resources/uninstaller.sh'
+  uninstall script: {
+                      executable: '/bin/sh',
+                      args:       ['-c', '/Applications/RealVNC/Uninstall\ VNC\ Server.app/Contents/Resources/uninstaller.sh; /Applications/RealVNC/Uninstall\ VNC\ Viewer.app/Contents/Resources/uninstaller.sh'],
+                    }
+
+  caveats "Uninstall scripts throw errors '/Applications/RealVNC/Advanced.localized: No such file or directory' but work correctly"
 end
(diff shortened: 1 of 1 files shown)


colindunn:
Not sure of the process for multiple uninstall scripts (this cask installs two separate applications) so have used /bin/sh to accomplish that. Couldn't see a better way in the docs, so let me know if there is a better method.

miccal:
See the Cask for [ransomwhere](https://github.com/caskroom/homebrew-cask/blob/master/Casks/ransomwhere.rb) for an example.

vitorgalvao:
Try `script:` twice. That might work.

Also, judging by the `caveat` you might need [`must_suceed: false`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#summary-of-keys) on one of them.

colindunn:
That was the first thing I tried, but unfortunately it doesn't work it just uses the last entry specified. The scripts still succeed correctly despite the error message, so I didn't add the must_succeed.

vitorgalvao:
Is it possible to add a `pkgutil:` as well?

colindunn:
The uninstallers take care of that. I haven't used app and pkg removal because the uninstallers should take care of the application through version changes, as well as remove registered services etc.

Package installed:

```
> pkgutil --pkgs|grep vnc
com.realvnc.vncserver.pkg
com.realvnc.vncviewer.pkg
```

Removal and result:

```
> brew cask uninstall real-vnc
==> Running uninstall process for real-vnc; your password may be necessary
==> Running uninstall script /bin/sh
==> ls: /Applications/RealVNC/Advanced.localized: No such file or directory
==> /Applications/RealVNC/Uninstall VNC Server.app/Contents/Resources/uninstallCommon.sh: line 27: /Applications/RealVNC/Uninstall VNC Server.app/Contents/Resources/removeFwExc.sh: No such file or directory
==> No receipt for 'com.realvnc.VNCServer.pkg' found at '/'.
==> Forgot package 'com.realvnc.vncserver.pkg' on '/'.
==> ls: /Applications/RealVNC/Advanced.localized: No such file or directory
==> No receipt for 'com.realvnc.VNCViewer.pkg' found at '/'.
==> Forgot package 'com.realvnc.vncviewer.pkg' on '/'.
 > pkgutil --pkgs|grep vnc
```

vitorgalvao:
I have no doubt the installers do it, but we usually prefer not to trust the installer scripts *only*, and also add out own uninstallation to make sure everything is done.

colindunn:
Put this one on hold and I'll update it to do removal via cask options

colindunn:
Changes made. Seems to work OK as far as I can tell and remove everything.

adidalal (approved):

vitorgalvao:
Oh, I meant adding `pkgutil:` **in addition to**. The point is to make really sure.

miccal:
Sorry @vitorgalvao, I saw the approval and just merged after a quick check.

vitorgalvao:
@miccal Not a problem.

@colindunn Would it be alright for you to please add it back? The more I think about, the more I’m unsure `script:` is the right place. We are indeed calling a script from the app, but using `bin/sh` to do it seems weird. Perhaps an `uninstall_preflight` would be more appropriate. And perhaps even redirect the error message, so we can get rid of the caveat.

miccal:
Can you have two `executable:`'s in an `uninstall script:`, like two of [these](https://github.com/caskroom/homebrew-cask/blob/master/Casks/ezeep.rb#L12#L14)?

colindunn:
Sorry missed these comments in my email. I will have another go at this later today when I get the chance, no worries.

PR #28676, opened 2017-01-06
Update marked.rb for zap

nuomi1:
Update marked.rb for zap

### DIFF
--- a/Casks/marked.rb
+++ b/Casks/marked.rb
@@ -6,7 +6,7 @@ cask 'marked' do
   appcast 'https://updates.marked2app.com/marked.xml',
           checkpoint: 'caf003f39ca2a87922e2d1f6afe17df55bd6bb6f36dcfd16b52e257cf629f5e0'
   name 'Marked'
-  homepage 'https://marked2app.com/'
+  homepage 'http://marked2app.com/'
 
   auto_updates true
 
--- a/Casks/marked.rb
+++ b/Casks/marked.rb
@@ -1,12 +1,12 @@
 cask 'marked' do
-  version '2.5.9'
+  version '2.5.9925'
   sha256 '81bcbfbe51c1ec987f1d20695a2187fcd848bc46c33c054821fe120c7d82a5c0'
 
-  url 'http://marked2app.com/download/Marked.zip'
+  url "https://updates.marked2app.com/Marked#{version}.zip"
   appcast 'https://updates.marked2app.com/marked.xml',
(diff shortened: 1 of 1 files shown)


Casks/marked.rb
```diff
@@ -1,8 +1,8 @@
 cask 'marked' do
-  version '2.5.9'
+  version '2.5.9925'
   sha256 '81bcbfbe51c1ec987f1d20695a2187fcd848bc46c33c054821fe120c7d82a5c0'
 
-  url 'http://marked2app.com/download/Marked.zip'
+  url "https://updates.marked2app.com/Marked#{version}.zip"
```
miccal: This `url` does not come from the `homepage`, it comes from the `appcast` - we prefer the link from the `homepage`, which is `http://marked2app.com/download/Marked.dmg`.

Casks/marked.rb
```diff
@@ -1,8 +1,8 @@
 cask 'marked' do
-  version '2.5.9'
+  version '2.5.9925'
```
miccal: This `version` does not come from the `app`, it comes from the `appcast` - we prefer the version from the `app`, which is `2.5.9`.

Casks/marked.rb
vitorgalvao: Slight correction, there: it’s fine for URLs to come from appcasts, especially if they are from the same domain and are versioned (vs the homepage’s unversioned). In that latter case, they are even preferred.

The comment above URL used to be different: it said the URL was official *as per the homepage*. Part of the reason it was changed was to keep it accurate with cases such as this, when a URL from outside the homepage is superior.

In cases such as this, where the URL comes from the same domain, it doesn’t really matter, so we should go with the superior (versioned) version.

miccal: Understood, thanks for the explanation - I though that it was preferred to verify the `url` from the `homepage`.

vitorgalvao: That’s understandable. Some of these smaller rules I’m not sure are written, even, as they are acquired as cases happen and are really specific (we’d need a whole `if`/`else` table to explain all of them).

Regarding versions, we generally don’t really give a hoot about the app’s version, we care about the version in the URL. Reason being nobody cares about the app’s version either, for the most part, so we do what’s the most convenient for interpolation.

Occasionally when the string for interpolation is incredibly obtuse, we may add both that the correct version (separated by `,`), but other than that, use whatever works best.

vitorgalvao (approved):

PR #28683, opened 2017-01-06
Add goad v1.2.0.ci6

zeph:

### DIFF
--- a/Casks/goad.rb
+++ b/Casks/goad.rb
@@ -1,0 +1,14 @@
+cask 'goad' do
+  version '1.2.0.ci6'
+  sha256 'bdd79b7c70a343c551e2d1903c46bda59b3d7426794b4ca9575a76eca2ef36ca'
+
+  # github.com/zeph/goad was verified as official when first introduced to the cask
+  url "https://github.com/zeph/goad/releases/download/v#{version}/goad-osx-x86-64.zip"
+  name 'Goad'
+  # github.com/zeph/goad is a fork with several improvements of goad.io
+  homepage 'https://goad.io'
+
+  container type: :naked
+
+  binary 'goad-osx-x86-64', target: 'goad'
+end
--- a/Casks/goad.rb
+++ b/Casks/goad.rb
@@ -8,7 +8,7 @@ cask 'goad' do
   # github.com/zeph/goad is a fork with several improvements of goad.io
   homepage 'https://goad.io'
 
-  container type: :naked
+  container type: :zip
 
-  binary 'goad-osx-x86-64', target: 'goad'
+  binary 'goad', target: 'goad'
 end


miccal:
So this is your own fork of an official tool? We cannot accept this.

zeph:
@miccal the tool was a proof of concept of some guys earlier January last year. They are barely maintaining it... I'm now actively developing it on my fork. And will continue to do so. 

Now, you can argue this is not the official fork. As they have prior art. Shall I change the name and detach myself from their original fork? Really "forking" the project, as pre-git times for any opensource project? I've been providing patches, but they are too slow to apply them. If they want they can pull from mine if they want to keep their fork active. 

Suggestions are accepted. Rejecting as I'm not the original creator of the first base of code, is against the concept of opensource to incentivate active development and sharing of improvements.

vitorgalvao:
I hadn’t seen this before, but https://github.com/caskroom/homebrew-cask/pull/28698#discussion_r95016500 is a clarification of @miccal’s point. We can accept forks, but not in the way you submitted it. You’re essentially passing your fork as official, which is inaccurate.

They have over a thousand stars, and you have a single one. I believe when you say yours is maintained and theirs is, but barely anyone knows your fork, and we’re not going to impose your fork disguised as the official app. *That* is the main issue, here.

vitorgalvao:
> Shall I change the name and detach myself from their original fork?

If you do that you can submit it with the name of your new project, *and you must not use their homepage*.

However, we will still not accept it, for two reasons:

+ Since this is CLI-only and open-source, it needs to be submitted to Homebrew first, as already explained.
+ Your fork is unknown, and hence will add overhead for little to no value (see https://github.com/caskroom/homebrew-cask/issues/28220). Homebrew’s rule are even stricter, here: the CI will refuse it because it has less than 50 stars and you are self-submitted, which they frown upon.

To be clear, no one is doubting your good intentions, but we all felt the need to start drawing the line somewhere, and barely-known projects are where it stands. If your fork indeed picks up, then it will be a wanted addition.

zeph:
@vitorgalvao ok, is there a way to propose a brew cask... via what... a Tap? I'm noob on the internals... so, I really would like to empower my colleagues to brew cask install goad (or similar command) ... is there a way to be explicit that this is NOT the official prj neither pretends to be it?

zeph:
> And with another comment claiming it is official.
as your "style" check was NOT allowing any other comment in order to pass the check... that is why 2 lines below I stated the real reference...  ;)

vitorgalvao:
> as your "style" check was NOT allowing any other comment in order to pass the check... that is why 2 lines below I stated the real reference... ;)

The correct way to deal with that was to use an accurate homepage, not add an inaccurate comment. The style check is there for a reason, and that reason is not that you circumvent it when it’s convenient.

To make a tap: [docs](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

zeph:
ok... I'll go the Tap way... thanks for the pointer @vitorgalvao

Issue #26345, opened 2016-11-02
Feature request: OR or XOR switches for depends_on stanzas


BenjaminHCCarr:
### Justification

There are casks that can work with multiple casks or brewed formulas which provide the same functionality and often the same named binary. This came up a very long time ago with #2001; and I am even going to use `sshfs` as the example.

Homebrew allows for arrays and switched , but the only array I can find in the docs:
https://github.com/caskroom/homebrew-cask/tree/master/doc
is for `depends_on macos: ['10.9', '10.10’]`

I can't seem to find any other arrays or switches.

Really I think the depends_on should be some sort of pseudocode like:
`depends_on: [formula: 'homebrew/science/r', cask: 'r-gui']`

or
```
depends on{
   formula: 'homebrew/science/r'
   OR
   cask: 'r-gui'
}
```

I just updated #26340, so r is on my mind.

is this functionality implemented but not documented; eg, am I missing something?

### Example use case

The example given at [depends_on doc](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/depends_on.md) is `sshfs` which could depend (until very recently) on `osxfuse` or `Fuse4x` and their are other places where someone may choose a binary/cask rather than a formula. I know this has been discussed.

### Similar issues:

`--flags on formula` #8784
`depends_on :cask` follow-ups #8492
using `depends_on` for intra-Cask dependencies and metapackages #2001
Feature request: XOR with depends_on #19369

### Possible solution condition statement or array:

The real crux is very similar to (though not NEARLY as complicated as):
Stanzas in conditional statements #14333

In the example of the package I bumped today `r-studio`  #26340; `/applications/R.app` is not an option in `homebrew/science/r` as there is no `--with-bundle` flag, so many choose to install both`homebrew/science/r-gui` as well; note, these do NOT conflict; though `r-gui` makes `brew doctor` very very upset. However both provide a "functional" `r` binary. With the way I see stanza `depends_on:` is written now; I had to use `depends_on formula: 'homebrew/science/r'`

Is there an undocumented way to use conditionals with depends_on:? I tried some hacks with `==>` and arrays `[..,..]` but thought merging to get everyone to stable (no less 1.0 finally) was worth more than me blindly throwing similar conditionals at the formula.

So my proposed solution would be something like we use for `depends_on macos:` or a ruby OR

Change `depends_on` to be something like this pseudocode:

[...]

PR #28717, opened 2017-01-07
rstudio.rb: replace depends_on with caveat

vitorgalvao:

---

Not the first time a user complains because this depends on the formula and they want the GUI. Most recently: https://github.com/caskroom/homebrew-cask/issues/28714.

### DIFF
--- a/Casks/rstudio.rb
+++ b/Casks/rstudio.rb
@@ -13,7 +13,7 @@ cask 'rstudio' do
 
   caveats <<-EOS.undent
     #{token} depends on R.
-    There are different ways to satisfy that dependency and we don’t want to impose one, so it is up to you to satify it.
+    There are different ways to satisfy that dependency and we don’t want to impose one, so it is up to you to satisfy it.
     We suggest you do so by running one of:
 
       brew install homebrew/science/r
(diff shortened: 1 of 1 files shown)


miccal (approved):

Casks/rstudio.rb
```diff
@@ -7,9 +7,16 @@
...
 
-  depends_on formula: 'homebrew/science/r'
-
   app 'RStudio.app'
 
   zap delete: '~/.rstudio-desktop'
+
+  caveats <<-EOS.undent
+    #{token} depends on R.
+    There are different ways to satisfy that dependency and we don’t want to impose one, so it is up to you to satify it.
```
Numbermaniac: "up to you to **satify** it"

miccal: Pardon?

BenjaminHCCarr:
Maybe this is a good place to bring up #26345, my [OR|XOR] thought.

As caveats seem to be growing, would it be possible to put an OR/XOR in a caveat form; we have drop throughs for other things and also handle pkg conflicts, it seems that allowing either a casked or bottle build should be possible. My suggestion was basically to change the form of `depends_on` to allow arrays.

miccal:
@vitorgalvao I would suggest this is good-to-go?

@BenjaminHCCarr let's keep the discussion regarding your [OR|XOR] thought in the original issue.

PR #28719, opened 2017-01-07
Add latest kerio-control-vpn-client.

sepehr:

### DIFF
--- a/Casks/kerio-control-vpn-client.rb
+++ b/Casks/kerio-control-vpn-client.rb
@@ -1,0 +1,15 @@
+cask 'kerio-control-vpn-client' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://download.kerio.com/dwn/kerio-control-vpnclient-mac.dmg'
+  name 'Kerio Control VPN Client'
+  homepage 'http://kerio.com/'
+
+  pkg 'Kerio Control VPN Client Installer.pkg'
+
+  uninstall kext:      'com.kerio.kext.kvnet',
+            launchctl: 'com.kerio.kvpncsvc',
+            pkgutil:   'com.kerio.control.kvpnclient.pkg',
+            quit:      ['com.kerio.VPN.status', 'com.kerio.VPN.agent']
+end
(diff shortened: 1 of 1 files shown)


miccal:
When you download the `dmg` there are two files, `Kerio Control VPN Client Installer.pkg` and `Kerio Control VPN Client Uninstaller.app`. The uninstall `app` contains an uninstall script at `Kerio Control VPN Client Uninstaller.app/Contents/Resources/uninstall` which seems to do a more thorough job of uninstalling this.

It might be better to add an `uninstall script:` to this Cask. See an example [here](https://github.com/caskroom/homebrew-cask/blob/master/Casks/ezeep.rb#L11#L15).

sepehr:
I went down that way at first. But it seems that the script only works through the UI. 

When you run that `uninstall` script, it halts with a non-zero exit status (without any output). I even tried to simulate the work of the `uninstall` script and execute those action scripts (residing in `action/`) separately which failed as well. 

So I took the `uninstall` script and simulated all its jobs using cask's uninstall stanza. The only thing that differs is that the uninstall script closes the *System Preferences* app. I didn't know how to do that in using uninstall stanza.

What do you think?

miccal:
No problem if you had already tried and it calls a UI then no problem. The only thing I can see that the uninstall script does "extra" is removing the following files:
```
rm -rf /usr/local/kerio/vpnclient/!(dbSSL|config|logs)
rm -rf /Library/PreferencePanes/KVpnClient.prefPane
rm -rf /Library/LaunchDaemons/com.kerio.kvpncsvc.plist
rm -rf /Library/Extensions/kvnet.kext
```

sepehr:
Yes, you are right and I saw those lines as well. But I though that's covered by [`pkgutil:   'com.kerio.control.kvpnclient.pkg'`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#uninstall-key-pkgutil). Isn't that?

miccal:
Ah yes, I believe you are correct.

miccal (approved):

vitorgalvao (approved):

PR #28721, opened 2017-01-07
Add uninstall launchctl to pritunl

zachhuff386:
Add uninstall launchctl to pritunl


miccal:

PR #28722, opened 2017-01-07
Updated tableau reader to 10.1.3

johandebeurs:
previous cask was pulling v9 of the reader


miccal:
Thanks you for your contribution.

PR #28739, opened 2017-01-07
Update real-vnc uninstall

colindunn:
Suggested uninstall script execution, as discussed on #28657

I am unable to get two scripts to run via uninstall, so settled on this method. Preflight should mean the scripts have done their work and the uninstall stanzas should be cleanup only if it scripts fail to do so.

---

### DIFF
--- a/Casks/real-vnc.rb
+++ b/Casks/real-vnc.rb
@@ -9,16 +9,8 @@ cask 'real-vnc' do
   pkg "VNC-#{version}-MacOSX-x86_64.pkg"
 
   uninstall_preflight do
-    IO.write "#{staged_path}/uninstall.sh", <<-EOS.undent
-      #!/bin/sh
-      "/Applications/RealVNC/Uninstall VNC Server.app/Contents/Resources/uninstaller.sh" &> /dev/null
-      "/Applications/RealVNC/Uninstall VNC Viewer.app/Contents/Resources/uninstaller.sh" &> /dev/null
-    EOS
-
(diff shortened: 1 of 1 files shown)


Casks/real-vnc.rb
```diff
@@ -8,6 +8,19 @@
 
   pkg "VNC-#{version}-MacOSX-x86_64.pkg"
 
+  uninstall_preflight do
```
vitorgalvao: Why all the trouble? Why not simply running the scripts as `system_command` directly? You can have as much of those as necessary.

colindunn: I'm trying to prevent the errors due to missing paths in the uninstall scripts using pipes to /dev/null to remove the caveat as per the discussion on the previous thread. I can't seem to find a way to do that via system_command alone without using /bin/sh and this seems cleaner than /bin/sh in each system_command.

vitorgalvao: I think something like

```ruby
system_command '/Applications/RealVNC/Uninstall VNC Server.app/Contents/Resources/uninstaller.sh', stderr: '/dev/null', sudo: true
system_command '/Applications/RealVNC/Uninstall VNC Viewer.app/Contents/Resources/uninstaller.sh', stderr: '/dev/null', sudo: true
```

should work.

colindunn: Ah, got it. I went and looked at the code for system_command in the brew repo and got the right options. Didn't realise that was available. Thanks @vitorgalvao

vitorgalvao:
Thank you for all the work!

PR #28805, opened 2017-01-09
Deleted stale Codebox cask

dcondrey:
Removed Codebox formula.  Codebox.io has been down for quite a while and attempts to install report `curl: (22) The requested URL returned error: 503 Service Unavailable`

App seems to only be available from the AppStore now.


miccal:
Verified [here](https://github.com/CodeboxIDE/codebox/issues/518).

PR #28822, opened 2017-01-10
Create binary.md documentation

miccal:
Closes https://github.com/caskroom/homebrew-cask/issues/25423.

### DIFF
--- a/doc/cask_language_reference/stanzas/binary.md
+++ b/doc/cask_language_reference/stanzas/binary.md
@@ -1,4 +1,4 @@
-# binary`
+# binary
 
 In the simple case of a string argument to `binary`, the source file is linked into the `$(brew --prefix)/bin` directory (typically `/usr/local/bin`) on installation. For example (from [operadriver.rb](https://github.com/caskroom/homebrew-cask/blob/60531a2812005dd5f17dc92f3ce7419af3c5d019/Casks/operadriver.rb#L11)):
 
--- a/doc/cask_language_reference/stanzas/binary.md
+++ b/doc/cask_language_reference/stanzas/binary.md
@@ -37,3 +37,11 @@ or just [cleaning up the name](https://github.com/caskroom/homebrew-cask/blob/14
 ```ruby
 binary 'minishift-darwin-amd64', target: 'minishift'
 ```
+
+## target: May Contain an Absolute Path
+
+If `target:` has a leading slash, it is interpreted as an absolute path. The containing directory for the absolute path will be created if it does not already exist. This is usseful for installing manual pages that are often included with binaries, for example (from [manopen.rb](https://github.com/caskroom/homebrew-cask/blob/312ae841f1f1b2ec07f4d88b7dfdd7fbdf8d4f94/Casks/manopen.rb#L12)):
+
+```ruby
+artifact 'openman.1', target: '/usr/local/share/man/man1/openman.1'
+```
--- a/doc/cask_language_reference/stanzas/binary.md
+++ b/doc/cask_language_reference/stanzas/binary.md
@@ -20,7 +20,7 @@ from a source file such as:
 
 ## Renaming the Target
 
-You can rename the target which appears in your `/usr/local/bin` directory by adding a `target:` key to `binary`, for example to be to be consistent with other command-line tools, like [changing case](https://github.com/caskroom/homebrew-cask/blob/070a3bdeadf339892268d79aded9108ba7960f32/Casks/praat.rb#L12):
+You can rename the target which appears in your `/usr/local/bin` directory by adding a `target:` key to `binary`, for example to be to be consistent with other command-line tools like [changing case](https://github.com/caskroom/homebrew-cask/blob/070a3bdeadf339892268d79aded9108ba7960f32/Casks/praat.rb#L12):
 
 ```ruby
 binary "#{appdir}/Praat.app/Contents/MacOS/Praat", target: 'praat'
--- a/doc/cask_language_reference/stanzas/binary.md
+++ b/doc/cask_language_reference/stanzas/binary.md
@@ -40,8 +40,9 @@ binary 'minishift-darwin-amd64', target: 'minishift'
 
 ## target: May Contain an Absolute Path
 
-If `target:` has a leading slash, it is interpreted as an absolute path. The containing directory for the absolute path will be created if it does not already exist. This is usseful for installing manual pages that are often included with binaries, for example (from [manopen.rb](https://github.com/caskroom/homebrew-cask/blob/312ae841f1f1b2ec07f4d88b7dfdd7fbdf8d4f94/Casks/manopen.rb#L12)):
+If `target:` has a leading slash, it is interpreted as an absolute path. The containing directory for the absolute path will be created if it does not already exist. This is usseful for installing manual pages that are often included with binaries, for example (from [manopen.rb](https://github.com/caskroom/homebrew-cask/blob/312ae841f1f1b2ec07f4d88b7dfdd7fbdf8d4f94/Casks/manopen.rb#L11#L12)):
 
 ```ruby
+binary 'openman'
 artifact 'openman.1', target: '/usr/local/share/man/man1/openman.1'
 ```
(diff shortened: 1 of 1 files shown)


doc/cask_language_reference/stanzas/binary.md
```diff
@@ -0,0 +1,48 @@
+# binary
+
+In the simple case of a string argument to `binary`, the source file is linked into the `/usr/local/bin` folder on installation. For example (from [operadriver.rb](https://github.com/caskroom/homebrew-cask/blob/60531a2812005dd5f17dc92f3ce7419af3c5d019/Casks/operadriver.rb#L11)):
```
vitorgalvao: I think we usually call them “directories” not “folders”. Or maybe we’re inconsistent on that. I prefer the former.

doc/cask_language_reference/stanzas/binary.md
```diff
@@ -0,0 +1,48 @@
...
+
+```ruby
+binary 'operadriver'
+```
+
+creates a symlink to:
+
+```bash
+/usr/local/bin/operadriver
```
vitorgalvao: This isn’t strictly true. In practice we use `$(brew --prefix)/bin`. Most of the time that will translate to `/user/local/bin`, but not always. See https://github.com/caskroom/homebrew-cask/issues/21372.

This also needs to be fixed in https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/all_stanzas.md.

doc/cask_language_reference/stanzas/binary.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+from a source file such as:
```
vitorgalvao: I don’t think this section is necessary. It’s not that useful to know where it’s linked from, that’s a more advanced detail.

doc/cask_language_reference/stanzas/binary.md
```diff
@@ -0,0 +1,48 @@
...
+
+```bash
+/usr/local/Caskroom/operadriver/0.2.2/operadriver
+```
+
+## Renaming the Target
+
+You can rename the target which appears in your `/usr/local/bin` directory by adding a `target:` key to `binary`, for example to be to be consistent with other command-line tools like [changing case](https://github.com/caskroom/homebrew-cask/blob/070a3bdeadf339892268d79aded9108ba7960f32/Casks/praat.rb#L12):
```
vitorgalvao: We can name it “binaries directory” or something, here, instead of `/usr/local/bin`, since as previously noted, it might not be that location.

vitorgalvao: The examples could be taken directly from https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases, for consistency.

doc/cask_language_reference/stanzas/binary.md
```diff
@@ -0,0 +1,48 @@
...
+binary 'FileBot.app/Contents/MacOS/filebot.sh', target: 'filebot'
+```
+
+or just [cleaning up the name](https://github.com/caskroom/homebrew-cask/blob/146917cbcc679648de6b0bccff4e9b43fce0e6c8/Casks/minishift.rb#L13):
+
+```ruby
+binary 'minishift-darwin-amd64', target: 'minishift'
+```
+
+## target: May Contain an Absolute Path
```
vitorgalvao: While useful, this part is referring to `artifact`, not `binary`. We don’t want to encourage absolute paths in `binary, target:`. The section should be removed or made clearer that it is referring to `artifact` as a companion to `binary`.

vitorgalvao (changes requested):

doc/cask_language_reference/stanzas/binary.md
miccal: Fixed.

doc/cask_language_reference/stanzas/binary.md
miccal: I followed the same language as used in [app.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md) which says the same thing.

Happy to remove it, though.

doc/cask_language_reference/stanzas/binary.md
miccal: Fixed.

doc/cask_language_reference/stanzas/binary.md
miccal: > We can name it “binaries directory” or something, here, instead of /usr/local/bin, since as previously noted, it might not be that location.

Fixed.

> The examples could be taken directly from https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases, for consistency.

That is where I got them from.

doc/cask_language_reference/stanzas/binary.md
miccal: Again, I followed the same language as used in [app.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md) which says the same thing.

Happy to remove it, though.

doc/cask_language_reference/stanzas/binary.md
vitorgalvao: I think we can remove it from `app.md` as well. As long as it’s consistent, I’ll leave it to what you find is clearer.

miccal: I thing that it is better in the binary because it actually _is_ a symlink, and users would like to know _where_ the symlink points to.

doc/cask_language_reference/stanzas/binary.md
vitorgalvao: Lets do one better. I think that I was the one to originally write these, and that I put `target` in `app.md` for lack of a better place. We could either:

+ Split the documentation of `target:` to a new document and link there from both `app.md` and `binary.md`
+ Remove all `target:` documentation from `binary.md` and link to the section in `app.md`.

I’m partial to the second one (seems to be as clear, and requires less files.

miccal: Sounds good to me.

miccal: I'm offline for a while - please fell free to edit the file.

vitorgalvao: I’m instead [removing mentions of `binary`](https://github.com/caskroom/homebrew-cask/pull/28840).

In this document, we can say that for `target:`, the `app.md` document should be followed, but that the select cases don’t apply as rigidly and it’s fine to take some extra liberties with `target:` to be consistent with other command-line tools, like [changing case](https://github.com/caskroom/homebrew-cask/blob/6e4eb6ba58ca0d9e6d42a1d78856cc8a35cf5fce/Casks/diffmerge.rb#L11), [removing an extension](https://github.com/caskroom/homebrew-cask/blob/312ae841f1f1b2ec07f4d88b7dfdd7fbdf8d4f94/Casks/filebot.rb#L12), or [cleaning up the name](https://github.com/caskroom/homebrew-cask/blob/146917cbcc679648de6b0bccff4e9b43fce0e6c8/Casks/minishift.rb#L13).

This way it’ll be more consistent with what we do with the other repos (we tell users to check the rules in this one, and point out the differences there.

vitorgalvao (approved):

doc/cask_language_reference/stanzas/binary.md
vitorgalvao: > I think that it is better in the binary because it actually _is_ a symlink, and users would like to know _where_ the symlink points to.

> It can be removed from `app.md` though, I think.

Agreed.

doc/cask_language_reference/stanzas/binary.md
miccal: I Agree.

miccal:
Thanks for all the help @vitorgalvao.

vitorgalvao:
@miccal Thank you for starting it in the first place.

PR #28823, opened 2017-01-10
Migrate rar 5.4.0 from homebrew-binary

miccal:

- [X] Checked that the cask was not already refused in [closed issues].

Ping @adidalal RE https://github.com/caskroom/homebrew-cask/issues/23938 for help about what else needs to be done for these migrations.

### DIFF
--- a/Casks/rar.rb
+++ b/Casks/rar.rb
@@ -12,7 +12,7 @@ cask 'rar' do
   artifact 'rar/rarfiles.lst', target: '/usr/local/etc/rarfiles.lst'
 
   caveats <<-EOS.undent
-    Instructions on using rar and unrar are aviable in
+    Instructions on using rar and unrar are available in
 
       #{staged_path}/rar/rar.txt
   EOS
--- a/Casks/rar.rb
+++ b/Casks/rar.rb
@@ -8,4 +8,13 @@ cask 'rar' do
 
   binary 'rar/rar'
   binary 'rar/unrar'
+  artifact 'rar/default.sfx', target: '/usr/local/lib/default.sfx'
+  artifact 'rar/rarfiles.lst', target: '/usr/local/etc/rarfiles.lst'
+
+  caveats <<-EOS.undent
+    Instructions on using rar and unrar are aviable in
+
+      #{staged_path}/rar/rar.txt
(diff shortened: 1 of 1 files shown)


adidalal:
Add it to the migrations json

miccal:
@adidalal is there a simple way to do that, or to I have to go through the whole fork/git remote add process? You wouldn't have that set up already would you?

miccal:
@adidalal Done.

Another question: I notice that the formula in homebrew/binary includes the following:
```
  def install
    bin.install "rar", "unrar"
    lib.install "default.sfx"
    etc.install "rarfiles.lst"
    doc.install "acknow.txt", "order.htm", "rar.txt", "whatsnew.txt"
    man1.install resource("man")
  end
```

The files included in the downloaded archive are:

screen shot 2017-01-10 at 22 36 53

I don't see a file to create a `man` page - am I missing something? Should I worry about any of these files via an `artifact ... target:`?

adidalal:
It's been a while but you may also need to copy over that .sfx (using artifact)

Not sure about the man pages, people can always do rar help though

miccal:
> It's been a while but you may also need to copy over that .sfx (using artifact)

To where?

> Not sure about the man pages, people can always do rar help though.

No, `rar --help` is not valid, according to the `rar.txt` file:
```
 Switches (used in conjunction with a command):

-?     Display help on commands and switches. The same as when none
         or an illegal command line option is entered.
```

MikeMcQuaid:
Sorry, I'm not sure about what to do about manages. I'd suggest we could add it without if need be.

miccal:
@MikeMcQuaid no problem, but could you explain where the commands `lib.install` and `etc.install` move the files to?

MikeMcQuaid:
As far as Cask is concerned: `/usr/local/lib` and `/usr/local/etc`

miccal:
Thanks @MikeMcQuaid.

Ping @adidalal for final review.

adidalal:
Typo in caveats but otherwise looks :+1

miccal:
Fixed, thanks @adidalal.

adidalal (approved):

PR #28847, opened 2017-01-10
Add TimeCamp v1.3.78

ghost:

### DIFF
--- a/Casks/timecamp.rb
+++ b/Casks/timecamp.rb
@@ -1,6 +1,6 @@
 cask 'timecamp' do
-  version '1.3.78'
-  sha256 '69049d5ca4b12166fa91724d01762310bc9d09326e4ad3f91822a3ac77b2fb46'
+  version :latest
+  sha256 :no_check
 
   url 'https://www.timecamp.com/downloadsoft/TimeCamp.dmg'
   name 'TimeCamp'
@@ -8,7 +8,5 @@ cask 'timecamp' do
 
   app 'TimeCamp.app'
 
-  zap rmdir: [
-               '~/Library/Application Support/TimeCamp',
-             ]
+  zap rmdir: '~/Library/Application Support/TimeCamp'
 end
--- a/Casks/timecamp.rb
+++ b/Casks/timecamp.rb
@@ -2,7 +2,6 @@ cask 'timecamp' do
   version '1.3.78'
   sha256 '69049d5ca4b12166fa91724d01762310bc9d09326e4ad3f91822a3ac77b2fb46'
 
-  # downloads page is https://www.timecamp.com/helper/download
   url 'https://www.timecamp.com/downloadsoft/TimeCamp.dmg'
   name 'TimeCamp'
   homepage 'https://www.timecamp.com/'
@@ -10,6 +9,6 @@ cask 'timecamp' do
   app 'TimeCamp.app'
 
   zap rmdir: [
-               '~/Library/Application Support/TimeCamp'
+               '~/Library/Application Support/TimeCamp',
              ]
 end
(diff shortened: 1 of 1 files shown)


Casks/timecamp.rb
```diff
@@ -0,0 +1,12 @@
+cask 'timecamp' do
+  version :latest
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, it needs `version :latest` and `sha256 :no_check`.

vitorgalvao (approved):

Casks/timecamp.rb
cboynton-cweil: Ahh. That makes sense. Thank you for teaching me. This was my first cask definition. I was looking at and copying from other examples. I remember seeing the same issue occurring elsewhere. Someone should go through and identify that problem everywhere. They should be easy to identify with a well crafted regular expression.

vitorgalvao: > Someone should go through and identify that problem everywhere. They should be easy to identify with a well crafted regular expression.

You’re assuming the problem exists. It doesn’t.

miccal:

Casks/timecamp.rb
cboynton-cweil: No assumption was made. I did irrefutably see the issue and many places. Where would you like me to send the evidence?

vitorgalvao: I remembered something later, but ended up not commenting. I’m betting those cases you saw have an `appcast` stanza.

Those casks can have a `version` even if `url` is unversioned, since we can leverage the `appcast` to know when the app was updated.

cboynton-cweil: The casks I've examined with the issue I mentioned do not contain appcast entries.

I'm not sure about all of the procedures necessary. I will paste elsewhere or open issues if you'd like. Please let me know.

Here is my evidence:

axure-rp
brow
citrix-receiver
clix
coyim
datadog-agent
dataurlmaker
deathtodsstore
displaperture
docear
fanny
find-empty-folders
fonttong
gridwars
idisplay
katana
kismac
liquifile
manageengine-mibbrowser
massreplaceit
mouse-locator
netgeargenie
overdrive-media-console
packages
pi-filler
recovery-disk-assistant
riffworkst4
smart-scroll
stattransfer
subclassed-mnemosyne
tamarin-prover
teensy
treemaker
tvmobili
ultimate
watts
younity

cboynton-cweil: I am not 100% sure on some of my evidence as the version is not part of the url, but it is used elsewhere. I may be mistaken on the following suggestions:

clix
coyim
datadog-agent
dataurlmaker
docear
fanny
fonttong
idisplay
liquifile
manageengine-mibbrowser
mouse-locator
netgeargenie
overdrive-media-console
recovery-disk-assistant
riffworkst4
smart-scroll
stattransfer
tamarin-prover
treemaker
tvmobili
ultimate
watts
younity.rb

miccal: All fixed, thanks @cboynton-cweil.

vitorgalvao: Ah, yes, there was one other case I forgot to mention. First, context. The reason casks without a versioning scheme need to use `:latest` and `:no_check` is that without it, the cask would break when there was an update upstream (since it would still download fine but fail the `sha256` check).

But look at the majority of casks you posted. They all need the version elsewhere. Specifically, they need it in the artifact that is the crux os the installation (`app`, `pkg`, …), which means that without the right version, those installations would break anyway. The version is very much needed. Look closely, as it’s (almost?) always interpolated (`#{version}`).

There was at least one that did not in fact adhere to this rule, but one or two does not constitute a problem. These ones already present themselves for fixing anyway, when they break. **If** they break actually, as some are so seldom updated (or abandoned) they rarely or never break anyway, meaning `:latest` isn’t needed.

vitorgalvao: > but one or two

Or thirteen, more accurately. I stand corrected in their number, thought I maintain they really aren’t an issue dor the aforementioned reasons.

Thank you for pushing through and presenting the offensive casks, @cboynton-cweil!

PR #28892, opened 2017-01-11
Update and rename rabbitmq.rb to rabbitmq-app.rb

vitorgalvao:
Part of https://github.com/caskroom/homebrew-cask/issues/15603.

From [the website](https://jpadilla.github.io/rabbitmqapp/), it seems indeed to be mostly a wrapper. If merged, we must also submit a fix to [the repo’s instructions](https://github.com/jpadilla/rabbitmqapp#installing-with-homebrew-cask).


miccal:
https://github.com/jpadilla/rabbitmqapp/issues/2.

PR #28966, opened 2017-01-12
appcast.md: add tr

vitorgalvao:
Addresses https://github.com/caskroom/homebrew-cask/issues/24454. Until https://github.com/caskroom/homebrew-cask/issues/24811 is done, this is still needed.

On hold until the new cask-repair release. The command will also need to be updated in https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/lib/hbc/audit.rb.

After we do so, we’ll need to recalculate every `appcast` for every cask.

### DIFF
--- a/doc/cask_language_reference/stanzas/appcast.md
+++ b/doc/cask_language_reference/stanzas/appcast.md
@@ -6,7 +6,7 @@ The value of the `appcast` stanza is a string, holding the URL for an appcast wh
 
 | key           | value       |
 | ------------- | ----------- |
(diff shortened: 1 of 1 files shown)


reitermarkus:
> After we do so, we’ll need to recalculate every appcast for every cask.

Well, not every cask. If the appcast is outdated already, we shouldn't recalculate it.

vitorgalvao:
> Well, not every cask. If the appcast is outdated already, we shouldn't recalculate it.

I meant to remove that but forgot. I since think we shouldn’t go through the trouble of recalculating everything on purpose. We can do it as casks are updated. I intend to build the new scripts to update casks based on appcasts sometime soon.

Related, could you give me some assistance with https://github.com/Homebrew/brew/pull/1876?

PR #29017, opened 2017-01-14
Add VESTA v3.4.0

mkhorton:
Adding VESTA, an app for visualizing crystallographic data. Unfortunately no https source is available on the official download page: http://jp-minerals.org/vesta/en/download.html

### DIFF
--- a/Casks/vesta.rb
+++ b/Casks/vesta.rb
@@ -1,9 +1,8 @@
 cask 'vesta' do
-  version :latest
-  sha256 :no_check
+  version '3.4.0'
+  sha256 'adc8e468ad53a79cccfa5aa3c0db113c9b6b656c032aa28fc8d6630b0714d133'
 
-  # geocities.jp/kmo_mma was verified as official when first introduced to the cask
-  url 'http://www.geocities.jp/kmo_mma/crystal/download/VESTA.dmg'
+  url "http://jp-minerals.org/vesta/archives/#{version}/VESTA.dmg"
   name 'VESTA'
   homepage 'http://jp-minerals.org/vesta/en/'
 
--- a/Casks/vesta.rb
+++ b/Casks/vesta.rb
@@ -1,0 +1,11 @@
+cask 'vesta' do
+  version :latest
+  sha256 :no_check
+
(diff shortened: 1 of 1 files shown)


Casks/vesta.rb
```diff
@@ -0,0 +1,11 @@
+cask 'vesta' do
+  version :latest
+  sha256 :no_check
+
+  # geocities.jp/kmo_mma was verified as official when first introduced to the cask
+  url 'http://www.geocities.jp/kmo_mma/crystal/download/VESTA.dmg'
+  name 'VESTA'
+  homepage 'http://jp-minerals.org/vesta/en/'
+
+  app 'VESTA/VESTA.app'
```
miccal: There is another app in the `VESTA` folder, namely `PowderPlot.app` - what is that for, is it required by the `VESTA.app`?

mkhorton: I believe that's a separate utility (it has a separate page on the author's website) but is included in the VESTA package for convenience, because VESTA can generate input files for PowerPlot.app. I have never had any problems running VESTA without PowderPlot installed.

miccal: OK, thank you.

miccal (approved):

Casks/vesta.rb
```diff
@@ -0,0 +1,11 @@
+cask 'vesta' do
+  version :latest
+  sha256 :no_check
```
neutric: Versioned releases, including the latest one, are available at http://jp-minerals.org/vesta/archives/. @miccal Shouldn't those be given preference?

miccal: Thanks @neutric yes it should. 

@mkhorton please change the `version` to `3.4.0` and the `url` to `"http://jp-minerals.org/vesta/archives/#{version}/VESTA.dmg"`.

You will also need to set the [sha256](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md).

mkhorton: Great, thanks for this. I'd missed the archive page. It has the benefit that it's hosted on the same domain too.

PR #29174, opened 2017-01-19
Fix install of kindlegen

halostatue:
When migrating Kindlegen from homebrew-binary to homebrew-cask (#22194), the binary artifact was incorrectly named to be in a subdirectory that only exists when the downloaded zipfile is opened through the Finder archiver integration. When using `ditto` or `unzip`, `kindlegen` is extracted without a base directory.

Kindlegen would not install without this change on my machine.


miccal:
Thank you.

miccal:
Closes https://github.com/caskroom/homebrew-cask/issues/29181

PR #29180, opened 2017-01-19
Update perforce.rb, remove p4.rb

miccal:
Adds https://github.com/Homebrew/homebrew-binary/blob/master/perforce-server.rb to the one Cask.

Ping @adidalal for review.


adidalal:
Are these the same perforce product? I know there's a p4 cask also

miccal:
@adidalal [p4v](https://github.com/caskroom/homebrew-cask/blob/master/Casks/p4v.rb) and [p4merge](https://github.com/caskroom/homebrew-cask/blob/master/Casks/p4merge.rb) are different, but [p4](https://github.com/caskroom/homebrew-cask/blob/master/Casks/p4.rb) is the same as the original perforce Cask.

Suggestions?

miccal:
Contents of `helix-versioning-engine.tgz`:

screen shot 2017-01-19 at 20 25 00

larsxschneider:
Is it possible to disable the hash comparison for this cask? I know it is a bad idea as you can never know what binary you are dealing with... but Perforce changes these files all the time. Example from the past: https://github.com/Homebrew/legacy-homebrew/issues/21799

miccal:
@larsxschneider done, thanks for the information.

Ping @adidalal and @MikeMcQuaid for final review.

larsxschneider:
I think the name `helix-versioning-engine` could be confusion to people. It looks like as if the `helix-versioning-engine.rb` contains all products:
p4 (the client app) and p4d (the "daemon", server app).

p4d was available here: https://github.com/Homebrew/homebrew-binary/blob/master/perforce-server.rb

miccal:
> I think the name helix-versioning-engine could be confusion to people.

Ping @vitorgalvao RE the token reference.

> p4d was available here: https://github.com/Homebrew/homebrew-binary/blob/master/perforce-server.rb

The whole point of this is it migrate all formula from hombrew-binary - see https://github.com/Homebrew/homebrew-binary/pull/364.

larsxschneider:
I would prefer 'perforce' as name but I am fine with 'helix-versioning-engine', too. Thanks a lot for working on this!

larsxschneider:
The git core TravisCI builds are blocked because of the Perforce ... I am looking forward to your merged patch 😄 
https://travis-ci.org/git/git/builds

miccal:
> I would prefer 'perforce' as name but I am fine with 'helix-versioning-engine', too. Thanks a lot for working on this!

No problem.

> The git core TravisCI builds are blocked because of the Perforce ... I am looking forward to your merged patch 

What do you mean exactly?

vitorgalvao:
I’m fine with either name. Since this is binary-only, the token reference doesn’t apply, but it does seem like users might expect `p4` or `perforce` instead of `helix-versioning-engine`.

miccal:
I think `perforce` might be better.

adidalal:
Let's also add in a json entry for the redirect. Otherwise looks good

miccal:
> Let's also add in a json entry for the redirect. Otherwise looks good

Yes, I will once https://github.com/Homebrew/homebrew-binary/pull/364 is merged.

adidalal:
Just to clarify, I meant p4 -> perforce (both Casks, but p4 is being renamed? now, so it makes sense to have that redirect there also)

miccal:
> Just to clarify, I meant p4 -> perforce (both Casks, but p4 is being renamed? now, so it makes sense to have that redirect there also)

Yes, I will fix it all, no problem.

miccal:
Closed by mistake.

adidalal (approved):

adidalal:
Not sure why Travis is lagging but LGTM from looking at the cask

miccal:
Thanks @adidalal - there are only three remaining in homebrew-binary, namely [eventstore.rb](https://github.com/Homebrew/homebrew-binary/blob/master/eventstore.rb), [mpress.rb](https://github.com/Homebrew/homebrew-binary/blob/master/mpress.rb) and [subgit.rb](https://github.com/Homebrew/homebrew-binary/blob/master/subgit.rb), which seem quite a bit more complicated than those I have already done.

I may need some help with those ones.

adidalal:
Just double check with @MikeMcQuaid about usage stats for those before putting in significant effort. All in all though, fantastic work!

miccal:
Thanks @adidalal - I am pretty sure @MikeMcQuaid has already removed a fair few bases on the analytics statistics, so those that are left are used.

miccal:
Well, the homebrew-binary formula for `mpress` is not working - when I run `brew test mpress` the output is:
```
-bash-4.4.5- /Users/raesy [> brew test mpress
Testing homebrew/binary/mpress
==> Using the sandbox
==> /usr/local/Cellar/mpress/1.21_1/bin/mpress
Error: homebrew/binary/mpress: failed
<1> expected but was
<22>.
-bash-4.4.5- /Users/raesy [>
```
and running `press --version` outputs nothing, so I cannot test a Cask install via a comparison with a homebred-binary install.

Thoughts @MikeMcQuaid?

miccal:
There is also a problem with the homebred-binary formula for `subgit` - I modified the formula to use the latest version (3.2.3 instead of 3.2.0) and update the download url, but there is no `test` with this formula, and running any command, such as `subgit --version` returns `Unable to find any JVMs matching version "(null)".` and request that I install Java (which I do not really want to do), so I cannot test a Cask install via a comparison with a homebrew-binary install on this either.

miccal:
And finally, I have no idea how to create a Cask for `eventstore`, it is beyond my skills, sorry.

MikeMcQuaid:
You've done a great job @miccal, well done.

miccal:
Update: `subgit` was actually quite straight-forward once I played around with it.

In regards to `press`, since the 64bit version is "stripped" out of the binary in homebrew/binary and it only installs the 32bit version, I believe that we cannot add it to Cask as 64bit is now required?

@vitorgalvao is that correct?

larsxschneider:
Everything works perfectly! Thank you!

PR #29205, opened 2017-01-19
Delete trailing spaces

mathbunnyru:
Only deletes trailing spaces in some markdown files.


vitorgalvao:
doc/cask_language_reference/stanzas/version.md and doc/reporting_bugs/error_unknown_command_cask.md are correct. As for the others, please revert them, that space is intentional (it’s for the user to start writing).

Also, those others are templates, used in other files, so those would have to be generated again as well.

mathbunnyru:
@vitorgalvao now only two files are changed

vitorgalvao:
Thank you.

PR #29209, opened 2017-01-19
Update deadbeef to 0.7.2

yurikoles:


miccal:
@vitorgalvao should we just change this to `version :latest` and `sha :no_check` and remove the `appcast` - it seems they keep changing, and the `version` is not actually used in the Cask?

yurikoles:
@miccal this is exactly what was in my first commit for this cask :)

vitorgalvao:
It’s not the appcast or the version are wrong, though. They aren’t changing due to an error, they are changing because the app keeps being updated. [This isn’t a stable version, yet](https://github.com/Alexey-Yakovenko/deadbeef#download-development-builds). When it becomes one, it’ll presumably be updated less frequently.

PR #29227, opened 2017-01-20
Migrate eventstore.rb from homebrew/binary

miccal:
This is a work-in-progress and I need some help - I know that what I currently have is probably not correct.

The original formula from homebrew/binary is reproduced below:
```
class Eventstore < Formula
  desc "functional database with complex event processing"
  homepage "http://geteventstore.com"
  url "http://download.geteventstore.com/binaries/EventStore-OSS-MacOSX-v3.7.0.tar.gz"
  sha256 "e5902899d45aca6414fae01fff152ddc2df1495ad29d36ea3bcffbe33025e396"

  bottle :unneeded

  depends_on :macos => :mavericks

  def install
    prefix.install Dir["*"]

    (bin/"eventstore").write <<-EOS.undent
      #!/bin/sh
      cd "#{prefix}"
      exec "#{prefix}/run-node.sh" "$@"
    EOS

    (bin/"eventstore-testclient").write <<-EOS.undent
      #!/bin/sh
      exec "#{prefix}/testclient" "$@"
    EOS
  end

  test do
    assert_match  /#{version}/, shell_output("#{bin}/eventstore --version")
  end
end
```

Ping @adidalal and @vitorgalvao for some help with this one please.

Files inside the `tar.gz` for reference:

screen shot 2017-01-20 at 20 09 26

### DIFF
--- a/Casks/eventstore.rb
+++ b/Casks/eventstore.rb
@@ -10,19 +10,19 @@ cask 'eventstore' do
   binary 'eventstore-testclient'
 
   # shim script (https://github.com/caskroom/homebrew-cask/issues/18809)
-  eventstore_shimscript = "#{staged_path}/eventstore"
-  testclient_shimscript = "#{staged_path}/eventstore-testclient"
+  eventstore_shimscript = "#{staged_path}/EventStore-OSS-MacOSX-v#{version}/eventstore"
+  testclient_shimscript = "#{staged_path}/EventStore-OSS-MacOSX-v#{version}/eventstore-testclient"
 
   preflight do
     IO.write eventstore_shimscript, <<-EOS.undent
       #!/bin/sh
-      cd "#{staged_path}"
-      exec "#{staged_path}/run-node.sh" "$@"
+      cd "#{staged_path}/EventStore-OSS-MacOSX-v#{version}"
+      exec "#{staged_path}/EventStore-OSS-MacOSX-v#{version}/run-node.sh" "$@"
     EOS
 
     IO.write testclient_shimscript, <<-EOS.undent
       #!/bin/sh
-      exec "#{staged_path}/testclient" "$@"
+      exec "#{staged_path}/EventStore-OSS-MacOSX-v#{version}/testclient" "$@"
     EOS
 
     set_permissions eventstore_shimscript, '+x'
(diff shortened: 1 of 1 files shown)


miccal:
Thank you @vitorgalvao, I am not familiar with the `shimscript` stanza and did not realise it was this simple.

I have tested it and it is working.

Question, though: in the Cask [`steamcmd`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/steamcmd.rb) (for example) after the shimscipt is created the command to make it executable is `FileUtils.chmod '+x', shimscript`, why did you change this one to `set_permissions eventstore_shimscript, '+x'` instead?

vitorgalvao:
`shimscript` isn’t exactly a stanza (not yet), it’s pure ruby done as a workaround until we support the feature.

`set_permissions` is our own (HBC) command, and it’s very likely it did not exist when those first shim scripts were made. As such, the old casks could get the update.

miccal:
> set_permissions is our own (HBC) command, and it’s very likely it did not exist when those first shim scripts were made. As such, the old casks could get the update.

Thanks for the explanation - would you like all existing Casks that use a `shimscript` to change from `FileUtils.chmod` to `set_permissions`?

vitorgalvao:
> would you like all existing Casks that use a `shimscript` to change from `FileUtils.chmod` to `set_permissions`?

Ideally yes, for consistency, but it’s not a big deal if they’re not. Though the change might be easily accomplished with a script.

miccal:
> Ideally yes, for consistency, but it’s not a big deal if they’re not. Though the change might be easily accomplished with a script.

[Done!](https://github.com/caskroom/homebrew-cask/pull/29240)

petterbj:
@vitorgalvao I have a question , `set_permissions` leads to unnecessary sudo usage, which might have some security risk, especially within installation vlc a media player.  [vlc.rb](https://github.com/caskroom/homebrew-cask/blob/master/Casks/vlc.rb)

vitorgalvao:
@petterbj https://github.com/Homebrew/brew/pull/1930.

petterbj:
@vitorgalvao Thanks for response, I was new to here, even new to developing a super huge project. Please pardon inconvenience, if any, on how to report something I found.

vitorgalvao:
@petterbj No inconvenience at all, that issue was opened *because you made that comment*. I thought we didn’t use `sudo` when `chmod`ing, and your post made me realise we do.

petterbj:
@vitorgalvao It seems I worried too much.😀 Anyway, I'm pleased to offer some help.

PR #29293, opened 2017-01-22
gtkwave 3.3.79 (new cask)

MikeMcQuaid:
Migrating this from homebrew/homebrew-gui.

  - Can't work out what's wrong with SourceForge format.

### DIFF
--- a/Casks/gtkwave.rb
+++ b/Casks/gtkwave.rb
@@ -2,7 +2,9 @@ cask 'gtkwave' do
   version '3.3.79'
   sha256 'dcdc3e9d356c00c356183cee07269734e8763098bb4958c5a355b1583cecd7e5'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Should pass `audit` now.

We’re so strict with sourceforge URLS because they gave us a lot of trouble at the start. It was frequent for people to submit specific URLs from specific mirrors (instead of the more generic download URLs that pick the best one).

By enforcing a certain pattern in audit (which we identified at the time), we make sure that doesn’t happen. So your previous link was actually fine (generic instead of specific mirror) but caught in the rule.

The reason we don’t extend the rule is it isn’t really needed. As you can see a small change (removing `/project`) makes it pass and work just the same.

vitorgalvao:
Passed. Thank you!

MikeMcQuaid:
Great, thanks @vitorgalvao. We've had similar problems with SourceForge URLs in the past so I feel your pain 😁

PR #29316, opened 2017-01-23
Inaccurate info about Ruby syntax

marnen:
The language reference seems to imply that `cask 'foo' do ... end` isn't "strict" Ruby, whatever that means. But that's not true: in standard Ruby syntax, no comma is required between simple arguments and a `do` block. So I'm suggesting removing the sentence that claims that this is not strict Ruby.


reitermarkus:
Thanks, no idea what this is supposed to mean either.

PR #29322, opened 2017-01-23
Add TouchBarServer 1.6

szhu:

~~Note: The installed app is called "Touch Bar Server" but the website clearly calls it "Touch Bar Demo App". My assumption is that "Touch Bar Server" is the original name, and that it will change on next release, which is why I named it according to the website here.~~

### DIFF
--- a/Casks/touchbarserver.rb
+++ b/Casks/touchbarserver.rb
@@ -1,0 +1,12 @@
+cask 'touchbarserver' do
+  version '1.6'
+  sha256 'c3adffbaed14d12feaf5300995f78e63fbbc2a99733f76549f5a4b071ce86a82'
+
+  url "https://github.com/bikkelbroeders/TouchBarDemoApp/releases/download/v#{version}/TouchBarServer.zip"
+  appcast 'https://github.com/bikkelbroeders/TouchBarDemoApp/releases.atom',
+          checkpoint: 'cdd8a73aac96791b9ec9e56cc6d214006f3a4d2d8c9ed1c8fbb3aa23fa4bf2bd'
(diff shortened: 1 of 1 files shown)


Casks/touchbarserver.rb
```diff
@@ -0,0 +1,10 @@
+cask 'touchbarserver' do
+  version '1.6'
+  sha256 'c3adffbaed14d12feaf5300995f78e63fbbc2a99733f76549f5a4b071ce86a82'
+
+  url "https://github.com/bikkelbroeders/TouchBarDemoApp/releases/download/v#{version}/TouchBarServer.zip"
```
vitorgalvao: Please add an [`appcast`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

vitorgalvao (changes requested):

vitorgalvao:
> Note: The installed app is called "Touch Bar Server" but the website clearly calls it "Touch Bar Demo App". My assumption is that "Touch Bar Server" is the original name, and that it will change on next release, which is why I named it according to the website here.

If it changes name, then we change the cask, no point in assuming a future name. Even if the name were in accordance with the repo, the token reference is pretty explicit that `app` should not be part of the name.

szhu:
It seemed like you were implying the name should be TouchBarServer, so I changed it to that. This should be ready for review again. Thanks!

Casks/touchbarserver.rb
szhu: Fixed.

vitorgalvao (approved):

PR #29335, opened 2017-01-24
Update rejected_casks.md

miccal:
Closes https://github.com/caskroom/homebrew-cask/issues/28220.

I think the third point in https://github.com/caskroom/homebrew-cask/issues/28220, namely:

> What maintainers feel is too much overhead for the value

is already covered by:

> The cask is so difficult to maintain, it constantly breaks. Happened only once so far, with Audacity, because it uses fosshub as a distribution site.


vitorgalvao (approved):

miccal:
Thanks @vitorgalvao.

Pinging @caskroom/maintainers for any other comments or suggestions before merging.

PR #29356, opened 2017-01-24
adding_a_cask: editions back to main repo and nomenclature additions

vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/issues/24418.

Bring back editions to main repo and separate drivers into new one. We should move the casks first and then merge this.

Also fixes the omission of fonts and eIDs.

### DIFF
--- a/doc/development/adding_a_cask.md
+++ b/doc/development/adding_a_cask.md
@@ -193,6 +193,7 @@ We maintain separate Taps for different types of binaries. Our nomenclature is:
 + **Vendorless**: A binary distributed without an official website, like a forum posting.
 + **Walled**: When the download URL is both behind a login/registration form and from a host that differs from the homepage.
 + **Font**: Data file containing a set of glyphs, characters, or symbols, that changes typed text.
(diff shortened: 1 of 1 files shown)


adidalal:
Re: moving Casks - I assume we'll keep /versions with just a tap_migrations.json?

vitorgalvao:
@adidalal caskroom/versions will continue to be alive and well. It’ll just not have the things that confused some users by being there (pro versions, community editions). It’ll still have all the betas, nightlies, developer editions, old versions, and all that.

miccal:
@adidalal and @vitorgalvao I have just gone through caskroom/versions and renamed a few Casks for consistency, which should make identifying which ones to migrate a little easier.

miccal:
I also updated several Casks in caskroom/versions yesterday, replacing `system` with `system_command` so those should already be nice and consistent with the main repo.

miccal:
@adidalal raises a good point, though - will we need a `tap_migrations.json` file?

vitorgalvao:
> will we need a `tap_migrations.json` file?

Sounds about right, yes.

PR #29364, opened 2017-01-24
Delete woodhouse.rb

vitorgalvao:
[Work in progress](https://github.com/phinze/woodhouse#woodhouse-is-an-incomplete-man) with no commits in five years and just 10 stars. Found it inadvertently and was going to update the appcast (to github releases), but looks like a good candidate for removal.

Lets check first, though. @phinze is this still useful?


phinze:
Oddly enough I just received word from one user that they still use this thing daily! But one user is not enough to warrant inclusion methinks. 😉 

+1 for Removal

PR #29368, opened 2017-01-25
Update opencpn homepage and appcast

colindunn:
- homepage gave an error that certificates were for another domain, likely the hosting provider. Changing back to http.
- git based appcast does not have download links

---

### DIFF
--- a/Casks/opencpn.rb
+++ b/Casks/opencpn.rb
@@ -4,8 +4,10 @@ cask 'opencpn' do
 
   # opencpn.navnux.org was verified as official when first introduced to the cask
   url "http://opencpn.navnux.org/#{version}/OpenCPN_#{version}.dmg"
+  appcast 'https://github.com/OpenCPN/OpenCPN/releases.atom',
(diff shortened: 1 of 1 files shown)


Casks/opencpn.rb
```diff
@@ -4,8 +4,10 @@
 
   # opencpn.navnux.org was verified as official when first introduced to the cask
   url "http://opencpn.navnux.org/#{version}/OpenCPN_#{version}.dmg"
+  appcast 'https://github.com/OpenCPN/OpenCPN/releases.atom',
```
vitorgalvao: Do they link to this repo from anywhere on the site?

vitorgalvao (approved):

Casks/opencpn.rb
colindunn: Not that I've found, but [github](https://github.com/OpenCPN/OpenCPN) links to [sourceforge](https://sourceforge.net/projects/opencpn/) which links to their [homepage](http://www.opencpn.org/). The sourceforge links are older, but this does show the overall linkage to the product.

PR #29394, opened 2017-01-25
Update a-better-finder-attributes to 6.03

colindunn:

### DIFF
--- a/Casks/a-better-finder-attributes.rb
+++ b/Casks/a-better-finder-attributes.rb
@@ -1,10 +1,10 @@
 cask 'a-better-finder-attributes' do
-  version '6.0.1'
-  sha256 '10d248b3580d8e48934c4e88d23a6f11f7b3d3a1de5114985b2999edf55b8da7'
+  version '6.03'
+  sha256 'fdc5e960faa943a78e0d9ecafb1c3ed0b0750035bc241cb78793c6b611c7493a'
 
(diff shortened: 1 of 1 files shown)


colindunn:
This cask raises a question I've had for a while but keep forgetting to ask.

The appcast is valid, but each time the app updates the cask breaks because the download checksum fails because there aren't versioned links.

I can find all their older versions [here](http://www.publicspace.net/download/), but the latest doesn't have a versioned download.

Should all the ones like this I find be changed to latest?
Should those with static links also be changed to latest when found?

miccal:
Ping @vitorgalvao, who will be able to provide a better and more detailed answer than I can.

vitorgalvao:
This cask is correct. As long as it has an appcast it must be versioned, even if the URL isn’t.

Appcasts exist precisely so we can detect when said changes happen (more scripts to do it are in the works).

colindunn:
Thanks @vitorgalvao. I'll just move to latest those I find with no appcast and an unversioned link

vitorgalvao:
> Thanks @vitorgalvao. I'll just move to latest those I find with no appcast and an unversioned link

No rush, though. If they aren’t broken, no need to do it. The only reason some casks need `:latest` and `:no_check` is otherwise they break constantly. If there’s an app that is so rarely updated it never breaks even with a version and an unversioned url, no point in weakening it.

PR #29524, opened 2017-01-28
Update fraise to 3.7.7

abentele:

### DIFF
--- a/Casks/fraise.rb
+++ b/Casks/fraise.rb
@@ -1,10 +1,10 @@
 cask 'fraise' do
-  version '3.7.3'
-  sha256 '48a5cf964188d01fb2115185403a96a8dd57f127d7fcb7480970266f5b0eb2e0'
+  version '3.7.7'
+  sha256 '22f127873db3546020660051c93a5b8ebe6fe5419f1918fc2b8d5fa9bb0385c2'
 
-  url "https://github.com/downloads/jfmoy/Fraise/Fraise%20#{version}.zip"
+  url 'https://github.com/abentele/Fraise/releases/download/3.7.7/Fraise-3.7.7.zip'
   name 'Fraise'
-  homepage 'https://github.com/jfmoy/Fraise'
+  homepage 'https://github.com/abentele/Fraise'
 
   app 'Fraise.app'
 end


StephenKing:
I would appreciate if homebrew would follow @abentele and use his fork of Fraise.

The original [Fraise](https://github.com/jfmoy/Fraise) hasn't been updated since 2013 and is incompatible with MacOS Sierra.

abentele:
@StephenKing: thanks for supporting my request!

The author of the original Fraise has definitely discontinued maintenance years ago. I've started to maintain Fraise in 2016 to support MacOS Sierra and add some improvements and bug fixes, and I will continue to maintain it in future.

There are already a few people who tell me that they are very happy with my current release. I hope updating the cask will distribute the release to more people.

Thanks!

Casks/fraise.rb
```diff
@@ -1,10 +1,10 @@
 cask 'fraise' do
-  version '3.7.3'
-  sha256 '48a5cf964188d01fb2115185403a96a8dd57f127d7fcb7480970266f5b0eb2e0'
+  version '3.7.7'
+  sha256 '22f127873db3546020660051c93a5b8ebe6fe5419f1918fc2b8d5fa9bb0385c2'
 
-  url "https://github.com/downloads/jfmoy/Fraise/Fraise%20#{version}.zip"
+  url 'https://github.com/abentele/Fraise/releases/download/3.7.7/Fraise-3.7.7.zip'
```
vitorgalvao: Don’t remove interpolation.

Also, need an [`appcast`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

vitorgalvao (changes requested):

vitorgalvao:
That’s all well and good, but you have to realise we’ll not just start imposing specific versions of forks on users. It has 16 stars, so it isn’t even that popular, there’s no reason to make this the default. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) is clear.

If you want this added, add it as a new cask as `abentele-fraise`.

Alternatively, you can change your fork to have a differently named app and submit it as that.

vitorgalvao:
Ah, one other way. You may also get the original developer to officially say the app was discontinued and linking to your fork. With that official endorsement, your fork would become the official cask.

abentele:
Thanks for fast response.

But it's not that simple: the current maintainer already said that he discontinued the project (sorry, but I don't find the source of this information any more), but doesn't respond to my mails: I already asked him to officially say that he discontinued his project. But it's ovious that there will be no new release from the original maintainer. He even doesn't respond to comments on issues in his repository.

It simply doesn't make sense for me to have a cask definition for an about seven years old software release that isn't working with the current macOS release.

And I don't want to publish it with a name like abentele-fraise. I want to publish it as "fraise".

For this and other reasons (e.g. there are many software update sites in the internet that publish the old Fraise 3.7.3), I think about renaming the software. But until that is done, I would appreciate if you publish my updated release.

vitorgalvao:
> It simply doesn't make sense for me to have a cask definition for an about seven years old software release that isn't working with the current macOS release.

That [is already being addressed](https://github.com/caskroom/homebrew-cask/pull/29527/files).

> And I don't want to publish it with a name like abentele-fraise. I want to publish it as "fraise".

Respectfully, that is your problem, not ours. This project is for all users, not your personal preferences. You don’t get to decide on your own to take ownership of the cask. We have rules for a reason. Even if the original app is discontinued, it still is more well-known than your fork (going by number of stars). Your fork is also relatively unknown (again, judging by number of stars). It’s fair to assume users will want a fork, but assuming they’ll want *your* fork is not.

Again, we don’t impose choices on users and won’t trick them into something they are not expecting.

> But until that is done, I would appreciate if you publish my updated release.

Submit it with the requested changes, and it’ll be accepted. Otherwise it won’t.

abentele:
Thanks for the discussion. I completely understand your principles. I think it's really important to have such rules for a software like this.

As I said, it makes more sense to first rename the software (because it's not only a conflict in this case...), and then come back with the request to add a completely new cask.

vitorgalvao:
> Thanks for the discussion. I completely understand your principles. I think it's really important to have such rules for a software like this.

Thank you as well, and for understanding.

> it makes more sense to first rename the software (…), and then come back with the request to add a completely new cask.

As you prefer. Looking forward to the contribution when it’s time.

PR #29589, opened 2017-01-30
Update anki to 2.0.41

colindunn:
* the normal version supports lion and up. Doesn't seem worthwhile to maintain and download check the alternate version so removing

---


miccal:
Thanks @colindunn as always.

PR #29595, opened 2017-01-30
sublime-text: remove files_in_usr_local caveat

### LINKED ISSUE #29593: sublime-text* and other formulaes have files_in_usr_local
It seems to be no issue now or it just with `/usr/local/bin`? That caveats was used from times when we generally did not symlink binaries to `/usr/local/bin`. Today many formulas install binaries there (#12030) and I think we can safely remove it from sublime-tex* and formulas that installs only in `/usr/local/bin` but no other places?

yurikoles:

Fixes #29593


vitorgalvao:
Either we remove the `caveat` itself from every cask, or this makes no sense. You do make a valid point, though.

Will look into it on the core.

PR #29604, opened 2017-01-30
Add wmail-pre-released channel v2.0.9

fdrouet:
The existing wmail cask give access to the last stable version of WMail.
This new cask give access to the pre-release channel (see http://thomas101.github.io/wmail/download for a description of the 2 channels)

### DIFF
--- a/Casks/wmail-pre-release.rb
+++ b/Casks/wmail-pre-release.rb
@@ -1,0 +1,15 @@
+cask 'wmail-pre-release' do
+  version '2.0.9'
+  sha256 '4978e60a7b92686d88c500aeda758893185183a5454c09d048801a487cac2a7a'
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Belongs in [caskroom/versions](https://github.com/caskroom/homebrew-versions).

Also, there was a [pull request template](https://github.com/blog/2111-issue-and-pull-request-templates) when you submitted. It contains [steps that must be followed](https://raw.githubusercontent.com/caskroom/homebrew-cask/master/.github/PULL_REQUEST_TEMPLATE.md).

The template would’ve indicated this is not the correct repo.

Alternatively, you may wish to [use `cask-repair`](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask). It takes care of all of this for you.

PR #29621, opened 2017-01-30
Updated Flixtools to 2.4

sbibauw:


vitorgalvao:
Thank you, but since the url is unversioned, this change would make the cask break when it is updated upstream. There’s a reason those symbols exist and are used in the cask.

PR #29632, opened 2017-01-31
Update tomtom-mysports-connect to 3.1.6

hynyna:

### DIFF
--- a/Casks/tomtom-mysports-connect.rb
+++ b/Casks/tomtom-mysports-connect.rb
@@ -2,7 +2,8 @@ cask 'tomtom-mysports-connect' do
   version :latest
   sha256 :no_check
 
-  url 'http://download.tomtom.com/sweet/fitness/MySportsConnectDesktop/current/TomTomMySportsConnectInstaller.pkg'
+  # sports.tomtom-static.com was verified as official when first introduced to the cask
+  url 'https://sports.tomtom-static.com/downloads/desktop/mysportsconnect/latest/TomTomMySportsConnectInstaller.pkg'
   name 'TomTom MySports Connect'
   homepage 'https://www.tomtom.com/mysports/getstarted'
 


Casks/tomtom-mysports-connect.rb
```diff
@@ -1,8 +1,9 @@
 cask 'tomtom-mysports-connect' do
-  version :latest
-  sha256 :no_check
+  version '3.1.6'
+  sha256 '677d4d5a870e113d8b226c6760a0c4f9b6406718ae665c490ce43a219bdbf428'
```
miccal: Since the `url` is un-versioned and there is no `appcast`, this should be left as `version :latest` and `sha256 :no_check` - please change these back to what they were.

miccal (changes requested):

Casks/tomtom-mysports-connect.rb
HobbinHood: Understood.
Can I use cask-repair for that to follow up on that. Or directly via git?

miccal: I can do it of you, no problem.

HobbinHood: Pushed it according to

> To make changes based on feedback:
> 
> 1. Checkout your branch again with `git checkout YOUR_BRANCH_NAME`
> 2. Make any requested changes and commit them with `git add` and `git commit`
> 3. Squash new commits into one commit per formula with `git rebase --interactive origin/master`
> 4. Push to the fork's remote branch and the pull request with `git push --force`

as per [How To Open a Homebrew Pull Request (and get it merged)](https://github.com/wting/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md)

Casks/tomtom-mysports-connect.rb
```diff
@@ -2,7 +2,8 @@
   version :latest
   sha256 :no_check
 
-  url 'http://download.tomtom.com/sweet/fitness/MySportsConnectDesktop/current/TomTomMySportsConnectInstaller.pkg'
+  # sports.tomtom-static.com was verified as official when first introduced to the cask
+  url 'https://sports.tomtom-static.com/downloads/desktop/mysportsconnect/latest/TomTomMySportsConnectInstaller.pkg'
```
vitorgalvao: What’s the download page for this? Also, the old URL is still live. Why the change?

HobbinHood: If you visit https://www.tomtom.com/mysports/getstarted you can select the device you own. You're then forwarded to a site with instructions for your device (mostly the same across the product range) and download of the TomTom MySports Connect software.

I don't know why the old URL ist still live, but they released a major update to their software and apparently made it available through a new URL.
When I downloaded and installed the (old) client through brew, the first thing it did was, presenting a message popup that an update is needed and after accepting, it downloaded the new version from the new URL and initiated installation of the pkg.

Afaik there is no reason to keep the old download URL in there, since the new one directly provides the most current client (that still works with the old TomTom hardware).

Let me know if there's anything else.

vitorgalvao: And the one for this is https://uk.support.tomtom.com/app/getstarted/p/2997, correct?

vitorgalvao:
@miccal After update we should move this one to caskroom/drivers, I’d say.

Casks/tomtom-mysports-connect.rb
HobbinHood: Well it doesn't really matter. The URL I stated above redirects to country  specific TomTom pages and after that it's just different landing pages for each product with different images and text to sooth the customer. But they all download the same software.

I think https://www.tomtom.com/mysports/getstarted should be the download page. This URL is more generic and (might) stay in place a lot longer. You never know if they retire a product and remove it's site.

But if a specific URL with direct download on the page is really needed, then: Yes, any of the currently ten product landing pages will suffice.

vitorgalvao: That’s OK. I though each had different software. In that case, it makes sense as it is. Thank you for clarifying.

vitorgalvao (approved):

vitorgalvao:
Travis is taking too long. Verified locally.

PR #29635, opened 2017-01-31
binary.md: fix typo, improve example consistency

neutric:
Remove duplicate word and update example casks where commit linked to shows license stanza that is no longer being used.


neutric:
Since the GitHub diff is rather unhelpful: the duplicate word was in "directory directory" and the examples updated were "changing case" and "removing an extension."

vitorgalvao:
No ideia why the diff sometimes just gives up like this.

PR #29638, opened 2017-01-31
Unison/Panic Unison: fix name

neutric:
Fix Unison/Panic Unison name mixup accidentally introduced in #16099.


vitorgalvao (approved):
Nice catch!

PR #29641, opened 2017-01-31
Remove connected-desktop.rb

colindunn:
* homepage was 404
* when searching the product pages for a new onc, couldn't find a link to the software
* this cask was originally added by @alexg0 in 2014
* has only had a homepage updated since in 2016 by @cafferata
* appcast shows updates that haven't been applied from 2014, so the app has been out of date since then
* there have been no further updates to the application from the vendor since 2014

I suspect that this product is no longer supported and is very out of date and may even be EOL.
Can @alexg0 or @cafferata confirm usage, or should this be removed?

---


cafferata:
I don't use 'connected-desktop', the homepage update was an massa check/update on homepages.

colindunn:
Thanks for the quick response @cafferata

vitorgalvao:
Thank you for the cleaning @colindunn. I’m always up to remove useless casks.

alexg0:
@codeindulgence There was a name change.  It's now called "Transporter Desktop".  The new download page is at https://secure.connecteddata.com/download/desktop.  The product is obsoleted, but not yet EOL.  I can take a stab at fixing the cask instead of deleting.

vitorgalvao:
@alexg0 It’s already been removed. If it’s obsolete I’d rather not have it back just to have to remove it later on.

PR #29676, opened 2017-02-01
Add 10 discovered appcasts

colindunn:
* all checkpoints stable for over a week
* ~~inkscape does have an available github download, but will update the download url on the next application update~~

---

### DIFF
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -2,8 +2,8 @@ cask 'inkscape' do
   version '0.91-1'
   sha256 '66c9af880f87d5b64006391fc8dd2d3f546d93d6d91235eba7856deacba29862'
 
-  # inkscape.global.ssl.fastly.net was verified as official when first introduced to the cask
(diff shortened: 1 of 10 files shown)


Casks/inkscape.rb
```diff
@@ -4,6 +4,8 @@
 
   # inkscape.global.ssl.fastly.net was verified as official when first introduced to the cask
   url "https://inkscape.global.ssl.fastly.net/media/resources/file/Inkscape-#{version}-x11-10.7-x86_64.dmg"
```
vitorgalvao: We can change this to the one on the releases page.

vitorgalvao (changes requested):

Casks/inkscape.rb
colindunn: Done.

vitorgalvao: Ah, only now noticed you mentioned this on the top post.

vitorgalvao (approved):

PR #29681, opened 2017-02-01
Update krita to 3.1.2

neutric:

### DIFF
--- a/Casks/krita.rb
+++ b/Casks/krita.rb
@@ -3,7 +3,7 @@ cask 'krita' do
   sha256 '7d8e5b97615327122887f1114a009093e2974aa4fca01e56950fa25173b9600a'
 
   # kde.org/stable/krita was verified as official when first introduced to the cask
-  url "http://download.kde.org/stable/krita/#{version}/krita-#{version}.0.dmg"
+  url "http://download.kde.org/stable/krita/#{version}/krita-#{version}.dmg"
   name 'Krita'
   homepage 'https://krita.org/'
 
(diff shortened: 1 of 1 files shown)


eMerzh:
@neutric you should be able to drop the ".0" from the download url 
i asked upstream to make a symlink bewteen the 3.1.2 and 3.1.2.0

miccal:
The download link from [here](https://krita.org/en/download/krita-desktop/) seems to have an extra ".0":
```
http://download.kde.org/stable/krita/3.1.2/krita-3.1.2.0.dmg
```
and it is not obvious it is part of the `version`.

eMerzh:
@miccal 
http://download.kde.org/stable/krita/3.1.2/krita-3.1.2.dmg also work (it the same file)
the .0/.1 was added because of a compilation issue.
so to me it would be better to keep the "version" part in the `#{version}` variable no ?

miccal:
@eMerzh `http://download.kde.org/stable/krita/3.1.2/krita-3.1.2.dmg` redirects to `http://download.kde.org/stable/krita/3.1.2/krita-3.1.2.0.dmg`, so for now it is best to leave it as @neutric has it, and hopefully this will be rectified in the next release.

vitorgalvao:
If the `.0` is being added to the cask now and will need to be removed on the next update *and* not having the `.0` works the same, why bother adding it now?

miccal:
@eMerzh can you provide a link to your suggestion that the `.0` was added in error?

eMerzh:
I have no link  like that but on IRC they said it was to fix a build issue in Windows. 
And they've added the symlink.
 
So .0 and without zero are the same file and so, 

I'm also in favor of removing the 0 as it will make next update "harder"  but I'll be fine if it stays there

miccal:
@eMerzh done - thank you for your help.

neutric:
Thanks and sorry, wasn't at the computer all day...

vitorgalvao:
Verified locally due to Travis delays.

PR #29748, opened 2017-02-02
Delete utorrent.rb

vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/issues/6872.

We can assume utorrent is quite popular, but it has also been reported multiple times by multiples sources of containing multiple types of malware (not necessarily at the same time).

We have [rules for casks that reportedly bundle malware](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/apps_with_malware.md). While that should mean the casks can stay (and be used at user’s discretion), in this case it seems like a proper `uninstall` matters more than in other cases. As such, I propose removing it until it is submitted with an acceptable `uninstall`. It’s not like this is Little Snitch that needs administrator privilegies to install itself and do changes to your system, so it should be feasible.


victorpopkov (approved):

miccal (approved):

Saklad5:
Alternatively, there needs to be a versioned download so we can have a hash of the application.

Saklad5:
Has anyone considered contacting the uTorrent team about this?

vitorgalvao:
> Alternatively, there needs to be a versioned download so we can have a hash of the application.

That’s entirely unrelated. If you find an appcast or a versioned download, we can change it, otherwise it stays as is. There’s a reason `:latest`/`:no_check` exists.

> Has anyone considered contacting the uTorrent team about this?

I’m not sure. They might have when we first discussed the malware on this cask, but I don’t recall. Either way, not too keen on trusting a team that has bundled their app with multiple types of malware (and a bitcoin miner!) multiple times.

But if you can get them to at least provide accurate and complete instructions to remove their software (CLI-reproducible and serious, no crap instructions on moving the application to the Trash), we can add those and not remove it.

vitorgalvao:
Oh, unless you mean “contacting about this”, “this” being the removal. In that case, I @mentioned them in a tweet https://twitter.com/homebrewcask/status/827209178007687169.

PR #29770, opened 2017-02-03
hacking.md: Update instructions to force OS version

neutric:
Update instructions to force OS version to reflect Homebrew/brew#1883.

I removed the bit concerning `$MACOS_RELEASE_WITH_PATCHLEVEL` as I couldn't find the corresponding bits in the code (using GitHub's search).


reitermarkus:
Thanks!

Issue #6872, opened 2014-10-21
`installer` Casks should provide `uninstall` stanzas


rolandwalker:
Otherwise, `brew cask uninstall <cask>` will only remove the staged files in the Caskroom.
### `installer :manual` Casks that need `uninstall`
- [ ] [adobe-creative-cloud.rb](../tree/master/Casks/adobe-creative-cloud.rb)
- [x] [default-folder-x.rb](../tree/master/Casks/default-folder-x.rb)
- [ ] [little-snitch.rb](../tree/master/Casks/little-snitch.rb)
- [x] [omnifocus-clip-o-tron.rb](../tree/master/Casks/omnifocus-clip-o-tron.rb)
- [x] [utorrent.rb](../tree/master/Casks/utorrent.rb)
- [x] [vlc-remote.rb](../tree/master/Casks/vlc-remote.rb)
### `installer :script` Casks that need `uninstall`
- [x] [google-cloud-sdk.rb](../tree/master/Casks/google-cloud-sdk.rb)
### `pkg` Casks that need `uninstall`
- [x] [serviio.rb](../tree/master/Casks/serviio.rb)

jawshooah:
@caskroom/maintainers I've updated the list in the top post. We should make it a priority to add `uninstall` stanzas to all Casks with `installer` and `pkg` stanzas that need them.

vitorgalvao:
@jawshooah Last time I checked this issue, I’ve edited it (and the casks) to the last remaining ones that are popular and hard to make uninstalls for.

That said, I’d be in favour of fixing them by force: outright remove them and if/when someone misses them and wants to readd, it will need `uninstall`.

vitorgalvao:
Remaining casks can’t be reliably uninstalled without GUI uninstaller. Closing in favour of https://github.com/caskroom/homebrew-cask/issues/24377 (added the casks there as part of the checklist).

PR #29806, opened 2017-02-04
Delete google-cloud-sdk.rb

vitorgalvao:
Same logic as https://github.com/caskroom/homebrew-cask/pull/29747, https://github.com/caskroom/homebrew-cask/pull/29748, https://github.com/caskroom/homebrew-cask/pull/29749.


victorpopkov (approved):

reitermarkus (approved):

toonetown:
I actually was using it...but I'm fine with its removal if it doesn't match policy.

The application, however, is run and installed *ONLY* into the actual caskroom directory...so just fully removing `/usr/local/Caskroom/google-cloud-sdk` will delete everything (I think that's what the "default" uninstall does - according to #6872 )  So I'm not clear on what actually needs to be added as an `uninstall` stanza.

The discussion at https://groups.google.com/forum/#!topic/google-cloud-sdk/ppP3vTW4aIo explains this - and the modifications to rc files described at that link are not done in the cask (the cask installs with `--bash-completion false`, `--path-update false` and `--rc-path false`.

Also - reading through the source of the `install.sh` script that gets run, it actually does just install everything into the caskroom staged directory.

However, as I said at the beginning, I'm fine with this cask being removed.  I just thought I'd add a bit more information around the uninstallation.

vitorgalvao:
@toonetown Actually, that means the cask was already fine. Thank you for the information and the research.

Readding in https://github.com/caskroom/homebrew-cask/pull/29832 with a new note https://github.com/caskroom/homebrew-cask/blob/4f21dfba2108ec68537e406203f2c4b9f269dd32/Casks/google-cloud-sdk.rb#L17.

PR #29884, opened 2017-02-06
Add Web Archive Extractor v0.2

MattiSG:

- [ ] Checked the cask is submitted to [the correct repo]. **→ Not sure of that one. The version is 0.2, which should point to not stable, but this has not been changed for years and is surely stable in terms of usage.**

### DIFF
--- a/Casks/webarchiveextractor.rb
+++ b/Casks/webarchiveextractor.rb
@@ -5,7 +5,7 @@ cask 'webarchiveextractor' do
   url "https://downloads.sourceforge.net/webarchivext/WebArchiveExtractor.#{version}.zip"
   appcast 'https://sourceforge.net/projects/webarchivext/rss',
           checkpoint: '6fac095518b79cf496e71a1929b8ea8a91aac73023e31a5c93d40605f56f916f'
-  name 'WebArchiveExtractor'
+  name 'Web Archive Extractor'
   homepage 'http://webarchivext.sourceforge.net'
 
   app 'WebArchiveExtractor.app'
--- a/Casks/webarchiveextractor.rb
+++ b/Casks/webarchiveextractor.rb
@@ -1,0 +1,11 @@
+cask 'webarchiveextractor' do
+  version '0.2'
+  sha256 'ff9d0debc9c718b58752c1c8163fea627e461238b7f315bdbe85329d427bd7f4'
+
+  # The homepage subdomain is provided by SourceForge, and the download domain is the shared SourceForge one.
+  url "https://downloads.sourceforge.net/webarchivext/WebArchiveExtractor.#{version}.zip"
+  name 'WebArchiveExtractor'
+  homepage 'http://webarchivext.sourceforge.net'
+
+  app 'WebArchiveExtractor.app'
+end
--- a/Casks/webarchiveextractor.rb
+++ b/Casks/webarchiveextractor.rb
@@ -3,6 +3,8 @@ cask 'webarchiveextractor' do
   sha256 'ff9d0debc9c718b58752c1c8163fea627e461238b7f315bdbe85329d427bd7f4'
 
   url "https://downloads.sourceforge.net/webarchivext/WebArchiveExtractor.#{version}.zip"
+  appcast 'https://sourceforge.net/projects/webarchivext/rss',
+          checkpoint: '6fac095518b79cf496e71a1929b8ea8a91aac73023e31a5c93d40605f56f916f'
   name 'WebArchiveExtractor'
   homepage 'http://webarchivext.sourceforge.net'
 
--- a/Casks/webarchiveextractor.rb
+++ b/Casks/webarchiveextractor.rb
@@ -2,7 +2,6 @@ cask 'webarchiveextractor' do
   version '0.2'
   sha256 'ff9d0debc9c718b58752c1c8163fea627e461238b7f315bdbe85329d427bd7f4'
 
-  # The homepage subdomain is provided by SourceForge, and the download domain is the shared SourceForge one.
   url "https://downloads.sourceforge.net/webarchivext/WebArchiveExtractor.#{version}.zip"
   name 'WebArchiveExtractor'
   homepage 'http://webarchivext.sourceforge.net'


Casks/webarchiveextractor.rb
```diff
@@ -0,0 +1,11 @@
+cask 'webarchiveextractor' do
+  version '0.2'
+  sha256 'ff9d0debc9c718b58752c1c8163fea627e461238b7f315bdbe85329d427bd7f4'
+
+  # The homepage subdomain is provided by SourceForge, and the download domain is the shared SourceForge one.
```
neutric: Please stick to the [template](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment): `# downloads.sourceforge.net/webarchivext was verified as official when first introduced to the cask`.

Casks/webarchiveextractor.rb
```diff
@@ -0,0 +1,11 @@
...
+  url "https://downloads.sourceforge.net/webarchivext/WebArchiveExtractor.#{version}.zip"
```
neutric: Please add an [appcast](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

Casks/webarchiveextractor.rb
```diff
@@ -0,0 +1,11 @@
+cask 'webarchiveextractor' do
```
neutric: This should be web-archive-extractor since the name is "Web Archive Extractor". The file would have to be renamed accordingly.

Casks/webarchiveextractor.rb
```diff
@@ -0,0 +1,11 @@
...
+  version '0.2'
+  sha256 'ff9d0debc9c718b58752c1c8163fea627e461238b7f315bdbe85329d427bd7f4'
+
+  # The homepage subdomain is provided by SourceForge, and the download domain is the shared SourceForge one.
+  url "https://downloads.sourceforge.net/webarchivext/WebArchiveExtractor.#{version}.zip"
+  name 'WebArchiveExtractor'
```
neutric: This should be "Web Archive Extractor".

neutric (changes requested):
Thanks for your contribution! A few minor things need fixing...

vitorgalvao:
> Not sure of that one. The version is 0.2, which should point to not stable, but this has not been changed for years and is surely stable in terms of usage.

The docs [inform there as well](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version):

> But There Is No Stable Version!
>

Casks/webarchiveextractor.rb
vitorgalvao: This is correct as it. As per the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md), the name of the app bundle determines the token, not whatever name the developers give it (that is often inconsistent).

Casks/webarchiveextractor.rb
vitorgalvao: The comment in this case is unnecessary. It needs to be removed altogether. In other cases yes, it must follow the pattern. `style` even warns you of that, if I recall correctly.

vitorgalvao (approved):

Casks/webarchiveextractor.rb
neutric: Ah, didn't know that yet! Thanks for the correction. :)

Casks/webarchiveextractor.rb
neutric: True. I glanced at one of the most recently updated casks that have a sf URL but overlooked that the project has their own domain... -.-

MattiSG: > style even warns you of that, if I recall correctly.

At least in that case, it didn't.

I did not get from reading `CONTRIBUTING.md` that the format was mandatory. I found it only after reading the [`url.md`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) file, which I went to only after getting the comment here.

> As a side note, I find it slightly weird to use a comment to affirm authenticity when the review process should be enough to prove a verification was made.

Casks/webarchiveextractor.rb
MattiSG: I don't know: the developer seems to use indifferently “Web Archive Extractor”, “WebArchiveExtractor”, and “WebArchive Extractor”. The latter would actually make the most sense, “WebArchive” being a file format.

Casks/webarchiveextractor.rb
MattiSG: I actually wrote the cask a first time naming it `web-archive-extractor`, then re-read the token reference once again after finding discrepancies between app name, doc, and HTML presentation file  ;)

vitorgalvao: > finding discrepancies between app name, doc, and HTML presentation file ;)

Those discrepancies led to a huge duplicates problem (since each person would interpret it differently), hence why we had to make the token reference.

Casks/webarchiveextractor.rb
vitorgalvao: > As a side note, I find it slightly weird to use a comment to affirm authenticity when the review process should be enough to prove a verification was made.

We’re only human, and as such can make mistakes. The comment serves to inform users we did our due diligence but that they should still be careful (hence why we say “when first introduced to the cask”). It also helps with the review, since the absence of a comment when it is needed will indeed trigger and error in `style`.

Casks/webarchiveextractor.rb
neutric: Again, sorry for the confusion! Vitor and Miccal and the other guys are putting in so much effort here that I thought I'll help out here and there. Unfortunately, being corrected like this in my case seems to help memorize "the rules" much better than reading the docs and memorizing intricacies dozens of times.

miccal: No problem @neutric, @vitorgalvao has corrected my stuff-ups more times than I can remember, but that is how we learn.

PR #29919, opened 2017-02-07
little-snitch: Add versions for older OS releases

neutric:
See https://www.obdev.at/products/littlesnitch/download-legacy.html

### DIFF
--- a/Casks/little-snitch.rb
+++ b/Casks/little-snitch.rb
@@ -1,12 +1,24 @@
 cask 'little-snitch' do
-  version '3.7.3'
-  sha256 'd3074fe0fb57bef70a77c6630071845800dde2eebeaa28e01673ee2a749668f3'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Do you personally need this, @neutric? If not, I’d rather not add it. There was never a need for it, and adding it without anyone needing seems like unnecessary overhead. This’ll just be removed later when it stops working, having helped no one.

neutric:
I don't need it personally, but thought it might make sense for a tool as useful/important as Little Snitch. I don't know the percentage of people running something older than 10.10, but I still have a Macbook 2,1 which now runs Linux but if it were to run macOS still, I wouldn't be able to get it anywhere near 10.10 iirc. So I don't insist on it, but the work of adding the bits in is done. Removing them only takes a few seconds... your call.

neutric:
:)

PR #29945, opened 2017-02-08
Update monolingual to 1.7.6

ghost:

### DIFF
--- a/Casks/monolingual.rb
+++ b/Casks/monolingual.rb
@@ -16,7 +16,7 @@ cask 'monolingual' do
     sha256 '24fa5ff0a5903c0eb07cd58a15292e3adab97ea0823f304241dc4187f9252ffc'
   else
     version '1.7.6'
-    sha256 '24fa5ff0a5903c0eb07cd58a15292e3adab97ea0823f304241dc4187f9252ffc'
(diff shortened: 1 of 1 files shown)


neutric:
Thanks for catching and fixing this! `cask-repair` must not like Casks catering to different OS releases...

neutric:
...or the problem more likely is that I am on 10.11 still and thus `cask-repair` downloaded and checksummed the version for 10.11.

vitorgalvao:
> ...or the problem more likely is that I am on 10.11 still and thus `cask-repair` downloaded and checksummed the version for 10.11.

Yep. `cask-repair` uses bash and a bunch of trickery. It always replaces the last `sha256`, but at the same time uses `fetch`, so it downloads according to your macOS version.

Will fix that shortly to always `fetch` for the latest macOS.

neutric:
I encountered the same thing a while ago, just forgot about it... looking forward to the new shiny.

vitorgalvao:
https://github.com/vitorgalvao/tiny-scripts/commit/ba00020f0b876e258114ed6cb25f78c630b826d3.

Thank you for the notice, @neutric.

neutric:
Thanks for the swift fix, @vitorgalvao!

PR #29956, opened 2017-02-09
Update Keybase to 1.0.18-20170208165022,dc66461

joshka:
Includes new Keybase chat feature.
See https://keybase.io/blog/keybase-chat for announcement details.

Add pre and post flight install / uninstall scripts.

### DIFF
--- a/Casks/keybase.rb
+++ b/Casks/keybase.rb
@@ -1,14 +1,27 @@
 cask 'keybase' do
-  version '1.0.18-20170207120017,ce0f90d'
-  sha256 '791f510ed8440e9fa59e5dae2b632ce8427ad50a682e5c3792e4b127a410a638'
+  version '1.0.18-20170209165653,17b641d'
+  sha256 'a5a974089e94e68ac3a5113570e903d0b88fe243d04474107645a5c792d13693'
 
   url "https://prerelease.keybase.io/darwin/Keybase-#{version.before_comma}%2B#{version.after_comma}.dmg"
   appcast 'https://prerelease.keybase.io/update-darwin-prod-v2.json',
-          checkpoint: 'a8454aca49d7dbb496f5fb4e5dc646c3ddeb2fb3f253d561600ea0682f3052f1'
+          checkpoint: '37f720c5a5c9be9cbf429c0b755a4b024ab900f5758f64b195bdf00c672634b4'
   name 'Keybase'
   homepage 'https://keybase.io/'
 
   auto_updates true
 
   app 'Keybase.app'
+
+  postflight do
+    system_command '/Applications/Keybase.App/Contents/Resources/KeybaseInstaller.app/Contents/MacOS/Keybase',
+                   args: ['--app-path=/Applications/Keybase.app', '--run-mode=prod', '--timeout=10']
+  end
+
+  uninstall_preflight do
+    if system_command('launchctl', args: ['list']).stdout =~ %r{/^\d+.*keybase.Electron/}
+      system_command 'killall', args: ['-kill', 'Keybase']
+    end
+    system_command '/Applications/Keybase.app/Contents/SharedSupport/bin/keybase',
+                   args: ['uninstall']
+  end
 end
--- a/Casks/keybase.rb
+++ b/Casks/keybase.rb
@@ -14,7 +14,7 @@ cask 'keybase' do
 
   postflight do
     system_command "#{appdir}/Keybase.App/Contents/Resources/KeybaseInstaller.app/Contents/MacOS/Keybase",
-                   args: ['--app-path=/Applications/Keybase.app', '--run-mode=prod', '--timeout=10']
+                   args: ["--app-path=#{appdir}/Keybase.app", '--run-mode=prod', '--timeout=10']
   end
 
   uninstall_preflight do
--- a/Casks/keybase.rb
+++ b/Casks/keybase.rb
@@ -13,7 +13,7 @@ cask 'keybase' do
   app 'Keybase.app'
 
   postflight do
-    system_command '/Applications/Keybase.App/Contents/Resources/KeybaseInstaller.app/Contents/MacOS/Keybase',
+    system_command "#{appdir}/Keybase.App/Contents/Resources/KeybaseInstaller.app/Contents/MacOS/Keybase",
                    args: ['--app-path=/Applications/Keybase.app', '--run-mode=prod', '--timeout=10']
   end
 
@@ -21,7 +21,7 @@ cask 'keybase' do
     if system_command('launchctl', args: ['list']).stdout =~ %r{/^\d+.*keybase.Electron/}
       system_command 'killall', args: ['-kill', 'Keybase']
     end
-    system_command '/Applications/Keybase.app/Contents/SharedSupport/bin/keybase',
+    system_command "#{appdir}/Keybase.app/Contents/SharedSupport/bin/keybase",
                    args: ['uninstall']
   end
 end


Casks/keybase.rb
```diff
@@ -1,14 +1,24 @@
...
   app 'Keybase.app'
+
+  postflight do
+    system_command '/Applications/Keybase.App/Contents/Resources/KeybaseInstaller.app/Contents/MacOS/Keybase',
+                   args: ['--app-path=/Applications/Keybase.app', '--run-mode=prod', '--timeout=10']
+  end
+
+  uninstall_preflight do
+    system 'killall', '-kill', 'Keybase'
+    system '/Applications/Keybase.app/Contents/SharedSupport/bin/keybase', 'uninstall'
```
miccal: Please use `system_command` instead of `system` in the `uninstall_preflight`, as you did in the `postflight`.

miccal:
@joshka you seem to updated an old version of `keybase.rb` which is causing a merge conflict.

Casks/keybase.rb
joshka: This is intentional. I do not want the killall / uninstall command to fail when the Keybase app is not running or has been removed.

joshka:
@miccal fixed the merge conflict

Casks/keybase.rb
miccal: `system` has been deprecated; you must use `system_command`. See discussion [here](https://github.com/caskroom/homebrew-cask/issues/23070).

joshka: I understand that's the policy. I'd like to be consistent with it, but it fails when Keybase is not running. There is a `:must_succeed` argument to `system_command`, but passing `false` has no effect.

If I remove the killall and Keybase is running during uninstall, a bar icon continues to be displayed. Quitting using Cmd+Q does not dismiss the application. Only quitting from the dock gets rid of the open application / bar icon.

TL;DR I need a way of running system_command that does not fail when the command fails. Until that's available, this is the best way to run these commands and prevent the orphaned icon / app.

reitermarkus: You can check if it is running first, and only then kill it.

```ruby
unless system_command('launchctl', args: ['list']).stdout.lines.grep(/keybase.Electron/).empty?
  # kill it
end
```

vitorgalvao: I never understood (and this is not a criticism but a serious question) why people “TL;DR” *at the end* of a post. If a post is so long I don’t want to read it, I’m not going to scroll to the very end to see if by any chance that person had a “TL;DR”, I just won’t read it.

Why not just put a “TL;WR” (too long; won’t read) or summary at the start?

joshka: I often use TL;DR to signify approximately "thanks for reading the reasons above, pay attention I'm asking a question down here that gets to the heart of the matter now that you have the context"

TL;DR "pay attention to this bit" ;)

joshka:
Review comments all addressed. I changed the regex necessary to only include running keybase.Electron instances not stopped instances.

Casks/keybase.rb
vitorgalvao: @joshka Ah, so you use it as a sort of “in conclusion”.

Casks/keybase.rb
```diff
@@ -1,14 +1,27 @@
...
   app 'Keybase.app'
+
+  postflight do
+    system_command "#{appdir}/Keybase.App/Contents/Resources/KeybaseInstaller.app/Contents/MacOS/Keybase",
+                   args: ["--app-path=#{appdir}/Keybase.app", '--run-mode=prod', '--timeout=10']
+  end
+
+  uninstall_preflight do
+    if system_command('launchctl', args: ['list']).stdout =~ %r{/^\d+.*keybase.Electron/}
+      system_command 'killall', args: ['-kill', 'Keybase']
```
vitorgalvao: Why is this in an `uninstall_preflight` instead of a regular `uninstall`? It can take care of quitting apps.

Casks/keybase.rb
```diff
@@ -1,14 +1,27 @@
...
+    end
+    system_command "#{appdir}/Keybase.app/Contents/SharedSupport/bin/keybase",
```
vitorgalvao: On that same token, why isn’t this script handled by an `uninstall` as well?

vitorgalvao: Also, isn’t `Keybase.app/Contents/SharedSupport/bin/keybase` a useful CLI? Shouldn’t it be added in a `binary`?

joshka: The app stanza runs before the uninstall stanza, so the binary gets removed and hence is not able to run in the uninstall stanza. Killing the app first is required before running the keybase uninstall command.

This app *is indeed* a useful CLI. It is installed to the right place (/usr/local/bin) during either the first run of Keybase.app, or by the command I added in postflight stanza above.

installer / first run installs services, cli, and taskbar app (maybe some more stuff). Requires app to be in /Applications or ~/Applications already hence cannot be run from a script stanza.
uninstall command requires that the app still exists in /Applications, hence cannot be run from an uninstall stanza.

joshka:
Also, it seems like this should deprecate the homebrew formula of the same name as it installs both the App and the CLI. The fact that keybase.io/download no longer ships a CLI only version of Keybase supports this. Once this merge goes through, I'll request a PR to drop the keybase formula with a reference to the cask.

gabriel:
You don't need to run the `KeybaseInstaller.app` postflight because it happens automatically on app start.

Also, the uninstall can call `/Applications/Keybase.app/Contents/SharedSupport/bin/keybase uninstall` which will properly remove all the things it installs.

joshka:
@gabriel yes. You're right. However by waiting until first run of the app start you have to run the app to get the command line to work.
The uninstall preflight does call the command you mentioned. See line 24

gabriel:
Oops, yes... 

Thanks for setting up the cask!

joshka:
No problem :)

Casks/keybase.rb
```diff
@@ -1,14 +1,27 @@
...
 
   app 'Keybase.app'
+
+  postflight do
+    system_command "#{appdir}/Keybase.App/Contents/Resources/KeybaseInstaller.app/Contents/MacOS/Keybase",
+                   args: ["--app-path=#{appdir}/Keybase.app", '--run-mode=prod', '--timeout=10']
+  end
+
+  uninstall_preflight do
+    if system_command('launchctl', args: ['list']).stdout =~ %r{/^\d+.*keybase.Electron/}
```
gabriel: On Sierra, launchd list presents Keybase app as:

```
> launchctl list | grep keybase
2166	0	keybase.Electron.19224
```

which fails this regex.

joshka: I'm not sure that it fails that regex. http://rubular.com/r/xXYsVBMOck

gabriel: Ahh maybe failing another way:

- Install Keybase (not from cask, from https://keybase.io/docs/the_app/install_macos)
- Run Keybase.app
- Run `brew cask install keybase`

It doesn't kill the running Keybase.

joshka: The uninstall preflight stanza only runs when uninstalling a version of keybase installed by cask. The regex works fine for cask installed keybase.

This behavior seems like it's more of a HBC bug rather than a bug of this cask. Perhaps HBC should stop the install rather than removing an existing app at that location / attempting to overwrite. I'll raise an issue to talk over this.

Here's the output that I see when attempting what you've described.

```
$ brew cask install keybase
==> Downloading https://prerelease.keybase.io/darwin/Keybase-1.0.18-20170209165653%2B17b641d.dmg
Already downloaded: /Users/joshka/Library/Caches/Homebrew/Cask/keybase--1.0.18-20170209165653,17b641d.dmg
==> Verifying checksum for Cask keybase
==> Removing App: '/Applications/Keybase.app'
Error: It seems there is already an App at '/Applications/Keybase.app'.
Error: nothing to install
```

To get past the issue, right click the icon in the dock and quit keybase.

gabriel: This is what I see:

```sh
kbmb ➜ ~ brew cask install keybase
==> Downloading https://prerelease.keybase.io/darwin/Keybase-1.0.18-20170209165653%2B17b641d.dmg
Already downloaded: /Users/gabe/Library/Caches/Homebrew/Cask/keybase--1.0.18-20170209165653,17b641d.dmg
==> Verifying checksum for Cask keybase
==> 2017-02-15T16:40:04.177053-08:00 ▶ [INFO keybase launchd.go:291] 001 Removing /Users/gabe/Library/LaunchAgents/keybase.kbfs.plist
==> 2017-02-15T16:40:04.181578-08:00 ▶ [INFO keybase launchd.go:217] 002 Waiting for keybase.kbfs to exit...
==> 2017-02-15T16:40:04.391075-08:00 ▶ [INFO keybase launchd.go:160] 003 Stopped keybase.kbfs
==> 2017-02-15T16:40:04.394875-08:00 ▶ [INFO keybase mounter_nix.go:29] 004 Trying to unmount: /keybase
==> 2017-02-15T16:40:04.407014-08:00 ▶ [INFO keybase launchd.go:291] 005 Removing /Users/gabe/Library/LaunchAgents/keybase.service.plist
==> 2017-02-15T16:40:04.412145-08:00 ▶ [INFO keybase launchd.go:217] 006 Waiting for keybase.service to exit...
==> 2017-02-15T16:40:04.624894-08:00 ▶ [INFO keybase launchd.go:160] 007 Stopped keybase.service
==> 2017-02-15T16:40:04.629004-08:00 ▶ [INFO keybase launchd.go:291] 008 Removing /Users/gabe/Library/LaunchAgents/keybase.updater.plist
==> 2017-02-15T16:40:04.632981-08:00 ▶ [INFO keybase launchd.go:217] 009 Waiting for keybase.updater to exit...
==> 2017-02-15T16:40:04.845010-08:00 ▶ [INFO keybase launchd.go:160] 00a Stopped keybase.updater
==> 2017-02-15T16:40:04.970214-08:00 ▶ [INFO keybase install_osx.go:694] 00b Removing KBFuse
==> 2017-02-15T16:40:05.055503-08:00 ▶ [INFO keybase install_osx.go:699] 00c Removing privileged helper tool
==> 2017-02-15T16:40:05.133359-08:00 ▶ [INFO keybase install.go:222] 00d Removing /usr/local/bin/keybase
Error: It seems there is already an App at '/Applications/Keybase.app'.
Error: nothing to install
```

joshka: Let's continue this over in #30142

PR #30032, opened 2017-02-12
Update homepage for 43 Casks

miccal:
Update homepage for 43 Casks.

### DIFF
--- a/Casks/sierra-cache-cleaner.rb
+++ b/Casks/sierra-cache-cleaner.rb
@@ -5,7 +5,7 @@ cask 'sierra-cache-cleaner' do
   url 'http://www.northernsoftworks.com/downloads/scc.dmg'
   name 'Sierra Cache Cleaner'
   name 'SCC'
-  homepage 'http://www.northernsoftworks.com/sierracachecleaner.html'
+  homepage 'http://www.northernsoftworks.com/sierracachecleaner.html/'
 
   depends_on macos: '<= :sierra'
   if MacOS.version >= :lion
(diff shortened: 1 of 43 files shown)


Casks/sierra-cache-cleaner.rb
```diff
@@ -5,7 +5,7 @@
   url 'http://www.northernsoftworks.com/downloads/scc.dmg'
   name 'Sierra Cache Cleaner'
   name 'SCC'
-  homepage 'http://www.northernsoftworks.com/sierracachecleaner.html'
+  homepage 'http://www.northernsoftworks.com/sierracachecleaner.html/'
```
vitorgalvao: Doesn’t make sense to have a trailing slash in cases like this. In this case, it even shows a broken version of the page.

Casks/wordservice.rb
```diff
@@ -5,7 +5,7 @@
   # amazonaws.com/DTWebsiteSupport was verified as official when first introduced to the cask
   url "https://s3.amazonaws.com/DTWebsiteSupport/download/freeware/wordservice/#{version}/WordService.zip"
   name 'WordService'
-  homepage 'http://www.devontechnologies.com/products/freeware.html#c1115'
+  homepage 'http://www.devontechnologies.com/products/freeware.html#c1115/'
```
vitorgalvao: Same here. The slash actually impeded following the anchor.

miccal: Fixed - sorry @vitorgalvao, I just use @victorpopkov's [list](https://caskroom.victorpopkov.com/homebrew-cask/homepages.csv) to update the homepages, and did not think about these cases.

PR #30034, opened 2017-02-12
Add wsk (OpenWhisk CLI binary)

shaftoe:

- [ ] The commit message includes the cask’s name and version: **no version available, using `:latest`**

- [x] `brew cask uninstall {{cask_file}}` worked successfully. **Caveat: asks for (sudo) password even though it should not be needed**

### DIFF
--- a/Casks/wsk.rb
+++ b/Casks/wsk.rb
@@ -1,0 +1,12 @@
+cask 'wsk' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://openwhisk.ng.bluemix.net/cli/go/download/mac/amd64/OpenWhisk_CLI-mac.zip'
+  name 'OpenWhisk CLI'
+  homepage 'https://console.ng.bluemix.net/openwhisk/cli'
+
+  binary 'wsk'
+
+  uninstall delete: '/usr/local/bin/wsk'
+end


miccal:
This seems to be a CLI-only open-source tool. Please submit it first to [Homebrew/core](https://github.com/Homebrew/homebrew-core) and only if they don’t accept it, submit here (linking back to the pull request where it was refused).

shaftoe:
@miccal all right, thanks for the explanation, will do

shaftoe:
@miccal I've tried but it proved way too complex for what I was willing to invest in this. Plus, I don't see that much of a value to force the (Mac) user to install that big stack (Java, Docker, Gradle) in order to run a Go binary (downloading which is by the way the official and suggested way to install the software), so I ask you again to please consider this as a valid Cask formula.

Homebrew-core pull request here: https://github.com/Homebrew/homebrew-core/pull/9879

Thank you!

miccal:
Ping @vitorgalvao.

MikeMcQuaid:
@shaftoe It's not really open source software if it's too hard for anyone to build it. You won't find any Linux package manager that's just willing to redistribute an upstream binary as-is, either.

shaftoe:
@MikeMcQuaid personally I don't think "easy to build" is a strict requirement to make a software "Open Source", but honestly I'm not really interested in neither building the tool from sources nor convincing package managers to ship that, I just want to have an easy way to drop the binary in my workstation's $PATH and given that I already use Homebrew Cask (between other tools) for this kind of task, I thought a new Cask formula could be a good fit instead of building my own solution, which also could benefits other OpenWhisk users too. I'm also already installing closed source software via Cask so I wasn't expecting this to be a problem. What is your suggestion?

MikeMcQuaid:
> What is your suggestion?

You'll probably be unsurprised but disappointed to hear: get it building from source into Homebrew or  [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

shaftoe:
@MikeMcQuaid creating my own tap seems actually quite an easy and good-enough solution to me! Thank you for your help!

shaftoe:
For the records: https://github.com/shaftoe/homebrew-wsk

vitorgalvao:
Considering the difficulties of building this from source and the dependencies it has (java), I’d be fine with adding this here.

It does need some `depends_on` and the removal of the `uninstall` (redundant).

markusthoemmes:
It's not dependend on Java, Gradle or Docker I think. It should be compilable using plain Go just fine.

shaftoe:
@vitorgalvao thanks! I'll remove`uninstall`, but I think I'm not understanding the semantics of `depends_on`, because I don't know which dependencies are you referring to: the zip file contains a single Go binary compiled for MacOS which by design has no dependencies.

markusthoemmes:
@shaftoe the source for it is plain Go though. OpenWhisk itself uses Scala and ships in Docker containers (thus the Java and Docker dependency). It should be manageable to build the CLI source using just Go though.

shaftoe:
@markusthoemmes yes you're right, but I was just following [the official docs](https://github.com/openwhisk/openwhisk/tree/master/tools/cli) on how to compile the source code for the CLI

MikeMcQuaid:
@vitorgalvao Up to you but I think you may well be opening a can of worms. It's always going to be harder to make a formula than a cask and I think it's a reasonable requirement that open source projects be buildable without too much effort

---

As a project management aside: this is somewhat an ideological thing that I think results in a better open source software ecosystem. Homebrew has done similarly with things that don't build without patching, don't build with clang, don't build without X11 etc. and this does seem to have had the effect of nudging upstreams in the right direction if they wish to support macOS. In this case, providing better and easier build instructions feels like a good, doable outcome of this gentle pressure.

vitorgalvao:
@MikeMcQuaid Agreed. We do tend to prefer that attitude as well (avoid workarounds, try to fix upstream first), so I see your point.

PR #30079, opened 2017-02-13
Add wine-devel 2.1

vszakats:
The install logic is identical to the existing `wine-staging` formula. The difference is in content, where this package only includes features already endorsed and integrated into the Wine development branch, i.e. nothing experimental or out-of-branch. So, it's useful to get an official Wine binary build that is as close as possible to the stable development branch.

I'm unsure if the package will be updated in-place, but if it turns out to be, the hash can be removed.

Another option is to install Wine via `brew install wine`, but that method has the downside that it needs to keep rebuilding all its dependencies from source from the point of installation, making it very resource and time hungry and therefore unpractical in practice.

### DIFF
--- a/Casks/wine-devel.rb
+++ b/Casks/wine-devel.rb
@@ -1,0 +1,33 @@
+cask 'wine-devel' do
+  version '2.1'
+  sha256 '2a84e1d3aeaabf82ebab71f9a546909ce9b40bb4cffae062dc96545150bf930a'
+
+  url "https://dl.winehq.org/wine-builds/macosx/i686/winehq-devel-#{version}.pkg"
+  name 'WineHQ-devel'
+  homepage 'https://wiki.winehq.org/MacOS'
+
+  depends_on cask: 'xquartz'
+
+  pkg "winehq-devel-#{version}.pkg",
+      choices: [
+                 {
+                   'choiceIdentifier' => 'choice3',
+                   'choiceAttribute'  => 'selected',
+                   'attributeSetting' => 1,
+                 },
+               ]
+
+  uninstall pkgutil: [
+                       'org.winehq.wine-devel',
+                       'org.winehq.wine-devel-deps',
+                       'org.winehq.wine-devel-deps64',
+                       'org.winehq.wine-devel32',
+                       'org.winehq.wine-devel64',
+                     ],
+            delete:  '/Applications/Wine Devel.app'
+
+  caveats <<-EOS.undent
+    #{token} installs support for running 64 bit applications in Wine, which is considered experimental.
+    If you do not want 64 bit support, you should download and install the #{token} package manually.
+  EOS
+end


vszakats:
Regarding the "Devel" status of this binary: It is, but, the existing `wine-stable` is a development version as well, but possibly less stable than this one due the inclusion of experimental features. So I'd argue that this package would be an improvement over the other, even if this one has 'devel' in its name, while 'staging' doesn't.

miccal (approved):

vszakats:
Made install/uninstall tests and cleaned up the commit message.

vitorgalvao:
I’m a bit torn on this one, but lets add it for now.

miccal:
> I’m a bit torn on this one, but lets add it for now.

@vitorgalvao out of interest, can you explain why? Is it because of Wine existing in the core?

vszakats:
I've tried the core Wine (`brew install wine`). For it to work properly it requires to rebuild all the dependent packages from source: `openssl freetype jpeg libtool libusb libusb-compat fontconfig libtiff gd little-cms2 libtasn1 libxml2`. This takes substantial amount of time when installing Wine, so f.e. installing it on a CI machine is out of question. It's not only that: From that point on, every time an update come for any of these dependencies, they will be rebuilt from source again an again.

Uninstalling Wine doesn't reset things to their original state (as expected): The user has to one-by-one remove-than-reinstall them in the correct order, to restore the pre-Wine configuration. [Maybe there exist a `brew` command to reinstall a package with default/bottled configuration, but I could not find it. ]

Compared to this, a `brew cask install wine-devel` and `brew cask uninstall wine-devel`, just works and the only downside is that the Wine `bin` directory needs to be added to `PATH` manually.

vitorgalvao:
I’m less concerned with it being in HB as well (thank you for the detailed explanation, @vszakats) and more with this being the correct repo for it.

As it stands, it’s likelier this gets removed to [caskroom/versions](https://github.com/caskroom/homebrew-versions) than be removed.

miccal:
> Regarding the "Devel" status of this binary: It is, but, the existing wine-staging is a development version as well, just possibly less stable than this one due the inclusion of experimental/out-of-branch features. So I'd argue this package would be an improvement over the existing formula, even if this one has 'devel' in its name, while 'staging' doesn't.

As pointed out by @vszakats, this one is actually more 'stable` than `wine-staging`.

@vitorgalvao it might make sense to move `wine-staging` to [caskroom/versions](https://github.com/caskroom/homebrew-versions), no?

vszakats:
I'd be glad if it stayed there, for these reasons:
- It's considered more stable than `wine-staging` which is already here.
- Currently there doesn't exist a superior binary build for macOS.
- Adding a tap is not a huge issue (though one more command to maintain), but it's not easily _discoverable_.
- Homebrew is actively working on migrating its `homebrew-versions` tap into the core one (though it has a somewhat different meaning there.)

I agree with @miccal.

vszakats:
It seems that switching to `wine-devel` from `wine-staging` fixes a slowdown bug on Travis CI machines that was causing a 4-minute delay each time `wine` was called. The delay could be somewhat reduced by using an obscure [hack](https://github.com/fracting/msys2-scripts/blob/dbb2bbc9b6f91d8107e5498ac3d0d4e7da71c765/src/msys2-init#L23).

Here are the running-time results:

1. Without Wine, **16 min 23 sec**:
   https://travis-ci.org/vszakats/harbour-core/jobs/201160889
2. With `wine-staging` as-is, **31 min 57 sec**:
   https://travis-ci.org/vszakats/harbour-core/jobs/201507887
3. With `wine-staging` and the hack `DISPLAY=:88.0; wineboot`, **24 min 23 sec**:
   https://travis-ci.org/vszakats/harbour-core/jobs/201631719
4. With `wine-devel` and the hack, **16 min 53 sec**:
   https://travis-ci.org/vszakats/harbour-core/jobs/201735494
5. With `wine-devel` as-is (without the hack), **16 min 42 sec**:
   https://travis-ci.org/vszakats/harbour-core/jobs/201825907

It means that using `wine-devel`, the only overhead is running those few Windows commands with the help of Wine, which is expected and correct.

This won't make `wine-devel` a non-devel package, but at least at this moment and in the above scenario, it's a more usable package compared to `wine-staging`.

vszakats:
(disregard my deleted comment about Travis CI comparison, there was a mistake in it)

vszakats:
After https://github.com/Homebrew/homebrew-core/commit/ca2b723507b79e91d9fcede838859b61058dd991, this formula has less, if any importance. The new native formula also resolves the wineboot hang/wait/slowdown (experienced with both `wine-devel and `wine-staging`), so it's better suited for CI environments where this problem has been seen.

Please feel free to decide to delete it or move it to development branch as planned originally.

Thanks to everyone for bringing wine to Homebrew.

PR #30102, opened 2017-02-14
Update couchbase-server-enterprise to 4.6.0-DP

cgriego:

### DIFF
--- a/Casks/couchbase-server-enterprise.rb
+++ b/Casks/couchbase-server-enterprise.rb
@@ -1,8 +1,8 @@
 cask 'couchbase-server-enterprise' do
-  version '4.5.1'
-  sha256 'de014c7c134eb97ff00be6b2e6f5d0da84295ce05bbb7bb3a4d3c747a365cd22'
+  version '4.6.0-DP'
+  sha256 '4b52ad87d703c235b2f9e4e63a8c1b6f4a720d36b28b81cbf593947256399a3f'
 
-  url "http://packages.couchbase.com/releases/#{version}/couchbase-server-enterprise_#{version}-macos_x86_64.zip"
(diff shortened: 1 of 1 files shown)


cgriego:
I wouldn't normally submit a pull request for a developer preview, but 4.6.0-DP is the only version that supports macOS Sierra. 4.5.1 doesn't function on Sierra.

reitermarkus:
Let's leave it at `4.5.1` for versions below Sierra. See https://github.com/caskroom/homebrew-cask/tree/master/doc/cask_language_reference#version-comparisons.

Casks/couchbase-server-enterprise.rb
```diff
@@ -1,8 +1,13 @@
 cask 'couchbase-server-enterprise' do
-  version '4.5.1'
-  sha256 'de014c7c134eb97ff00be6b2e6f5d0da84295ce05bbb7bb3a4d3c747a365cd22'
+  if MacOS.version < :sierra
```
reitermarkus: We usually prefer `<=` for checks, so please use `<= :el_capitan` here.

cgriego: I saw that the majority of checks used `<=`, but I also saw an exception in the r-app cask that matched this use case and the `<` code read like your comment. I've updated to `<=` for consistency.

reitermarkus: Thanks, seems like this was overlooked for `r-app`.

miccal: > Thanks, seems like this was overlooked for `r-app`.

Fixed in https://github.com/caskroom/homebrew-cask/pull/30113.

miccal:

PR #30121, opened 2017-02-15
Update RawTherapee to 5.0

BenjaminHCCarr:
New packaging format and URL structure for the 5.0 releases
4.2.1:`http://rawtherapee.com/releases_head/mac/RawTherapee_OSX_10.7_64_4.2.1.zip`
5.0-r1:`http://www.rawtherapee.com/shared/builds/mac/RawTherapee_OSX_10.9_5.0-r1_gtk3.app.zip`

5.0 format has no `DMG` just a zipped `.app`. Also the 5.0 version is named `RawTherapee-gtk3.app`

As this is developer change, I left as is, but not sure if we need to nuke the old one as it won't over-write.
The other option is an `app 'RawTherapee-gtk3.app', target: 'RawTherapee.app'`

Either way we are left with possible duplicates of non-concurrent versions. Unless a zap (I think) is put in for the `pre-5.0` version.
---

---
Note on `Line 10`, no matter how many times I tried I keep ending up with 4 spaces on line 10 for no clear reason. Please delete, even pasting the:
`1 file inspected, 1 offense detected, 1 offense corrected`
code I ended up with the whitespace.
Sorry.

### DIFF
--- a/Casks/rawtherapee.rb
+++ b/Casks/rawtherapee.rb
@@ -1,31 +1,23 @@
 cask 'rawtherapee' do
-  if MacOS.version <= :mavericks
-    version '4.2.270'
-    sha256 '231f991cf43c92e21aace5ebc34c229542cbd3902633d091a8cd2c049216b29d'
+  if MacOS.version <= :lion
+    version '4.2.1'
+    sha256 'd553f04cab4e09a1d66d8811be5ec8468171abe584bdb541ff97d9fa0841e32f'
 
-    url "http://rawtherapee.com/releases_head/mac/RawTherapee_OSX_10.6_64_#{version}.zip"
-  elsif MacOS.version <= :yosemite
-    version '4.2.349'
-    sha256 'd88f4dc10740b9a105d0c6d72b8d8801957b5e3f2923912afc76b308ebb049bb'
-
-    url "http://rawtherapee.com/releases_head/mac/RawTherapee_OSX_10.10_64_#{version}.zip"
+    url "http://rawtherapee.com/releases_head/mac/RawTherapee_OSX_10.7_64_#{version}.zip"
   else
-    version '4.2.1025'
-    sha256 'e3e8f4b919528884c2e51dcb0cd0b746ea1ca32cae5317495bca80807648f84b'
-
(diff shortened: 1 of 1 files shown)


Casks/rawtherapee.rb
```diff
@@ -1,31 +1,23 @@
 cask 'rawtherapee' do
-  if MacOS.version <= :mavericks
-    version '4.2.270'
-    sha256 '231f991cf43c92e21aace5ebc34c229542cbd3902633d091a8cd2c049216b29d'
+  if MacOS.version <= :lion
```
miccal: I think you can remove this - it seems version 5 works for >= macOS 10.8, so lets just use the latest version.

BenjaminHCCarr: > I  think you can remove this - it seems version 5 works for >= macOS 10.8

Version 5 lists 10.8/9/10/11/12 64-bit as supported.

Have we deprecated lion and below (didn't see a final verdict)? The 4.2.1 release still supports 10.6/7, which is why I included it as the old version has this tiered format for older versions.

Happy to oblige if that is consensus though.

BenjaminHCCarr:
The other questions is what to do with the new `.app` name:
5.0-r1 at least is `RawTherapee-gtk3.app`
Previous version were `RawTherapee.app`

---

Do we want in this case to follow the developers convention and keep the gtk string, or to do an app rename?

Casks/rawtherapee.rb
miccal: We officially only support the current OS version and at most two behind, so I suggest we just remove it.

miccal:
> Do we want in this case to follow the developers convention and keep the gtk string, or to do an app rename?

Go with the developer's convention I would say.

BenjaminHCCarr:
Agreed, just wanted to check, as they are bifurcating the code: `gtk2` and `gtk3`; but we're not building it anyway.

So I should yank lion and below, and we'll merge then?

miccal:
> So I should yank lion and below

Yes please.

miccal (approved):

alanorth:
Works for me, clean install on macOS 10.12.

PR #30189, opened 2017-02-17
Windows, Linux, and WebGL build support for Unity, version 5.5.0f3

sjml:

- [X] Checked the cask was not already refused in [closed issues].

(This pull request includes multiple casks; if that's not considered appropriate, please let me know and I can break them up.)

These casks (and the ones I based them on, `unity-ios-support-for-editor` and `unity-android-support-for-editor`) might more ideally be supported as installation options on the `unity` cask. Is that something that's possible? Even with these additions, we're still missing the Tizen, Xbox One, PS Vita, PS4, and Samsung TV support packages. Making a separate cask for each one doesn't really scale; as it is there are already multiple casks that will need to be updated when the core `unity` package gets a version bump. Any thoughts?

### DIFF
--- a/Casks/unity-linux-support-for-editor.rb
+++ b/Casks/unity-linux-support-for-editor.rb
@@ -1,0 +1,14 @@
+cask 'unity-linux-support-for-editor' do
+  version '5.5.0f3,38b4efef76f0'
+  sha256 'fa76f2b509c825040a1466a3b9f2f7ae16eaea7124ad478a1b56d750d26e0cb5'
+
+  url "http://netstorage.unity3d.com/unity/#{version.after_comma}/MacEditorTargetInstaller/UnitySetup-Linux-Support-for-Editor-#{version.before_comma}.pkg"
+  name 'Unity Linux Build Support'
+  homepage 'https://unity3d.com/unity/'
+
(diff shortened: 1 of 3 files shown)


adidalal:
Multiple related casks are fine. It's possible to script updates, for example the navicat casks call all be updated together with a for loop in bash and cask-repair

sjml:
Looks like a download timeout made one of the checks fail erroneously (the other check passed fine). Is there a way to kick the Travis build again or do I need to do a dummy push?

adidalal:
I can restart it and if it still times out, I can verify locally. Just post again if it fails so I get a notification.

sjml:
Thanks for kicking the build! 

As for the multiple related casks -- so as long as whoever updates the Unity cask knows to update the dependent casks, too, it'll be fine. My concern is that with every dependent cask it becomes easier for them to get out of sync, and getting the versions lined up is pretty important. 

But I'm not a maintainer here, so if you say it's cool, I trust you!

adidalal:
I think we have flash, xamarin, navicat, and unity so far in this bucket. If it gets unreasonable we'll think of something.

Thanks for the contribution!

PR #30208, opened 2017-02-19
new cask "erbele" 3.8.2

abentele:
please add my new cask for Erbele.app


vitorgalvao:
Thank you for the submissions, but this cask is self-submitted and too obscure, so we’ll have to refuse it. [This policy is documented](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/rejected_casks.md).

Also, next time please do fill the pull request template.

PR #30305, opened 2017-02-22
Fix capitalisation typo in keybase.rb.

kirb:

---

Fixes the postflight failing on a case-sensitive filesystem because `Keybase.app` had a capital `A`.

(Is the version needed in the commit message in this case? Seems that point could do with clarification.)

### DIFF
--- a/Casks/keybase.rb
+++ b/Casks/keybase.rb
@@ -13,7 +13,7 @@ cask 'keybase' do
   app 'Keybase.app'
 
   postflight do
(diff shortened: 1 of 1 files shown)


miketheman:
This has been bugging me for a few days, and thanks to this change, I was able to install on my case-sensitive drive.

vitorgalvao:
> Is the version needed in the commit message in this case?

No. It’s better to not tick what doesn’t apply then just tick everything.

PR #30433, opened 2017-02-27
Revert "Update gcc-arm-embedded to 6-2017-q1"

miccal:
Reverts caskroom/homebrew-cask#30356.

Download links for the new version are down. so reverting to the previous version for now.


jinyuanxie:
The url work now: https://developer.arm.com/-/media/Files/downloads/gnu-rm/6_1-2017q1/gcc-arm-none-eabi-6-2017-q1-update-mac.tar.bz2

screen shot 2017-03-05 at 01 06 13

PR #30441, opened 2017-02-28
Update qiyimedia.rb

nuomi1:
Update qiyimedia.rb


miccal:
Since the `url` is un-versioned and there is no appcast, the `version` should remain as `:latest` and the `sha256` should remain as `:no_check`.

PR #30447, opened 2017-02-28
Update rdm to 0.9.0-alpha3,0.9.0.41

josser:


vitorgalvao:
Thank you, but this repo is for stable versions.

josser:
Oh, I understand. 
But this alpha is from november 2016 and looks like actually stable
Also, author didn't release stable builds to public asking for subscription. So, this is the only way to get rdm on mac via cask. 
However, I think it would be better to create brew formula to build stable version from source.

vitorgalvao:
Took a closer look, and it seems like [they’re working on it](https://github.com/uglide/RedisDesktopManager#release-schedule).

Looks like this cask should remain as it was. Thank you again.

PR #30472, opened 2017-03-01
Update kap to 1.0.1

neutric:

### DIFF
--- a/Casks/kap.rb
+++ b/Casks/kap.rb
@@ -1,9 +1,9 @@
 cask 'kap' do
-  version '1.0.1'
+  version '1.0.1,1'
   sha256 'f51902982b80925b6cf1fe81c920d4a3a0bb841dfb5bbae5879f934beeb2eae7'
 
   # github.com/wulkano/kap was verified as official when first introduced to the cask
-  url "https://github.com/wulkano/kap/releases/download/v#{version}/Kap-#{version}.1.dmg"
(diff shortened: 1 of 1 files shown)


Casks/kap.rb
```diff
@@ -1,11 +1,11 @@
 cask 'kap' do
-  version '1.0.0'
-  sha256 '04f3a950988fb60b6f52b9962ad8adccd9786a4b00e2ae6638c6a3130788afcd'
+  version '1.0.1'
+  sha256 'f51902982b80925b6cf1fe81c920d4a3a0bb841dfb5bbae5879f934beeb2eae7'
 
   # github.com/wulkano/kap was verified as official when first introduced to the cask
-  url "https://github.com/wulkano/kap/releases/download/v#{version}/Kap-#{version}.dmg"
+  url "https://github.com/wulkano/kap/releases/download/v#{version}/Kap-#{version}.1.dmg"
```
vitorgalvao: The extra `.1` should be added to the `version`.

vitorgalvao (changes requested):

Casks/kap.rb
neutric: Even if the `.1` is an exception judging by the (admittedly not overly long) [release history](https://github.com/wulkano/kap/releases) and is not part of the version when looking at "About" in the app itself? It doesn't make future updates easier if you have to go and change back the interpolation strings. I however can see that it documents the exception a tad bit better, provided that commit title was updated to `Update kap to 1.0.1,1`.

PR #30478, opened 2017-03-01
Update tableau to 10.2.0

dmohns:


dmohns:
Apparently the format on the tableau installer has changed. It now provides and `pkg`-installer file inside the `dmg`. I would assume the update therefore would require more work to be done, namely transform the cask into a pkg-installer based one.

dmohns:
Will close and transform into issue.

PR #30492, opened 2017-03-01
Add "vscode" as an alias for Visual Studio Code

ebraminio:
Not sure if this is accepted but I hoped brew guide me better when I wrote "brew cask install vscode" and I guessed this can help on that.


vitorgalvao:
Thank you, but `VS Code` should be enough. `name` doesn’t yet work, anyway.

PR #30589, opened 2017-03-05
Update uninstall.md

miccal:
See https://github.com/caskroom/homebrew-cask/pull/30570 for reference.


miccal:
Thanks for the updates @vitorgalvao.

PR #30696, opened 2017-03-08
Update gcc-arm-embedded.rb

miccal:
Closes https://github.com/caskroom/homebrew-cask/issues/30634.


adidalal (approved):

vitorgalvao:
`20161216` is *really* specific, while `2016q4` is more geral. Even if the former changed 365 times in a year, the latter would only change four times. As such, if we have to miss interpolating one, I’d rather lose the latter than the former.

adidalal:
It simplifies the rest of the Cask though, IMO

vitorgalvao (approved):

PR #30740, opened 2017-03-08
Delete waterroof.app

gobinathm:
- The Home page & app download URL seems to be broken. It returns a HTTP 503


gobinathm:
Closes the bug #30741 , i have opened. 

I understand that doing a PR & opening a bug will be duplicate of work. But i'm not sure about the directions we will be taking on this issue.

Issue #30791, opened 2017-03-10
Bug report: yasu returns 406 Not Acceptable


gobinathm:
#### Description of issue

The URL returns 406 Not Acceptable. Tested it from few different networks (ISP)

<details><summary>Output of ` brew cask fetch --download --verbose yasu`</summary>

```
==> Downloading external files for Cask paymowidget
==> Downloading https://app.paymoapp.com/desktop-apps/paymowidget/PaymoWidget_mac_4.0.16.dmg
/usr/bin/curl --remote-time --location --user-agent Homebrew/1.1.11 (Macintosh; Intel macOS 10.12.3) curl/7.51.0 --fail https://app.paymoapp.com/desktop-apps/paymowidget/PaymoWidget_mac_4.0.16.dmg -C 0 -o /Users/gobinathm/Library/Caches/Homebrew/Cask/paymowidget--4.0.16.dmg.incomplete
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
Error: Download failed on Cask 'paymowidget' with message: Download failed: https://app.paymoapp.com/desktop-apps/paymowidget/PaymoWidget_mac_4.0.16.dmg
The incomplete download is cached at /Users/gobinathm/Library/Caches/Homebrew/Cask/paymowidget--4.0.16.dmg.incomplete
Error: Kernel.exit
```

</details>

<details><summary>Output of `brew doctor`</summary>

```
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: You have unlinked kegs in your Cellar
Leaving kegs unlinked can lead to build-trouble and cause brews that depend on
those kegs to fail to run properly once built. Run `brew link` on these:
  php70
```
</details>

<details><summary>Output of `brew cask doctor`</summary>

```
==> Homebrew-Cask Version
Homebrew-Cask 1.1.11-23-gb38c52f
caskroom/homebrew-cask (git revision 3cf30; last commit 2017-03-10)
==> Homebrew-Cask Install Location
<NONE>
==> Homebrew-Cask Staging Location
/usr/local/Caskroom
==> Homebrew-Cask Cached Downloads
[...]

PR #30795, opened 2017-03-10
Deleting yasu.rb

gobinathm:
- Package download is no more possible
- Ref: #30791


vitorgalvao:
[As mentioned](https://github.com/caskroom/homebrew-cask/issues/30791#issuecomment-285603523), the download does work manually.

PR #30823, opened 2017-03-11
Add `uninstall :pkgutil` to SpiderOakOne.

reitermarkus:

### DIFF
--- a/Casks/spideroakone.rb
+++ b/Casks/spideroakone.rb
@@ -8,5 +8,6 @@ cask 'spideroakone' do
 
   pkg 'SpiderOakONE.pkg'
 
-  uninstall delete: '/Applications/SpiderOakONE.app'
+  uninstall delete:  '/Applications/SpiderOakONE.app',
+            pkgutil: 'org.python.python'
 end


Casks/spideroakone.rb
```diff
@@ -8,5 +8,6 @@
 
   pkg 'SpiderOakONE.pkg'
 
-  uninstall delete: '/Applications/SpiderOakONE.app'
+  uninstall delete:  '/Applications/SpiderOakONE.app',
```
vitorgalvao: Won’t the `delete:` be unnecessary now, due to the recent change that gets rid of empty `app` bundles?

reitermarkus: No, because there is no package id for SpiderOakONE, only for the Python which gets installed inside `/Applications/SpiderOakONE.app`:

```text
$ pkgutil --pkg-info-plist org.python.python
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>install-location</key>
	<string>Applications/SpiderOakONE.app</string>
	<key>install-time</key>
	<integer>1489239030</integer>
	<key>pkg-version</key>
	<string>2.7.7</string>
	<key>pkgid</key>
	<string>org.python.python</string>
	<key>receipt-plist-version</key>
	<real>1</real>
	<key>volume</key>
	<string>/</string>
</dict>
</plist>
```

This is some really bad design. I don't get why they don't simply include it in the `.app` in the first place, when they only use the `.pkg` to move the `.app`, and *then* install Python into it.

PR #30923, opened 2017-03-14
Add cask avast-secureline-vpn

gtmgianni:
I can't get the uninstall script to show me its arguments (if it has any), and additionally the user needs to restart their system immediately after the uninstall (because the script launches the GUI uninstaller, there is no other option for the user). Also, I included a commit in this pull request which changes the cask `avast` to `avast-mac-security` to avoid confusion between the two casks.

### DIFF
--- a/Casks/avast-secureline-vpn.rb
+++ b/Casks/avast-secureline-vpn.rb
@@ -1,0 +1,23 @@
+cask 'avast-secureline-vpn' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://mac.sl.d.avast.com/mac/avast_secureline.dmg'
+  name 'Avast SecureLine VPN'
+  homepage 'https://www.avast.com/secureline-vpn#mac'
+
+  pkg 'Avast SecureLine VPN.pkg'
+
+  uninstall script:   '/Applications/AvastSecureLine.app/Contents/Applications/AvastSecureLineUninstaller.app/Contents/MacOS/AvastSecureLineUninstaller',
+            pkgutil:   'com.avast.secureline',
+            launchctl: [
+                         'com.avast.secureline.userinit',
+                         'com.avast.secureline.update-agent',
+                         '*.com.avast.osx.secureline.avastsecurelinehelper',
+                       ]
+
+  zap delete: [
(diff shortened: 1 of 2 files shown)


giannitm:
Additionally, the `zap` stanza cannot be executed because it always first runs the `uninstall` sequence, which will either result in the computer having to be restarted after uninstallation, or throwing up an error because the cask is already uninstalled and the uninstaller no longer exists. Either way, the `zap` stanza is not run.

giannitm:
I assume this behaviour is intentional, but if a user runs the `zap` command and the uninstall stanza fails (for whatever reason), the `zap` command is not executed afterwards. This issue occurs with the cask `trusteer-rapport` also.

Casks/avast-secureline-vpn.rb
```diff
@@ -0,0 +1,23 @@
...
+  version :latest
+  sha256 :no_check
+
+  url 'http://mac.sl.d.avast.com/mac/avast_secureline.dmg'
+  name 'Avast SecureLine VPN'
+  homepage 'https://www.avast.com/secureline-vpn#mac'
+
+  pkg 'Avast SecureLine VPN.pkg'
+
+  uninstall script:   '/Applications/AvastSecureLine.app/Contents/Applications/AvastSecureLineUninstaller.app/Contents/MacOS/AvastSecureLineUninstaller',
```
vitorgalvao: This is a wrong use of `uninstall script:`. Like you mentioned, this is not a script but a GUI. Please remove it. Also see https://github.com/caskroom/homebrew-cask/issues/24377.

vitorgalvao (changes requested):

giannitm:
Actually, don't bother answering my other query @vitorgalvao, I've just realised the reason I'm having issues with `zap`.
`zap` is designed to be run instead of `uninstall` if you want a more thorough uninstallation, right? I've been using `zap` as a sort of "cleanup" command for cleaning up files after uninstallation, rather than an "uninstall" command in its own right. Just using `zap` instead of `uninstall` fixes the problems I was having with `trusteer-rapport` and this new cask.

PR #30933, opened 2017-03-15
Update aws-vault to 3.7.1

mtibben:

### DIFF
--- a/Casks/aws-vault.rb
+++ b/Casks/aws-vault.rb
@@ -1,10 +1,10 @@
 cask 'aws-vault' do
-  version '3.7.0'
-  sha256 'd17dfd93591efeaaceaca8d2eba8371f2ecc1543c4dc5d4441333c557330b04a'
+  version '3.7.1'
+  sha256 '6ca2ac45cd227f38fb8374509f0ea0665fa0b34db5cb8724e238b87622dec459'
 
   url "https://github.com/99designs/aws-vault/releases/download/v#{version}/aws-vault-darwin-amd64"
   appcast 'https://github.com/99designs/aws-vault/releases.atom',
-          checkpoint: 'f1668dcf1b7e0b144d30ce76b3d1af473448bcc13579cf6a03a5419c41ecbe44'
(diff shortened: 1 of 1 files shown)


Casks/aws-vault.rb
```diff
@@ -1,10 +1,8 @@
 cask 'aws-vault' do
-  version '3.7.0'
-  sha256 'd17dfd93591efeaaceaca8d2eba8371f2ecc1543c4dc5d4441333c557330b04a'
+  version '3.7.1'
+  sha256 '6ca2ac45cd227f38fb8374509f0ea0665fa0b34db5cb8724e238b87622dec459'
 
   url "https://github.com/99designs/aws-vault/releases/download/v#{version}/aws-vault-darwin-amd64"
-  appcast 'https://github.com/99designs/aws-vault/releases.atom',
-          checkpoint: 'f1668dcf1b7e0b144d30ce76b3d1af473448bcc13579cf6a03a5419c41ecbe44'
```
miccal: Do not remove the `appcast`, please add it back.

miccal (changes requested):

Casks/aws-vault.rb
mtibben: For what purpose? Cask doesn't use it at all

vitorgalvao: @mtibben Yes, we do. Why would we have the feature if we had no intentions of using it? Furthermore, why would you jus remove something you *think* has no use? Why not ask about it?

`appcast` is for maintainer use. Please add is back, as requested.

mtibben: That's a pretty rude reply to an innocent question, I did just ask about it. "For what purpose?". Frankly if that's the level of discourse in this project, I'll pull this cask and start my own tap

adidalal: @mtibben It's needed for the (eventual) implementation of a auto-update mechanism. Until then. you can use `cask-repair` instead of a manual edit/calculation. (details in [CONTRIBUTING](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md))

mtibben: Thank you @adidalal for the polite explanation

Issue #30901, opened 2017-03-14
Bug report: reinstall logitech-control-center fails


sheim:
#### Cannot uninstall or reinstall logitech-control-center

I'm on macOS 10.12.3 (16D32).  
Both "reinstall" and "uninstall" result in the same error:

<details><summary>`brew cask uninstall logitech-control-center --verbose`</summary>

```
==> Downloading https://www.logitech.com/pub/techsupport/mouse/mac/lcc3.9.5.66.zip
Already downloaded: /Users/sheim/Library/Caches/Homebrew/Cask/logitech-control-center--3.9.5.66.zip
==> Verifying checksum for Cask logitech-control-center
==> Running uninstall process for logitech-control-center; your password may be necessary
==> Running uninstall script LCC Installer.app/Contents/Resources/LCC Uninstaller Tool
==> 2017-03-14 15:41:50.249 LCC Uninstaller Tool[7986:123550] Running LCC Uninstall tool. Arguments = (
==>     "/usr/local/Caskroom/logitech-control-center/3.9.5.66/LCC Installer.app/Contents/Resources/LCC Uninstaller Tool"
==> )
==> Uninstalling packages:
com.Logitech.Control Center.pkg
Error: incompatible encoding regexp match (UTF-8 regexp with ASCII-8BIT string)
Follow the instructions here:
  https://github.com/caskroom/homebrew-cask/blob/master/doc/reporting_bugs/pre_bug_report.md

If this doesn’t fix the problem, please report this bug:
  https://github.com/caskroom/homebrew-cask#reporting-bugs

/usr/local/Homebrew/Library/Homebrew/vendor/plist/plist/parser.rb:91:in `scan'
/usr/local/Homebrew/Library/Homebrew/vendor/plist/plist/parser.rb:91:in `parse'
/usr/local/Homebrew/Library/Homebrew/vendor/plist/plist/parser.rb:29:in `parse_xml'
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/system_command.rb:160:in `_parse_plist'
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/system_command.rb:131:in `plist'
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/pkg.rb:75:in `info'
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/pkg.rb:66:in `pkgutil_bom_all'
[...]

PR #30938, opened 2017-03-15
Remove logitech-control-center Cask

gobinathm:
- PR can be leverage, in case the maintainer decision is to delete logitech-control-center Cask
- Ref : #30901


dzhus:
Does it mean that there's no more way to make extra buttons on Logitech mice work?

gobinathm:
I guess so. when tried a manually install of existing software it redirected me to https://secure.logitech.com/maclcc, but that page is issuing a 404. I believe Logitech discontinues this development

screen shot 2017-03-29 at 12 59 58 pm

dzhus:
Ugh, what a mess. Particularly funny since this mouse works on Linux out of the box since, like, forever.

PR #30967, opened 2017-03-16
Deleting anonym Cask

gobinathm:
- Website & download URL’s are returning 404.
- Tried Tracing the APP, but could not locate alternatives


vitorgalvao:
This one works fine for me. Maybe it’s back up from when you tried.

gobinathm:
@vitorgalvao The site was returning 404 & even i save a google cached version of 404. But guess its fixed. Thanks

PR #30971, opened 2017-03-16
Deleting identify cask

gobinathm:
- Both the Website URL & Application URL returns 404. Could not track any alternatives.


vitorgalvao:
The website does not return a 404 to me, but:
image

Either way the download link was one of the old Dropbox links that will stop working.

PR #31062, opened 2017-03-18
Updating ringtones.rb

gobinathm:
- SSL seems to be expired about 3 months back. While installation this gets in to a error


miccal:
This Cask works if you switch to `http`, so just change the `https`'s to `http`.

gobinathm:
@miccal sure, done with the change

miccal:
Thank you.

PR #31073, opened 2017-03-18
textwrangler: Mark as discontinued

neutric:

---

See [announcement](http://us11.campaign-archive1.com/?u=d7ffaa16f302eaf61e416e389&id=5c0940cf9b). Textwrangler will be replaced by BBEdit, which is free to use with reduced feature set after the evaluation period is over.


vitorgalvao:
Let's outright remove it. It's not only being discontinued, it’s being replaced, so there is no need to keep it around.

PR #31076, opened 2017-03-18
apple-events: Update appcast hash and add zap delete stanza

gtmgianni:
The appcast hash was outdated (though the cask version remains the same) and I also added a `zap delete` stanza. Passed audit and style checks.

### DIFF
--- a/Casks/apple-events.rb
+++ b/Casks/apple-events.rb
@@ -13,7 +13,7 @@ cask 'apple-events' do
   app 'Apple Events.app'
 
   zap delete: [
-                '~/Library/Application\ Support/br.com.guilhermerambo.Apple-Events',
+                '~/Library/Application Support/br.com.guilhermerambo.Apple-Events',
                 '~/Library/Caches/br.com.guilhermerambo.Apple-Events',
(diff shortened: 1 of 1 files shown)


Casks/apple-events.rb
```diff
@@ -4,11 +4,17 @@
...
+          checkpoint: 'cdf5faa46cdf2b6215c68424a681bba8462d5f4e0d39791929f2411b4bd08b9f'
   name 'Apple Events'
   homepage 'https://github.com/insidegui/AppleEvents'
 
   auto_updates true
 
   app 'Apple Events.app'
+
+  zap delete: [
+                '~/Library/Application Support/br.com.guilhermerambo.Apple-Events',
```
vitorgalvao: You keep using `\ ` In quoted strings. The `\` is an escape character that tells the interpreter “treat the next character as a literal space, instead of a divisor to the next argument”. If you include it in a quoted string, the it’ll include the literal `\` character, and fail, because such path does not exist.

giannitm: Got it.

PR #31110, opened 2017-03-19
Add Frescobaldi 2.20.0.

marnen:
Add [Frescobaldi](http://www.frescobaldi.org) 2.20.0.

***

### DIFF
--- a/Casks/frescobaldi.rb
+++ b/Casks/frescobaldi.rb
@@ -4,13 +4,14 @@ cask 'frescobaldi' do
 
   # github.com/wbsoft/frescobaldi was verified as official when first introduced to the cask
   url 'https://github.com/wbsoft/frescobaldi/releases/download/v2.20.0/Frescobaldi-2.20.0-x86_64.dmg'
-  appcast 'https://github.com/wbsoft/frescobaldi/releases.atom', checkpoint: '1edb6cbd5767024e224f7aecda697d206b49a180066180dd3201359cf4f6a598'
+  appcast 'https://github.com/wbsoft/frescobaldi/releases.atom',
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):

vitorgalvao:
> Many users of this cask will already have a LilyPond installation, which may or may not be managed through Cask

That’s the case for every `depends_on` case. By that logic, we’d never use `depends` on and would always have `caveats`. Lilypond, as installed by as, is merely an app, so depending on it won’t be harmful.

marnen:
@vitorgalvao I don't think you understood my point. This is not a case of mandatory dependency, so `depends_on` is inappropriate in this case. 

Does Caskroom have a concept of "recommended but optional dependency"? If so, I'll use that.

vitorgalvao:
> Does Caskroom have a concept of "recommended but optional dependency"?

We do not.

In that case, [I’ve removed the dependency](https://github.com/caskroom/homebrew-cask/commit/2fe6d65fe5b6bc0983c3bd6495ccf6aa86d54c0b). A `caveat` is also not appropriate, as those are to be used for necessary installation instructions. The correct case here is to have nothing as [we are not a discoverability service and users are expected to have reasonable knowledge of what they are installing](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md).

marnen:
I suppose that works. I was thinking that the caveat was good for the case of "oh crap, why can't I use the IDE to compile LilyPond source", while not forcing experienced LilyPond users to install a second copy of LilyPond. But I understand your point about discoverability.

PR #30955, opened 2017-03-15
macvim.rb: remove extensive link renames

vitorgalvao:
After making all changes to the cask:

- [ ] `brew cask audit --download {{cask_file}}` is error-free.
- [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
- [ ] The commit message includes the cask’s name and version.

---

Closes #30953.

PR #31178, opened 2017-03-22
Update broken binary in MacVim 8.0.129

tinhkyaw:
- [x] brew cask audit --download {{cask_file}} is error-free.
- [x] brew cask style --fix {{cask_file}} left no offenses.

### DIFF
--- a/Casks/macvim.rb
+++ b/Casks/macvim.rb
@@ -12,7 +12,10 @@ cask 'macvim' do
   depends_on macos: '>= :mountain_lion'
 
   app 'MacVim.app'
+  binary "#{appdir}/MacVim.app/Contents/MacOS/MacVim"
+  binary "#{appdir}/MacVim.app/Contents/MacOS/macvim-askpass"
   binary "#{appdir}/MacVim.app/Contents/MacOS/mvim"
+  binary "#{appdir}/MacVim.app/Contents/MacOS/Vim"
 
   zap delete: [
                 '~/Library/Caches/org.vim.MacVim',
--- a/Casks/macvim.rb
+++ b/Casks/macvim.rb
@@ -1,10 +1,10 @@
 cask 'macvim' do
-  version '8.0.128'
-  sha256 'e3763cf5978850eb85c8c333fc2b064c6ef1615c2d9f89445d4b385ca7fa916d'
+  version '8.0.129'
+  sha256 'a75d4999986f65b12b5adecde9dbdcf30ae04a19545a65b0131783c547897cf2'
 
   url "https://github.com/macvim-dev/macvim/releases/download/snapshot-#{version.patch}/MacVim.dmg"
   appcast 'https://github.com/macvim-dev/macvim/releases.atom',
-          checkpoint: '7f624738f4adf4fb2a6c19fe6d04dfb5577805140f05df0a9e512df8d2fdce73'
+          checkpoint: 'b94f207852bed3a9ece4d15be8163ed9a9b6abd18a751e0ae02123f42bb8b8c2'
   name 'MacVim'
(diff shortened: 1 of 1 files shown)


lifepillar:
I have tried this, but running `mvim` results in

```
/usr/local/bin/mvim: line 86: /usr/local/bin/../../../MacVim.app/Contents/MacOS/Vim: No such file or directory
/usr/local/bin/mvim: line 86: exec: /usr/local/bin/../../../MacVim.app/Contents/MacOS/Vim: cannot execute: No such file or directory
```

Also, this cask should create links for `mvimdiff`, `mview`, and possibly `gview`, `gvim`, `gvimdiff` (but *not* `view` and `vimdiff`). Although they are all symlinks to `mvim`, they behave differently (e.g., `mview` opens a file in read-only mode).

fedde-s:
Is this related to issue #31178?

lifepillar:
@fedde-s Yes, this pull request aims at addressing that issue.

fedde-s:
Thanks! About the symlinks, they seem to have been removed in an attempt to solve an unrelated issue in pull request #30955.

tinhkyaw:
Thanks for updating this.  I just knew it was broken and I only looked back 1 PR before it was broken.  Should have looked further :-).

13k:
@lifepillar that shell script error seems to be a bug from upstream. For some reason the `mvim` shell script in the app package sets the vim application directory by using a rather naive method:

```
VIM_APP_DIR=`dirname "$0"`/../../..
```

When dealing with symlinks, this will most certainly break, which is almost 100% of the cases when using macvim. What's odd, though, is that that file in [macvim's repository](https://github.com/macvim-dev/macvim/blob/snapshot-128/src/MacVim/mvim#L14) doesn't set `VIM_APP_DIR` and if you check file history, it was never changed in that way. Only the shipped file within the app package is setting that variable.

miccal:
I have updated this to version 8.0.129 - @lifepillar can you test this?

miccal:
Tested locally - seems to be working now.

lifepillar:
@miccal No, it's not fixed yet. Please note:

- snapshot-129 has moved binaries into `MacVim.app/Contents/bin/`, not `MacVim.app/Contents/MacOS`.
- you shouldn't link `MacVim` and `Vim`.

@13k How about creating a shim script that sets `$VIM_APP_DIR$`, instead of symlinking?

lifepillar:
Ok, this is what works for me:
https://gist.github.com/lifepillar/5540d93b197308b0b8820ce746b1415a

miccal:
https://github.com/caskroom/homebrew-cask/pull/31217

lifepillar:
@miccal That looks fine.

Is there a reason why the other links cannot be reinstated, as in my gist above? I cannot reproduce https://github.com/caskroom/homebrew-cask/issues/30953 (which seems the reason why they were removed in the first place) and, according to https://github.com/caskroom/homebrew-cask/pull/30955, that issue should have been fixed by https://github.com/Homebrew/brew/pull/2347.

Having `mview`, `mvimdiff` etc… would be useful.

miccal:
@lifepillar you are welcome to open a PR yourself if you like.

13k:
@lifepillar answering your question directed to me: Despite installers being probably the best place where to configure paths, since they know the concrete installation paths, I don't think we need. MacVim seems to do a pretty good job at guessing the paths already.

lifepillar:
@13k In fact, with the latest snapshot MacVim's scripts work just fine. I have opened a [PR](https://github.com/caskroom/homebrew-cask/pull/31218) as suggested by @miccal to create the necessary symlinks.

13k:
:+1:

PR #31240, opened 2017-03-24
Migrate exa to Homebrew

iKevinY:
`exa` is purely a command line app, so it probably should have been a core formula rather than a Cask. However, at the time, there were no formula for projects built using Rust, so it seemed easier to simply copy over a prebuilt binary. Now that there are quite a few Rust-based core formulae, it makes sense to migrate `exa` over to Homebrew. (Also, the current release is [completely broken](https://github.com/ogham/exa/issues/140) on Sierra.)

Link to new `homebrew-core` formula PR: ~~Homebrew/homebrew-core#11529~~ [#11817](https://github.com/Homebrew/homebrew-core/issues/11817)


vitorgalvao:
Closing because it got rejected in https://github.com/Homebrew/homebrew-core/pull/11529.

But now we support this, and that if you can get it into HB, we’ll gladly remove it from here.

fxcoudert:
`exa` has been added to Homebrew.

miccal:
Thanks @iKevinY and @fxcoudert.

PR #31242, opened 2017-03-24
update most jetbrains casks to include build number

leipert:
Recently we had to update the casks intellij-idea (https://github.com/caskroom/homebrew-cask/pull/31190) and intellij-idea-ce (https://github.com/caskroom/homebrew-cask/pull/31209), because jetbrains swapped out the `2017.1` with another build.

This commit adds the `build` number to the version string. The `build` number is publicly announced at the app casts, for example here for clion: https://data.services.jetbrains.com/products/releases?code=CL&latest=true&type=release


miccal:
If every one of these Casks uses `#{version.before_comma}` only in the interpolation, then what is the purpose of the `version` after the comma?

leipert:
The purpose of it is being a bit more verbose if a background swap of binaries by jetbrains happens (See linked merge requests in initial comment).

The ` #{version.after_comma}` could actually be used in the appcast url for example:

https://data.services.jetbrains.com/products/releases?code=AC&build=163.13906.24

instead of 

https://data.services.jetbrains.com/products/releases?code=AC&latest=true&type=release

I do not know if the appcast url should contain a version number or not

victorpopkov (approved):

miccal:
@leipert are you able to fix the merge conflicts?

leipert:
done 👍 
Do you think, we should update the app cask urls to contain the build number?

reitermarkus:
> Do you think, we should update the app cask urls to contain the build number?

Does the appcast change if the build number is given as a parameter, or does it just show this build number's release notes? It should always change if there is a new version.

leipert:
> It should always change if there is a new version.

Then the cast url should remain as they are

leipert:
So what do the maintainers think? Do we want do this change or not?

leipert:
👍 

Updated @jcbot to 1.1.0 for automatic updates including build number:
https://github.com/leipert/jetbrains-cask-bot/releases/tag/v1.1.0

PR #31259, opened 2017-03-24
Fix deprecated brew-cask-completion

dadoonet:
Fix this message:

```
Warning: Use brew-cask-completion instead of deprecated homebrew/completions/brew-cask-completion
```


vitorgalvao:
Thank you.

PR #31303, opened 2017-03-25
Update curse.rb

Jerzerak:

No need for cask version in commit message because it isn't being updated, just a few stanzas


vitorgalvao:
Thank you, but lets remove it instead. It has been discontinued in favour of another app, and we have that other app.

Jerzerak:
👍🏻

PR #31344, opened 2017-03-27
Rename AppGate-Client to lowercase

numbermaniac:
`brew cask info` fails on the current version of this cask, and I'm pretty sure it's because of the casing. Hopefully this fixes that. (discovered this by accident when `cask _stanza` threw a warning)

```
>cask info AppGate-Client
Error: Cask 'appgate-client' definition is invalid: Bad header line: 'AppGate-Client' does not match file name
>cask info appgate-client
Error: Cask 'appgate-client' definition is invalid: Bad header line: 'AppGate-Client' does not match file name
```

### DIFF
--- a/Casks/appgate-client.rb
+++ b/Casks/appgate-client.rb
@@ -1,4 +1,4 @@
-cask 'AppGate-Client' do
+cask 'appgate-client' do
   version '11.2.3'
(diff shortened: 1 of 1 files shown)


miccal:
Thank you.

PR #31361, opened 2017-03-27
Delete xrg cask

gobinathm:
- the application is not downloadable manually. Checked with multiple ISP's


giannitm:
@gobinathm This appears to work on my computer, and `brew cask install xrg` also completes with no errors.

gobinathm:
Strange now its working for me as well :(

PR #31591, opened 2017-03-29
add new Cask for ManicTime 1.3.3

waldyrious:

### DIFF
--- a/Casks/manictime.rb
+++ b/Casks/manictime.rb
@@ -4,9 +4,13 @@ cask 'manictime' do
 
   url "http://cdn.manictime.com/setup/mac/ManicTime-v#{version}.dmg"
   name 'ManicTime'
-  homepage 'http://www.manictime.com/'
+  homepage 'http://www.manictime.com/Mac'
 
   pkg "ManicTime-#{version}.pkg"
 
   uninstall pkgutil: 'com.finkit.manictime.tracker'
+  
+  caveats "This is not a full ManicTime Windows port to Mac. "\
+          "It is only a tracker, and  requires ManicTime Server v3.2 or later"\
+          "installed on a Windows machine to receive and display the reports."\
 end
--- a/Casks/manictime.rb
+++ b/Casks/manictime.rb
@@ -9,7 +9,7 @@ cask 'manictime' do
   pkg "ManicTime-#{version}.pkg"
 
   uninstall pkgutil: 'com.finkit.manictime.tracker'
-  
+
   caveats 'This is not a full ManicTime Windows port to Mac. '\
           'It is only a tracker, and  requires ManicTime Server v3.2 or later '\
           'installed on a Windows machine to receive and display the reports.'
--- a/Casks/manictime.rb
+++ b/Casks/manictime.rb
@@ -8,5 +8,5 @@ cask 'manictime' do
 
   pkg "ManicTime-#{version}.pkg"
 
-  uninstall pkgutil 'com.finkit.manictime.tracker'
+  uninstall pkgutil: 'com.finkit.manictime.tracker'
 end
--- a/Casks/manictime.rb
+++ b/Casks/manictime.rb
@@ -9,8 +9,4 @@ cask 'manictime' do
   pkg "ManicTime-#{version}.pkg"
 
   uninstall pkgutil: 'com.finkit.manictime.tracker'
-
-  caveats 'This is not a full ManicTime Windows port to Mac. '\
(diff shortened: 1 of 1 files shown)


waldyrious:
Uninstalling failed with the following output:

```
$ brew cask uninstall manictime
Warning: Unexpected method 'pkgutil' called on Cask manictime.
Follow the instructions here:
  https://github.com/caskroom/homebrew-cask/blob/master/doc/reporting_bugs/pre_bug_report.md

If this doesn’t fix the problem, please report this bug:
  https://github.com/caskroom/homebrew-cask#reporting-bugs

Warning: Unexpected method 'pkgutil' called on Cask manictime.
Follow the instructions here:
  https://github.com/caskroom/homebrew-cask/blob/master/doc/reporting_bugs/pre_bug_report.md

If this doesn’t fix the problem, please report this bug:
  https://github.com/caskroom/homebrew-cask#reporting-bugs
```

Any help appreciated.

Moisan:
You need `pkgutil:` instead of `pkgutil` to fix the uninstall.

waldyrious:
Thanks @Moisan. Shouldn't the audit or the style checker have caught this issue?

Moisan:
I am not deeply familiar with Homebrew audit and style checker but my feeling is that those will only make sure that the formula is properly formatted (e.g. correct indentation, all attributes needed are present), not that it will necessarily work at runtime.

waldyrious:
Yes, that makes sense, but since "pkgutil:" is a known keyword, I don't think it would be out of scope to check for the necessary punctuation. At the very least (IMHO), one of these checkers (or the uninstall command) should print out an informative error message that would lead cask authors towards the solution.

Casks/manictime.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  url "http://cdn.manictime.com/setup/mac/ManicTime-v#{version}.dmg"
+  name 'ManicTime'
+  homepage 'http://www.manictime.com/Mac'
+
+  pkg "ManicTime-#{version}.pkg"
+
+  uninstall pkgutil: 'com.finkit.manictime.tracker'
+
+  caveats 'This is not a full ManicTime Windows port to Mac. '\
```
gobinathm: Not sure if we need this caveat here.

When visited the software homepage there is clear indication of this message on the webpage. Also this message has nothing to do with how homebrew-cask is installing this software.

> >Sometimes there are particularities with the installation of a piece of software that cannot or should not be handled programmatically by Homebrew-Cask


Ref: [Caveat Doc](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md)

Casks/manictime.rb
```diff
@@ -0,0 +1,16 @@
+cask 'manictime' do
```
gobinathm: Since it's only a Client App, it might be good idea to use the token like ```manictime-client```

gobinathm (changes requested):
While uninstalling there are some residues left in Application Support folder. Hence that needs to be deleted as well

> > ~/Library/Application\ Support/ManicTime/

May be include

```
zap delete: '~/Library/Application Support/ManicTime'
```

Casks/manictime.rb
waldyrious: I'd rather keep it since people might then be confused that the app is not working as they expect (considering that the Windows client is fully-featured), but I understand the reasoning on the caveat guidelines.

Note, however, that the warning is quite visible in http://www.manictime.com/Mac, but not as much on http://www.manictime.com/mac/download, which is the page you get to by following the download links from the homepage.

That said, if you still think the caveat does more harm than good, let me know and I'll remove it.

gobinathm: @waldyrious personally i feel its not required, but i will defer the decision to Maintainers

Casks/manictime.rb
miccal: The token reference goes by the `name` stanza, so this is fine.

miccal:

waldyrious:
@gobinathm I'm sorry, I missed some of your review comments. I agree with you about the token, but the one I used was generated according to the instructions in [adding_a_cask.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md), which is why I didn't think of changing it.

As for the `zap delete`, that makes sense. Should I send a new PR adding that, @miccal?

gobinathm:
@waldyrious i guess new PR is not need for the token change, that can be ignored.

miccal:
@waldyrious you are welcome to open a new PR to add the `zap` stanza.

PR #31592, opened 2017-03-29
add caveat about expected fork for submitting a PR

waldyrious:
I ran into this issue when preparing a cask for submission, and thought it would be useful to add the note here so that other people don't have to face the same issue.

### DIFF
--- a/doc/development/adding_a_cask.md
+++ b/doc/development/adding_a_cask.md
@@ -294,6 +294,10 @@ Examples of difficult, unclear commit summaries:
 
 Push your changes to your GitHub account:
 
+(Note: this assumes you have cloned this repository
+and added it to your local clone
+under a remote named after your github account.)
+
 ```bash
 $ github_user='<my-github-username>'
 $ git push "$github_user" my-new-cask


gobinathm:
It's that how PR works in Github ? do we really need to add this ?

vitorgalvao:
Thank you, but that is not a prevalent issue (don’t remember it ever having happened before) and that is indeed true for every repository.

waldyrious:
> It's that how PR works in Github ?

> that is indeed true for every repository.

Could you expand a bit on what you mean by this? IIRC local clones do not automatically get two remotes unless made through a specific client (e.g. [hub](https://github.com/github/hub/)). A plain `git clone` only adds the default `origin` remote pointing to whatever url it was originally cloned from. Am I missing something?

vitorgalvao:
You’re not supposed to clone the upstream repo. You’re supposed to fork the repo to your account and then clone that one.

waldyrious:
@vitorgalvao I'm not claiming otherwise. Regardless of which repo one clones (the original or their fork), the upstream name in the local clone always defaults to "origin", not the person's github username.

gobinathm:
Article [How To Create a Pull Request on GitHub](https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github) articulates how to work with GITHUB pull request.

@vitorgalvao is this something we can ref in contributors guide ? thoughts ?

vitorgalvao:
[As I said](https://github.com/caskroom/homebrew-cask/pull/31592#issuecomment-290129608), doubts on this are rare, so it’s not worth adding to the documentation.

Most cask updates can be handled by `cask-repair` anyway, and to an extent that’s the recommended method.

waldyrious:
@gobinathm note how the article you linked to describes pushing to a remote named "origin", rather than the person's username: `$ git push --set-upstream origin new-branch`. They even have a [dedicated section for setting up a custom remote name](https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github#configure-a-remote-for-the-fork), but they use it to setup an "upstream" remote for the original project, rather than a "<github_username>" remote for the fork, which is what the caskroom documentation assumes.

That said, I can understand @vitorgalvao's argument that this problem appears to be rare -- still, I would prefer if this guide was self-contained, to avoid the unnecessary risk of discouraging contributors. I kindly ask the projects' maintainers to reconsider.

vitorgalvao:
> to avoid the unnecessary risk of discouraging contributors

Complete honesty: we have *too many contributors*, but not nearly enough where we need them most — the core itself. If you’re updating a cask, than the contribution is always welcome and you should be using `cask-repair` anyway. If you’re *adding* a cask, however, there’s always someone that knows how to use git and understands HBC enough to add it. **If such a person does not exist, the app isn’t that popular anyway**, and to me that is a wonderful deterrent.

We already have too many casks that nobody uses. How do I know that? Because occasionally we find casks that have been outdated for over two major versions, or have been dead for months without anyone noticing. Casks that nobody uses are unnecessary overhead, I’d much rather we have less casks that are well maintained.

That does not mean we do not care about our documentation. We do, and I wrote much of it and have at least been involved in nearly all of it. But one thing is HBC-specific documentation, and the other is git-specific documentation. The latter had its place in this repo, but is not beneficial anymore.

waldyrious:
I appreciate the context, and agree in principle with your reasoning. The thing is, this detail is not git-specific, since git's default behavior is using a remote named "origin", and HBC's documentation is deviating from that, which is why it should IMHO take the the onus of notifying users of the nonstandard expectations. That said, I won't insist further on this issue, since it doesn't seem to be that consequential anyway. Thank you for the patience and replies.

vitorgalvao:
Was checking the documentation to make the necessary fixes, and noticed the instructions are already there.

[At the top of `adding_a_cask`](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#adding-a-cask) we tell you to follow [these instructions](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#getting-set-up-to-contribute) that would have set you up correctly for the `Pushing` section.

waldyrious:
Ah, my bad, indeed I missed that -- and if as you say this issue hasn't been raised before, that probably means that the documentation is already clear enough that other people haven't missed the link. I'm sorry for taking your time, and thanks again for the patience.

vitorgalvao:
> I'm sorry for taking your time, and thanks again for the patience.

Thank you for being reasonable and polite throughout it all. An unreasonable amount of users would have lashed out and demanded the change like they were the queen of England. Those are the ones that burn maintainers out. I’m glad to say you’re on the positive side of that spectrum.

waldyrious:
I appreciate your words. That said (and speaking as a maintainer of a few other projects myself), steering discussions towards constructiveness takes effort on both sides, so I highly applaud your efforts toward that end as well :)

vitorgalvao:
Checked your github page and I recognise some of the projects. Also, pleasantly surprised to see you’re also Portuguese!

waldyrious:
Heheh in your case the name doesn't allow much room for guessing ;) um abraço!

PR #31598, opened 2017-03-29
android-sdk add symlink logic

yurikoles:
---
I had added symlink creation and remove of old content from homebrew. But I need your advice how to better handle old data on `"#{HOMEBREW_PREFIX}/opt/android-sdk`. I mean it's safe to remove it first time since there is old symlink from homebrew, but what to do if user makes an upgrade from previous version of this cask?

### DIFF
--- a/Casks/android-sdk.rb
+++ b/Casks/android-sdk.rb
@@ -7,6 +7,8 @@ cask 'android-sdk' do
   name 'android-sdk'
   homepage 'https://developer.android.com/index.html'
 
+  conflicts_with cask: 'android-platform-tools'
+
   build_tools_version = '25.0.2'
 
   binary "#{staged_path}/build-tools/#{build_tools_version}/aapt"
@@ -40,8 +42,16 @@ cask 'android-sdk' do
     system_command "#{staged_path}/tools/bin/sdkmanager", args: ['tools', 'platform-tools', "build-tools;#{build_tools_version}"], input: 'y'
   end
 
+  postflight do
+    # default unversioned path symlink
+    share_path = "#{HOMEBREW_PREFIX}/share/android-sdk"
+    FileUtils.ln_sf(staged_path.to_s, share_path)
+  end
+
   caveats <<-EOS.undent
     We will install android-sdk-tools, platform-tools, and build-tools for you.
     You can control android sdk packages via the sdkmanager command.
+    You may want to add to your profile:
+      'export ANDROID_HOME=#{HOMEBREW_PREFIX}/share/android-sdk'
   EOS
 end


Casks/android-sdk.rb
```diff
@@ -25,10 +25,14 @@
 
   postflight do
     system_command "#{staged_path}/tools/bin/sdkmanager", args: ['tools', 'platform-tools', 'build-tools;25.0.2'], input: 'y'
+    FileUtils.rm_rf ["#{HOMEBREW_PREFIX}/opt/android-sdk"]
+    File.symlink(staged_path.to_s, "#{HOMEBREW_PREFIX}/opt/android-sdk")
```
MikeMcQuaid: I don't think this should be using the HOMEBREW_PREFIX/opt; that's for Homebrew and not Cask. Creating another, unversioned directory symlink seems like a good idea, though.

victorpopkov:
This will need a rebase since https://github.com/caskroom/homebrew-cask/pull/31568 is merged.

yurikoles:
@victorpopkov done

yurikoles:
I had added conflict with `android-platform-tools`

victorpopkov:
@yurikoles Thank you for the rebase and `conflicts_with` addition.

However, I can't figure out a good solution to this problem. At least for now.

As @MikeMcQuaid has pointed out it's better not to use `#{HOMEBREW_PREFIX}/opt` path to store the symlink for the Cask. Most likely the `#{HOMEBREW_PREFIX}/share` will make more sense in order to avoid messing up with Homebrew directory structure.

I understand, however, that this will make the migration from Formula to Cask more painful, but I agree with @MikeMcQuaid. It's better not to use `#{HOMEBREW_PREFIX}/opt` for this. Another approach would be to adapt the cask to check whether the the Homebrew Formula has been used before and if yes replace the `#{HOMEBREW_PREFIX}/opt` symlink.

For example, you can always create an unversioned directory symlink in `#{HOMEBREW_PREFIX}/share` by default and if you is migrating from Homebrew also replace the old symlink with the new one:

```ruby
cask 'android-sdk' do
  version '25.2.3'
  sha256 '593544d4ca7ab162705d0032fb0c0c88e75bd0f42412d09a1e8daa3394681dc6'

  # google.com/android/repository/tools_r was verified as official when first introduced to the cask
  url "https://dl.google.com/android/repository/tools_r#{version}-macosx.zip"
  name 'android-sdk'
  homepage 'https://developer.android.com/index.html'

  conflicts_with cask: 'android-platform-tools'

  build_tools_version = '25.0.2'

  binary "#{staged_path}/build-tools/#{build_tools_version}/aapt"
  binary "#{staged_path}/build-tools/#{build_tools_version}/aapt2"
  binary "#{staged_path}/build-tools/#{build_tools_version}/aidl"
  binary "#{staged_path}/build-tools/#{build_tools_version}/apksigner"
  binary "#{staged_path}/build-tools/#{build_tools_version}/dexdump"
  binary "#{staged_path}/build-tools/#{build_tools_version}/dx"
  binary "#{staged_path}/build-tools/#{build_tools_version}/llvm-rs-cc"
  binary "#{staged_path}/build-tools/#{build_tools_version}/zipalign"
  binary "#{staged_path}/emulator/emulator"
  binary "#{staged_path}/emulator/emulator-check"
  binary "#{staged_path}/emulator/emulator64-arm"
  binary "#{staged_path}/emulator/emulator64-mips"
  binary "#{staged_path}/emulator/emulator64-x86"
  binary "#{staged_path}/platform-tools/adb"
  binary "#{staged_path}/platform-tools/dmtracedump"
  binary "#{staged_path}/platform-tools/etc1tool"
  binary "#{staged_path}/platform-tools/fastboot"
  binary "#{staged_path}/platform-tools/hprof-conv"
  binary "#{staged_path}/platform-tools/sqlite3"
  binary "#{staged_path}/tools/android"
  binary "#{staged_path}/tools/bin/avdmanager"
  binary "#{staged_path}/tools/bin/lint"
  binary "#{staged_path}/tools/bin/monkeyrunner"
  binary "#{staged_path}/tools/bin/sdkmanager"
  binary "#{staged_path}/tools/mksdcard"
  binary "#{staged_path}/tools/monitor"

  preflight do
    system_command "#{staged_path}/tools/bin/sdkmanager", args: ['tools', 'platform-tools', "build-tools;#{build_tools_version}"], input: 'y'
  end

  postflight do
    # default unversioned path symlink
    share_path = "#{HOMEBREW_PREFIX}/share/android-sdk"
    FileUtils.rm_rf [share_path]
    File.symlink(staged_path.to_s, share_path)

    # when migrating/migrated from Homebrew replace the old symlink
    opt_path = Pathname.new("#{HOMEBREW_PREFIX}/opt/android-sdk")
    if opt_path.exist?
      opt_dirname = opt_path.realpath.dirname
      if opt_dirname.to_s == "#{HOMEBREW_CELLAR}/android-sdk" || opt_dirname.to_s == caskroom_path.to_s
        FileUtils.rm_rf [opt_path]
        File.symlink(staged_path.to_s, opt_path)
      end
    end
  end

  caveats <<-EOS.undent
    We will install android-sdk-tools, platform-tools, and build-tools for you.
    You can control android sdk packages via the sdkmanager command.
    You may want to add to your profile:
      'export ANDROID_HOME=#{HOMEBREW_PREFIX}/share/android-sdk'
  EOS
end
```

Not the best solution, though. Also, it's a little bit hackish and stability should be concerned.

yurikoles:
@victorpopkov Could you push it as commit to this PR?

yurikoles:
@victorpopkov never mind, I had it already integrated and rebased.

Casks/android-sdk.rb
```diff
@@ -40,8 +42,27 @@
...
   end
 
+  postflight do
+    # default unversioned path symlink
+    share_path = "#{HOMEBREW_PREFIX}/share/android-sdk"
+    FileUtils.rm_rf [share_path]
+    File.symlink(staged_path.to_s, share_path)
+
+    # when migrating/migrated from Homebrew replace the old symlink
+    opt_path = Pathname.new("#{HOMEBREW_PREFIX}/opt/android-sdk")
```
MikeMcQuaid: I don't think this section is necessary; Homebrew will clean up its own files and should be responsible for doing so.

Casks/android-sdk.rb
```diff
@@ -40,8 +42,27 @@
     system_command "#{staged_path}/tools/bin/sdkmanager", args: ['tools', 'platform-tools', "build-tools;#{build_tools_version}"], input: 'y'
   end
 
+  postflight do
+    # default unversioned path symlink
+    share_path = "#{HOMEBREW_PREFIX}/share/android-sdk"
```
MikeMcQuaid: Maybe `android-home`?

Casks/android-sdk.rb
```diff
@@ -40,8 +42,27 @@
...
+    FileUtils.rm_rf [share_path]
```
MikeMcQuaid: Not sure how Homebrew Cask normally handles this but `rm_rf` without a prompt feels weird to me.

Casks/android-sdk.rb
yurikoles: @MikeMcQuaid but how homebrew will handle uninstall when formulae is gone from repository?

MikeMcQuaid: @yurikoles Formulae don't need to exist to be uninstalled or unlinked.

yurikoles:
I had replaced `FileUtils.rm_rf` by `FileUtils.ln_sf`

yurikoles:
bump?

victorpopkov (approved):
I think this should be merged since there is no better solution at the moment. Other PR are welcome if you know how this can be improved.

victorpopkov:
> It's hanging during install. Specifically, after it attempts to verify the checksum.

After verifying the checksum It's running the `sdkmanager` that with specified arguments as defined in preflight. It takes approximately up to 5 minutes to finish since it downloads and installs those tools. So if you quit it before it finishes, it won't link the binaries.

plastiv:
@victorpopkov btw, it really takes some time depending on internet connection. is it possible to redirect the output from sdkmanager? For e.g. with `--verbose` option user would see that something is happening

Alhadis:
@victorpopkov I left it running for at least 10 minutes, and my Internet is really slow. If a lengthy operation is happening, it needs to be brought to the user's attention.

victorpopkov:
> If a lengthy operation is happening, it needs to be brought to the user's attention.

@Alhadis I fully agree with you. There is definitely needs to be a message pointing that. You can submit PR to extend the current caveat with appropriate message:

```ruby
  caveats <<-EOS.undent
    We will install android-sdk-tools, platform-tools, and build-tools for you.
    You can control android sdk packages via the sdkmanager command.
    You may want to add to your profile:
      'export ANDROID_HOME=#{HOMEBREW_PREFIX}/share/android-sdk'

    This operation may take up to 10 minutes depending on your internet connection.
    Please, be patient.
  EOS
```

victorpopkov:
The lengthy operation notice added in https://github.com/caskroom/homebrew-cask/pull/31802.

PR #31573, opened 2017-03-29
Update docker to 17.03.1-ce

commitay:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.

PR #31599, opened 2017-03-29
Update docker

gobinathm:
- Since we are using an URL that points to latest version, updating the version/sha256 stanza

### DIFF
--- a/Casks/docker.rb
+++ b/Casks/docker.rb
@@ -1,6 +1,6 @@
 cask 'docker' do
-  version '17.03.1-ce'
-  sha256 '403b167dff238f5cbdb5fb085a842b8c8e1d6812956873f47df875a01adb0200'
+  version :latest
+  sha256 :no_check
 
   url 'https://download.docker.com/mac/stable/Docker.dmg'
   appcast 'https://download.docker.com/mac/stable/appcast.xml',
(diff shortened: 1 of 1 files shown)


Casks/docker.rb
```diff
@@ -1,6 +1,6 @@
 cask 'docker' do
-  version '17.03.1-ce'
-  sha256 '403b167dff238f5cbdb5fb085a842b8c8e1d6812956873f47df875a01adb0200'
+  version :latest
```
victorpopkov: Since there's is an appcast it's better to leave the cask versioned. However, I would recommend adding the build number to the version and reuse it in the download URL taken from the appcast to avoid sha256 mismatches in the future:
```ruby
  version '17.03.1-ce,16048'
  sha256 '403b167dff238f5cbdb5fb085a842b8c8e1d6812956873f47df875a01adb0200'

  url "https://download.docker.com/mac/stable/#{version.after_comma}/Docker.dmg"
```

gobinathm: @victorpopkov thats precisely the same thing i suggested in #31573. But i thought we did not want to do it based on the discussion there. NVM, i guess i misunderstood the point :)

Updating the PR 👍

victorpopkov (approved):

PR #31625, opened 2017-03-30
Fix incorrect sha256 for cask airvpn

n055:
See https://airvpn.org/macosx/


miccal:

PR #31714, opened 2017-03-31
Update torustrooper to 0.22.2

gobinathm:

### DIFF
--- a/Casks/torustrooper.rb
+++ b/Casks/torustrooper.rb
@@ -4,7 +4,7 @@ cask 'torustrooper' do
 
   url "https://workram.com/downloads/TorusTrooper-for-OS-X-#{version}.dmg"
   appcast 'https://workram.com/games/',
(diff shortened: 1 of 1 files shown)


Casks/torustrooper.rb
```diff
@@ -4,7 +4,7 @@
 
   url "https://workram.com/downloads/TorusTrooper-for-OS-X-#{version}.dmg"
   appcast 'https://workram.com/games/',
```
victorpopkov: I suggest removing the appcast from this cask since it just duplicates the homepage.

victorpopkov (changes requested):

Casks/torustrooper.rb
vitorgalvao: It duplicates the homepage, yes, but it can be used to check for updates, so it should stay. The `checkpoint` matters.

PR #31724, opened 2017-03-31
Adds Kite.app v0.20170331.0

deiga:


g-bon:
I think this should become kite and the current kite should be called kite-compositor because it's the complete name, and this is likely to be more popular.

"For unrelated Apps that share a name, the most popular one (usually the one already present) stays unprefixed."

deiga:
@victorpopkov Nice find, I didn't know they had appcast

victorpopkov:
@deiga You can use `find_sparkle_appcast` script to retrieve an appcast from the app, if you suspect that it has it.

You can find more info about it in [appcast.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) of the cask language reference.

victorpopkov:
The cask needs an update since the new version `0.20170331.0` has released.

@g-bon Personally, I agree with you. Maybe other maintainers have other suggestions.

vitorgalvao:
> Since the vendor name of the current `kite.rb` is "Rhodes Software", it also can be named `rhodes-software-kite.rb`.

That is the correct approach, though `rhodes-kite` should suffice. `kite-compositor` would break the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) rule.

Considering the amount of dislike I’ve already seen for this kite due to privacy/security concerns regarding the code and some shady tactics (replaced the code in a popular atom extension to use this without asking for consent), if I had to bet on one of them to run out of money and be discontinued, it’d be this one, not the current cask.

That said, I’d have no quarrel with renaming the other one. The token on this on is wrong, anyway

vitorgalvao:
@deiga Care to add the rename of the other one to this PR?

deiga:
@vitorgalvao Yeah, sure. No problem

vitorgalvao (approved):

PR #31733, opened 2017-03-31
update Bridge CC to 6.3.1

ghost:
according to this helpxsite bridge is currently at v6.3.1 whereas this formula still installs 6.0.0
https://helpx.adobe.com/bridge/using/whats-new.html
maybe this is by design, because I can not really see how one would change the formula to reflect the update?

if it was possible to update the version though, that would be really nice because I appreciate how you made this install process work without the rest of the CC hurdle 

thank you very much !


reitermarkus:
I don't think Adobe provides downloads for minor version, so updating is probably the only way to get to version 6.3.1.

vitorgalvao:
This shouldn’t be a PR, though.

PR #31736, opened 2017-04-01
Update deadbeef to latest

ibrokemypie:

### DIFF
--- a/Casks/deadbeef.rb
+++ b/Casks/deadbeef.rb
@@ -1,6 +1,6 @@
 cask 'deadbeef' do
-  version '1.0'
-  sha256 '57b1c473f4897e6ad2f20f81d53cf9441c1d50a23162d999d7eec7ee171141db'
+  version :latest
+  sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


Casks/deadbeef.rb
```diff
@@ -1,6 +1,6 @@
 cask 'deadbeef' do
-  version '1.0'
-  sha256 '57b1c473f4897e6ad2f20f81d53cf9441c1d50a23162d999d7eec7ee171141db'
+  version :latest
```
victorpopkov: As long as the cask has an appcast, the cask should be versioned. Even though the download URL points to latest. So there is no need to update this.

Appcasts exist precisely so we can detect when said changes happen (more scripts to do it are in the works).

ibrokemypie: should I just update the shasum then?

vitorgalvao: > should I just update the shasum then?

Update the version and let `cask-repair` do the rest.

PR #31775, opened 2017-04-02
migrate wine-devel to cask/versions

yurikoles:
see also https://github.com/caskroom/homebrew-cask/pull/31768#issuecomment-290995041, https://github.com/caskroom/homebrew-versions/pull/3568


vitorgalvao:
Thank you.

PR #31783, opened 2017-04-03
Update mqttfx.rb

### LINKED ISSUE #31686: `mqttfx` is now using an installer
More precisely, it’s using an app installer, meaning it will also need an `uninstall`.

miccal:

Closes https://github.com/caskroom/homebrew-cask/issues/31686.

### DIFF
--- a/Casks/mqttfx.rb
+++ b/Casks/mqttfx.rb
@@ -13,6 +13,6 @@ cask 'mqttfx' do
 
   zap delete: [
                 '~/Library/Application Support/MQTT-FX',
-                '~//Library/Application Support/MQTT.fx',
+                '~/Library/Application Support/MQTT.fx',
               ]
 end
--- a/Casks/mqttfx.rb
+++ b/Casks/mqttfx.rb
@@ -5,7 +5,14 @@ cask 'mqttfx' do
   # jensd.de/apps/mqttfx was verified as official when first introduced to the cask
   url "http://www.jensd.de/apps/mqttfx/#{version}/mqttfx-#{version}-macos.dmg"
   name 'MQTT.fx'
-  homepage 'http://mqttfx.jfx4ee.org/'
+  homepage 'http://www.mqttfx.org/'
 
-  app 'MQTT.fx.app'
+  installer manual: 'MQTT.fx Installer.app'
+
+  uninstall delete: '/Applications/MQTT.fx.app'
+
+  zap delete: [
+                '~/Library/Application Support/MQTT-FX',
+                '~//Library/Application Support/MQTT.fx',
+              ]
 end
--- a/Casks/mqttfx.rb
+++ b/Casks/mqttfx.rb
@@ -7,7 +7,11 @@ cask 'mqttfx' do
   name 'MQTT.fx'
   homepage 'http://www.mqttfx.org/'
 
-  installer manual: 'MQTT.fx Installer.app'
+  installer script: {
+                      executable: 'MQTT.fx Installer.app/Contents/MacOS/JavaApplicationStub',
+                      args:       %w[-q],
+                      sudo:       true,
+                    }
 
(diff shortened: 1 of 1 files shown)


Casks/mqttfx.rb
```diff
@@ -5,7 +5,14 @@
   # jensd.de/apps/mqttfx was verified as official when first introduced to the cask
   url "http://www.jensd.de/apps/mqttfx/#{version}/mqttfx-#{version}-macos.dmg"
   name 'MQTT.fx'
-  homepage 'http://mqttfx.jfx4ee.org/'
+  homepage 'http://www.mqttfx.org/'
 
-  app 'MQTT.fx.app'
+  installer manual: 'MQTT.fx Installer.app'
+
+  uninstall delete: '/Applications/MQTT.fx.app'
```
vitorgalvao: So this is the only thing the installer does?

miccal: I have installed it, run it and then `sudo find / -iname "*mqtt*"` to find everything, and this seems to be it. I also ran `pkgutil --pkgs` and `launchctl list` but they produced nothing.

miccal: On an unrelated note, what _is_ that a picture of in your avatar?

vitorgalvao: > Anything else I should try?

Ideally there would be some official document saying what is installed. Perhaps check `~/Library/LaunchAgents` and the list of apps that start with the system.

> On an unrelated note, what is that a picture of in your avatar?

(Part of) a broccoli and pumpkin. It’s [an art print I’ve made recently](https://society6.com/product/obri-maur_print) that I ended up liking more than I initially though. I like seeing it as an avatar, I find it fun.

miccal: > Ideally there would be some official document saying what is installed.

I looked for it, but could not find any.

> Perhaps check ~/Library/LaunchAgents and the list of apps that start with the system.

I check those also, there was nothing.

> (Part of) a broccoli and pumpkin.

It's very cute!

vitorgalvao: > I looked for it, but could not find any.
> I check those also, there was nothing.

I hate `app` installers. And if the only thing they do is move an app, they’re even worse. But it seems to have a `JavaApplicationStub`, so you might be able to do [this](https://github.com/caskroom/homebrew-cask/blob/16b26d3eda47d309b5a1e2b46cedbebbadb3bcf4/Casks/burp-suite.rb#L9L13).

> It's very cute!

Thank you!

miccal: > But it seems to have a `JavaApplicationStub`

Is there a trick to viewing this? When I open it in my editor, I just get gibberish.

numbermaniac: What if you try running it with a `--help` flag or similar? Does it list the options available?

miccal: > What if you try running it with a --help flag or similar? Does it list the options available?

No, just relaunches the manual installer.

vitorgalvao: > Is there a trick to viewing this?

Right click → Show Package Contents. Then Contents → MacOS.

I always See Package Contents on `installer manual:` apps, to see if there’s a script.

miccal: You misunderstood - I can see the file no problem, but I cannot view it in my editor.

miccal: `atom /Users/miccal/Downloads/MQTT.fx\ Installer.app/Contents/MacOS/JavaApplicationStub` opens a file of gibberish.

vitorgalvao: Ah, yes, that’s because it’s a compile executable.

Most (all?) `JavaApplicationStub` do follow that installation pattern, though, so it must be a standard command. Worth a try.

PR #31863, opened 2017-04-04
Update intellij-idea-ce.rb

yogendra:
Replacing uninstall 'delete:' key 'script:'
`/usr/local/bin/idea` file is not installed as part of this package.
This file is created by a user action in the application. As a result it could be present anywhere on path.
`uninstall delete:` key requires sudo, which may not be available.
Using uninstall script seems a better option.

### DIFF
--- a/Casks/intellij-idea-ce.rb
+++ b/Casks/intellij-idea-ce.rb
@@ -13,7 +13,12 @@ cask 'intellij-idea-ce' do
 
   app 'IntelliJ IDEA CE.app'
 
-  uninstall delete: '/usr/local/bin/idea'
+  uninstall script: {
+                      executable:   '/bin/rm',
+                      args:         [`which idea`.strip!],
+                      sudo:         false,
+                      must_succeed: false,
+                    }
 
   zap delete: [
                 "~/Library/Application Support/IdeaIC#{version.major_minor}",


vitorgalvao:
`uninstall script` is for uninstallation scripts the apps provide. Use `postflight` instead. Also verify first if `idea` exists and only then try to delete it.

Finally, shouldn’t the other Intellij casks benefit from this as well?

> `uninstall delete:` key requires sudo, which may not be available.

That true of any app, so it’s not a valid argument here.

yogendra:
Thanks for that. Let me check post flight. 

Is there a preferred way of checking for a file? Any existing cask I could check?

As per sudo part, I agree it could be a common problem. But, when we install the launcher from gui, we do not need sudo. In case of uninstall delete: there is no way to avoid sudo. Am I missing something?

Once this cask is fixed I will fix other intellj casks, including once that do not have this.

vitorgalvao:
I’ll do the first one and you do the others.

https://github.com/caskroom/homebrew-cask/pull/31867.

yogendra:
Ok. Thanks

vitorgalvao:
Ignore the failing test in that PR. It is correct, it’s just Travis that is having issues. I’m trying to fix those now. You can go ahead.

yogendra:
In #31870 , `File.rm(file)` wasn't working on my machine. I am using macOS system ruby 2.0.0. So, I changed it to `File.delete(file)`. 
Made changes to all 10 tools.

PR #31919, opened 2017-04-05
visual-studio: add articats to zap stanza

yurikoles:


Casks/visual-studio.rb
```diff
@@ -17,5 +17,7 @@
                 '~/Library/Logs/VisualStudio',
                 '~/Library/Preferences/Visual*Studio',
                 '~/Library/Application Support/CrashReporter/VisualStudio*',
+                '/Applications/Xamarin Workbooks.app',
```
vitorgalvao: This really warrants an explanation. Why are full apps here?

vitorgalvao (changes requested):

Casks/visual-studio.rb
yurikoles: This apps are installed by Visual Studio.

PR #31973, opened 2017-04-06
Update newmoon to 27.2.1

gobinathm:

### DIFF
--- a/Casks/newmoon.rb
+++ b/Casks/newmoon.rb
@@ -1,10 +1,14 @@
 cask 'newmoon' do
-  version '26.5.0-131'
-  sha256 '617581f55863b2ab0f55538791d2c82a06378febffebce14b8d1c1d625f465cb'
+  version '27.2.1'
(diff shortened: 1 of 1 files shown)


ibrokemypie:
Multiple issues with this PR.
The newmoon cask exists separately to the [palemoon cask](https://github.com/caskroom/homebrew-versions/blob/77338054b8ef4c31ea24c072cdb9ca3f72578af3/Casks/palemoon.rb) as there were major changes between versions 26 and 27 of palemoon which may render the application no longer suitable for certain usecases upon updating. The filename of the pre 27 palemoon was newmoon, and as such, the token was correct. The filename for palemoon 27+ is palemoon.app, which is why the token is correct there.
Another issue is the regression of the download link from https to http, why do this?
The palemoon cask is in versions and should have been been updated accordingly. This PR should be reverted.

miccal:
I cannot revert this for some reason - @ibrokemypie would you mind opening a new PR to fix this?

PR #31988, opened 2017-04-06
Update xnconvert to 1.74

tiiiecherle:

### DIFF
--- a/Casks/xnconvert.rb
+++ b/Casks/xnconvert.rb
@@ -1,6 +1,6 @@
 cask 'xnconvert' do
-  version :latest
-  sha256 :no_check
+  version '1.74'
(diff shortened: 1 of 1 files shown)


Casks/xnconvert.rb
```diff
@@ -1,6 +1,6 @@
 cask 'xnconvert' do
-  version :latest
-  sha256 :no_check
+  version '1.74'
```
victorpopkov: There is no need in this change. Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`.

victorpopkov:
The same thing applies to other PR as well. Thank you for your contributions, but there is no need for all those changes if the download URL doesn't point to any specific version.

PR #32002, opened 2017-04-06
Update copper to latest

gobinathm:
- App Update was available and the latest version is 1.3.1
- modified version to ```:latest``` and sha256 to ```:no_check``` as the download URL is not versioned. 


vitorgalvao:
When there’s an appcast, the cask must stay versioned. Please update it instead.

PR #32010, opened 2017-04-06
dropbox: remove appcast

neutric:
What hasn't been updated for months is now finally gone.

### DIFF
--- a/Casks/dropbox.rb
+++ b/Casks/dropbox.rb
@@ -1,9 +1,8 @@
 cask 'dropbox' do
-  version '23.4.17'
-  sha256 '1c92b57810a677243caa194b5c0edd383ba88f58f62852c08c1e4c8aec7fbc0f'
+  version :latest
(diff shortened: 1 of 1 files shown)


victorpopkov (approved):

vitorgalvao:
I propose we revert this cask back to an always up-to-date URL and `:latest`/`:no_check`.

Dropbox is one of those rare cases where having the latest version trumps having it versioned. We made it versioned due to the appcast, but even that only kind of worked. Without an appcast, this should go back to being unversioned.

gobinathm:
agreed with @vitorgalvao.

Also with _Drobbox Update_  there might not even a been to for this app to be upgrade / updated via homebrew as the app is going to self update itself periodically 

Ref# https://www.dropbox.com/help/9300

neutric:
Thanks for doing the necessary changes, @miccal! 🍒

PR #32054, opened 2017-04-07
Deleting Kismac Cask

gobinathm:
- Domain is not accessible any more

screen shot 2017-04-07 at 12 15 08 am
screen shot 2017-04-07 at 12 15 22 am


numbermaniac:
Looks like they still exist on GitHub: https://github.com/IGRSoft/KisMac2
but:
> Mac OS 10.7 - 10.10 (64-bit only)

miccal:
@numbermaniac it looks like it has not been updated in a while, so probably best to remove it.

PR #32064, opened 2017-04-07
Update unity to 5.6.0f3,497a0f351392

xuleman:

### DIFF
--- a/Casks/unity.rb
+++ b/Casks/unity.rb
@@ -1,6 +1,6 @@
 cask 'unity' do
-  version '5.5.2f1,3829d7f588f3'
-  sha256 '9f12dfa47058ce024d3e776a6a5847cd3d2cac6ff3be7d0001999d8569cf0b86'
+  version '5.6.0f3,497a0f351392'
+  sha256 'd46dee16c4c6d7edf22d2af1ff33d78884934c3c341836bd84fd6f8780e8e604'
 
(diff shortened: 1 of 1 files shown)


numbermaniac:
(to whichever maintainer looks at this): is there a reason why the `uninstall` stanza doesn't use `#{version.major}` instead of just `5`?

victorpopkov:
> (to whichever maintainer looks at this): is there a reason why the `uninstall` stanza doesn't use `#{version.major}` instead of `5`?

Nope, there is no reason.

mubeeniqbal:
The build is failing on Travis although the only change is the version and the checksum which I believe should be the case. Is there a specific reason as to why is this happening on Travis?

victorpopkov:
> The build is failing on Travis although the only change is the version and the checksum which I believe should be the case. Is there a specific reason as to why is this happening on Travis?

Nope, there is no specific reason here. At the moment, Travis CI results in fail even when there is nothing wrong with the changes in approximately 30% of cases. Build restart usually helps.

Why exactly it happens? I, personally, don't know. It needs some digging to find out the core reason behind that.

victorpopkov:
Closed in favour of https://github.com/caskroom/homebrew-cask/pull/32052 since it was made 1 hour earlier.

Thank you for all your contribution. We really appreciate them!

PR #32086, opened 2017-04-07
Rename formula: studio -> bricklink-studio latest

apetresc:
The cask name "studio" for [Bricklink Studio](https://studio.bricklink.com/v2/build/studio.page) is a bit misleading, since it's such a generic name that so many different products use. Nothing about the name "studio" implies "this is a LEGO program'. It is referred to as "Stud.io" on the Bricklink site, but that's because in that setting it is so clearly underneath the "Bricklink" umbrella. This setting is lost in the current name, and it's hard to find.

I propose renaming it to `bricklink-studio`.

### DIFF
--- a/Casks/bricklink-studio.rb
+++ b/Casks/bricklink-studio.rb
@@ -1,4 +1,4 @@
-cask 'studio' do
+cask 'bricklink-studio' do
   version :latest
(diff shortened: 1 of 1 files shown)


miccal:
Thank you.

PR #32128, opened 2017-04-09
Update installer.md

commitay:
This was going to be a issue but I thought opening a PR to ask / propose a change might be better.

Which style for install scripts with args/sudo is preferred?

The docs for [installer.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/installer.md#installer-script) show one style, but some casks have a style that matches the example for [uninstall.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#uninstall-key-script).

- [burp-suite](https://github.com/caskroom/homebrew-cask/blob/master/Casks/burp-suite.rb#L9-L13)
- [mqttfx](https://github.com/caskroom/homebrew-cask/blob/master/Casks/mqttfx.rb#L10-L14)
- [ransomwhere](https://github.com/caskroom/homebrew-cask/blob/master/Casks/ransomwhere.rb#L12-L16)

I'm not sure if there is any function difference but for readability, I personally prefer the uninstall style `{ }` scripts.

I've updated `installer.md` to use `ransomwhere` as an example if it is preferable.

### DIFF
--- a/doc/cask_language_reference/stanzas/installer.md
+++ b/doc/cask_language_reference/stanzas/installer.md
@@ -18,7 +18,7 @@ installer manual: 'Little Snitch Installer.app'
 
 | key             | value
 | ----------------|------------------------------
-| `script:`       | path to an install script to be run via `sudo`. (Required first key.)
+| `executable:`   | path to an install script to be run via `sudo`. (Required first key.)
 | `args:`         | array of arguments to the install script
 | `input:`        | array of lines of input to be sent to `stdin` of the script
 | `must_succeed:` | set to `false` if the script is allowed to fail
--- a/doc/cask_language_reference/stanzas/installer.md
+++ b/doc/cask_language_reference/stanzas/installer.md
@@ -24,10 +24,12 @@ installer manual: 'Little Snitch Installer.app'
 | `must_succeed:` | set to `false` if the script is allowed to fail
 | `sudo:`         | set to `true` if the script needs `sudo`
 
-The path may be absolute, or relative to the Cask. Example (from [adobe-air.rb](https://github.com/caskroom/homebrew-cask/blob/312ae841f1f1b2ec07f4d88b7dfdd7fbdf8d4f94/Casks/adobe-air.rb#L10-#L12)):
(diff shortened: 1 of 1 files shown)


miccal (approved):

vitorgalvao:
We should also update the docs for `installer`.

vitorgalvao (approved):

miccal:
> We should also update the docs for `installer`.

@vitorgalvao which docs are you referring to exactly?

vitorgalvao:
> @vitorgalvao which docs are you referring to exactly?

These ones. I got confused and thought the changes were being made somewhere else.

@commitay `executable:` should be added to the table.

doc/cask_language_reference/stanzas/installer.md
```diff
@@ -18,7 +18,7 @@ installer manual: 'Little Snitch Installer.app'
 
 | key             | value
 | ----------------|------------------------------
-| `script:`       | path to an install script to be run via `sudo`. (Required first key.)
```
vitorgalvao: `script:` should still stay. It’s valid, after all. Maybe we can take from the `uninstall` docs and follow the same pattern.

commitay: I'm not sure I understand as `install script:` is valid without `{ }` but `executable:` is used as the key within `{ }`.

commitay:
I've update all the `installer script:` I could find to `{ }`.

- https://github.com/caskroom/homebrew-drivers/pull/28
- https://github.com/caskroom/homebrew-versions/pull/3613
- #32145

- #32146 
- #32147 
- #32148

doc/cask_language_reference/stanzas/installer.md
commitay: We could add a second example,

If `installer script:` does not require any key-values it can point to the path of the install script. Example from [amazon-drive.rb](https://github.com/caskroom/homebrew-cask/blob/427c52acdc3ce0ab1e97950e6cee9896480d7353/Casks/amazon-drive.rb#L10):
```
installer script: 'Amazon Drive Installer.app/Contents/MacOS/Amazon Drive Installer'
```

vitorgalvao: > We could add a second example

Sounds good.

PR #32214, opened 2017-04-11
Update xpra to latest

gobinathm:

### DIFF
--- a/Casks/xpra.rb
+++ b/Casks/xpra.rb
@@ -2,7 +2,7 @@ cask 'xpra' do
   version :latest
   sha256 :no_check
 
-  url 'https://www.xpra.org/dists/osx/x86/Xpra.dmg'
+  url 'https://www.xpra.org/dists/osx/x86_64/Xpra-x86_64.pkg'
   name 'Xpra'
   homepage 'https://www.xpra.org/'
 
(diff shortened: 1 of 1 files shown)


Casks/xpra.rb
```diff
@@ -2,7 +2,7 @@
   version :latest
   sha256 :no_check
 
-  url 'https://www.xpra.org/dists/osx/x86/Xpra.dmg'
+  url 'https://www.xpra.org/dists/osx/x86_64/Xpra-x86_64.pkg'
```
victorpopkov: Since now it uses `pkg` instead of `dmg`, the installer should be changed to `pkg` ([pkg.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/pkg.md)) and be accompanied with the `uninstall` ([uninstall.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md)).

victorpopkov (changes requested):

Casks/xpra.rb
gobinathm: mm, that so bad of me to miss it. Will update.

gobinathm: @victorpopkov Its updated now.

PR #32452, opened 2017-04-17
Use HTTPS to download adobe-digital-editions

raindev:
- [x] The commit message includes the cask’s name ~~and version.~~

### DIFF
--- a/Casks/adobe-digital-editions.rb
+++ b/Casks/adobe-digital-editions.rb
@@ -2,7 +2,7 @@ cask 'adobe-digital-editions' do
   version '4.5.4'
   sha256 :no_check # required as upstream package is updated in-place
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Welcome, and thank you for the contribution!

Forgot to mention on twitter there’s another special case when cask download URLs are HTTP. When the developer does provide an HTTPS URL *but* it redirects to an HTTP one. In those cases we keep the HTTP one, so as to not give a false sense of security to users when it isn’t there.

We do have people who regularly run scripts over all casks to check for those cases, and also update the ones that were converted to true HTTPS with no redirects.

raindev:
Thank you! Checked for lack of redirects for this particular cask ✅

Issue #32518, opened 2017-04-18
sha256 mismatch in deadbeef 


wodim:
The [deadbeef cask](https://github.com/caskroom/homebrew-cask/blob/master/Casks/deadbeef.rb) points to

https://downloads.sourceforge.net/deadbeef/travis/osx/master/deadbeef-devel-osx-x86_64.zip

Look, I may not know much about so-called computahs, but that url seems to link to a snapshot of master, which changes every time I commit is made by the developers, so it'll be outdated forever. Either change the link or just remove the checksum check.

gobinathm:
Just wondering if we should make this cask to 
```
version: latest
sha256: no check
```

Multiple reasons

* #30607 changed the version to 1.0, but I'm not seeing any 1.0 on the official site. Not sure if I'm overlooking something
* After that we had 2 more commits which just update SHA256 & CHECKPOINT (#31221 & #31764) 

Any thoughts?  

I can update the PR if we have an agreement on this !!

MTCoster:
The latest release version is 0.7.2, although there are no binaries available for macOS at any version other than the nightlies already used by Cask.

The reason for this is hinted at on [the wiki](https://github.com/Alexey-Yakovenko/deadbeef/wiki/FAQ#os-x-port).  Should this cask possibly be removed?

gobinathm:
We can Take 2 approaches here

* Remove the Cask
* Change the version / sha256 so mismatch issue never occurs. 

I will defer the decision to maintainers.

vitorgalvao:
> although there are no binaries available for macOS at any version other than the nightlies already used by Cask.
> (…)
> Should this cask possibly be removed?

I’d be fine with removal.

yurikoles:
Why to remove? It just works.
I'll suggest move to `-versions` and change it to 

```
version: latest
sha256: no check
```

MTCoster:
It's not been deemed 'complete enough' to be released as a binary yet. If people want to use it anyways, they can still download a nightly build or build it from source themselves as per the project's Github page.

vitorgalvao:
> It's not been deemed 'complete enough' to be released as a binary yet.

That’s the argument that sells the removal, for me.

PR #32520, opened 2017-04-18
Delete deadbeef Cask

gobinathm:
- Removing deadbeef cask

---
Closes: #32518


gobinathm:
Just wondering if we should make this cask to version: latest & sha256: no check. Multiple reasons

* #30607 changed the version to 1.0, but I'm not seeing any 1.0 on the official site. Not sure if I'm overlooking something
* After that we had 2 more commits which just update SHA256 & CHECKPOINT (#31221 & #31764) 

Any thoughts?

PR #32561, opened 2017-04-19
Fix El Capitan support for SageMath 7.6

jdfm:

I'm on a tethered connection at home, the sage package is 1GB, I don't want to burn through my data limits this early in the month. I did have to download the package at least once (to generate the sha256 and to verify that the `SageMath-7.6.app` file existed in the `.dmg` file) and I did test that the link started to download with `brew cask install sage` with the new `sage.rb` file.
I accidentally used the project name of the application and not the cask name, I was unaware this would be problematic. The [documentation](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#commit-messages) spoke of the Application name, which I assume would have been SageMath in this case, but this condition makes it seem like it's the cask name (referred to as token in the documentation, it's sage in this case).

### DIFF
--- a/Casks/sage.rb
+++ b/Casks/sage.rb
@@ -1,10 +1,5 @@
 cask 'sage' do
-  if MacOS.version <= :lion
-    version '7.2'
-    sha256 'f88e0686ae8fe31f2684163a57bea938d93f732842cab7c263ee6e4cdeb271cc'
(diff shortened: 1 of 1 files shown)


miccal:
Verified locally.

PR #32608, opened 2017-04-19
Update siilo to 0.3.2

gtmgianni:


giannitm:
The appcast doesn't work because the TLS certificate is broken. Whenever `curl` tries to fetch the appcast, the web server redirects it automatically to the HTTPS version of the site, and it consequently fails with a certificate error.

PR #32683, opened 2017-04-20
Update sonoair to 1.0-beta-6.1

colindunn:
- [x] The commit message includes the cask\'s name and version.

### DIFF
--- a/Casks/sonoair.rb
+++ b/Casks/sonoair.rb
@@ -1,6 +1,6 @@
 cask 'sonoair' do
-  version '1.0-beta-4.0'
-  sha256 '2f14f818322f5bbf96833880edb5aee46cda5fa74456d7c951e3d36372c09067'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you for the continued correct and helpful contributions, @colindunn. It’s always a pleasure to see a contribution from you.

Which leads me to: why don’t you have a profile picture set? I ask because profile pictures help me greatly when reviewing multiple pull requests at once (which happens almost every time, considering we get an inordinate amount of them). They make it easy for my brain to identify the contributor at once and speed up the process.

If you prefer not having one, that is fine, but I thought I’d let you know they can be helpful.

colindunn:
I've only had the account for about 8 years @vitorgalvao why would I do that?!?

You raise a valid point though. Done.

vitorgalvao:
Thank you!

PR #32692, opened 2017-04-20
Update iridium-extra to 57.0

ghost:

### DIFF
--- a/Casks/iridium-extra.rb
+++ b/Casks/iridium-extra.rb
@@ -1,14 +1,14 @@
 cask 'iridium-extra' do
-  version '51.1.0'
-  sha256 'b0de307fef0bc1ac1123ac8f480c4a2178b6ba2849578f38d6c10dad35c04d6d'
+  version '57.0'
+  sha256 '3fec33a6426d1ee1a52371a721b4760fa8d0bd355cf5e0060f84201859210b66'
 
-  url "https://downloads.iridiumbrowser.de/macosx/#{version}/iridium_browser_extra_#{version}_osx_x64.dmg"
+  url 'https://downloads.iridiumbrowser.de/macosx/iridium_browser_osx_latest.dmg'
   appcast 'https://downloads.iridiumbrowser.de/macosx/',
-          checkpoint: '0a338c863d60192ea9b00df8b50aa90fc4b7d298b7e895284fab5b19d29888ca'
+          checkpoint: '8eecacd9cc8db0fca9979234468787faedff9a7a902f0740a23aa70a5cceca95'
   name 'Iridium Browser'
   homepage 'https://iridiumbrowser.de/'
 
-  app 'Iridium-extra.app'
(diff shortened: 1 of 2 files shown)


gobinathm (changes requested):
It's the App name have changed, it's better to update the CASK name as well to be consistent

Casks/iridium-extra.rb
```diff
@@ -1,14 +1,14 @@
 cask 'iridium-extra' do
```
victorpopkov: As @gobinathm has already suggested, since the app name has changed, the cask should also be renamed to `iridium.rb` to match the [token_reference.md](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md). The cask header should also be changed accordingly:

```ruby
cask 'iridium' do

```

victorpopkov (changes requested):

Casks/iridium.rb
```diff
@@ -1,14 +1,14 @@
-cask 'iridium-extra' do
-  version '51.1.0'
-  sha256 'b0de307fef0bc1ac1123ac8f480c4a2178b6ba2849578f38d6c10dad35c04d6d'
+cask 'iridium' do
+  version '57.0'
+  sha256 '3fec33a6426d1ee1a52371a721b4760fa8d0bd355cf5e0060f84201859210b66'
 
-  url "https://downloads.iridiumbrowser.de/macosx/#{version}/iridium_browser_extra_#{version}_osx_x64.dmg"
+  url 'https://downloads.iridiumbrowser.de/macosx/iridium_browser_osx_latest.dmg'
```
vitorgalvao: Are there no longer versioned downloads?

vitorgalvao (changes requested):

Casks/iridium.rb
gobinathm: we have versioned downloads available 

```
https://downloads.iridiumbrowser.de/macosx/
```

additionally, we also have the latest file available. Maybe it's good idea to move this cask to version: latest / sha256: no_check

vitorgalvao: > Maybe it's good idea to move this cask to version: latest / sha256: no_check

Why? Versioned URLs are always preferred.

gobinathm: We don't have any real appcast so if we leave it to version to latest & sha256  to nocheck. Then, We don't touch this cask until there are any change in other stanza's

```
"$(brew --repository)/Library/Taps/caskroom/homebrew-cask/developer/bin/find_sparkle_appcast"  ~/Desktop/Iridium.app
It appears this app does not have a Sparkle appcast
```

vitorgalvao: @gobinathm There is an appcast. But even if there was none, versioned `url`s are always preferred.

gobinathm: :thumbsup:

PR #32706, opened 2017-04-20
Update anaconda to use --installation-path option

### LINKED ISSUE #32699: Feature request: Add --installation-path option to anaconda formula
Hi.

### Description of feature/enhancement

Add an option to change the path the [Anaconda](https://github.com/caskroom/homebrew-cask/blob/master/Casks/anaconda.rb) installation script installs _itself_ to.

### Justification
By default the Anaconda installation script installs to ~/anaconda[3]. Users who don’t want software installed in their home directory, which is not meant to contain software, have no option to change this path atm. Luckily the install script offers to change this path by supplying a `-p PREFIX` option but it is not supported by the anaconda formula.

From Anaconda's usage message:

    -p PREFIX    install prefix, defaults to /Users/dirk/anaconda3

_Unfortunately I don’t know ruby and haven’t found documentation about adding options to formulas so I cannot do it myself._

### Example use case

Installing anaconda to `/opt/anaconda[3]` to avoid polluting one's home directory.

gobinathm:
- Move the installation to /usr/local instead of home directory

---
Closes #32699

### DIFF
--- a/Casks/anaconda.rb
+++ b/Casks/anaconda.rb
@@ -11,10 +11,11 @@ cask 'anaconda' do
 
   installer script: {
                       executable: "Anaconda3-#{version}-MacOSX-x86_64.sh",
-                      args:       ['-b'],
+                      args:       ['-b', '-p/usr/local/anaconda3'],
+                      sudo:       true,
                     }
 
-  uninstall delete: '~/anaconda3'
+  uninstall delete: '/usr/local/anaconda3'
 
   zap delete: [
                 '~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.continuum.io.sfl',
@@ -22,6 +23,7 @@ cask 'anaconda' do
               ]
 
   caveats do
-    path_environment_variable '~/anaconda3/bin'
+    path_environment_variable '/usr/local/anaconda3/bin'
+    files_in_usr_local
   end
 end
--- a/Casks/anaconda.rb
+++ b/Casks/anaconda.rb
@@ -11,7 +11,7 @@ cask 'anaconda' do
 
   installer script: {
                       executable: "Anaconda3-#{version}-MacOSX-x86_64.sh",
-                      args:       ['-b', '-p/usr/local/anaconda3'],
+                      args:       ['-b', '-p', '/usr/local/anaconda3'],
                       sudo:       true,
                     }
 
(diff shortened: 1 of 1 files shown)


gobinathm:
If Maintainers accept to move ahead with #32699

victorpopkov (approved):

reitermarkus:
Shouldn't we rather be using `HOMEBREW_PREFIX` here, instead of hard-coding `/usr/local`?

vitorgalvao:
> Shouldn't we rather be using `HOMEBREW_PREFIX` here, instead of hard-coding `/usr/local`?

Definitely.

gobinathm:
@reitermarkus @vitorgalvao Removed the hardcoding

Casks/anaconda.rb
```diff
@@ -11,17 +11,19 @@
 
   installer script: {
                       executable: "Anaconda3-#{version}-MacOSX-x86_64.sh",
-                      args:       ['-b'],
+                      args:       ['-b', '-p', "#{HOMEBREW_PREFIX}/anaconda3"],
+                      sudo:       true,
```
reitermarkus: @gobinathm, why is `sudo` needed now, but wasn't before?

wedi:
It’s not very flexible, but it solves the issue for me.

Casks/anaconda.rb
gobinathm: @reitermarkus Since its writing in /usr/local , elevated privilege is a must. Else the execution fails

vitorgalvao: @gobinathm Are you sure? Because `/usr/local`’s owner is the user (HB does that on install).

commitay: @vitorgalvao Install fails without `sudo`.
```
==> Running installer script Anaconda3-4.3.1-MacOSX-x86_64.sh
==> mkdir: /usr/local/anaconda3: Permission denied
==> ERROR: Could not create directory: /usr/local/anaconda3

==> Failed command:
#<Pathname:/usr/local/Caskroom/anaconda/4.3.1/Anaconda3-4.3.1-MacOSX-x86_64.sh> -b -p /usr/local/anaconda3
```

gnyman: Apologies if I'm reviving things unnecessarily, but I reacted on the sudo request when installing anaconda and looked into why as personally I think sudo should be avoided if possible.

After looking into it for some time, the obvious thing that stands out would be to install it in `/usr/local/Caskroom` ? 

Would that be a good idea? That folder seems to be designed to be used for things like this?

I can do a PR on that change if so. It might of course be a terrible idea I am very new to both anaconda and casks so I might have missed something important.

PR #32711, opened 2017-04-20
itsycal: Pin 0.10.16 for MacOS.version <= :el_capitan

mkhl:
Versions 0.11.X and later require at least :sierra,
see [the version history](https://www.mowglii.com/itsycal/versionhistory.html)

### DIFF
--- a/Casks/itsycal.rb
+++ b/Casks/itsycal.rb
@@ -5,6 +5,12 @@ cask 'itsycal' do
 
     # s3.amazonaws.com/itsycal was verified as official when first introduced to the cask
     url "https://s3.amazonaws.com/itsycal/Itsycal-#{version}.zip"
+  elsif MacOS.version <= :el_capitan
+    version '0.10.16'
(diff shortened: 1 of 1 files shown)


vitorgalvao:

Alternatively, you may wish to [use `cask-repair`](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask). It takes care of all of this for you.

Everything seems fine with this submission so I’ll go ahead and merge it, but do please fill the template next time.

mkhl:
> If you were not presented a template, you may have submitted via a CLI tool (likely hub). If that was the case, you may wish to see if the tool has a way to let you use the template (hub seems to have already implemented it, so you may be using an old version).

Yes, that’s what I did (use hub). FYI, the current version ([2.2.9](https://github.com/github/hub/releases/tag/v2.2.9)) does **not** include that feature though, it is only in the [2.3.0 series](https://github.com/github/hub/releases/tag/v2.3.0-pre1) which still [has not been released](https://github.com/github/hub/releases).

Thank you for taking the time to explain, I’ll take care to use the Web UI next time!

PR #32726, opened 2017-04-20
Add 8 discovered appcasts

colindunn:
* all appcasts have been stable for at least 7 days

---

### DIFF
--- a/Casks/alchemy.rb
+++ b/Casks/alchemy.rb
@@ -3,6 +3,8 @@ cask 'alchemy' do
   sha256 '7527f4e2231db8167c57d2639fcba166d8fefea091cca7884bd355fb52a3449a'
 
   url "http://al.chemy.org/files/Alchemy-#{version}.dmg"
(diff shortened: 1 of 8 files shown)


colindunn:
Just retested `brewtarget` on my end and still downloads and audits fine. If someone would mind doing a restart or a manual test of just that one?

Casks/alchemy.rb
```diff
@@ -3,6 +3,8 @@
   sha256 '7527f4e2231db8167c57d2639fcba166d8fefea091cca7884bd355fb52a3449a'
 
   url "http://al.chemy.org/files/Alchemy-#{version}.dmg"
+  appcast 'http://al.chemy.org/download/',
```
reitermarkus: Maybe http://al.chemy.org/download/version-history/ would be better here?

colindunn: Makes sense to me. Done

PR #32794, opened 2017-04-22
Update electron to 1.6.5

spaquet:


numbermaniac:
Looks like it was already 1.6.5 but then #32731 brought it _down_ to 1.3.15. I don't think that was meant to happen, so this should be good to go.

PR #32862, opened 2017-04-23
Update ostiarius to 1.2.0

gtmgianni:


vitorgalvao:
Nice find!

commitay:
@vitorgalvao Just FYI, I've found that the `changelog.txt` file for Objective-See apps is usually updated with the new version a week or two in advance of it actually being available for download.

vitorgalvao:
@commitay That’s quite a long time. Lets leave as is and revert later if we need.

PR #32893, opened 2017-04-23
mpv.rb: add binary

vitorgalvao:
The binary is a full-featured CLI app. The `app` bundle around it is just a simple wrapper and [they encourage direct usage of the binary](https://github.com/mpv-player/mpv/issues/4354#issuecomment-296448420).

### DIFF
--- a/Casks/mpv.rb
+++ b/Casks/mpv.rb
@@ -10,4 +10,5 @@ cask 'mpv' do
   homepage 'https://mpv.io/'
 
   app 'mpv.app'
+  binary "#{appdir}/mpv.app/Contents/MacOS/mpv"
 end


Akemi:
i don't know if i would call it encouraging but from https://github.com/mpv-player/mpv/commit/a5b97104cff358beebdb7ed1e96f7ab441295afa (fixed https://github.com/mpv-player/mpv/issues/2061) onwards the binary in the bundle will behave like the binary without the bundle. so i can only recommend using the binary in the bundle from 0.25 or that commit onwards.

vitorgalvao:
Thank you for the input, @Akemi! Welcome.

> i can only recommend using the binary in the bundle from 0.25 or that commit onwards.

Duly noted. I hope we can upgrade soon to 0.25, so that should not be a problem.

PR #32961, opened 2017-04-25
Add UniFi 5.4.14

ghost:


commitay:
Already exists in `caskroom/drivers` : [unifi-controller](https://github.com/caskroom/homebrew-drivers/blob/master/Casks/unifi-controller.rb)

PR #33015, opened 2017-04-26
Delete aliedit

gtmgianni:
Download link and cask homepage do not work for me. Can others confirm this, please?


giannitm:
Actually, this appears to be an issue with my VPN – closing pull request.

PR #33021, opened 2017-04-27
Update travis xcode to 8.3

commitay:
I just noticed this and I'm not sure if it should be updated or not.


commitay:
@vitorgalvao Should I do the other repos?

vitorgalvao:
@commitay [No need](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/sync_templates_and_ci), thank you.

PR #33117, opened 2017-04-28
Add "require 'open-uri'" to defer execution snippet

ebraminio:
It was needed for this work https://github.com/caskroom/homebrew-versions/pull/3752 so thought would be good idea to share it with others.


miccal:
Thank you.

PR #33154, opened 2017-04-29
Update beyerdynamicvs to 1.011

gtmgianni:


giannitm:
This isn't actually an appcast, it seems: https://north-america.beyerdynamic.com/virtual-studio.html

colindunn:
HTML pages that remain static are considered OK @giannitm

vitorgalvao:
> HTML pages that remain static are considered OK @giannitm 

Yep.

giannitm:
What I mean is that the page has no version information whatsoever on it, I think. @vitorgalvao

giannitm:
@vitorgalvao I don't think this should be used as an appcast because it doesn't look like it changes whenever new versions of the cask application are released.

giannitm:
That makes sense, thank you @vitorgalvao.

PR #33229, opened 2017-04-29
Update crypt to 3

gtmgianni:

### DIFF
--- a/Casks/crypt.rb
+++ b/Casks/crypt.rb
@@ -5,7 +5,7 @@ cask 'crypt' do
   # voluntary.net.s3.amazonaws.com was verified as official when first introduced to the cask
   url "http://voluntary.net.s3.amazonaws.com/Crypt#{version}_20100429.zip"
   appcast 'https://voluntary.net/crypt/',
-          checkpoint: '9754a5d7fca5ad8c9d5ad8543f574dcb4cd053b20de1b98c110698d28ea4488a'
+          checkpoint: '91ae34bf38675d39e737c284ff49d61190cf219840e0ff2a4161ea7cc4a1ab19'
   name 'Crypt'
   homepage 'https://voluntary.net/crypt/'
 


gobinathm:
>> vitorgalvaoVítor Galvão - commented 28 days ago
>I’ve merged it, but when the change is just the ```checkpoint```, please don’t submit it. There’s no point.

During my initial day of contribution to this repo, I was suggested not to submit any PR. But today I noticed a lot of PR's got merged with just checkpoint changes ie., similar to this one. *So just wondering if anything has changed? now it's allowed to push checkpoint changes?*  

Ref # https://github.com/caskroom/homebrew-cask/pull/31751#issuecomment-290937708

victorpopkov:
> So just wondering if anything has changed? now it's allowed to push checkpoint changes?

@giannitm, @gobinathm No, there is no point to submit PR separately to just change the checkpoint. The main reason behind that it simply creates a lot of unnecessary noise in the first place without actually making some useful changes to the casks.

I assume that the main reason why you are thinking that the idea of updating could be worth is that you can automate the finding of outdated casks using scripts. However, that is not the case. Many appcasts update even without the addition of the new version release: _change in old releases, the addition of the unstable release that doesn't belong in [caskroom/homebrew-cask](https://github.com/caskroom/homebrew-cask)_ and etc.

This results in the need of the checkpoint update each time in order to eliminate the number of wrong positives in the automated script that uses only the checkpoint to verify if the cask is outdated. But as soon as you manually verify that the latest appcast doesn't serve the new version for the current repository you will want to update the checkpoint again in order to skip the case in the script.

It could be reasonable if this would happen only a few times per month, but this happens a lot… and for many appcasts… This eliminates the usefulness of this kind of automation scripts that only trust the checkpoint. You will still need to manually verify the latest version of each outdated checkpoint each time and then submit another PR. You will get tired sooner or later including the maintainers.

giannitm:
@victorpopkov @gobinathm The scripts I'm using are not fully automated – the first one checks all casks for outdated `appcast checkpoint`s, then outputs all the casks which possibly need updating to a file. 

The next script will go through each cask in the file and use `brew cask audit --download $cask` to check it. If there is a problem (e.g. outdated appcast, broken appcast URL) it will output the error message. It then proceeds to open the cask.rb file, so I can manually check the version number. Next, it prints the raw data from the cask's appcast, so I can check manually to see if there is a version bump or if something else has changed the appcast (e.g. a beta release, which caskroom/homebrew-cask does not support).

Finally, I use `cask-repair` to submit the appcast change, and a version bump if necessary.

What I was thinking of doing, to prevent duplicate pull requests, is to put a list of all the casks which need to be checked, somewhere on the Internet. That way, any contributor can check the appcast themselves, and submit a pull request, then mark it off on the list. As of now, I haven't found a website/service that would let everyone edit the list I upload, but I'm still looking. (if you have a suggestion, please tell me).

I feel like updating these appcasts will not be wasted effort. Of the ~40 appcasts I updated 3 days ago, 2 or 3 of them changed the next day, and as a result, 2 of them were removed because they were not very good. Obviously we won't do all of the appcasts in one go – we'll go through them slowly so as not to wear ourselves out too much.

If later on we find that there are too many appcasts changing for us to keep on top of everything, then we will stop. 

If I'm just reiterating arguments that people have had in the past about appcasts, please fill me in because I'm only going off what I already know.

In summary, I feel that the `appcast` stanza is being wasted if we don't actually use them for checking for version bumps.

I have a few more ideas, but this comment is long enough (I apologise in advance) so I'll leave it there.

giannitm:
I've also just seen this pull request that was merged about 4 months ago:
https://github.com/Homebrew/brew/pull/1894

This may be why so many appcasts are showing up as outdated. Once we've gone through all of them, the workload will be much smaller.

vitorgalvao:
> What I was thinking of doing, to prevent duplicate pull requests, is to put a list of all the casks which need to be checked, somewhere on the Internet. That way, any contributor can check the appcast themselves, and submit a pull request, then mark it off on the list. As of now, I haven't found a website/service that would let everyone edit the list I upload, but I'm still looking. (if you have a suggestion, please tell me).

Which is my problem as well, and a big reason why the new scripts don’t yet exist.

@gobinathm You are indeed correct, I asked that of you. I told @giannitm the same, but then he mentioned he was using the scripts and I [recalled this point I had made](https://github.com/Homebrew/brew/pull/1894#issuecomment-274553902) which I forgot when I first talked to you.

Yes, `appcasts` should be fixed now. Apologies for the inducing you in error before.

PR #33266, opened 2017-04-30
Remove colorpicker.rb

gtmgianni:
https://www.northernspysoftware.com/ has been down for more than 2 days now. Could others confirm that `brew cask audit --download colorpicker` fails?


vitorgalvao:
Fails here too.

victorpopkov (approved):
Doesn't work on my side also.

PR #33274, opened 2017-05-01
Update kicad with proper artifact install location and add uninstall and zap stanzas

decrement:

### DIFF
--- a/Casks/kicad.rb
+++ b/Casks/kicad.rb
@@ -12,8 +12,10 @@ cask 'kicad' do
   artifact 'kicad', target: '/Library/Application Support/kicad'
 
   preflight do
-    FileUtils.mkdir "#{staged_path}/Kicad-apps"
-    FileUtils.mv Dir.glob("#{staged_path}/Kicad/*.app"), "#{staged_path}/Kicad-apps/"
+    FileUtils.cd staged_path do
+      FileUtils.mkdir 'Kicad-apps'
+      FileUtils.mv Dir.glob('Kicad/*.app'), 'Kicad-apps'
+    end
   end
 
   zap delete: '~/Library/Preferences/kicad'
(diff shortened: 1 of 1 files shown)


Casks/kicad.rb
```diff
@@ -9,10 +9,14 @@
   homepage 'http://kicad-pcb.org/'
 
   suite 'Kicad-apps', target: 'Kicad'
-  artifact 'kicad', target: "#{ENV['HOME']}/Library/Application Support/kicad"
+  artifact 'kicad', target: '/Library/Application Support/kicad'
```
vitorgalvao: Why?

decrement: The correct location of the `kicad` directory placement is `/Library/Application Support/`, as evident by the symlink/alias provided in the DMG. Furthermore, the `KISYSMOD` environmental variable defaults to `/Library/Application Support/` as well, which is used by kicad-extras to locate the offline footprints. Of course, the end-user is able to redefine `KISYSMOD`, but what I am proposing is what I think the KiCad team intended and most friendly for the end-user.

decrement:
The `brew cask uninstall kicad` did not remove the directory `/Library/Application Support/kicad/` without this line (that is why I included it):
```
uninstall delete: '/Library/Application Support/kicad'
```

vitorgalvao:
> The `brew cask uninstall kicad` did not remove the directory `/Library/Application Support/kicad/`

Just tested it and can confirm, but that is clearly a bug. `uninstall` even says:

```
==> Uninstalling Cask kicad
==> Removing App Suite '/Applications/Kicad'.
==> Removing Generic Artifact '/Library/Application Support/kicad'.
==> Using sudo to gain ownership of path '/Library/Application Support/kicad'
```

And then it doesn’t remove it. The solution is not to add an `uninstall`, but to fix the bug. Pinging @reitermarkus.

decrement:
I have updated the formula to use Ruby methods.

decrement:
I found the reason for the failure in removing the Generic Artifact; I have submitted a pull request: Homebrew/brew#2587

vitorgalvao (approved):

PR #33275, opened 2017-05-01
Add kicad-extras 4.0.6

decrement:

### DIFF
--- a/Casks/kicad-extras.rb
+++ b/Casks/kicad-extras.rb
@@ -18,7 +18,7 @@ cask 'kicad-extras' do
   end
 
   uninstall_preflight do
-    if File.directory?('/Library/Application Support/kicad/modules.github') == false
+    unless File.directory?('/Library/Application Support/kicad/modules.github')
       system_command '/bin/mv', args: ['--', '/Library/Application Support/kicad/modules', '/Library/Application Support/kicad/modules.github']
     end
   end
--- a/Casks/kicad-extras.rb
+++ b/Casks/kicad-extras.rb
@@ -1,0 +1,32 @@
+cask 'kicad-extras' do
+  version '4.0.6'
+  sha256 '06a9f9c7ad99dd82115011e63dd40e21519ccc433ccd091b363106afe45302aa'
+
+  url "http://downloads.kicad-pcb.org/osx/stable/kicad-extras-#{version}.dmg"
+  name 'KiCad Extras'
+  homepage 'http://kicad-pcb.org/'
+
+  depends_on cask: 'kicad'
+
+  artifact 'modules', target: '/Library/Application Support/kicad/modules'
+
+  preflight do
+    system_command '/bin/mv', args: ['--', '/Library/Application Support/kicad/modules', '/Library/Application Support/kicad/modules.github']
+    system_command '/bin/cp', args: ['-f',
+                                     '/Library/Application Support/kicad/template/fp-lib-table.for-pretty',
+                                     '/Library/Application Support/kicad/template/fp-lib-table']
+  end
+
+  uninstall_preflight do
+    if File.directory?('/Library/Application Support/kicad/modules.github') == false
+      system_command '/bin/mv', args: ['--', '/Library/Application Support/kicad/modules', '/Library/Application Support/kicad/modules.github']
+    end
+  end
+
+  uninstall_postflight do
+    system_command '/bin/mv', args: ['--', '/Library/Application Support/kicad/modules.github', '/Library/Application Support/kicad/modules']
+    system_command '/bin/cp', args: ['-f',
+                                     '/Library/Application Support/kicad/template/fp-lib-table.for-github',
+                                     '/Library/Application Support/kicad/template/fp-lib-table']
+  end
+end
--- a/Casks/kicad-extras.rb
+++ b/Casks/kicad-extras.rb
@@ -11,22 +11,24 @@ cask 'kicad-extras' do
   artifact 'modules', target: '/Library/Application Support/kicad/modules'
 
   preflight do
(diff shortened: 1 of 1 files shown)


vitorgalvao:
@adidalal I don’t remember exactly why this wasn’t included before, but I remember you were involved in the discussions. Mind taking a look?

adidalal:
https://github.com/caskroom/homebrew-cask/pull/15692 is the earlier thread - not sure if the problems raised there are still applicable as I haven't touched the Cask in quite a while

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,32 @@
+cask 'kicad-extras' do
+  version '4.0.6'
+  sha256 '06a9f9c7ad99dd82115011e63dd40e21519ccc433ccd091b363106afe45302aa'
+
+  url "http://downloads.kicad-pcb.org/osx/stable/kicad-extras-#{version}.dmg"
+  name 'KiCad Extras'
+  homepage 'http://kicad-pcb.org/'
+
+  depends_on cask: 'kicad'
```
vitorgalvao: [Same question](https://github.com/caskroom/homebrew-cask/pull/33274/files#r114116096).

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  artifact 'modules', target: '/Library/Application Support/kicad/modules'
+
+  preflight do
+    system_command '/bin/mv', args: ['--', '/Library/Application Support/kicad/modules', '/Library/Application Support/kicad/modules.github']
+    system_command '/bin/cp', args: ['-f',
```
vitorgalvao: Why?

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,32 @@
...
+  version '4.0.6'
+  sha256 '06a9f9c7ad99dd82115011e63dd40e21519ccc433ccd091b363106afe45302aa'
+
+  url "http://downloads.kicad-pcb.org/osx/stable/kicad-extras-#{version}.dmg"
+  name 'KiCad Extras'
+  homepage 'http://kicad-pcb.org/'
+
+  depends_on cask: 'kicad'
+
+  artifact 'modules', target: '/Library/Application Support/kicad/modules'
```
vitorgalvao: Why are you making this and then renaming it further down (`mv`) instead of doing it all at once?

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  preflight do
+    system_command '/bin/mv', args: ['--', '/Library/Application Support/kicad/modules', '/Library/Application Support/kicad/modules.github']
+    system_command '/bin/cp', args: ['-f',
+                                     '/Library/Application Support/kicad/template/fp-lib-table.for-pretty',
+                                     '/Library/Application Support/kicad/template/fp-lib-table']
+  end
+
+  uninstall_preflight do
```
vitorgalvao: Again, why? I don’t get this section at all. Why are you moving around and copying things that should just be deleted?

decrement:
The original problem raised was the `README.txt` file stated the end-user must run KiCad once before installing the extras. It was concluded that this was undesirable and no solution was found, so only `kicad`, and not `kicad-extras`, was accepted. I looked into the requirement and the reason is because the directory `~/Library/Preferences/kicad/` is created on the first run of KiCad. This was necessary so that the end-user had a place to copy the `fp-lib-table`, as called out in the `README.txt`.

The contents of `fp-lib-table` is actually the same as `fp-lib-table.for-pretty` located in `/Library/Application Support/kicad/template/`, and then I realized that the default `fp-lib-table` created on the first run of KiCad comes from this directory. You can see my *flight blocks in the formula that handles this neatly for the end-user. Now you can just run `brew cask install kicad kicad-extras` and have KiCad ready to go without pausing between to run KiCad once first.

Since I am on the topic of explaining the formula, the `uninstall_preflight` block is to help the end-user remove `kicad-extras` in the event the end-user accidentally removes `kicad` first (as this will remove the files that came with `kicad-extras` as well).

```
brew cask install kicad kicad-extras
brew cask uninstall kicad
```

In this instance, the end-use can perform the following steps:

```
brew cask install kicad
brew cask uninstall kicad-extras
brew cask uninstall kicad
```

Casks/kicad-extras.rb
decrement: The directory `/Library/Application Support/kicad/modules` already exists with content from the install of `kicad`. I am preserving it so if the end-user decides to uninstall `kicad-extras`, the original expected behavior (fetch footprints from GitHub) is reinstated.

Casks/kicad-extras.rb
```diff
@@ -0,0 +1,32 @@
...
+  url "http://downloads.kicad-pcb.org/osx/stable/kicad-extras-#{version}.dmg"
+  name 'KiCad Extras'
+  homepage 'http://kicad-pcb.org/'
+
+  depends_on cask: 'kicad'
+
+  artifact 'modules', target: '/Library/Application Support/kicad/modules'
+
+  preflight do
+    system_command '/bin/mv', args: ['--', '/Library/Application Support/kicad/modules', '/Library/Application Support/kicad/modules.github']
```
vitorgalvao: Also, instead of forking over to the system utilities, just use the Ruby methods. It can handle copying and moving just fine.

Casks/kicad-extras.rb
vitorgalvao: Got it. But please use ruby methods.

decrement:
I have updated the formula to use Ruby methods.

vitorgalvao (approved):

PR #33511, opened 2017-05-05
Update tap migrations - tomtom-mysports-connect

commitay:
This was renamed with https://github.com/caskroom/homebrew-drivers/pull/36. Not sure if the entry in `tap_migrations.json` needs to be removed or renamed.


vitorgalvao:
Let’s remove. Rename is a different beast.

PR #33524, opened 2017-05-06
Update gotomeeting to 8.4.6871

Conan1985:


miccal:
Since the `url` is un-versioned and there is no `appcast`, this should remain as `version :latest` and `sha256 :no_check`.

PR #33538, opened 2017-05-07
Add policy to better prevent malware.

reitermarkus:
Addresses https://github.com/caskroom/homebrew-cask/issues/33536.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,6 +5,8 @@ After making all changes to the cask:
 - [ ] `brew cask audit --download {{cask_file}}` is error-free.
 - [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
 - [ ] The commit message includes the cask’s name and version.
+- [ ] `sha256` was updated, but `version` remained the same.
+      [Post a link to a public confirmation by the developer.][version-checksum]:
 
 Additionally, **if adding a new cask**:
 
@@ -19,3 +21,4 @@ Additionally, **if adding a new cask**:
 [open pull requests]: https://github.com/caskroom/homebrew-cask/pulls
 [closed issues]: https://github.com/caskroom/homebrew-cask/issues?q=is%3Aissue+is%3Aclosed
 [the correct repo]: https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask
+[version-checksum]: https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256
--- a/doc/cask_language_reference/stanzas/sha256.md
+++ b/doc/cask_language_reference/stanzas/sha256.md
@@ -15,3 +15,13 @@ The special value `sha256 :no_check` is used to turn off SHA checking whenever c
 `version :latest` requires `sha256 :no_check`, and this pairing is common. However, `sha256 :no_check` does not require `version :latest`.
 
 We use a checksum whenever possible.
+
+## Updating the SHA256
+
+When updating the `sha256` stanza of an existing Cask, the `version` also has to have changed. Otherwise, the new checksum has to be confirmed by the developer.
+
+The confirmation of the updated `sha256` should ideally be publicly available. Specifically:
+
+ - If confirmed on Twitter, @mention [@homebrewcask](https://twitter.com/homebrewcask).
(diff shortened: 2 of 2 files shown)


mholt:
This is a great first step and I definitely approve (for what it's worth). :+1:

However, this is just someone ticking a box. Now might be a good time to start talking about a technical method for enforcing version+hash immutability. I don't have a good answer for Homebrew off-hand, at least, not through the pull request system...

vitorgalvao:
Yes, the thing we also need to add is a requirement for the person ticking the box to link to a public discussion where they can show they confirmed with the developer that the shasum changed.

@mholt You can’t enforce version+sha immutability. This problem did not occur due to human error and cannot be fixed by software. It was a correctly enforced bad policy, which is why the policy needs to change.

It’s **common**, especially for closed-source (but not only) software developers to update the upstream package in place without changing its version. So the vast majority of times the shasum changes but not the version are legitimate. There are cases where we do check with upstream, but not always (like in Handbrake’s case) due to it being so common. Making the version and the shasum immutable would mean you break multiple casks with no way of fixing them.

doc/development/updating_a_cask.md
```diff
@@ -0,0 +1,8 @@
+# Updating a Cask
```
vitorgalvao: This document is mostly a repeat of the PR_TEMPLATE, so I feel it’ll be ignored. I assume most documents are unread unless shoved in the user’s face (i.e. the templates).

So just like the [comment security measure](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) is in [`url.md`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md), I feel like the shasum warning can be in [`sha256.md`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md) and linked to from the PR_TEMPLATE.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,6 +5,7 @@ After making all changes to the cask:
 - [ ] `brew cask audit --download {{cask_file}}` is error-free.
 - [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
 - [ ] The commit message includes the cask’s name and version.
+- [ ] [If the checksum changed, the version also has to have changed.][version-checksum]
```
vitorgalvao: Should have a note asking for a link to a public proof.

vitorgalvao (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
vitorgalvao: Even better, in the longer rules we should mention that if it is asked on Twitter they should @mention [@homebrewcask](twitter.com/homebrewcask), and if on Github they should @mention one of us.

doc/cask_language_reference/stanzas/sha256.md
```diff
@@ -20,8 +20,8 @@ We use a checksum whenever possible.
 
 When updating the `sha256` stanza of an existing Cask, the `version` also has to have changed. Otherwise, the new checksum has to be confirmed by the developer.
 
-The confirmation of the updated `sha256` should ideally be publicly available. Specifically,
+The confirmation of the updated `sha256` should ideally be publicly available. Specifically:
 
- - if confirmed on Twitter, @mention [@homebrewcask](https://twitter.com/homebrewcask),
- - if confirmed on GitHub, @mention [one of the Homebrew-Cask maintainers](https://github.com/orgs/caskroom/people),
- - otherwise, post a link to the developer's confirmation.
+ - If confirmed on Twitter, @mention [@homebrewcask](https://twitter.com/homebrewcask),
```
reitermarkus: These should probably end with a dot now, since each point is a complete sentence.

vitorgalvao (approved):
Fine by me as it is @reitermarkus, but feel free to tweak some more.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,8 +5,8 @@ After making all changes to the cask:
 - [ ] `brew cask audit --download {{cask_file}}` is error-free.
 - [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
 - [ ] The commit message includes the cask’s name and version.
-- [ ] If the checksum changed, the version also has to have changed.
-      [Otherwise it has to be publicly confirmed by the developer.][version-checksum]
+- [ ] `sha256` was update, but `version` remained the same.
```
reitermarkus: *update**d**

PR #33634, opened 2017-05-08
Deleting quick radar cask

gobinathm:
- the homepage is not accessible

quickradar-2
quickradar-1


reitermarkus:
Thank you for the thorough research on this. 👍

PR #33664, opened 2017-05-09
Add/remove binaries in the emacs cask

midchildan:
- Remove `ctags` binary, since it conflicts with the [ctags formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ctags.rb) in homebrew-core.
- Add `emacs` binary.
- Add `conflicts_with formula: 'emacs'`

### DIFF
--- a/Casks/emacs.rb
+++ b/Casks/emacs.rb
@@ -9,6 +9,7 @@ cask 'emacs' do
   homepage 'https://emacsformacosx.com/'
 
   app 'Emacs.app'
+  binary "#{appdir}/Emacs.app/Contents/MacOS/Emacs", target: 'emacs'
   binary "#{appdir}/Emacs.app/Contents/MacOS/bin/emacsclient"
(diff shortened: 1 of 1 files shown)


miccal:
Thank you, but I think it is better to add `ctags` to `conflict_with` than remove it.

midchildan:
I think the installation of the `ctags` binary in this cask should not be a requirement, for two reasons:

- `ctags` is a very popular formula in homebrew-core (see https://brew.sh/analytics/install/), so introducing conflicts will affect a lot of people
- the [emacs formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/emacs.rb) in homebrew-core makes the installation of the `ctags` binary optional, causing consistency problems

vitorgalvao (approved):
I’m fine with it. Since I don’t use emacs, this is one of those cases where I prefer to defer to people who actually use it. If there’s disagreement later, we can revisit the best way to approach it.

PR #33692, opened 2017-05-09
PULL_REQUEST_TEMPLATE.md: updated message for sha256 change without version

vitorgalvao:
After seeing it live, it seemed a bit verbose/confusing still. Suggesting this small change.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,7 @@ After making all changes to the cask:
 Additionally, if **updating a cask**:
 
 - [ ] [If the `sha256` changed but the `version` didn’t][version-checksum],  
-      provide a link to public confirmation by the developer: {{link}}
+     provide a link to public confirmation by the developer: {{link}}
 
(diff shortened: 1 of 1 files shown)


reitermarkus (changes requested):
Maybe:

`Confirmation by developer: {{link}}`

giannitm:
@vitorgalvao @reitermarkus Just wondering, do you think it would be wise to add new tag for pull requests which still need to be verified (e.g. https://github.com/caskroom/homebrew-cask/pull/33570). Or will the "on hold" tag suffice?

vitorgalvao:
@giannitm Currently, it should be `on hold` + `upstream`. By I wouldn’t oppose a new label.

vitorgalvao:
@reitermarkus How about like this? We don’t need to repeat it’s a link. The amount of earlier commits was me trying to align the text, but apparently Github always has that small indentation.

reitermarkus (approved):

reitermarkus:
> The amount of earlier commits was me trying to align the text, but apparently Github always has that small indentation.

That was also driving me nuts yesterday. 😂

PR #33812, opened 2017-05-10
Update iboostup to 5.6.1

gobinathm:

### DIFF
--- a/Casks/iboostup.rb
+++ b/Casks/iboostup.rb
@@ -4,7 +4,7 @@ cask 'iboostup' do
 
   url 'https://www.iboostup.com/iboostup.dmg'
   appcast 'https://www.iboostup.com/updates',
-          checkpoint: 'ef359e1ee2d461e285517fc50650db354234423fb5a5a70b3dd0e81bae19414d'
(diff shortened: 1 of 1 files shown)


gtmgianni (approved):

giannitm:
@vitorgalvao Just out of curiosity, is it beneficial for contributors to review other's pull requests, or is it not really necessary?

gobinathm:
@giannitm based on my experience its good to check the PR which has failed Travis builds.

giannitm:
@gobinathm Fair point.

vitorgalvao:
There are three main instances when I review with Github’s feature:

+ ✘ There are changes that need to be made.
+ ✔ Everything looks fine but I’m not merging yet because Travis hasn’t finished.
+ ✔ Everything looks fine but I’d like for another maintainer to check/merge. This usually happens when the PR suffered a lot of changes and this is the final “alright, everything looks fine but I’d like another pair of eyes in case I missed something amidst all the changes”.

There’s no point for contributor reviews in the second or third cases, since they’ll still need to be signed off. The first case is useful, though, for two reasons:

+ You may catch something we didn’t.
+ If you start the conversation early with the user, the issue might already be solved when the maintainer arrives, which saves a lot of time and helps _tremendously_.

PR #34007, opened 2017-05-13
Update linein to 2.3.0

gtmgianni:

### DIFF
--- a/Casks/linein.rb
+++ b/Casks/linein.rb
@@ -9,4 +9,8 @@ cask 'linein' do
   homepage 'https://www.rogueamoeba.com/freebies/'
 
   app 'LineIn.app'
+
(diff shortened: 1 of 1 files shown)


giannitm:
Not sure if this should be moved to caskroom/homebrew-versions.

reitermarkus:
This implies that it is discontinued.

```ruby
  caveats do
    discontinued
  end

```

reitermarkus:
I think there was another app by RogueAmoeba that you updated with the legacy link. Could you add the caveat there as well? Thanks!

reitermarkus:
I think there was another app by RogueAmoeba that you updated with the legacy link. Could you add the caveat there as well? Thanks!

PR #34088, opened 2017-05-14
Update moom to 3.2.6

gtmgianni:

### DIFF
--- a/Casks/moom.rb
+++ b/Casks/moom.rb
@@ -2,9 +2,9 @@ cask 'moom' do
   version '3.2.6'
   sha256 '482dd69788ac707f13483631eae5aaef7fcc356b688d51e359a3b69fc531a143'
 
-  url 'https://manytricks.com/download/moom'
-  appcast 'https://manytricks.com/moom/appcast.xml',
-          checkpoint: 'ea75ccd05a0a5ffbff3da7692f5fb60315b6a13dc055a44562c07c38588edb41'
+  url "https://manytricks.com/download/_do_not_hotlink_/moom#{version.no_dots}.dmg"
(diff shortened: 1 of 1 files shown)


Casks/moom.rb
```diff
@@ -2,9 +2,9 @@
   version '3.2.6'
   sha256 '482dd69788ac707f13483631eae5aaef7fcc356b688d51e359a3b69fc531a143'
 
-  url 'https://manytricks.com/download/moom'
-  appcast 'https://manytricks.com/moom/appcast.xml',
-          checkpoint: 'ea75ccd05a0a5ffbff3da7692f5fb60315b6a13dc055a44562c07c38588edb41'
+  url "https://manytricks.com/download/_do_not_hotlink_/moom#{version.no_dots}.dmg"
```
vitorgalvao: I think that maybe they might not want this in HBC. We should ask.

vitorgalvao (changes requested):

giannitm:
@vitorgalvao I was wondering if that would be an issue, but I'm not sure if that will apply for HBC. I interpreted the message to be targeting sites which offer file downloads, but direct the download request to the Manytricks server, while still earning money themselves from advertising revenue. That is the true spirit of "hotlinking".

gobinathm:
many appcast have different download URL's compare to the once that's found on their website.  They might be doing it specifically to track downloads (or) something for analytics purpose. As @vitorgalvao it worth checking that, instead of we deciding it

giannitm:
@gobinathm Good point, didn't think of that.

giannitm:
New version of Moom has just been released – I'm making another pull request. @vitorgalvao

PR #34158, opened 2017-05-14
Process monitor 1.1

timsutton:

This is a new one. The public download URL listed on the website and the one used in the Appcast feed are the same, and not versioned. It seems possible then that future updates will simply use the same URL and the feed would be updated to have only the latest item.

For this reason it seems also reasonable to use `version :latest` instead of `1.1`, but I wasn't sure. Since this is so new there's not an established pattern for how this will get updates.

### DIFF
--- a/Casks/process-monitor.rb
+++ b/Casks/process-monitor.rb
@@ -1,6 +1,6 @@
 cask 'process-monitor' do
   version '1.1'
-  sha256 :no_check
+  sha256 'ef074c433a75ab347856b26a7fa0568558c2b938d7b531c12cd53ae1ddfebea8'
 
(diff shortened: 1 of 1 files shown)


Casks/process-monitor.rb
```diff
@@ -0,0 +1,16 @@
+cask 'process-monitor' do
+  version '1.1'
+  sha256 :no_check
```
vitorgalvao: I understand your reasoning, but in this situation the `sha256` is to be set anyway.

vitorgalvao (changes requested):

vitorgalvao (approved):

PR #34349, opened 2017-05-18
Update plex-media-server to 1.6.1.3722-4955e31cf

vin047:

### DIFF
--- a/Casks/plex-media-server.rb
+++ b/Casks/plex-media-server.rb
@@ -1,6 +1,6 @@
 cask 'plex-media-server' do
-  version '1.5.6.3790-4613ce077'
-  sha256 'a94b39d162029396eb7afa9131e1209cdf2bceb8296c0b9859806b31e469ef35'
+  version '1.6.1.3722-4955e31cf'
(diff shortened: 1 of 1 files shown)


Casks/plex-media-server.rb
```diff
@@ -1,6 +1,6 @@
 cask 'plex-media-server' do
-  version '1.5.6.3790-4613ce077'
-  sha256 'a94b39d162029396eb7afa9131e1209cdf2bceb8296c0b9859806b31e469ef35'
+  version '1.6.1.3722-4955e31cf'
```
commitay: @vin047 Is `1.6.1` publicly available or plex pass only?

https://www.plex.tv/downloads/ is still at `1.5.6` and the `appcast` hasn't been updated.

vin047: You're right, its a plex pass download. Thought that was the version that was being tracked in Cask though? At least, i'm pretty sure it used to be...

commitay: > You're right, its a plex pass download

Will `1.6.1` work without plex pass?

miccal: We can only use the one available via the `homepage`, which as @commitay pointed out is still at `1.5.6`.

PR #34358, opened 2017-05-18
spatial: latest - After installation, set the command structure to v2 which is current.

petemounce:
This eliminates a manual step on our setup guide for new users: https://spatialos.improbable.io/docs/reference/10.3/setup-spatialos/mac#2-set-up-the-spatialos-cli

- [x] [If the `sha256` changed but the `version` didn’t][version-checksum],  
      provide public confirmation by the developer:

  _I'm making the contribution from my company's fork._

n/a

### DIFF
--- a/Casks/spatial.rb
+++ b/Casks/spatial.rb
@@ -10,4 +10,9 @@ cask 'spatial' do
   container type: :naked
 
   binary 'mac', target: 'spatial'
+
+  postflight do
+    puts 'Setting cli-structure to v2 (current).'
+    `spatial config set cli-structure v2`
+  end
 end


vitorgalvao:
Welcome, @petemounce.

Going to pass on this. For one, the style on your `postflight` is vastly different from the style in every other `postflight`. It’s not consistent at all, and looks like a hack.

More importantly, though, it’s running a non-mandatary-for-correct-installation command on behalf of the user, which we don’t do. We try to impose the least on users. See [our FAQ on the matter](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md).

petemounce:
@vitorgalvao thanks for taking a look; I wasn't aware of the FAQ on the matter.

PR #34479, opened 2017-05-20
Update traktable to 0.6-alpha1,a

gtmgianni:

### DIFF
--- a/Casks/traktable.rb
+++ b/Casks/traktable.rb
@@ -1,11 +1,11 @@
 cask 'traktable' do
-  version '0.5.2'
-  sha256 '77d3800f02f2b5a2aef8a56e7986b731831738673e35c633d743df87bc556225'
+  version '0.6-alpha1,a'
+  sha256 '6bf5b2e30e8378d46e5c69494bc23617a3f410856ee4682899683607cb0dee79'
 
   # github.com/yo-han/Traktable was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


giannitm:
Is there a better way of implementing this? The way I've done it looks a bit messy in my opinion.

colindunn:
Only other way I can think of is to truncate the version on the first entry which is just as ugly. Or if this is a rare one, then don't bother at all and just add the `a` into the download url to be removed later because you'll be in the same position later when it gets removed and you have to remove the comma because there isn't anything after the 1.

Having said that, if this is an alpha, should it even be updated to that version?

vitorgalvao:
> Having said that, if this is an alpha, should it even be updated to that version?

Latest stable is over three years old, and this alpha is over two years old and [reportedly does not have many changes](https://github.com/yo-han/Traktable/releases/tag/0.6-alpha1). I guess this falls into the [unstable being the standard rule](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

I agree the current `version`/`url` is a bit messy, but I also doubt this app will get any updates soon (or at all). We can deal with it when it happens.

vitorgalvao:
Thank you for the contribution. It was merged directly as efb1a2c33a450909ec78d39444906cea09197c13. Your contribution is still credited to you.

PR #34740, opened 2017-05-22
Update dbeaver-enterprise to 4.0.5

gtmgianni:

### DIFF
--- a/Casks/dbeaver-enterprise.rb
+++ b/Casks/dbeaver-enterprise.rb
@@ -3,8 +3,6 @@ cask 'dbeaver-enterprise' do
   sha256 'c58b03056e0a845f2d7fff7ca37dcdf8b4dc2ded74e1932bc2654f8676510bf0'
 
   url "http://dbeaver.jkiss.org/files/#{version}/dbeaver-ee-#{version}-macos.dmg"
-  appcast 'http://dbeaver.jkiss.org/files/',
-          checkpoint: 'ffef67bd5da18feb5a7143c0365679819ee456b260ec7ccec9b1b9199cef5bd9'
   name 'DBeaver Enterprise Edition'
   homepage 'http://dbeaver.jkiss.org/'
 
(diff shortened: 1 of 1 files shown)


colindunn:
This appcast has been updated 3 times in the last 20 days without an application update. It's time to consider removing the appcast.

miccal:
@colindunn I agree, `appcast` removed.

commitay:
@miccal There are several casks like this with either development or alternate versions released out of sync on the same `appcast`, [brave](https://github.com/caskroom/homebrew-cask/pulls?utf8=%E2%9C%93&q=is%3Apr%20brave%20) is one that has had the `appcast` updated several times without a `version` bump. Should we consider removing those `appcast`s as well?

miccal:
@colindunn I will let @vitorgalvao make that call.

vitorgalvao:
`brave` uses a github appcast, as does atom and many others. It did indeed get version bumps, it’s just that those bumps were not made to the stable version. As such, the `appcast` is still valuable.

We can safely say `appcast` is useful. What we need now is a type of [`appcast do`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#using-a-block-to-defer-code-execution) that will let us be granular about what to ignore/keep from the `appcast`.

Maybe we even change the whole system and all appcasts become blocks that output a string that we can transform anyway we want. That way we’ll even be able to parse webpages that frequently change and just get the piece we want. Then we compare the result from the appcast to something else (like the previous value, or `version`), and use that to determine if an update occurred.

I’ve add a system like this in mind for a long time, but not to be included in casks. It’s basically be an external system that would then leverage `cask-repair` to automate the updates. The advantage would be we could output a common format that other apps/users would be able to take advantage of as well.

colindunn:
Sorry @vitorgalvao I missed this comment somehow.

Love the idea of an `appcast do` or similar but agree that it doesn't really seem to belong in the casks themselves. The idea of an external shared system of some kind would be great here too and probably more appropriate.

When I have a bit more spare time I'd be happy to open an issue on this one for discussion. I had already promised I would share some of my update finding methods so will wrap it in there as well.

vitorgalvao:
> When I have a bit more spare time I'd be happy to open an issue on this one for discussion. I had already promised I would share some of my update finding methods so will wrap it in there as well.

I’d appreciate that. I’d like to read your insights before working on it.

PR #34762, opened 2017-05-22
Remove appcast from objectivesharpie

colindunn:
* Appcast has changed 3 times over the last week. Removing as it is ineffective.

---


vitorgalvao:
It’s not so much ineffective, as it’s shared amongst a lot of xamarin apps. Unsure what to do about this one. We’d have to remove it from all xamarin casks.

colindunn:
Fair point @vitorgalvao.
Basically this one comes up for me every time when I do my scans which means that I end up effectively ignoring it. Happy either way, will figure something else out then.

PR #34866, opened 2017-05-25
Tweak wording of readme for clearer definition of correct repo.

sgleadow:
I submitted a cask for an EAP version this morning to homebrew-versions in caskroom/homebrew-versions#3897, but that was incorrect and I moved to caskroom/homebrew-cask#34865

I was confused by the wording that said in the situation where the app is only available in beta, it _can_ go in the main repo.

I took this to mean it _can_ but it doesn't have to, so submitted there anyway.

This PR tries to make the wording to make it clear in this case it _should_ go to the main repo, and _not_ in the versions repo.

I hope that helps, it could just be user error on my part!

### DIFF
--- a/doc/development/adding_a_cask.md
+++ b/doc/development/adding_a_cask.md
@@ -202,7 +202,7 @@ Stable versions live in the main repository at [caskroom/homebrew-cask](https://
 
 ### But There Is No Stable Version!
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you anyway but this is not really mandatory — it’s a guideline. It’s still up to maintainer discretion. Current wording is correct. This is not a prevalent issue.

PR #34879, opened 2017-05-25
Remove natron.rb

colindunn:
* The download url requires a login now, which means install fails. Unable to find a download link that does not require login so removing

---


colindunn:
Never mind. Retested and it's working now.

PR #34946, opened 2017-05-26
Update things 3.0.2

phatblat:


svenax:
This cask seems pretty useless, since it only installs a 7 day demo version. The real version is only available from the App store.

vitorgalvao:
@svenax Agreed. I thought you could buy direct from them but that seems to no longer be the case.

ldgmini:
@svenax It's a 15-day trial, but you have a good point. I guess they ditched the non-MAS licensing with 3.0.

PR #34948, opened 2017-05-26
Add transmit-disk 4.4.12

phatblat:

## Notes 

Transmit Disk is a helper app for Transmit which is separate due to Mac App Store sandboxing restrictions. The helper app is necessary to mount a remote server as a disk in the Finder. I believe it uses [FUSE](https://osxfuse.github.io) to do this as it installs a 2nd bundle to `/Library/Filesystems/transmitdisk.fs`.

The `com.panic.TransmitDisk.pkg` package installs `Transmit Disk.app` into `/tmp/` which doesn't seem like a good permanent home.

I tried setting `customLocation` in choices, which worked to put the .app in `/Applications`, but the installer validation failed. So, instead this uses a `postflight` move the .app into `/Applications`, paired with an `uninstall`/`trash` of the .app.

### DIFF
--- a/Casks/transmit-disk.rb
+++ b/Casks/transmit-disk.rb
@@ -6,40 +6,11 @@ cask 'transmit-disk' do
   name 'transmit-disk'
   homepage 'https://panic.com/transmit/'
 
-  depends_on cask: 'transmit'
-
-  pkg "Transmit Disk #{version}.pkg",
-      choices: [
-                 {
-                   'choiceIdentifier' => 'com.panic.TransmitDisk.pkg',
-                   'choiceAttribute'  => 'visible',
-                   'attributeSetting' => false,
-                 },
-                 {
-                   'choiceIdentifier' => 'com.panic.TransmitDisk.pkg',
-                   'choiceAttribute'  => 'enabled',
-                   'attributeSetting' => true,
-                 },
-                 {
-                   'choiceIdentifier' => 'com.panic.TransmitDisk.pkg',
-                   'choiceAttribute'  => 'selected',
-                   'attributeSetting' => 1,
-                 },
-                 # FIXME: Puts .app in correct place but fails package validation
-                 # {
-                 #   'choiceIdentifier' => 'com.panic.TransmitDisk.pkg',
-                 #   'choiceAttribute'  => 'customLocation',
-                 #   'attributeSetting' => '/Applications',
-                 # },
-               ]
+  pkg "Transmit Disk #{version}.pkg"
 
   postflight do
-    system_command 'mv',
-                   args: [
-                           '/tmp/Transmit/Transmit Disk.app',
-                           '/Applications/',
-                         ],
-                   sudo: true
+    set_ownership ['/tmp/Transmit']
+    FileUtils.move('/tmp/Transmit/Transmit Disk.app', "#{appdir}/Transmit Disk.app")
   end
 
   uninstall quit:    [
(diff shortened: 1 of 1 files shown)


Casks/transmit-disk.rb
```diff
@@ -0,0 +1,61 @@
...
+                 # FIXME: Puts .app in correct place but fails package validation
+                 # {
+                 #   'choiceIdentifier' => 'com.panic.TransmitDisk.pkg',
+                 #   'choiceAttribute'  => 'customLocation',
+                 #   'attributeSetting' => '/Applications',
+                 # },
+               ]
+
+  postflight do
+    system_command 'mv',
```
vitorgalvao: Don’t use `mv` (also, always use the full path). Instead, use ruby itself (`FileUtils`).

Casks/transmit-disk.rb
```diff
@@ -0,0 +1,61 @@
...
+                   args: [
+                           '/tmp/Transmit/Transmit Disk.app',
+                           '/Applications/',
```
vitorgalvao: Use `appdir` so it uses the user’s choice. Also, don’t end directory paths in `/`.

Casks/transmit-disk.rb
```diff
@@ -0,0 +1,61 @@
...
+  uninstall quit:    [
+                       'com.panic.Transmit',
+                       'com.panic.Transmit.TransmitDisk',
+                       'com.panic.Transmit.TransmitMenu',
+                     ],
+            pkgutil: [
+                       'com.panic.TransmitDisk.pkg',
+                       'com.panic.TransmitDisk.filesystem.tdfuse.pkg',
+                     ],
+            trash:   '/Applications/Transmit Disk.app'
```
vitorgalvao: `appdir` here as well. Also, use `delete`, `trash` does not work yet.

Casks/transmit-disk.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  zap delete: [
+                '~/Library/Preferences/com.panic.Transmit.TransmitDisk.plist',
+                '/Library/Filesystems/transmitdisk.fs/',
+                '/Applications/Transmit Disk.app',
```
vitorgalvao: `zap` always runs `uninstall` first. Remove this line.

Casks/transmit-disk.rb
```diff
@@ -0,0 +1,61 @@
+cask 'transmit-disk' do
+  version '4.4.12'
+  sha256 '3a4e3515450e597fc7f0d75f0d53df215bea3b6d9615c0a960f471b1185602ef'
+
+  url "https://download.panic.com/transmit/Transmit%20Disk%20#{version}.pkg.zip"
+  name 'transmit-disk'
+  homepage 'https://panic.com/transmit/'
+
+  depends_on cask: 'transmit'
```
vitorgalvao: If `transmit` can be installed from the App Store, don’t force it being installed by us.

vitorgalvao (changes requested):

Casks/transmit-disk.rb
phatblat: I don't know if `FileUtils` will work. Is there a way to run it through `sudo`?

```
Error: Permission denied - (/tmp/Transmit/Transmit Disk.app, /Applications/Transmit Disk.app)
```

Installer packages get ran [through `sudo`](https://github.com/Homebrew/brew/blob/d1f599f1ed51976fc76ad19803de04864d12ef2b/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb#L55), so the installed files are owned by root.

```
ls -ld /Applications /tmp /tmp/Transmit /tmp/Transmit/Transmit\ Disk.app                                                                                                                                                                                                          5  18:41:42
drwxrwxr-x+ 163 root        admin        5542 May 27 18:05 /Applications/
lrwxr-xr-x@   1 root        wheel          11 Sep 26  2016 /tmp -> private/tmp
drwxr-xr-x    3 root        wheel         102 May 27 18:40 /tmp/Transmit/
drwxr-xr-x    3 root        wheel         102 May 27 18:40 /tmp/Transmit/Transmit Disk.app/
```

phatblat: nvm, I found [set_ownership](https://github.com/Homebrew/brew/blob/b6f3399cb3facababf4f6b1e6f7821edc2db554a/Library/Homebrew/cask/lib/hbc/staged.rb#L33)

vitorgalvao (approved):

PR #34969, opened 2017-05-27
Add DollyDrive.app v3.94.12

Saklad5:

### DIFF
--- a/Casks/dollydrive.rb
+++ b/Casks/dollydrive.rb
@@ -1,0 +1,13 @@
+cask 'dollydrive' do
+  version '3.94.12'
+  sha256 '654c2b0d5a6eab71e6bf60769ace36f5c24b287d00ce8f5b6fcb47ac904825ee'
+
+  url "http://dollydrive.com/download-center/dollydrive/DollyDrive_#{version.major_minor}_#{version.no_dots}_CERTIFIED.zip"
+  appcast 'http://www.dollydrive.com/dolly3.xml',
+          checkpoint:
+                      '3ff4fe0e5a38053b5179774599043aa76b718d6af4286d28ace8a8bce93524d1'
+  name 'DollyDrive'
+  homepage 'http://www.dollydrive.com/'
+
+  app 'DollyDrive.app'
+end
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):

Saklad5:
I don’t think http://www.dollydrive.com/dolly2.xml exists. It may not be a good idea to assume that the “3” in http://www.dollydrive.com/dolly3.xml will change with the version number. All we know is that the current URL is what the application itself checks.

vitorgalvao:
@Saklad5 On the contrary. It’s likely it’ll change with the version number, and it’s better to have it break if it doesn’t, forcing us to fix it, then not break and be outdated, making us not notice for a while.

Casks/dollydrive.rb
```diff
@@ -4,8 +4,7 @@
 
   url "http://dollydrive.com/download-center/dollydrive/DollyDrive_#{version.major_minor}_#{version.no_dots}_CERTIFIED.zip"
   appcast "http://www.dollydrive.com/dolly#{version.major}.xml",
-          checkpoint:
-                      '3ff4fe0e5a38053b5179774599043aa76b718d6af4286d28ace8a8bce93524d1'
+          checkpoint: '3ff4fe0e5a38053b5179774599043aa76b718d6af4286d28ace8a8bce93524d1'
```
Saklad5: Why didn’t the style check change that if it was incorrect?

vitorgalvao: `style` isn’t magical, it only checks what we told it to. It’s likely this is not a rule we thought about before. You’re the first case I ever remember doing it this way, which is non-standard.

Saklad5: I copied directly from `find_sparkle_appcast`’s output. Odd.

vitorgalvao: Is it possible that your terminal width made the (long) result jump to a different line?

Saklad5: It wasn’t doing that before, but I just switched from `screen` to `tmux`, so that is definitely a possibility.

PR #35058, opened 2017-05-31
Update capture-one to add major version to app name

garand:
When installing this the installer choked because it couldn't find the app file, I presume because this was removed so looks like the major version needs to be re-added to the app name.


victorpopkov:
Yes, I can confirm that the app still has a number in its name. Thank you for submitting a PR to fix it.

PR #35226, opened 2017-06-05
Update virtualbox - add uninstall preflight

commitay:

Checks to see if `virtualbox-extension-pack` exists at `"#{HOMEBREW_PREFIX}/Caskroom/virtualbox-extension-pack"` and uninstalls if it is found.

> https://github.com/caskroom/homebrew-cask/issues/29301#issuecomment-295478018 
> `virtualbox` was updated today with a minor version change which happens to be a dependent of `virtualbox-extension-pack`. Reinstall works fine, but then the extension pack is missing because everything had been deleted even though cask shows the package as installed.

I've had this happen a few times with updates or switching between `virtualbox` / `virtualbox-beta`,  also you can't uninstall `virtualbox-extension-pack` without `virtualbox`.

/cc @colindunn

### DIFF
--- a/Casks/virtualbox.rb
+++ b/Casks/virtualbox.rb
@@ -10,6 +10,12 @@ cask 'virtualbox' do
 
   pkg 'VirtualBox.pkg'
 
+  uninstall_preflight do
+    if File.exist?("#{HOMEBREW_PREFIX}/Caskroom/virtualbox-extension-pack")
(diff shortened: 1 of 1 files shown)


colindunn:
Makes sense to me but it would mean that you have to install the extension pack again as a slight gotya. Seems like a reasonable trade-off.

PR #35256, opened 2017-06-06
Add shim script for copyq 3.0.2

hluk:

### DIFF
--- a/Casks/copyq.rb
+++ b/Casks/copyq.rb
@@ -5,10 +5,19 @@ cask 'copyq' do
   # github.com/hluk/CopyQ was verified as official when first introduced to the cask
   url "https://github.com/hluk/CopyQ/releases/download/v#{version}/CopyQ.dmg"
   appcast 'https://github.com/hluk/CopyQ/releases.atom',
-          checkpoint: '4d5aa488e9fef7dd477eca8c7756c03e4eb81eb0a3988b89345fbd6295e3f7f9'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> Hmm, the commit message not correct -- I used `cask-repair -e ...` -- it should be something like `Add shim script for copyq`.

`-e` does not change the commit message, it only opens the cask for editing before checkign everything else. You can’t change the commit message with `cask-repair` (and I don’t intend to add the feature).

vitorgalvao:
Why is this change necessary? I don’t see a difference.

hluk:
Symlinks doesn't work with Qt apps - when the symlink is executed the target app doesn't search the correct directories with libraries and in this specific case fails to load Cocoa Qt plugin.

hluk:
BTW, I fixed the commit message.

vitorgalvao:
Are you sure the `exec` is necessary? Also, use `/bin/bash`, that’s what `sh` is on macOS, anyway.

gibfahn:
I can confirm that this fixes the error we were seeing before (e.g. https://github.com/hluk/CopyQ/issues/719#issuecomment-306408808).

Thanks @hluk

hluk:
OK, I replaced `/bin/sh` with `/bin/bash`.

The `exec` is not needed but it replaces the bash process and probably saves a bit of memory.

vitorgalvao (approved):
Everything looks fine. Good to go when Travis passes.

PR #35371, opened 2017-06-10
Update tunnelblick to 3.7.1a,4812

vitorgalvao:

---

The app changes its own ownership on first run.

### DIFF
--- a/Casks/tunnelblick.rb
+++ b/Casks/tunnelblick.rb
@@ -14,7 +14,7 @@ cask 'tunnelblick' do
 
   app 'Tunnelblick.app'
 
-  postflight do
+  uninstall_postflight do
     set_ownership "#{appdir}/Tunnelblick.app"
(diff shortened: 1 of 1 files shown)


commitay:
Still prompts to change ownership and permissions even with the `postflight set_ownership`.

vitorgalvao:
> Still prompts to change ownership and permissions even with the `postflight set_ownership`.

I meant to do `uninstall_postflight`. I don’t want to change it’s permissions on usage (it does so for security reasons), only make it uninstallable (otherwise it fails). Fixed.

commitay:
Shouldn't it be `uninstall_preflight` so it comes before the `app` is removed?

vitorgalvao:
> Shouldn't it be `uninstall_preflight` so it is called before the `app` is removed?

Yes. Fixed. Thank you.

PR #35395, opened 2017-06-11
Update miniconda - change the default installation path

ghost:
Move the installation to `/usr/local` instead of home directory (same as anaconda #32706).

### DIFF
--- a/Casks/miniconda.rb
+++ b/Casks/miniconda.rb
@@ -12,12 +12,18 @@ cask 'miniconda' do
 
   installer script: {
                       executable: 'Miniconda3-latest-MacOSX-x86_64.sh',
-                      args:       ['-b'],
+                      args:       ['-b', '-p', "#{HOMEBREW_PREFIX}/miniconda3"],
+                      sudo:       true,
                     }
 
-  uninstall delete: '~/miniconda3'
+  postflight do
+    set_ownership "#{HOMEBREW_PREFIX}/miniconda3"
+  end
+
+  uninstall delete: "#{HOMEBREW_PREFIX}/miniconda3"
 
   caveats do
-    path_environment_variable '~/miniconda3/bin'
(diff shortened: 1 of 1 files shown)


Casks/miniconda.rb
```diff
@@ -12,12 +12,18 @@
...
   installer script: {
                       executable: 'Miniconda3-latest-MacOSX-x86_64.sh',
-                      args:       ['-b'],
+                      args:       ['-b', '-p', "#{HOMEBREW_PREFIX}/miniconda3"],
+                      sudo:       true,
                     }
 
-  uninstall delete: '~/miniconda3'
+  postflight do
+    set_ownership "#{HOMEBREW_PREFIX}/miniconda3"
```
reitermarkus: Is this really needed? You should already have permissions for `HOMEBREW_PREFIX`.

reitermarkus:
I think it would make more sense to set the prefix to `"#{staged_path}/miniconda"` and only linking the needed `binary`s.

Casks/miniconda.rb
zhangpj: On my machine:

```
ls -l /usr
```
gives
```
drwxr-xr-x    17 root  wheel    578 Jun 11 10:16 local
```

So I added that.

zhangpj:
@reitermarkus I have no idea which one is better, but `anaconda.rb` uses `#{HOMEBREW_PREFIX}` ([anaconda.rb#L14](https://github.com/caskroom/homebrew-cask/blob/aa07f4b7479bfc1127ed8f1a65db6986050c032a/Casks/anaconda.rb#L14)).

All the binary files are in `miniconda/bin`, including the executable files of packages installed afterwards with `conda`. Usually people add `miniconda/bin` to `PATH` instead of symlinking some files.

reitermarkus:
Ok, let's leave it as is then, thanks!

PR #35399, opened 2017-06-11
Update antconc to 3.5.0

gtmgianni:

VirusTotal: https://www.virustotal.com/en-gb/file/f5d4e2c87effc3dd98ad7ac4ebd86a5fc22110cf6de608b670daf66207c949ba/analysis/1497208824/

### DIFF
--- a/Casks/antconc.rb
+++ b/Casks/antconc.rb
@@ -3,13 +3,13 @@ cask 'antconc' do
     version '3.4.1'
     sha256 '03c353c059b8c0762b01d9be83f435321f5396cbf203bd8b36c6a56682b6a240'
   else
-    version '3.5.0'
-    sha256 'd79cdc444f7284c34dc51b8aebf4bdf356ef82ea6a9ad17858db083277ef5369'
+    version '3.4.4'
+    sha256 '2c346728c70dce3279647005f8dd704e48368c91aada684aaee4ce01017c1327'
   end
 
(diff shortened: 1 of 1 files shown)


giannitm:
The file does not appear to be signed (I think)?

vitorgalvao:
Yep. Not signed, so the virus total link is not enough in this case.

commitay:
@vitorgalvao Should we add this to the `cask-repair` PR template?
```

```

vitorgalvao:
@commitay I wouldn’t oppose it.

giannitm:
Just realised that this is a development version of `antconc`. That explains why the `sha256` changed and the version number didn't.

The latest version is 3.4.3, but that was released in October 2016. Should it be downgraded or not? Pinging @vitorgalvao.

vitorgalvao:
> Should it be downgraded or not? 

[Is this true?](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version):

> where such a version is the general standard

Do they treat this development version as a development version or as “this is the version you should use”?

giannitm:
The download link on the homepage links to version 3.4.4, so I assumed that that would be the latest release. Having said that, the development version says it is only compatible up to OS X Yosemite, whereas the older "latest" release is compatible with macOS Sierra? So I'm going to downgrade the version to 3.4.4, but I'll make the pull request later.

PR #35434, opened 2017-06-12
Migrate plantronics-hub to caskroom/drivers

commitay:
Updated `tap_migrations.json`

cask has hardware dependencies.

Closes https://github.com/caskroom/homebrew-cask/issues/35433

https://github.com/caskroom/homebrew-drivers/pull/72


tap_migrations.json
```diff
@@ -78,6 +78,7 @@
   "pgloader": "homebrew/core",
   "phantomjs": "homebrew/core",
   "pl2303": "caskroom/drivers",
+  "plantronics-hub": "caskroom/drivers",
```
reitermarkus: This doesn't work for Casks.

commitay: > This has no effect for Casks.

I don't understand?

commitay: Ping @reitermarkus , @vitorgalvao 

I'm not sure what I should do with this as almost everything in `caskroom/drivers` has an entry in `tap_migrations.json`?

vitorgalvao: I’d leave it. It has no effect so it’s not detrimental, and if it ever has an effect, it’s already done.

`conflicts_with` is in a similar position. It exists but does nothing yet.

PR #35684, opened 2017-06-19
Update osxfuse to 3.6.0

tiiiecherle:

- [ ] [If the `sha256` changed but the `version` didn’t](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256),


commitay:
Version `3.6.0` is a developer preview.

tiiiecherle:
could you change the cask that way that every 10.12 system gets the stable one and 10.13 gets the new one? would be nice as the stable is not working on 10.13. I think I have seen a modification of a cask like that but I don`t know how to do it.

thanks a lot

PR #35711, opened 2017-06-20
Update ISSUE_TEMPLATE.md - `brew config`

commitay:
Add `brew config` to ISSUE_TEMPLATE.md

Useful for `java` related issues and MacOS versions, reduces the need for follow up questions.


commitay:
I went back through some old issues and `brew config` is less useful than I thought it was.

PR #35931, opened 2017-06-27
Add AutoFirma Latest

ignic:

Hello,

This is my first attempt to make a new cask, please be gentle :)

The URL has no version, so I have used "version: latest". But the zip that is downloaded has a pkg with version number in the filename. I think that this would break when a new version is updated. Is there a better way to handle this?

Thank you all for the Homebrew project!

Regards

### DIFF
--- a/Casks/autofirma.rb
+++ b/Casks/autofirma.rb
@@ -26,6 +26,6 @@ cask 'autofirma' do
                    sudo: true
   end
 
-  uninstall pkgutil: 'es.gob.afirma'
-  uninstall delete:  '/Applications/AutoFirma.app'
+  uninstall pkgutil: 'es.gob.afirma',
+            delete:  '/Applications/AutoFirma.app'
 end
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> But the zip that is downloaded has a pkg with version number in the filename. I think that this would break when a new version is updated. Is there a better way to handle this?

Make the cask versioned and add a `sha256`. It’ll break anyway without it, so no point in not having it.

ignic:
Thanks to @commitay for the fixes! I have spent more than an hour to figure out that I have to use double quotes instead of single quotes if I want to use the #{version} variable.

vitorgalvao:
> I have spent more than an hour to figure out that I have to use double quotes instead of single quotes if I want to use the #{version} variable.

That is common in languages where you use interpolation (`bash` also works like that): single quotes are literal, double quotes expand interpolation.

vitorgalvao:
Thank you for your first contribution!

PR #36046, opened 2017-07-01
Remove phocus

commitay:
App download is behind a login page.

http://www.hasselblad.com/support/manuals/software-current/


victorpopkov (approved):

vitorgalvao:
Note that the download link being behind a download page is only a problem when `homepage` and `url` have different domains. That is so a malicious actor can’t simply point the cask to a malicious source without users being able to verify with ease.

When they have the same domain, like in this case, it is fine. A user can look at the cask and be confident the `url` is legitimate (a malicious actor would have to target the upstream domain; which is harder to do and keep up).

commitay:
Just for clarification: are we okay with making the app available to users who would otherwise have to register to download it?

vitorgalvao:
Unless upstream objects, yes.

Naturally, if you disagree we can discuss it. I initially conceived the rule, but am not attached to it or anything. I could certainly be persuaded otherwise.

commitay:
> Naturally, if you disagree we can discuss it. I initially conceived the rule, but am not attached to it or anything. I could certainly be persuaded otherwise.

No disagreement from me. Just wanted to clarify as I've already rejected a cask request for being behind a login/registration page. I'll take another look at it to see if it uses the same domain for `homepage` and `url`.

> Note that the download link being behind a download page is only a problem when `homepage` and `url` have different domains.

I'll add this to the docs somewhere.

____

After adding `anka-run` I realised it was behind a login page and removed it (https://github.com/caskroom/homebrew-cask/pull/35801), I'll re-add it.

vitorgalvao:
> I'll add this to the docs somewhere.

It’s already in the docs. See “walled builds” in the “adding a cask” document.

As for `anka`, that one was also fine, as I recall, as you found a page with download links that was not behind registration.

commitay:
> It’s already in the docs. See “walled builds” in the “adding a cask” document.

Oops, sorry  @vitorgalvao. Thank you for your explanations and patience, they are appreciated. 🥇

PR #36052, opened 2017-07-01
updated Gravit Designer URL

olierxleben:


Casks/gravitdesigner.rb
```diff
@@ -2,7 +2,7 @@
   version :latest
   sha256 :no_check
 
-  url 'https://app.designer.io/_downloads/mac/GravitDesigner.zip'
+  url 'https://designer.gravit.io/_downloads/mac/GravitDesigner.zip'
```
commitay: @olivererxleben Where did you get this link? https://www.designer.io/ only has a MAS version.

olivererxleben: Gravit changed urls in latest release. 

app.deisgner.io pointed to web version so I manually tested that.

olivererxleben: Thanks for adding the url stanza - I forgot to do so.

commitay:
Ping @vitorgalvao This seems to have moved to MAS, no download link at https://www.designer.io/

vitorgalvao:
I’m fine with removing.

olivererxleben:
This was not my intention in any way when I started the pull request. I wanted to update the brew version to the latest release. And help maintaining up to date versions. You know, giving back to the community etc. 

You suggest to use the MAS Version? Before my pull request there was just a MAS version either. How do you decide what can be removed and what not? Are there any official guidelines here?

vitorgalvao:
> How do you decide what can be removed and what not? Are there any official guidelines here?

Yes. In several documents we mention we cannot install MAS-only apps. I understand this is not technically MAS-only, but as @commitay noted, nowhere on the website does this non-MAS link seem to be listed (and we also cover that case). Add that to the fact this cask isn’t versioned, and from experience what’ll happen is:

+ Eventually the non-MAS version will stop being updated, and we won’t know (again, no version on the URL), making so that users only download an old version without realising.´
+ Eventually they’ll remove the direct link, breaking the cask.

See specifcially the “[rejected casks](https://github.com/caskroom/homebrew-cask/blob/a363322d7649061c0b3c3a55d1400231ec0bb614/doc/faq/rejected_casks.md)” FAQ:

> Similarly (and trickier to spot), the app has moved to the Mac App Store but still provides old versions via direct download. We reject these in all official repos so users don’t get stuck using an old version, wrongly thinking they’re using the most up-to-date one (which, amongst other things, might be a security risk).

commitay:
https://github.com/caskroom/homebrew-cask/pull/36070

PR #36211, opened 2017-07-05
Style fix: use %i for array of symbols

vitorgalvao:


reitermarkus:
I think we should disable this, it's easier to grep for `:****` symbols.

vitorgalvao:
> I think we should disable this, it's easier to grep for `:****` symbols.

I’m fine with that.

PR #36523, opened 2017-07-14
Remove google-earth

commitay:
https://www.google.com/earth/download/ge/agree.html has different downloads for `google-earth` and `google-earth-pro` but they both download the same `dmg` containing:
`Install Google Earth Pro #{version}.pkg`

Removed the `caveat` from `google-earth-pro` as a license key not required.
____

Note: There are versioned downloads available for both "Google Earth" versions but they are outdated and do not auto update.


et:
Should there be a comment to explain that you can use the pro version for the free version?

PR #36725, opened 2017-07-18
Certificate issue on installation

suxxes:
Certificate used to sign package is not trusted. Use -allowUntrusted to override.

### DIFF
--- a/Casks/ksdiff.rb
+++ b/Casks/ksdiff.rb
@@ -8,7 +8,7 @@ cask 'ksdiff' do
 
   conflicts_with cask: 'kaleidoscope'
 
-  pkg 'Install ksdiff.pkg'
+  pkg 'Install ksdiff.pkg', allow_untrusted: true
 
   uninstall pkgutil: 'com.blackpixel.kaleidoscope.ksdiff.installer.pkg'
 


vitorgalvao:
Thank you for the contribution. Next time, please fill the template.

I’m going to refuse this. There’s a significant difference between an unsigned packaged and one with an invalid signature. This is the latter. By adding `allow_untrusted`, you’re making so that when the certificate is fixed, we won’t notice and then be vulnerable.

The correct fix here is to contact upstream about the issue. Care to do so?

vitorgalvao (changes requested):
There’s a significant difference between an unsigned packaged and one with an invalid signature. This is the latter (expired certificate).

vitorgalvao:
Closing for lack of reply.

If someone else stubs against this problem before upstream fixes it, we’ll deal with it then.

BM5k:
Hitting this for ksdiff today. Manual installation worked without a certificate error. Not sure what needs to be changed to remove the error from cask-based install.

vitorgalvao:
> Not sure what needs to be changed to remove the error from cask-based install.

You need to contact the developers of the app and tell them the certificate they used to sign the package has expired.

BM5k:
@vitorgalvao the cask points to this [package][0] with this sha256 checksum `9570f53dcbeb558c53f4808ba58e8c9f394a3026e8bdd122277200a1cdf11e52`

That's the same package linked to [here][1] (same download URL).
Manually downloading the file results in the same checksum:

```
openssl dgst -sha256 ~/Downloads/ksdiff-122.zip
SHA256(/Users/bm5k/Downloads/ksdiff-122.zip)= 9570f53dcbeb558c53f4808ba58e8c9f394a3026e8bdd122277200a1cdf11e52
```

Installing from the manual download doesn't complain about the certificate warning.
Installing from the cask does.

So, it seems like there is a problem with the cask _not_ the certificate or the package (unless cask is more strictly checking the cert than installer, which would be a lame decision on apple's part).

[0]: https://cdn.kaleidoscopeapp.com/releases/ksdiff-122.zip
[1]: https://www.kaleidoscopeapp.com/ksdiff2

commitay:
```
pkgutil --check-signature Install\ ksdiff.pkg
Package "Install ksdiff.pkg":
   Status: signed by a certificate that has since expired
   Certificate Chain:
    1. Developer ID Installer: Black Pixel
       SHA1 fingerprint: B2 0D D5 48 F4 61 57 48 92 F2 2F 3F 7F A3 82 3A 3D 25 1C F5
       -----------------------------------------------------------------------------
    2. Developer ID Certification Authority
       SHA1 fingerprint: 3B 16 6C 3B 7D C4 B7 51 C9 FE 2A FA B9 13 56 41 E3 88 E1 86
       -----------------------------------------------------------------------------
    3. Apple Root CA
       SHA1 fingerprint: 61 1E 5B 66 2C 59 3A 08 FF 58 D1 4A E2 24 52 D1 98 DF 6C 60

```

BM5k:
@commitay OK, I verified that the manual download also fails that check.

Does installer not display those errors when installing? I received no warning in the GUI when installing _this package_ today.

commitay:
> Does installer not display those errors when installing?

`/usr/sbin/installer` seems to have stricter cert checks.

PR #36748, opened 2017-07-19
fix(brave): set auto_updates dependencies

koalabz:
- fix annoying `brew cask outdated`message version mismatch
 - add minimal install dependencies on osx


miccal:

PR #36799, opened 2017-07-20
Delete cleanarchiver.rb

miccal:
Both the `homepage` and `url` are dead.


etienned:
Both the homepage and url are now alive.

miccal:
@etienned please feel free to re-add this Cask if you wish.

PR #36801, opened 2017-07-20
Update punto-switcher.rb

miccal:
Update `url`.


Dreamisdead:
Hey. Can we disable installation of optional yandex utils?

PR #36803, opened 2017-07-20
Delete tk-suite-client.rb

miccal:
The `url` is dead and this software does not seem to be available any more.


commitay:
I think their FTP server is misconfigured. If anyone wants the cask it can be re-added.

PR #36891, opened 2017-07-23
Remove appcast from 17 discontinued casks

commitay:


reitermarkus:
Them being discontinued doesn't make their appcasts invalid.

commitay:
This was to avoid checkpoint update only PRs on these casks, a few have outdated appcasts but no new development or releases.

miccal:
I agree with @reitermarkus, the `appcast`'s are still valid and should remain.

PR #36947, opened 2017-07-25
Add versioning to ZeroNet ('0.5.7') current

BenjaminHCCarr:
The `sha256` and `version` will change but the `URL` currently won't.

Unfortunately, I don't see a way to automate versioning as their own versioning is out of date: https://github.com/HelloZeroNet/ZeroNet/tags versus the target URLs on https://zeronet.io/

Though this slightly increases maintainer overhead considering it is a security/privacy app, I think a functioning `sha` is good.

---

### DIFF
--- a/Casks/zeronet.rb
+++ b/Casks/zeronet.rb
@@ -1,6 +1,6 @@
 cask 'zeronet' do
-  version :latest
-  sha256 :no_check
+  version '0.5.7'
+  sha256 'beac83a1299415a90d58b680c0f466337d527337d9214b96e188fcc8ce733d8f'
 
   # github.com/HelloZeroNet/ZeroNet-mac was verified as official when first introduced to the cask
   url 'https://github.com/HelloZeroNet/ZeroNet-mac/archive/dist/ZeroNet-mac.zip'


commitay:
Without an `appcast` that matches the current release this cannot be versioned.

If you still want this to be versioned you should contact the developers and get them to provide a versioned release.

BenjaminHCCarr:
@GitHub !?!?!?!??!!?! Locked me out.....

Your account has been flagged.
Because of that, your profile is hidden from the public. If you believe this is a mistake, contact support to have your account status reviewed.

---

I got "Flagged" / Locked / Private
Thinking it's @KeybaseIO related https://github.com/keybase/proofs/issues/39

commitay:
I noticed that your account 404ed and all issues and PRs authored by you disappeared.

Not sure I understand what you mean by updating the required stanzas?

BenjaminHCCarr:
I understand that `appcast` makes automation much easier. but there are many cask
in `/usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask/Casks` we get:
```
tyr:/usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask/Casks (master) benc$ ls *.rb | wc
    3663    3663   51214
tyr:/usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask/Casks (master) benc$ grep -ci appcast *.rb | grep -c 1$
1991
```
So just about half of all Cask do not have an appcast.

BenjaminHCCarr:
Homebrew cask has never requeired the `appcast` stanza; it is an optional, see: https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/all_stanzas.md

Over half our casks don't have one at all.

commitay:
I think this is a misunderstanding over my previous comment.

> Without an appcast that matches the current release this cannot be versioned.

What I meant was **_this_** cask with the current download `url` cannot be versioned without an appcast.

If it had a versioned `url` we can use it doesn't necessarily need to have an `appcast`.

BenjaminHCCarr:
*abuses pipes*
`tyr:/usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask/Casks (master) benc$ grep -ci appcast *.rb | grep 0$ | awk -F: '{print $1}' | xargs grep url | grep -v \#\{version | grep 0$ | awk -F: '{print $1}'`

So yes, everything either has a versioned URL or an appcast.

And I understand the confusion you and I had about the nomenclature of the `appcast` stanza.

That sadly still leaves us with ~`732` unverified casks.
1) using no appcast
2) using sha256 no_check

I understand this makes maintenance easier. But from an end user perspective it is frustrating. `latest` doesn't mean anything particularly if there is no `auto_updates`.

Maybe we should revisit mainlining: https://github.com/buo/homebrew-cask-upgrade

Which has made my user experience better; particularly with all of the moving-bits and re-orgs in the last year.

commitay:
> Maybe we should revisit mainlining: https://github.com/buo/homebrew-cask-upgrade

See https://github.com/caskroom/homebrew-cask/issues/29301 for discussion about why this won't happen.
___

>That sadly still leaves us with ~`732` unverified casks.
>1) using no appcast
>2) using sha256 no_check

A lot of these casks casks we quite simply cannot version as there is no appcast / changelog and no versioned downloads.

PR #37056, opened 2017-07-26
Migrate wootility to drivers

commitay:
https://github.com/caskroom/homebrew-drivers/pull/107


petems:
"Software to make a hardware peripheral recognisable and usable by the system. If the software is useless without the peripheral, it’s considered a driver."

@commitay @reitermarkus 

They keyboard works without a drive... so I'm not sure this applies? But not super fussed

PR #37090, opened 2017-07-27
Update teamviewer casks (2) - change to sha256 no_check

commitay:
The changelog for these casks is unstable and cannot be used as an appcast. 

The version is shown here https://www.teamviewer.com/en/download/mac/ but the downloads seem to be updated in place without a version bump.


commitay:
@miccal Thanks!

PR #37288, opened 2017-08-01
Update dotnet-sdk to 1.0.4

brucemontegani:

- [ ] [If the `sha256` changed but the `version` didn’t](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256),

### DIFF
--- a/Casks/dotnet-sdk.rb
+++ b/Casks/dotnet-sdk.rb
@@ -1,32 +1,41 @@
 cask 'dotnet-sdk' do
-  version '1.1.1'
-  sha256 '5b432bc4cf63ea63e8c28a089747feb8a10364e91dae734ca69942ea14188c81'
+  version '1.0.4'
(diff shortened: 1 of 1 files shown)


bmontegani:
The official version number for the latest SDK is 1.0.4. I am not sure where the previous version number 1.1.1 came from. It probably had something to do with the runtime not the SDK. Also the latest SDK contains two runtimes. The current runtime is: 1.1.2 while the long term support(LTS) is 1.0.5. The download locations are straight from the Microsoft .net core website as well as the checksum number.
[.NET Core](https://www.microsoft.com/net/download/core)

bmontegani:
I only added the version to the name so that a user of the Cask might have more information. What I really wanted is to let the user know which runtimes were included. is there another place to put this information?

commitay (approved):

miccal:

PR #37310, opened 2017-08-02
soapui fix - program hangs on first launch

TheLonelyGhost:
Due diligence:

- [x] If `sha256` changed but the version didn't, included public confirmation by developer.

---

See [StackOverflow answer](https://stackoverflow.com/a/44325050/1236035) about problem with running SOAP UI after installing. This automates that fix, with slightly future-proof checks in case it is fixed upstream.

Also includes some configuration file additions to zap, since there was a missing zap stanza.

### DIFF
--- a/Casks/soapui.rb
+++ b/Casks/soapui.rb
@@ -66,8 +66,9 @@ cask 'soapui' do
 
   uninstall delete: "/Applications/SoapUI-#{version}.app"
 
-  zap delete: [
-                '~/default-soapui-workspace.xml',
-                '~/soapui-settings.xml',
-              ]
+  zap trash: [
+               '~/.soapuios',
+               '~/default-soapui-workspace.xml',
+               '~/soapui-settings.xml',
+             ]
 end
--- a/Casks/soapui.rb
+++ b/Casks/soapui.rb
@@ -46,5 +46,23 @@ cask 'soapui' do
                                   ],
                     }
 
+  postflight do
+    "/Applications/SoapUI-#{version}.app/Contents/vmoptions.txt".tap do |file|
+      break unless File.exist? file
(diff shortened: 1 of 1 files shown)


Casks/soapui.rb
```diff
@@ -46,5 +46,28 @@
...
+
+      File.open(file, 'a') { |f| f.puts '-Dsoapui.browser.disabled=true' }
+    end
+
+    "/Applications/SoapUI-#{version}.app/Contents/java/app/bin/soapui.sh".tap do |file|
+      break unless File.exist? file
+      # Does it already exist in the file?
+      break if File.read(file) =~ %r{^(?<!#)JAVA_OPTS=".+-Dsoapui\.browser\.disabled=true(?:\s.*)?"(?:\s+.*)?$}i
+
+      system %Q{sed -i'' -e 's/#.*\\(JAVA_OPTS=".*-Dsoapui\\.browser\\.disabled=true".*\\)$/\\1/g' "#{file}" 1>/dev/null}
```
vitorgalvao: Use `system_command`.

vitorgalvao (changes requested):

commitay:
I can't reproduce any issues with launching `soapui`, is this something we should be fixing here if it isn't a consistent problem?

Casks/soapui.rb
reitermarkus: This could be done in Ruby, so we don‘t have to duplicate the regex.

reitermarkus (changes requested):

commitay:
Ping @TheLonelyGhost

miccal:
Closing in favour of https://github.com/caskroom/homebrew-cask/pull/37594.

PR #37317, opened 2017-08-03
Make google-cloud-sdk versioned to 165.0.0

cglong:

### DIFF
--- a/Casks/google-cloud-sdk.rb
+++ b/Casks/google-cloud-sdk.rb
@@ -1,8 +1,8 @@
 cask 'google-cloud-sdk' do
-  version :latest
-  sha256 :no_check
+  version '165.0.0'
+  sha256 'e4ff9f899ae94efecb601feb5b24ca48918317986a5f29200467736ff4132c7a'
 
-  url 'https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz'
+  url "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-#{version}-darwin-x86_64.tar.gz"
   name 'Google Cloud SDK'
   homepage 'https://cloud.google.com/sdk/'
 
--- a/Casks/google-cloud-sdk.rb
+++ b/Casks/google-cloud-sdk.rb
@@ -14,18 +14,11 @@ cask 'google-cloud-sdk' do
   binary 'google-cloud-sdk/bin/gcloud'
   binary 'google-cloud-sdk/bin/git-credential-gcloud.sh', target: 'git-credential-gcloud'
   binary 'google-cloud-sdk/bin/gsutil'
+  artifact 'google-cloud-sdk/completion.bash.inc', target: "#{HOMEBREW_PREFIX}/etc/bash_completion.d/google-cloud-sdk-completion.bash"
+  artifact 'google-cloud-sdk/completion.zsh.inc', target: "#{HOMEBREW_PREFIX}/share/zsh/site-functions/google-cloud-sdk-completion.zsh"
+  artifact 'google-cloud-sdk/path.bash.inc', target: "#{HOMEBREW_PREFIX}/etc/bash_completion.d/google-cloud-sdk-path.bash"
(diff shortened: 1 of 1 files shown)


commitay:
@cglong This will break the completions and PATH.

cglong:
@commitay Sorry, what do you mean about the PATH? And re: completions, is there a way to still support those (e.g. symlinking or something)? The main benefit I personally receive from Caskroom is auto-updating and asking users to instead manually enter `gcloud components update` seems counter to that goal.

commitay:
> Sorry, what do you mean about the PATH?
```
      for bash users
        source '#{staged_path}/#{token}/path.bash.inc'
        source '#{staged_path}/#{token}/completion.bash.inc'

      for zsh users
        source '#{staged_path}/#{token}/path.zsh.inc'
        source '#{staged_path}/#{token}/completion.zsh.inc'
```

> The main benefit I personally receive from Caskroom is auto-updating and asking users to instead manually enter gcloud components update seems counter to that goal.

Caskroom doesn't auto-update.

> And re: completions, is there a way to still support those (e.g. symlinking or something)?

Yes, using `artifact` or `postflight` and `uninstall_preflight`. We don't have support for installing completions like Homebrew.

It was suggested a while ago that this could be moved to `Homebrew-core` (https://github.com/Homebrew/homebrew-core/issues/583#issuecomment-214115437) which would auto-update and properly support installing completions.

cglong:
FWIW, it looks like the PATHs are resolving correctly, albeit inconveniently as you mentioned:

```
  for bash users
    source '/usr/local/Caskroom/google-cloud-sdk/165.0.0/google-cloud-sdk/path.bash.inc'
    source '/usr/local/Caskroom/google-cloud-sdk/165.0.0/google-cloud-sdk/completion.bash.inc'

  for zsh users
    source '/usr/local/Caskroom/google-cloud-sdk/165.0.0/google-cloud-sdk/path.zsh.inc'
    source '/usr/local/Caskroom/google-cloud-sdk/165.0.0/google-cloud-sdk/completion.zsh.inc'
```

> Caskroom doesn't auto-update.

Oh, I never noticed this (I'm pretty new to Caskroom tbh)! Either way, I do like having a centralized place for managing my command line tools, rather than having to memorize vendor-specific commands.

> Yes, using `artifact` or `postflight` and `uninstall_preflight`.

Awesome! I'll toy around with these and see if I can get something working 🙂

commitay:
> it looks like the PATHs are resolving correctly, albeit inconveniently as you mentioned

Yes, I meant they will break for everyone currently using them.
___
I'm not sure I understand how making this versioned will avoid having to run `gcloud components update`? Only `bq`, `core`, `gsutil` and `gcloud` are installed by default.

cglong:
> Updating the Cask (running `brew cask reinstall google-cloud-sdk`) will remove any installed components.

Ah yeah, that's true. It could be interesting to explore making additional Casks for auxiliary components that `depends_on` this one, but I'm not sure how easy it would be to replicate what was done for the extra .debs/.rpms.

I could take this either way and will let you decide whether it should be merged in or not 🙂

commitay:
> It could be interesting to explore making additional Casks for auxiliary components that depends_on this one, but I'm not sure how easy it would be to replicate what was done for the .debs/.rpms corresponding to those components.

This would also work better as a Homebrew formula which would allow install options and `resource` blocks.

Whilst we do support `depends_on cask` until we have a `brew cask upgrade` (https://github.com/caskroom/homebrew-cask/issues/29301) it isn't going to work very well and it could have issues if the casks are not updated at the same time.
____

Going to pass on this as versioning this cask will mean downloaded components will be overwritten.

Thank you anyway for the PR @cglong!

cglong:
@commitay I just noticed `install.sh` has an `--additional-components` flag where we can augment the list of what gets installed. I'm wondering if it'd make sense to just install *all* components by default? It would also help with situations like the `datalab` component which can be optionally installed but then isn't symlinked.

commitay:
We could install everything but this will mean downloading an additional 340Mb of components (installed size is 800Mb larger than the default of 250Mb) that users may not need or want. 

`gcloud components update` would update all installed components so users would have to uninstall what they don't use.

PR #37323, opened 2017-08-03
Update java - add uninstall_preflight

commitay:

[`java-jdk-javadoc`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/java-jdk-javadoc.rb) and [`jce-unlimited-strength-policy`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/jce-unlimited-strength-policy.rb) are casks installed using `system_command` into `java_home`.

Trying to uninstall/reinstall either of these casks after `java` is uninstalled/reinstalled fails.

This adds an `uninstall_preflight` that checks `#{HOMEBREW_PREFIX}/Caskroom/` for both of these casks and uninstalls them if found.

### DIFF
--- a/Casks/java-jdk-javadoc.rb
+++ b/Casks/java-jdk-javadoc.rb
@@ -31,5 +31,7 @@ cask 'java-jdk-javadoc' do
     License Agreement for Java SE at
 
       https://www.oracle.com/technetwork/java/javase/terms/license/index.html
+
+     #{token} will be uninstalled when the Java Cask is uninstalled or reinstalled
   EOS
 end
--- a/Casks/java.rb
+++ b/Casks/java.rb
@@ -45,6 +45,16 @@ cask 'java' do
     end
   end
 
+  uninstall_preflight do
+    if File.exist?("#{HOMEBREW_PREFIX}/Caskroom/java-jdk-javadoc")
+      system_command 'brew', args: ['cask', 'uninstall', 'java-jdk-javadoc']
+    end
+
+    if File.exist?("#{HOMEBREW_PREFIX}/Caskroom/jce-unlimited-strength-policy")
+      system_command 'brew', args: ['cask', 'uninstall', 'jce-unlimited-strength-policy']
+    end
+  end
+
   uninstall pkgutil:   [
                          "com.oracle.jdk#{version.minor}u#{java_update}",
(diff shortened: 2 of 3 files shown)


adidalal:
https://github.com/caskroom/homebrew-cask/issues/36933

Might also be worth adding a postinstall_preflight to reinstall those two specific casks on install?

commitay:
> Might also be worth adding a postinstall_preflight to reinstall those two specific casks on install?

This would be nice but I'm not sure how to check for a previously installed cask as part of a `postflight`.

miccal:
Add a `caveat` maybe?

commitay:
The `caveat` on this cask is already rather large.
```
  caveats <<-EOS.undent
    This Cask makes minor modifications to the JRE to prevent issues with
    packaged applications, as discussed here:

      https://bugs.eclipse.org/bugs/show_bug.cgi?id=411361

    If your Java application still asks for JRE installation, you might need
    to reboot or logout/login.

    Installing this Cask means you have AGREED to the Oracle Binary Code
    License Agreement for Java SE at

      https://www.oracle.com/technetwork/java/javase/terms/license/index.html
  EOS
```
____

Just to clarify uninstalling or reinstalling `java` with either of these to casks installed also removes them as `java` removes/overwrites their files, the purpose of this is to tidy up `{HOMEBREW_PREFIX}/Caskroom/` (and `brew cask list`) so users don't have to do it manually.

reitermarkus:
> Just to clarify uninstalling or reinstalling `java` with either of these casks installed also removes them as `java` removes/overwrites their files

We have `conflicts_with` for that.

commitay:
>We have conflicts_with for that.

It isn't a conflict, they install into `java_home`.

reitermarkus:
So it is a conflict, but only when uninstalling?

commitay:
> So it is a conflict, but only when uninstalling?

Yes, only when uninstalling/reinstalling.

commitay:
Added a `caveat` to `java-jdk-javadoc` and `jce-unlimited-strength-policy`

vitorgalvao (approved):
I don’t really like that we’re calling ourselves via a system command, but if there’s no better way and this is needed, so be it.

PR #37388, opened 2017-08-04
PULL_REQUEST_TEMPLATE.md: add "How?" to sha256 part

vitorgalvao:
Most contributors either ignore this section or mark it when they shouldn’t. Seems to me like no one clicks through. Perhaps a “How?” will make it more explicit.

To do after merging:

- [x] Update in `cask-repair`
- [x] Run the sync script

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,8 +8,8 @@ After making all changes to the cask:
 
 Additionally, if **updating a cask**:
 
-- [ ] [If the `sha256` changed but the `version` didn’t][version-checksum],  
-      provide public confirmation by the developer: {{link}}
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,8 +8,8 @@ After making all changes to the cask:
 
 Additionally, if **updating a cask**:
 
-- [ ] [If the `sha256` changed but the `version` didn’t][version-checksum],  
-      provide public confirmation by the developer: {{link}}
+- [ ] If the `sha256` changed but the `version` didn’t,  
+      provide public confirmation by the developer ([How?][version-checksum]): {{link}}
```
commitay: We could probably remove the words "by the developer".

PR #37519, opened 2017-08-08
PULL_REQUEST_TEMPLATE.md: clarify sha change with "I"

vitorgalvao:
Trying to make the message even clearer:

+ Removed `{{link}}`:
  + Few people actually use it.
  + It’s inconsistent with the rest of the template where `{{}}` means something you change *locally*, not in the template itself.
  + Confirmation is frequently accompanied by more than just a link.
+ Made the message about providing confirmation more personal.
+ “What is this” instead of “how”. More “in your face” regarding “we know this might be confusing, so see a longer explanation here”.

To do after merging:

- [ ] Update in `cask-repair`
- [ ] Run the sync script

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,8 +8,8 @@ After making all changes to the cask:
 
 Additionally, if **updating a cask**:
 
-- [ ] If the `sha256` changed but the `version` didn’t,  
-      provide public confirmation ([How?][version-checksum]): {{link}}
+- [ ] `sha256` changed but the `version` didn’t ([what is this?][version-checksum]).  
+      I’m providing public confirmation below.
 
 Additionally, if **adding a new cask**:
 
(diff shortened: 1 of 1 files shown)


commitay:
I realise this was argued against in https://github.com/Homebrew/brew/pull/2601 but I'd like to have `only_sha_changed` enabled for local `audit` to warn contributors of this requirement. 

This should be something that alerts the contributor __*before*__ opening a PR. 

With `cask-repair` PRs we don't know if the contributor even sees the PR template after submitting a PR.

Maintainers can comment on the PR requesting verification after it is submitted but I'd like us to consider a more proactive approach.

vitorgalvao:
@commitay Alright. Your argument makes sense. It’s fine by me.

Pinging @reitermarkus.

reitermarkus:
Also fine by me.

vitorgalvao:
https://github.com/caskroom/homebrew-cask/issues/37572

PR #37568, opened 2017-08-10
Update Hyper.app to v1.4.0

chrisfinazzo:


chrisfinazzo:
I tested this offline and found that `https://github.com/zeit/hyper/archive/1.4.0.zip` works. This returns a 302 and redirects to `https://codeload.github.com/zeit/hyper/zip/1.4.0`. Seems like this should be correct URL.

Thoughts?

commitay:
> `https://github.com/zeit/hyper/archive/1.4.0.zip`

This is the source, not a zipped `app`. 

https://github.com/zeit/hyper/releases and https://hyper.is/ still point to `1.3.3` as the current `app` release.

PR #37593, opened 2017-08-11
Update openzfs - uninstall_preflight

commitay:

___

Uses ruby instead of `sed` to edit the `uninstall script`

### DIFF
--- a/Casks/openzfs.rb
+++ b/Casks/openzfs.rb
@@ -22,10 +22,9 @@ cask 'openzfs' do
 
   if MacOS.version >= :el_capitan
     uninstall_preflight do
-      system_command '/usr/bin/sed',
-                     args: ['-i', '.bak', 's|/usr/sbin/zpool|/usr/local/bin/zpool|', "#{staged_path}/Docs & Scripts/uninstall-openzfsonosx.sh"]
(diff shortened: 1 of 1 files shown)


miccal:
> Uses ruby instead of sed to edit the uninstall script

Why?

commitay:
I thought ruby methods were preferred over `system_command`?

miccal:
Ping @vitorgalvao - I am not sure myself which is preferred.

vitorgalvao:
Ruby is preferred because casks are themselves ruby code. The only times we should use `system_command` is when we can’t do something with ruby alone.

miccal:
Thanks for the clarification @vitorgalvao.

PS. I like your new avatar, but I think I will miss the  broccoli and pumpkin!

vitorgalvao:
> PS. I like your new avatar

Thank you.

> but I think I will miss the broccoli and pumpkin!

So will I. This one is quite busier and doesn’t compress as well to a tiny avatar, but it’s part of a later series I’ve been working on that I think ended quite well. Video of one of the processes (heavily compressed by twitter, unfortunately): https://twitter.com/vhgalvao/status/897984999743410176

PR #37733, opened 2017-08-17
Update pingplotter - zap

ruiokada:

Couldn't run `brew cask style --fix {{cask_file}}` since it kept crashing for some reason. Specifically I'm getting a seg fault from the `json` gem and it might be from mixing system Ruby and RVM Ruby. Instead I just ran rubocop on the file directly and it passed.


commitay:
Thanks for the PR!

PR #37862, opened 2017-08-21
Update issue templates

daylightbrightledlight:
Comment out a part so that it wouldn't be submitted

This is an updated template so that the comment can be seen when writing the issue, but it will not be previewed because it becomes repetitive to appear in every request.


daylightbrightledlight:
I hope you like the edit. Let me know what you think. I think it will clean up some stuff.

vitorgalvao (approved):
Didn’t know we could do this. Yes, if you find other documentation that would benefit, please add it as well.

commitay (approved):

vitorgalvao:
Alright, now there are too many changes. Making some parts comments is fine, but all the rest was carefully worded and exactly as intended after many iterations. Please revert everything except making the comments.

vitorgalvao (changes requested):
Please revert every change you made but making the comments (`<!-- -->`)

commitay (approved):
Other changes reverted.

PR #37968, opened 2017-08-24
Update murus.rb zaps

daylightbrightledlight:
I've added two items to the Murus zaps:

- `/private/etc/murus`
- `/private/etc/murus.sh`


daylightbrightledlight:
Actually `/private/etc` is the same directory as `/etc`, I think. Never mind.

PR #38146, opened 2017-08-31
Add duplicati.rb zaps

daylightbrightledlight:
I've added some zaps for Duplicati.


daylightbrightledlight:
I've closed this request. See https://github.com/caskroom/homebrew-cask/pull/38152 instead.

PR #38147, opened 2017-08-31
Update java.rb uninstall file

daylightbrightledlight:
I've added the file `com.oracle.java.Helper-Tool.plist` to be deleted when uninstalling Java.


daylightbrightledlight:
See https://github.com/caskroom/homebrew-cask/pull/38153 instead

PR #38148, opened 2017-08-31
Update vmware-fusion.rb zaps

daylightbrightledlight:
Add zap for `/Library/Logs/VMware USB Arbitrator Service.log`


daylightbrightledlight:
See https://github.com/caskroom/homebrew-cask/pull/38156 instead.

PR #38149, opened 2017-08-31
Update little-snitch.rb zap

daylightbrightledlight:
Added zap for `/Library/Logs/LittleSnitchDaemon.log`


daylightbrightledlight:
See https://github.com/caskroom/homebrew-cask/pull/38154 instead

PR #38150, opened 2017-08-31
Update real-vnc.rb

daylightbrightledlight:
Add zap for `/Library/Logs/vncserver.log.bak`


daylightbrightledlight:
See https://github.com/caskroom/homebrew-cask/pull/38155 instead

PR #38151, opened 2017-08-31
Update some zaps and uninstalls

daylightbrightledlight:
I've decided to combine some of the commits


daylightbrightledlight:
Actually I've separated the patches into their own pull requests again

PR #38157, opened 2017-08-31
Update teamviewer.rb

daylightbrightledlight:
I've added some uninstall files


commitay:
These are removed by `uninstall launchctl`

PR #38159, opened 2017-08-31
Update virtualbox.rb

daylightbrightledlight:
Update virtualbox.rb zaps


commitay:
This should be removed by the uninstall script?

Casks/virtualbox.rb
```diff
@@ -25,6 +25,7 @@
 
   zap delete: [
                 '/Library/Application Support/VirtualBox',
+                '/Library/LaunchDaemons/org.virtualbox.startup.plist',
```
commitay: This should be removed by the uninstall script?

PR #38250, opened 2017-09-04
Remove ship

kogir:

- [ ] The commit message includes the cask’s **name** and ~version~. No version because it now tracks `:latest`

The way we publish releases and their frequency is not conducive to SHA hashes. If you could validate code signatures using `codesign -v -R="identifier com.realartists.Ship2 and anchor apple generic" Ship.app` that would be awesome.


commitay:
This can be versioned with an `sha256` because of the `appcast`

Casks/ship.rb
```diff
@@ -1,9 +1,9 @@
 cask 'ship' do
-  version :latest
-  sha256 :no_check
+  version '2.6.4,702'
+  sha256 '358c2ea4f409f5688ac971086512b45fb55cb63cc11f166c5c72ca67ff407ce6'
```
kogir: These are brittle and will break again next release, as soon as this week.

Casks/ship.rb
```diff
@@ -12,9 +12,9 @@
...
-  zap delete: [
-                '~/Library/Application Support/com.realartists.Ship2',
-                '~/Library/Preferences/com.realartists.Ship2.plist',
-                '~/Library/RealArtists/Ship2',
-              ]
+  zap trash: [
+               "~/Library/Application Support/com.realartists.Ship#{version.major}",
+               "~/Library/Preferences/com.realartists.Ship#{version.major}.plist",
+               "~/Library/RealArtists/Ship#{version.major}",
+             ]
```
kogir: These files won't change even if we release later major versions, so the hard-coded `2`s were correct. Ship v1 was an exception.

Also the docs discourage use of `trash` - should they be updated?

Casks/ship.rb
```diff
@@ -1,9 +1,9 @@
 cask 'ship' do
-  version :latest
-  sha256 :no_check
+  version '2.6.4,702'
+  sha256 '358c2ea4f409f5688ac971086512b45fb55cb63cc11f166c5c72ca67ff407ce6'
 
-  url 'https://www.realartists.com/builds/2.0/Ship.app.zip'
-  appcast 'https://www.realartists.com/builds/2.0/sparkle.xml',
+  url "https://www.realartists.com/builds/#{version.major}.0/Ship.app.zip"
+  appcast "https://www.realartists.com/builds/#{version.major}.0/sparkle.xml",
```
kogir: These also won't change in the future, even if the major version does.

kogir (commented):
I think this commit defeats the purpose of my PR.

If Ship's cask (submitted by a user, not us) is going to break regularly, can we please just be de-listed? Having it broken here will only confuse users and lead to needless support inquiries. In fact, that's the origin of this PR: https://twitter.com/nhoizey/status/904642924067127296

kogir:
Sorry, just saw your comment. If `appcast` fixes things, why did it not work before? I'm obviously less familiar with cask than you and trust it will work if you say so.

commitay:
> If appcast fixes things, why did it not work before?

The version still has to be updated manually. See [updating a cask](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask)
____

Our users overwhelming prefer versioned downloads despite that it occasionally means a Cask is broken until it is updated.

commitay:
We can remove the cask if you prefer but we don't change versioned casks to use `latest` unless they repeatedly break (i.e the `sha256` changes without the `appcast` being updated.)

kogir:
> unless they repeatedly break (i.e the sha256 changes without the appcast being updated.)

I can arrange that 😃 
<hr />

In all seriousness, in that case it may be best to remove the cask. Please do. Sorry for the trouble, and thanks for your time and help.

commitay (approved):
Removed at the request of the developer .

vitorgalvao:
Agreed with the removal.

PR #38794, opened 2017-09-20
Remove boom

commitay:
Refs: https://github.com/caskroom/homebrew-cask/pull/38791#issuecomment-330986641


reitermarkus:
You can still buy Boom 2, so I'm not sure about this.

commitay:
I guess while you can still buy it it's more of an "edition" and belongs here rather than a legacy "version".

PR #38807, opened 2017-09-21
Remove jce-unlimited-strength-policy

commitay:
Java Cryptography Extension Policy Files are included in Java 1.9 

> https://docs.oracle.com/javase/9/migrate/#GUID-D6EE05FB-6791-43B3-A610-3F4416DEE508

> If your application previously required the Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy Files, then you no longer need to download or install them. They are included in the JDK and are activated by default.


pavel-sakun:
@commitay What's for 1.8? It still does require JCE. Wouldn't it better renaming the cask or any other approach except just deleting it?

commitay:
https://github.com/caskroom/homebrew-versions/pull/4633

pavel-sakun:
@commitay thank you! We're using this in our product so that it's good to know about upcoming changes.

PR #38951, opened 2017-09-25
Travis CI

commitay:
Refs: https://github.com/caskroom/homebrew-cask/issues/38950

- `xcode9` 
- vendored ruby

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: true
 matrix:
   include:
     - os: osx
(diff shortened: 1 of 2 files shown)


commitay:
It seems we have to unset this because we want `brew` to use its own vendored ruby.

```
# ensure that brew uses the ruby we want it to
run export HOMEBREW_RUBY_PATH="${SYSTEM_RUBY_BINDIR}/ruby"
```

PR #39043, opened 2017-09-27
Update catlight: remove `depends_on macos`

commitay:
Reverts https://github.com/caskroom/homebrew-cask/pull/39038

I'm 👎  on adding these to Casks that don't _currently_ work because of a bug.  In most cases it will only be a temporary problem that will be fixed in a new version. People may also want to install the Cask so they can report issues upstream.

I think this should only be added if upstream says they aren't going to support 10.13 at all.

### DIFF
--- a/Casks/catlight.rb
+++ b/Casks/catlight.rb
@@ -6,7 +6,5 @@ cask 'catlight' do
   name 'catlight'
   homepage 'https://catlight.io/'
 
-  depends_on macos: '<= 10.12'
-
   app 'Catlight.app'
 end


OranShuster:
i guess... however,
1.for that version the app doesn't work at all. there is nothing to "report upstream"
2.a caveat could have been added to explain that *in this current version* it does not support 10.13

commitay:
> 1.for that version the app doesn't work at all. there is nothing to "report upstream"

I was referring to other Casks that have may issues on `10.13`, not this Cask in particular.

> a caveat could have been added to explain that in this current version it does not support 10.13

`caveats` are usually reserved for installation related information. Adding caveats to Casks that don't work on 10.13 (and removing them when they do work) is a lot of effort for what would likely be a short term problem.

miccal:
I think it is fine to add this to Casks that do not currently work - the added `depends_on` can easily be removed one the Cask is updated to a new version.

commitay:
But that means _we_ are deciding based on user reports that it doesn't work. Unless upstreams say they don't support 10.13 we shouldn't make that decision.

> the added `depends_on` can easily be removed once the Cask is updated to a new version.

That assumes someone actually remembers to remove it.

miccal:
> But that means we are deciding based on user reports that it doesn't work. Unless upstreams say they don't support 10.13 we shouldn't make that decision.

In this case I thought the developers had stated it did not work on High Sierra?

I agree, we should only add this if the developer states it is not supported.

> Maintainers can always check this before merging.

commitay:
> In this case I thought the developers had stated it did not work on High Sierra?

The bug report is a day old and the developers have already identified the issue.

> Maintainers can always check this before merging.

That mean _**we**_ have to verify that it does work!
___

10.13 has only been out a couple of days, I'd prefer to wait at least a couple of weeks (or longer) before we start adding `depends_on` to cover these cases.

miccal:
Agreed - I think we should wait for, say 4 weeks, from the official release date of High Sierra, and then start adding `depends_on`.

On a side note, I upgraded one of my machines to High Sierra yesterday, and all `brew` formulas and all `casks` are working for me.

commitay:
> On a side note, I upgraded one of my machines to High Sierra yesterday, and all brew formulas and all casks are working for me.

I've upgraded one and done clean installs on another two. No issues with `brew` or `cask`.

OranShuster:
i only found problems with python/pip install
any package that required building, clang was failing
had to reinstall python (that upgraded from 27.13 to 17.14) and reinstalled xcode command line tools

PR #39122, opened 2017-09-30
Upgrade OCLint to v0.13

ryuichis:

### DIFF
--- a/Casks/oclint.rb
+++ b/Casks/oclint.rb
@@ -1,11 +1,11 @@
 cask 'oclint' do
-  version '0.12,16.5.0'
-  sha256 '12fda8c59ddd1896880a3055b701850d89550ea686b8d82aa94fffb74220950d'
+  version '0.13,16.7.0'
+  sha256 '7a8a2645e47d069e17b9239e8a8f7bfee642124337ac5b451bc1ed2accd7da66'
 
   # github.com/oclint/oclint was verified as official when first introduced to the cask
   url "https://github.com/oclint/oclint/releases/download/v#{version.before_comma}/oclint-#{version.before_comma}-x86_64-darwin-#{version.after_comma}.tar.gz"
   appcast 'https://github.com/oclint/oclint/releases.atom',
-          checkpoint: '4f8feba9928bb7318c46a3b7d4eeb798f7b6b30cbdfa7bfe86c13eff09790757'
+          checkpoint: 'e95e5635fdec4c6c86f8c4a39ea670d387fcb98142df47325bbd049de42dc031'
   name 'OCLint'
   homepage 'http://oclint.org/'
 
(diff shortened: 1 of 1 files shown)


Casks/oclint.rb
```diff
@@ -1,17 +1,18 @@
...
-          checkpoint: '4f8feba9928bb7318c46a3b7d4eeb798f7b6b30cbdfa7bfe86c13eff09790757'
+          checkpoint: 'e95e5635fdec4c6c86f8c4a39ea670d387fcb98142df47325bbd049de42dc031'
   name 'OCLint'
   homepage 'http://oclint.org/'
 
   binary "oclint-#{version.before_comma}/bin/oclint"
   binary "oclint-#{version.before_comma}/bin/oclint-json-compilation-database"
   binary "oclint-#{version.before_comma}/bin/oclint-xcodebuild"
   binary "oclint-#{version.before_comma}/lib/oclint", target: "#{HOMEBREW_PREFIX}/lib/oclint"
+  binary "oclint-#{version.before_comma}/lib/clang", target: "#{HOMEBREW_PREFIX}/lib/clang"
```
miccal: Are you sure this will not conflict with anything from Homebrew-Core? `/usr/local/lib/clang` seems like something that might already exist for some users?

commitay: AFAIK all formulas that would have a conflict are keg-only.

ryuichis: @miccal I wondered if it's possible to specify a version of `clang` as dependency? so that if user has installed clang, then we can just use that (because the ones bundled by oclint are the same thing as the ones by clang); if user doesn't have clang installed, then we install it as a dependency. That way, we can remove this line.

@commitay can you explain what is `keg-only`?

commitay: > https://github.com/Homebrew/homebrew-core/blob/master/Formula/llvm.rb#L132
```
This formula is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.
```

ryuichis: I see. I am fine with making this keg-only if the three `oclint-*` binaries can be invoked with their names directly. e.g., I don't want users to type the full path `/usr/local/bin/oclint`, instead, they should use the tool with `oclint` only. Can you confirm? And guide me with the suggested changes?

commitay: keg-only is a type of Formula it does not apply to Casks.

ryuichis: @commitay got you. Then is there a way that we can mark this formula depends on clang? So that we can ensure at the time `oclint` is installed, the correct version of `clang` is also installed? 

Another solution might be checking the existence of clang, when clang is absent, this formula can copy over the ones from its own package.

commitay: This is a Cask not a Formula. Formula live in https://github.com/Homebrew/homebrew-core

> Then is there a way that we can mark this formula depends on clang? So that we can ensure at the time `oclint` is installed, the correct version of `clang` is also installed? 

No. I think this might be possible if this was a Formula in `homebrew/core` but we can't do it with Cask.

> Another solution might be checking the existence of clang, when clang is absent, this formula can copy over the ones from its own package.

To install Homebrew you need to have the CLT package or Xcode installed. I believe both of these include clang.

commitay (approved):

PR #39130, opened 2017-09-30
Update google-backup-and-sync Casks: conflict

commitay:
Pending https://github.com/caskroom/homebrew-cask/pull/39129

### DIFF
--- a/Casks/google-backup-and-sync.rb
+++ b/Casks/google-backup-and-sync.rb
@@ -6,7 +6,10 @@ cask 'google-backup-and-sync' do
   name 'Google Backup and Sync'
   homepage 'https://www.google.com/drive/download/'
 
-  conflicts_with cask: 'google-photos-backup-and-sync'
+  conflicts_with cask: [
+                         'google-photos-backup-and-sync',
+                         'google-drive-file-stream',
+                       ]
   depends_on macos: '>= :mavericks'
 
   app 'Backup and Sync.app'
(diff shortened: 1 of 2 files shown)


commitay:
`install` fails because of conflict.
```
>>> brew cask _audit_modified_casks 2a827fffa892ca082cdfe1be8a371c8adb622bb3...5b68967d5153072ed9366a34679fe0ea5da458b5
==> 2 modified casks: google-backup-and-sync google-photos-backup-and-sync
audit for google-backup-and-sync: passed
audit for google-photos-backup-and-sync: passed

>>> brew cask style Casks/google-backup-and-sync.rb Casks/google-photos-backup-and-sync.rb
2 files inspected, no offenses detected
```

guaguanco:
I don't understand why is conflicting with 'google-drive-file-stream', I'm currently using both on my mac (one with the personal and the other with the work accounts) and I don't get any error or warning.

I understand that it wouldn't be wise to use both with the same account, but that's up to the user and it's not a conflict between the two packages

MTCoster:
I have both a personal Google account and a GSuite account for work.  Is it really not possible to install both apps side by side?

guaguanco:
@MTCoster I'm using both together and they happily co-exist, I just had to install one of them the old way because of the Homebrew recipe.

I assume that Google recommends to uninstall B&S to prevent naive users to sync the same account with two different tools.

PR #39143, opened 2017-09-30
update Principle to latest

marcobiedermann:

Updated Principle to latest, so no need for checksum

### DIFF
--- a/Casks/principle.rb
+++ b/Casks/principle.rb
@@ -1,11 +1,9 @@
 cask 'principle' do
-  version '2.1.6'
-  sha256 '1648d2fd4693770f91bcb2c5cc00b54b6b9d3b416e3915d8ee7168ebbbb0c955'
+  version :latest
+  sha256 :no_check
 
   # principleformac.com.s3.amazonaws.com was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


vitorgalvao:
No cask *needs* a checksum. Most have them because users prefer them, in part for security reasons.

Thank you anyway.

marcobiedermann:
@vitorgalvao 
How about updating Principle? Currently used: 2.1.6, latest is 3.2.

Also when running `brew cask install principle` I get a checksum error.
That's why I removed the checksum to always support latest version.

vitorgalvao:
> How about updating Principle? Currently used: 2.1.6, latest is 3.2.

[Already been done](https://github.com/caskroom/homebrew-cask/pull/39146).

> That's why I removed the checksum to always support latest version.

Again, thank you for the PR, but that’s not the way we do it. If a cask is outdated we update it. We also don’t remove appcasts unless they’re no longer accurate or give problems.

marcobiedermann:
@vitorgalvao 
Thanks for updating.
Just a quick question: How do I get the hash for checkpoint of appcast?

vitorgalvao:
https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md

PR #39202, opened 2017-10-01
Upgrade Paperspace from latest to 5.4.0.20

kirktrue:
Paperspace download URLs switched to including the specific version, so
now including that along with the new checksum.

`sha256` changed because the `version` changed.


commitay:
Thanks for the PR!

PR #39481, opened 2017-10-10
Update PULL_REQUEST_TEMPLATE.md: sha256

commitay:


vitorgalvao:
Nice idea of changing “updating a cask” to “updating the `sha256` only”. Makes the whole section more obvious.

Made a few consistency changes (“if” and “:” not bold) and shortened the link (it was such a long stretch, I feel it would be ignored as it looked more like blue text than a link).

PR #39512, opened 2017-10-10
Update mono-mdk to 5.2.0.224

commitay:
Update `url`

Reverts https://github.com/caskroom/homebrew-cask/pull/39495


commitay:
Verified locally.

PR #39557, opened 2017-10-12
Delete sticker.rb

miccal:
Dead `homepage` and `url`.


commitay:
The homepage still works, the download has moved to devmate.

miccal:
Strange, the homepage is not working for me.

PR #39631, opened 2017-10-14
Update OpenMW to 0.42.0

benwebber:
OpenMW no longer packages the software in a suite.

The DMG also includes a game editor (OpenMW CS) as a separate app. I'd like to provide that through Homebrew-Cask as well.

1. Is it possible to install multiple apps in the same cask?
2. If (1) is not recommended, would you prefer a separate cask for the game editor?
3. Does Homebrew-Cask support cask options like Homebrew does? For example:

    ```
    brew cask install openmw --with-cs
    ```

---

### DIFF
--- a/Casks/openmw.rb
+++ b/Casks/openmw.rb
@@ -10,4 +10,5 @@ cask 'openmw' do
   homepage 'https://openmw.org/'
 
   app 'OpenMW.app'
+  app 'OpenMW-CS.app'
 end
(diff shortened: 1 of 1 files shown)


commitay:
> 1. Is it possible to install multiple apps in the same cask?

Yes

```
  app 'OpenMW.app'
  app 'OpenMW-CS.app'
```

> 3. Does Homebrew-Cask support cask options like Homebrew does? 

No

PR #39633, opened 2017-10-14
New Cask Mac-Cantonese-Input

hanxue:
I see that the `openvanilla` Cask is an `input_method`. This cask similarly is an input_method, and needs to be copied over to `~/Library/Input\ Methods/`

Could you help me to write the proper Cask formula? Thank you.


commitay:
This isn't notable enough to be included in Cask. 

[Rejected Casks](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/rejected_casks.md#rejected-casks)

I'd suggest [creating your own tap.](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md)

PR #39701, opened 2017-10-16
Update sogouinput to 4.3.0.4743

xiangzhuyuan:

Additionally, if **updating only the `sha256`**:

- [x] I verified this change is legitimate [<sup>How do I do that?</sup>](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256) and **am providing confirmation below**:


commitay:
The homepage and `url` should remain http://pinyin.sogou.com/mac/

xiangzhuyuan:
ok, fine! Actually current vision is temporary, since we can not access the official download url.
and why I try to change that homepage url since it offense some rules here.

ok drop this pr. @commitay tks

PR #39704, opened 2017-10-16
Fix get-iplayer-automator appcast checkpoint

jabley:
Not sure what's changed, but it's not happy.


commitay:
I've merged this but we've been trying to discourage `checkpoint` only PRs. 

A `checkpoint` mismatch doesn't impact impact installing the Cask.

PR #39707, opened 2017-10-16
Add protonmail-unofficial 0.5.2

neilbartley:

### DIFF
--- a/Casks/protonmail-desktop.rb
+++ b/Casks/protonmail-desktop.rb
@@ -1,0 +1,12 @@
+cask 'protonmail-desktop' do
+  version '0.5.1'
+  sha256 'a02877dcc1760f75133245f12eb6768eb5c4201b152e2b56ae61720146e484ea'
+
+  url "https://github.com/protonmail-desktop/application/releases/download/v#{version}/protonmail-desktop-#{version}.dmg"
+  appcast 'https://github.com/protonmail-desktop/application/releases.atom',
+          checkpoint: 'c9eb231a43b5d3d14a33925ddb979afae31b5a642881e5627bac0b16bb8da4ce'
+  name 'Protonmail Desktop'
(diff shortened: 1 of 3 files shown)


vitorgalvao:

Next time please don’t collapse the checkboxes. Keeping the template as was makes it easier to parse what you ticked.

vitorgalvao (approved):

vitorgalvao:
On hold for https://github.com/caskroom/homebrew-cask/pull/39708.

commitay:
@vitorgalvao As this is an unofficial app I'm not sure about calling this just `protonmail`.

I'm not forgetting "not a discoverability service", but it would be easy to mistake this as an offical app if you only inspect the Cask (brew cask cat). 

Protonmail also has a Github https://github.com/protonmail and an app https://github.com/ProtonMail/WebClient (has to be built)

Normally I would say it's not an issue but as this is an app for an encrypted email service I feel that explicitly calling this unofficial would be appropriate.

vitorgalvao:
@commitay That’s a good point. It actually falls into a new rule I had been thinking about today due to https://github.com/caskroom/homebrew-cask/pull/39706.

I was thinking that every unofficial app that uses the official name unofficially (i.e. without permission from the developers) should mandatorily have the fork name prefix, regardless of it there is another app with its name or not.

commitay:
> I was thinking that every unofficial app that uses the official name unofficially (i.e. without permission from the developers) should mandatorily have the fork name prefix, regardless of it there is another app with its name or not.

I think this is a good idea that should cover most cases but sometimes the fork / Github org name duplicates the `app` name or would otherwise give the appearance of being official. When this occurs (this Cask) I'd suggest adding an `-unoffical` suffix.

vitorgalvao:
Let’s make it on hold again for https://github.com/caskroom/homebrew-cask/issues/39739.

commitay (approved):

PR #39755, opened 2017-10-17
Migrate 3 Casks to drivers

commitay:
https://github.com/caskroom/homebrew-drivers/pull/194


boboldehampsink:
Why move `discovercare`? We've had a discussion about this in the original pull request: #30287

PR #40038, opened 2017-10-25
Move Native Access to subfolder

danielbayley:
Most NI products install to `/Applications/Native Instruments`.

### DIFF
--- a/Casks/native-access.rb
+++ b/Casks/native-access.rb
@@ -6,5 +6,5 @@ cask 'native-access' do
   name 'Native Access'
   homepage 'https://www.native-instruments.com/en/specials/native-access/'
 
-  app 'Native Access.app'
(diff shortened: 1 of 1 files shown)


danielbayley:
I feel like this should use the [`suite` stanza](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/suite.md), but it fails if there is another NI app already installed in that folder…

miccal:
Cask tries to mimic what installing the `.app` manually would do, and in this case opening the `.dmg` file gives:

screen shot 2017-10-26 at 07 13 34

So by default the `.app` is **not** installed to `/Applications/Native Instruments`, so I do not think we should be changing this behaviour.

vitorgalvao:
Agreed with @miccal. Same point made in https://github.com/caskroom/homebrew-cask/pull/40039#issuecomment-339502640.

PR #40058, opened 2017-10-26
Update ISSUE_TEMPLATE.md

commitay:
Reverts https://github.com/caskroom/homebrew-cask/pull/40018 in favour of adding macOS and Java versions to `brew cask doctor`.

https://github.com/Homebrew/brew/commit/1d5e1afa28517d3e8445c29b7e33c6a5570f4f6f
___

Most of the output of `brew config` and `brew doctor` isn't relevant to Cask.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -22,15 +22,3 @@
 ```
 {{replace this}}
 ```
-
-#### Output of `brew doctor`
-
-```
-{{replace this}}
-```
-
-#### Output of `brew config`
-
-```
-{{replace this}}
-```


miccal:
I think it is worthwhile having all three outputs of `brew cask doctor`, `brew doctor` and `brew config`.

commitay:
What output from `brew doctor` and `brew config` do you think is useful?

It could be added to `brew cask doctor` rather than expecting (hoping) that users provide all three.

miccal:
The more info we have on a user's `brew` configuration the better in my opinion.

vitorgalvao:
> The more information we have on a user's `brew` configuration the better in my opinion.

I do think there’s a point of too much information.

Remember we did use to ask for a lot more, and at that time the issue template was filled with `<details>` tags, because all the output was so extensive, we needed that organisation. But `<details>` is harder for users to understand (it’s a little-known tag) when filling the template. It’s also more trouble for us, having to constantly expand the tags.

I’d rather strike a balance of asking upfront for the information that’ll be short and enough 90% of the time, and then asking for more when it’s needed.

miccal:
Fine, reverting.

PR #40065, opened 2017-10-26
New Cask: gfortran 6.1

jeroen:

---

Provides the official GNU gfortran builds. Technically, this is a dependency of the `r-app` Cask, but in practice it is only needed to install R packages with fortran code. More details are given on the [r-project website](https://cran.r-project.org/bin/macosx/tools/) in particular this fragment:

> __Important note__: R 3.4.0 El Capitan binaries are using Clang 4.0.0 and GNU Fortran 6.1 to provide OpenMP parallelization support and C++17 standard features. If you want to compile R packages from sources, please download GNU Fortran binary from the official GNU Fortran Binaries page - in particular OS X 10.11 gfortran 6.1. 

The build gfortran that ships with homebrew `gcc` is incompatible with the gfortran runtime that ships with `r-app`. Thanks for considering.

### DIFF
--- a/Casks/gfortran.rb
+++ b/Casks/gfortran.rb
@@ -1,16 +1,22 @@
 cask 'gfortran' do
-  version '6.1'
-  sha256 'eb817bce64bf9032595e09166bdaaf740c83bf7258f900b79cd6786437bacbf4'
-
   # coudert.name/software was verified as official when first introduced to the cask
-  url "http://coudert.name/software/gfortran-#{version}-ElCapitan.dmg"
+  if MacOS.version == :el_capitan
+    version '6.1'
+    sha256 'eb817bce64bf9032595e09166bdaaf740c83bf7258f900b79cd6786437bacbf4'
+    url "http://coudert.name/software/gfortran-#{version}-ElCapitan.dmg"
+    pkg "gfortran-#{version}-ElCapitan/gfortran.pkg"
+  else
+    version '6.3'
+    sha256 '38b81bc878dba41cfdbb0c335aec5a97554a5d1766fb3e3ca6be7da0df9e8e09'
+    url "http://coudert.name/software/gfortran-#{version}-Sierra.dmg"
+    pkg 'gfortran.pkg'
+  end
+
   name 'gfortran'
   homepage 'https://gcc.gnu.org/wiki/GFortranBinaries'
 
   conflicts_with formula: 'gcc'
   depends_on macos: '>= :el_capitan'
-
-  pkg "gfortran-#{version}-ElCapitan/gfortran.pkg"
 
   uninstall delete:  [
                        '/usr/local/gfortran',
--- a/Casks/gfortran.rb
+++ b/Casks/gfortran.rb
@@ -7,14 +7,16 @@ cask 'gfortran' do
   name 'gfortran'
   homepage 'https://gcc.gnu.org/wiki/GFortranBinaries'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you for the submission.

This seems to be a CLI-only open-source tool. Please submit it first to [Homebrew/core](https://github.com/Homebrew/homebrew-core) and only if they don’t accept it, submit here (linking back to the pull request where it was refused).

jeroen:
I don't understand. Should I submit the cask to core? The problem is that `gfortran` is part of `gcc` which is already in homebrew, but we would like to provide the official binary build.

commitay:
@jeroen Is `6.1` needed specifically or should be be dependent on macOS version?

jeroen:
We specifically like use the 6.1 version. This build targets el-capitain so it works on all currently supported versions of MacOS (10.11, 10.12, 10.13).

commitay:
@vitorgalvao I think this would be better suited in a personal tap because the use case won't support the usual `if MacOS.version` and `6.1` is not the latest release.

jeroen:
The 6.1 build is the latest version that works on all currently supported MacOS versions. The latest build from [the website](https://gcc.gnu.org/wiki/GFortranBinaries#MacOS) is 6.3 for Sierra. We could conditionally install the 6.3 version on Sierra and up. For simplicity R just uses the 6.1 build so that we have a single toolchain across different MacOS users, but either way is fine by me.

Sure I can host it in my own tap, but I do think that the official Fortran for MacOS (that does not require the entire gcc) may be more useful to other users, not just me :)

commitay:
Can you add the Sierra version, `conflicts with formula: 'gcc'` and `pkgutil` to the uninstall please?

miccal:
screen shot 2017-10-27 at 10 47 29

> To install any of them, you need to have the Apple Developer Tools installed, as well as "command line developer tools".

So does this need the full Xcode, or are the CLT enough?

miccal:
screen shot 2017-10-27 at 10 50 20

jeroen:
I'v been using it on a VM with only CLT and it works fine as far as I can tell.

commitay:
@vitorgalvao What do you think about adding this to `versions` as `gfortran61`?

jeroen:
I have conditioned the version on the MacOS version in the [last commit](https://github.com/caskroom/homebrew-cask/pull/40065/commits/c90cd8a6b4d2b7c542b6d4ffa9a29d9bb9722803). Not sure I like it but it seems to work.

Casks/gfortran.rb
```diff
@@ -0,0 +1,30 @@
+cask 'gfortran' do
+  # coudert.name/software was verified as official when first introduced to the cask
```
vitorgalvao: This should be located directly above `url`, even if it means repeating it. Each URL is verified independently.

vitorgalvao (changes requested):

Casks/gfortran.rb
jeroen: You sure? It seems to pass the audit. Anyway will update.

vitorgalvao:
> @vitorgalvao What do you think about adding this to `versions` as `gfortran61`?

If there’s a reason a Sierra (and up) user would want that version, yes. Otherwise, one this one should be enough. I’m doubting the usefulness (amount of people that’ll take advantage of) this cask already as it is; it’s harder to justify two casks across two repos.

commitay (approved):

PR #40109, opened 2017-10-27
Keybase 1.0.34 20171027003854

ondrejfuhrer:

- [x] `brew cask audit --download keybase` is error-free.
- [x] `brew cask style --fix keybase` reports no offenses.


ondrejfuhrer:
I also changed the URL to the one, that is mentioned in the `app_digest` (*.zip version) so it is possible to check the SHA against the value inside the `app_digest`.

It that is for whatever reason not wanted, feel free to remove the last commit (the commit before works correctly with the *.dmg URL)

PR #40113, opened 2017-10-27
Anaconda auto updates

domoritz:

### DIFF
--- a/Casks/anaconda.rb
+++ b/Casks/anaconda.rb
@@ -6,9 +6,8 @@ cask 'anaconda' do
   url "https://repo.continuum.io/archive/Anaconda3-#{version}-MacOSX-x86_64.sh"
   name 'Continuum Analytics Anaconda'
   homepage 'https://www.anaconda.com/what-is-anaconda/'
-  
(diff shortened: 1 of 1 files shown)


commitay:
@domoritz Does this "auto update" i.e prompt for updates / silently install updates or do you need to run a `conda update` command?

domoritz:
Good point. I run `conda update --all` from time to time and so my anaconda is up to date even though cask thinks it is not. I don't think anaconda auto updates in the background (it might pop up a dialog, but not sure).

So what's the right setting for a cask that does not auto update in the background but that does have an update mechanism?

commitay:
> So what's the right setting for a cask that does not auto update in the background but that does have an update mechanism?

We don't have a setting to cover this case. 

We have a few command line Casks that can self update but because this requires an explicit `update` / `upgrade` command we don't set `auto_update`. Most of our users depend on Cask to handle the updates.

domoritz:
Thanks for the explanation. I have to check whether anaconda shows a popup or not. I will close this PR until then.

PR #40127, opened 2017-10-28
Add Korg M1 1.7.1

danielbayley:

### DIFF
--- a/Casks/m1.rb
+++ b/Casks/m1.rb
@@ -1,0 +1,20 @@
+cask 'm1' do
+  version '1.7.1'
+  sha256 '222d6fad47d09baae310767ba67ee822de1842ac24546bd1768144fc1da465af'
+
+  # korg-datastorage.jp was verified as official when first introduced to the cask
+  url "http://korg-datastorage.jp/Update/KLCDE/klc_m1_mac_#{version.no_dots}.dmg"
+  name 'Korg M1'
+  homepage 'http://korg.com/products/software/korg_legacy_collection/page_4.php'
+
+  pkg 'KLC M1 Installer.pkg'
+
+  uninstall pkgutil: 'jp.co.korg.pkg.M1',
+            delete:  [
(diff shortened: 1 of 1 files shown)


Casks/m1.rb
```diff
@@ -0,0 +1,20 @@
+cask 'm1' do
+  version '1.7.1'
+  sha256 '222d6fad47d09baae310767ba67ee822de1842ac24546bd1768144fc1da465af'
+
+  # korg-datastorage.jp was verified as official when first introduced to the cask
+  url "http://korg-datastorage.jp/Update/KLCDE/klc_m1_mac_#{version.no_dots}.dmg"
```
commitay: Is this URL public?

danielbayley: Define public? These are the official servers and don't seem to require a login to download. This is one of the official mirrors they list, the other is `https://korguser.net/downloadfile/files/klc_m1_mac_#{version.no_dots}.dmg`, which might seem slightly more 'proper' on the face of it, but for some reason that one requires `--ciphers DEFAULT` to be passed to `curl`, making it difficult to use in a cask properly. It also seems to be a lot slower. Once downloaded, it can be registered without any additional download as per the rules…

commitay: This there a public page with this URL or is it behind a login?

danielbayley: Hmm, the links seem to be presented after login actually, although the urls don't contain anything special. Looking at _Rejected Casks_ again; perhaps that counts as being behind a login/registration form, but not sure about the host technically differing from the homepage? They're all Korg servers…

vitorgalvao: > They're all Korg servers

There’s no way for a user to know that without considerable effort. As far as they know, the URL domain was setup by a malicious actor to look genuine.

Seems like this is indeed a walled build. You may submit to [`alehouse/homebrew-unofficial`](https://github.com/alehouse/homebrew-unofficial), if you want.

PR #40153, opened 2017-10-30
Link Zsh completion for Docker CLI

FranklinYu:

Version is not included because this does not involve a new version. It only installs completions to proper location.

If this looks good then I may also update `docker-edge`.

### DIFF
--- a/Casks/docker.rb
+++ b/Casks/docker.rb
@@ -14,7 +14,18 @@ cask 'docker' do
 
   app 'Docker.app'
 
-  uninstall delete:    '/Library/PrivilegedHelperTools/com.docker.vmnetd',
+  preflight do
+    ['docker', 'docker-compose', 'docker-machine'].each do |cmd|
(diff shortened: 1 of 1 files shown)


commitay:
This is already provided by the docker competition formulas.

```
docker-completion
docker-compose-completion
docker-machine-completion
```

franklinyu:
Sorry, I didn't know that. But why do we maintain another three formula when Docker.app comes with the completion already? The approach feels like more work to do… I guess there might have been some discussion about it?

commitay:
The formulas have a better way of installing completions and work with all Docker Casks.

```
  def install
    bash_completion.install "components/cli/contrib/completion/bash/docker"
    fish_completion.install "components/cli/contrib/completion/fish/docker.fish"
    zsh_completion.install "components/cli/contrib/completion/zsh/_docker"
  end
```

franklinyu:
Just curious, is there any Docker Casks besides `docker` and `docker-edge`?

PR #40278, opened 2017-11-03
Update ThingsMacHelper to 3.9

mdavis1982:


commitay:
The `app` name is different in finder, normally we would do something like this but I'm not sure if it will break anything in this case.

https://github.com/caskroom/homebrew-cask/blob/7bed7fb0041eec4988680ca9188155c51496c7e0/Casks/teamviewer-quickjoin.rb#L10-L11

commitay:
Thanks for the PR!

mdavis1982:
@commitay No problem 👍 It's my first time updating a cask, so any tips like that are really useful! Thank you 😄

PR #40400, opened 2017-11-07
Update atom to 1.21.2

miccal:


Casks/atom.rb
```diff
@@ -19,14 +19,16 @@
   zap delete: [
                 '~/.atom',
```
commitay: Is this user config? If it is it should probably be trashed.

PR #40513, opened 2017-11-09
Add mavericks version to hammerspoon

burakguven:
The version number isn't documented anywhere but based on my tests
0.9.46 is the latest version working on mavericks.

### DIFF
--- a/Casks/hammerspoon.rb
+++ b/Casks/hammerspoon.rb
@@ -1,6 +1,11 @@
 cask 'hammerspoon' do
-  version '0.9.57'
-  sha256 '33844074066446e98e97c10deb68e3934126b2e7ce740bc925ec58060ba59b92'
+  if MacOS.version <= :mavericks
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> The version number isn't documented anywhere

It needs to be. We specify (specified?) in the documentation that we don’t want to assume those versions numbers. We prefer to trust the official word of the developer.

Thank you for doing the work, though!

burakguven:
I just noticed that the minimum system version is [specified in the appcast file](https://github.com/Hammerspoon/hammerspoon/blob/master/appcast.xml#L163). It goes from 10.8 on version 0.9.46 to 10.10 on version 0.9.47. Does that count as sufficient documentation?

vitorgalvao:
> Does that count as sufficient documentation?

It does indeed!

commitay (approved):

commitay:
Thanks for the PR!

PR #40593, opened 2017-11-10
Update freecad to 0.16-6712.da2d364

gtmgianni:

### DIFF
--- a/Casks/freecad.rb
+++ b/Casks/freecad.rb
@@ -1,11 +1,11 @@
 cask 'freecad' do
-  version '0.16-6706.f86a4e4'
-  sha256 '9d05b21103bfe0dd7de2ec006fc4c1baca3e7ff32d4b570d91e627caded40178'
+  version '0.16-6712.da2d364'
+  sha256 '8b1d6410a2b276de393b99ff21bbd4888c31bcae6bdf0d400e198b291333f345'
 
   # github.com/FreeCAD/FreeCAD was verified as official when first introduced to the cask
-  url "https://github.com/FreeCAD/FreeCAD/releases/download/0.16/FreeCAD_#{version}-OSX-x86_64.dmg"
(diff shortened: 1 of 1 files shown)


Casks/freecad.rb
```diff
@@ -1,11 +1,11 @@
 cask 'freecad' do
-  version '0.16-6706.f86a4e4'
-  sha256 '9d05b21103bfe0dd7de2ec006fc4c1baca3e7ff32d4b570d91e627caded40178'
+  version '0.16-6712.da2d364'
+  sha256 '8b1d6410a2b276de393b99ff21bbd4888c31bcae6bdf0d400e198b291333f345'
 
   # github.com/FreeCAD/FreeCAD was verified as official when first introduced to the cask
-  url "https://github.com/FreeCAD/FreeCAD/releases/download/0.16/FreeCAD_#{version}-OSX-x86_64.dmg"
+  url "https://github.com/FreeCAD/FreeCAD/releases/download/0.16.6712/FreeCAD_#{version}-OSX-x86_64.dmg"
```
gtmgianni: Could someone more experienced in Ruby suggest a better method of substitution for this URL? At the moment, I'm having to manually type in the new version number.

gtmgianni: Could someone more experienced in Ruby suggest a better method of substitution for this URL? At the moment, I'm having to manually type in the new version number.

gtmgianni (commented):
A suggestion from someone more experienced with Ruby would be greatly appreciated (see line 6).

Casks/freecad.rb
commitay: I tried pushing a simple fix that I had to revert because it didn't work.

I'm okay with leaving this as is, looking at the releases page this `version` is not standard and anything we do to the `url` will break on the next release anyway.

PR #40603, opened 2017-11-11
Update little-snitch to 4.0.3

miccal:


miccal:
There are four more items to `zap`, namely:
```
/Library/StagedExtensions/Library/Extensions/LittleSnitch.kext
/Library/Extensions/LittleSnitch.kext
```
Should these be in an `uninstall kext:`?

commitay:
> Should these be in an `uninstall kext:`?

No, they are removed by the GUI uninstall and it seems to cause issues with the GUI uninstaller if we try to unload and remove them first.

Can you add them to `zap delete` with a comment please? 

> these kexts should not be uninstalled by Cask

PR #40617, opened 2017-11-11
Update virtualbox-extension-pack: uninstall_postflight

commitay:
https://github.com/caskroom/homebrew-cask/pull/40615

This allows the uninstall step to be skipped if the dependant Cask has been uninstalled, which would have removed the files installed by this Cask.

### DIFF
--- a/Casks/virtualbox-extension-pack.rb
+++ b/Casks/virtualbox-extension-pack.rb
@@ -24,6 +24,7 @@ cask 'virtualbox-extension-pack' do
   end
 
   uninstall_postflight do
+    next unless File.exist?('/usr/local/bin/VBoxManage')
     system_command '/usr/local/bin/VBoxManage',
                    args: [
                            'extpack', 'uninstall',


Casks/virtualbox-extension-pack.rb
```diff
@@ -24,6 +24,7 @@
   end
 
   uninstall_postflight do
+    next unless File.exist?('/usr/local/bin/VBoxManage')
```
vitorgalvao: Should be a `break` instead, as it’s not a loop.

commitay: `break` doesn't seem to work but I'm probably not using it correctly.

vitorgalvao: What about `return`?

commitay: `return` errors in both cases.

reitermarkus: It's a block, so it must be `next`.

PR #40654, opened 2017-11-13
Update 9 Casks: depends_on

commitay:
@vitorgalvao Not sure if the `depends_on` is meant to reoccur, it doesn't seem to make a difference?

### DIFF
--- a/Casks/cdock.rb
+++ b/Casks/cdock.rb
@@ -9,8 +9,8 @@ cask 'cdock' do
   name 'cDock2'
   homepage 'https://w0lfschild.github.io/cdock'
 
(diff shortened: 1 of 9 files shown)


vitorgalvao:
I think we decided on multiple `depends_on` as a stylistic choice. It’s not documented, though, so either way is fine with me, as long as we’re consistent.

commitay:
Just found the relevant discussion https://github.com/caskroom/homebrew-cask/issues/8784

The single value for each `depends_on` make sense for  future implementation of https://github.com/caskroom/homebrew-cask/issues/16383 and  https://github.com/caskroom/homebrew-cask/issues/26345

PR #40800, opened 2017-11-15
Update adobe-photoshop-lightroom: remove uninstall_preflight

commitay:
This is also break when https://github.com/Homebrew/brew/pull/3396 is merged.

As this is a patch installer on top of another Cask it doesn't need an `uninstall` itself, the `app` can be removed by uninstalling `lightroom600`.

The patch upgrade can be applied on top of the previous patch.

### DIFF
--- a/Casks/adobe-photoshop-lightroom.rb
+++ b/Casks/adobe-photoshop-lightroom.rb
@@ -27,10 +27,6 @@ cask 'adobe-photoshop-lightroom' do
                    sudo: true
   end
 
-  uninstall_preflight do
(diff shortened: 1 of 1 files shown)


commitay:
```
>>> brew cask _audit_modified_casks 5c0d832f7d5846170347cc6c573400160379ba5c...a4b7d4a84cd0833500349fcce205256e7f4857ff
==> 1 modified cask: adobe-photoshop-lightroom
audit for adobe-photoshop-lightroom: passed

>>> brew cask style Casks/adobe-photoshop-lightroom.rb
```

commitay:
Going to leave this as is for now. It won't actually "break" the upgrade, it will just install a clean version and then patch it.

PR #40897, opened 2017-11-17
Update screens-connect to 4.4.5450,1509106836

colindunn:

- [x] \`brew cask audit --download {{cask_file}}\` is error-free.
- [x] \`brew cask style --fix {{cask_file}}\` reports no offenses.

### DIFF
--- a/Casks/screens-connect.rb
+++ b/Casks/screens-connect.rb
@@ -1,11 +1,11 @@
 cask 'screens-connect' do
-  version '4.2'
-  sha256 '87546e86b647c03ccea2cf3f0f1dff0ab8adc33e19888077ea35956fb7848e09'
+  version '4.4.5450,1509106836'
+  sha256 'a4c530e8af65dd6b29ca96894356f9b815d86f8cdbe0ffe91c84446cd4907960'
 
-  # edovia.com was verified as official when first introduced to the cask
-  url "https://download.edovia.com/screensconnect/ScreensConnect_#{version}.zip"
+  # dl.devmate.com/com.edovia.Screens-Connect was verified as official when first introduced to the cask
+  url "https://dl.devmate.com/com.edovia.Screens-Connect/#{version.patch}/#{version.after_comma}/ScreensConnect-#{version.patch}.zip"
   appcast 'https://updates.devmate.com/com.edovia.Screens-Connect.xml',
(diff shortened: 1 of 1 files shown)


Casks/screens-connect.rb
```diff
@@ -21,7 +21,8 @@
                          'com.edovia.screens.connect',
                        ],
             script:    {
-                         executable:   'Screens Connect.app/Contents/Resources/sc-uninstaller.tool',
```
commitay: Should be able to prepend `#{appdir}` here

colindunn: Unfortunately I can't and this is the only way I could get it to work.

```
==> Running uninstall script /Applications/Screens Connect.app/Contents/Resources/sc-uninstaller.tool
==> Executing: ["/bin/chmod", "--", "+x", "/Applications/Screens Connect.app/Contents/Resources/sc-uninstaller.tool"]
==> Executing: ["/Applications/Screens Connect.app/Contents/Resources/sc-uninstaller.tool"]
Error: No such file or directory - /Applications/Screens
```

commitay: Okay, that an issue with the core. It's still skipping the script with `/bin/bash -c`
```
==> Running uninstall script /bin/bash
==> /bin/bash: /Applications/Screens: No such file or directory
```

colindunn: Do you want to add an "on hold" and I'll create a bug report then?

commitay: > Do you want to add an "on hold"

Done.

> and I'll create a bug report then?

That would be good, thanks!

commitay: Sorry, don't worry about the issue, turns out it does work.

colindunn: You cheated. `sudo` makes a difference here :-)

---

So it looks to me that when `.app` installs moved to /Applications instead of linking that this cask was broken because of that.

If an uninstall script is relative, the staged path is pre-pended to the executable which kept it as a full path, but a full path gets split by the uninstall script at processing time. Except when you put a sudo on it.

Suspect there are other broken casks because of this because they expected to be in staged? Maybe not

commitay: Found a way to reproduce it with and without sudo. https://github.com/caskroom/homebrew-cask/issues/40908

commitay: Merging as `sudo` is required for the `uninstall script` for this Cask.

commitay (approved):

PR #41111, opened 2017-11-21
script.sh: add pkgutil

commitay:
This is intended to help with checking `uninstall` stanzas that remove pkgs with `pkgutil` or `script`.

```
==> Uninstalling Cask powershell
==> Running uninstall process for powershell; your password may be necessary
==> Uninstalling packages:

>>> /usr/sbin/pkgutil --pkgs
com.hipbyte.rubymotion.pkg
com.microsoft.powershell
com.oracle.jdk8u112
com.oracle.jre
com.vmware.tools.macos.pkg.files
org.macosforge.xquartz.pkg
```

### DIFF
--- a/ci/travis/script.sh
+++ b/ci/travis/script.sh
@@ -25,6 +25,9 @@ elif [[ ${#modified_casks[@]} -gt 0 ]]; then
     for cask in "${modified_casks[@]}"; do
       run brew cask reinstall --verbose "${cask}"
       run brew cask uninstall --verbose "${cask}"
+      if brew cask _stanza uninstall "${cask}" &> /dev/null; then
+        run "$(brew --repository)/Library/Taps/caskroom/homebrew-cask/developer/bin/list_loaded_launchjob_ids"
+      fi
       if brew cask _stanza pkg "${cask}" &> /dev/null; then
         run "$(brew --repository)/Library/Taps/caskroom/homebrew-cask/developer/bin/list_recent_pkg_ids"
       fi
--- a/ci/travis/script.sh
+++ b/ci/travis/script.sh
@@ -26,7 +26,7 @@ elif [[ ${#modified_casks[@]} -gt 0 ]]; then
       run brew cask reinstall --verbose "${cask}"
       run brew cask uninstall --verbose "${cask}"
       if brew cask _stanza pkg "${cask}" &> /dev/null; then
-        run /usr/sbin/pkgutil --pkgs | /usr/bin/grep --ignore-case --invert-match apple | /usr/bin/sort
+        run "$(brew --repository)/Library/Taps/caskroom/homebrew-cask/developer/bin/list_recent_pkg_ids"
       fi
     done
   else
--- a/ci/travis/script.sh
+++ b/ci/travis/script.sh
@@ -26,7 +26,7 @@ elif [[ ${#modified_casks[@]} -gt 0 ]]; then
       run brew cask reinstall --verbose "${cask}"
       run brew cask uninstall --verbose "${cask}"
       if brew cask _stanza pkg "${cask}" &> /dev/null; then
-        run /usr/sbin/pkgutil --pkgs | /usr/bin/grep -iv apple | /usr/bin/sort
+        run /usr/sbin/pkgutil --pkgs | /usr/bin/grep --ignore-case --invert-match apple | /usr/bin/sort
       fi
     done
   else
--- a/ci/travis/script.sh
+++ b/ci/travis/script.sh
@@ -25,6 +25,9 @@ elif [[ ${#modified_casks[@]} -gt 0 ]]; then
     for cask in "${modified_casks[@]}"; do
       run brew cask reinstall --verbose "${cask}"
       run brew cask uninstall --verbose "${cask}"
+      if brew cask _stanza pkg "${cask}" > /dev/null 2>&1; then
+        run /usr/sbin/pkgutil --pkgs | /usr/bin/grep -iv apple | /usr/bin/sort
+      fi
     done
   else
     ohai 'More than 3 Casks modified, skipping install'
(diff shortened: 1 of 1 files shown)


ci/travis/script.sh
```diff
@@ -25,6 +25,9 @@ elif [[ ${#modified_casks[@]} -gt 0 ]]; then
     for cask in "${modified_casks[@]}"; do
       run brew cask reinstall --verbose "${cask}"
       run brew cask uninstall --verbose "${cask}"
+      if brew cask _stanza pkg "${cask}" &> /dev/null; then
```
vitorgalvao: Same result, less code.

ci/travis/script.sh
```diff
@@ -25,6 +25,9 @@ elif [[ ${#modified_casks[@]} -gt 0 ]]; then
...
+        run /usr/sbin/pkgutil --pkgs | /usr/bin/grep --ignore-case --invert-match apple | /usr/bin/sort
```
vitorgalvao: In a script we should never use the short version of flags when there’re a long ones. Short flags make it harder to understand the command at a glance.

vitorgalvao:
This is a nice idea. We could instead tell it to run [`list_recent_pkg_ids`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/list_recent_pkg_ids), since we already have the script.

And while we’re at it we could run the other `list_` commands in [`/developer/bin`](https://github.com/caskroom/homebrew-cask/tree/master/developer/bin). It’d make our job easier on new submissions.

commitay:
> We could instead tell it to run [`list_recent_pkg_ids`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/list_recent_pkg_ids), since we already have the script.

Forgot about this script, I'll update the PR to use it.

> And while we’re at it we could run the other `list_` commands in [`/developer/bin`](https://github.com/caskroom/homebrew-cask/tree/master/developer/bin).

Which scripts do you think would be useful to run?

vitorgalvao:
> Which scripts do you think would be useful to run?

Ideally they’d be run only on new cask submissions and not cask updates, but:

+ [`list_recent_pkg_ids`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/list_recent_pkg_ids)
+ [`list_installed_launchjob_ids`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/list_installed_launchjob_ids) **or** [`list_loaded_launchjob_ids`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/list_loaded_launchjob_ids)
+ [`list_loaded_kext_ids`](https://github.com/caskroom/homebrew-cask/blob/master/developer/bin/list_loaded_kext_ids)

commitay:
> list_recent_pkg_ids

This is useful for updates, the top post has the example that gave me the idea when `pkgutil` didn't remove anything after the pkg id changed. https://github.com/caskroom/homebrew-cask/pull/41013/commits/550233207cd58451f19a897d82309c815c24e7f1

> list_installed_launchjob_ids or list_loaded_launchjob_ids

These would be helpful but sometimes the app needs to be launched first, not sure how easy that would be to do or if we even want to.

> list_loaded_kext_ids

Casks with kexts will break during `install` on travis when we switch to a 10.13 image.

vitorgalvao:
> > list_installed_launchjob_ids or list_loaded_launchjob_ids
>
> These would be helpful but sometimes the app needs to be launched first, not sure how easy that would be to do or if we even want to.

True, but catching some is better than catching none. I’m thinking of it more like as a reference of “oh, I missed this one before”.

commitay:
```
fatal: Invalid symmetric difference expression c4b0f95e3a97ca3416fe09a5b71091d62957e93c...cb0ae1da713a36a79619e80ea725febcbca065ba
```

commitay:
An alternative approach to this could be that we just run the scripts before and after the main block, save the output and then run `diff`. It would make the final output cleaner and we don't need to add conditionals.

reitermarkus:
> Casks with kexts will break during install on travis when we switch to a 10.13 image.

Since this security feature won't go away, I think it would make sense to add some information about kexts in the Casks, to at least give a better error message. It took me about two weeks to realize VirtualBox wasn't broken but that I had to click a button in `Preferences -> Security`.

commitay:
>  I think it would make sense to add some information about kexts in the Casks, to at least give a better error message. 

A `caveat` that only displays on >=10.13 would work.

ci/travis/script.sh
```diff
@@ -25,6 +25,12 @@ elif [[ ${#modified_casks[@]} -gt 0 ]]; then
     for cask in "${modified_casks[@]}"; do
       run brew cask reinstall --verbose "${cask}"
       run brew cask uninstall --verbose "${cask}"
+      if brew cask _stanza uninstall "${cask}" &> /dev/null; then
+        run "$(brew --repository)/Library/Taps/caskroom/homebrew-cask/developer/bin/list_loaded_launchjob_ids"
```
vitorgalvao: I was thinking of this inside the `pkg` conditional as well. Point being that we find these when the contributor hasn’t included it.

commitay: I've updated https://github.com/caskroom/homebrew-cask/pull/41111#issuecomment-346278461 with a (rough) alternative and a couple of extra checks that are useful, especially for pkgs.

reitermarkus:
> It seems to have been badly implemented by Apple, the Virtualbox forums and a few of the users posts here say that sometimes they have to approve kexts multiple times. I removed one post here that recommended disabling SIP altogether (with detailed instructions).

Yeah, there was one instance (in my VM I think) where clicking “Allow” was essentially bricked and just didn't do anything at all.

commitay:
Created an issue for the kext discussion. https://github.com/caskroom/homebrew-cask/issues/41244

vitorgalvao:
Regarding https://github.com/caskroom/homebrew-cask/pull/41111#issuecomment-346278461, do those work as `alias` instead of functions? If not, could they at least be a single line per function? To do so you’ll likely need to end it with `;` (`something() { command ; }`).

commitay:
They should work as `alias`.  I only switched to functions so the post was clearer, the first version looked weird and was kinda hard to read.

I'll clean up the script and push it it tomorrow.

commitay:
Closing, will open a new PR.

PR #41262, opened 2017-11-23
Boxcryptor 2.17.892

ondrejfuhrer:

### DIFF
--- a/Casks/boxcryptor.rb
+++ b/Casks/boxcryptor.rb
@@ -8,7 +8,7 @@ cask 'boxcryptor' do
   name 'Boxcryptor'
   homepage 'https://www.boxcryptor.com/en/'
 
-  depends_on macos: '>= :mavericks'
+  depends_on macos: '>= :yosemite'
 
   app 'Boxcryptor.app'
 end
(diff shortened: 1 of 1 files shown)


ondrejfuhrer:
They changed the Appcast for some reason, I would suggest to wait, what happened (they removed the 2.17.892 version from appcast)

vitorgalvao:
Let’s give it until Monday.

ondrejfuhrer:
I just checked and it is changed again (with the latest version there), and the appcast checkpoint from my commit is valid again. Can you remove your commit?

commitay:
> they removed the 2.17.892 version from appcast

And it's been removed again.

https://rink.hockeyapp.net/api/2/apps/7fd6db3e51a977132e3b120c613eaea8

commitay:
The homepage and the changelog both show `2.17.892`.

The appcast versioning might be an issue with HockeyApp itself.

ondrejfuhrer:
I asked them on twitter and the reply was:
```
Hi! 
This behavior occurs because we do staged rollouts, which means that not all users get the new updates at exactly the same time. 
Best,
Rebecca from Boxcryptor
```

Which does not make much sense, but at least we know, they consider that "normal"

commitay:
Appcast has `2.17.892` again.

commitay (approved):
Verified locally.

PR #41374, opened 2017-11-27
Remove fiddler

commitay:
> https://news.ycombinator.com/item?id=15791781

> `https://github.com/chocolatey/chocolatey-coreteampackages/issues/923`

> Telerik reached out and is asking that the Fiddler package be removed as it is giving direct access to a download location they no longer show on the site.


commitay:
As upstream is asking downstreams to remove packages, we may as well just remove it now and save any hassle.

PR #41449, opened 2017-11-29
Update libreoffice.rb

miccal:
Pending https://github.com/caskroom/homebrew-versions/pull/4882


miccal:
Never mind, the binaries will still conflict.

PR #41520, opened 2017-12-01
Switch spotify from latest to 1.0.68.407.g6864aaaf-37

sparhomenko:
After making all changes to the 
---
This switches Spotify cask from `:latest` and `:no_check` to the current version. The versioned download is the same as the one used by the auto-updater, it seems to have the same content as the offline installer but without `Spotify.app` root dir. For further updates the version can typically be found in `CFBundleVersion` property of `/Applications/Spotify.app/Contents/Info.plist` after auto-updating. In rare cases of a mismatch - eg. at the moment `CFBundleVersion` is `1.0.68.407.g6864aaaf-36` but the download is `1.0.68.407.g6864aaaf-37` - it can be found by inspecting the app calls with a proxy tool.

I've tried (hard) to find an `appcast` without any success: app seems to get it via an encrypted TCP connection to one of the ["ap" servers](http://apresolve.spotify.com/), and they [decided](https://community.spotify.com/t5/Live-Ideas/All-Platforms-Changelog-after-each-Update-In-App/idc-p/1412187#M85424) to stop providing release notes. So although automating version upgrades for this cask at the moment doesn't look easy, I've decided to give this a try since this is one of the [few remaining](https://github.com/Homebrew/brew/pull/3396#issuecomment-347625451) popular casks with both `:latest` and `auto_updates` that won't work well with the shiny new `brew cask upgrade`.

Please feel free to weigh in, I would love to get maintainers' input before looking into other `:latest` casks I use.

### DIFF
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -1,17 +1,18 @@
 cask 'spotify' do
-  version :latest
-  sha256 :no_check
+  version '1.0.68.407.g6864aaaf-37'
+  sha256 '56081655d0b559d69fc2b3ed894c0ef77d82c53b6c9ae0336c6dd94131885264'
 
-  url 'https://download.spotify.com/Spotify.dmg'
+  url "https://upgrade.spotify.com/upgrade/client/osx-x86_64/spotify-autoupdate-#{version}.tbz"
   name 'Spotify'
   homepage 'https://www.spotify.com/'
 
   auto_updates true
   depends_on macos: '>= :lion'
 
-  app 'Spotify.app'
+  app '.', target: 'Spotify.app'
 
-  uninstall launchctl:  'com.spotify.webhelper',
+  uninstall quit:       'com.spotify.client',
+            launchctl:  'com.spotify.webhelper',
             login_item: 'Spotify'
 
   zap trash: [


commitay:
How would we even know what the new version is when it changes?

sparhomenko:
> How would we even know what the new version is when it changes?

I've tried to document it here:

>For further updates the version can typically be found in `CFBundleVersion` property of `/Applications/Spotify.app/Contents/Info.plist` after auto-updating. In rare cases of a mismatch - eg. at the moment CFBundleVersion is `1.0.68.407.g6864aaaf-36` but the download is `1.0.68.407.g6864aaaf-37` - it can be found by inspecting the app calls with a proxy tool.

You can also see the version in the About window but unfortunately it's not enough to update the cask - it looks like `1.0.68.407.g6864aaaf` and doesn't have the last `-37` component, thus the more complicated alternatives.

commitay:
Without an `appcast` that publishes the version we can't merge this. 

Also things like this: `app '.', target: 'Spotify.app'` we try to avoid unless absolutely necessary.

Parsing `plist`s and other similar approaches for versioning Casks have been attempted previously but they are all to difficult to reliably (and easily) implement.

I appreciate that you've put work into trying to version this but it isn't practical.
____

RE: the list in https://github.com/Homebrew/brew/pull/3396#issuecomment-347625451

Several of those Casks are deliberately unversioned even if versioned downloads or appcasts exist.

A couple of them should actually have `auto_updates true` removed.

sparhomenko:
I see, thanks for the input. It might be good to update [Adding a Cask](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md) doc - at the moment it recommends using `:latest` only if only an unversioned download is available.

commitay:
> at the moment it recommends using :latest only if only an unversioned download is available.

This is generally correct, we prefer versioned downloads *unless* they are impractical / regularly break, etc.

This is also overridden by other docs depending on the Cask, i.e. for nightlys:

> https://github.com/caskroom/homebrew-versions/blob/master/CONTRIBUTING.md#nightly-builds-nightlies 

> For Casks of nightlies, version, sha256, and url change frequently, making the Cask constantly outdated. To alleviate having to regularly update them, the use of a url do block in conjunction with version :latest and sha256 :no_check is encouraged, to ensure the latest version is always fetched.

sparhomenko:
I guess the desire to make it versioned is mostly triggered by the decision in Homebrew/brew#3396 to include `:latest` into `brew cask upgrade --greedy`, i.e. not supporting `brew cu -a` behaviour. Normally it's minor annoyance - upgrade redownloads some casks even when it's not needed - but here it actually leads to a downgrade since `:latest` currently means `1.0.67.582.g19436fa3-17`. So my scheduled `brew cask upgrade --greedy` ends up in a daily fight with the auto-updater.

>This is generally correct, we prefer versioned downloads unless they are impractical / regularly break, etc.

My comment on the doc was more specific - if not having an `appcast` that publishes the version makes it impractical to use versioned cask when `:latest` is available it might be worthwhile to mention in that doc to prevent PRs like mine.

PR #41699, opened 2017-12-06
rstudio 1.1.383: Recommend brew cask install r-app

sjackman:

-----

Rstudio depends on R. There are two ways of satisfying that dependency.
Option 1 is recommended to be able to use precompiled binary packages.
Option 2 is not recommended, as it requires compiling R packages from source.

1. `brew cask install r-app`
2. `brew install r`

### DIFF
--- a/Casks/rstudio.rb
+++ b/Casks/rstudio.rb
@@ -13,10 +13,18 @@ cask 'rstudio' do
 
   caveats <<~EOS
     #{token} depends on R.
-    There are different ways to satisfy that dependency and we don’t want to impose one, so it is up to you to satisfy it.
-    We suggest you do so by running one of:
+    There are different ways to satisfy that dependency. RStudio recommends installing R from The R Project, which is required to install binary R packages, without needing to compile packages from source.
+
+    https://support.rstudio.com/hc/en-us/articles/217799238
+
+    To install the R Project package run:
+
+      brew cask install r-app
+
+    Alternative ways to satisfy the dependency are:
 
       brew install r
-      brew cask install r-app
+
+    This requires compiling R packages from source.
   EOS
 end


commitay:
> Option 1 is recommended

By upstream? If they recommend it we should link to their documentation.

commitay:
Just out of curiosity (I'm not familiar with anything R related) would [`microsoft-r-open`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/microsoft-r-open.rb) also fulfil the requirement?

sjackman:
> Just out of curiosity (I'm not familiar with anything R related) would microsoft-r-open also fulfil the requirement?

Zowee. I didn't know that existed. Maybe? Hopefully? I haven't tested it.

commitay:
> Maybe? Hopefully? I haven't tested it.

I probably wouldn't include it in the caveat unless upstream explicitly supports it.

___
```
  caveats <<~EOS
    #{token} depends on R.
    There are different ways to satisfy that dependency, upstream recommends the binary package.

      {upstream link}

    To install the binary package run:

      brew cask install r-app
    
    Alternative ways to satisfy the dependency are:

      brew install r  (This requires compiling R packages from source)
  EOS
```

sjackman:
>> Option 1 is recommended

> By upstream? If they recommend it we should link to their documentation.

By me. =) https://twitter.com/sjackman/status/910916874191257601
By this person: https://github.com/r-hub/homebrew-cran#prerequisites
I assume that the upstream project recommends using the binaries that they provide.
https://cran.r-project.org/bin/macosx/

commitay:
By upstream I mean `rstudio`

commitay:
> https://support.rstudio.com/hc/en-us/articles/200486548-Frequently-Asked-Questions

> What versions of R is RStudio compatible with?
In order to run RStudio you need to have already installed R 2.11.1 or higher. You can download the most recent version of R for your environment from CRAN.

commitay:
I can't find anything that explicitly says the CRAN binary is preferred over Homebrew.

sjackman:
Also here:

https://support.rstudio.com/hc/en-us/articles/217799238-When-I-start-RStudio-Desktop-I-get-the-message-unable-to-locate-R-binary-by-scanning-standard-location

> https://support.rstudio.com/hc/en-us/articles/217799238-When-I-start-RStudio-Desktop-I-get-the-message-unable-to-locate-R-binary-by-scanning-standard-location

sjackman:
The Homebrew package of R is not able to install binary R packages. It took me years to figure that out.

sjackman:
I've revised my PR based on your suggestion above.

commitay:
Homebrew only seems to be referenced _once_ in RStudio docs, they definitely prefer the binary!
___

> I've revised my PR based on your suggestion above.

Thanks, sorry for the run around over a few lines of text!

sjackman:
No worries.

sjackman:
Thanks for merging, @commitay!

PR #41733, opened 2017-12-08
Update qt-creator version to 4.5.0

sp-1234:
√    `brew cask audit --download {{cask_file}}` is error-free.
√    `brew cask style --fix {{cask_file}}` left no offenses.
√    The commit message includes the cask’s name and version.

Update qt-creator version to 4.5.0 and use https for downloading

Version and hash taken from https://download.qt.io/official_releases/qtcreator/4.5/4.5.0/qt-creator-opensource-mac-x86_64-4.5.0.dmg.mirrorlist
HTTPS link checked and it works

P.S.: I don't know why but Github didn't insert the template message here and I don't know how to do these square checkbox thing, I hope it's readable enough like it is.

### DIFF
--- a/Casks/qt-creator.rb
+++ b/Casks/qt-creator.rb
@@ -2,7 +2,7 @@ cask 'qt-creator' do
   version '4.5.0'
   sha256 'cb3aac65ddadaccef1bf62a95120f8d74d3e4454f7afcdcb5425f20051d24b5b'
 
-  url "https://download.qt.io/official_releases/qtcreator/#{version.major_minor}/#{version}/qt-creator-opensource-mac-x86_64-#{version}.dmg"
+  url "http://download.qt.io/official_releases/qtcreator/#{version.major_minor}/#{version}/qt-creator-opensource-mac-x86_64-#{version}.dmg"
   name 'Qt Creator'
   homepage 'https://www1.qt.io/developers/'
 
--- a/Casks/qt-creator.rb
+++ b/Casks/qt-creator.rb
@@ -1,8 +1,8 @@
 cask 'qt-creator' do
-  version '4.4.1'
-  sha256 '6e048a113f34bd3333fbcecdbf916aa292c5fe510662c4ad7cf8d65a515a0d64'
+  version '4.5.0'
+  sha256 'cb3aac65ddadaccef1bf62a95120f8d74d3e4454f7afcdcb5425f20051d24b5b'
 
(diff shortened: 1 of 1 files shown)


commitay:
> HTTPS link checked and it works

It downgrades to HTTP

sp-1234:
Well it may stop doing that in the future. And from my address it redirected to another `https` url, so maybe it depends.

sp-1234:
kind of funny that the tool doesn't agree with that
>  - The URL http://download.qt.io/official_releases/qtcreator/4.5/4.5.0/qt-creator-opensource-mac-x86_64-4.5.0.dmg should use HTTPS rather than HTTP

from Travis log at https://travis-ci.org/caskroom/homebrew-cask/builds/313411266

commitay:
https://github.com/caskroom/homebrew-cask/issues/41735

vitorgalvao:
> Well it may stop doing that in the future.

We don’t typically make changes to casks based on conjecture.

It’s important to us to not have HTTPS URLs that redirect to HTTP because it gives users a false sense of security.

sp-1234:
@victorpopkov as long as redirects are allowed, **all** HTTPS URLs are false sense of security. Because the host may start behaving differently at any moment, especially if it uses redirecting to serve from mirrors and they can be configured very differently.

Until there's no option to automatically and surely forbid redirects to plaintext connections or forbid  installation of formulas without sha256, I don't see how it makes any significant difference. And last time I checked these 2 things were true.

vitorgalvao:
> as long as https->http redirects are allowed, **all** HTTPS URLs are false sense of security.

Which is why they aren’t. Those casks now should fail audit, which is the whole point.

> Until there's no option to (…) forbid installation of formulas without sha256

There is. It’s [`--require-sha`](https://github.com/caskroom/homebrew-cask/blob/master/USAGE.md#options).

This is not a discussion. All the points you mentioned have already been extensively discussed and decided upon in the past. They’re either implemented or on the list. We’re not arguing with you about what the best method is — we’re telling you what we already concluded.

If you want to add to those decisions, you’ll need to research the old issues and the documentation and make your arguments based on that. Respectfully, we don’t want to start to explain everything from scratch.

---

Two extra small notes:

+ You @mentioned the wrong person. It’s a common mistake, don’t worry about it. But I have missed points made to me in the past due to it.
+ We call them “casks”, not “formulas”. They are different things and behave differently. They do not share features.

commitay (approved):
Verified locally. Disregarding the new `audit` false positive.

PR #41851, opened 2017-12-12
Add BonitaStudioCommunity.app v7.6.0

claui:

    **Note:** For simplicity’s sake, I left out the `community` suffix as this is currently the only edition available online.

___

Cross-platform BPM modeling app; its Community Edition is under GPLv2.

### DIFF
--- a/Casks/bonitastudio.rb
+++ b/Casks/bonitastudio.rb
@@ -1,0 +1,24 @@
+cask 'bonitastudio' do
+  version '7.6.0'
+  sha256 'bb5a12a233ca7f0ede813070fedf34ecd3343c16dfae0f255f4707da130dba32'
+
+  # download.forge.ow2.org was verified as official when first introduced to the cask
+  url "https://download.forge.ow2.org/bonita/BonitaStudioCommunity-#{version}-x86_64.dmg"
+  appcast 'https://documentation.bonitasoft.com///?page=release-notes',
+          checkpoint: 'a66974da63b8e9383a454678b31a65a9008ce0ca0525949ce26cc9a1e884d075'
+  name 'Bonita Studio Community Edition'
+  homepage 'https://www.bonitasoft.com/downloads'
+
+  installer script: {
+                      executable: "#{staged_path}/BonitaStudioCommunity-#{version}-osx.app/Contents/MacOS/installbuilder.sh",
+                      args:       ['--mode', 'unattended'],
+                    }
+
+  uninstall quit:   'org.bonitasoft.studio.product',
+            delete: "#{appdir}/BonitaStudioCommunity.app"
+
+  zap trash: [
+               '~/Library/Preferences/org.bonitasoft.studio.product.plist',
+               '/Library/Caches/org.bonitasoft.studio.product',
+             ]
+end
--- a/Casks/bonitastudiocommunity.rb
+++ b/Casks/bonitastudiocommunity.rb
@@ -1,4 +1,4 @@
-cask 'bonitastudio' do
+cask 'bonitastudiocommunity' do
   version '7.6.0'
   sha256 'bb5a12a233ca7f0ede813070fedf34ecd3343c16dfae0f255f4707da130dba32'
 
@@ -15,7 +15,7 @@ cask 'bonitastudio' do
                     }
 
   uninstall quit:   'org.bonitasoft.studio.product',
-            delete: "#{appdir}/BonitaStudioCommunity.app"
+            delete: '/Applications/BonitaStudioCommunity.app'
 
   zap trash: [
                '~/Library/Preferences/org.bonitasoft.studio.product.plist',


Casks/bonita-studio.rb
```diff
@@ -0,0 +1,28 @@
+cask 'bonita-studio' do
```
commitay: No spaces in the `app` name so probably `'bonitastudio'`

Casks/bonita-studio.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  # download.forge.ow2.org was verified as official when first introduced to the cask
+  url "https://download.forge.ow2.org/bonita/BonitaStudioCommunity-#{version}-x86_64.dmg"
+  appcast 'https://documentation.bonitasoft.com///?page=release-notes',
+          checkpoint: 'a66974da63b8e9383a454678b31a65a9008ce0ca0525949ce26cc9a1e884d075'
+  name 'Bonita Studio Community Edition'
+  homepage 'https://www.bonitasoft.com/downloads'
+
+  installer script: {
+                      executable: "#{staged_path}/BonitaStudioCommunity-7.6.0-osx.app/Contents/MacOS/installbuilder.sh",
```
commitay: `{version}`

Casks/bonita-studio.rb
claui: Good catch!

Casks/bonitastudio.rb
```diff
@@ -0,0 +1,28 @@
...
+  uninstall quit:   'org.bonitasoft.studio.product',
+            delete: "#{appdir}/BonitaStudioCommunity.app"
+
+  zap trash: [
+               '~/Library/Preferences/org.bonitasoft.studio.product.plist',
+               '/Library/Caches/org.bonitasoft.studio.product',
+             ]
+
+  caveats do
+    depends_on_java('6+')
```
commitay: Upstream mentions Java 8 for the `zip` version?

https://www.bonitasoft.com/downloads?version=7.6.0&os=mac

claui: You’re right in that the `zip` version (which is not the version the Cask uses) does state the 1.8 dependency.

Lacking consistent hints with respect to the _installer_ version – which claims to be a 32-bit app while it’s actually a x86_64 binary – I had taken for granted what the app’s `Info.plist` says after I’d install the Cask:

```
	<key>Eclipse</key>
		<array>
			<!-- to use a specific Java version (instead of the platform's default) uncomment one of the following options,
					or add a VM found via $/usr/libexec/java_home -V
				<string>-vm</string><string>/System/Library/Java/JavaVirtualMachines/1.6.0.jdk/Contents/Commands/java</string>
				<string>-vm</string><string>/Library/Java/JavaVirtualMachines/1.8.0.jdk/Contents/Home/bin/java</string>
			-->
			[…]
		</array>
```

Following your comment, I’ve just tried to figure out things using the instructions from the plist. Turns out the app silently uses 1.8 all the time, no matter which `jenv` version I use to launch the app, and also regardless of what `JAVA_HOME` I put in the plist.

Turns out there’s **another config item** hard-coded into `/Contents/Eclipse/BonitaStudioCommunity.ini`, which enforces a minimum Java version of 1.8:

```
-Dosgi.requiredJavaVersion=1.8
```

And indeed, this seems to override any other setting. Once more, good catch @commitay! ☺️

commitay:
> Turns out the app silently uses 1.8 all the time

Does it bundle a JRE? The Java `caveat` may not be required.

claui:
> Does it bundle a JRE? The Java `caveat` may not be required.

Correct. The bundle indeed ships with a JRE in `/Contents/Eclipse/jre/Contents/Home`.  
A quick `pgrep -afl java` shows that the app actually runs on that. The app also uses the bundled JVM (by default) for the child VMs it spawns.

claui:
```>
 ==> Downloading https://download.forge.ow2.org/bonita/BonitaStudioCommunity-7.6.0-x86_64.dmg


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

The Travis build seems to consistently time out mid-download.

The upstream package weighs roughly 430 MB; according to my calculations, a Travis timeout sounds plausible only if the box, on average, used no more than 70 kbit/s of dedicated downstream bandwidth for that particular build job. However, I don’t know how much downstream bandwidth the Travis box actually has, and among how many concurrent jobs it is split.

@caskroom/maintainers Is it worth the effort to increase the Travis timeout, then check if the problem goes away?

vitorgalvao:
> For simplicity’s sake, I left out the `community` suffix as this is currently the only edition available online.

“Currently” can be a big caveat. If we leave out `community` for now and then a non-`community` app comes out, when running `brew cask upgrade` users will be switched to an essentially different cask (and open repeated bug reports). I’d rather have less simplicity but more clarity.

> Is it worth the effort to increase the Travis timeout, then check if the problem goes away?

Huge casks are a rare occurrence. Travis is a make-sure-everything-is-alright, but we can (and sometimes do) ignore it in favour of manual verification. Since you are a maintainer, your own local verification is enough.

commitay:
> Is it worth the effort to increase the Travis timeout

I've tried a couple of workarounds for this and couldn't find a good solution. `travis_wait` increases the timeout but hides the output and adding `brew cask fetch --verbose` if =! `latest` cuts off the CURL output and still times out.

Casks/bonitastudio.rb
```diff
@@ -0,0 +1,24 @@
...
+  name 'Bonita Studio Community Edition'
+  homepage 'https://www.bonitasoft.com/downloads'
+
+  installer script: {
+                      executable: "#{staged_path}/BonitaStudioCommunity-#{version}-osx.app/Contents/MacOS/installbuilder.sh",
+                      args:       ['--mode', 'unattended'],
+                    }
+
+  uninstall quit:   'org.bonitasoft.studio.product',
+            delete: "#{appdir}/BonitaStudioCommunity.app"
```
commitay: This should probably be `/Applications/` as the install doesn't have an option for install location.

claui:
@vitorgalvao Thanks for pointing this out. I’m fine with tacking on the `-community` suffix.

Casks/bonitastudiocommunity.rb
```diff
@@ -0,0 +1,24 @@
...
+          checkpoint: 'a66974da63b8e9383a454678b31a65a9008ce0ca0525949ce26cc9a1e884d075'
+  name 'Bonita Studio Community Edition'
+  homepage 'https://www.bonitasoft.com/downloads'
+
+  installer script: {
+                      executable: "#{staged_path}/BonitaStudioCommunity-#{version}-osx.app/Contents/MacOS/installbuilder.sh",
+                      args:       ['--mode', 'unattended'],
+                    }
+
+  uninstall quit:   'org.bonitasoft.studio.product',
```
vitorgalvao: Isn’t there something missing? Does the `installbuilder.sh` serve only to build an app?

claui: The script seems to build a command line to launch the actual installer, depending on the architecture, and then executes that command line. I believe that this is why the script bears that confusing name.

claui:
@vitorgalvao Any idea why Travis suddenly says _Invalid symmetric difference expression_ (instead of timing out like it used to)?

```
fatal: Invalid symmetric difference expression a37d9810621879277f4404e63778409f8bcff9cd...5c307818054d03e32c3798b95e420a224218c151
```

commitay:
Travis doesn't like "old" branches. https://github.com/caskroom/homebrew-cask/issues/40648

commitay:
I've added the Travis commit range workaround we used to have to see if that resolves the issue.

vitorgalvao (approved):

PR #41890, opened 2017-12-13
Remove SIP Casks

commitay:

Requires SIP to be disabled for > yosemite. https://totalfinder.binaryage.com/sip

Similar case: https://github.com/caskroom/homebrew-cask/issues/38746


commitay:
```
Error:
 Cask totalfinder depends on macOS release being one of [10.9, 10.10], but you are running release 10.12.

Error:
 Kernel.exit
```

vitorgalvao:
According to the rule to support only the latest two versions, at a certain point it stops making sense we even have these casks.

How about we start removing them and add a rule about not supporting casks that need a disabled SIP to install?

commitay:
> How about we start removing them and add a rule about not supporting casks that need a disabled SIP to install?

Yes, makes sense. I'll update this to remove them and add SIP to `rejected_casks`.

vitorgalvao (approved):

sscotth:
Tap for posterity: https://github.com/sscotth/homebrew-sip

wickles:
@sscotth I take it you do not plan to maintain that tap?

sscotth:
@wucjkes Thanks for the ping. I can add the #41971 casks if that's what you are looking for. I didn't know they existed.

PR #41916, opened 2017-12-13
Remove totalspaces

commitay:
Refs: https://github.com/caskroom/homebrew-cask/pull/41890


nass600:
How can we install total spaces with brew after this change?

commitay:
https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md

nass600:
Thanks, I assume there is no tap already created for this then

PR #41947, opened 2017-12-14
Update dyn-updater to 5.4.2

zach-adams:


commitay:
As the `url` is unversioned and there isn't an `appcast`, this should stay as `latest` and `no_check`.

PR #42141, opened 2017-12-20
Add free TAL audio plugins

mofeing:
- [x] ``brew cask audit --download {{cask_file}}`` is error-free.
- [x] ``brew cask style --fix {{cask_file}}`` left no offenses.

The added casks are:
- ``tal-reverb-2``
- ``tal-reverb-3``
- ``tal-reverb-4``
- ``tal-vocoder``
- ``tal-noisemaker``

### DIFF
--- a/Casks/tal-noisemaker.rb
+++ b/Casks/tal-noisemaker.rb
@@ -1,0 +1,16 @@
+cask 'tal-noisemaker' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://tal-software.com//downloads/plugins/TAL-NoiseMaker-installer.pkg'
+  name 'TAL-NoiseMaker'
+  homepage 'https://tal-software.com/products/tal-noisemaker'
+
+  pkg 'TAL-NoiseMaker-installer.pkg'
+
+  uninstall pkgutil: [
+                       'com.talsoftware.pkg.TAL-NoiseMaker-AAX',
+                       'com.talsoftware.pkg.TAL-NoiseMaker-AU',
+                       'com.talsoftware.pkg.TAL-NoiseMaker-VST',
+                     ]
+end
--- a/Casks/tal-reverb-2.rb
+++ b/Casks/tal-reverb-2.rb
@@ -1,0 +1,16 @@
+cask 'tal-reverb-2' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://tal-software.com/downloads/plugins/TAL-Reverb-2-installer.pkg'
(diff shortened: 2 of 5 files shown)


vitorgalvao:
* In the future, please submit one PR per cask.
* Numbers in the name and token should not be separate from letters.
* Casks that have a number [need a comment](https://github.com/caskroom/homebrew-cask/blob/81f45bafe5011f93ae1e5693120f65fdfbbf84d6/Casks/4k-youtube-to-mp3.rb#L2).
* I’m 👎 on adding these: unversioned obscure pkgs that share a homepage? Four characteristics that make them harder to maintain and unlikely to ever be updated. Casks like these usually stay a while, give us a ton of headaches, eventually break and we remove them. I’d rather save the trouble.

Keeping open for a while for other maintainers to disagree.

mofeing:
@vitorgalvao 

> - In the future, please submit one PR per cask.

Okie dokie.


> - Numbers in the name and token should not be separate from letters.
> - Casks that have a number need a comment.

They are actually different software (not upgraded versions) and those are their names but just lowercase (TAL-Reverb-2, TAL-Reverb-3, TAL-Reverb-4, TAL-Vocoder and TAL-NoiseMaker). I should have pointed that.


> - I’m 👎 on adding these: unversioned obscure pkgs that share a homepage? Four characteristics that make them harder to maintain and unlikely to ever be updated. Casks like these usually stay a while, give us a ton of headaches, eventually break and we remove them. I’d rather save the trouble.

They do have version. Since there is already a plugin from the same developer inside this repository (``tal-chorus-lx``) I limited to copy the cask and adapt it to each case.

vitorgalvao:
> They are actually different software (not upgraded versions) and those are their names

Yes, I understand, that’s why I mentioned [the comment stating they are not versions](https://github.com/caskroom/homebrew-cask/blob/81f45bafe5011f93ae1e5693120f65fdfbbf84d6/Casks/4k-youtube-to-mp3.rb#L2). What I mean is that instead of `tal-reverb-2` it should be `tal-reverb2` (same for the others).

> They do have version.

By “unversioned”, I mean it in the sense we use it here, as in “the URL is unversioned, as it always points to the latest version”.

commitay:
The uninstalls for the reverb Casks conflict.

commitay:
Closing as per https://github.com/caskroom/homebrew-cask/pull/42141#issuecomment-353110738, feel free to submit separate PRs for `tal-vocoder` and `tal-noisemaker`.

PR #42188, opened 2017-12-21
Update zoomus to version 4.1.17379.1218

mikemimik:
Just updated the version to the newest published on their website

Ref: [Zoom download page](https://zoom.us/download)

### DIFF
--- a/Casks/zoomus.rb
+++ b/Casks/zoomus.rb
@@ -1,6 +1,6 @@
 cask 'zoomus' do
   version '4.1.17379.1218'
-  sha256 '111640bbf99308aa57b60b49ad4f62a9a20a3c44b71870a7856e901e41690b72'
+  sha256 '68d4a9b58ee523f66fcbe3e8d6ea6443adf32819a33861e89d60b701ef16580d'
 
   url "https://www.zoom.us/client/#{version}/zoomusInstaller.pkg"
   name 'Zoom.us'
--- a/Casks/zoomus.rb
+++ b/Casks/zoomus.rb
@@ -1,5 +1,5 @@
 cask 'zoomus' do
-  version '4.1.11049.1024'
+  version '4.1.17379.1218'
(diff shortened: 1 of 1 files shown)


Casks/zoomus.rb
```diff
@@ -1,5 +1,5 @@
 cask 'zoomus' do
-  version '4.1.11049.1024'
+  version '4.1.17379.1218'
   sha256 '111640bbf99308aa57b60b49ad4f62a9a20a3c44b71870a7856e901e41690b72'
```
commitay:
```
audit for zoomus: failed
 - download not possible: Checksum for Cask 'zoomus' does not match.
```

Please update the `sha256`

commitay (dismissed):

Casks/zoomus.rb
mikemimik: @commitay Thanks I didn't think of this. How do I get the new checksum?

commitay: Run `shasum -a 256 {file}`

mikemimik: @commitay got it! Thanks! c816216 should fix it

vitorgalvao: @mikemimik Didn’t you follow the template before checking the boxes? The commands would have caught this.

vitorgalvao (approved):

Casks/zoomus.rb
vitorgalvao: @mikemimik I see you didn’t, because this was (likely) submitted via the Github web interface. When you do so, please do not just click the checkboxes. It’s fine if you can’t do what’s in them and leave them empty, but checking them without actually following the instructions is not — it means the information you’re giving us is not true and it makes our work harder.

We’d rather have truthful unchecked boxes than falsely checked ones. With the former we know what we’re dealing with; with the latter we don’t.

PR #42443, opened 2018-01-02
Update lyx to 2.2.3

ldeck:


vitorgalvao:
Thank you, but we prefer the URL that is on the homepage, as well as keeping the version interpolation.

ldeck:
Yes, it would be preferred except that it was returning connection refused. Stripping the Version Interpolation was an oversight, sorry.

PR #42477, opened 2018-01-02
Updated Spotify Download URL from actual installer

childnode:
old URL actually resolves to 1.0.69.336.g7edcc575, new URL to 1.0.70.388.g8e1ed5af

  - no cask version as long it is a permanent URL which is updated anytime

### DIFF
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -2,7 +2,7 @@ cask 'spotify' do
   version :latest
   sha256 :no_check
 
-  url 'https://download.spotify.com/Spotify.dmg'
+  url 'https://download.scdn.co/Spotify.dmg'
   name 'Spotify'
   homepage 'https://www.spotify.com/'
 
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -2,6 +2,7 @@ cask 'spotify' do
   version :latest
   sha256 :no_check
 
+  # scdn.co was verified as official when first introduced to the cask
   url 'https://download.scdn.co/Spotify.dmg'
   name 'Spotify'
   homepage 'https://www.spotify.com/'


Casks/spotify.rb
```diff
@@ -2,7 +2,8 @@
   version :latest
   sha256 :no_check
 
-  url 'https://download.spotify.com/Spotify.dmg'
+  # scdn.co was verified as official when first introduced to the cask
+  url 'https://download.scdn.co/Spotify.dmg'
```
commitay: We need to be able to publicly verify this link _without_ inspecting the installer.

adidalal: Spotify auto updates, using a `http://upgrade.scdn.co/upgrade/client/osx-x86_64/__` url.

We should keep the URL as is linked on the homepage.

commitay:
Closing as the download URL cannot be verified.

childnode:
just fmi: the "correct" way would be a cask with the installer instead of the nowadays hidden app container image?

Casks/spotify.rb
childnode: Hi @adidalal I don't know what URL is prompted to you, but for me, https://www.spotify.com/de/download/mac/ only tells me the https://download.scdn.co/SpotifyInstaller.zip

adidalal:
Ah, I see what you mean.

@commitay, This is valid - Install Spotify.app is a stub that downloads from that URL.

```
	<key>SpotifyDownloadURL</key>
	<string>https://download.scdn.co/Spotify.dmg</string>
```

commitay:
If the url doesn't match the homepage, we need to be able to verify it.

I'm not sure I understand how we can do that without downloading and inspecting the installer.

vitorgalvao:
Seems like the `scdn.co` URL is now on the homepage.

1. Enter https://www.spotify.com/
2. Click to Get it free.
3. The “try again” link points to the domain


It does point to `zip`, though, and not a `dmg`.

adidalal:
Correct, and the `.zip` is just a stub installer that downloads the `.dmg`

PR #42600, opened 2018-01-08
Add Evergreen pre-release

ramsey:
Evergreen is currently in a pre-alpha state, described by the author as "barely usable at this point. Not even that. It’s not beta — it’s not even alpha." Therefore, I am submitting this to the main repo. The author previously created NetNewsWire and MarsEdit, and there is a lot of anticipation for this new project, so I expect it to continue to a version 1 in the near-ish future.

I originally attempted to add a `zap` section, but after removing the user-specific files and then reinstalling, the app did not return to "factory settings." Only after [following these instructions](https://github.com/brentsimmons/Evergreen/blob/master/Technotes/RestoreToFactory.md) could I get it to return to factory settings, and I was unable to run these commands using the `uninstall script:` section. I figured the best approach was to leave all the user-specific files in place for now.

---

### DIFF
--- a/Casks/evergreen.rb
+++ b/Casks/evergreen.rb
@@ -1,0 +1,15 @@
+cask 'evergreen' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://ranchero.com/downloads/Evergreen-latest.zip'
+  appcast 'https://ranchero.com/downloads/evergreen-beta.xml',
+          checkpoint: '8e32778a4f5c31bf091df662426dd12834af89fde814bcb6dd2668c57644302b'
+  name 'Evergreen'
+  homepage 'https://ranchero.com/evergreen/'
+
+  auto_updates true
+  depends_on macos: '>= :high_sierra'
+
+  app 'Evergreen.app'
+end
--- a/Casks/evergreen.rb
+++ b/Casks/evergreen.rb
@@ -1,8 +1,8 @@
 cask 'evergreen' do
-  version :latest
-  sha256 :no_check
+  version '1.0d30'
+  sha256 'f9290d0fb028980e95d3d1cb1770f1f352d376317684cef6ea63ecd7aa844dc2'
 
-  url 'https://ranchero.com/downloads/Evergreen-latest.zip'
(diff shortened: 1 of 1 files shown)


Casks/evergreen.rb
```diff
@@ -0,0 +1,15 @@
+cask 'evergreen' do
+  version :latest
```
commitay: `1.0d30`

Casks/evergreen.rb
```diff
@@ -0,0 +1,15 @@
...
+  sha256 :no_check
```
commitay: Please add the `sha256`

Casks/evergreen.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  url 'https://ranchero.com/downloads/Evergreen-latest.zip'
```
commitay: `"https://ranchero.com/downloads/Evergreen#{version}.zip"`

commitay (changes requested):

Casks/evergreen.rb
ramsey: Thanks for catching these. I should have read the contents of the appcast. :smile:

commitay:
> ...  the app did not return to "factory settings."

`zap` isn't a "reset" function, it's only for file cleanup. You can the `application support` folder back if you want but we don't run things like `defaults delete` and `killall cfprefsd` as part of `uninstall` or `zap`.

ramsey:
I understand that `zap` is not a reset function. That's not what I was trying to use it for. However, the first time you launch the app, it comes pre-configured with a list of feeds. If you modify that list, then uninstall the app with `zap` (removing the Application Support folder), it blows everything away, so the next time you install it, there are no default feeds—or any feeds at all.

I found the full list of user-specific files and preferences to move to trash with `zap`, but I could only "reset" the app to the clean install settings by using `defaults delete` and then `killall cfprefsd`.

ramsey:
It looks like the Travis CI build is failing because Travis doesn't support High Sierra yet.

commitay (approved):
Install verified locally.

victorpopkov (approved):

PR #42767, opened 2018-01-13
LICENSE: 2013

commitay:
Saves updating it every year.

[brew/LICENSE.txt](https://github.com/Homebrew/brew/blob/651988bf5538c78d4a81428ea4153baf6c4b0318/LICENSE.txt#L3)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright © 2013-2017, Paul Hinze & Contributors
+Copyright © 2013, Paul Hinze & Contributors
 All rights reserved.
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Technically, we don’t even need “present”. The copyright notice only needs to include the starting year, to be correct.

commitay:
Should I change them to `2031` only?

vitorgalvao:
> Should I change them to ~~`2031`~~ `2013` only?

Yes, that sounds better.

PR #42800, opened 2018-01-14
Update pokerstars to fix urls

kiendang:
- Website changed from www.pokerstars.com to www.pokerstars.net.
- The following regions now use the same default .net download url instead of their own: AT, DK, GR, IT, RO, UK.

### DIFF
--- a/Casks/pokerstars.rb
+++ b/Casks/pokerstars.rb
@@ -3,31 +3,7 @@ cask 'pokerstars' do
   sha256 :no_check
 
   language 'US', default: true do
-    ['.com', '']
-  end
-
-  language 'AT' do
-    ['.eu', 'EU']
-  end
-
-  language 'DK' do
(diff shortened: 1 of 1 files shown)


Casks/pokerstars.rb
```diff
@@ -3,31 +3,7 @@
...
-
-  language 'AT' do
-    ['.eu', 'EU']
-  end
-
-  language 'DK' do
-    ['.dk', 'DK']
-  end
-
-  language 'GR' do
```
vitorgalvao: This one seems to still exist.

kiendang: Yes but the actual download links all got redirected to http://www.pokerstars.net/PokerStars.net.app.zip.
Only PT is different.

kiendang: Oh yeah but the homepage's still different. I'll modify.

PR #42976, opened 2018-01-19
Update mplayer-osx-extended to rev16-test3

isolinear:


isolinear:
As a first-time contributor I used the script, but after I hit submit I realized this should probably be in homebrew-cask-versions instead since it's a "test" build.  Sorry for the trouble.

rev15, the last packaged version, crashes on startup under High Sierra.  rev16-test* versions do run, and I think should be made available for cask install.

PR #43050, opened 2018-01-21
Add zap and appcast for monero-wallet

gtklocker:

---

I'd like to mention that ideally I'd like to use the app target to rename the application to something more sensible like "Monero Wallet.app", however this would break backwards compatibility for people who have already installed this cask. Is there any way to deal with this? Do you agree that a renaming would be a good idea?

### DIFF
--- a/Casks/monero-wallet.rb
+++ b/Casks/monero-wallet.rb
@@ -9,4 +9,11 @@ cask 'monero-wallet' do
   homepage 'https://getmonero.org/'
 
   app 'monero-wallet-gui.app'
+
+  zap trash: [
+               '~/.bitmonero',
+               '~/Monero',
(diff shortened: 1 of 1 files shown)


commitay:
[target: Should Only Be Used in Select Cases](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases)

Have you asked upstream to rename it?

gtklocker:
I read the reference earlier and thought it could be considered a case of consistency. I just asked upstream: monero-project/monero-gui#1090.

I suppose this PR as is (with no naming changes) can be merged in the meantime as a naming change upstream could take very long (if it even happens). We can still keep this thread for discussion.

Cheers!

commitay (approved):
Verified locally.

PR #43164, opened 2018-01-25
Travis commit range fix (again!)

commitay:
We don't need it most of the time but it does save us from having to rebase PRs that are "out of sync" with master.

cc @reitermarkus https://github.com/caskroom/homebrew-cask/pull/38970

https://github.com/caskroom/homebrew-cask/pull/41851#issuecomment-352207180, https://github.com/caskroom/homebrew-cask/issues/40648#issue-273305278

### DIFF
--- a/ci/travis/before_script.sh
+++ b/ci/travis/before_script.sh
@@ -10,6 +10,18 @@
 
 header 'Running before_script.sh...'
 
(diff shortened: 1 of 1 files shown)


reitermarkus (approved):

reitermarkus:
Can you add a comment explaining that this is normally not needed and why it is sometimes needed? Just so we don't keep removing/adding this.

PR #43220, opened 2018-01-27
.travis.yml cache

commitay:
Update cache location

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ branches:
 
 cache:
   directories:
-    - $HOME/.gem
+    - /usr/local/Homebrew/Library/Homebrew/vendor/bundle
 
 install: true # skip install step
 


.travis.yml
```diff
@@ -15,7 +15,7 @@ branches:
 
 cache:
   directories:
-    - $HOME/.gem
+    - $(brew --repo)/Library/Homebrew/vendor/bundle
```
reitermarkus: I don't think Travis evaluates this as `bash`, also I don't think it supports cache directories outside of `$HOME`.

commitay:
> I don't think Travis evaluates this as `bash`

Good point.

> also I don't think it supports cache directories outside of `$HOME`.

There are some restrictions (symlinks, spaces, etc) but I think `/usr/local/Homebrew/Library/Homebrew/vendor/bundle` would work.

reitermarkus (approved):
> There are some restrictions (symlinks, spaces, etc) but I think /usr/local/Homebrew/Library/Homebrew/vendor/bundle would work.

Well, we can at least try. 😉

PR #43286, opened 2018-01-29
Use version number for beaker-browser, starting at 0.7.11

optikfluffel:
Also use download url [from official website](https://beakerbrowser.com/docs/install) and add appcast.

---

- [x] `brew cask audit --download beaker-browser` is error-free.
- [x] `brew cask style --fix beaker-browser` reports no offenses.

---
I couldn't pass the style check because my comment above `url`

> does not match the expected comment format

and I didn't really know what to do about it because simply putting the wanted

> was verified as official when first introduced to the cask

just to let it pass felt wrong.

### DIFF
--- a/Casks/beaker-browser.rb
+++ b/Casks/beaker-browser.rb
@@ -1,9 +1,11 @@
 cask 'beaker-browser' do
-  version :latest
-  sha256 :no_check
+  version '0.7.11'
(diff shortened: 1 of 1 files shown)


gtmgianni:
You could write:

`# github.com/beakerbrowser/beaker was verified as official when first introduced to the cask`

vitorgalvao (approved):

PR #43360, opened 2018-01-30
Update addaps to 0.4.35

brianbrownton:

### DIFF
--- a/Casks/addaps.rb
+++ b/Casks/addaps.rb
@@ -1,6 +1,6 @@
 cask 'addaps' do
-  version '0.4.31'
-  sha256 '8feddc813f8006bd588b40077ce2548087141a1fda4e34e23419496202298788'
+  version '0.4.35'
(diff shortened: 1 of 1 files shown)


commitay:
0.4.38 is the latest version.

commitay:
@brianmorton I do appreciate the PRs updating Casks but if your script isn't checking what the actual latest version is that means maintainers have to check manually.

commitay:
Closing in favour of https://github.com/caskroom/homebrew-cask/pull/43361

brianmorton:
@commitay it does check based on the URL. So I just check based on the current package version and increment various parts of the version identifier and to see if a newer version is available. If, on the other hand, you are only interested in what the latest version is available on the package's homepage (which seems to be the case), please confirm.

commitay:
newer version doesn't necessarily mean it is the newest version or the correct stable version.

We have lots of beta / dev Casks that use the same URL as stable but with a newer version.

brianmorton:
@commitay alright I will make an effort to verify it is their latest official stable package before submitting PRs.

PR #43499, opened 2018-02-01
Update pretzel to 0.0.18

brianbrownton:


brianmorton:
The homepage doesn't list a version or release notes, but does provide a download to a non-versioned file ( https://download.pretzel.rocks/PretzelDesktop.dmg )... this is another situation which I'm not sure which you'd rather go with: `:latest` or this version... even though this version was only found through URL sniffing.

vitorgalvao:
Versioned is preferred. There’s no sniffing necessary in this case, the URL is right there in the appcast.

PR #43569, opened 2018-02-02
remove kanmusmemory

brianbrownton:
Don't know if you take PRs for removals, but the original cask is from 2013, doens't work on OSX 10.10+ seemingly, requires an old version of flash, etc.

The homepage is also a blogpost which is a little unusual IMHO.

If you want to keep the cask, there are updates available (`0.16` and `0.17`) which were found through URL sniffing however I was unable to find out if they are actual releases.


vitorgalvao:
> Don't know if you take PRs for removals

Gladly.

PR #43805, opened 2018-02-08
remove alarm-clock

brianbrownton:
It's at least 6 years old, and requires using a third party site for the `homepage` and `url` (wayback.archive.org).


commitay:
Ping @vitorgalvao 

Looks like this was originally added with wayback.archive.org links. https://github.com/caskroom/homebrew-cask/pull/5664

PR #43845, opened 2018-02-09
Fix sha256 checksum for MAMP 4.4

lead0r:

### DIFF
--- a/Casks/mamp.rb
+++ b/Casks/mamp.rb
@@ -4,7 +4,7 @@ cask 'mamp' do
     sha256 'b935d118b6e14d43cf2570be2a05166a15b824c3edc0ede98cc30e6ab5af3697'
   else
     version '4.4'
-    sha256 '390791c2a9e47e0b1c605290d6e05f09e634c44392d394c7eee43bf84ea3bfeb'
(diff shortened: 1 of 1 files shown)


vitorgalvao (changes requested):

lead0r:
The correct checksum can be found at the developer's website: https://www.mamp.info/en/downloads/

The file is too big to upload it to VirusTotal and I do not know if it works with .pkg files

commitay (approved):
Matches public `sha256`
https://www.mamp.info/en/downloads/
```
MAMP & MAMP PRO 4.4 (macOS)
Published: 2018-02-07
SHA-256: 9a86de2f1321851693bbf24f96c0aaca29fe7fb4d9b3c887aa464d41c36eed9c
```

vitorgalvao:
I’m not sure that in this case we can trust the public shasum without contacting the developer.

Both the download and the page with the shasum are on the same domain, and presumably on the same server. A malicious third-party that tampers with the first can easily tamper with the second (it’s just a tiny string of characters in an HTML page).

lead0r:
@vitorgalvao I agree with you, and even with the new checksum I got the following error: https://github.com/caskroom/homebrew-cask/issues/43859

PR #43928, opened 2018-02-12
Update xquartz: remove conflict

commitay:
Pending https://github.com/caskroom/homebrew-versions/pull/5241


commitay:
Seems to pick up that the existing pkg has been uninstalled, maybe?

```
>>> Leftover: pkgs
```

PR #43930, opened 2018-02-12
Update dingtalk to 4.2.39.2

brianbrownton:

### DIFF
--- a/Casks/dingtalk.rb
+++ b/Casks/dingtalk.rb
@@ -1,6 +1,6 @@
 cask 'dingtalk' do
-  version '4.2.60'
-  sha256 '0b3174e5e619e9ebc6f945ffce58463e8594eddfc061e364a25318c6b5aa8233'
+  version '4.2.39.2'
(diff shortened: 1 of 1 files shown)


brianmorton:
4.2.39.2 is the latest on their website... I think the early update jumped the gun and this should be reverted.

commitay:
>  I think the early update jumped the gun

I think they released a beta as stable for a few hours,  it was correct as the time it was merged.

> and this should be reverted

Yep, thanks.

brianmorton:
@commitay i'm sure there had to be a reason, trying to get a beta or non-stable passed you lot is an exercise in futility!

commitay:
Yeah, we're kinda ruthless about "correct" versions but it's because we need to be. 

I don't think I've linked [this](https://github.com/caskroom/homebrew-cask/issues/41211) previously, it's an example of what happens if users get switched to a beta by accident, we end up getting several follow up PRs to upgrade to the next beta because they might not realise they have been switched to the beta channel.

PR #43944, opened 2018-02-13
Fix logisim-evolution 2.14.4 sha256

miedzinski:

### DIFF
--- a/Casks/logisim-evolution.rb
+++ b/Casks/logisim-evolution.rb
@@ -1,6 +1,6 @@
 cask 'logisim-evolution' do
   version '2.14.4'
-  sha256 '5b659436dc719862a3c9815ebf6f8deb2499e31d2e91574c8dea789578d83908'
+  sha256 '3ca55caac762ca63d928e52f0f92013fb2a14d8d7031e378ef187c5969aef4f9'
 
   url "https://github.com/reds-heig/logisim-evolution/releases/download/v#{version}/logisim-evolution.jar"
   appcast 'https://github.com/reds-heig/logisim-evolution/releases.atom',


miedzinski:
I see that Travis check failed, but I'm pretty sure about that sha256, because

a) downloaded the .jar on two different machines in two different countries and computed digests there
b) downloads were over TLS connection
c) I believe there is no MiTM between me and Amazon's servers

vitorgalvao:
Welcome, and thank you for the PR.

miedzinski:
https://www.virustotal.com/#/url/e39ac3743b6c21a8c05b7f8a950f61473b1d2060e609419ea191d68c09aa1381/details

vitorgalvao:
You need to download and update the actual file, not the URL.

miedzinski:
Okay then...

https://www.virustotal.com/#/file/3ca55caac762ca63d928e52f0f92013fb2a14d8d7031e378ef187c5969aef4f9/details

URL is the same.

miedzinski:
I did like 4 checks already, is it enough?

vitorgalvao:
The posted instructions say:

> If the Cask is an `.app` that is [codesigned](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/codesign.1.html) (in a `.dmg` or `.zip` container) it can be uploaded and verified using [VirusTotal](https://www.virustotal.com/) by looking at the “Details” tab. 
>
>   **If there is no *Signature Info* section, VirusTotal verification is not enough**.

This is neither an `app` nor codesigned, so VirusTotal isn’t enough. Public confirmation by the developer is necessary.

miedzinski:
@alexandremalki could you help me?

```curl -L https://github.com/reds-heig/logisim-evolution/releases/download/v2.14.4/logisim-evolution.jar 2>/dev/null | openssl dgst -sha256```

computed on your side is all we need, I hope.

@brianmorton did you discover new version manually or using https://github.com/brianmorton/cask-automation? Maybe the checksum was wrong in the first place because your tool is somehow bugged?

**Comment by brianmorton** [comment]
@miedzinski 
the actual pull requests (and therefore the sha checksum) come from the `cask-repair` tool in almost all cases (including this case).

**Comment by brianmorton** [comment]
@miedzinski further, if I run your code snippet I get the already existing checksum:

```
$ curl -L https://github.com/reds-heig/logisim-evolution/releases/download/v2.14.4/logisim-evolution.jar 2>/dev/null | openssl dgst -sha256
3ca55caac762ca63d928e52f0f92013fb2a14d8d7031e378ef187c5969aef4f9
```

alexandremalki:
Czesc @miedzinski, i got the follwing hash --> 3ca55caac762ca63d928e52f0f92013fb2a14d8d7031e378ef187c5969aef4f9 .

miedzinski:
@alexandremalki thanks!

miedzinski:
@vitorgalvao I believe we just got public confirmation.

miedzinski:
Thank you! 🌴

PR #43957, opened 2018-02-13
Update bitpay to 3.14.1

brianbrownton:


brianmorton:
This one is actually a little confusing... as the github for it is `copay` but the homepage goes to bitpay... howver the github lists copay.io as the site...

So either this is not bitpay, or should have a second name stanza for copay?

commitay:
> So either this is not bitpay, or should have a second name stanza for copay?

Neither, bitpay and copay are different apps with releases from the same repo.

PR #43962, opened 2018-02-13
remove appshelf - homepage unavil since mid-2017

brianbrownton:
http://www.kedisoft.com/appshelf/
https://web.archive.org/web/*/http://www.kedisoft.com/appshelf/


commitay:
homepage was working two weeks ago https://github.com/caskroom/homebrew-cask/pull/43324#issuecomment-361405004

brianmorton:
Hey good idea to search previous PRs too ;)

Up to you - clearly they have uptime issues!

brianmorton:
Another cask is also affected:
totals 3.1.5
http://www.kedisoft.com/totals/

vitorgalvao:
`totals2` in [caskroom/versions](https://github.com/caskroom/homebrew-versions) is also affected. The whole kedisoft hosting is down.

brianmorton:
So is there a guideline for this? If we fail a check 3 days in a row is it now a candidate for removal? Should I add `totals` to the PR for removal or drop the PR?

commitay:
> So is there a guideline for this?

No specific rules.

> If we fail a check 3 days in a row is it now a candidate for removal? 

Probably want to wait a week or two so we aren't deleting and then re-adding Casks.

brianmorton:
OK so if I were to implement an SQLite DB and submit a PR for removal after either the homepage or URL failed for two straight weeks that would be fine?

commitay:
PRs for deletion should be checked first be being submitted, Casks may just need their homepage or url updated.

brianmorton:
Page is back up.

PR #44099, opened 2018-02-17
Remove deckard

gtmgianni:
The download link is broken, even if you remove the `www` subdomain from the download URL. Also, [the developer's Twitter account](https://twitter.com/deckard_ai) has been inactive since July 2017.


commitay:
https://deckard.ai/ works as a homepage but the download is behind a signup form.

commitay:
We may as well remove this as we can't fix it without signing up.

gtmgianni:
@commitay If you go to https://deckard.ai/download there is a download link available there, but that does not work either.

PR #44136, opened 2018-02-19
Update liteide to 33.2

noelvillador:

### DIFF
--- a/Casks/liteide.rb
+++ b/Casks/liteide.rb
@@ -1,11 +1,11 @@
 cask 'liteide' do
   version '33.2'
-  sha256 '2b773c078f46d880d6f8f3980ff153c47790068e647d9a9f9fecb0ad7628c67f'
+  sha256 'abaf9492c156048bff09fb3c6dac7657fbc6fb73389264f24b2cd03d81ad2e47'
 
   # github.com/visualfc/liteide was verified as official when first introduced to the cask
   url "https://github.com/visualfc/liteide/releases/download/x#{version}/liteidex#{version}.macosx-qt5.zip"
   appcast 'https://github.com/visualfc/liteide/releases.atom',
-          checkpoint: 'dcc2fc7a31e3f829790c067b09ab80ccbe27ddbc96625091cdd619a5fe0f7e45'
+          checkpoint: 'c19004e6cdb7414f81805b957fc88d0ad1dffc265952ffcf17455407b2d6ec43'
   name 'LiteIDE'
   homepage 'http://liteide.org/'
 


commitay:
Please open an issue at https://github.com/visualfc/liteide/issues to confirm that the `sha256` change is legitmate.

brianmorton:
@commitay @vitorgalvao this isn't the first (or even the second) PR we've seen updating a checksum recently for one of my PRs that was submitted through `cask-repair` - any idea if there's a pattern or problem here?

commitay:
Your PRs are fine, I think the problem is upstreams doing bugfix updates in-place soon after the initial release.

Usually it takes a day or two (particularly with versioned URLs) for someone to send an update PR so normally we might not see these.

brianmorton:
@commitay I agree that what you have described seems like the case... but then isn't it on us to fix the issue? Like maybe there is a 3 or 5 day window after the update PR where the author of the PR can submit an updated checksum without having the app developer chime in or something?

It just kind of sucks as an end user experience that because we are on the ball with updates that checksums may be different if there is a update-in-place afterwards, and `brew cask install <app>` won't work anymore... would using `--force` even solve this or the cask just not installable?

commitay:
> would using --force even solve this or the cask just not installable?

`sha256` mismatch means the Cask cannot be installed.

> isn't it on us to fix the issue? Like maybe there is a 3 or 5 day window after the update PR where the author of the PR can submit an updated checksum without having the app developer chime in or something?

While it is our problem to deal with, their isn't much we can do. We use VirusTotal to check codesigning which covers a lot of `apps` (this one isn't signed), apart from that contacting the developer is really our only option. 

Having a grace period for `sha256` changes would mean that we assume that the change is legitimate, which isn't something we can do.

> It just kind of sucks as an end user experience

Yeah but installing malware is worse.

brianmorton:
First, I recognize the technical merits of what you're saying and agree generally about checksums and the risk of installing malware.

BUT

Wouldn't having the grace period assume essentially no more risk than the originally committed PR/sha256 anyway? If we forget about VirusTotal for a moment, the builds go through CI which independently verifies the checksum, and if "upstream" was compromised before pushing the original update the end result is the same. What if it was compromised before pushing the version bump? We're already part way down a slippery slope where we can play the what if game for quite a while.

The "essentially" part in the above paragraph refers to the 3-5 day grace period window. If the source is compromised during that grace period then I agree this represents a risk. But the pragmatic side of me thinks the likelihood of it happening exactly like this is small compared to the major annoyance and decreased usage of homebrew cask of having more and more casks having... debated... checksums because we update too quickly and/or devs update-in-place.

There is already a certain amount of trust involved in using a tool like homebrew (and homebrew cask) - all security is a tradeoff between the experience/usability and how secure the system is. If people wanted to be as secure as possible I don't think they would use homebrew - they would get it from source. They use homebrew because they want it to work quickly and easily.

The other options are to use `:latest` and `:no_check`, or to take an array of checksums (which would obviously involve possible exploitation). Again, I'm not saying security is bad and checksums are useless, I just recognize there is a tradeoff between security and usability and I think the marker on this one may be in the wrong place.

My two cents.

commitay:
> Wouldn't having the grace period assume essentially no more risk than the originally committed PR/sha256 anyway? If we forget about VirusTotal for a moment, the builds go through CI which independently verifies the checksum, and if "upstream" was compromised before pushing the original update the end result is the same. What if it was compromised before pushing the version bump? We're already part way down a slippery slope where we can play the what if game for quite a while.

While you're not wrong, currently we have a reasonably straightforward rule for how we handle this. A grace period would significantly complicate it.

- Does the grace period reset each time the `sha256` changes?

- If it doesn't, should we accept multiple `sha256` changes within the original window?

- For some upstreams that we know to do new releases and not update in-place a grace period could very well mean we would accept a checksum change that we shouldn't.

> But the pragmatic side of me thinks the likelihood of it happening exactly like this is small compared to the major annoyance and decreased usage of homebrew cask of having more and more casks having... debated... checksums because we update too quickly and/or devs update-in-place.

> There is already a certain amount of trust involved in using a tool like homebrew (and homebrew cask) - all security is a tradeoff between the experience/usability and how secure the system is.

Yeah, but we're in a position that we can't really win here. People will complain if they can't install Casks with checksums mismatches and complain even louder if they can but it contains malware.

> They use homebrew because they want it to work quickly and easily.

They also expect security. That's why the current system was implemented.

PR #44248, opened 2018-02-23
Update neat to 4.5.0.99

qayshp:

- [x] `brew cask audit --download neat` is error-free.
- [x] `brew cask style --fix neat` reports no offenses.

### DIFF
--- a/Casks/neat.rb
+++ b/Casks/neat.rb
@@ -1,9 +1,8 @@
 cask 'neat' do
-  version '4.3.0.36'
-  sha256 '4db396a5e9e6dbf1d03670830959fbba5753caeb77d279a60f6f251f376f3f1d'
+  version '4.5.0.99'
+  sha256 '84bc54c01a239997bf18b86347161bde0a94ed66c222fe9bb1e977f4df1e3cdb'
 
-  # neatco.com was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


qayshp:
Note that the domain used to host the file has also changed, but as far as I can tell, this is the official URL.
That said, I may need to split this off into an older version and newer one, as the 4.X series is now discontinued (with 4.5.0.99 being the latest) and they have a newer one with a different feature set.

commitay:
> as the 4.X series is now discontinued (with 4.5.0.99 being the latest) and they have a newer one with a different feature set.

We should update to the current version or delete this Cask.

qayshp:
I was debating moving this cask to [caskroom/homebrew-versions](https://github.com/caskroom/homebrew-versions/). My main concern was that licenses from 4.X do not carry forward, so the upgrade experience would be bad. But now looking at some other versioned casks, this seems to be expected. When I get a chance, I'll update this to the latest and create a PR in versions.

(An example I looked at was [homebrew-versions/istat-menus5](https://github.com/caskroom/homebrew-versions/blob/master/Casks/istat-menus5.rb) vs. [homebrew-cask/istat-menus](https://github.com/caskroom/homebrew-cask/blob/master/Casks/istat-menus.rb) )

commitay:
Closing in favour of https://github.com/caskroom/homebrew-cask/pull/44963

PR #44325, opened 2018-02-25
Add AirPort Utility v5.6.1 Launcher Latest

luispuerto:

### DIFF
--- a/Casks/old-airport-utility.rb
+++ b/Casks/old-airport-utility.rb
@@ -1,0 +1,11 @@
+cask 'old-airport-utility' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://zcs.zyniker.org/download/67/'
+  name 'AirPort Utility v5.6.1 Launcher'
+  homepage 'https://zcs.zyniker.org/airport-utility-v5-6-1/'
+
+  app 'AirPort Utility 5.6.1/AirPort Utility 5.6.1 Launcher.app'
+  app 'AirPort Utility 5.6.1/AirPort Utility 5.6.1.app'
+end


vitorgalvao:
Thank you for the submission, @luisspuerto, but I think we’re going to pass on this one.

Making the decision on the grounds of this appearing to be too obscure (which is a documented policy). It was never submitted nor requested before, which from the age this appears to have, is an indication of lack of interest. That coupled with the lack of confidence the website transpires, what the app does, and how niche it is.

I’m still going to comment on how the cask should’ve been improved, for future reference.

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
+cask 'old-airport-utility' do
```
vitorgalvao: Does not conform to the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md)

vitorgalvao: Does not conform to the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
...
+  version :latest
+  sha256 :no_check
+
+  url 'https://zcs.zyniker.org/download/67/'
+  name 'AirPort Utility v5.6.1 Launcher'
```
vitorgalvao: Versions do not belong in `name`.

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
+cask 'old-airport-utility' do
+  version :latest
```
vitorgalvao: Should have a `version`.

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
...
+  sha256 :no_check
```
vitorgalvao: Should have a `sha256`.

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  url 'https://zcs.zyniker.org/download/67/'
```
vitorgalvao: `67` should be interpolated, and part of `version`.

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
...
+  name 'AirPort Utility v5.6.1 Launcher'
+  homepage 'https://zcs.zyniker.org/airport-utility-v5-6-1/'
```
vitorgalvao: `version` should be interpolated here.

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  app 'AirPort Utility 5.6.1/AirPort Utility 5.6.1 Launcher.app'
```
vitorgalvao: `version` should be interpolated here.

Casks/old-airport-utility.rb
```diff
@@ -0,0 +1,11 @@
...
+  app 'AirPort Utility 5.6.1/AirPort Utility 5.6.1.app'
```
vitorgalvao: `version` should be interpolated here.

luisspuerto:
Thanks a lot for the comments on how the cask has to be summited. 

I also thought that probably was a problematic app to have in homebrew-cask, but finally I wanted to give it a run, even for the sake of try to submit a cask for the first time and see the process. 

I'll try to reply to your comments. Thanks a lot, again, for taking the time to make.

Casks/old-airport-utility.rb
luisspuerto: OK!! The thing is, this app, or script, it's just running another app specific version that without that script couldn't be launched in the current versions of macOS.

Casks/old-airport-utility.rb
luisspuerto: I don't know the specific version of the script. It isn't provided. Not even in the web.

Casks/old-airport-utility.rb
luisspuerto: I tried to provided and even calculated it, but since I wasn't able to pin point or provide a version, the submission script suggested me that better to put ":latest".

Casks/old-airport-utility.rb
luisspuerto: I don't know if they upload in different directories by the version.

Casks/old-airport-utility.rb
luisspuerto: This version is static since the purpose of the app/script is to run only this specific version in the systems where it isn't longer possible.

Casks/old-airport-utility.rb
luisspuerto: This version is static since the purpose of the app/script is to run only this specific version in the systems where it isn't longer possible.

Casks/old-airport-utility.rb
luisspuerto: This version is static since the purpose of the app/script is to run only this specific version in the systems where it isn't longer possible.

luisspuerto:
The purpose of this app / script if to be able to run an old version of the Apple Airport Utility in the modern versions of the macOS. Some people prefer this older version basically because it is more rich-festured that the new one provided in the modern versions of the macOS. For instance, you can change the hostname of the Time Capsules in this version, while in the modern ones you can't. 

Again, I understand that perhaps it is not the specific type of apps that you have in homebrew-cask, but it's definitely an utility that some os us has to use from time to time. 

Perhaps I can talk with the developer to crear things a little bit. 

Anyhow, thanks for your time.

PR #44455, opened 2018-02-28
Update thingsmacsandboxhelper to 3.11

brianbrownton:

### DIFF
--- a/Casks/thingsmacsandboxhelper.rb
+++ b/Casks/thingsmacsandboxhelper.rb
@@ -1,6 +1,6 @@
 cask 'thingsmacsandboxhelper' do
-  version '3.10'
-  sha256 '238763169c381b0a518f8bd771507bbb14fb36f0ca9845575ec407db041a623b'
+  version '3.11'
+  sha256 'ae0face6c56b8da9aceaf157176d7ef2d4492be65471687740761b6b142e310c'
 
   # culturedcode.cachefly.net was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


brianmorton:
Update found by URL sniffing, although it is the same as on the stated homepage.

I do think this should probably be renamed to `things-helper` or something similar, and perhaps the `things` cask added (there is a 15-day trial). Rationale being it is strange to have a helper and not the root program available, and the naming convention would not help anybody looking for this actually find it IMHO.

Thoughts?

commitay:
> I do think this should probably be renamed to things-helper

https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md

> and perhaps the things cask added (there is a 15-day trial).

https://github.com/caskroom/homebrew-cask/pull/34955

brianmorton:
@commitay thank you, I am aware. Their own website references it as `Things Helper` and it is marketed as such. So much to the technical chagrin of developers, I would contend it should be named the same here... no matter if the actual app is slightly different. But I don't really care, other than trying to help people at large.

#34955 is outdated, it is available as a stand-alone DL from their homepage:
https://culturedcode.com/things/download/

commitay:
Can the trial be upgraded to the full version?

brianmorton:
things

commitay:
So the full version has to be downloaded from MAS.

> I would contend it should be named the same here... no matter if the actual app is slightly different

I concede that following the token reference sometimes leads to nonsensical naming but it was implemented because of severe problems with duplicates.  We still get PRs for "new" Casks because they don't follow the token reference.

PR #44498, opened 2018-03-02
Update URL of fastlane.zip installer

KrauseFx:
I'm the author of fastlane, working at Google. We recently moved all fastlane related infra services away from crashlytics/fabric, so the URL changed from AWS to the official website.

Let me know if you need any other kind of proof or verification


adidalal (approved):
Verified product owner status from https://github.com/fastlane/fastlane

KrauseFx:
Thanks for the fast merge 👍

PR #44712, opened 2018-03-10
Add Godot Mono.app v3.0.2

Sposito:
Godot has a separated build which supports C# Mono runtime, add a new cask which supports the stable version of it.


commitay:
Duplicate of https://github.com/caskroom/homebrew-cask/pull/44709

PR #44751, opened 2018-03-12
travis.yml: xcode9.3

commitay:
XCode 9.3 is on 10.13.

Casks with kexts will fail to install.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: true
 matrix:
   include:
     - os: osx
-      osx_image: xcode9.2
+      osx_image: xcode9.3
       rvm: system
   fast_finish: true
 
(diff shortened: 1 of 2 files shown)


commitay:
I was thinking of adding something to skip the install / uninstall step for Casks with kexts so the CI build doesn't always fail, but that would mean that we wouldn't be checking if the `install` is valid.

commitay:
Opened an issue with Travis CI about disabling `kext-consent` for the 10.13 image.

https://github.com/travis-ci/travis-ci/issues/9377

commitay:
@TravisBuddy stop

commitay:
The Xcode 9.3 / 10.13 image has been released.
http://blog.travis-ci.com/2018-04-09-say-bonjour-to-xcode-93-and-macos-high-sierra

`kext-consent` is enabled.

commitay:
@vitorgalvao I'm leaning towards merging this so don't have to manually check urls and checksums of Casks with conditionals or depends_on high_sierra. We'll still get a failed install for Casks with kexts so we'll know the install is valid at least.

vitorgalvao (approved):

PR #44763, opened 2018-03-12
add Family Tree Builder 7.0.2.155

brianbrownton:

Note some of my installs failed:
```
curl: (56) Unexpected EOF
Error: Download failed on Cask 'family-tree-builder' with message: Download failed: https://mhcache-myheritage.netdna-ssl.com/FP/FamilyTreeBuilder/family_tree_builder_7155.dmg
```

But I believe this is an issue with curl: https://github.com/curl/curl/issues/1618

### DIFF
--- a/Casks/family-tree-builder.rb
+++ b/Casks/family-tree-builder.rb
@@ -1,0 +1,11 @@
+cask 'family-tree-builder' do
+  version '7.0.2,155'
+  sha256 '67e346320510ee1e0be3ad5d10b3a68e68f29c63693e67470160be4da9140623'
+
+  # mhcache-myheritage.netdna-ssl.com was verified as official when first introduced to the cask
+  url "https://mhcache-myheritage.netdna-ssl.com/FP/FamilyTreeBuilder/family_tree_builder_#{version.major}#{version.after_comma}.dmg"
+  name 'Family Tree Builder'
+  homepage 'https://www.myheritage.com/'
+
+  app 'Family Tree Builder.app'
+end


brianmorton:
Even though CI passed, I believe there may be some additional files created or installed by this app. How would I go about figuring out what these are for addition to a `zap` stanza?

commitay:
CI only checks for some types of "leftovers" and it won't find files from `app`s because the app hasn't been launched. The CI checks are mostly only helpful for `pkgs`.

Usually for zaps I'd check the Cask in a clean VM so you don't have to sort through unrelated files.

`com.myheritage.familytreebuilder` is the bundle ID so we can search for that.

`$ find ~ -iname "*myheritage*"`

That will also find a plist that can have an `uninstall launchctl:`

You can use the dev scripts ([docs](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md)) to find the bundle id and launchjob.

`list_running_app_ids` `list_installed_launchjob_ids`

brianmorton:
@commitay really appreciate that. Any tips on how one runs a "clean VM" of macOS? I admit I haven't spent very much time trying to virtualize macOS recently, but back when I did look into this it was... not very feasible.

Is something like [this](https://github.com/geerlingguy/macos-virtualbox-vm) what you mean? How did you guys get Travis to run macOS?

Apologies for all the questions...

commitay:
> Apologies for all the questions...

No problem!

> Any tips on how one runs a "clean VM" of macOS?

Just to be clear this isn't necessary as I'm guessing that you're using this Cask personally as you are adding it.

Maintainers usually use a VM because installing some Casks will litter your system with files, pkgs, launchjobs and installing every Cask we work on will make our systems a mess (even with uninstall/zap).

Also to reproduce/fix user issues we might need to "break" things that we wouldn't want to do on our main install.

> ... not very feasible.

Yeah, using Virtualbox for mac can be kinda painful. Personally I use Fusion and a couple of other tools that make spinning up VMs pretty easy.

> Is something like this what you mean?

Yep, that would work. https://github.com/timsutton has some stuff about virtualisation as well and https://github.com/busterc/osx-iso may also work for ISOs (haven't used it personally).

> How did you guys get Travis to run macOS?

Travis has macOS hardware running vSphere, we just set what version we want in the travis config.
https://github.com/caskroom/homebrew-cask/blob/d216a98c086e9be44828a8e9b1f59fae84018ec1/.travis.yml#L7-L8

vitorgalvao (approved):

PR #44849, opened 2018-03-14
Use explicit `URI#open`.

reitermarkus:


vitorgalvao:
What’s the difference?

reitermarkus:
The new `Security/Open` cop complains about it. See https://github.com/Homebrew/brew/pull/3923.

PR #44930, opened 2018-03-17
Add uninstall_preflight and _postflight to Virtualbox 5.2.8,121009

RulerOf:
Virtualbox cask installation and uninstallation will fail if the
application or any virtual machines are running. This commit adds an
uninstall_preflight stanza to the cask that saves the state of all
running VMs for all users, then cleanly closes the application with a
ten second timeout, after which it will kill -9 any still-running
VirtualBox binaries.

Currently, uninstalling the virtualbox cask deletes the extension pack
binaries, leaving the associated cask in an inconsistent state. This
commit adds an uninstall_postflight stanza that forces uninstallation of
the virtualbox-extension-pack cask when the virtualbox cask is
uninstalled.

### DIFF
--- a/Casks/virtualbox.rb
+++ b/Casks/virtualbox.rb
@@ -10,6 +10,31 @@ cask 'virtualbox' do
 
   pkg 'VirtualBox.pkg'
 
+  uninstall_preflight do
+    def get_process_list_array(filter = nil)
+      psaux = system_command('/usr/bin/env', args: ['ps', 'aux'], sudo: true).stdout.split("\n")
+      filter.nil? ? psaux : psaux.select { |line| line.match(filter) }
+    end
+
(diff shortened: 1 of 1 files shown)


commitay:
> adds an uninstall_postflight stanza that forces uninstallation of the virtualbox-extension-pack cask

This used to be part of the Cask and was removed because it breaks with `brew cask upgrade`

RulerOf:
> This used to be part of the Cask and was removed because it breaks with brew cask upgrade

Hmm... damned if you do, damned if you don't. I have to agree that it probably makes more sense to defer to keeping `brew cask upgrade` error-free than the particular edge case I ran into.

I'll close this PR and resubmit without the `uninstall_postflight` stanza.

commitay:
Please reopen this and force push.

RulerOf:
Sorry... Should I still work with this PR or stick with the new one?

commitay:
It's already open so the new one is fine.

PR #44931, opened 2018-03-17
Add uninstall_preflight to Virtualbox 5.2.8,121009

RulerOf:
Virtualbox cask installation and uninstallation will fail if the
application or any virtual machines are running. This commit adds an
uninstall_preflight stanza to the cask that saves the state of all
running VMs for all users, then cleanly closes the application with a
ten second timeout, after which it will kill -9 any still-running
VirtualBox binaries.

This is a fixed version of #44930

### DIFF
--- a/Casks/virtualbox.rb
+++ b/Casks/virtualbox.rb
@@ -10,6 +10,27 @@ cask 'virtualbox' do
 
   pkg 'VirtualBox.pkg'
 
+  uninstall_preflight do
+    def get_process_list_array(filter = nil)
+      psaux = system_command('/usr/bin/env', args: ['ps', 'aux'], sudo: true).stdout.split("\n")
+      filter.nil? ? psaux : psaux.select { |line| line.match(filter) }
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you for the submission, but this is really atypical, complex, and flimsy. I can already see predict the bug reports; relying so much on system tools for such a simple thing is unlikely to end well.

I’d rather use an [`uninstall :quit`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#summary-of-keys), if that saves the state. If not and the installation fails when there are open VMs, so be it. I don’t think it’s hard to figure out why it failed and proceed accordingly.

This case also imposes our will on the user by force quitting — in the same situation, I’d rather the installation fails and it lets me do things my way. It’s more likely that I don’t want to risk an upgrade at this point and finish my work. We should err on the side of doing less.

Still, the contribution is appreciated!

PR #45033, opened 2018-03-21
JIRA Client: adding zap stanza

ondrejfuhrer:


ondrejfuhrer:
Question: There were also some files in `/private/var/folders/` under some hash, should that be a part of `zap` as well? Or what are those files?

commitay:
We almost always leave `/var` alone, only 7 Casks have it in either`delete` or `zap` and a couple actually need for it to be removed.

ondrejfuhrer:
@commitay Cool, thanks!

PR #45034, opened 2018-03-21
Update the Amethyst cask to 0.12.0

issyl0:
- I installed this via `brew cask install` and when I launched it it
  offered me an update. I thought I'd give this update back to Cask as
  well.
- The `amethyst/versions` path appears to have changed to be
  `Amethyst/releases/download/v[version]`, so I updated that too and
  left a comment. The `cask audit` step is unhappy with this, as it
  doesn't like the homepage and download URLs to be different. The
  homepage URL links to the download URL for the specific version,
  though.

- [] `brew cask style --fix {{cask_file}}` reports no offenses.

### DIFF
--- a/Casks/amethyst.rb
+++ b/Casks/amethyst.rb
@@ -6,13 +6,15 @@ cask 'amethyst' do
     version '0.10.1'
     sha256 '9fd1ac2cfb8159b2945a4482046ee6d365353df617f4edbabc4e8cadc448c1e7'
   else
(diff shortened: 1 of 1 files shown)


issyl0:
Thanks, 🤖. Should I squash those two commits together to make for a cleaner history? I should have looked at URLs for previous versions and seen that they were still live.

commitay:
Thanks for the PR!

PR #45138, opened 2018-03-24
xpra: remove allow_untrusted

commitay:


commitay:
`zap trash` error, `install` is successful.

PR #45142, opened 2018-03-24
Remove hoster

commitay:
Can't be installed via CLI.


commitay:
```
==> Standard Output of failed command:
installer: Cannot install on volume / because it is disabled.
installer: version check failure. to install require MacOSX 10.4.4 later.
```

PR #45154, opened 2018-03-24
pkg.md, rejected_casks.md: allow_untrusted

commitay:
https://github.com/caskroom/homebrew-cask/issues/45146


vitorgalvao:
The [main docs](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/pkg.md#pkg-allow_untrusted) also need to be updated, to explain the option exists for users who want to do in their own taps, but to not be used in the main repos.

commitay:
Updated the `pkg` docs.

commitay:
Added to `audit` so it doesn't sneak back in. https://github.com/Homebrew/brew/pull/3974

PR #45464, opened 2018-04-03
Update adobe-photoshop-lightroom: uninstall preflight

commitay:
The `system_command` uninstall calling `brew` is broken.

### DIFF
--- a/Casks/adobe-photoshop-lightroom.rb
+++ b/Casks/adobe-photoshop-lightroom.rb
@@ -28,7 +28,7 @@ cask 'adobe-photoshop-lightroom' do
   end
 
   uninstall_preflight do
-    system_command 'brew', args: ['cask', 'uninstall', 'adobe-photoshop-lightroom600']
+    system_command "#{HOMEBREW_PREFIX}/bin/brew", args: ['cask', 'uninstall', 'adobe-photoshop-lightroom600']
   end
 
   zap trash: [


vitorgalvao:
> The `system_command` uninstall calling `brew` is broken.

What happens?

commitay:
Same thing as `virtualbox` https://github.com/caskroom/homebrew-cask/issues/45428

```
==> Un-installing artifact of class Hbc::Artifact::PreflightBlock
==> Executing: ["brew", "cask", "uninstall", "virtualbox-extension-pack"]
Error: No such file or directory - brew
```

vitorgalvao:
Looks like it may not be reading `PATH`.

commitay:
Yes, `homebrew_prefix/bin` is filtered.

commitay:
Ah, actually thinking about this some more it might not work. I think this Cask needs to be applied on a clean install of `adobe-photoshop-lightroom600`.

vitorgalvao:
> > There’s always *something* with Adobe. We’re not going to keep bending over backwards.
>
> Should we just delete it?

I’m fine with that. Pinging @danielbayley as he’ll probably want to port this over to his adobe casks repo.

danielbayley:
> he’ll probably want to port this over to his adobe casks repo

😅  Not really! Particularly, given that I don't use it… although it should probably _technically_ be in there.

If one of you wants to offload it in the form of a PR and it passes the CI, then I'll merge it.

Also, would this problem not be solved with the following?
~~~rb
uninstall_preflight do
  system_command "#{HOMEBREW_PREFIX}/bin/brew", args: ['cask', 'uninstall', 'adobe-photoshop-lightroom600']
end
~~~

commitay:
> Also, would this problem not be solved with the following? ...

Yep, thanks! Guess we can keep this until the next time it breaks.

commitay:
>  since calling brew recursively in a Cask is a pretty ghetto workaround to begin with

Agreed, I now regret having adding it to `virtualbox`

> This should be fixed in SystemCommand

I think this is the third or forth time filtering `#{HOMEBREW_PREFIX}/bin` has been an issue, should we disable it completely for Cask rather than specific cases when needed?

Having `#{HOMEBREW_PREFIX}/bin` filtered for some cases and not others is inconsistent and somewhat confusing.

PR #45637, opened 2018-04-05
Add parts-library caveats to Fritzing 0.9.3b

ELLIOTTCABLE:
(No checkboxes; didn't change anything but a caveats clause.)

### DIFF
--- a/Casks/fritzing.rb
+++ b/Casks/fritzing.rb
@@ -8,6 +8,8 @@ cask 'fritzing' do
   name 'Fritzing'
   homepage 'http://fritzing.org/home/'
 
+  app 'Fritzing.app'
+
   caveats <<~EOS
     There's a known issue with Fritzing 0.9.3b, causing it to corrupt its own
     parts-library on first-launch. To avoid this, you must launch Fritzing
@@ -24,6 +26,4 @@ cask 'fritzing' do
 
     See: https://github.com/fritzing/fritzing-app/issues/3308 for more.
   EOS
-
-  app 'Fritzing.app'
 end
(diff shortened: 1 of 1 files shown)


commitay:
We don't have version specific caveats. Asking upstream to make a release that we can upgrade the Cask to would solve this problem.

ELLIOTTCABLE:
@commitay The problem is that there are going to be no moer upstream releases (the maintainers have been working on a full rewrite for a few years, and have ignored requests for a release including bugfixes like these. `)=`

Given,

1. This isn't going to be fixed upstream,
2. it's a popular, necessary project for particular tasks, despite the unmaintained status of the (supposedly-legacy?) Qt codebase,
3. the problem in question *is* technically install-related (it's the first-time-launch self-setup behaviour that's broken),
4. and finally, that a Cask user is *already* staring at their Terminal, right after installing the app, and is presumably about to launch the app for the first time …

I dunno. It's your call! But it really seems like a worthwhile tip, to save the user of all the trouble of launching it the normal way, Googling the resulting errors, and finding they have to entirely uninstall and re-install the app to proceed …

I know it's technically against policy, but I *really* urge you to consider this a special case. `(=`

vitorgalvao:
There’s no point in making the caveat version specific — users of a previous version wouldn’t see it anyway.

This message seems acceptable to me; I can see the point to saying it’s installation-specific.

vitorgalvao (approved):

vitorgalvao:
@ELLIOTTCABLE And that Travis error is exactly why you *always* need to run the commands in the boxes, even if you think your changes don’t warrant it. They do.

ELLIOTTCABLE:
@vitorgalvao I ran them, but they both passed on my local machine! Yikes.

Cool! Let me know if there's anything else I can do. 😊

PR #45718, opened 2018-04-09
Update teamviewer, teamviewer-host: remove uninstall

commitay:
Fixes https://github.com/caskroom/homebrew-cask/issues/15322#issuecomment-379603267

Different issue than originally reported in https://github.com/caskroom/homebrew-cask/issues/15322

### DIFF
--- a/Casks/teamviewer-host.rb
+++ b/Casks/teamviewer-host.rb
@@ -11,12 +11,21 @@ cask 'teamviewer-host' do
 
   pkg 'Install TeamViewerHost.pkg'
 
-  uninstall pkgutil:   'com.teamviewer.*',
+  # This pkg and file should not be added to uninstall or zap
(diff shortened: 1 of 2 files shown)


commitay:
@TravisBuddy stop

commitay:
@vitorgalvao https://github.com/caskroom/homebrew-cask/pull/45718/commits/55ae7b059228a21a41d9dfd3ba8a93d1c34d5820 fixes the issue but any upstream changes to the pkgs (i.e. the files within `com.teamviewer.teamviewer13AuthPlugin` are added to a `pkg` id that we `uninstall`) will cause further issues. I think removing the `uninstall` altogether is the safest option.

vitorgalvao:
> I think removing the `uninstall` altogether is the safest option.

Agreed.

vitorgalvao:
Added both casks to https://github.com/caskroom/homebrew-cask/issues/24377.

PR #45747, opened 2018-04-09
Update psychopy to 1.90.1

gtmgianni:


gtmgianni:
There are two "flavours" of this application – one for Python 2 and one for Python 3. Which should be used for this cask?

vitorgalvao:
Whenever we can, let’s go with the future (python 3).

vitorgalvao:
Whenever we can, let’s go with the future (python 3). But if there are two different apps with different names, we may also include both.

vitorgalvao:
Whenever we can, let’s go with the future (python 3). But if there are two different apps with different names, we may also include both.

PR #45778, opened 2018-04-10
Update yandex-disk cask

glukki:

Today, without this fix, installation was failing with `Error: Checksum for Cask 'yandex-disk' does not match`.
Unfortunately, they publish new versions under the same URL, so checksum breaks install.
Luckily, the image is downloaded over HTTPS.


glukki:
CI check result is false-negative

vitorgalvao:
> Unfortunately, they publish new versions under the same URL, so checksum breaks install.

So far it has happened once. There’s no reason yet to believe this is a pattern. Please follow the normal procedure to update the `sha256`.

glukki:
@vitorgalvao checks will fail again, as I’ll need to update only `sha256`, right?

PR #45921, opened 2018-04-13
Update qcad to 3.20.1

brianbrownton:

### DIFF
--- a/Casks/qcad.rb
+++ b/Casks/qcad.rb
@@ -1,10 +1,18 @@
 cask 'qcad' do
   version '3.20.1'
-  sha256 'e24fb8ef40bf1dac7c830bfe1ab2c41810961b72d6231dcec883dbc15b1f1d8c'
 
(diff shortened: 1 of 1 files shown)


claui (commented):
Is there a reason we choose the `10.9–10.13.dmg` edition over the `10.10–10.13.dmg` one?

The [download page](https://www.qcad.org/en/qcad-downloads-trial) appears to advertise the former just for macOS 10.9 while it says the latter is for macOS 10.10–10.13.

Does the 10.10–10.13 edition have any feature the 10.9 edition doesn’t?

brianmorton:
You might need @commitay or @vitorgalvao to chime in, but I believe the cask could probably be updated with an `if/else` block.

vitorgalvao:
Since we only officially support the current macOS version and one back, we could even get rid of the `if`. No harm in having it there if it’s not broken and poses no maintenance burden, though.

PR #45939, opened 2018-04-14
Remove travis buddy

commitay:
With the high volume of traffic we have in Caskroom combined with Casks that fail with unfixable errors make the notifications somewhat annoying.

Improving our build scripts to collect and handle errors might help but that is a separate issue that will require a reasonable amount of work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,6 @@ notifications:
   webhooks:
     urls:
       - "https://webhooks.gitter.im/e/712d699360b239db14a5"
-      - "https://www.travisbuddy.com/"
     on_success: never
     on_failure: always
     on_start: never


vitorgalvao:
> make the notifications somewhat annoying

Do you get emails for everything?

commitay:
No, web notifications.

vitorgalvao:
I also use web notifications, and TravisBuddy is a net win for me. Not having to click around and waiting for Travis to load to see an error is tremendously useful.

Since we’re the ones that spend most time with the PRs, let’s see if we can get a consensus.

**How does TravisBuddy make your web notifications more annoying?** Because to me it’s neutral, notifications-wise. Here’s what I do:

1. Open https://github.com/notifications in the browser.
2. Run (via Alfred) a JXA script I made that will:
3. Check if there are unread notifications. If yes, go to **4**, else jump to **5**.
4. Go into https://github.com/notifications/participating and stop.
5. Grab all notifications (by this point, they’re all for things I haven’t participated in).
6. Ignore closed and merged issues and PRs.
7. Open all remaining in new browser tabs.

commitay:
I leave GitHub open in a browser most of the day, so I already get a notification whenever an issue or PR is opened or commented on. Getting notifications for _every_ build failure is annoying and makes notifications less useful to me. 

Seeing the build output without having to open Travis is useful, _the notification_ is not.

I've been preemptively commenting on builds that I _know_ will fail with `@TravisBuddy stop` just so I don't get a notification a couple of minutes later.

Blocking the Travis Buddy user doesn't work so I think the the only other option would be to stop watching the repo which isn't something that I want to do.

vitorgalvao:
Alright, if it’s that annoying to you, let’s get rid of it.

vitorgalvao (approved):

bluzi:
@commitay @vitorgalvao 
> I've been preemptively commenting on builds that I know will fail with @TravisBuddy stop just so I don't get a notification a couple of minutes later.

Sorry to bump this old issue, but just had an idea: 
You've said that you use `@TravisBuddy stop` on builds since you already know it will fail - What if TravisBuddy won't comment on those builds at all? Do you think we could tag those builds somehow? Maybe using regex on build logs? 

Would love to hear your thoughts on this idea :)

commitay:
> Sorry to bump this old issue

No problem!

> What if TravisBuddy won't comment on those builds at all? Do you think we could tag those builds somehow? Maybe using regex on build logs?

Something like this could work it but it would need to be very finely tuned to a point that not be practical.  Some of the expected failures are very similar to an actual error that shouldn't be disregarded.

I think overhauling our build scripts would be the the best approach as it would allow us to handle these type of expected failures and other general build problems at the same time.

bluzi:
I see, thank you for sharing! 
We've earned so many great features by learning from failed integrations like this one, so I really appreciate your response. :) 

If you come up with an idea of how we could improve the service to better fit your needs (regardless of your in-house changes that needs to be done before trying the integration again), please let me know.

PR #45943, opened 2018-04-14
Delete Festify cask

NeoLegends:
Hey!

The new version of Festify is completely browser-based and the old versions are deprecated, so there's no need for a brew cask anymore. :)


vitorgalvao:
Thank you.

PR #46031, opened 2018-04-17
Update GHDL 0.35

sww1235:
Wasn't able to audit or check style due to editing purely via github.


sww1235:
The GHDL project changed to a github organization, thus I could not use cask-repair. 

I am not certain exactly how to add the artifact stanza correctly.

Also a question. This application does not seem to fit in homebrew casks, as it does not provide a .app, it just provides binary files. This seems like a better fit now for normal homebrew.

commitay:
> This application does not seem to fit in homebrew casks, as it does not provide a .app, it just provides binary files. This seems like a better fit now for normal homebrew.

Agreed. 

@vitorgalvao Should we remove this?

vitorgalvao:
> This seems like a better fit now for normal homebrew.

That is correct. The policy is documented in [faq/rejected_casks](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/rejected_casks.md):

> The GHDL project changed to a github organization, thus I could not use cask-repair.

What do you mean? `cask-repair` doesn‘t care.

> Should we remove this?

Yes. @sww1235 Will you submit to Homebrew/core?

sww1235:
I am working on submitting to homebrew/core, but I believe I have discovered the reason why this is in homebrew-casks in the first place. 

GHDL is written almost entirely in Ada, which requires a special compiler from GCC. There is no similar support for LLVM/clang as far as I can tell. 

I have submitted a discourse post asking for assistance on using a specific compiler, and will keep this pull request updated with any further info I find.

commitay:
https://discourse.brew.sh/t/creating-new-formula-for-code-in-ada/2086

commitay:
Adding ada to gcc is possible but it's kind of niche and would further complicate an already complex (and popular) formula.

Cask can install the binaries but installing includes and libs probably isn't something we want to do.

I think the best way forward would be for this to removed from Cask in favour of a third-party formula that can perform a complete install of upstreams pre-built tarball.

vitorgalvao:
Fine with removal.

sww1235:
Thank you for working on this. I understand that it doesn't necessarily fit under cask or will work in core.

PR #46077, opened 2018-04-18
Add Todo Cloud

ondrejfuhrer:

### DIFF
--- a/Casks/todo-cloud.rb
+++ b/Casks/todo-cloud.rb
@@ -1,0 +1,16 @@
+cask 'todo-cloud' do
+  version '10.3.1'
+  sha256 '76eaed95d790bdaf62431b9284f3ce504801ba2de2d5a4978a1ce74d69b61704'
+
+  url "http://builds.appigo.com/todo-cloud/macos/Todo+Cloud-PROD-#{version}-mac.dmg"
+  name 'Todo Cloud'
+  homepage 'http://www.appigo.com/'
+
+  app 'Todo Cloud.app'
+
+  zap trash: [
+               '~/Library/Preferences/com.appigo.todo10.plist',
+               '~/Library/Preferences/com.appigo.todo10.helper.plist',
+               '~/Library/Saved Application State/com.appigo.todo10.savedState',
+             ]
+end
--- a/Casks/todo-cloud.rb
+++ b/Casks/todo-cloud.rb
@@ -9,8 +9,8 @@ cask 'todo-cloud' do
   app 'Todo Cloud.app'
 
   zap trash: [
-               '~/Library/Preferences/com.appigo.todo10.plist',
-               '~/Library/Preferences/com.appigo.todo10.helper.plist',
-               '~/Library/Saved Application State/com.appigo.todo10.savedState',
+               "~/Library/Preferences/com.appigo.todo#{version.major}.plist",
+               "~/Library/Preferences/com.appigo.todo#{version.major}.helper.plist",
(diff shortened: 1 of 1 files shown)


Casks/todo-cloud.rb
```diff
@@ -0,0 +1,16 @@
...
+  sha256 '76eaed95d790bdaf62431b9284f3ce504801ba2de2d5a4978a1ce74d69b61704'
+
+  url "http://builds.appigo.com/todo-cloud/macos/Todo+Cloud-PROD-#{version}-mac.dmg"
+  name 'Todo Cloud'
+  homepage 'http://www.appigo.com/'
+
+  app 'Todo Cloud.app'
+
+  zap trash: [
+               '~/Library/Preferences/com.appigo.todo10.plist',
```
commitay: Use `#{version.major}` here instead of `10`

ondrejfuhrer: Done 👍

vitorgalvao: We should think about starting to use `*` instead, in `zap`s. I see no strong reason why `zap`s shouldn’t remove these files for *all* versions of the software.

ondrejfuhrer: Will it work here already? Or support for that has to be implemented

vitorgalvao: It’ll work already.

ondrejfuhrer: There is maybe one thing to consider. `*` in that case means everything, so if you have multiple apps from the same company, you could break the other, in theory (that could but not must be installed via `brew cask`).

So i.e. you have, as here, `Todo Cloud` and then let's imagine you have `Todo Desktop`, which will have following plists:
`~/Library/Preferences/com.appigo.todo10.plist`
`~/Library/Preferences/com.appigo.todo-desktop-10.plist`

By setting `~/Library/Preferences/com.appigo.todo*.plist` you would then trash also the plist for the other one, which is not wanted. So I'm not sure if it is worth setting that way, since it can cause those troubles.

ondrejfuhrer: So in this case it would be better to have an option to put in `\d+` instead.

vitorgalvao: > By setting `~/Library/Preferences/com.appigo.todo*.plist` you would then trash also the plist for the other one

That is indeed a problem.

> So in this case it would be better to have an option to put in `\d+` instead.

`zap` uses wildcards for glob expansion, not regular expressions. So this would have to be `+([0-9])`, or something, which I’m not sure we support.

Pinging @rolandwalker.

rolandwalker: Ping of the Devil, and the Devil himself may appear.

From memory, I only implemented tilde expansion in `zap`; did globs come later?

rolandwalker: Whatever the order of implementation, globs are in there now per https://github.com/Homebrew/brew/blob/1d9746db84d63acd4fd7c2b9b28569892d9a0d9f/Library/Homebrew/cask/lib/hbc/artifact/abstract_uninstall.rb#L236 .

`Pathname.glob` supports the syntax at https://ruby-doc.org/core-1.9.3/Dir.html#method-c-glob , so you can use `[0-9]`.

Best.

vitorgalvao: I meant to ping @reitermarkus, but glad to read from you again! Hope you’ve been well.

Thank you for the clarification. `[0-9]*` seems like the best we can do (which still just means one digit and then anything), but that might be enough for most cases.

Hope my erroneous ping wasn’t inconvenient. Thank you for taking the time to reply!

PR #46211, opened 2018-04-22
Remove scribbleton

gtmgianni:
The server returns an HTTP 500 error whenever `curl` attempts to download the installer for this cask. Is this reproducible for others?

Note: I can download the installer via Safari without any issues, but can't download it via the command-line.


adidalal:
Confirmed, they're doing some cookie token checking.

vitorgalvao:
Have you tried with any of the [parameters](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#additional-https-url-parameters)?

gtmgianni:
@adidalal @vitorgalvao I'm not sure which of those cookies are not necessary.

commitay:
These cookies are likely of break easily, lets just remove this if upstream wants to make it that hard to download via CLI.

commitay:
Yes, we can do that in ruby using a `url do` block but we aren't going to add it to the Cask.

commitay:
@gtmgianni Can you update the PR to remove the Cask please?

commitay:
Thanks!

PR #46215, opened 2018-04-22
Add Solstice Client :latest

iandrewt:
Adds Solstice client, an app used for mirroring the display to a TV in education/enterprise

### DIFF
--- a/Casks/solstice.rb
+++ b/Casks/solstice.rb
@@ -2,9 +2,11 @@ cask 'solstice' do
   version :latest
   sha256 :no_check
 
-  url 'http://www.mersive.com/downloads/SolsticeClientMacWeb.zip'
+  url 'https://www.mersive.com/downloads/SolsticeClientMacWeb.zip'
   name 'Solstice Client'
   homepage 'https://www2.mersive.com/'
 
-  app 'SolsticeClientMacWeb.app'
+  installer manual: 'SolsticeClientMacWeb.app'
+
+  uninstall delete: '~/Desktop/Solstice Client.app'
 end
--- a/Casks/solstice.rb
+++ b/Casks/solstice.rb
@@ -1,0 +1,10 @@
+cask 'solstice' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.mersive.com/downloads/SolsticeClientMacWeb.zip'
+  name 'Solstice Client'
+  homepage 'https://www2.mersive.com/'
+
+  app 'SolsticeClientMacWeb.app'
+end


Casks/solstice.rb
```diff
@@ -0,0 +1,12 @@
...
+  version :latest
+  sha256 :no_check
+
+  url 'https://www.mersive.com/downloads/SolsticeClientMacWeb.zip'
+  name 'Solstice Client'
+  homepage 'https://www2.mersive.com/'
+
+  installer manual: 'SolsticeClientMacWeb.app'
+
+  uninstall delete: '~/Desktop/Solstice Client.app'
```
vitorgalvao: It’s unlikely moving this app is the only thing the installer says. Furthermore, it’s unlikely users will keep the installed app in their Desktop.

vitorgalvao (changes requested):
We need further documentation on this app’s un/install process. As it is, we cannot accept this cask.

Casks/solstice.rb
commitay: > it’s unlikely users will keep the installed app in their Desktop

There isn't much we can do about this. It runs a GUI installer that doesn't seem to accept any flags so we can't move it ourselves.

vitorgalvao: > There isn't much we can do about this.

We can `postflight` to move it to `appdir`.

commitay: It's a GUI installer?

vitorgalvao: Ah, right. Still, deleting it from the Desktop is bound to be useless. This would be another good candidate for https://github.com/caskroom/homebrew-cask/issues/8211.

commitay: Should we refuse it then?

vitorgalvao: Yes, depending on what @iandrewt can gather, if anything. For all we know this might setup an uninstall script in a fixed location that would acceptably deal with the matter.

iandrewt: Hi, sorry I've been inactive since submitting this. Did some testing, moving the app from the desktop to Applications doesn't result in it being moved back when opening it again. Is this the kind of info you're after?

vitorgalvao: No, the issue is that since this has an installer (as opposed to just a .app bundle) it likely does some modifications to your system (like installing `launchd` jobs). For this to be accepted, we need to have an `uninstall` that undos those changes.

iandrewt: Ah, did some digging and it looks like the only other file system thing it does is create a couple files in `~/Library/Application Support` (the file `solsticeMRU_V2.dat` and the folder `Solstice`). Nothing at all in `/Library`, and no `launchd` jobs

vitorgalvao:
Going to refuse since we can’t build a proper uninstall. Still, thank you for the contribution and the fast responses.

PR #46223, opened 2018-04-23
Update nomachine to 6.1.6_10

luispuerto:

### DIFF
--- a/Casks/nomachine.rb
+++ b/Casks/nomachine.rb
@@ -9,15 +9,15 @@ cask 'nomachine' do
   pkg 'NoMachine.pkg'
 
   # A launchctl job ordinarily manages uninstall once the app bundle is removed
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):

commitay:
I'd think we should remove the `zap` altogether rather than make it the `uninstall`. 

Removing `com.nomachine.uninstall` means that we aren't duplicating the uninstall and I'm also getting `kextunload` errors with this.

commitay:
I don't know but running the script without any args doesn't do anything.

vitorgalvao:
That’s because the script only runs if the app bundld isn’t present anymore. If you inspect it, it’s a few functions at the too and then a huge `if` statement checking for that condition.

commitay:
Cask will run the script before the app is removed.

PR #46271, opened 2018-04-24
Add UHK Agent v1.2.0

katrielalex:
This is the configuration software for a CrowdSupply keyboard that's just starting to ship.


commitay:
https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md

> Driver: Software to make a hardware peripheral recognisable and usable by the system. If the software is useless without the peripheral, it’s considered a driver.

Please submit this to [`caskroom/drivers`](https://github.com/caskroom/homebrew-drivers)

PR #46274, opened 2018-04-24
Update vuescan to 9.6.07

kuerb:


vitorgalvao (changes requested):

vitorgalvao:
> I downloaded the file from the same link as before.

But the version on the server could have been compromised (it has happened before).

> I also still have to old dmg

Not necessary. The VirusTotal verification is enough. Thank you.

PR #46330, opened 2018-04-25
Make Mendeley-Desktop 1.17.13+ powersave-friendly

physicalist:
This post-install rule patches Mendeley Desktop.app's Info.plist, adding
the NSSupportsAutomaticGraphicsSwitching flag which enables significant
power savings on MacBook Pros with a dedicated graphics card.
Currently, since this flag is missing Mendeley Desktop.app forces MacOS
to switch to discrete graphics which drains the battery much more than
necessary. However, the app doesn't really need the discrete card to
function, as I have extensively tested on a late 2012 Macbook Pro Retina
for over a year.

### DIFF
--- a/Casks/mendeley.rb
+++ b/Casks/mendeley.rb
@@ -7,4 +7,9 @@ cask 'mendeley' do
   homepage 'https://www.mendeley.com/'
 
   app 'Mendeley Desktop.app'
+
+  preflight do
+    # FIX: add Application variable to prevent Mendeley from requiring discrete graphics card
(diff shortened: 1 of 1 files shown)


commitay:
Thanks for the PR but we don't "patch" Casks unless absolutely necessary.

physicalist:
Well, as far as "necessary" goes, this is something that more than halves battery time if unfixed and bug reports (by myself and many(!) others) to upstream AKA Mendeley itself have fallen on deaf ears multiple times and for several years now, even when supplied with the above solution.
But OK, if you're not interested in accepting this patch, it appears I'll have to start my own tap for this one lonely cask-formula :/

Issue #46382, opened 2018-04-26
'brew cask outdated' does not identify iterm2


dwmarshall:
#### General troubleshooting steps

- [x] I have retried my command with `--force` and the issue is still present.
- [x] I have checked the instructions for [reporting bugs](https://github.com/caskroom/homebrew-cask#reporting-bugs) (or [making requests](https://github.com/caskroom/homebrew-cask#requests)) before opening the issue.
  - [x] None of the templates was appropriate for my issue, or I’m not sure.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).

#### Description of issue

`brew cask outdated` does not report that iterm2 3.1.5 can be upgraded to 3.1.6

#### Command that failed

```
$ brew cask outdated
minikube (0.25.2) != 0.26.1
virtualbox (5.2.8,121009) != 5.2.10,122088
virtualbox-extension-pack (5.2.8,121009) != 5.2.10,122088
```

#### Output of command with `--force --verbose --debug`

```
$ brew cask outdated --verbose --debug
==> Determining which containers to use based on 'container :type'
==> Determining which containers to use based on 'container :type'
==> Checking update info of Cask atom
==> Checking update info of Cask chefdk
==> Checking update info of Cask docker
==> Checking update info of Cask iterm2
==> Checking update info of Cask minikube
minikube (0.25.2) != 0.26.1
==> Checking update info of Cask skitch
==> Checking update info of Cask virtualbox
virtualbox (5.2.8,121009) != 5.2.10,122088
==> Checking update info of Cask virtualbox-extension-pack
virtualbox-extension-pack (5.2.8,121009) != 5.2.10,122088
```

#### Output of `brew cask doctor`

```
$ brew cask doctor
==> Homebrew-Cask Version
Homebrew-Cask 1.6.2-32-gae6ae30
caskroom/homebrew-cask (git revision cdf45; last commit 2018-04-26)
==> macOS
10.13.3
==> SIP
Enabled
==> Java
1.7.0_75
==> Homebrew-Cask Install Location
<NONE>
==> Homebrew-Cask Staging Location
/usr/local/Caskroom
==> Homebrew-Cask Cached Downloads
[...]

Issue #38784, opened 2017-09-20
brew cask outdated doesn't report all outdated casks


jordigg:
#### General troubleshooting steps

- [x] I have checked the instructions for [reporting bugs](https://github.com/caskroom/homebrew-cask#reporting-bugs) (or [making requests](https://github.com/caskroom/homebrew-cask#requests)) before opening the issue.
- [x] None of the templates was appropriate for my issue, or I’m not sure.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).

#### Description of issue

`brew cask outdated` doesn't report all outdated casks. Using `--greedy` is possible to see that there is a newer version available for `iterm2` for example.

#### Output of your command with `--verbose --debug`

```
brew cask outdated --verbose --debug
==> Determining which containers to use based on 'container :type'
==> Checking update info of Cask atom
==> Checking update info of Cask google-backup-and-sync
==> Checking update info of Cask google-chrome
==> Checking update info of Cask google-drive
==> Checking update info of Cask gpgtools
==> Checking update info of Cask intellij-idea
==> Checking update info of Cask iterm2
==> Checking update info of Cask java
java (1.8.0_131-b11,d54c1d3a095b4ff2b6607d096fa80163) != 1.8.0_144-b01,090f390dda5b47b9b721c7dfaa008135
==> Checking update info of Cask jetbrains-toolbox
==> Checking update info of Cask launchcontrol
==> Checking update info of Cask microsoft-office
==> Checking update info of Cask slack
==> Checking update info of Cask spotify
==> Checking update info of Cask steam
==> Checking update info of Cask sublime-text
==> Checking update info of Cask tunnelblick
==> Checking update info of Cask vagrant
vagrant (1.9.5) != 2.0.0
==> Checking update info of Cask virtualbox
virtualbox (5.1.22-115126) != 5.1.28-117968
==> Checking update info of Cask virtualbox-extension-pack
virtualbox-extension-pack (5.1.22-115126) != 5.1.28-117968
```

```
brew cask outdated --greedy --verbose --debug
==> Determining which containers to use based on 'container :type'
==> Checking update info of Cask atom
atom (1.18.0) != 1.20.1
[...]

Issue #37294, opened 2017-08-01
brew cask outdated does not do what is described


jrwiegand:
#### General troubleshooting steps

- [x] I have checked the instructions for [reporting bugs](https://github.com/caskroom/homebrew-cask#reporting-bugs) (or [making requests](https://github.com/caskroom/homebrew-cask#requests)) before opening the issue.
  - [x] None of the templates was appropriate for my issue, or I’m not sure.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).

#### Description of issue

`brew cask outdated` does not work as described. From the help section of the client (`brew cask help`) "outdated - list the outdated installed Casks". Does this mean that it will display a list of software that is outdated? If so, then the current version is not doing that.

#### Output of your command with `--verbose --debug`
```
$ brew cask outdated --verbose --debug
==> Checking update info of Cask alfred
==> Checking update info of Cask docker
==> Checking update info of Cask docker-toolbox
==> Checking update info of Cask filezilla
==> Checking update info of Cask firefox-beta
==> Checking update info of Cask flycut
==> Checking update info of Cask gimp
==> Checking update info of Cask google-chrome-beta
==> Checking update info of Cask intellij-idea
==> Checking update info of Cask iterm2-beta
==> Checking update info of Cask itsycal
==> Checking update info of Cask java
==> Checking update info of Cask mattermost
==> Checking update info of Cask postage
==> Checking update info of Cask postgres
==> Checking update info of Cask postman
==> Checking update info of Cask sourcetree-beta
==> Checking update info of Cask the-unarchiver
==> Checking update info of Cask virtualbox-beta
==> Checking update info of Cask visual-studio-code

$ brew cask outdated
```

#### Output of `brew cask doctor`

```
$ brew cask doctor
==> Homebrew-Cask Version
Homebrew-Cask 1.3.0-15-g2ad03b8
caskroom/homebrew-cask (git revision 43e55; last commit 2017-08-02)
==> Homebrew-Cask Install Location
<NONE>
==> Homebrew-Cask Staging Location
/usr/local/Caskroom
==> Homebrew-Cask Cached Downloads
~/Library/Caches/Homebrew/Cask (28 files, 2.2GB)
[...]

PR #46417, opened 2018-04-26
note that some Casks don't show up as outdated

dwmarshall:
There are a number of issues (all quickly closed) about how one Cask or another wasn't listed as being outdated. When I reported such an issue, I didn't know that the --greedy switch exists.

This is a documentation-only change, in hopes of preventing issues such as #46382, #38784, and #37294 (to choose just a few)

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -128,7 +128,7 @@ Caffeine.app (app)
 
 ## Updating/Upgrading Casks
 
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):

commitay:
Thanks for the PR @dwmarshall!

PR #46449, opened 2018-04-27
Add docker-credential-gcloud binary to google-cloud-sdk

jpvelez:
Using Google Cloud Platform's Container Registry (GCR) to store docker images requires giving Docker access to GCR. There are a couple of authentication methods: https://cloud.google.com/container-registry/docs/advanced-authentication#using_the_standalone_docker_credential_helper

One way is to use `gcloud docker ` instead of `docker`. However, this method is now deprecated for Docker client versions above 18.03.

The recommended way authentication now is to configure docker to use a standalone docker credential helper, `docker-credential-gcloud`, a binary that ships with new versions of google-cloud-sdk. To do so, you use the following command:

```
$ gcloud auth configure-docker 
WARNING: `docker-credential-gcloud` not in system PATH.
gcloud's Docker credential helper can be configured but it will not work until this is corrected.
Docker configuration file updated.
```

This gcloud auth command uses the `docker-credential-gcloud` command to configure docker. It works fine if you install google-cloud-sdk the normal way, but doesn't work if you install it via brew cask. The reason is that even though cask actually installs the binary to `/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin`, it doesn't symlink it to `usr/local/bin`, as documented here: https://stackoverflow.com/questions/49780218/docker-credential-gcloud-not-in-system-path. This fixes that.

### DIFF
--- a/Casks/google-cloud-sdk.rb
+++ b/Casks/google-cloud-sdk.rb
@@ -14,6 +14,7 @@ cask 'google-cloud-sdk' do
   binary 'google-cloud-sdk/bin/gcloud'
   binary 'google-cloud-sdk/bin/git-credential-gcloud.sh', target: 'git-credential-gcloud'
   binary 'google-cloud-sdk/bin/gsutil'
+  binary 'google-cloud-sdk/bin/docker-credential-gcloud'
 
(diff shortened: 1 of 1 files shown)


commitay:
Thanks for the PR!

PR #46489, opened 2018-04-29
Update clip-studio-paint to 1.7.4

zthxxx:


zthxxx:
emmmH~ 
thanks, i have find the space in file name now.

PR #46517, opened 2018-04-29
Update lulu to 0.9.4

commitay:

### DIFF
--- a/Casks/lulu.rb
+++ b/Casks/lulu.rb
@@ -1,11 +1,11 @@
 cask 'lulu' do
-  version '0.9.3'
-  sha256 '6309a469e3620cd4a3934ae0e417be7d8d3f9b6fe9f363561c01235ae57c9486'
+  version '0.9.4'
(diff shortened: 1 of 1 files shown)


objective-see:
Aloha!

One point to note - that's new to 0.9.4

After `LuLu Installer.app` runs the config script, it now also invokes a method to add the login item (see:  https://github.com/objective-see/LuLu/blob/790cd977bc13d28846faa0ec41d9299a14e71767/configure/Configure/Configure.m#L343) 

AFAIK, there wasn't a way to do this via script/shell commands - so this had to be done in the objective-c code. Before the script invoked the main binary to do this, but had to do this via the 'open' command to ensure it ran in the context of the logged in user (vs. as root)...this method spawned a separate process, dock icon, etc...so it was simpler just to move this logic into the `LuLu Installer.app`. However, this means that now executing `configure.sh` isn't enough to install LuLu fully. 

Would it help if I created a cmdline option that could be passed to the `LuLu Installer.app` that just took care of everything? (i.e. ran the script, installed the login item, etc etc)?

PR #46520, opened 2018-04-29
script.sh: uninstall instead of zap

commitay:
Reverts https://github.com/caskroom/homebrew-cask/pull/44407

Running `zap` makes the leftover check less useful in some cases because it [explodes](https://github.com/caskroom/homebrew-cask/issues/44993) trying to remove files it doesn't have permissions to and doesn't run the check.


reitermarkus:
Also, `uninstall` should usually be enough to remove everything `install` installs, except for the special case which prompted https://github.com/caskroom/homebrew-cask/pull/44407.

reitermarkus (approved):

PR #46615, opened 2018-05-02
Update adobe-acrobat-pro to 2018

core-code:

### DIFF
--- a/Casks/adobe-acrobat-pro.rb
+++ b/Casks/adobe-acrobat-pro.rb
@@ -1,8 +1,8 @@
 cask 'adobe-acrobat-pro' do
-  version '2017'
+  version '2018'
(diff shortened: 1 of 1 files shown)


commitay:
There are significant differences between these files, the new one seems to be a "update" package that's half the size of the current download.

Also, this was updated [September 2017](https://github.com/caskroom/homebrew-cask/pull/38688) from `2015` to `2017` and inspecting the current download shows it was modified Feb 2018.

Thanks for the PR nonetheless!

core-code:
i found the real, full download URL for Adobe Acrobat 2018 (DC) now:

https://trials3.adobe.com/AdobeProducts/APRO/18/osx10/Acrobat_DC_Web_WWMUI.dmg

(the won't work in browser because of the cookie requirement)

should i submit another pull request?

PR #46687, opened 2018-05-04
probot: lock closed issues/pull requests

commitay:
https://probot.github.io/apps/lock/
https://github.com/dessant/lock-threads

> only 30 issues and pull requests will be locked per hour. If your repository has more than that, it will just take a few hours or days to lock them all.

@vitorgalvao I don't have permissions to enable this for Caskroom but I've sent a "request", which seems to be a new GitHub feature.


commitay:
If we merge this, we can run this https://github.com/MikeMcQuaid/LockOldIssues as well so it doesn't take the bot a month and half to catch up.

PR #46749, opened 2018-05-05
Update IDA free 7.0 SHA

fletchto99:
The sha provided is actually the SHA-1 and not SHA256 thus the checksum check fails. I've updated it to use the SHA256

Verification: https://www.virustotal.com/#/file/7cbcfbcbff4154e358428987e8511196cc84a1562de4406ba13c6df5d2230bcb/detection

### DIFF
--- a/Casks/idafree.rb
+++ b/Casks/idafree.rb
@@ -1,6 +1,6 @@
 cask 'idafree' do
   version '7.0'
-  sha256 'aaf962bf02134f147cba56fe6b2c146549f5b270512e1406a4004f417e2799ac'
+  sha256 '7cbcfbcbff4154e358428987e8511196cc84a1562de4406ba13c6df5d2230bcb'
 
   url "https://out7.hex-rays.com/files/idafree#{version.no_dots}_mac.tgz"
   name 'IDA Free'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> The sha provided is actually the SHA-1 and not SHA256

That is incorrect. SHA-1 is shorter (40 characters long) and if the shasum were wrong at the moment of submission, Travis would have caught it.

This represents a file changed in place. Seeing as there isn’t a signed app we can check, the VirusTotal check does us no good. You need to follow the other procedure and contact the developer in a way they publicly confirm this shasum is correct.

vitorgalvao (changes requested):
Awaiting shasum verification

fletchto99:
Ahh true it makes sense that the SHA1 wouldn't have been used in the first place 😓. The file was changed in place hence the change in the shasum. 

The developer publishes the SHA1: https://www.hex-rays.com/products/ida/support/download_freeware.shtml

Is that enough to verify that the file is correct?

fletchto99:
Sorry for the delayed response! I got busy and totally forgot to message the developer. Anyhow I've got confirmation now.

screen shot 2018-06-02 at 4 56 45 pm

screen shot 2018-06-02 at 4 57 42 pm

fletchto99:
Got an update as per why the SHA changed:

capture

fletchto99:
@vitorgalvao Is there anything else I must do before this PR is good to go?

vitorgalvao:
Thank you, @fletchto99. I see an inconsistency between the published version and their shasum on the page, so give me a bit to confirm that.

PR #46819, opened 2018-05-08
Fix Casks using `gpg`.

reitermarkus:

### DIFF
--- a/Casks/bitpay.rb
+++ b/Casks/bitpay.rb
@@ -1,11 +1,11 @@
 cask 'bitpay' do
-  version '3.15.2'
-  sha256 '793961e1b1e0a2cdf1b9a770191399531cff5c5388e38461ffc72b0f1dc5c7d4'
+  version '4.3.2'
+  sha256 'afdcc1556ee3294e10045324907075e98bb831791f4503b6a9ee4f171c973185'
 
   # github.com/bitpay/copay was verified as official when first introduced to the cask
   url "https://github.com/bitpay/copay/releases/download/v#{version}/BitPay.dmg"
   appcast 'https://github.com/bitpay/copay/releases.atom',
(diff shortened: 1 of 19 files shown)


commitay:
Mixing version bumps in with other changes isn't great as the CI install/uninstall step isn't run.

reitermarkus:
> the CI install/uninstall step isn't run.

Why not? 

> `>>> More than 3 Casks modified, skipping install`

We should just let CI fail then, instead of skipping it.

commitay:
CI builds that try to install all the Casks modified in a PR will run until they timeout which impacts the whole organisation queue.

reitermarkus:
Then maybe let's fail if there is more than one Cask updated. 3 is an arbitrary number, updating `mactex` and `mactex-no-gui` at the same time would probably time out too, despite being only two.

commitay:
> Then maybe let's fail if there is more than one Cask updated.

Why? Mass edits shouldn't fail, we just don't want to run the install.

reitermarkus:
But if mass edits don't fail, how am I supposed to know I should split them up in *this* case?

Basically, if you argue that “mixing version bumps in with other changes isn't great”, I argue that CI should fail and suggest to split them up.

commitay:
Are you proposing:

1) CI fails if more than one Cask is modified?

2) CI fails if more than one Cask is modified _and_ one of those Casks is version bumped?

reitermarkus:
For now, I would suggest 1, since for 2, we would basically have to rewrite the Travis script in Ruby in order to properly check if the version (or rather url, since that is what the installation really depends on) has changed.

reitermarkus:
> I'm not sure I understand how 1 would would help in _this_ case

It would fail, hence I would have to split it up, hence the installation would be tested.

PR #46862, opened 2018-05-08
Remove winds

gtmgianni:


Casks/winds.rb
```diff
@@ -1,11 +1,11 @@
 cask 'winds' do
-  version '0.3.0'
-  sha256 'f3d97376a9f060c6d4c1b3f32383332cfa4a96c294aed760306111759adc292b'
+  version '2.0.74'
+  sha256 'dd11b1a6b019c9437a8bf3597faafb9c77c6b7d67e95201205e72647e6a5256f'
 
-  # github.com/GetStream/Winds was verified as official when first introduced to the cask
-  url "https://github.com/GetStream/Winds/releases/download/v#{version}/Winds.zip"
+  # s3.amazonaws.com/winds was verified as official when first introduced to the cask
+  url "https://s3.amazonaws.com/winds-#{version.major_minor}-releases/releases/Winds-#{version}-mac.zip"
```
vitorgalvao: We can’t use this link, because it’s walled.

vitorgalvao (changes requested):

Casks/winds.rb
gtmgianni: Does it count as walled if you don't need to register "fully" (you just need to enter an email address, and the download begins)? If so, we should probably just remove this cask.

vitorgalvao: > Does it count as walled if you don't need to register "fully"

Yes. Even if it works with a fake email, it still counts as walled.

The point is to make every cask easily auditable by anyone, without having to resort to “tricks”.

commitay:
Thanks!

commitay:
Re-added in https://github.com/Homebrew/homebrew-cask/pull/48699 with less wall.

PR #46868, opened 2018-05-08
Update openconnect-gui to 1.5.3

gtmgianni:


gtmgianni:
Not sure if this pull request should be merged – according the the developer's Releases page, the latest version `1.5.3` has no `.dmg` file to download, but it does have a Homebrew `bottle`, which contains the application file.

Since this is a GUI application, should it stay in Homebrew-Cask, despite the developer trying to create a Homebrew formula out of it?

Also note that the developer has said that this package is experimental, and is for testing out the Homebrew installation.

commitay:
This isn't meant for us to install, it's for use with upstreams own tap. https://github.com/horar/homebrew-openconnect-gui/blob/develop/Formula/openconnect-gui.rb

We might need to remove this at some point if they don't publish an new download we can use.

PR #46917, opened 2018-05-10
dynamodb-local (latest) as a service

ikonst:
Installs dynamodb-local as a launchd service.

### DIFF
--- a/Casks/dynamodb-local.rb
+++ b/Casks/dynamodb-local.rb
@@ -19,6 +19,51 @@ cask 'dynamodb-local' do
     EOS
   end
 
+  launchd_plist = "#{staged_path}/dynamodb-local.plist"
+
+  postflight do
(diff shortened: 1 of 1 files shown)


commitay:
We don't support adding plists to Casks.

Thanks for the PR nonetheless!

ikonst:
@commitay What would be a good way forward to ensure dynamodb-local can be installed on equal grounds with e.g. the mongodb formula? My goal is to create a smooth local development experience.

commitay:
> What would be a good way forward to ensure dynamodb-local can be installed as conveniently as e.g. the mongodb formula?

Short answer: There isn't one.

dynamodb was a `core` formulae (where it had proper plist service support) but it was removed because it is closed-source and has an unversioned download.

The only other Cask I can think of that would benefit from a plist service is in the process of being migrated to `core` so it isn't something we're considering adding support for.

PR #47028, opened 2018-05-12
Update wowza-streaming-engine to 4.7.5

gtmgianni:


vitorgalvao:
Tried in a VM but can’t finish installation without a license key. Pinging people that previously updated this cask. Keep in mind that if we cannot devise a proper download strategy, we’ll have to remove the cask.

Pinging @yurikoles @colindunn @reitermarkus @mandrean

adidalal:
Did a bit of poking around. It uses installbuilder, which has no easy way to unpack it. (ref: https://support.bitrock.com/hc/en-us/community/posts/115002748989-Unpack-installer-package)

adidalal:
Think the issue is we're not sure if the uninstall .app has changed its format.
However, it seems unlikely, based on its history.

vitorgalvao:
Let’s merge and wait for bug reports if applicable.

PR #47267, opened 2018-05-15
Update bitcoin-unlimited to 1.0.3.0

core-code:

### DIFF
--- a/Casks/bitcoin-unlimited.rb
+++ b/Casks/bitcoin-unlimited.rb
@@ -1,6 +1,6 @@
 cask 'bitcoin-unlimited' do
-  version '1.0.2.0'
-  sha256 'e0360d108cb4f3ea996e5d59c41e95065e793c140bc12042e217d037bb587d63'
+  version '1.0.3.0'
+  sha256 'b9f5e275450632fe595c9cd776694ffa2fc049071ee7d0ffe14b899c95255959'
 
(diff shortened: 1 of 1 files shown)


Casks/bitcoin-unlimited.rb
```diff
@@ -1,8 +1,9 @@
 cask 'bitcoin-unlimited' do
   version '1.0.2.0'
-  sha256 'e0360d108cb4f3ea996e5d59c41e95065e793c140bc12042e217d037bb587d63'
+  sha256 '41c977acb2dfa4d44c2dc8a11a7585330036af274b038ff0d5b592399491cf1d'
```
commitay: The checksum shouldn't change if this is a mirrored copy.

core-code: i was wondering about that as well, but the official download ( https://www.bitcoinunlimited.info/download ) now links there

core-code:
ok but we have the same checksum mismatch as the original host.

the checksum failure on install is why i sent the pull request

commitay:
Please update this to the current version `1.3.0.1` and keep the bitcoinunlimited.info URL.

core-code:
i'm a bit hesitant to submit an pull request to 
• the newest version BUcash 1.3.0 because i don't know if its the same thing and also
• to the 1.0.3 version you mentioned, because while there seems to be a valid download there, its not mentioned at all on their download website

commitay:
I posted a link to a different version by accident, the URL and version currently used in this PR is correct.

PR #47289, opened 2018-05-15
Update airtool to 1.6.1

core-code:

### DIFF
--- a/Casks/airtool.rb
+++ b/Casks/airtool.rb
@@ -10,15 +10,17 @@ cask 'airtool' do
 
   pkg "airtool_#{version}.pkg"
 
+  uninstall_preflight do
+    set_ownership '/Library/Application Support/Airtool'
+  end
+
   uninstall pkgutil:    [
                           'com.adriangranados.airtool.airtool-bpf.*',
                           'com.adriangranados.airtool.Airtool.pkg',
                         ],
             launchctl:  'com.adriangranados.airtool.airtool-bpf.pkg',
-            login_item: 'Airtool'
+            login_item: 'Airtool',
+            delete:     '/Library/Application Support/Airtool'
 
-  zap trash: [
-               '/Library/Application Support/Airtool',
-               '~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.adriangranados.airtool.sfl*',
-             ]
+  zap trash: '~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.adriangranados.airtool.sfl*'
 end
--- a/Casks/airtool.rb
+++ b/Casks/airtool.rb
@@ -18,7 +18,7 @@ cask 'airtool' do
                           'com.adriangranados.airtool.airtool-bpf.*',
                           'com.adriangranados.airtool.Airtool.pkg',
                         ],
-            launchctl:  'com.adriangranados.airtool.airtool-bpf.pkg',
+            launchctl:  'com.adriangranados.airtool.airtool-bpf',
             login_item: 'Airtool',
             delete:     '/Library/Application Support/Airtool'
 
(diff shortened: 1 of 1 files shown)


Casks/airtool.rb
```diff
@@ -1,25 +1,26 @@
...
-  sha256 'ba7db87984d2f18f2d46cdbb0b75fdfdda920d3f43dfa3cf18aff1866bac0701'
+  version '1.6.1'
+  sha256 '68fa07f9c3d452a34f505f454640e3920f919236b6017dda0a16d5f486f41b2d'
 
-  # amazonaws.com/apps.adriangranados.com was verified as official when first introduced to the cask
-  url "https://s3.amazonaws.com/apps.adriangranados.com/airtool_#{version}.pkg"
-  appcast 'https://www.adriangranados.com/appcasts/airtoolcast.xml',
-          checkpoint: 'e2801e382b08cc44b507979619bfef3c1bb019d803a269148564e8366426d583'
+  url "https://www.adriangranados.com/downloads/airtool_#{version}.pkg"
+  appcast 'https://updates.devmate.com/com.adriangranados.Airtool.xml',
```
commitay: How did you get this appcast?

core-code: many apps where the sparkle appcast url (SUFeedURL) is not in the info.plist have their feeds at https://updates.devmate.com/BUNDLEID.xml  , especially if they have the DevMateKit framework instead of the Sparkle framework in their embedded frameworks folder. in any case, you can find many sparkle appscast URLs by using a proxy like 'Charles'

if you are looking for 'hidden' sparkle appcast URLs for popular apps, i have a list of > 100 i can share if needed.

commitay: I meant this appcast specifically?

core-code: as mentioned the newest version doesn't have the the sparkle framework but the DevMateKit framework, so i suspected the feed must be there (its the same for nearly all devmate apps), and running the charles proxy (need to enable SSL proxying) while doing 'check for updates' in the app confirmed the actual feed location.

vitorgalvao: > many apps where the sparkle appcast url (SUFeedURL) is not in the info.plist have their feeds at https://updates.devmate.com/BUNDLEID.xml , especially if they have the DevMateKit framework instead of the Sparkle framework in their embedded frameworks folder.

For the Sparkle appcast, we can easily query `Info.plist` for the `SUFeedURL`. Is there a similar easy method for the DevMateKit framework, rather than blindly trying to to access `https://updates.devmate.com/BUNDLEID.xml` or going through the trouble of intercepting connections?

> if you are looking for 'hidden' sparkle appcast URLs for popular apps

We are. You can add them to casks directly or open an issue where you list all of them if that’s more convenient. If you go the list route, we’d prefer a checklist format so we can check them as we add them, but whatever works for you. Thank you.

core-code: > similar easy method 
not that i know of. but i find "blindly trying" quite easy ;)

note that apart of DevMate apps, there are also quite many Hockeyapp apps which also have hidden SUFeedURLs. those are always at 
https://rink.hockeyapp.net/api/2/apps/APPIDENTIFIER
you can find the APPIDENTIFIER by running:

strings binary | grep ^[a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9][a-z0-9]$

> We are.

i've opened issue #47351 with the list of URLs i have

vitorgalvao: Thank you for the tips. Here’s one in return. Since what you’re doing in `grep` is looking for an hexadecimal string of 32 characters, you can simplify the code: `strings binary | grep --extended-regexp '^[a-f0-9]{32}$'`. `a-f` because hexadecimal strings don’t have letters beyond that; `{32}` to look for the pattern that repeats itself exactly 32 times, and `--extended-regexp` (or the shorter flag `-E`) because without it `{32}` won’t work.

core-code: nice, thanks ;)

PR #47365, opened 2018-05-18
Update expandrive to 6.1.12

core-code:


core-code:
even when specifying the .dmg extension, actually a zip file gets downloaded. so lying doesn't make much sense

PR #47404, opened 2018-05-19
Update privatevpn to 2.3.5

core-code:

### DIFF
--- a/Casks/privatevpn.rb
+++ b/Casks/privatevpn.rb
@@ -2,7 +2,8 @@ cask 'privatevpn' do
   version '2.3.5'
   sha256 '98d55627a704767a24502112491415be70c1e8800b245ea5cd022cb29f7b099d'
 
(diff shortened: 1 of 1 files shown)


Casks/privatevpn.rb
```diff
@@ -2,7 +2,8 @@
   version '2.3.5'
   sha256 '98d55627a704767a24502112491415be70c1e8800b245ea5cd022cb29f7b099d'
 
-  url 'https://privatevpn.com/client/PrivateVPN.dmg'
```
commitay: The homepage still has this as the URL?

vitorgalvao: I’m guessing the replacement was taken from the appcast.

Thank you, @core-code, but while the appcast is useful for knowing when new versions are out, we’d rather have download links that doesn’t need the “verified as official” comment.

vitorgalvao: It’s Cloudflare’s DDoS protection preventing the download.

PR #47442, opened 2018-05-21
New find_appcast script

vitorgalvao:
Restructured script and included new methods, following the tips by @core-code in https://github.com/caskroom/homebrew-cask/pull/47289.


developer/bin/find_appcast
```diff
@@ -0,0 +1,113 @@
...
+
+# exit if no argument (or more than one) was given
+if [[ -z "${1}" ]] || [[ -n "${2}" ]]; then
+  usage
+  exit 1
+fi
+
+readonly path_to_app="$(absolute_path "${1}")"
+
+if [[ ! -d "${path_to_app}" ]] || [[ "${path_to_app}" != *'.app' ]]; then
```
vitorgalvao: This was previously `if ! mdls "${path_to_app}" | grep --quiet 'com.apple.application-bundle'; then`, but that won’t work the bundle isn’t yet indexed by Spotlight, which is likely if you’ve just downloaded it.

core-code:
very cool. small bug...some apps are 'hockeyapps' and have the hockey-sparkle-feed, but choose to actually not use it, and use another feed instead.

e.g. for Jump Desktop Connect ( https://jumpdesktop.com/automatic-setup.html ) the script finds the hockey feed:

An appcast was found. You should add it to your cask as:

      appcast 'https://rink.hockeyapp.net/api/2/apps/4cefa4642fd4c0a2029ef2f4d83dfd37',

but this one is actually empty - it has no single entry - and thus should not be returned as a result. 

(the actual feed Jump Desktop Connect uses is https://s3.amazonaws.com/jumpdesktop.com/downloads/connect/connect-mac.xml )

vitorgalvao:
* Updated script to ignore empty DevMate appcasts.
* Removed URL checks (how useful were they? Even a cask with a 302 might be useful).
* Check for all appcast types instead of stopping on the first. This together with the previous point should encourage looking for the best one.

commitay (approved):
Nice!

core-code:
for some reason the new script doesn't find the hockey appcast
from this app:
https://fiplab.com/apps/infinity-dashboard-for-mac
=>
https://rink.hockeyapp.net/api/2/apps/a46bae8fd523424580942ac903f4561d

vitorgalvao:
That one is an outlier. It finds *two* matching strings:

```
59519fbfa4311e9a2dcbc8796c75992f
a46bae8fd523424580942ac903f4561d
```

And the script is only prepared for one.

PR #47465, opened 2018-05-22
Update synergy to 2.0.12,b1807-50472cde

moritzreiter:

### DIFF
--- a/Casks/synergy.rb
+++ b/Casks/synergy.rb
@@ -1,8 +1,8 @@
 cask 'synergy' do
-  version '2.0.11,b1783-8c94f00f'
-  sha256 'af13c4b26d00c4cd87173af7196b5d713a09e9c2818fa1538264f6f692a87355'
+  version '2.0.12,b1807-50472cde'
+  sha256 '41927a6a52a523797122a7d6a9c8817d09cce9966ad08118c693345f51aa41e9'
 
-  url "https://binaries.symless.com/v#{version.before_comma}/Synergy_v#{version.before_comma}-stable_#{version.after_comma}.dmg"
+  url "https://binaries.symless.com/v#{version.before_comma}/Synergy_v#{version.before_comma}-beta_#{version.after_comma}.dmg"
   name 'Synergy'
   homepage 'https://symless.com/synergy'
 


Casks/synergy.rb
```diff
@@ -1,8 +1,8 @@
 cask 'synergy' do
-  version '2.0.11,b1783-8c94f00f'
-  sha256 'af13c4b26d00c4cd87173af7196b5d713a09e9c2818fa1538264f6f692a87355'
+  version '2.0.12,b1807-50472cde'
+  sha256 '41927a6a52a523797122a7d6a9c8817d09cce9966ad08118c693345f51aa41e9'
 
-  url "https://binaries.symless.com/v#{version.before_comma}/Synergy_v#{version.before_comma}-stable_#{version.after_comma}.dmg"
+  url "https://binaries.symless.com/v#{version.before_comma}/Synergy_v#{version.before_comma}-beta_#{version.after_comma}.dmg"
```
commitay: Why is this a beta version? The Cask should track stable versions if they are available.

anothernode: [Symless decided to label the whole 2.x branch as beta now](https://symless.com/forums/topic/5725-2012-beta/), presumably because they keep having some hard to fix connection problems in the 2.x version.

So, in order to stick with the rules, the cask would have to be downgraded to the 1.x branch, which is the only official stable version now.

Me personally, I would plead against that and instead make an exception to the rule here as for me the 2.x branch works good enough (although not perfect).

vitorgalvao: It’s not an exception. [The rule](https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version) covers this (emphasis added):

> When an App is only available as beta, development, or unstable versions, **or in cases where such a version is the general standard**, then said version can go into the main repo.

ndobbs:
From what I understand; the beta version is lacking the paid 'Pro' features. Staying on the v1 line of code makes sense for paying customers at least until v2 becomes stable and the pro feature upgrade becomes available.

screen shot 2018-07-27 at 9 39 48 am

PR #47471, opened 2018-05-22
Update 1password to 7.0

core-code:

### DIFF
--- a/Casks/1password.rb
+++ b/Casks/1password.rb
@@ -40,7 +40,7 @@ cask '1password' do
 
   auto_updates true
 
-  uninstall pkgutil: '1Password*'
+  uninstall pkgutil: 'com.agilebits.pkg.onepassword'
 
   zap trash: [
                '~/Library/Application Scripts/2BUA8C4S2C.com.agilebits.onepassword-osx-helper',
--- a/Casks/1password.rb
+++ b/Casks/1password.rb
@@ -2,8 +2,8 @@ cask '1password' do
   version '7.0'
   sha256 '9c79de6b4ff9dadc7476cd140156087bd9ad12d66de0b7b05dbcb0a2f660d7ec'
 
-  # cache.agilebits.com/dist/1P was verified as official when first introduced to the cask
-  url "https://cache.agilebits.com/dist/1P/mac#{version.major}/1Password-#{version}.zip"
+  # 1password.com was verified as official when first introduced to the cask
+  url "https://c.1password.com/dist/1P/mac#{version.major}/1Password-#{version}.zip"
   appcast "https://app-updates.agilebits.com/product_history/OPM#{version.major}",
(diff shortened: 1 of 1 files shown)


Casks/1password.rb
```diff
@@ -15,23 +15,33 @@
     url "https://cache.agilebits.com/dist/1P/mac4/1Password-#{version}.zip"
 
     app "1Password #{version.major}.app"
-  else
+  elsif MacOS.version <= :sierra
```
vitorgalvao: This implies this version only works on Sierra and below, which I don’t think is the case.

cpmsmith: According to AgileBits, [v7 supports 10.12 onward](https://support.1password.com/upgrade-mac/#get-help), so this should say El Capitan.

brianmorton:
https://github.com/caskroom/homebrew-cask/pull/47470

Casks/1password.rb
core-code: > This implies this version only works on Sierra and below, which I don’t think is the case.
 
i think you mis-read the diff. version 7 of 1password only runs on 10.12+ so older systems still have to install the old version 6

> This needs to be part of only the conditional of version 7.

fixed thanks

commitay (changes requested):
Version 7 is a paid upgrade, I think this Cask should contain version 7 only and 6 should be added to versions.

commitay:
Also, there is a zip for version 7.

https://cache.agilebits.com/dist/1P/mac7/1Password-7.0.zip

core-code:
the zip file wasn't there when i checked earlier today. 

i can change the cask to contain only version 7 if you want.

commitay (approved):

Casks/1password.rb
```diff
@@ -1,36 +1,18 @@
...
-
-    # cache.agilebits.com/dist/1P/mac4 was verified as official when first introduced to the cask
-    url "https://cache.agilebits.com/dist/1P/mac4/1Password-#{version}.zip"
-
-    app "1Password #{version.major}.app"
-  end
-
-  appcast 'https://app-updates.agilebits.com/product_history/OPM4',
-          checkpoint: 'eb8b60476f30a53f496a84e809b948faf81ea9f9ab442f4c923e48720e5a1f2a'
+  # 1password.com was verified as official when first introduced to the cask
```
commitay: The comment is required, same as https://github.com/caskroom/homebrew-versions/pull/5451

commitay:
PR for `1password6` https://github.com/caskroom/homebrew-versions/pull/5720

karlbright (approved):

vitorgalvao:
> the zip file wasn't there when i checked earlier today.

Can confirm. I also checked and thought they had given rid of it for good.

PR #47599, opened 2018-05-25
Rollback Gitter link to caskroom/homebrew-cask

JBallin:
The updated link (Homebrew/homebrew-cask) responds with a 404 error but the old link (caskroom/homebrew-cask) still works. Even if planning to update link on Gitter, the README link should be updated temporarily.


reitermarkus:
Thanks!

PR #47600, opened 2018-05-25
Add puppet-bolt 0.2.3

lucywyman:

- [ ] Named the cask according to the [token reference]. I'm not entirely sure what this is?

### DIFF
--- a/Casks/puppet-bolt.rb
+++ b/Casks/puppet-bolt.rb
@@ -1,0 +1,22 @@
+cask 'puppet-bolt' do
+  version '0.20.3'
+
+  if MacOS.version == :el_capitan
+    sha256 'd4cbc0a9c824cfa05dd540f7eaa5c036484bd651613a60ec7a929a22fb099441'
+  elsif MacOS.version == :sierra
+    sha256 'ca2407e8b258705d998b6d3143dcfb6d208a5453b02b09ab945ce5b21dd93d00'
+  else
+    sha256 'd623713f352a318fcdbc105a03c87faa055190819b171f1604e5e147579be21a'
+  end
+
+  # downloads.puppetlabs.com/mac was verified as official when first introduced to the cask
+  url "http://downloads.puppetlabs.com/mac/#{MacOS.version}/PC1/x86_64/puppet-bolt-#{version}-1.osx#{MacOS.version}.dmg"
+  name 'Puppet Bolt'
+  homepage 'https://github.com/puppetlabs/bolt'
+
+  depends_on macos: '>= 10.11'
+
+  pkg "puppet-bolt-#{version}-1-installer.pkg"
+
+  uninstall pkgutil: 'com.puppetlabs.puppet-bolt'
+end
--- a/Casks/puppet-bolt.rb
+++ b/Casks/puppet-bolt.rb
@@ -14,7 +14,7 @@ cask 'puppet-bolt' do
   name 'Puppet Bolt'
   homepage 'https://github.com/puppetlabs/bolt'
 
-  depends_on macos: '>= 10.11'
+  depends_on macos: '>= :el_capitan'
 
   pkg "puppet-bolt-#{version}-1-installer.pkg"
 


Casks/puppet-bolt.rb
```diff
@@ -0,0 +1,23 @@
+cask 'puppet-bolt' do
+  version '0.20.3'
+
+  case MacOS.version
```
victorpopkov: First of all, thank you for your contribution. We really do appreciate this!

Just for consistency reasons, please change `MacOS.version` case block to:
```ruby
  if MacOS.version == :el_capitan
    sha256 'd4cbc0a9c824cfa05dd540f7eaa5c036484bd651613a60ec7a929a22fb099441'
  elsif MacOS.version == :sierra
    sha256 'ca2407e8b258705d998b6d3143dcfb6d208a5453b02b09ab945ce5b21dd93d00'
  else
    sha256 'd623713f352a318fcdbc105a03c87faa055190819b171f1604e5e147579be21a'
  end
```

You can learn more here:
https://github.com/Homebrew/homebrew-cask/tree/master/doc/cask_language_reference#conditional-statements

lucywyman:
Fixed! Thanks for the quick feedback :)

Casks/puppet-bolt.rb
```diff
@@ -0,0 +1,22 @@
...
+  if MacOS.version == :el_capitan
+    sha256 'd4cbc0a9c824cfa05dd540f7eaa5c036484bd651613a60ec7a929a22fb099441'
+  elsif MacOS.version == :sierra
+    sha256 'ca2407e8b258705d998b6d3143dcfb6d208a5453b02b09ab945ce5b21dd93d00'
+  else
+    sha256 'd623713f352a318fcdbc105a03c87faa055190819b171f1604e5e147579be21a'
+  end
+
+  # downloads.puppetlabs.com/mac was verified as official when first introduced to the cask
+  url "http://downloads.puppetlabs.com/mac/#{MacOS.version}/PC1/x86_64/puppet-bolt-#{version}-1.osx#{MacOS.version}.dmg"
```
commitay: `#{MacOS.version}` will break with the next macOS release.

Please change this to hardcode the version in the URL similar to 
https://github.com/Homebrew/homebrew-cask/blob/master/Casks/pdk.rb

commitay:
https://github.com/puppetlabs/bolt

@vitorgalvao This is open-source and CLI-only (so is `pdk`, which I merged without noticing) should we accept this?

vitorgalvao:
> This is open-source and CLI-only (so is pdk, which I merged without noticing) should we accept this?

Nope. We should also remove `pdk`, and follow the usual process.

lucywyman:
Ok, I will put up a PR to homebrew-core! Thank you.

vitorgalvao:
> Is the "CLI-only" criteria a reference to the app itself or its means of installation? 

The app itself.

lucywyman:
@vitorgalvao In reading https://docs.brew.sh/Acceptable-Formulae it looks like they also don't want binaries to be in homebrew-core?  We'll put up a PR to core that installs the .dmg, but we're just not sure what "the right way" to distribute our package is.

lucywyman:
Opened https://github.com/Homebrew/homebrew-core/pull/28439

lucywyman:
We don't want to build our package from source, for 2 main reasons:

1. The puppet-bolt `.pkg` includes Gem dependencies and ruby 2.3 (since most Linux OSes don't have a ruby version that high by default), which we don't want to ask users to install on their own. That workflow already exists, and is fine, but we want to make installation as easy as possible and have already gone to the trouble of packaging everything together. 
2. We also don't want to have to maintain our Formula alongside our package as dependencies change, especially since our CI pipeline is already packaging Bolt with it's dependencies so requires no extra effort.  It's likely that we would, at some point, forget to update the Formula with a critical change, and would end up with out-dated or broken installations for our OSX users.

Do those seem unreasonable?

commitay:
Until a formula source build has been judged by maintainers as being unsuitable to add and maintain, it isn't going to be accepted as a cask. 

Upstream wishes as to how this is distributed aren't really a factor.

Unless https://github.com/Homebrew/homebrew-core/pull/28439 is going to be updated with a source build, I suggest closing both PRs and [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

Note that creating a tap will not preclude it from being accepted as a formula if it is submitted by someone else.

Issue #41990, opened 2017-12-15
Consider moving into Homebrew GitHub organisation


MikeMcQuaid:
- [x] None of the templates was appropriate for my issue, or I’m not sure.

I think the Homebrew/brew merge of `brew cask` code has been a massive success. I'd like to propose that you consider moving caskroom/homebrew-cask into Homebrew/homebrew-cask. I think this would help unify the projects and communities a little bit more and generally strengthen both projects.

I think it would be cool to also consider moving the other taps over, too. The only problem is a naming collision with caskroom/homebrew-versions and Homebrew/homebrew-versions but I'd also propose considering the addition of the versioned casks into caskroom/homebrew-cask (this would also be consistent with homebrew/homebrew-core).

Thoughts? Concerns? Feelings?

wickles:
I would just like to point out the issue that there is still some duplication/overlap between homebrew formulae and casks (see https://github.com/caskroom/homebrew-cask/issues/15603). In particular: emacs, macvim, wireshark, and mono/mono-mdk. I would also add ppsspp, which ostensibly should be a cask not a homebrew formula, although I realize that isn't directly possible because they don't seem to provide official stable builds for some reason (however, official latest/working dev builds are provided at https://build.ppsspp.org/?page/downloads#osx).

commitay:
@wickles Most are intentional duplicates. From the issue you linked: https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-271920038

`mono` / `mono-mdk` are very different in how and what they install, the `mono` formula would require extensive changes to replace `mono-mdk` and we would still need to somehow provide compatibility for Casks that require mono and expect it to reside in `/Library/Frameworks`.

The purpose of this issue at this stage is to discuss _if we should move_, please keep comments on topic or it will be locked.

reitermarkus:
I‘m definitely in favour of this. Just not sure what would be the best place to start with. Probably allowing versioned Casks in `caskroom/versions`, i.e. allowing `@` in tokens.

commitay:
I don't really have an opinion either way with regards to moving.

Personally I would oppose splitting up the taps between different organisations, so I think if we do move, we should move all of the taps.

Moving Casks from `versions` into the main tap would be something I am against, I find to easier to manage and maintain "versions" when they are separated from the main repo.

[...]

PR #47609, opened 2018-05-26
Migrate java8

commitay:
I [suggested](https://github.com/Homebrew/homebrew-cask/issues/41990#issuecomment-389100136) adding JDK8 to this repo as a compromise rather than adding all of the casks in versions.

Opened this to discuss it separately rather than in the migration issue.

Things to do if this is going to be merged:

- [ ] Update brew java_requirement, cask java caveats DSL
https://github.com/Homebrew/brew/pull/4220
- [ ] Remove `java8` from versions 
https://github.com/Homebrew/homebrew-cask-versions/pull/5733
- [ ] New brew tag so the caveats and requirement are updated for people not on master.

### DIFF
--- a/Casks/java8.rb
+++ b/Casks/java8.rb
@@ -1,0 +1,85 @@
+cask 'java8' do
+  version '1.8.0_172-b11,a58eab1ec242421181065cdc37240b08'
+  sha256 'b0de04d3ec7fbf2e54e33e29c78ababa0a4df398ba490d4abb125b31ea8d663e'
+
+  java_update = version.sub(%r{.*_(\d+)-.*}, '\1')
+  url "http://download.oracle.com/otn-pub/java/jdk/#{version.minor}u#{version.before_comma.split('_').last}/#{version.after_comma}/jdk-#{version.minor}u#{java_update}-macosx-x64.dmg",
+      cookies: {
+                 'oraclelicense' => 'accept-securebackup-cookie',
+               }
+  name 'Java Standard Edition Development Kit'
+  homepage "https://www.oracle.com/technetwork/java/javase/downloads/jdk#{version.minor}-downloads-2133151.html"
+
+  # auto_updates true: JDK does not auto-update
+
+  pkg "JDK #{version.minor} Update #{java_update}.pkg"
+
+  postflight do
+    system_command '/usr/libexec/PlistBuddy',
+                   args: ['-c', 'Add :JavaVM:JVMCapabilities: string BundledApp', "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Info.plist"],
+                   sudo: true
+    system_command '/usr/libexec/PlistBuddy',
+                   args: ['-c', 'Add :JavaVM:JVMCapabilities: string JNI', "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Info.plist"],
+                   sudo: true
+    system_command '/usr/libexec/PlistBuddy',
+                   args: ['-c', 'Add :JavaVM:JVMCapabilities: string WebStart', "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Info.plist"],
+                   sudo: true
+    system_command '/usr/libexec/PlistBuddy',
+                   args: ['-c', 'Add :JavaVM:JVMCapabilities: string Applets', "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Info.plist"],
+                   sudo: true
+    system_command '/bin/ln',
+                   args: ['-nsf', '--', "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Home", '/Library/Java/Home'],
+                   sudo: true
+    system_command '/bin/mkdir',
+                   args: ['-p', '--', "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Home/bundle/Libraries"],
+                   sudo: true
+    system_command '/bin/ln',
+                   args: ['-nsf', '--', "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Home/jre/lib/server/libjvm.dylib", "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents/Home/bundle/Libraries/libserver.dylib"],
+                   sudo: true
+  end
+
+  uninstall pkgutil:   [
+                         "com.oracle.jdk#{version.minor}u#{java_update}",
+                         'com.oracle.jre',
+                       ],
+            launchctl: [
+                         'com.oracle.java.Helper-Tool',
+                         'com.oracle.java.Java-Updater',
+                       ],
+            quit:      [
+                         'com.oracle.java.Java-Updater',
+                         'net.java.openjdk.cmd', # Java Control Panel
+                       ],
+            delete:    [
+                         '/Library/Internet Plug-Ins/JavaAppletPlugin.plugin',
+                         "/Library/Java/JavaVirtualMachines/jdk#{version.split('-')[0]}.jdk/Contents",
+                         '/Library/PreferencePanes/JavaControlPanel.prefPane',
+                         '/Library/Java/Home',
+                       ],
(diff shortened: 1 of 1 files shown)


adidalal:
Is this the appropriate place to discuss naming? ie java@8 vs java8

commitay:
Just moving `java8` to this repo shouldn't impact anyone that has it installed as the cask itself would still has the same name, preserving upgrade functionality, etc. Renaming it would break it.

Also, people who:
```
$ brew tap homebrew/cask-versions
or
$ brew tap caskroom/versions

$ brew cask install java8
```` 
won't notice the difference if we keep the current name. 

It will break regardless of the name for people who try to install using the full tap path.
```
$ brew cask install homebrew/cask-versions/java8
or
$ brew cask install caskroom/versions/java8
```

claui:
Not sure if I’m stating the obvious here but considering the fact that Java 8 is an LTS version and Oracle’s habit of offering embarrassingly tiny support windows for everything else, I believe it’s worthy of treating Java 8 on par with, and keeping it in the same tap as, `java`.

MikeMcQuaid:
> Not sure if I’m stating the obvious here but considering the fact that Java 8 is an LTS version and Oracle’s habit of offering embarrassingly tiny support windows for everything else, I believe it’s worthy of treating Java 8 on par with, and keeping it in the same tap as, java.

Agreed 👍 

I'm 👍 on this move and 👍 on `java@8`.

commitay:
> Oracle’s habit of offering embarrassingly tiny support windows for everything 

java-se-lifecycle-131-a

As we will likely end up removing this cask in the not too distant future I don't see the benefit in a rename that will break it for everyone that already has it installed.

vitorgalvao:
> As we will likely end up removing this cask in the not too distant future I don't see the benefit in a rename that will break it for everyone that already has it installed.

Agreed.

Also, renaming it would introduce the weird case of having two different naming rules for versioned casks, which will only add to the confusion. And there’s no way to move all versioned casks to the new name system, [because the core doesn’t support it](https://github.com/Homebrew/homebrew-cask/issues/41990#issuecomment-353674034):

> In terms of current use and implementation `@` is only valid for version numbers.

Even if it were possible, though, I’m [still not convinced about `cask@beta`](https://github.com/Homebrew/homebrew-cask/issues/41990#issuecomment-353671524), though [I could probably be, given the right argument](https://github.com/Homebrew/homebrew-cask/issues/41990#issuecomment-353654377).

claui:
@commitay and @vitorgalvao Good points.

As an added benefit, Java 8 remains the latest supported LTS release for now anyway. Omitting the `@` reflects that fact nicely.

MikeMcQuaid:
> Also, renaming it would introduce the weird case of having two different naming rules for versioned casks, which will only add to the confusion.

I'd argue pretty strongly that the current system of `@` vs. not causes confusion as-is and we should plan to make things better rather than similar or worse.

It seems there's no current support to rename Casks but I'd also argue this is something that's worth doing and that consistency will make Homebrew Cask easier to use for Homebrew users. I also think  it's probably worth attempting and holding off a `java8` migration until this rename/migration support exists.

> And there’s no way to move all versioned casks to the new name system, because the core doesn’t support it

Well, not quite. You can just do `java-nightly` and `java@8`. I don't think that's a contradiction there; that again matches what Homebrew already does.

> As an added benefit, Java 8 remains the latest supported LTS release for now anyway. Omitting the @ reflects that fact nicely.

`java-lts` would better reflect that, IMO.

vitorgalvao:
> I'd argue pretty strongly that the current system of `@` vs. not causes confusion as-is and we should plan to make things better rather than similar or worse.

I don’t disagree, but seeing as we can’t have `@` for everything, I’d say it’s more confusing to say “on casks we sometimes use some and sometimes the other” than “on casks we always use this one”. Consistency across all casks seems more important than having users guess which one works at which time.

> It seems there's no current support to rename Casks but I'd also argue this is something that's worth doing

You’ll find no opposition to this. All cask maintainers want this and agree it’s desirable (I think there’s an open issue).

> You can just do `java-nightly` and `java@8`. I don't think that's a contradiction there; that again matches what Homebrew already does.

Looks weird to me. But maybe it’s just me.

> `java-lts` would better reflect that, IMO.

Hadn’t thought of that. I’m fine with it.

---

On a tangential note, one reason I’m not keen on having versioned casks moved to the main repo is the mess that’ll create. Keeping the separate versions in their separate repo is hard enough as it is, but having them separate makes it easier. If we could have directories inside the main repo to keep them separated, perhaps that’d help. And it that case it might make sense to do the same for the other repos (except `fonts`, those are too many and different enough to justify a separate repo).

Tangential to the tangent, we should also rethink the policy of not including casks in the analytics. From all the active maintainers from that time, I’m the only one that’s left (active), and since we’re now under the same organisation, having different policies on that will make sense to no one.

claui:
> `java-lts` would better reflect that, IMO.

I feel that most users who want LTS are going to appreciate a `java-lts` cask.

Just to clarify: would that imply we’re going to drop Java 8 support as soon as 11 LTS is around? (Per Cask’s philosophy, I wouldn’t even mind that.)

**tl;dr** I’m cool with `java-lts`.

commitay:
> would that imply we’re going to drop Java 8 support as soon as 11 LTS is around?

I disagree with `java-lts` as there isn't going to be another public LTS release after JDK8.

commitay:
> It seems there's no current support to rename Casks but I'd also argue this is something that's worth doing and that consistency will make Homebrew Cask easier to use for Homebrew users. I also think  it's probably worth attempting and holding off a `java8` migration until this rename/migration support exists.

I'm fine with this but currently there are no plans to implement it so it may not happen in time to be of use with this cask.

> You’ll find no opposition to this. All cask maintainers want this and agree it’s desirable (I think there’s an open issue).

Tap migrations and renames would be useful but we would need to consider what justifies adding something to the list of renames and how we eventually (hopefully) remove Casks from the list as with the frequency of cask renames it will end up being difficult to maintain. 

> If we could have directories inside the main repo to keep them separated, perhaps that’d help.

I think it would be worth discussing this further as I can see it both helping and hurting.

> Tangential to the tangent, we should also rethink the policy of not including casks in the analytics.

I've been thinking about this for the last couple of weeks as it would be useful in some cases. I'm just not sure about collecting analytics on _all_ casks. Should we open an issue to discuss this further?

claui:
@commitay Are you implying Oracle JDK 11 (LTS) is going to be behind a paywall? (It seems to me that Oracle hasn’t made a clear announcement yet, at least not in those three articles.)

Well, if paywalls are the plan, then no `java-lts` cask for now. Fine with me, and thanks for the heads up!

claui:
Also, 👍 for analytics on all casks in non-`private` taps, and 👍 for discussing it in a separate issue.

MikeMcQuaid:
> I don’t disagree, but seeing as we can’t have @ for everything, I’d say it’s more confusing to say “on casks we sometimes use some and sometimes the other” than “on casks we always use this one”. Consistency across all casks seems more important than having users guess which one works at which time.

I think we need to stop thinking about formulae and casks (and indeed Homebrew and Homebrew Casks) as separate things and separate projects because our users don't and the current situation is confusing to them.

> Yes. The Oracle docs are rather hilariously vague. JDK11 will be free until JDK12 is released and then you either pay for LTS or upgrade.

I think this adds another 👍 to `java-lts` for me.

commitay:
> > Yes. The Oracle docs are rather hilariously vague. JDK11 will be free until JDK12 is released and then you either pay for LTS or upgrade.

> I think this adds another 👍 to `java-lts` for me.

Can you explain you thoughts on this please @MikeMcQuaid?

MikeMcQuaid:
> when is it "long enough" to remove something from https://github.com/Homebrew/homebrew-cask/blob/master/tap_migrations.json, for example

It doesn't really matter in that case because it's a big JSON file that users don't really have to deal with and maintains add to very often. The overhead there is nil.

> I would like the docs to be unified, that's something I can look into separately.

That would be good 👍, same with the website(s). It should be pretty easy to add any Markdown files to Homebrew/brew's docs directory and they will just show up on docs.brew.sh.

joshka:
As a dev who relies on a stable JDK being available, I'm strongly for explicit versioning of the java cask. I'm fine with java8, but would prefer java@8. It is not OK to automatically upgrade / break a java-lts version even though public downloads have ended / support gone commercial etc.

commitay:
Closing as there seems to be a consensus that this shouldn't happen until after the implementation of tap migrations/renames.

PR #47877, opened 2018-05-30
D16 vsts

joshka:

--
Note, these are two separate VSTs from one company. I know that it's possible that we don't usually want to include the manufacturer's name in the cask, but I think given the quantity of possible instruments (http://d16.pl/products), it may be worth considering relaxing this idea for VSTs. An alternative might be to split audio plug ins out to another repo.

### DIFF
--- a/Casks/d16-frontier.rb
+++ b/Casks/d16-frontier.rb
@@ -1,0 +1,12 @@
+cask 'd16-frontier' do
+  version '1.0.0'
+  sha256 '7ed4762003b3a40b3306523404a1665a41bd7ba0016e6ed66073967cbfabdfd4'
+
+  url "http://d16.pl/pub/install/Frontier%20-%20#{version}.dmg"
+  name 'D16 Frontier'
+  homepage 'http://d16.pl/frontier'
+
+  pkg "Frontier-#{version}.pkg"
+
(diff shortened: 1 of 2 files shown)


commitay:
If these don't work without the associated equipment they should be in [`Homebrew/cask-drivers`](https://github.com/Homebrew/homebrew-cask-drivers)

joshka:
It’s not equipment related. Entirely software only. A VST is like an application with a UI that is loaded as a library in other music software (e.g. Ableton, Cubase, Bitwig, ProTools, Logic, etc). The VST software generally acts as either an audio instrument (a synth) or an effect.

commitay:
> given the quantity of possible instruments (http://d16.pl/products)

@vitorgalvao I can't find them but IIRC we've refused a couple of PRs that are similar to this because we don't want to end up with casks for all of them. I think this would be better suited to a personal tap.

vitorgalvao:
> IIRC we've refused a couple of PRs that are similar to this because we don't want to end up with casks for all of them.

I remember that.

> I think this would be better suited to a personal tap.

I agree. Thank you anyway, @joshka.

joshka:
No problem, that makes sense. I moved these to https://github.com/joshka/homebrew-audio and added the rest.

PR #47905, opened 2018-05-30
Remove chef casks

commitay:
Open-source and cli-only.

https://github.com/chef/chef-dk

https://github.com/inspec/inspec


coderanger:
Hey there, one of the devs from Chef here. We are getting a few user questions about why these were removed. From looking at a few of the similar PRs I'm guessing you are trying to focus on GUI apps and removing Casks for non-GUI apps? Just wanted to make sure we were on the same page, thanks!

commitay:
> From looking at a few of the similar PRs I'm guessing you are trying to focus on GUI apps and removing Casks for non-GUI apps?

Just open-source, cli-only software. These can be submitted as formulae to `homebrew/core`. If upstream would prefer to ship pre-built binaries I suggest [creating your own tap.](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md)

https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/rejected_casks.md
> The app is both open-source and CLI-only (i.e. it only uses the binary artifact). In that case, and in the spirit of deduplication, submit it first to Homebrew/core as a formula. If it is rejected, you may then try again as a cask (link us to the issue so we can see the discussion and reasoning for rejection).

coderanger:
Yeah, we don't support normal Homebrew-ish stuff since we only (officially) distribute ChefDK as a "everything and the kitchen sink" installer. Sounds like making our own tap is probably the best option. Thanks again!

coderanger:
The FAQ does mention you'll allow installers that aren't otherwise allowed in Homebrew, my understanding is that Homebrew's formula repo doesn't allow binary-blob installers, has that changed?

commitay:
> Homebrew's formula repo doesn't allow binary-blob installers, has that changed?

No, Homebrew formulae are required to build from source, the only exceptions are cross-platform binaries e.g. java and mono.

coderanger:
Okay, is the process for creating a Cask tap the same as for a normal formula? I've not tried it before but hopefully it works similarly :)

commitay:
Yes it's almost the same, the casks just need to be inside a `/Casks` directory.

coderanger:
For any Chef or InSpec users finding this thread later, we've moved these into a [`chef/chef` tap](https://github.com/chef/homebrew-chef).

You can install using `brew cask install chef/chef/chefdk` or `brew cask install chef/chef/inspec`.

commitay:
@coderanger We also had a PR for Workstation (https://github.com/Homebrew/homebrew-cask/pull/47503) if you want to add it to the Chef tap.

darkn3rd:
I am confused a little, as these are both open source, so not sure why these were knocked off for that reason.  For CLI only, that makes sense.

If open to suggestion, I would recommend having some policy to deprecate packages with some notification; to just purge previously accepted packages without notification leaves users hanging.

vitorgalvao:
@darkn3rd https://github.com/Homebrew/homebrew-cask/pull/47905#issuecomment-393687068

To have these reinstated officially, the way is to make formulae that build from source and submit to [Homebrew/core](https://github.com/Homebrew/homebrew-core).

coderanger:
@vitorgalvao That is not an option, we don't support building this from source because we need to ensure the installers are relatively identical across all platforms. I'm happy to keep hosting them in a tap, but ChefDK will never go into homebrew-core.

PR #47908, opened 2018-05-31
Remove haskell-platform

commitay:
This seems to be equivalent to `brew install ghc cabal-install stack` https://www.haskell.org/platform/contents.html

I don't think this was covered in the deduplication issue but I might have missed it if it was discussed elsewhere.


commitay:
When I merged this I didn't realise it was blacklisted in [`missing_formula`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/missing_formula.rb#L68-L75) and pointed to the cask.
```
        when "haskell-platform" then <<~EOS
          We no longer package haskell-platform. Consider installing ghc
          and cabal-install instead:
            brew install ghc cabal-install

          You can install with Homebrew-Cask:
            brew cask install haskell-platform
          EOS
```
Do we want to restore the cask or update the blacklist message to something like:
```
          We no longer package haskell-platform. Consider installing ghc,
          cabal-install and stack instead:
            brew install ghc cabal-install stack
          EOS
```

cc @MikeMcQuaid @ilovezfs @vitorgalvao

ilovezfs:
the latter

commitay:
https://github.com/Homebrew/brew/pull/4302

PR #47930, opened 2018-05-31
Update nestopia to 1.4.2

core-code:

### DIFF
--- a/Casks/nestopia.rb
+++ b/Casks/nestopia.rb
@@ -6,6 +6,8 @@ cask 'nestopia' do
   name 'Nestopia'
   homepage 'http://www.bannister.org/software/nestopia.htm'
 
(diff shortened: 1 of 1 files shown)


core-code:
i've updated the cask because it fails to install because the download now contains a folder with the version number. i'm not sure if there is a way to support that without introducing the version number into the cask.

adidalal:
The cask will break on update immediately - is there an appcast that can be monitored so we can minimize that time?

core-code:
didn't find any. btw nestopia 1.4.2 now seems to require 10.11, should the cask be updated to note that?

vitorgalvao:
> btw nestopia 1.4.2 now seems to require 10.11, should the cask be updated to note that?

Yes, please.

PR #47972, opened 2018-06-01
Update tikzit to 1.1

gtmgianni:

### DIFF
--- a/Casks/tikzit.rb
+++ b/Casks/tikzit.rb
@@ -3,7 +3,7 @@ cask 'tikzit' do
   sha256 '7e00a64fe5c96963ae5b09f1f59698b8b487239c53eed01a3bf012cf024f0686'
 
   # downloads.sourceforge.net/tikzit was verified as official when first introduced to the cask
-  url "https://downloads.sourceforge.net/tikzit/TikZiT-osx-#{version}.zip"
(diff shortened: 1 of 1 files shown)


Casks/tikzit.rb
```diff
@@ -2,11 +2,12 @@
   version '1.1'
   sha256 '7e00a64fe5c96963ae5b09f1f59698b8b487239c53eed01a3bf012cf024f0686'
 
+  # downloads.sourceforge.net/tikzit was verified as official when first introduced to the cask
   url "https://downloads.sourceforge.net/tikzit/TikZiT-osx-#{version}.zip"
```
vitorgalvao: The current download URL on the homepage is from `bintray`.

gtmgianni: The `sha256` for the file from `bintray` is different though...

gtmgianni: Not sure why, I thought we could stick with this URL to save the trouble of verifying the `sha256`.

vitorgalvao: But as it is, none of the official sources link to this link on sourceforge, so we cannot claim it as official anymore.

Change it to the bintray one. We‘re changing most of the important things about this cask, like `home` and `appcast`, we can consider it makes sense for them to have altered the package as well and not require verification.

vitorgalvao (approved):

PR #47980, opened 2018-06-01
Update p4v to 2018.2

gtmgianni:


gtmgianni:
The SHA-256 hash listed on the `appcast` has also changed to reflect the new `sha256`, but I doubt that is enough to verify the file wasn't maliciously altered?

vitorgalvao:
> but I doubt that is enough to verify the file wasn't maliciously altered?

Yeah. Someone who could’ve altered the download file could just as easily alter the shasum document.

commitay:
With perforce casks we actually have three different ways the version is handled because they update in place.

https://github.com/Homebrew/homebrew-cask/blob/master/Casks/perforce.rb
https://github.com/Homebrew/homebrew-cask/blob/master/Casks/p4merge.rb

PR #48031, opened 2018-06-03
Update wavesurfer to 1.8.8p5

gtmgianni:


gtmgianni:
File isn't code signed, but the developer did re-release this version of `wavesurfer` [according to this feed.](https://sourceforge.net/p/wavesurfer/activity/?page=0&limit=100#5b0e53feee24ca74371540aa)

I can't find a way to publicly contact the developer to confirm the `sha256` change, and since the Sourceforge forum for this application doesn't appear to be active I doubt there's much point in using that.

vitorgalvao:
That feed seems to be auto-generated, like the atom feed for github releases, so we can’t rely on it’s trustworthiness.

brianmorton:
I know HBC uses sha256, but sourceforge does confirm the md5/sha1 on the site:
image

So surely if you verify either (or both) of those, then presumably you can trust the sha256 of the same binary.

```
$ wget https://downloads.sourceforge.net/wavesurfer/wavesurfer-1.8.8p5-osx-i386.dmg
[...]
wavesurfer-1.8.8p5-osx-i386.dmg   100%[==========================================================>]  12.26M  

2018-07-16 16:19:15 (3.11 MB/s) - ‘wavesurfer-1.8.8p5-osx-i386.dmg’ saved [12859412/12859412]

$ md5 wavesurfer-1.8.8p5-osx-i386.dmg
MD5 (wavesurfer-1.8.8p5-osx-i386.dmg) = 92565b5ea8763b8d2701a0016f9a9889

$ shasum wavesurfer-1.8.8p5-osx-i386.dmg
2062e7d73faeee9da9e8e4fef5a9dd6492395c91  wavesurfer-1.8.8p5-osx-i386.dmg
```

brianmorton:
@commitay ? @reitermarkus ?

vitorgalvao:
> I know HBC uses sha256, but sourceforge does confirm the md5/sha1 on the site

Isn’t that auto-generated?

brianmorton:
I have no idea, but it's there.

What is the difference if the dev posts it or if it's auto generated? Presumably only the dev would be able to upload to that account...?

vitorgalvao:
> What is the difference if the dev posts it or if it's auto generated? Presumably only the dev would be able to upload to that account...?

The point of verifying sha-only changes is to ensure the dev account wasn’t compromised (which happened once and was what triggered the policy).

brianmorton:
If the dev account was compromised couldnt they just post the new (bad) checksum? I'm kind of confused.

vitorgalvao:
> If the dev account was compromised couldnt they just post the new (bad) checksum?

Yes, which is why we need public confirmation from the developer that *that* specific case was their doing, and we do so by *asking* them (that part is important).

1. Developer account is compromised. Malicious actor pushes a new build and leaves.
2. Dev is unaware, as everything looks the same (who remembers and checks their own checksums?).
3. We get a submission for the changed checksum and ask the developer to confirm it.
4. Dev doesn’t remember making the change in place so checks their own local copy/rebuilds and sees the checksum does not match the online version.
5. Conclusion: online version was compromised.

vitorgalvao:
Verified checksum locally. Merging since our rules regarding verification have changed.

Issue #48051, opened 2018-06-04
Remove appcast checkpoints?


commitay:
Refs: https://github.com/Homebrew/homebrew-cask/pull/47728, https://github.com/Homebrew/homebrew-cask/issues/38376, https://github.com/Homebrew/homebrew-cask/issues/43605, https://github.com/Homebrew/homebrew-cask/issues/41240 and others

This has been discussed briefly in a few places, opened this for a proper discussion.

This would basically be as the title says, the `checkpoint` string would be removed along with the `_appcast_checkpoint` command while keeping the `appcast` URL that can be parsed with external scripts/tools.

cc @brianmorton @colindunn @gtmgianni @leipert People that use their own scripts/tools for cask updates.

Would also like to know what other contributors think about this so please feel free to comment.

leipert:
Thanks for pinging me!

Are they used for anything relating to security? If not, remove them! It makes no sense for the casks which @jcbot updates. I always check against the version string, if we need updates and not against the hash of the URL.

gtmgianni:
Although my scripts rely entirely on the appcast `checkpoint`, I don't think using the `checkpoint` is a long term solution for updating casks. This is mainly due to the fact that there are so many casks with outdated `checkpoint` hashes, that don't actually need a version bump.

I have a blacklist of around 1000 casks that fall into this category, although by the time I run the script which scrubs the blacklist of any invalid entries (i.e. casks whose `checkpoint` hashes have changed yet again), that number decreases to around 700. In an ideal scenario, all ~300 of those casks that were removed from the blacklist would need a version bump, but in reality many of them are just unstable appcasts which get added back to the blacklist with their new `checkpoint` hash once I've manually checked them.

I think removing the `checkpoint` is something we should consider because many appcasts are unstable.

I like the idea of tailored code for each cask being used to find the latest version, as discussed in https://github.com/Homebrew/homebrew-cask/issues/41240. Out of curiosity, would we have some sort of bot which automatically submits the PRs, or would different contributors be able to update the casks themselves? If the latter is the case, how would we avoid pull request duplicates?

vitorgalvao:
My initial idea:

* Have the tailored update code for each cask, and store it in a different repo.
* Have a bot that submit the PRs.
* In `cask-repair`, blacklist casks that are updated by the bot.

[...]

PR #48060, opened 2018-06-04
Update riverdesign-sparkle to latest

core-code:

### DIFF
--- a/Casks/riverdesign-sparkle.rb
+++ b/Casks/riverdesign-sparkle.rb
@@ -1,9 +1,9 @@
 cask 'riverdesign-sparkle' do
-  version :latest
-  sha256 :no_check
+  version '2.6.1'
+  sha256 'cc9e72bbb09e22379b69430fc9e94ceedc7316db952a2d6fa392f6762a147f74'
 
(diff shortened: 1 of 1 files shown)


adidalal:
Mind adding an [appcast](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md)?

Appcast is at `https://sparkleapp.com/update/appcast.php?lang=en` and version looks to be `3.6.1`

core-code:
i am not sure if an appcast is beneficial if we have version latest, but thanks, i've added it

vitorgalvao:
When using an appcast, even if `url` in unversioned, we add a version. This is because with an appcast we can detect updates faster and it won’t break as much.

vitorgalvao (approved):

core-code:
thanks for the info. is there actually an automated system in place to find outdated apps with appcasts? (because it seems to remember submitting updates for a lot of those)

vitorgalvao:
> is there actually an automated system in place to find outdated apps with appcasts? 

Several. See https://github.com/Homebrew/homebrew-cask/issues/48051.

PR #48111, opened 2018-06-05
Update isyncr to 5.14.0

brianbrownton:

### DIFF
--- a/Casks/isyncr.rb
+++ b/Casks/isyncr.rb
@@ -1,14 +1,18 @@
 cask 'isyncr' do
-  version '5.14.0'
-  sha256 '68e65554b12083c35b0180c6535cefcc7897c0aa4902848dcf706459ecfab3ad'
+  if MacOS.version <= :lion
+    version '5.0.8'
+    sha256 '1b1c0e6054b890fa7dd903cc680fd2543bc6b40b34cde2d23f4c2e355dcffb87'
+  elsif MacOS.version <= :sierra
+    version '5.6.5'
+    sha256 '8cd6b1c96a902d8810e52aab6a980424370237617bfd3ff574367ff1ce8d4f4e'
+  else
+    version '5.14.0'
+    sha256 '68e65554b12083c35b0180c6535cefcc7897c0aa4902848dcf706459ecfab3ad'
+  end
 
   url "http://www.jrtstudio.com/files/iSyncr%20Desktop%20#{version}.pkg"
-  appcast 'http://www.jrtstudio.com/files/SlashiSyncr25.js',
-          checkpoint: '65385df99037be7b37ea118f503a51036b2af6c11a3c6230ff5b4f3576dee4ab'
   name 'iSyncr Desktop'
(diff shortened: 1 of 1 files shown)


brianmorton:
Actually pls hold on merging this - there are different versions for different macOS versions.

brianmorton:
Should be good to merge now, if you all are happy with it.

vitorgalvao:
Why drop the appcast?

brianmorton:
@vitorgalvao 

It is unusual for an appcast to point to a JS file ( http://www.jrtstudio.com/files/SlashiSyncr25.js ), so I had a look.

The contents of the file seem to facilitate downloading from an HTML page, however all the version numbers were wrong/outdated - the mac version was for `5.0.8` - which was the lowest common denominator in the mac versions (lion and older). So clearly it was no longer being updated for recent versions, and it really never should have served as an "appcast".

brianmorton:
To expand further:

I looked at a candidate for a new appcast. The equivalent up to date file of the old appcast is http://www.jrtstudio.com/files/SlashiSyncr32.js , however that doesn't seem to correlate to the version numbers (specifically the `32` at the end) so it would be hard to tell if/when it became outdated, like the old appcast.

I also looked at the changelog ( http://www.jrtstudio.com/iSyncr/Changelog ), however even that is outdated compared to the version offered for download (`5.12` on the changelog vs `5.14` for download) - not a great start. After that, I have had issues with using a webpage for appcasts before due to the risk of dynamic content, so since it was outdated and I didn't want to run that risk I just felt it was better to drop the appcast.

I think no information is better than bad information.

Casks/isyncr.rb
```diff
@@ -1,15 +1,19 @@
...
+    sha256 '68e65554b12083c35b0180c6535cefcc7897c0aa4902848dcf706459ecfab3ad'
+  end
 
   url "http://www.jrtstudio.com/files/iSyncr%20Desktop%20#{version}.pkg"
-  appcast 'http://www.jrtstudio.com/files/SlashiSyncr25.js',
-          checkpoint: '65385df99037be7b37ea118f503a51036b2af6c11a3c6230ff5b4f3576dee4ab'
   name 'iSyncr Desktop'
   homepage 'http://www.jrtstudio.com/iSyncr-iTunes-for-Android'
 
-  depends_on macos: '>= :mountain_lion'
```
commitay: Lets drop the lion version and keep the depends_on mountain_lion

brianmorton: With respect, I believe 10.7 is Lion and I was just trying to follow what is listed on the site's download page:

image

Based on the bit about
> Mac 10.7 & older

I dropped the `depends_on` stanza.

commitay: We don't need to have a 10.7 version just because the homepage has one.

brianmorton: Right... well now I'm not really clear on what exactly you want changed. So I will leave this PR as it is, and you can either close it or modify it to suit your goal.

commitay:
Thanks @vitorgalvao and apologies for being unclear @brianmorton

PR #48256, opened 2018-06-09
Update marked to 2.5.18,952

core-code:

### DIFF
--- a/Casks/marked.rb
+++ b/Casks/marked.rb
@@ -1,10 +1,10 @@
 cask 'marked' do
-  version '2.5.17950'
-  sha256 '38f0bc6d5ea13ef202a18ea770cb5ee0f2097755c80e6632da3f368036202278'
+  version '2.5.18,952'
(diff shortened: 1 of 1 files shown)


Casks/marked.rb
```diff
@@ -1,10 +1,10 @@
 cask 'marked' do
-  version '2.5.17950'
-  sha256 '38f0bc6d5ea13ef202a18ea770cb5ee0f2097755c80e6632da3f368036202278'
+  version '2.5.18,952'
```
commitay: Please change this back to `2.5.18952`

commitay (changes requested):

Casks/marked.rb
core-code: actually the version (short version string) is 2.5.18
and the bundle version is 952
i don't think it makes sense to just mix them together in one long string just because they are next to each other in the download URL

commitay: It isn't needed, please change it back.

PR #48303, opened 2018-06-11
Update retroarch-cg appcast

brianbrownton:

### DIFF
--- a/Casks/retroarch-cg.rb
+++ b/Casks/retroarch-cg.rb
@@ -3,6 +3,8 @@ cask 'retroarch-cg' do
   sha256 'c47c19a3a3774de04bfeb9bba0b41f396fb2e0d14570062d1704a37d32cf527b'
 
   url "https://buildbot.libretro.com/stable/#{version}/apple/osx/x86_64/RetroArch_CG.dmg"
+  appcast "https://buildbot.libretro.com/stable/#{version}/apple/osx/x86_64/",
+          checkpoint: '6ee351187a709b0d7c8a545a59d3d1152cf9666a15da94b721639da670675e77'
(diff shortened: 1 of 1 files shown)


Casks/retroarch-cg.rb
```diff
@@ -3,6 +3,8 @@
   sha256 'c47c19a3a3774de04bfeb9bba0b41f396fb2e0d14570062d1704a37d32cf527b'
 
   url "https://buildbot.libretro.com/stable/#{version}/apple/osx/x86_64/RetroArch_CG.dmg"
+  appcast "https://buildbot.libretro.com/stable/#{version}/apple/osx/x86_64/",
```
commitay: Not sure I understand what you're trying to do with this, when the next version is released this won't find it?

brianmorton: @commitay Yep I think you're absolutely right. The current appcast isn't working well (checkpoint/contents always change) so I was trying to refine it down to something less mutable... but I think this goes too far.

commitay: Within the next day or so `checkpoint` will be removed, just waiting until we are ready to update everything at the same time.

brianmorton: I have long since abandoned the official checkpoints, and now use an outside db and blah blah blah to manage this stuff. Appcasts that result in dynamic content, however, are still problematic (which is what I was trying to solve here).

commitay: Once checkpoint is removed we can probably switch the appcast to http://www.retroarch.com/?page=platforms and eventually try to parse either the text on page or the hrefs for the macOS releases.

PR #48388, opened 2018-06-13
Update file-juicer to 4.66

brianbrownton:


brianmorton:
Just wanted to add an appcast, but it seems the sha256 for this version also changed. I don't want to get into that web, so I revert the checksum...

PR #48553, opened 2018-06-16
token_reference: add links to tap specific naming conventions

commitay:
Ref: https://github.com/Homebrew/homebrew-cask-versions/pull/5828#issuecomment-397688694


joewiz (approved):
It’s going to be so helpful to have these in one place!

PR #48654, opened 2018-06-18
Update stereo-tool to 9.00

brianbrownton:

### DIFF
--- a/Casks/stereo-tool.rb
+++ b/Casks/stereo-tool.rb
@@ -1,5 +1,5 @@
 cask 'stereo-tool' do
-  version '8.54'
+  version '9.00'
   sha256 'e2fa88bdaa32517adb8b515ff30f6e00fbfb9b09dfce9e69665c9d5ec9f53d57'
 
   url 'http://www.stereotool.com/download/stereo_tool.zip'


vitorgalvao:
They upped the version but did not change the download?

brianmorton:
@vitorgalvao well it uses a static (unversioned) download url - `http://www.stereotool.com/download/stereo_tool.zip` so that's not unusual I guess.

The unusual part is that the change announcement is in the appcast which is tagged for version `8.50` or more generiously, version `8` releases:
image

vitorgalvao:
> well it uses a static (unversioned) download url - `http://www.stereotool.com/download/stereo_tool.zip` so that's not unusual I guess.

It is a bit unusual. It either means they’ve bumped the version but not updated the tool or that on the last update to the cask, the version wasn’t updated.

brianmorton:
There are ~~lots~~ some casks that do this. I have previously recommended that we should just change those casks to be `:latest` and `:no_check`, but I think you all prefer to have versions and checksums... which I can understand. However in these cases I don't think it makes sense.

vitorgalvao:
> but I think you all prefer to have versions and checksums

At one point we decided we had to make a decision on which we’d prefer (because there are many apps where we can have one or the other) and we let the users vote. They overwhelmingly preferred versioned casks.

brianmorton:
I understand, but I still think it makes no sense at all in these cases. They are essentially just a waste of contrib time if the download URL is the same.

vitorgalvao:
They’re not a waste because the shasum is (usually) updated.

PR #48656, opened 2018-06-18
unity casks: update appcast

brianbrownton:


commitay:
Why?

brianmorton:
Great question.

The top two reasons are both of utility to me during maintenance:
• I can see previous versions on the same page (helps with automated checking)
• I can get a link to the release notes, which include the revision (eg `732dbf75922d`) which we use as part of the version. Granted I could probably programmatically find this page based on the version number (eg `https://unity3d.com/unity/whatsnew/unity-2018.1.5`), however the version we store (`2018.1.5f1,732dbf75922d`) would require additional parsing to format it correctly. Version before comma, drop last two digits... etc).

PR #48666, opened 2018-06-19
Remove `brew cask search` from documentation.

reitermarkus:


USAGE.md
```diff
@@ -14,29 +14,26 @@ $ brew --version
...
+==> Casks
 google-chrome
-```
-
-A `search` command with no search term will list all available Casks:
-
-```bash
-$ brew cask search
-# <list of all available Casks>
+homebrew/cask-versions/google-chrome-beta
```
vitorgalvao: I’m not getting the `homebrew/cask-versions/` prefix. Does that only happen if the repo is untapped?

reitermarkus: > Does that only happen if the repo is untapped?

Yes.

PR #48698, opened 2018-06-19
Update securityspy to 4.2.6

brianbrownton:


brianmorton:
another with a static url which I think should be `:latest` `:no_check`

vitorgalvao:
> another with a static url which I think should be `:latest` `:no_check`

Again, if it were, we’d lose the `sha256` verification.

PR #48700, opened 2018-06-19
Update resolume-avenue to 6.0.9

brianbrownton:

### DIFF
--- a/Casks/resolume-avenue.rb
+++ b/Casks/resolume-avenue.rb
@@ -10,5 +10,8 @@ cask 'resolume-avenue' do
 
   pkg 'Resolume Avenue Installer.pkg'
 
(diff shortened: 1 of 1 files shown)


Casks/resolume-avenue.rb
```diff
@@ -1,8 +1,9 @@
 cask 'resolume-avenue' do
-  version '6.0.8.60677'
-  sha256 'f6f45145d66966ad29ac1e0d0a8e1ae28f73cb94b6005cca47724e99ede284de'
+  version '6.0.9'
```
commitay: Probably want to keep the full sparkle version.

brianmorton: Sorry for the delay, I just realized you were asking me to make a change. 

Since we don't use the "sparkle" version for anything in the cask, why do we want to keep it?

PR #48796, opened 2018-06-21
Update pocket-casts to 0.9.9

brianbrownton:

### DIFF
--- a/Casks/pocket-casts.rb
+++ b/Casks/pocket-casts.rb
@@ -1,6 +1,6 @@
 cask 'pocket-casts' do
-  version '0.9.8'
-  sha256 'af64a8f2d7cf9cf669c338ffb48f7430c332db09cdfa16a891055ad671fdb8fe'
+  version '0.9.9'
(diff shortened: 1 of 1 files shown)


brianmorton:
Static download URL, no version interpolation in the rest of the cask... I see no reason why this can't be `:latest' `:no_check`

reitermarkus:
If the `appcast` contains the version, it should use this version, even if the URL is unversioned, but there is `https://static2.pocketcasts.com/mac/pc_0_9_9.zip` in the appcast.

brianmorton:
I understand there is a preference for this (for some reason), but the practical outcome is that it has contribs doing busy work (or at best marginally valuable).

When it becomes outdated (such as before I subbed this PR), all that happens is the `brew cask install`s start having a checksum problem when they didn't need to. Going to `:latest` and `:no_check` fixes that and prevents me from having to sub a PR for appearances/posterity.

I will grant you that there is perhaps some marginal security benefit to the version + checksum, but generally speaking if an attacker could compromise the website to modify the binary, they could also just issue a new version of the binary and we would update it anyway...

PR #48802, opened 2018-06-21
Update rest-time to 1.10

brianbrownton:

### DIFF
--- a/Casks/rest-time.rb
+++ b/Casks/rest-time.rb
@@ -1,6 +1,6 @@
 cask 'rest-time' do
-  version '1.08'
-  sha256 '3a9659e17af7cb60de4a1309b800c791c8f8a16a586fd9faefe5e5ddbecc4d0f'
+  version '1.10'
+  sha256 '8fac7e3f8caec3e68744c7d4b26cd36c7d4f0a4712711ba878d92ae0a2caba4f'
 
   url 'http://www.publicspace.net/download/RestTime.dmg'
   appcast 'http://www.publicspace.net/app/resttime.xml'


brianmorton:
another with no version interpolation that should be `:latest` `:no_check`

vitorgalvao:
> another with no version interpolation that should be `:latest` `:no_check`

Please stop posting that. Posting that will not change the policy, because **this is as intended**. Users prefer versioned casks because they can have a `sha256`. And since the cask has an `appcast`, we can more accurately know when it’s outdated, like presumably you just did. You’re demonstrating the system works.

brianmorton:
@vitorgalvao we will have to agree to disagree then, and I will just blacklist the casks I don't believe to be worth maintaining if HBC is unwilling to consider changing the policy.

vitorgalvao:
As I’ve explained previously, the policy was made at the behest of users. At the time I was already a maintainer and I preferred `:latest` and `:no_check` to be the default whenever possible, but it was decided users should be the ones to make that choice, and they did.

I don’t get why you want to blacklist these. They have an appcast — the effort to update them is the same as if the URL were versioned.

brianmorton:
I understand your points, and indeed it is almost the same effort (cask-repair -b interrupts the automation to ask for a URL on these casks, which is a minor annoyance we've discussed before where you proposed a workaround).

My point is one of protest - "users" are free to maintain them since they supposedly feel so strongly about their position. If you would prefer I not publicly state it on the casks (while updating them), then I think the civilized thing to do moving forward would be to just not update them and let the users maintain them. That way it does not annoy me or you, and users can have what they wanted.

Win-win-win.

brianmorton:
Comes out to 910 casks currently: https://pastebin.com/TDGfL1ur

PR #48819, opened 2018-06-21
Update chronicle to 8.0.0

core-code:


core-code:
i am not sure why it says only the SHA changed, previously the cask downloaded version 7 (although it claimed to download version 8), so obviously downloading the right file instead of the wrong file has another SHA

vitorgalvao:
The check expects the the `sha256` and `version` to change. If `sha256` but not `version`, it’ll shw that message, even it the `url` changed.

core-code:
sorry we have a problem. because the SHA check failed, i didn't notice that there is another problem too:
Error: It seems the App source '/usr/local/Caskroom/chronicle/8.0.0/Chronicle.app' is not there.

app line should be changed to

app "Chronicle #{version.major}.app"

PR #48941, opened 2018-06-25
Delete lastpass.rb

### LINKED ISSUE #48300: Lastpass cask doesn't install lastpass app, but browser extension installer
When installing the lastpass cask, only the browser extension installer is downloaded/"installed". It's not the regular lastpass app, which would be put into the applications folder, which you can currently only get from the app store, you would expect.

I propose either adding the "real" lastpass app or switching it for the current app or making the name of the current app more descriptive and clear.

vitorgalvao:
This only installs the browser extension and not the app (MAS-only), so it’s a bit confusing. I doubt many people care about this cask.

Closes https://github.com/Homebrew/homebrew-cask/issues/48300


commitay (approved):

jhabdas:
Does this mean those with LastPass installed via dotfiles will no longer receive updates to the browser extension should [another security vulnerability](https://blog.lastpass.com/2017/03/security-update-for-the-lastpass-extension.html/) be discovered? Or is that process managed by the plug-in code?

vitorgalvao:
> Does this mean those with LastPass installed via dotfiles will no longer receive updates to the browser extension

You probably weren’t getting them from us anyway, because this cask was not included in normal `upgrade` procedure.

> Or is that process managed by the plug-in code?

That is likely, but you’ll have to ask the LastPass developers.

vitorgalvao:
We don’t have many casks for browsers plugins (do we even another one?). Casks that are excluded from normal `upgrade` are the ones that auto upgrade themselves and the ones who have an unversioned link and no appcast, since we have no practical way to discern when they’ve been updated upstream. Do `man brew-cask` and look for `upgrade` for more info (include how to force those casks to upgrade).

PR #48995, opened 2018-06-26
Update tunnelblick to 3.7.7beta02,5090

brianbrownton:


core-code:
huh, why update to a beta version?

brianmorton:
Good catch - their GH releases don't really mark a latest release (it's on the uninstall tool currently) and the previous release being `3.7.6a` I assumed was an alpha, so going to a beta seemed fine to me.

But after looking at their homepage, they clearly state `3.7.6a` is their stable, so I will close this.

PR #49018, opened 2018-06-27
Update TablePlus hash.

huyphams:

I updated the sha256 because I need to update the current version but don't want to bump the build (which can trigger an update for all users).


vitorgalvao:
[Confirmed as the developer](https://github.com/TablePlus/TablePlus/commits?author=huyphams).

PR #49090, opened 2018-06-29
Update 5 casks: uninstall/zap

commitay:


MindTooth:
For docker; can this be included?  Just did a zap, but after a `fd docker /`, this file were left behind.
```sh
$ rm /private/var/run/docker.sock 
  rm: /private/var/run/docker.sock: Permission denied
```

vitorgalvao:
@MindTooth Those are temporary resources that macOS will take care of automatically.

PR #49126, opened 2018-06-29
Update bean to 3.2.10

suschizu:


vitorgalvao:
Thank you, but:

* We technically only support the latest two macOS versions, adding these right now will likely have no benefit.
* You’re adding a version to an unversioned url with no appcast, which will cause the cask to break.
* Installation is failing.

PR #49149, opened 2018-07-01
Update sha256 for cask rar 5.6.0

g-bon:

[closed issues]: https://github.com/Homebrew/homebrew-cask/issues?q=is%3Aissue+is%3Aclosed

The sha256 for rar 5.6.0 https://www.rarlab.com/rar/rarosx-5.6.0.tar.gz was not correct and causan ing error. Fixed with correct sha256.

### DIFF
--- a/Casks/rar.rb
+++ b/Casks/rar.rb
@@ -1,6 +1,6 @@
 cask 'rar' do
   version '5.6.0'
-  sha256 '381eb0c0645bfbd603e7e20a6a8cb34e484c6eeca161a1b3017cf93415448fde'
+  sha256 '381ca05ca2701cf72148da49d36a4dda4e8e29cd0c3879f697b2861326277d5d'
 
   url "https://www.rarlab.com/rar/rarosx-#{version}.tar.gz"
   name 'RAR Archiver'


vitorgalvao (changes requested):

g-bon:
Thanks, VirusTotal link:

https://www.virustotal.com/#/file/381ca05ca2701cf72148da49d36a4dda4e8e29cd0c3879f697b2861326277d5d/details

vitorgalvao:
That is expected and not the root cause of the issue. The problem is the package was updated in place upstream. Since this is not codesigned the virustotal verification is not enough. We need public confirmation from the developer that they changed the shasum to that one.

g-bon:
Ok I don't fully understand why the Sha should be different, but happy to learn if someone has time to explain.

This said I don't think the file was ever changed, couldn't just be that it was just pushed with the wrong SHA255 in this pull request [48968](https://github.com/Homebrew/homebrew-cask/pull/48968) 4 days ago and for some reason the CI didn't catch the problem?Never worked for me since it was updated

vitorgalvao:
> Ok I don't fully understand why the Sha should be different, but happy to learn if someone has time to explain.

The result of a shasum is always the same when applied to the same file. So if it changed, it means the file also changed. That means the file was replaces in place on the server. That’s somewhat common, and developers usually do that for updates they deem too small to be worth bumping the version.

> couldn't just be that it was just pushed with the wrong SHA256 in this pull request 48968 4 days ago and for some reason the CI didn't catch the problem?

That is highly unlikely. The former case is happens quite a lot.

g-bon:
Thanks a lot for the explanation. Makes sense.

Viktor45:
Hello,



This is a legitimate modification.

32 bit app support will be removed from OS X soon and latest OS X beta
already issues a warning when running 32 bit apps. We received user
complains about it, so we replaced 32 bit RAR for OS X version
with 64 bit version.

We plan to publish only 64 bit version in the future.

Eugene

Viktor45:
from: roshal(at)rarlab.com

> Hello,
> 
> This is a legitimate modification.
> 
> 32 bit app support will be removed from OS X soon and latest OS X beta
> already issues a warning when running 32 bit apps. We received user
> complains about it, so we replaced 32 bit RAR for OS X version
> with 64 bit version.
> 
> We plan to publish only 64 bit version in the future.
> 
> Eugene

vitorgalvao:
Thank you @Viktor45, but we need *public* confirmation. For all we know you could’ve made up that email.

Viktor45:
@vitorgalvao you can contact rar developer yourself

vitorgalvao:
@g-bon Any of those is fine. They can tweet at [@homebrewcask](https://twitter.com/homebrewcask) from an official account, or comment here (if the person can prove they work for them). Any method is fine, as long as a person who can prove they’re from the team says it’s OK in a way we can vet.

euroshal:
I've added info about this change to "News" section of top www.rarlab.com page. Search for "64 bit RAR for OS X". If it is still not convincing, you can write to developers email from "Feedback" section of same site and I'll confirm it by email.

g-bon:
Thanks a lot @euroshal. Let's wait @vitorgalvao for confirmation but I guess that should do it.

vitorgalvao:
That does it. Thank you, everyone!

PR #49197, opened 2018-07-02
Update yt-music to 1.0.4,5:1530537206

brianbrownton:

### DIFF
--- a/Casks/yt-music.rb
+++ b/Casks/yt-music.rb
@@ -1,9 +1,10 @@
 cask 'yt-music' do
-  version '1.0.3'
-  sha256 'b72fccdd965d590dfb0d75af26ac1fd0c1a0abd712db63dba753eaa16f2362b4'
+  version '1.0.4,5:1530537206'
+  sha256 'd69f7e854b5db387f04c8eee2e2479cfe71368d3c6c05b2b736645e6ba6db539'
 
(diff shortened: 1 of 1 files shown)


brianmorton:
per this conversation:
https://github.com/steve228uk/YouTube-Music/issues/17

Seems like the github releases page is less reliable at having the latest release than devmate / new appcast.

vitorgalvao (changes requested):
From the conversation, this seems like a momentary lapse rather than something that’ll keep occurring. I’d rather keep it as it is currently. The less we need the `was verified as official when first introduced to the cask` comment, the better.

brianmorton:
@vitorgalvao I understand.

I will say that since I started the automated appcast based scanning, I have noticed that a lot of developers tend to treat the github releases page as secondary/tertiary (or an after thought), which is why I subbed this PR - because of this feeling if there is another release channel I would tend to prefer it, even though I like the uniformity provided by having a lot of casks using a github appcast.

vitorgalvao:
@brianmorton Is this something your system would notice if it happens again? I say this in the sense of: keep this cask with the github release, but if it happens again let’s change it.

brianmorton:
Not in an automated way - when this stuff happens I usually open an issue asking about a macOS binary for the latest release. The issues that get attention the result is usually that the dev forgot to upload the release to github. Had I never inquired I'm not sure if or when they would have noticed.

brianmorton:
https://github.com/Homebrew/homebrew-cask/pull/49243

PR #49297, opened 2018-07-06
Update soundtoys 5.2.4.13665

yutarody:


yutarody:
I do not know the cause of fail ... The local test is successful

vitorgalvao:
It’s failing because you’re changing a bunch of useful broad stuff to be specific and not catch everything.

vitorgalvao:
It’s failing because you’re changing a bunch of useful broad stuff to be specific and not catch everything. Don’t replace the `*`. Also, keep the comma logic, `split` is harder to understand.

Either way, this now seems to be a walled build (they ask for an email and host on a different domain) so we’ll have to remove it.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/49399

PR #49306, opened 2018-07-06
Update deezer to 0.9.2

csi-adziahel:

### DIFF
--- a/Casks/deezer.rb
+++ b/Casks/deezer.rb
@@ -1,6 +1,6 @@
 cask 'deezer' do
-  version '0.8.0'
-  sha256 '6ff9dbae14de1dbbca6f3a51f8b618bf87c6d45f4a8692113b5f3e5354f7cc60'
+  version '0.9.2'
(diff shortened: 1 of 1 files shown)


core-code:
@csi-adziahel: 0.9.1 is the latest version that is linked for download on the homepage. what is the rationale for updating to version 0.9.2 instead?

csi-adziahel:
@core-code see, when I've submitted 0.8.0 &rarr; 0.9.1, it wasn't linked at the time as well, so it made no difference what official non-advertised version to update to. Makes sense?

Feel free to issue a rollback PR anyway.

vitorgalvao:
> when I've submitted 0.8.0 → 0.9.1, it wasn't linked at the time as well, so it made no difference what official non-advertised version to update to. Makes sense?

If the version deemed as stable on the website was `0.8.0`, submitting an update to `0.9.1` or `0.9.2` (really anything other than `0.8.0`) would be equally wrong.

There have been too many of these slipping through. We clearly need a better method to stop non-stable versions from slipping into these repos.

PR #49357, opened 2018-07-06
Update navicat-premium appcast

brianbrownton:


brianmorton:
I  have updated the appcast, as the old one used version interpolation which I have been previously advised to be not very useful in an appcast (how will you know when a new version is out?).

There are a bunch of other `navicat` casks that I can do this for, but I want to see if this is ok before I go to the trouble of doing all the others.

reitermarkus:
Yes, seems sensible that appcasts should not depend on the version, at least not minor versions.

commitay:
See https://github.com/Homebrew/homebrew-cask/pull/49362#issuecomment-403164853

PR #49362, opened 2018-07-06
update navicat appcasts

brianbrownton:
per https://github.com/Homebrew/homebrew-cask/pull/49357#issuecomment-403124726


commitay:
I switched the casks to use the appcast rather than the changelog as it sometimes doesn't match what version the actually download is.

vitorgalvao:
@commitay So you mean we should revert this?

brianmorton:
I would prefer no appcast over a bad appcast.

PR #49363, opened 2018-07-06
Update navicat-data-modeler to 2.1.18

brianbrownton:

### DIFF
--- a/Casks/navicat-data-modeler.rb
+++ b/Casks/navicat-data-modeler.rb
@@ -1,5 +1,5 @@
 cask 'navicat-data-modeler' do
-  version '2.1.17'
+  version '2.1.18'
   sha256 '94fb444b011d77d634f08d650bd636b6a726b6818ebb7a5ccbc2dc44044ebd41'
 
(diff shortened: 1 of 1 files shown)


brianmorton:
Not sure how I feel about this PR... the interpolation for the download only takes the major and minor (`21`) and the sha256 is unchanged, however their changelog says this is `2.1.18`... but if the binary is the same as `2.1.17` what does that mean? The content is dynamic? The cask/sha256 we had for `2.1.17` is really from `2.1.18`?

commitay:
IIRC this never matches the changelog, what is the version of the app itself?

brianmorton:
That would require me actually installing this... :P 

Give me a minute.

brianmorton:
image

brianmorton:
I think what was submitted as 2.1.17 was actually 2.1.18, based on this:
image

This is about a week after the modified date on the 2.1.18 .app - so I think what we thought was 2.1.17 was actually this binary (2.1.18) and that is why the checksum didn't change.

Thoughts?

commitay:
> Thoughts?

Yeah, Navicat casks are really annoying. I thought I had checked the version number locally when I submitted that PR but I guess I must have forgotten.

PR #49380, opened 2018-07-07
Update subsurface to 4.8.0

brianbrownton:

### DIFF
--- a/Casks/subsurface.rb
+++ b/Casks/subsurface.rb
@@ -1,6 +1,6 @@
 cask 'subsurface' do
   version '4.8.0'
-  sha256 'b3b62120bb3c90e378d3b574ca8274b8ad474352c6b32d68fdf28699e2478d5a'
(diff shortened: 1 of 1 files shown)


brianmorton:
I got an email from the devs saying they did an update in place on the binary which is why the sha256 changed. I will reply back asking them to confirm in this PR the new checksum.

dirkhh:
I'm the Subsurface maintainer and I'm the one creating our binary releases. Unfortunately I had a script bug which caused the original 4.8.0 Mac DMG of Subsurface to be built against the wrong version of libdivecomputer (which actually is a submodule). So technically the original 4.8.0 Mac DMG wasn't really 4.8.0 :-(
Because of that I recreated the DMG and replaced it without bumping the version number. Bad style - I know. We are working on a 4.8.1 that addresses some of the other bugs that we have found so far.

Anyway, I can confirm that 51fe5c49eb726f5eda780cc3e845574648c3521b350a56d9d6574d035cabf3fc is the correct SHA256 for the official Subsurface 4.8.0 DMG

reitermarkus:
Thanks!

PR #49427, opened 2018-07-09
Update resolume-arena to 6.0.10

brianbrownton:

### DIFF
--- a/Casks/resolume-arena.rb
+++ b/Casks/resolume-arena.rb
@@ -9,9 +9,9 @@ cask 'resolume-arena' do
 
   pkg 'Resolume Arena Installer.pkg'
 
-  uninstall pkgutil: [
-                       "com.resolume.pkg.ResolumeArena#{version.major}",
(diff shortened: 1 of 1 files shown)


brianmorton:
I'm a little concerned about the uninstall stanza for this and [resolume-avenue](https://github.com/Homebrew/homebrew-cask/pull/49428), as there seem to be two shared pkgs that both use. If both casks have them in the uninstall stanza, however, and a user has both apps/casks installed but then removes one it would break the other.

I believe these are the shared packages:
```
> com.resolume.pkg.ResolumeDXV
> com.resolume.pkg.ResolumeQuickLook
```

Thoughts?

vitorgalvao:
The solution is to remove the shared packages from `uninstall` and add them to `zap`.

vitorgalvao:
> Maybe travis doesn't do zap?

It used to, but we removed that option because it caused trouble in some situations.

brianmorton:
bump

brianmorton:
bump

Issue #49640, opened 2018-07-13
android-platform-tools links sqlite3 binary (macOS-provided software) without readline support


marlencrabapple:
After months of dealing with a sqlite3 without readline that I assumed was the default macOS binary, I finally ran `which sqlite3` on it and discovered it was linked from a Cask. Sometime before this I tried to link a sqlite3 from regular Homebrew and got the following warning:

> Warning: Refusing to link macOS-provided software: sqlite

Fast forward to when I found out the sqlite3 binary was from a Cask, I put 2+2 together and began to think maybe android-platform tools, the Cask in question shouldn't link sqlite3.

I also noticed a similar issue (#31797) in which a pull request that contained an updated Cask formula didn't link the sqlite3 binary, so I figured I should try and do the same for android-platform-tools.

Hopefully I'm doing this correctly and not breaking any rules 😬

commitay:
Please don't open an issue if you are also submitting a PR for it.

Issue #31797, opened 2017-04-03
Android SDK Install Fails


datwheat:
#### Description of issue

I'm trying to install android-sdk with the new Cask, but it fails because of sqlite3.

#### Output of your command with `--verbose --debug`

```
brew cask reinstall android-sdk --verbose --debug
==> Printing caveats
==> Caveats
We will install android-sdk-tools, platform-tools, and build-tools for you.
You can control android sdk packages via the sdkmanager command.
You may want to add to your profile:
  'export ANDROID_HOME=/usr/local/share/android-sdk'

==> Hbc::Installer#fetch
==> Downloading
==> Downloading https://dl.google.com/android/repository/tools_r25.2.3-macosx.zip
Already downloaded: /Users/datwheat/Library/Caches/Homebrew/Cask/android-sdk--25.2.3.zip
==> Downloaded to -> /Users/datwheat/Library/Caches/Homebrew/Cask/android-sdk--25.2.3.zip
==> Verifying download
==> Determining which verifications to run for Cask android-sdk
==> Checking for verification class Hbc::Verify::Checksum
==> 1 verifications defined
Hbc::Verify::Checksum
==> Running verification of class Hbc::Verify::Checksum
==> Verifying checksum for Cask android-sdk
==> SHA256 checksums match
==> Hbc::Installer#uninstall
==> Un-installing artifacts
==> Determining which artifacts are present in Cask android-sdk
==> 3 artifact/s defined
#<Hbc::Artifact::PreflightBlock:0x007f84d98e5c78>
#<Hbc::Artifact::Binary:0x007f84d98e5b10>
#<Hbc::Artifact::PostflightBlock:0x007f84d98e5a48>
==> Un-installing artifact of class Hbc::Artifact::PreflightBlock
==> Un-installing artifact of class Hbc::Artifact::Binary
==> Un-installing artifact of class Hbc::Artifact::PostflightBlock
==> Purging files for version 25.2.3 of Cask android-sdk
==> Purging all staged versions of Cask android-sdk
==> Hbc::Installer#stage
==> Extracting primary container
==> Determining which containers to use based on filetype
==> Checking container class Hbc::Container::Pkg
==> Checking container class Hbc::Container::Ttf
==> Checking container class Hbc::Container::Otf
==> Checking container class Hbc::Container::Air
==> Checking container class Hbc::Container::Cab
==> Checking container class Hbc::Container::Dmg
[...]

PR #49643, opened 2018-07-13
android-platform-tools: remove sqlite3 symlink

marlencrabapple:
Moved from issue https://github.com/Homebrew/homebrew-cask/issues/49640

After months of dealing with a sqlite3 without readline that I assumed was the default macOS binary, I finally ran `which sqlite3` on it and discovered it was linked from a Cask. Sometime before this I tried to link a sqlite3 from regular Homebrew and got the following warning:

> Warning: Refusing to link macOS-provided software: sqlite

Fast forward to when I found out the sqlite3 binary was from a Cask, I put 2+2 together and began to think maybe android-platform tools, the Cask in question shouldn't link sqlite3.

I also noticed a similar issue (#31797) in which a pull request that contained an updated Cask formula didn't link the sqlite3 binary, so I figured I should try and do the same for android-platform-tools.

Hopefully I'm doing this correctly and not breaking any rules 😬

### DIFF
--- a/Casks/android-platform-tools.rb
+++ b/Casks/android-platform-tools.rb
@@ -13,5 +13,4 @@ cask 'android-platform-tools' do
   binary "#{staged_path}/platform-tools/fastboot"
   binary "#{staged_path}/platform-tools/hprof-conv"
   binary "#{staged_path}/platform-tools/mke2fs"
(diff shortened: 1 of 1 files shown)


commitay:
Thanks for the PR!

PR #49713, opened 2018-07-16
fix formatting in bug template

yurikoles:
There is some typo.


reitermarkus:
Thanks!

commitay:
I've always thought it was intentional, @vitorgalvao?

vitorgalvao:
Yes, it is intentional. It’s a sub-item of the previous one and so it is indented.

PR #49870, opened 2018-07-20
Update popo to 2.6.7

SteamedFish:

### DIFF
--- a/Casks/popo.rb
+++ b/Casks/popo.rb
@@ -1,8 +1,8 @@
 cask 'popo' do
-  version '2.6.7'
+  version '2.6.7,5615'
   sha256 'a24c516110a1aafc329e909f4e566cb6d1674ef4724649d0ac750b6ce691cd7f'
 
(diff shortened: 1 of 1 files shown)


SteamedFish:
popo sometimes updates its dmg (and build number) without update its version number. 

I don't know if it is the correct way to submit like this.

vitorgalvao (changes requested):

SteamedFish:
popo doesn't have Signature Info section. Is this means that the only way to verify is to get developer's confirmation?

popo updates its build pretty frequently, but it does not update its version number very often. 

does this means that i have to get developer's confirmation everytime popo update its build number？

the build number is at popo's about menu. There's no build number on the download URL。can I include the build number as part of the version number？

vitorgalvao:
> can I include the build number as part of the version number？

Yes.

SteamedFish:
I have updated my pull requests with build number included.
popo updated with a new build number today, so I have updated with a new sha256.

PR #49896, opened 2018-07-20
Update mirador to 1.5.1

brianbrownton:


brianmorton:
This one is a little interesting - it seems they may have changed their way of naming releases to no longer use a version string - or it may just be a one time thing for this release.

If the next update still uses `latest-macos.zip` I would probably suggest updating to be `:latest` `:no_check`

brianmorton:
It seems this is source code and not a binary - I will follow up with them to see if this is an error.

brianmorton:
This should be good to merge now.

PR #49916, opened 2018-07-22
Update musicbrainz-picard to 2.0.1

jomo:
Also upgraded URL to https

### DIFF
--- a/Casks/musicbrainz-picard.rb
+++ b/Casks/musicbrainz-picard.rb
@@ -2,8 +2,7 @@ cask 'musicbrainz-picard' do
   version '2.0.1'
   sha256 '2a652ff5c562216033568cb33d0a2525cfd3043f43c67afdefaf4f8cd471cd5a'
 
-  # ftp.osuosl.org was verified as official when first introduced to the cask
-  url "https://ftp.osuosl.org/pub/musicbrainz/picard/MusicBrainz-Picard-#{version}.dmg"
+  url "http://ftp.musicbrainz.org/pub/musicbrainz/picard/MusicBrainz-Picard-#{version}.dmg"
   appcast 'https://picard.musicbrainz.org/changelog/'
   name 'MusicBrainz Picard'
   homepage 'https://picard.musicbrainz.org/'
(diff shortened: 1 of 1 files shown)


Casks/musicbrainz-picard.rb
```diff
@@ -1,8 +1,9 @@
 cask 'musicbrainz-picard' do
-  version '2.0'
-  sha256 'a8ae8f9d863bbf7ae244762d986b45c3f62e57a3ff5f3776992db0a4b4defc35'
+  version '2.0.1'
+  sha256 '2a652ff5c562216033568cb33d0a2525cfd3043f43c67afdefaf4f8cd471cd5a'
 
-  url "http://ftp.musicbrainz.org/pub/musicbrainz/picard/MusicBrainz-Picard-#{version}.dmg"
+  # ftp.osuosl.org was verified as official when first introduced to the cask
+  url "https://ftp.osuosl.org/pub/musicbrainz/picard/MusicBrainz-Picard-#{version}.dmg"
```
commitay: Why is the `url` changed?

jomo: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#https-urls-are-preferred

"If available, an HTTPS URL is preferred. A plain HTTP URL should only be used in the absence of a secure alternative."

jomo: Also, it's the same server. ftp.musicbrainz.org is just an alias for ftp.osuosl.org.

commitay: > "If available, an HTTPS URL is preferred. A plain HTTP URL should only be used in the absence of a secure alternative."

Yes this is correct. If `https://ftp.musicbrainz.org/` worked we would prefer it but it it doesn't. We to avoid using the verified comments unless we have too.

jomo: In that case I suggest you update the documentation, clarifying that secure alternatives are not preferred if the domain name changes. (Which I find a bit weird here, given that it's not a mirror).

FWIW musicbrainz.osuosl.org also has working HTTPS if that's any better.

Issue #49929, opened 2018-07-23
Licecap installation fails: ` Directory not empty @ dir_s_rmdir - /var/folders/4d/d_nhdwh14m1f6pf07417q4kw0000gn/T/d20180723-49512-zd6fma`


Taytay:
I did not see this problem listed. If I missed the template, feel free to close.

I just installed Home-brew tonight. Other casks are working great. 
This is the error I see:
```
Installing licecap cask. It is not currently installed.
==> Satisfying dependencies
==> Downloading https://www.cockos.com/licecap/licecap128.dmg
Already downloaded: /Users/taytay/Library/Caches/Homebrew/Cask/licecap--1.28.dmg
==> Verifying checksum for Cask licecap
==> Installing Cask licecap
==> Purging files for version 1.28 of Cask licecap
Error: Directory not empty @ dir_s_rmdir - /var/folders/4d/d_nhdwh14m1f6pf07417q4kw0000gn/T/d20180723-49512-zd6fma
Follow the instructions here:
  https://github.com/Homebrew/homebrew-cask#reporting-bugs
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1446:in `rmdir'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1446:in `block in remove_dir1'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1457:in `platform_support'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1445:in `remove_dir1'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1438:in `remove'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:780:in `block in remove_entry'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1495:in `ensure in postorder_traverse'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1495:in `postorder_traverse'
[...]

Issue #48862, opened 2018-06-22
Scraping shasum-verification policy


vitorgalvao:
So far, I’d call the policy of requiring verification for `sha256`-only changes a monumental failure. It places a burden on maintainers and contributors that beared no fruit. Everyone is annoyed by it and no one wants to do it.

It’s also easily circumventable, as an attacker that replaces a binary needs only change its version on the download page. Version bumps not initiated by a developer might be easy to catch in an open-source world, but with the tools we deal with a simple change to the HTML of the download page could go unnoticed indefinitely.

By now it looks little less than a rushed policy taken out of proportion because of a single occurrence. It’s not like shasums are a meaningful way of providing security, anyway. At best they provide a check for verifying a correct download.

So how do we fix this?

Codesigning by itself doesn’t save a user from being infected (see Transmission’s case in 2016), but it can prevent the infection from spreading too far (same case, after Apple revoked the certificate). It’s also currently our preferred method of verification — we ask users to upload the file to VirusTotal but what we care about is the codesigning verification it provides. While not a perfect method, it’s the best we have.

It’s impossible for us to know when an app was tampered with and we need to set expectations accordingly. Just like we have `--require-sha` for security-minded users, we also need to make it their decision the risk of accepting or refusing non-codesingned apps.

macOS already provides an acceptable way of warning users the application they downloaded may not be trustworthy, by means of quarantining. The problem with that method is quarantining needs to be set by the downloading app. Since `curl` doesn’t do it, no apps we download are quarantined and users never get a warning.

Quarantining [is a desirable feature](https://github.com/Homebrew/homebrew-cask/issues/22388), but unfortunately Apple does not provide a simple way to set a download as quarantined. @reitermarkus [started work on this](https://github.com/reitermarkus/quarantine), though I’m unsure it it’s already usable.

So my proposal is to scrape the policy and finish what needs to be done to allow us to quarantine downloads (ideally releasing it as a stand-alone CLI tool that other projects can benefit from as well). We may also add a `--no-quaratine` option for users that don’t care

Pinging @Homebrew/cask, but everyone is welcome to comment.

amyspark:
In re quarantining - I believe implementing it in Ruby may be too prone to breakage (manually handling the SQLite database and the extended attribute led me to a broken policy db when I dealt with @reitermarkus's patch).

[...]

PR #49931, opened 2018-07-23
readme: update URL for reporting bugs

commitay:
Currently this sends people to the blank template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ If your issue persists, search for it before opening a new one. If you find an o
   * [Unlisted reason](doc/reporting_bugs/a_cask_fails_to_install.md#unlisted-reason)
 * [`uninstall` wrongly reports cask as not installed](doc/reporting_bugs/uninstall_wrongly_reports_cask_as_not_installed.md)
 * [`Error: Unknown command: cask` error](doc/reporting_bugs/error_unknown_command_cask.md)
-* [My problem isn’t listed](https://github.com/Homebrew/homebrew-cask/issues/new)
+* [My problem isn’t listed](https://github.com/Homebrew/homebrew-cask/issues/new?template=01_bug_report.md)
 
 ## Requests
 


vitorgalvao:
That was on purpose. As in “if my problem isn’t listed, it’s something else”. We don’t get many bug reporters that click on that instead. Even https://github.com/Homebrew/homebrew-cask/issues/49929, who looks like it, asserts they’ve seen the template.

commitay:
If there clicking on this link don't we want a full report anyway?

vitorgalvao:
> If they're clicking on this link don't we want a full report anyway?

But we don’t know what they’re reporting, so we don’t know what to ask. They’re not making a bug report or a feature/cask request, so it’s something else we don’t yet know which info we need.

Also, note [the template is not empty](https://github.com/Homebrew/homebrew-cask/issues/new). It’s just that it’s message is hidden, and it’s discouraging people from not using a template:

> It’s extremely unlikely your issue isn’t covered by one the templates. Please go back and use one of them. If you still decide to open an issue without a template and we deem it was covered by one of them, your issue will be closed without comment or review. Repeat offences might get you banned. Those templates exist to help us help you. When you don’t use them, you’re making our work more difficult. Please respect our time. Thank you for your understanding.

commitay:
> But we don’t know what they’re reporting, so we don’t know what to ask.

What else would they be reporting? The cask URL that is printed with the error points to this section which has various options and then "My problem isn’t listed" which takes them to the "go back and try again" message.

vitorgalvao:
> What else would they be reporting?

Something not related to code. Asking a question about the our policies/documentation or making a suggestion like https://github.com/Homebrew/homebrew-cask/issues/48862.

> takes them to the "go back and try again" message, which doesn't link to the bug report template.

Good point. [Submitted a change](https://github.com/Homebrew/homebrew-cask/pull/49936) to link them back directly from the message and not hide the message.

commitay:
> Something not related to code. Asking a question about the our policies/documentation or making a suggestion like #48862.

The heading for this section is "Reporting bugs"?

vitorgalvao:
> The heading for this section is "Reporting bugs"?

Ignore! I didn’t notice you were changing the README, I thought you were changing the default template.

Yes, this looks good.

PR #49935, opened 2018-07-23
PULL_REQUEST_TEMPLATE.md: confirm version is stable

vitorgalvao:
There have been too many non-stable versions slipping through to non-versions repos.

I’m not sure this checkbox will be effective at all, especially for `cask-repair` submissions (will leave unchecked). Maybe `cask-repair` will need to be modified to make that question as well (and giving a flag for override).

Naturally, in [Homebrew/homebrew-cask-versions](https://github.com/Homebrew/homebrew-cask-versions) it would be left unchecked.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,6 +5,7 @@ After making all changes to the cask:
 - [ ] `brew cask audit --download {{cask_file}}` is error-free.
 - [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
 - [ ] The commit message includes the cask’s name and version.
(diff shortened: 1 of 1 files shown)


commitay:
TBH, I don't think it will have any effect at all.

For manually submitted PRs:
1) People check it, it's a stable version and we check anyway.
2) People check it, it isn't a stable version and we check anyway.
3) People don't check it, it's a stable version and we check anyway.
4) People don't check it, it isn't a stable version and we check anyway
5) It isn't submitted because it isn't a stable version.

I don't really object to it but it seems like something that just going to end up annoying _us_ when it isn't used correctly like the `sha256` change checkbox.

vitorgalvao:
I was thinking more like:

* People see the box and pay better attention, reducing our need to check. This will also reduce the number of (wrong) PRs.
* People ignore repeatedly, at which point they’re asked to correct behaviour or get banned.

brianmorton:
I would support greater consistency/transparency in these rules as they seem to change depending on who is reviewing PRs.

vitorgalvao:
@brianmorton I’d say we’re all on the same page. When an unstable version slips through it’s due to distraction/error rather than a different interpretation of the rules.

commitay:
Let's try it for now and we can discuss if we think it's working/worthwhile in a month or so.

commitay (approved):

PR #49979, opened 2018-07-23
Create haptickey.rb

chriscn:
Before I am tried to do this locally and it hasn't gone very well so this was all done through GitHub web. I can't see anything wrong so can someone run me through the steps to audit this?

### DIFF
--- a/Casks/haptickey.rb
+++ b/Casks/haptickey.rb
@@ -1,9 +1,10 @@
 cask 'haptickey' do
-  version 0.44
+  version '0.44'
   sha256 :no_check
   
-  url: "https://github.com/niw/HapticKey/releases/download/{version}/HapticKey.app.zip"
-  name: "haptickey"
-  homepage: "https://github.com/niw/HapticKey"
-  
+  url: 'https://github.com/niw/HapticKey/releases/download/{version}/HapticKey.app.zip'
+  name: 'haptickey'
+  homepage: 'https://github.com/niw/HapticKey'
+
+  app: 'HapticKey.app"
 end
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Hello @chriscn,

Have you checked the [`adding-a-cask` document](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md)? Pretty much everything about this cask is wrong. `version`’s value isn’t quoted, it has `:` in stanzas, there’s no `sha256`, there’s nothing being installed, it uses double quotes instead of single quotes.

Check the linked document as it will help you making a cask from scratch. Give it another go and we’ll continue from there.

vitorgalvao:
Didn’t mean to close.

chriscn:
@vitorgalvao thank you for your reply and having read that document it has helped a lot! Currently improving it and quick question is there anyway I can test a brew cask on its own? Don't really want to override my local version of brew.

vitorgalvao:
You can clone the project and then call that specific `brew`. But for this it’s easier create your cask file and then `brew cask {whatever} /path/to/your/cask`.

chriscn:
@vitorgalvao follow your suggestion and fixed it. Thank you for your help.

vitorgalvao:
* You’re still adding a `:` to stanzas. Don’t.
* `sha256` needs to be set. Run `shasum -a 256` on the downloaded zip.
* Try to add an [`appcast`](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

chriscn:
Appcast is there, as well as the hash. I am unsure what you mean when you say stanzas.

chriscn:
I think the reason why the build is failing is due to trying to get the version into the url and I am not sure how to do that because I don't really program in Ruby. If it is unknown how to fix it the PR can just be closed.

vitorgalvao (approved):

PR #49984, opened 2018-07-23
Add osquery

szhu:

[closed issues]: https://github.com/Homebrew/homebrew-cask/issues?q=is%3Aissue+is%3Aclosed

### DIFF
--- a/Casks/osquery.rb
+++ b/Casks/osquery.rb
@@ -1,0 +1,12 @@
+cask 'osquery' do
+  version '2.7.0'
+  sha256 'e5eaf91b5582252f136476391cdd17985cc614536f964f38ddf7eb9cb69d0213'
+
+  url "https://pkg.osquery.io/darwin/osquery-#{version}.pkg"
+  name 'osquery'
+  homepage 'https://osquery.io/'
+
+  pkg "osquery-#{version}.pkg"
+
+  uninstall pkgutil: 'com.facebook.osquery'
+end


commitay:
This already exists as a formula: [`osquery`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/osquery.rb)

commitay:
> osquery's binary package, on the other hand, takes seconds to install and has zero dependencies.

Because it bundles a _limited_ set of the same dependencies as the formula.

https://osquery.io/downloads/official/3.2.6 lists the formula as a way of installing `osquery`, I don't think that we should duplicate it as a cask.

szhu:
Are you sure that's not circular reasoning? I think it might be possible that the official site lists the formula as the official way to install because the cask does not exist yet!

Also, note that installing from Homebrew is listed as an "alternative install option". Which seems to imply that the `.pkg` is the preferred way to install. It would be great if installing from Homebrew could be the top thing listed on the page!

commitay:
> Are you sure that's not circular reasoning? I think it might be possible that the official site lists the formula as the official way to install because the cask does not exist yet!

> Also, note that installing from Homebrew is listed as an "alternative install option". Which seems to imply that the .pkg is the preferred way to install. It would be great if installing from Homebrew could be the top thing listed on the page!

`osquery` upstream has made changes to their codebase to improve the formula and has sent PRs to apply those chnages downstream.

> Also, sure, the cask has no additional features over the formula, but it does have the advantage of having 22 less dependencies.

Yes, but the cask would actually have _less_ features as not all of the dependencies are bundled.

> I would be okay with not including the cask if there was a bottle available for osquery, but when I tried installing it, it did not pour from a bottle.

It does have a bottle so I'm not sure why it isn't working for you.

szhu:
@commitay, I appreciate your thoughtful replies, and I understand the rationale behind most of these decisions. Let me know, though, (1) if you could have another maintainer also take a look at this issue as a second opinion and/or (2) if it's possible for you to try installing the formula and cask yourself and seeing the significant difference in speed for yourself. I'll also file an issue with the osquery team to see what they think.

---

I understand that github issues aren't particularly efficient for repeated back-and-forth discussion, but if my replies below are relevant, please let me know what you think.

>> Also, note that installing from Homebrew is listed as an "alternative install option". Which seems to imply that the .pkg is the preferred way to install. It would be great if installing from Homebrew could be the top thing listed on the page!
>
> osquery upstream has made changes to their codebase to improve the formula and has sent PRs to apply those changes downstream.

That makes it an official install option, but it still doesn't appear to be the preferred install option.

>> Also, sure, the cask has no additional features over the formula, but it does have the advantage of having 22 less dependencies.
>
> Yes, but the cask would actually have less features as not all of the dependencies are bundled.

https://osquery.io/downloads doesn't imply this though, so I feel like the osquery team thinks that the `.pkg` is good enough.

>> I would be okay with not including the cask if there was a bottle available for osquery, but when I tried installing it, it did not pour from a bottle.
>
> It does have a bottle so I'm not sure why it isn't working for you.

You are correct, it is actually pouring from a bottle. The fact that it is pouring 22 bottles before this one still makes installation slow.

reitermarkus:
If the software is open source and CLI-only, it should be a formula. That’s just a line we have to draw, otherwise we end up with duplicates.

You can always set up your own tap and host this as a cask.

PR #50144, opened 2018-07-26
Add Esperanto language for Firefox cask (61.0.1)

neoeinstein:
Mozilla offers an Esperanto-language version of Firefox.

Mozilla ofertas Esperantan linvgan version de Firefox.

### DIFF
--- a/Casks/firefox.rb
+++ b/Casks/firefox.rb
@@ -21,6 +21,11 @@ cask 'firefox' do
     'en-US'
   end
 
+  language 'eo' do
+    sha256 'cf6be4d799cf6a066c59c7c226fc96816ccac454c3adeaec8d9f7eb1c41fb19c'
+    'eo'
+  end
+
   language 'es-AR' do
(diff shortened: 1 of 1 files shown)


neoeinstein:
This currently fails because only the sha256 changed, but not the cask version. This is because the cask version is tied to the version of Firefox, and this only adds access to an additional language. The normal audit command is indeed error-free, so I propose that this PR _should_ be accepted despite the reported failure.

neoeinstein:
Fixed by using a new version and updating the cask to support the non-standard version upgrade

vitorgalvao:
> Fixed by using a new version and updating the cask to support the non-standard version upgrade

Please don’t. Don’t complicate the cask just to make CI pass. We’re not automatons, we can make decisions over the CI and yes we will accept it just with the added language. Please revert this to your previous commit.

neoeinstein:
@vitorgalvao Completed the reversion. Thanks. I wasn't sure how strictly to conform to CI rules.

vitorgalvao:
Thank you.

PR #50111, opened 2018-07-26
Update a-better-finder-rename to 10.31

brianbrownton:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #50147, opened 2018-07-26
Revert "Update a-better-finder-rename to 10.31"

victorpopkov:
Reverts Homebrew/homebrew-cask#50111

Just to be safe.


reitermarkus:
Only the major version is used in the URL, so this doesn‘t help.

victorpopkov:
Yes, you are completely right.

PR #50276, opened 2018-07-31
Update photo-supreme-single-user to 4.2

core-code:


core-code:
since the file is updated in-place (for a given major number) and we don't have an appcast, shouldn't the SHA be removed?

vitorgalvao:
This is one of those “if it starts becoming an issue, we’ll remove. If not, no harm in having the extra verification”.

PR #50321, opened 2018-08-01
Update cctalk to 7.3.1,692

brianbrownton:


brianmorton:
I realize this is quite a big version bump. This is what was being offered on the homepage/download page.

vitorgalvao:
`app` applies to moving an app bundle in the download. Since it’s now a pkg, We need to use the `pkg` stanza to install.

Issue #48383, opened 2018-06-13
CI testing command


commitay:
Somewhat related to the proposed changes in https://github.com/Homebrew/homebrew-cask/issues/48267.

Our CI scripts have become more complicated in the past year now that we run install/uninstall, `gpg` and "leftovers" checks.

Implementing a ruby CI test command similar to https://github.com/Homebrew/homebrew-test-bot might be useful as we could add more checks and better handle errors.

It should also allow us to readd `travisbuddy` and "tune" it for better error output. e.g  commenting only on errors which should be actionable by the contributor: audit, style, "source not found" and "leftover" issues.

reitermarkus:
I actually have been experimenting on a local branch and integrated all the leakes file checks into `Audit` and added a `--install` option.

commitay:
Can you open a PR or push you changes so we can take a look? My main concern is that audit is already rather complicated. I'd like to break it up into stages with a separate command rather than trying to fit everything into audit.

commitay:
Forgot to include it in the top post but one check I'd like to add is a `postinstall` for pkgs to see if they install an `app`.

commitay:
Found it https://github.com/Homebrew/brew/compare/master...reitermarkus:audit-install

commitay:
Just to elaborate futher on why I'm not keen on adding non-audit checks to audit is we have no control over the test environment. 

Changes made by Travis could potentially break anything at anytime. Also I think the amount of weird errors we encounter and have to work around means we would still need a separate way of _choosing_ what (and when) we would want to run/install/check. Having everything as part of the core `brew cask audit` command means we don't have that ability available as we can't easily add/remove/edit/test the core code.

reitermarkus:
> adding non-audit checks to audit

Right, I hadn't thought about that. Then I think it should be part of `_audit_modified_casks` and not `audit`. And `_audit_modified_casks` should be renamed to `_test_modified_casks` because it would then be extended to do everything that the current Travis script now does, not just `audit`.

commitay:
> Then I think it should be part of `_audit_modified_casks`

Sorry, I actually meant `_audit_modified_casks` when I said `audit`.

reitermarkus:
So where should it be added then?

commitay:
A separate command that lives outside of the `brew` core code and workflow similar to test-bot.

commitay:
Mentioned in https://github.com/Homebrew/homebrew-cask/issues/49200. 

[...]

PR #50358, opened 2018-08-02
Add `brew cask ci` command.

reitermarkus:
This replaces the old Bash scripts with a shiny new Ruby command, which was discussed in https://github.com/Homebrew/homebrew-cask/issues/48383.


commitay:
I was about to leave some comments on your repo. 😆

.travis.yml
```diff
@@ -16,14 +16,39 @@ branches:
...
+install:
+  - |
+    # Force strict error checking.
+    set -o errexit
+    set -o pipefail
+  - |
+    # Update Travis commit range.
+    # This is not normally required but does prevent problems with outdated forks and
+    # deleted casks (see https://github.com/Homebrew/homebrew-cask/pull/43164).
+    BRANCH_COMMIT="${TRAVIS_COMMIT_RANGE##*.}"
```
commitay: Thoughts on using `brew pull` so we can drop all of this?

reitermarkus: Sure, I never thought about that. Although I'd rather stick with the current solution for this PR and leave the trial-and-error of using `brew pull` up to you. 😉

cmd/brewcask-ci.rb
```diff
@@ -0,0 +1,260 @@
...
+        end
+
+        ruby_files_in_wrong_directory = modified_ruby_files - (modified_cask_files + modified_command_files)
+
+        unless ruby_files_in_wrong_directory.empty?
+          raise CaskError, "Casks are in the wrong directory:\n" +
+                           ruby_files_in_wrong_directory.join("\n")
+        end
+
+        if modified_cask_files.count > 1 && pr_author && !maintainers.include?(pr_author)
```
commitay: If not really sure if the benefit on this justifies having a token for it and the extra GitHub module.

cmd/brewcask-ci.rb
```diff
@@ -0,0 +1,260 @@
...
+          return
+        end
+
+        raise CaskError, "Build failed."
+      end
+
+      private
+
+      def step(name, travis_id)
+        Travis.fold travis_id do
```
commitay: Maybe have it fold only if the step doesn't error?

cmd/brewcask-ci.rb
commitay: On the other hand if we're going to have a token we could have it [comment](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils/github.rb#L296) on PRs.

.travis.yml
commitay: 😆 Cool, don't look like it should be too hard to get working, I'll do it after this gets merged.

.travis.yml
```diff
@@ -16,14 +16,39 @@ branches:
...
+    brew update
+  - |
+    # Mirror the repo as a tap.
+    TAP_DIR="$(brew --repository)/Library/Taps/${TRAVIS_REPO_SLUG}"
+    mkdir -p "${TAP_DIR}"
+    rsync -az --delete "${TRAVIS_BUILD_DIR}/" "${TAP_DIR}/"
+    export TRAVIS_BUILD_DIR="${TAP_DIR}"
+    builtin cd "${TRAVIS_BUILD_DIR}"
+
+script: brew cask ci
```
commitay: More specific command perhaps so it's clear to users what it is?

reitermarkus: Not sure what's more fitting than `ci` for a command that runs the continuous integration.

commitay:
Dependencies are being uninstalled after the diff check.
```console
==> Uninstalling Cask lando
==> Running uninstall process for lando; your password may be necessary
==> Uninstalling packages:
io.lando.pkg.lando
==> Purging files for version 3.0.0-beta.47 of Cask lando
Error: Some applications were not uninstalled.
+ /Applications/Docker.app
==> Uninstalling Cask docker
```

cmd/brewcask-ci.rb
```diff
@@ -0,0 +1,267 @@
...
+      end
+
+      private
+
+      def step(name, travis_id)
+        success = false
+        output = nil
+
+        Travis.fold travis_id do
+          print "#{name} "
```
commitay: Can we colorise or otherwise highlight the steps?

https://travis-ci.org/Homebrew/homebrew-cask/builds/411515226#L500

cmd/brewcask-ci.rb
reitermarkus: Apparently the environment variables are only decrypted if the PR comes from the same repo, not when it comes from a fork. I thought they would be decrypted also if the PR author has commit access.

cmd/brewcask-ci.rb
```diff
@@ -0,0 +1,271 @@
...
+            Style.run(path)
+          end
+
+          was_installed = cask.installed?
+          cask_dependencies = CaskDependencies.new(cask).reject(&:installed?)
+
+          checks = {
+            installed_apps: Diffable.new do
+              sleep(5) # Allow `mdfind` to refresh.
+              system_command!("/usr/bin/mdfind", args: ["-onlyin", "/", "kMDItemContentType == com.apple.application-bundle"], print_stderr: false)
```
commitay: Only seen it happen once so far on a longer build, this can pick up apps already installed in the Travis image.

https://travis-ci.org/Homebrew/homebrew-cask/builds/411591672#L678

cmd/brewcask-ci.rb
commitay: 😢 I was already working on adding comments.

builds with modified casks > 1 seem to work?

cmd/brewcask-ci.rb
reitermarkus: Then let's use `find` again.

commitay:
Looks really good, thanks for doing this @reitermarkus! 🥇

PR #50511, opened 2018-08-06
Update navicat-for-mariadb to 12.1.5

vitorgalvao:


brianmorton:
@vitorgalvao How did you find these? The appcast only has 12.1.4 (for this and the others of similar version: mysql, oracle, sql-server, postgres, sqlite)

vitorgalvao:
These are all updated at the same time.

I went to the website and saw that one of them (I forget which) was at `12.1.15`. I checked another one and saw the same. So I ran a loop to update them all.

vitorgalvao:
> I realize i the cask we only use the maj/min and it is currently passing an audit, and both of the last PRs did change the hash...

That’s what triggered it. Someone complained of a sha256 error and I checked the website.

PR #50580, opened 2018-08-08
Update Google Chrome to allow old mac installation

My2ndAngelic:


My2ndAngelic:
Note: the newest version of Chrome will not work with Mavericks and earlier if you download chrome from latest link. It seems that they have move old Chrome version to the 'legacy' download link instead, hence it was added there.

I'm not sure if this will work with older mac since I only have 10.9.5 installed on my machine.

vitorgalvao:
Thank you, but we officially only support the current macOS version and one behind. Some casks have this `if/else`, yes, but we remove it when pertinent.

Having an old version of a web browser is a particular security risk which I’d rather not condone.

As for the submission itself, not the conditional is always done in reverse (the `else` is for the latest versions; the others are `<=`).

Thank you anyway.

PR #50626, opened 2018-08-08
Update intel-power-gadget to 3.5.3,655065

Jahhein:


Jahhein:
Cask-repair made the request even when it failed.

vitorgalvao:
> Cask-repair made the request even when it failed.

What was the command you used?

Jahhein:
`cask-repair -e -i formula`

Jahhein:
It has never asked me. It only asks if the formula file is correct. I already closed it.

vitorgalvao:
> It only asks if the formula file is correct.

That’s the trigger. If you say everything is correct, it submits (the point of `cask-repair` is quick edits submitted fast).

Small note: In this case they’re casks, not formulae (built differently and provide different software in different ways).

PR #50652, opened 2018-08-09
Update qqmusic to 5.5.0.5

wangkezun:


vitorgalvao:
Seeing as `url` doesn’t point to a specific version, this cask needs `version :latest` and `sha256 :no_check`, or it’ll break every time there’s an update upstream. If you want it to be versioned, help us find an [`appcast`](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

Issue #44993, opened 2018-03-20
zap trash: elevated permissions


commitay:
Currently it launches a popup even if `uninstall` prompts for sudo.

Also causes CI failures. 
https://github.com/caskroom/homebrew-cask/pull/45540#issuecomment-378613552
https://travis-ci.org/caskroom/homebrew-cask/builds/355729073#L1929

```
$ brew cask zap virtualbox
==> Implied "brew cask uninstall virtualbox"
==> Running uninstall process for virtualbox; your password may be necessary
==> Running uninstall script VirtualBox_Uninstall.tool
Password:
==>
==> Welcome to the VirtualBox uninstaller script.
==>
==> The following files and directories (bundles) will be removed:
==>     /Users/commitay/Library/LaunchAgents/org.virtualbox.vboxwebsrv.plist
==>     /usr/local/bin/VirtualBox
==>     /usr/local/bin/VBoxManage
==>     /usr/local/bin/VBoxVRDP
==>     /usr/local/bin/VBoxHeadless
==>     /usr/local/bin/vboxwebsrv
==>     /usr/local/bin/VBoxBugReport
==>     /usr/local/bin/VBoxBalloonCtrl
==>     /usr/local/bin/VBoxAutostart
==>     /usr/local/bin/VBoxDTrace
==>     /usr/local/bin/vbox-img
==>     /Library/LaunchDaemons/org.virtualbox.startup.plist
==>     /Library/Python/2.7/site-packages/vboxapi/VirtualBox_constants.py
==>     /Library/Python/2.7/site-packages/vboxapi/VirtualBox_constants.pyc
==>     /Library/Python/2.7/site-packages/vboxapi/__init__.py
==>     /Library/Python/2.7/site-packages/vboxapi/__init__.pyc
==>     /Library/Python/2.7/site-packages/vboxapi-1.0-py2.7.egg-info
==>     /Library/Application Support/VirtualBox/LaunchDaemons/
==>     /Library/Application Support/VirtualBox/VBoxDrv.kext/
==>     /Library/Application Support/VirtualBox/VBoxUSB.kext/
==>     /Library/Application Support/VirtualBox/VBoxNetFlt.kext/
==>     /Library/Application Support/VirtualBox/VBoxNetAdp.kext/
==>     /Applications/VirtualBox.app/
==>     /Library/Python/2.7/site-packages/vboxapi/
==>
==> And the following KEXTs will be unloaded:
[...]

PR #50718, opened 2018-08-12
Improve CI messages.

reitermarkus:
Tell contributors what to add to the cask when there are leaked files.


commitay:
I've noticed in some PRs contributors as adding things that are already in the `zap` stanza to `uninstall` to get it past CI as directed by the messaging.

commitay:
Always running `zap` like we used to might be the better option but we would need to fix https://github.com/Homebrew/homebrew-cask/issues/44993

PR #50882, opened 2018-08-15
Update vuescan 9.6 checksum

pauliusm:
Just updated checksum to match download.


vitorgalvao (changes requested):

reitermarkus:
Seems like it changed again https://github.com/Homebrew/homebrew-cask/pull/51284.

PR #51017, opened 2018-08-20
Update jprofiler to 10.1.2

Maarc:


vitorgalvao (changes requested):

Maarc:
Looks good: https://www.virustotal.com/#/file/49a6e4afb5c4725616770ff47111e041adef9d6a16fa5e11b4ddb845caf47e74

vitorgalvao:
> Looks good

Doesn’t look good, because it’s not signed, which is what we’re looking for. This means we need public confirmation from the developer that the software was modified.

PR #51333, opened 2018-08-27
Update bettertouchtool to 2.608

brianbrownton:

### DIFF
--- a/Casks/bettertouchtool.rb
+++ b/Casks/bettertouchtool.rb
@@ -18,16 +18,16 @@ cask 'bettertouchtool' do
     # bettertouchtool.net/releases was verified as official when first introduced to the cask
     url "https://bettertouchtool.net/releases/btt#{version}_final_10_9.zip"
   else
-    version '2.604'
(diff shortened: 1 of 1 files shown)


brianmorton:
I would love to get some feedback on the following:

The appcast currently lists latest as 2.606 - however the link on the appcast goes to a 404. Sniffing around the site I did find this:

https://bettertouchtool.net/releases/
image

This may be better as an appcast (or at least has the potential to be)... however care would need to be taken. I think the reason the 2.606 on the appcast link is broken is because on that releases page it is clear the filename is `btt2.606_alpha.zip` which is a little worrying as the releases  page didn't indicate it was an alpha...

In summary, I guess what I'm saying is that I think `https://bettertouchtool.net/releases/` would be a better appcast as long as we don't update to anything which says `alpha`, `beta`, or `rc` in the filename.

Thoughts?

vitorgalvao:
Sounds good.

brianmorton:
I have updated this PR with those changes.

I also changed the homepage, as the old homepage was just redirecting to what I have submitted.

PR #51336, opened 2018-08-27
Update mediathekview to 13.2.0

brianbrownton:

### DIFF
--- a/Casks/mediathekview.rb
+++ b/Casks/mediathekview.rb
@@ -1,6 +1,6 @@
 cask 'mediathekview' do
-  version '13.1.1'
-  sha256 '5266e92735d262faaf1705c2c8339936edfc1cd830291928257eb4dd70742e57'
+  version '13.2.0'
+  sha256 '0f2eea2cd6ae188c16ecebe220e372cb27ed2e1e5307c447e9f1d1b060296155'
 
   url "https://download.mediathekview.de/stabil/MediathekView-#{version}.dmg"
   appcast 'https://mediathekview.de/changelog/index.xml'


brianmorton:
this is maybe a better appcast:
https://download.mediathekview.de/stabil/

similarly, the latest release (currently `13.2.0` doesn't get a version number, it's just `latest` on that page) so I have updated the cask accordingly.

mmoehring:
For this app curl get's a HTTP 503 when called from within homebrew cask. It works when curl is called standalone. I opened an issue at the maintainers website/server repository (https://github.com/mediathekview/MediathekViewDE/issues/38) for investigation.

brianmorton:
```
I think homebrew-cask makes too many queries. I limited the download with limits. Have these now increased times. can test again?
```

we can also try using the `:fake` symbol for a UA to get around this... thoughts?

mmoehring:
I did not find the code/value for ``:fake``. Is it randomized for each request or is it a fixed? A fixed UA might soon get the same problem.

brianmorton:
Maintainers: I have tried this with both the `:fake` UA symbol and a real one and both times it was 503. Not sure how to make this work (it doesn't seem to take a referrer or cookies).

reitermarkus:
Wow, this was a hard bug to figure out, but I'm not sure how this would be fixable.

It is failing because we make two requests, one to get the final URL, and one for the actual download.

If I put a `sleep 1` in between these steps, it works.

reitermarkus:
The download is now working, but the checksum changed.

reitermarkus:
> currently `13.2.0` doesn't get a version number

Seems like it does now.

alex1702:
@brianmorton change checksum to `0f2eea2cd6ae188c16ecebe220e372cb27ed2e1e5307c447e9f1d1b060296155`

brianmorton:
dropped the URL change, updated the shasum

PR #51504, opened 2018-08-31
Update doomsday-engine to 2.0.3 (rollback to stable)

brianbrownton:

### DIFF
--- a/Casks/doomsday-engine.rb
+++ b/Casks/doomsday-engine.rb
@@ -1,10 +1,9 @@
 cask 'doomsday-engine' do
-  version '2.1.0-build2774'
-  sha256 '37b4b4ab75dcf3122ebcfd7b4d04122c3a04af0af27f21354b766d1ae808c2f0'
+  version '2.0.3'
+  sha256 '0a6258ed1c061b3a80f1573c7845bce39cd8c8d66ade8d3f29531b7cc658e9d5'
 
-  # sourceforge.net/deng was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


claui:
Thanks @brianmorton!

[The upstream page says 2.1.0 is a nightly, unstable version.](http://dengine.net/macos) Shouldn’t this Cask be downgraded to 2.0.3 instead? And have the nightly 2.1.0 builds over on homebrew-cask-versions?

brianmorton:
the appcast `https://sourceforge.net/projects/deng/rss` didn't indicate anything about nightlies vs stables... so maybe we should roll this back to `2.0.3` AND update the appcast to be http://dengine.net/macos

Thoughts?

claui:
> the appcast `https://sourceforge.net/projects/deng/rss` didn't indicate anything about nightlies vs stables

There’s [a second feed](http://api.dengine.net/1/builds/feed). The summary element contains a (somewhat) parseable stable/unstable status but our appcast feature can’t select elements, or can it?

> update the appcast to be http://dengine.net/macos

I’m worried that the site is updated far too often (see the `time` elements in the page source).

Given the (sometimes) year-long hiatus between stable releases, maybe we should drop the appcast.

brianmorton:
I would prefer to update to the second appcast you listed. "We" (being HBC) don't do anything really with the appcasts, by on my side for the automation I do extensive stuff with them... with the eventuality being automatic updates (customized to each cask).

vitorgalvao:
> but our appcast feature can’t select elements, or can it?

Nope. It’s basically just a string.

claui:
Thanks @brianmorton and @vitorgalvao!

PR #51526, opened 2018-08-31
Update manager to 18.8.85

core-code:

### DIFF
--- a/Casks/manager.rb
+++ b/Casks/manager.rb
@@ -1,6 +1,6 @@
 cask 'manager' do
-  version '18.8.81'
-  sha256 'f00db6ff59f94c35133e3049f6dc1641cb3b69df8e2921265d3d2076e4401a30'
+  version '18.8.85'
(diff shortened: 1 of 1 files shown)


brianmorton:
https://github.com/Homebrew/homebrew-cask/pull/51510

adidalal:
Duplicate, thanks anyway @core-code !

core-code:
@brianmorton: i have a cronjob-observation-script  running for the manager cask so that i can send the updates. however, you seem to do this very reliably already and i don't think there is any need to duplicate our efforts. can i turn off the script i have in place for the manager cask?

brianmorton:
@core-code i am slowly moving towards something similar for each cask. i usually do the updates weekday mornings between 9am-10am PT. having said that, however, I will not be contributing next week so it may be good for us to keep doing this for a little while longer ;)

core-code:
@brianmorton: ok thanks ;)

PR #51579, opened 2018-09-03
Update adobe-acrobat-reader to 18.011.20058

core-code:


core-code:
maybe 'http://get.adobe.com/reader/' should be added as an appcast since it contains the version number?

PR #51587, opened 2018-09-03
README: Homebrew updates

MikeMcQuaid:
- `Homebrew-Cask` reads really weirdly in English so let's remove the dash.
- Use https://brew.sh for SSL goodness
- It's not really "Homebrew-style" as it's part of Homebrew itself now

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
-# Homebrew-Cask
+# Homebrew Cask
 
 _“To install, drag this icon…” no more!_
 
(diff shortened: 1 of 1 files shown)


amyspark (approved):

vitorgalvao:
I’d rather `Homebrew-cask` or `HomebrewCask`, even, or `Homebrew/Cask` (might be ambiguous) but I’m not going to fight anyone on this choice. I agree the current naming is strange, and the only times I don’t feel weird typing it are `homebrew-cask` and `HBC`. By this point, even `Caskroom` would be the superior choice, but that can lead to even more confusion.

PR #51642, opened 2018-09-05
Update cleanmymac to 4.0.1,1536142248

toonetown:

### DIFF
--- a/Casks/cleanmymac-x.rb
+++ b/Casks/cleanmymac-x.rb
@@ -1,11 +1,11 @@
-cask 'cleanmymac' do
+cask 'cleanmymac-x' do
   version '4.0.1,1536142248'
   sha256 '5aaa5a233884fb02b9c47ccdc30a3546f4a59368a9ba3babe5881be84e922368'
 
   # devmate.com/com.macpaw.CleanMyMac was verified as official when first introduced to the cask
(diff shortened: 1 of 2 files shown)


vitorgalvao (commented):
`uninstall` uses `version.major` extensively. It’s likely to require a revamp.

toonetown:
In my testing, all the files that get written to `/Library/` are named with version "4"...only the app name itself is changed to be named `CleanMyMac X.app` instead of `version.major`

toonetown:
Also, it is my understanding this is an update to the current "CleanMyMac" product...I don't know that the cask should be renamed...

core-code:
not sure on the re-naming either. its a version number, not the product name. also the real version number is 4, the 'X' is just a marketing thing.

reitermarkus:
> the 'X' is just a marketing thing.

They should also have added a notch to their new icon while they were at it.

core-code:
@reitermarkus ha, how true. i've actually launched that thing and was quite appalled. fascinating that this seems to be in spot 10 of the most-used mac apps (i've got some stats there).

toonetown:
So - is there something still waiting on me for this?

fantasticfears:
It's a new version and needs a new license. The BOM indicates the version 4. I think two casks for now would be much better.

fantasticfears:
Good idea. For ref: https://github.com/Homebrew/homebrew-cask-versions/pull/6262

PR #51657, opened 2018-09-05
Update manager to 18.9.14

core-code:

### DIFF
--- a/Casks/manager.rb
+++ b/Casks/manager.rb
@@ -1,6 +1,6 @@
 cask 'manager' do
-  version '18.9.13'
-  sha256 'e583fdf23bd2200655078e842a9517ad5e9c3e28315ddb35a09f94669edd786a'
+  version '18.9.14'
+  sha256 'b6dd9a5b43ddd418b69e46138472600398504be07a426f12be93c24b2325b53e'
 
   # d2ap5zrlkavzl7.cloudfront.net was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


core-code:
actually i'd propose to change that cask to version :latest ...their release schedule is just insane, see here  https://www.manager.io/desktop/releases/

i can keep an automatically updated cask in my local repository but since this is quite an unpopular app having that much work for it is kinda pointless. i'd rather see some more important other casks with version :latest converted to being versioned if they have an appcast than continue updating this one each day

vitorgalvao:
Aren’t these updates scripted?

core-code:
well, i have a script that tells me when its updated and i don't mind sending a pull request. i assume the same is true for @brianmorton 

however, someone has to approve the pull request each day. so i guess there is some overhead. but if everyone is fine with the current situation, no problem

vitorgalvao:
> i assume the same is true for @brianmorton

I think @brianmorton’s script also `cask-repair`s directly.

> however, someone has to approve the pull request each day. so i guess there is some overhead. but if everyone is fine with the current situation, no problem

True on the overhead, but they’re so many daily casks, I don’t think it’ll make that much of a diference. I use [a script for merging](https://github.com/vitorgalvao/tiny-scripts/blob/master/climergebutton) together with a custom Alfred Workflow. When I’m ready to merge, I press a keyboard shortcut that grabs the URL of Chrome’s frontmost tab, closes it, then merges the PR in the background.

core-code:
ok good - a cheers for automation ;) then lets close this in favour of #51707

PR #51690, opened 2018-09-06
Update evergreen to 5.0d3

core-code:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #51703, opened 2018-09-07
restore stable netnewswire and rename new netnewswire

core-code:
i am trying to fix the situation created by '#51690'

i didn't realize that a cask named 'netnewswire' already existed in the repository. therefore a cask for a stable product has been overwritten with a cask for an alpha version of an upcoming new project.

this pull request restores the netnewswire cask to the state is was before #51690 and renames the new cask for netnewswire/evergreen to netnewswire-evergreen

### DIFF
--- a/Casks/netnewswire-evergreen.rb
+++ b/Casks/netnewswire-evergreen.rb
@@ -1,0 +1,14 @@
+cask 'netnewswire-evergreen' do
+  version '5.0d3'
+  sha256 'b0ae90ac9a3e2f01856ed8679aa1fb10993c48d62b59c6230be381a350667702'
+
+  url "https://ranchero.com/downloads/NetNewsWire#{version}.zip"
+  appcast 'https://ranchero.com/downloads/netnewswire-beta.xml'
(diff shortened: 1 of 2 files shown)


vitorgalvao:
If you look at the previous website (http://netnewswireapp.com/), all it has are the announcements of the app reverting to Brent. I don’t even see a download link. That’s why it made sense to me to overwrite the cask with the new in-progress version.

I’m not a NetNewsWire user, though, so I’m open to whatever its users think is vest.

core-code:
i am not a user either. but the download is still available and i get the impression that the new netnewswire/evergreen is not really stable or even  useable yet, so i'd think it would make sense to keep them separate for the time being.

as you say, it would be nice to hear opinions from actual NetNewsWire users.

vitorgalvao:
Lets keep this open for a bit and wait for a complaint.

vitorgalvao:
We’re [already getting other people updating this one](https://github.com/Homebrew/homebrew-cask/pull/51789), let’s keep it.

PR #51770, opened 2018-09-09
Update cajviewer to 2.0

core-code:


core-code:
they still seem to have the old 1.5 version for download, but this new 'cloud' version 2.0 seems to be newer.
its the same app and has the same bundle-identifier, just in a newer version.

PR #51774, opened 2018-09-09
Fix CoolTerm.app not found inside nested .dmg

futek:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [ ] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

The .zip file at url must have changed such that the application is now nested inside a .dmg.

PR #51775, opened 2018-09-09
play: test

commitay:


commitay:
This and https://github.com/Homebrew/homebrew-cask/pull/51774 would seem to be a regression, `dmg` containers used to be extracted automatically.

reitermarkus:
Seems to be because `unzip` doesn't remove skip `__MACOSX` like `ditto` does.

PR #51793, opened 2018-09-10
zoomus: 4.1.31275.0831

jasontedor:
This commit upgrades the Zoom meeting client from version 4.1.27695.0702
to 4.1.31275.0831.


vitorgalvao:
Thank you. Next time please do leave the template in place, even if you don’t fill it.

jasontedor:
Sure thing, thanks for the feedback.

PR #51834, opened 2018-09-11
Update SpiderOakONE to 7.3.0

g-a-c:

### DIFF
--- a/Casks/spideroakone.rb
+++ b/Casks/spideroakone.rb
@@ -14,6 +14,12 @@ cask 'spideroakone' do
 
   pkg 'SpiderOakONE.pkg'
 
-  uninstall quit:    'com.spideroak.orange.client',
+  uninstall signal:  [
+                       ['TERM', 'com.spideroak.orange.client'],
+                       ['QUIT', 'com.spideroak.orange.client'],
+                       ['INT',  'com.spideroak.orange.client'],
+                       ['HUP',  'com.spideroak.orange.client'],
+                       ['KILL', 'com.spideroak.orange.client'],
+                     ],
             pkgutil: 'org.python.python'
 end
--- a/Casks/spideroakone.rb
+++ b/Casks/spideroakone.rb
@@ -14,6 +14,5 @@ cask 'spideroakone' do
 
   pkg 'SpiderOakONE.pkg'
 
-  uninstall quit: 'com.spideroak.orange.client'
-  uninstall pkgutil: 'org.python.python'
+  uninstall quit: 'com.spideroak.orange.client', pkgutil: 'org.python.python'
 end
--- a/Casks/spideroakone.rb
+++ b/Casks/spideroakone.rb
@@ -14,5 +14,6 @@ cask 'spideroakone' do
 
   pkg 'SpiderOakONE.pkg'
 
-  uninstall quit: 'com.spideroak.orange.client', pkgutil: 'org.python.python'
+  uninstall quit:    'com.spideroak.orange.client',
+            pkgutil: 'org.python.python'
 end
--- a/Casks/spideroakone.rb
+++ b/Casks/spideroakone.rb
@@ -16,10 +16,6 @@ cask 'spideroakone' do
 
   uninstall signal:  [
                        ['TERM', 'com.spideroak.orange.client'],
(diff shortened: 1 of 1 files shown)


Casks/spideroakone.rb
```diff
@@ -2,17 +2,18 @@
...
   name 'SpiderOakOne'
   homepage 'https://spideroak.com/'
 
-  depends_on macos: '>= :mountain_lion'
+  depends_on macos: '>= :mavericks'
 
   pkg 'SpiderOakONE.pkg'
 
-  uninstall pkgutil: 'org.python.python'
+  uninstall quit:    'com.spideroak.orange.client',
```
vitorgalvao: If Travis still fails, try [`signal`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#uninstall-key-signal) instead.

Casks/spideroakone.rb
```diff
@@ -2,17 +2,26 @@
...
   name 'SpiderOakOne'
   homepage 'https://spideroak.com/'
 
-  depends_on macos: '>= :mountain_lion'
+  depends_on macos: '>= :mavericks'
 
   pkg 'SpiderOakONE.pkg'
 
-  uninstall pkgutil: 'org.python.python'
+  uninstall signal:  [
```
vitorgalvao: It should be just one (the lowest one that gets the job done).

vitorgalvao (changes requested):

Casks/spideroakone.rb
```diff
@@ -2,17 +2,26 @@
...
+                       ['TERM', 'com.spideroak.orange.client'],
+                       ['QUIT', 'com.spideroak.orange.client'],
+                       ['INT',  'com.spideroak.orange.client'],
+                       ['HUP',  'com.spideroak.orange.client'],
+                       ['KILL', 'com.spideroak.orange.client'],
+                     ],
+            pkgutil: 'org.python.python'
```
vitorgalvao: Is this correct? Why is it uninstalling `python`?

g-a-c: That's the ID that's used in the SpiderOak installer:

```
gac@MBP /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/developer/bin ±master » ./list_ids_in_pkg /Volumes/SpiderOakONE/SpiderOakONE.pkg
org.python.python (+)
```

I'm sure it might be considered a bug in SpiderOak, but that's what it does, so if we want it in Homebrew then that's how it should be. Since Python isn't distributed as a Cask it seems like it should be safe from conflicting with an actual version of Python, which is (I assume) why it was allowed into Homebrew to begin with (I'm not the original committer).

Casks/spideroakone.rb
```diff
@@ -2,17 +2,22 @@
   version :latest
   sha256 :no_check
 
-  if MacOS.version <= :sierra
-    url 'https://spideroak.com/release/spideroak/osx'
-  else
+  if MacOS.version == :high_sierra
```
reitermarkus: Use `<=`.

Casks/spideroakone.rb
```diff
@@ -2,17 +2,22 @@
...
     url 'https://spideroak.com/release/spideroak/osx_hs'
+  else
+    url 'https://spideroak.com/release/spideroak/osx'
   end
   name 'SpiderOakOne'
   homepage 'https://spideroak.com/'
 
-  depends_on macos: '>= :mountain_lion'
+  depends_on macos: '>= :mavericks'
```
reitermarkus: This doesn't match `MacOS.version == :high_sierra` above.

g-a-c: It isn't supposed to match - the SpiderOak download offering is supported on Mavericks and above, but for some reason (which I don't know) they offer one package for 10.9, 10.10, 10.11, 10.12, and 10.14 and a separate package for 10.13 only.

Casks/spideroakone.rb
g-a-c: Please see the other comment - for some reason 10.13 is the only version of macOS that needs this package, both older and newer versions of macOS use the alternative package.

Maybe this would be easier to read with multiple ifs for  "<= el_capitan", "== high_sierra", ">= mojave"?

PR #51909, opened 2018-09-13
Update integrity to :latest

sscotth:

### DIFF
--- a/Casks/integrity.rb
+++ b/Casks/integrity.rb
@@ -1,9 +1,10 @@
 cask 'integrity' do
-  version :latest
-  sha256 :no_check
+  version '8.1.12'
+  sha256 '6ae35493504efc0395c64b0a927eb445907fa6a6195209b99b865d716ff22826'
 
   # peacockmedia.co.uk/integrity was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


sscotth:
Would `:latest` & `:no_check` be preferred here?

sscotth:
I updated the PR just in case

vitorgalvao (approved):
Reinstated appcast, since it still works.

sscotth:
I got an audit warning not to use appcast with `:latest`

```
• audit for integrity: warning
 - Casks with an appcast should not use version :latest
```

https://github.com/Homebrew/brew/blob/ca86f0703982e243e3940b03b32fc91f4a8b7fb8/Library/Homebrew/cask/audit.rb#L213

vitorgalvao:
> I got an audit warning not to use appcast with `:latest`

The cask is not using `:latest`. I think I see where your confusion is coming from, so I’ll clarify. The rule is: if a cask’s `url` is unversioned, it requires `:latest`/`:no_check` (because we can’t know when it changes upstream) *unless* it has an `appcast`, in which case it must have a set `version` and `sha256` (because we can use the `appcast` to know when there are changes upstream).

PR #51913, opened 2018-09-13
add spoilers to bug template

yurikoles:
This adds spoilers to simplify reading of issues. Example: https://github.com/yurikoles/homebrew-cask/issues/1, which is clone of #51907 with use of this template.


vitorgalvao:
Thank you, but this makes reading issues harder. Output is generally short enough to scroll; I don’t want to have to expand three outputs every time.

PR #51924, opened 2018-09-13
delete cisco-spark cask

rajiv:
"Cisco Spark" was renamed to "Webex Teams", and there is already a cask called `webex-teams`.


claui (approved):
Good catch @rajiv!

PR #52046, opened 2018-09-16
Update lingon-x to 6.3.2

suschizu:

### DIFF
--- a/Casks/lingon-x.rb
+++ b/Casks/lingon-x.rb
@@ -1,6 +1,14 @@
 cask 'lingon-x' do
-  version '6.3.2'
-  sha256 'b5f60becd9e5afa2ea8f094c4d0e268b5dd16d6727babbe37c4588569a67dbad'
+  if MacOS.version <= :el_capitan
+    version '4.3.9'
+    sha256 '85d5b7e28f3b404b285d11e35311dbb7d9cc23df20ebb089c08a7483e37e62ef'
+  elsif MacOS.version <= :sierra
+    version '5.2.9'
+    sha256 '7169aaa861626f62112c6ac24256226cd5a5f3b73dade41a4a803232d5942063'
+  else
+    version '6.3.2'
+    sha256 'b5f60becd9e5afa2ea8f094c4d0e268b5dd16d6727babbe37c4588569a67dbad'
+  end
 
   url "https://www.peterborgapps.com/downloads/LingonX#{version.major}.zip"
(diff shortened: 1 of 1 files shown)


commitay:
Previous versions are in https://github.com/Homebrew/homebrew-cask-versions

https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/lingon-x5.rb
https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/lingon-x4.rb
https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/lingon-x2.rb

core-code:
@commitay isn't it much better to have proper version information for different OS versions in the cask? this way a user just needs to 'brew cask install' and the right thing happens automatically instead of having to do a lot of research himself for the latest version.

i think it would be ideal if any cask would install on the last 3 released OS versions. there are quite few apps that raise minimum requirements that fast anyway.

in any case, there are quite a few casks that have different versions for different OS versions. also, there are quite a few old versions stored in 'homebrew-cask-versions'. is there any clear guideline which one is preferred in which case?

commitay:
> isn't it much better to have proper version information for different OS versions in the cask? this way a user just needs to 'brew cask install' and the right thing happens automatically instead of having to do a lot of research himself for the latest compatible version.

No, as it isn't just a "compatible version", they are paid upgrades. Including them all in this cask would mean that a user can't install an older version that they have a license for on a newer version of macOS.

> is there any clear guideline which one is preferred in which case?

Usually for "free" software when there isn't a reason not to install the latest version (unless it isn't compatible) we do include it in the same cask.  Paid upgrades go to `cask-versions`.

core-code:
> Including them all in this cask would mean that a user can't
> install an older version that they have a license for on a newer version of macOS.

well, including them in the main cask wouldn't necessitate deleting the existing casks in homebrew-versions

> Usually for "free" software when there isn't a reason not to install the latest version
> (unless it isn't compatible) we do include it in the same cask.
>  Paid upgrades go to cask-versions.

ok, thanks for the clarification

core-code:
don't want to beat a dead horse here. but i kind of depend on casks being installable all the way back to 10.11. i have to fork casks that don't work on older OS versions. which means the energy we spend on keeping those casks up-to-date won't be of use to anyone else since they are stored only on our server. i'd rather see us contributing to something that is of use to everyone. and i can't see any harm in adding the versions for old OS releases.

if you disagree, just ignore this message.

PR #52071, opened 2018-09-16
Delete privatevpn.rb

core-code:
downloading this cask fails because they employ some download protection. even faking user agent and referer is not enough because they rely on some cookies that the website sets. i see no easy way around this.


vitorgalvao:
> they rely on some cookies that the website sets.

[We can set cookies to](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#additional-https-url-parameters).

core-code:
well yes, but i believe the website will set different cookies all the time and the old ones won't work anymore. but of course it would be good to keep the cask. here is a working curl as of today:


`curl -H 'Host: privatevpn.com' -H 'Cookie: LaSID=eluyffubivumy05tq6c8wzfq0424c; LaVisitorId=uyi7v8d2znqcyq94n2zy7u8684wue; LaVisitorNew=Y; PAPVisitorId=db3642a320f10ba3020X1DnLpKH7mdUA; PHPSESSID=rnob6d8b1bvdbva839emijmra0; discount_code_id=35; cf_clearance=0145cdbd13c6f1fbed87cfab11b74f13d5223f3b-1537130153-2592000-150; __cfduid=d68426f1fd112f4d2a891c73e87e6dfac1537130149' -H 'accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15' -H 'accept-language: en-gb' -H 'referer: https://privatevpn.com/why-privatevpn/view-our-software/' --compressed 'https://privatevpn.com/client/PrivateVPN.dmg'`

core-code:
ok this is the minimal curl i get it to download:
`
curl -H 'Cookie: cf_clearance=0145cdbd13c6f1fbed87cfab11b74f13d5223f3b-1537130153-2592000-150;' -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15' 'https://privatevpn.com/client/PrivateVPN.dmg'`

however, we could also just switch to using the download url from the appcast as this one doesn't seem protected at all

vitorgalvao:
> however, we could also just switch to using the download url from the appcast as this one doesn't seem protected at all

Might be the best choice.

core-code:
see #52157

PR #52349, opened 2018-09-23
conflicts_with.md: add cask

commitay:


reitermarkus:
Should we remove `conflicts_with arch:` and `conflicts_with macos:`?

commitay:
We could probably remove everything except `conflicts_with formula:`.

vitorgalvao:
> We could probably remove everything except `conflicts_with formula:`.

Agreed. We might as well remove the table and introduce a new `formula` section like you did with `cask` and comment there it‘s not yet working.

vitorgalvao (approved):

PR #52449, opened 2018-09-25
Upgrade Plug to latest (2.0.19)

jesse-c:
After making all changes to the cask:

- [x] `brew cask audit --download plug` is error-free.
- [x] `brew cask style --fix plug` reports no offenses.
- [x] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #52480, opened 2018-09-26
Use specific link for Plug upgrade

jesse-c:
- [x] `brew cask audit --download plug` is error-free.
- [x] `brew cask style --fix plug` reports no offenses.

Corrected https://github.com/Homebrew/homebrew-cask/pull/52449.

Edit: I'm a contributor to the Plug project, and got the specific version URL. Due to the sha conflict without this change, it means that the previous link using the version wasn't working.

### DIFF
--- a/Casks/plug.rb
+++ b/Casks/plug.rb
@@ -6,8 +6,8 @@ cask 'plug' do
   else
     version '2.0.19'
     sha256 '418d1adcf0ca099e7d02f9773df133c8c3276f1e65c47ff60f587861fa07d063'
-    url "https://www.plugformac.com/updates/plug#{version.major}/Plug-latest.dmg"
-    appcast "https://www.plugformac.com/updates/plug#{version.major}/sparklecast.xml"
+    url 'https://www.plugformac.com/updates/plug2/Plug-2067.dmg'
+    appcast 'https://www.plugformac.com/updates/plug#{version.major}/sparklecast.xml'
   end
 
   name 'Plug'


Casks/plug.rb
```diff
@@ -6,8 +6,8 @@
   else
     version '2.0.19'
     sha256 '418d1adcf0ca099e7d02f9773df133c8c3276f1e65c47ff60f587861fa07d063'
-    url "https://www.plugformac.com/updates/plug#{version.major}/Plug-latest.dmg"
-    appcast "https://www.plugformac.com/updates/plug#{version.major}/sparklecast.xml"
+    url 'https://www.plugformac.com/updates/plug2/Plug-2067.dmg'
+    appcast 'https://www.plugformac.com/updates/plug#{version.major}/sparklecast.xml'
```
vitorgalvao: Don’t change the quotes. It breaks the interpolation.

Casks/plug.rb
```diff
@@ -6,8 +6,8 @@
   else
     version '2.0.19'
     sha256 '418d1adcf0ca099e7d02f9773df133c8c3276f1e65c47ff60f587861fa07d063'
-    url "https://www.plugformac.com/updates/plug#{version.major}/Plug-latest.dmg"
-    appcast "https://www.plugformac.com/updates/plug#{version.major}/sparklecast.xml"
+    url 'https://www.plugformac.com/updates/plug2/Plug-2067.dmg'
```
vitorgalvao: We can’t use a hidden link other users don’t have access to. Also, by removing the interpolation you’re making this harder to update.

Either make this the official link on the page or update the shasum.

vitorgalvao (changes requested):

jesse-c:
Thanks for the feedback. Will leave it as is for now.

PR #43424, opened 2018-02-01
java 9.0.4 link MacOS

tduehr:
Adds MacOS link to /Library/Java

After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.

PR #52556, opened 2018-09-27
add cask for openjdk 11

ryenus:
With the license change to Oracle JDK, namely the `java` cask,
it makes more sense to use openjdk instead of Oracle JDK.
~See a recent reddit discussion [Do not fall into Oracle's Java 11 trap](https://www.reddit.com/r/programming/comments/9j25je/do_not_fall_into_oracles_java_11_trap/).~
**Edit: see: https://blogs.oracle.com/java-platform-group/oracle-jdk-releases-for-java-11-and-later**, specifically this:
> If you are used to getting Oracle Java SE binaries for free, you can simply continue doing so with Oracle’s OpenJDK builds available at jdk.java.net.  If you are used to getting Oracle Java SE binaries as part of a commercial product or service from Oracle, then you can continue to get Oracle JDK releases through My Oracle Support (MOS), and other locations.


ryenus:
This cask can install openjdk 11 correctly, which is now functionally equivalent to Oracle JDK except the latter has a commercial license.

Feature wise, openjdk is the same, however, many core formula(s) are (hard)coded to `depends_on :java`, so we probably need a way to say this `openjdk` cask also provides `:java` like what the Oracle JDK does. But this is something I'm too dumb to figure out, so someone please help :-)

Another caveat, both Oracle JDK 11 and OpenJDK 11 install to a directory named `jdk-11.jdk`, therefore I have to create the symbol link to openjdk as `/Library/Java/JavaVirtualMachines/openjdk-11.jdk` to avoid conflict/confusion if any.

vitorgalvao:
> many core formula(s) are (hard)coded to `depends_on :java`, so we probably need a way to say this `openjdk` cask also provides `:java`

That’s an issue to be opened in [Homebrew/brew](https://github.com/Homebrew/brew), if you don’t mind. In this repo we have casks (internally different) and those never depend on java (because we haven’t yet added the functionality); they just *tell you* you need java.

Casks/openjdk.rb
```diff
@@ -0,0 +1,28 @@
+cask 'openjdk' do
+  version '11+28'
+  version_parts = version.split('+')
```
vitorgalvao: Don’t. Instead use [the methods we provide](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md#version-methods).

vitorgalvao (dismissed):

vitorgalvao:
Pinging @commitay since you’re dealing with [the other submission](https://github.com/Homebrew/homebrew-cask/pull/52182/).

Casks/openjdk.rb
ryenus: Thanks for the feedback, I was kinda hesitate for that because Oracle/OpenJDK 11 has a non-standard version string `11+28`, in which `+28` kinda indicates the build number, rather than minor/patch:

>$ java -version
openjdk version "11" 2018-09-25
OpenJDK Runtime Environment 18.9 (build 11+28)
OpenJDK 64-Bit Server VM 18.9 (build 11+28, mixed mode)

So which is the preferred way? What about something like `version '11,28'` so that `version.(before/after)comma` can be used?

ryenus:
> > many core formula(s) are (hard)coded to `depends_on :java`, so we probably need a way to say this `openjdk` cask also provides `:java`
> 
> That’s an issue to be opened in [Homebrew/brew](https://github.com/Homebrew/brew), if you don’t mind. In this repo we have casks (internally different) and those never depend on java (because we haven’t yet added the functionality); they just _tell you_ you need java.

What if we change the `java` cask directly to use OpenJDK instead of Oracle JDK?

ryenus:
> Pinging @commitay since you’re dealing with [the other submission](https://github.com/Homebrew/homebrew-cask/pull/52182/).

My bad, that one has/had **OpenJDK 1.8**, I thought it was outdated. ;-)

gjmveloso:
I had changed my original PR regarding OpenJDK to use version 10 per @commitay feedback guidance.

https://github.com/Homebrew/homebrew-cask/pull/52182

Thanks

commitay:
https://github.com/Homebrew/homebrew-cask/pull/52182 has a build from the `AdoptOpenJDK` project, this cask as the Oracle "open" JDK build.

I think this cask should use the `openjdk` token and we use `adoptopenjdk` for https://github.com/Homebrew/homebrew-cask/pull/52182.

Casks/openjdk.rb
```diff
@@ -0,0 +1,27 @@
+cask 'openjdk' do
+  version '11,28' # build 11+28
```
commitay: Remove the comment.

Casks/openjdk.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  postflight do
+    system_command '/bin/mkdir',
+                   args: ['-p', '--', "#{staged_path}/jdk-#{version.before_comma}.jdk/Contents/Home/bundle/Libraries"]
+
+    system_command '/bin/ln',
+                   args: ['-nsf', '--', "#{staged_path}/jdk-#{version.before_comma}.jdk/Contents/Home/lib/server/libjvm.dylib",
+                          "#{staged_path}/jdk-#{version.before_comma}.jdk/Contents/Home/bundle/Libraries/libserver.dylib"]
+
+    system_command '/bin/ln',
```
commitay: Is there a reason for this to be linked from staged_path rather than being moved?

ryenus: Good point, first of all, this is to fix issue https://github.com/Homebrew/homebrew-cask/issues/5332, see [the related commit to the java cask][1]:

> This fixes issues where applications using java were unable to start
because this lib could not be found.

IMHO, the problem is more about some apps assuming `libserver.dylib` to be on an outdated path, rather than it's missing or misplaced, hence the symbol link.

BTW, there's one apparent difference between the Oracle JDK and OpenJDK, the former is delivered as a **.pkg** file, which installs directly to `/Library/Java/JavaVirtualMachines`, meanwhile openjdk is only available as a **.tar.gz** file, and gets extracted to `#{stage_path}`, that's why I create the link there, then link the whole directory to `/Library/Java/JavaVirtualMachines` afterward.

[1]: https://github.com/Homebrew/homebrew-cask/commit/cdf9d1cf5324b68902a09849acd581f34b9cf7de

Casks/openjdk.rb
ryenus: done

Casks/openjdk.rb
commitay: I'm referring to this:

```
    system_command '/bin/ln',
                   args: ['-nsf', '--', "#{staged_path}/jdk-#{version.before_comma}.jdk", "/Library/Java/JavaVirtualMachines/openjdk-#{version.before_comma}.jdk"],
                   sudo: true
```
Wouldn't it be easier to copy all the files from `staged_path` first and then create the symlinks within `/Library/Java/`?

ryenus: > Wouldn't it be easier to copy all the files from `staged_path` first and then create the symlinks within `/Library/Java/`?

May I ask why do you think that's easier? I think creating a symbol link is much cheaper compared to copying the whole `/usr/local/Caskroom/openjdk/11+28/jdk-11.jdk/` directory, which is about **`287M`**. One thing I can think of is that way there's no modification to the files in the Caskroom, does that justify the copy?

ryenus: How about moving instead of copying? Then we can link afterward in the `/Library/Java/JavaVirtualMachines` directory.

commitay (approved):

Casks/openjdk.rb
ryenus: @vitorgalvao FYI, the version string is now changed to `11,28` to work with `version.(before/after)_comma`, thank you.

bhaavanmerchant:
> What if we change the java cask directly to use OpenJDK instead of Oracle JDK? Or make it an alias for our openjdk cask?

This makes perfect sense to me. I think the default java cask should point to OpenJDK.

ryenus:
> > This makes perfect sense to me. I think the default java cask should point to OpenJDK.
> 
> I’m not going to fight anyone on this, as I don’t really care about Java and I’m not familiar with it’s development. I use it [in the context of Processing](https://processing.org/) and that’s about it. But if I understand the facts, I’m against this.
> 
> Why should we push a specific implementation of Java on users? We don’t do it for any other casks, and for good reason. I’m sure Oracle’s license is bad for a large portion of developers and all, but what about the people that want the official Oracle implementation and don’t care for the license? We make a `closed-java` cask?
> ...
> But again, I’m not going to fight on this. If another maintainer thinks it makes sense to make OpenJDK the default `java` cask, fine by me.

Well, at first I wasn't sure about this either, then I [was pointed][1] to this:
https://blogs.oracle.com/java-platform-group/oracle-jdk-releases-for-java-11-and-later

There it was clarified that:

*  **If you are used to getting Oracle Java SE binaries for free, you can simply continue doing so with Oracle’s OpenJDK builds available at jdk.java.net.**

[1]: https://twitter.com/DonaldOJDK/status/1045656479326322688

HerrSpace:
> Why should we push a specific implementation of Java on users? [...]

It's mostly the same implementation and the same code, just a different packaging and licensing.
https://web.archive.org/web/20150629214653/https://blogs.oracle.com/henrik/entry/java_7_questions_answers

Further OpenJDK is the official reference implementation of Java 7.
https://blogs.oracle.com/java/moving-to-openjdk-as-the-official-java-se-7-reference-implementation

Since both OpenJDK and OracleJDK are maintained by Oracle I'd hardly say this is pushing an ideology, just taking the easy solution.

vitorgalvao:
@ryenus @HerrSpace Thank you both for the explanation. You’ve convinced me into being pro making OpenJDK the `java` cask.

> Since both OpenJDK and OracleJDK are maintained by Oracle I'd hardly say this is pushing an ideology

Small note on this, a quick search on this lead me to erroneously believe it was maintained by a different company, hence my initial comment.

ryenus:
Is there anything yet to be done for this cask to be added/merged? @vitorgalvao @commitay Thank you!

commitay:
I've renamed this so `openjdk` is `java` and the non-free Oracle build is now `oracle-jdk`.

commitay (approved):

vitorgalvao (approved):

ryenus:
It's noteworthy that previously there're two extra symbol links created in `/Library/Java/`:

* `/Library/Java/Home`, see [Fix Java 8 missing Home symlink](https://github.com/Homebrew/homebrew-cask/commit/10c9aa1ece54b69f7cbc5a26bfdc440b1abef407)
* `/Library/Java/MacOS`, see [java 9.0.4 link MacOS (#43424)](https://github.com/Homebrew/homebrew-cask/commit/e3482d5948d6cf890d99cf7615b060aab0a6bcdf) (#43424)

The former, `/Library/Java/Home`, committed by @zwaldowski, was to *Fix Java 8 missing Home symlink*, but no more details. 

Meanwhile the latter, [according to](https://github.com/Homebrew/homebrew-cask/pull/43424#issuecomment-362258554) @tduehr, was for the now removed tool `javapackager`:
> So you can point applications generated by `javapackager` at it and `/Library/Java/Home` to keep the Java install they use up to date without delinking them every new install.

However `javapackage` has been removed from JDK 11, and possibly could be substituted by the currently candidate [JEP 343: Packaging Tool](https://openjdk.java.net/jeps/343).

Should we add them as before? However this would mean that this cask would become conflict with the Oracle JDK cask, otherwise we might need to change the latter to not creating these links.

Thoughts?

commitay:
I don't think we should add them unless there are demonstrable issues with JDK 11.

PR #52589, opened 2018-09-28
Update gisto to 1.10.5

vitorgalvao:

### DIFF
--- a/Casks/gisto.rb
+++ b/Casks/gisto.rb
@@ -1,6 +1,6 @@
 cask 'gisto' do
   version '1.10.5'
-  sha256 '1435b12c75777a97ddd3e28577fcd54247015326954e6aebb1befb72f2e5f66c'
+  sha256 'e3def1a5950b3c34db6393e8e78316445311f00079e073fc875876b21f5e1a0b'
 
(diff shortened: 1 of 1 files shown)


sanusart:
@vitorgalvao Not that I am aware of, but the file on the server (github releases) did not changed, so maybe calculation happened on a partially downloaded file.

In any case it seems that this PR corrects the hash key, I've rechecked just in case.
```
> shasum -a 256 Gisto-1.10.5.dmg
e3def1a5950b3c34db6393e8e78316445311f00079e073fc875876b21f5e1a0b  Gisto-1.10.5.dmg
```

Thanks!

vitorgalvao:
> so maybe calculation happened on a partially downloaded file.

That’s unlikely, as it’d mean both you as the submitter and Travis would have caught the exact same partial download.

> Also checked via VirusTotal

That doesn’t prove it wasn’t tampered with, because there’s no signing info.

Going to merge since you’re the dev, but we should keep our eyes open for further occurrences.

sanusart:
Thanks @vitorgalvao 
I've checked release creation time stamp of the dmg via GitHub API and it looks legit.
I'll re-release it with version bump anyway just to be save. Will keep an eye on it.

PR #52667, opened 2018-10-01
Upgrade Aria2D to Latest

uetcis:
The author stopped offering download for previous versions, instead, now he provides a general url. So here I'm updating for this.

### DIFF
--- a/Casks/aria2d.rb
+++ b/Casks/aria2d.rb
@@ -1,9 +1,9 @@
 cask 'aria2d' do
-  version '1.2.1,1535339003'
-  sha256 '8338d48fa8c90c8f785568b7b16ca86613b6a8c64d107933a1bc72d6aa401e25'
+  version :latest
+  sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


Casks/aria2d.rb
```diff
@@ -1,9 +1,9 @@
 cask 'aria2d' do
-  version '1.2.1,1535339003'
-  sha256 '8338d48fa8c90c8f785568b7b16ca86613b6a8c64d107933a1bc72d6aa401e25'
+  version :latest
```
vitorgalvao: The cask has an `appcast`, so `version` and `sha256` should remain (`1.2.4`).

vitorgalvao (changes requested):

Casks/aria2d.rb
CaptainYukinoshitaHachiman: but we don't have the link for different versions

vitorgalvao: And set the `sha256`.

CaptainYukinoshitaHachiman: But I don't have the version after comma

PR #52670, opened 2018-10-01
Update magicavoxel to 0.99.2

miccal:


miccal:
Same error as before:
```
==> Purging files for version 0.99.2 of Cask magicavoxel
It seems the App Suite source '/usr/local/Caskroom/magicavoxel/0.99.2/MagicaVoxel-0.99.2-alpha-mac' is not there.
```

PR #52689, opened 2018-10-01
Update deegain to 1.1.0

core-code:


core-code:
since deegain has had its SHA changed at least twice since the 'first' version 1.1.0, i'd suggest changing it to no_check.

we've tried previously 
• asking the developer for confirmation. didn't answer for months
• waiting for a new version to be released. i don't think there'll be another update to this


if changing to no_check is not an option, i can send a delete request, since carrying a non-functional cask has little point.

vitorgalvao:
Let’s delete it. I doubt it’s a popular cask.

PR #52690, opened 2018-10-01
Update managenameext to 1.6.5

core-code:

### DIFF
--- a/Casks/managenameext.rb
+++ b/Casks/managenameext.rb
@@ -1,6 +1,6 @@
 cask 'managenameext' do
   version '1.6.5'
-  sha256 '6b127ff94ff3b16b8eda7d9b44c791105f98c86bcda6298f15bc0cc6553773b4'
+  sha256 '3835e138ef5f7a6cf7d875bf870e842f8c0b5c57c9f4f0dc0b0dab74b4c54139'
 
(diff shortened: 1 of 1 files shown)


core-code:
this is a SHA only change but the checksum matches what the developer states:

http://tom.25.free.fr/prg/SHA-256_Checksum/ManageNameExt_c-cs.txt

vitorgalvao (changes requested):
We can’t trust that checksum because a malicious third-party might have changed that as well. That’s why we still require developer confirmation.

core-code:
i've sent an e-mail to the developer

core-code:
i've already got a response from the developer stating the change is legit. but AFAIK the confirmation needs to be public. so should i
• either ask him to confirm it here on this thread or
• since he also said he replaces files all the time for minor changes without bumping the version number, shouldn't be just convert the cask to no-SHA?

vitorgalvao:
You’ve shown yourself to be trustworthy and this is an `app` (meaning quarantine will remove the need for verifications), so post a screenshot of the email (for other users) and we’re good.

Pinging @Homebrew/cask in case anyone disagrees with the decision. Feel free to email me in private if you prefer. If a single maintainer disagrees, I’ll revert the decision and there’s no need to share who the maintainer is.

core-code:
thanks attached is a screenshot of the e-mail including headers

screen shot 2018-10-03 at 15 49 36

PR #52691, opened 2018-10-01
Update retas-studio.rb

core-code:
RETAS STUDIO has been failing to install because of a SHA-only change for quite some time. its unsigned, so i can't send an update and it seems hard to get confirmation from the developer. furthermore 'RETAS STUDIO' seems deprecated, and 'CLIP STUDIO PAINT' seems to be their official successor app. so i propose updating the cask to the newest version of CLIP STUDIO PAINT

### DIFF
--- a/Casks/retas-studio.rb
+++ b/Casks/retas-studio.rb
@@ -1,11 +1,20 @@
 cask 'retas-studio' do
-  version '6.6.0'
-  sha256 '387d299ac77dad4a4116605e492e7c69614c463cde2ce5df439caaa55c9befb5'
+  version '1.8.2'
+  sha256 '55069bfd6ece3bacde37db2eef4fe37c013e0d6f3aed026e3894ed197fa392a2'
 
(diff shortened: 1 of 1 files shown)


Casks/retas-studio.rb
```diff
@@ -1,11 +1,20 @@
 cask 'retas-studio' do
```
vitorgalvao: Also needs to be updated.

vitorgalvao (changes requested):

Casks/retas-studio.rb
core-code: actually clip-studio-paint already exists in HBC, so there is no point in this pull request.

should i rather
• send a SHA-only change to keep retas-studio working or
• send a pull request which deletes retas-studio since it seems to be obsolete anyway
?

vitorgalvao: Let’s remove it.

PR #52764, opened 2018-10-03
chromium: Add version information

tikki:
Enables checksum checking.

### DIFF
--- a/Casks/chromium.rb
+++ b/Casks/chromium.rb
@@ -1,9 +1,10 @@
 cask 'chromium' do
-  version :latest
-  sha256 :no_check
+  version '596179'
+  sha256 '4aac476233875efb22d0da8c081189727aceb624170f8e961cb5b13395fa60ae'
 
-  # download-chromium.appspot.com was verified as official when first introduced to the cask
-  url 'https://download-chromium.appspot.com/dl/Mac?type=snapshots'
+  # commondatastorage.googleapis.com/chromium-browser-snapshots was verified as official when first introduced to the cask
+  url "https://commondatastorage.googleapis.com/chromium-browser-snapshots/Mac/#{version}/chrome-mac.zip"
+  appcast 'https://download-chromium.appspot.com/rev/Mac?type=snapshots'
   name 'Chromium'
   homepage 'https://www.chromium.org/Home'
 


Casks/chromium.rb
```diff
@@ -1,9 +1,10 @@
 cask 'chromium' do
-  version :latest
-  sha256 :no_check
+  version '596179'
+  sha256 '4aac476233875efb22d0da8c081189727aceb624170f8e961cb5b13395fa60ae'
 
-  # download-chromium.appspot.com was verified as official when first introduced to the cask
-  url 'https://download-chromium.appspot.com/dl/Mac?type=snapshots'
+  # commondatastorage.googleapis.com/chromium-browser-snapshots was verified as official when first introduced to the cask
+  url "https://commondatastorage.googleapis.com/chromium-browser-snapshots/Mac/#{version}/chrome-mac.zip"
```
vitorgalvao: Where’s this coming from?

vitorgalvao (changes requested):

Casks/chromium.rb
tikki: It's what https://download-chromium.appspot.com/dl/Mac?type=snapshots redirects to, where all the snapshots are actually hosted.

```
curl -v 'https://download-chromium.appspot.com/dl/Mac?type=snapshots'
[...]
> GET /dl/Mac?type=snapshots HTTP/2
> Host: download-chromium.appspot.com
> User-Agent: curl/7.54.0
> Accept: */*
>
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
< HTTP/2 302
< content-type: text/html; charset=utf-8
< cache-control: no-cache
< location: https://commondatastorage.googleapis.com/chromium-browser-snapshots/Mac/596185/chrome-mac.zip
< x-cloud-trace-context: 5d6e3729d5be75831fc8567ec1f54e9b;o=1
< date: Wed, 03 Oct 2018 13:44:07 GMT
< server: Google Frontend
< content-length: 0
< alt-svc: quic=":443"; ma=2592000; v="44,43,39,35"
<
```

vitorgalvao:
How often does Chromium get updates?

tikki:
> How often does Chromium get updates?

The build bot that creates these compiled snapshots does it every few commits – I don't how it decides when to create another snapshot, but it doesn't do it for *every* commit.
This results in multiple updates per day, I'd guess at least a dozen.
This is because Chromium is really the dev project for Chrome, and not meant for end users – they recommend to use Chrome Canary (i.e. nightly builds) instead.

If you're wondering if that means the recipe will almost always be outdated, then yes, that's correct. 
But I think
1. that's not really a problem because we can update the recipe just like with every other cask; this could even be automated by a script that monitors the appcast URL
2. having a version number and checksum outweighs the drawback of not having the most up-to-the-second current release. Without the version number chromium won't even show up in `brew cask outdated` (and also won't be automatically upgraded), even though it's probably *always* outdated on the user's system.

vitorgalvao:
Doesn’t matter if it’s automated, it’ll still result in constant PRs that have to be reviewed and merged. That’s a huge maintenance burden. Multiple updates a day is definitely too much, and I feel we might’ve already discuss this cask and that specific problem.

tikki:
I understand, then let's definitely not automate it. :)

Do you think people will manually update this recipe so often that it will be problematic? Otherwise I'd still prefer having a pinned version in the recipe.

vitorgalvao:
> Do you think people will manually update this recipe so often that it will be problematic?

I guarantee it. We have multiple people that auto-submit from checking the `appcast`s (which is the point).

core-code:
FWIF i'd prefer no information to information thats bound to be always incorrect

tikki:
> FWIF i'd prefer no information to information thats bound to be always incorrect

It is very much correct though, just outdated, and that's a systematic problem, i.e. pretty much all homebrew recipes are outdated at some point and some stay outdated even for a long time.

If we're not adding version information to the chromium recipe I there should be a discussion (maybe one's already happening/already happend?) how to handle these cases of frequently updating software.
Keeping the version at "latest" **breaks** the upgrade mechanism (calling `brew cask upgrade` skips chromium, so I have to remember to upgrade it manually every so often), and having no checksum is really a shame safety wise.

One way I can think of would be to have brew dynamically check the current version, so the `version` stanza would have to offer an optional block syntax (like e.g. `url` already does). For chromium we could then retrieve the information from the appcast url.

commitay:
Does Chromium self-update like Chrome?

vitorgalvao:
> Does Chromium self-update like Chrome?

No.

commitay:
I think we need to either version it (and try to discourage people from sending PRs for it every five minutes) or delete it, as we can't expect all users to run `upgrade --greedy` and running an outdated browser isn't good.

core-code:
> (and try to discourage people from sending PRs for it every five minutes)

i think whats needed is a way to update casks without sending PRs. 

some appcasts are 100% reliable, and automation would work perfectly without the need for additional overhead

vitorgalvao:
> and running an outdated browser isn't good.

I agree. Though this is not a consumer browser, and is not intended to be ran as such.

For added info/context, an example of a cask we made `latest` when we know there are versioned links is `dropbox`.

> (and try to discourage people from sending PRs for it every five minutes)

Ever since I’ve been thinking of a system to auto-submit cask updates, I’ve been thinking of a way to discourage users from submitting manual updates to those. Adding some info to the cask is a bad solution. The best I’ve came up with was adding a blacklist to `cask-repair` — it doesn’t cover all cases, but it covers a lot.

Ideally, the auto-updating system would be such that in the end, unless there’s a change that needs to be made to fix something (like an `app` changing location inside the `dmg`), users should no longer submit version bumps, instead submitting fixes to the auto-update system as necessary.

> i think whats needed is a way to update casks without sending PRs.

I’ve also thought about this, but I’ve always been wary. It’s simple enough to create a dummy Github account with write permissions and have a script sit on a server that fires with `cron` and auto-pushes on changes. But there’s always the nagging feeling in the back of my mind that that server suddenly becomes an appetising target to break into, as it’ll essentially give write permissions to HBC. At least with the PRs, we have an extra line of defence.

Regarding Chromium/Dropbox, I mean, maybe if we had them versioned and the script ran only once a day it’d be fine?

We could also have a script to be ran manually by maintainers that would auto-merge PRs of trusted users/bots *if* the only changes were `version` and `sha256`.

core-code:
> The best I’ve came up with was adding a blacklist to cask-repair — it doesn’t cover all cases, but it covers a lot.

couldn't something be added to Travis to catch those cases that cask-repair didn't catch?

> But there’s always the nagging feeling in the back of my mind that that
> server suddenly becomes an appetising target to break into, 
> as it’ll essentially give write permissions to HBC

wouldn't there be easier ways to infiltrate more people? (hm, how many users do you have?)
just think of all the apps doing version-checking to appcasts sitting on expired domains that can be easily purchased...

> At least with the PRs, we have an extra line of defence

yes - on the downside it creates a lot of work, there is no guarantee that the 'defence' will work, and there can be considerable delay until the updates are merged. 

> We could also have a script to be ran manually by maintainers that
> would auto-merge PRs of trusted users/bots if the only changes were version and sha256

sounds good and could work in conjunction with what i proposed in issue #52837

claui:
> having no checksum is really a shame safety wise.

@tikki Please keep in mind that the SHA checksum is meant to serve as an integrity check, and [by no means a security feature](https://github.com/Homebrew/homebrew-cask/issues/48862).

vitorgalvao:
> couldn't something be added to Travis to catch those cases that cask-repair didn't catch?

That would only catch them *after* they’re submitted, at which point it’s too late.

> just think of all the apps doing version-checking to appcasts sitting on expired domains that can be easily purchased

URLs don’t change from appcasts alone. If an appcast shows a different URL and the old is dead, that’s an automatic flag for us to verify it. Even if all the domains of an expired app were compromised, HBC wouldn’t be the primary mode of sharing the app, and it would be an abandoned app without many users anyway. The server script could rewrite *any* popular cask and infect a good many users in a few hours.

> on the downside it creates a lot of work, there is no guarantee that the 'defence' will work, and there can be considerable delay until the updates are merged.

It may not work always, but it does work a lot of time. Bypassing the current method takes so much work, targeting HBC alone is unfeasible.

commitay:
Just remembered about this that I added to the formula `audit`: 

Adding something similar for cask would block `cask-repair` and manual PRs (assuming people run the `audit`).
```ruby
      throttled = %w[
        aws-sdk-cpp 10
        awscli 10
        heroku 10
        quicktype 10
        vim 50
      ]


      throttled.each_slice(2).to_a.map do |a, b|
        next if formula.stable.nil?


        version = formula.stable.version.to_s.split(".").last.to_i
        if @strict && a == formula.name && version.modulo(b.to_i).nonzero?
          problem "should only be updated every #{b} releases on multiples of #{b}"
        end
      end
```

vitorgalvao:
> considering we have 115 open pull requests and some obviously-right PRs have been stuck for more than a week i am not convinced anymore that bypassing the PR system is such a bad idea

All the stuck PRs that shouldn’t be are stuck due to problems with CI. If we had a system that bypassed merging, those would still not be merged because an automated system would not merge what’s not passing. We’d be exactly where we are.

vitorgalvao:
Passing on this for now.

PR #52865, opened 2018-10-04
Secure URLs (a-b-c)

vszakats:
Used `chkhttps` tool to verify any possible URLs to use HTTPS. Applied upgrades to `url` with versioned strings. This is for casks starting with letters A-B-C.

### DIFF
--- a/Casks/0-ad.rb
+++ b/Casks/0-ad.rb
@@ -3,7 +3,7 @@ cask '0-ad' do
   sha256 'c9de859ad0e2cc6b1085803d1fc107094423520e16c55b32175f43df11475441'
 
   # wildfiregames.com was verified as official when first introduced to the cask
-  url "http://releases.wildfiregames.com/0ad-#{version}-osx64.dmg"
+  url "https://releases.wildfiregames.com/0ad-#{version}-osx64.dmg"
   appcast 'https://play0ad.com/download/mac/'
   name '0 A.D.'
   homepage 'https://play0ad.com/'
--- a/Casks/0xed.rb
+++ b/Casks/0xed.rb
@@ -2,9 +2,9 @@ cask '0xed' do
   version :latest
   sha256 :no_check
 
-  url 'http://www.suavetech.com/download/0xED.tar.bz2'
+  url 'https://www.suavetech.com/download/0xED.tar.bz2'
   name '0xED'
-  homepage 'http://www.suavetech.com/0xed/'
+  homepage 'https://www.suavetech.com/0xed/'
 
   app '0xED.app'
 
--- a/Casks/5kplayer.rb
+++ b/Casks/5kplayer.rb
@@ -3,7 +3,7 @@ cask '5kplayer' do
   sha256 'b5f02f1cb9ecdd832c4759c9e7b33eb7ef5d1a91c33dcdfa194de68de3cac753'
 
   url 'https://www.5kplayer.com/download/5kplayer.dmg'
-  appcast 'http://www.5kplayer.com/upgrade/mac/updatecast.xml'
+  appcast 'https://www.5kplayer.com/upgrade/mac/updatecast.xml'
   name '5KPlayer'
   homepage 'https://www.5kplayer.com/'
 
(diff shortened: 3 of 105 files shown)


commitay:
Are you sure none of the downgrade to HTTP? Currently `HOMEBREW_NO_INSECURE_REDIRECT` doesn't work for cask.

vszakats:
The checker does not test for the downgrade case.

I've seen some homepages doing that while manually checking those flagged for manual verification, but since those are opened in browsers (typically), the downgrade should be apparent. (There are some weirder things, like infinite redirect regardless of protocol.)

For binary/appcast URLs I believe there is not much, if anything lost compared to a completely insecure link, and my anecdotal experience is that protocol downgrades for file downloads is becoming less and less common, and rather atypical by now (though I don't have stats for Cask links in particular.)

The stats I _can_ see is that out of 2383 insecure links, about 600-700 is likely upgradable, possibly more. It would seem useful to do it!

commitay:
@Homebrew/cask Any opinions on this?

vitorgalvao:
> The checker does not test for the downgrade case.

It must. It has been long-standing policy that cask URLs should stay truthful to the last redirect for auditability purposes. If it says `https`, then the user should be safe in assuming it’ll stay `https` throughout.

I could see the argument for scraping that policy if we had a technical solution to prevent `https` from becoming `http`, but since we don’t, the check has to take it into account or we can’t trust the results at all.

vszakats:
I could find much more opportunities for upgrades by writing a custom script to upgrade `url`s with macros.

But, if downgrade check is a requirement I'm rather (and sadly) pulling off this project and leave stuff insecure.

I believe this policy serves no useful purpose, especially when compared to the benefit of upgrading about a thousand URLs to secure. Downgrades _are rare_ and the checksum is there as a final check anyway, so unless I'm missing something obvious an insecure redirect will not introduce extra security issue over a plain insecure starting link. For those extra picky people (like me), a personal firewall like Little Snitch will report the port 80 access anyway. This isn't something desired of course, but in such case this is better to be reported to site maintainers to fix.

I hope there will come an opportunity to rethink this policy.

vszakats:
A total of about 850 links could be upgraded. That's 35% of all insecure links used in Casks at the moment.

vszakats:
One more thought: I think it's misleading to decide about protocol downgrades on one-off points in time such as when this PR was created. URL handling and redirections are inherently dynamic, so they may (and sometimes do) change over time. It means there is no guarantee that a perfectly secure URL doesn't get redirected or routed-through a non-secure hop any day in the future. In fact they may also be redirected depending on geo-location, time of day, server load or any other factor.

For these reasons I strongly think it makes not much sense for me to develop a system to detect these and then use it at the _arbitrary time/place_ when looking for protocol upgrades. It may or may not redirect for me/today, and do the reverse tomorrow and/or for a different user.

I'd argue that the _only_ correct place to check for insecure redirects (_if_ we're interested in them) is when the URL is _used_, like it is correctly done in non-Cask Homebrew situations.

Other methods will just add a false-sense of guarantee for something that cannot be guaranteed.

Please give this some more though and/or maybe discuss with other maintainers.

vitorgalvao:
@vszakats You’ve convinced me.

But I do have a question. Let’s ignore the redirects. With your system, if an `https` URL is no longer accessible but the same `http` link is, will it be detected/corrected?

reitermarkus:
> Currently `HOMEBREW_NO_INSECURE_REDIRECT` doesn't work for cask.

Why not? They are using the same download strategies now.

commitay:
It wasn't working when I tried it recently, can't remember which casks I tested though.

vszakats:
@vitorgalvao: No, the checker is specifically meant for upgrading links from insecure to HTTPS. Meaning it's not a general URL checker/fixer, so initially-secure URLs are automatically omitted from processing. The typical, internet-wide trend is HTTP → HTTPS, and the script's goal is to help following up any references pointing to an already upgraded URL. I haven't explored this deeply, but I'd assume that majority of downgrades should be the result of a server configuration mistake or site restructure, and usually it's better to try finding out where a new secure link is available or reporting the breakage upstream.

Probably a general link checker would be useful as well, but it falls outside my interest (and free time.)

PR #53029, opened 2018-10-07
doc: update OSDN URL rules

vszakats:
Ref: https://github.com/Homebrew/brew/pull/5052

Affected Casks:
- letterfix
- miditrail
- nndd

### DIFF
--- a/Casks/letterfix.rb
+++ b/Casks/letterfix.rb
@@ -1,11 +1,11 @@
 cask 'letterfix' do
-  version '2.6.0,68346'
+  version '2.6.0'
   sha256 '89de44a2b0e6cb43915d9584615226811f939a42aed6d2a14ba099de18db2768'
 
-  url "http://onet.dl.osdn.jp/letter-fix/#{version.after_comma}/LetterFix-#{version.before_comma}.dmg"
-  appcast 'https://ja.osdn.net/projects/letter-fix/releases/rss'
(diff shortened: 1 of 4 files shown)


commitay:
Can you add the affected casks as well to please?

commitay:
Not sure what to change this to:

http://osdn.dl.osdn.jp/users/8/8642/genjyuugothic-20150607.zip

The homepage has this URL:
https://osdn.jp/downloads/users/8/8644/genjyuugothic-x-20150607.zip

but the `audit` doesn't like it.

vszakats:
This is the closest URL I could find:
https://osdn.net/downloads/users/8/8644/genjyuugothic-x-20150607.zip/
(via <https://osdn.net/users/mm-rounded/pf/genshin/files/?id=8632>)

One solution is to also allow `/downloads/` instead of `/dl/`.

commitay:
Whoops. I copied the wrong URL from the homepage, updated my comment above.

> One solution is to also allow /downloads/ instead of /dl/.

This seems like the easist option. 👍

vszakats:
It looks like these are files associated with a specific username, not with a project, hence the different URL scheme.

commitay:
> Can you add the affected casks as well to please?

Sorry, I meant add the actual casks to this PR.

vszakats:
There you go.

commitay:
Thanks!

PR #52975, opened 2018-10-06
Secure URLs

vszakats:
Ref: https://github.com/Homebrew/homebrew-cask/pull/52865

Stats:
```
              before          after
   Casks:       4041      →    4041
   HTTPS:       8291      →    9254
Insecure:       2382      →    1423
    HTTP:       2374      →    1420
     FTP:          8      →       3
   Ratio:         77.68 % →      86.67 %
```

PR #53032, opened 2018-10-07
Secure URLs (2 of 3)

vszakats:
Ref: https://github.com/Homebrew/homebrew-cask/pull/52975

### DIFF
--- a/Casks/headset.rb
+++ b/Casks/headset.rb
@@ -6,7 +6,7 @@ cask 'headset' do
   url "https://github.com/headsetapp/headset-electron/releases/download/v#{version}/Headset-#{version}.dmg"
   appcast 'https://github.com/headsetapp/headset-electron/releases.atom'
   name 'Headset'
-  homepage 'http://headsetapp.co/'
+  homepage 'https://headsetapp.co/'
 
   app 'Headset.app'
 
--- a/Casks/hedgewars.rb
+++ b/Casks/hedgewars.rb
@@ -7,7 +7,7 @@ cask 'hedgewars' do
     sha256 '2a5fbfa005ec6aeea172270397025c17a2c117224dd21db5214b8cbbeade411b'
   end
 
-  url "http://www.hedgewars.org/download/releases/Hedgewars-#{version}.dmg"
+  url "https://www.hedgewars.org/download/releases/Hedgewars-#{version}.dmg"
   appcast 'https://www.hedgewars.org/download/appcast.xml'
   name 'Hedgewars'
   homepage 'https://hedgewars.org/'
--- a/Casks/heimdall-suite.rb
+++ b/Casks/heimdall-suite.rb
@@ -5,7 +5,7 @@ cask 'heimdall-suite' do
   # bitbucket.org/benjamin_dobell/heimdall was verified as official when first introduced to the cask
   url "https://bitbucket.org/benjamin_dobell/heimdall/downloads/heimdall-suite-#{version}-mac.dmg"
   name 'Heimdall Suite'
-  homepage 'http://glassechidna.com.au/heimdall/'
+  homepage 'https://glassechidna.com.au/heimdall/'
 
   pkg "Heimdall Suite #{version}.pkg"
 
--- a/Casks/helio.rb
+++ b/Casks/helio.rb
@@ -2,7 +2,7 @@ cask 'helio' do
   version '1.7.6'
   sha256 '7f0499eb9266e65f3535795c20b18b82bf926a9966f757ea1a5854d861e8cfa6'
 
-  url "http://helioworkstation.com/app/Helio-#{version}.dmg"
+  url "https://helioworkstation.com/app/Helio-#{version}.dmg"
   name 'Helio'
   homepage 'https://helioworkstation.com/'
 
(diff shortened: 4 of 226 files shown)


Casks/inc.rb
```diff
@@ -2,8 +2,8 @@
   version :latest
   sha256 :no_check
 
-  # inc-static.kippt.com.s3.amazonaws.com was verified as official when first introduced to the cask
-  url 'http://inc-static.kippt.com.s3.amazonaws.com/apps/inc-osx.zip'
+  # s3.amazonaws.com/inc-static.kippt.com was verified as official when first introduced to the cask
+  url 'https://s3.amazonaws.com/inc-static.kippt.com/apps/inc-osx.zip'
```
commitay: This need to be verifiable from the homepage.

Casks/iteleport-connect.rb
```diff
@@ -2,10 +2,12 @@
   version '6.0.0.2'
   sha256 '834098a95c95183782e498de6427e3106b867a1f1793e4fe1306bb065baa732f'
 
-  url "http://www.iteleportmobile.com/download/iTeleport%20Connect.v#{version}.app.zip"
-  appcast 'http://www.iteleportmobile.com/download/sparkle.xml'
+  # Broken URL
```
commitay: Seems to have different URLs on the homepage, I'd suggest dealing with this separately.

Casks/kega-fusion.rb
```diff
@@ -2,9 +2,10 @@
   version '3.63i'
   sha256 'f329d3c700a4b1f66e49753f91d27182b1e67bdd0e9bd16c6347f989aa7131ba'
 
-  url "http://www.carpeludum.com/download/Fusion#{version.no_dots}.zip"
+  url "https://www.carpeludum.com/download/Fusion#{version.no_dots}.zip"
+  # Alternative without an insecure redirect: "https://www.carpeludum.com/wp-content/files/Fusion#{version.no_dots}.zip"
```
commitay: Lets deal with this in a follow up, we don't really want to add comments in casks.

Casks/mylio.rb
```diff
@@ -2,10 +2,10 @@
   version :latest
   sha256 :no_check
 
-  # myliodownloads.s3-website-us-east-1.amazonaws.com was verified as official when first introduced to the cask
-  url 'http://myliodownloads.s3-website-us-east-1.amazonaws.com/mylio.dmg'
+  # myliodownloads.s3.amazonaws.com was verified as official when first introduced to the cask
+  url 'https://myliodownloads.s3.amazonaws.com/mylio.dmg'
```
commitay: Needs to be verified from the homepage.

Casks/open-sankore.rb
```diff
@@ -5,7 +5,7 @@
   # cndp.fr/open-sankore was verified as official when first introduced to the cask
   url "http://www.cndp.fr/open-sankore/OpenSankore/Releases/v#{version}/Open-Sankore_MacOSX_#{version}.dmg"
   name 'Open-Sankore'
-  homepage 'http://open-sankore.org/'
+  homepage 'https://web.archive.org/web/20171028190857/open-sankore.org/'
```
commitay: Lets deal with this in a follow up, probably need to find a current homepage or delete it.

Casks/papers.rb
```diff
@@ -3,7 +3,8 @@
   sha256 '15c226fd763479319fe115c710d1ee40180a7d8137954da71cf777b9ae0dfd7c'
 
   # appcaster.papersapp.com/apps/mac/production/download was verified as official when first introduced to the cask
-  url "http://appcaster.papersapp.com/apps/mac/production/download/#{version.after_comma}/papers_#{version.before_comma.no_dots}_#{version.after_comma}.dmg"
+  url "https://appcaster.papersapp.com/apps/mac/production/download/#{version.after_comma}/papers_#{version.before_comma.no_dots}_#{version.after_comma}.dmg"
+  # Alternative without an insecure redirect: "https://downloads.mekentosj.com/papers_#{version.before_comma.no_dots}_#{version.after_comma}.dmg"
```
commitay: Lets deal with this in a follow up, we don't really want to add comments in casks.

Casks/inc.rb
vszakats: These are URL synonyms by default. The bucker name can be used as a subdomain or a subdirectory inside the S3 root domain. They point to the same location. But, the sad thing about bucket names with a dot `.` is that they break HTTPS access, which can be (in almost all cases) be worked around by specifying the bucket as a subdirectory. This is what's happening here.

Unfortunately it's still possible to find bucket names with dots and used as subdomains. Probably something to report upstream, but meanwhile these URLs are just as perfect as the wrong "official" ones. Well, more perfect because they offer security.

Casks/open-sankore.rb
vszakats: I only use archive.org if there was no reasonable way to find the live homepage.

Casks/mylio.rb
vszakats: See above.

Casks/inc.rb
commitay: Yes, but as they are _required_ to be verified from the homepage and we can't do that.

Adding an audit rule to deal with and documenting it this might be one option but I'd drop casks that are like this for now so we can get these PRs merged.

commitay (approved):

commitay:
@vszakats Thanks for these, they are appreciated but in future please don't submit so many changes at once as these weren't easy to review. I'd strongly suggest separating them by HTTP -> HTTPS only changes vs actual URL changes and a maxium of 50 casks per PR.

PR #53033, opened 2018-10-07
Secure URLs (3 of 3)

vszakats:
Ref: https://github.com/Homebrew/homebrew-cask/pull/52975

### DIFF
--- a/Casks/plotdevice.rb
+++ b/Casks/plotdevice.rb
@@ -2,10 +2,10 @@ cask 'plotdevice' do
   version '0.10.0'
   sha256 '8c224f228c5b014e76925339171f4207a0b4a09f595500703d791cb689a32e8d'
 
-  url "http://plotdevice.io/app/PlotDevice_app-#{version}.zip"
-  appcast 'http://plotdevice.io/app.xml'
+  url "https://plotdevice.io/app/PlotDevice_app-#{version}.zip"
+  appcast 'https://plotdevice.io/app.xml'
   name 'PlotDevice'
-  homepage 'http://plotdevice.io/'
+  homepage 'https://plotdevice.io/'
 
   app 'PlotDevice.app'
 end
--- a/Casks/plotdigitizer.rb
+++ b/Casks/plotdigitizer.rb
@@ -2,10 +2,11 @@ cask 'plotdigitizer' do
   version '2.6.8'
   sha256 '56ddfb70fd6eeb9598a83c1ffa44125b46944d86fb69de8e40a59c939f6876cf'
 
+  # downloads.sourceforge.net/plotdigitizer was verified as official when first introduced to the cask
   url "https://downloads.sourceforge.net/plotdigitizer/PlotDigitizer_#{version}_MacOSX.dmg"
   appcast 'https://sourceforge.net/projects/plotdigitizer/rss'
   name 'Plot Digitizer'
-  homepage 'http://plotdigitizer.sourceforge.net/'
+  homepage 'https://plotdigitizer.sourceforge.io/'
 
   app 'PlotDigitizer.app'
 end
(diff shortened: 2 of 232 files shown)


Casks/quickpvr.rb
```diff
@@ -4,7 +4,7 @@
 
   url "https://www.limbic.com/downloads/QuickPVR_v#{version.major}.zip"
   name 'QuickPVR'
-  homepage 'http://www.limbic.com/quickpvr.html'
+  homepage 'https://www.limbic.com/#games'
```
commitay: This doesn't seem to be the homepage?

Casks/soube.rb
```diff
@@ -6,7 +6,7 @@
   url "https://github.com/DracotMolver/Soube/releases/download/#{version}/soube-#{version}.dmg"
   appcast 'https://github.com/DracotMolver/Soube/releases.atom'
   name 'Soube'
-  homepage 'http://soube.diegomolina.cl/'
+  homepage 'https://web.archive.org/web/20171224130516/soube.diegomolina.cl/'
```
commitay: I'd change this to the GitHub repo.

Casks/td-agent.rb
```diff
@@ -2,8 +2,8 @@
   version '2.3.5-1'
   sha256 '2f2973ed29a3edd416cc5dad6b924751a2cdb8cdbb2792dc5344f3e13529b190'
 
-  # packages.treasuredata.com.s3.amazonaws.com was verified as official when first introduced to the cask
-  url "http://packages.treasuredata.com.s3.amazonaws.com/2/macosx/td-agent-#{version}.dmg"
+  # s3.amazonaws.com/packages.treasuredata.com was verified as official when first introduced to the cask
+  url "https://s3.amazonaws.com/packages.treasuredata.com/2/macosx/td-agent-#{version}.dmg"
```
commitay: How can we verify this URL from the homepage or appcast?

Casks/quickpvr.rb
vszakats: The project is no longer presented on this homepage.

Casks/td-agent.rb
vszakats: It's a synonym. No need to.

vszakats:
This more and more looks like an impossible task to achieve :( I've been dealing with these for 16 hours straight just today and the situation is worse than ever. At this point I'd prefer to abandon it and push my efforts into the trash.

To keep sanity I'd suggest to pull these as-is and commit follow up fixes (e.g. Cask deletions) where it makes sense, but I strongly think that using an archive.org homepage (or the vendor's generic product page) is better than a broken URL, at least until the Cask is there.

For S3 URLs, my URL use is positively safe, and I don't see them being thrown away on the grounds that the app maintainers didn't configure it with HTTPS in mind.

Casks/soube.rb
vszakats: The archive URL has different and more rich content than the GitHub page. In fact it's still being linked from the top of the GitHub page.

commitay:
> This more and more looks like an impossible task to achieve

Hence why dealing them in smaller batches is easier.

> For S3 URLs, my URL use is positively safe, and I don't see them being thrown away on the grounds that the app maintainers didn't configure it with HTTPS in mind.

We have a specific rule to verify URLs, it doesn't that these URLs are safe to use. We can adjust our rules to deal with it but I'm not happy to just merge these as-is.

commitay:
If you want to abandon them I'll deal with them from here.

vszakats:
Please do, thank you — I can't deal with these anymore at this point.

commitay:
Okay, I won't bother preserving the commit attribution when making changes but as the PRs are yours the changes will still be credited to you when we squash and merge.

commitay (approved):

PR #53031, opened 2018-10-07
Secure URLs (1 of 3)

vszakats:
Ref: https://github.com/Homebrew/homebrew-cask/pull/52975

PR #53035, opened 2018-10-07
Secure URLs 

commitay:
Leftovers from:
https://github.com/Homebrew/homebrew-cask/pull/53031
https://github.com/Homebrew/homebrew-cask/pull/53032
https://github.com/Homebrew/homebrew-cask/pull/53033

cc @vszakats

### DIFF
--- a/Casks/bitpost.rb
+++ b/Casks/bitpost.rb
@@ -2,8 +2,8 @@ cask 'bitpost' do
   version '0.9.8.3'
   sha256 '8e58ddf022eae721bfca6ecb456edf40baa23ea31076ee9c1ad75e947b0cea2b'
 
-  # voluntary.net.s3.amazonaws.com was verified as official when first introduced to the cask
-  url "http://voluntary.net.s3.amazonaws.com/Bitpost.#{version}.zip"
+  # s3.amazonaws.com/voluntary.net was verified as official when first introduced to the cask
+  url "https://s3.amazonaws.com/voluntary.net/Bitpost.#{version}.zip"
   appcast 'https://voluntary.net/bitpost/'
   name 'Bitpost'
   homepage 'https://voluntary.net/bitpost/'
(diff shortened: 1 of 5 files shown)


commitay:
Some of these don't comply with our current verification policy, others probably need to be deleted or otherwise need changes for merging.

vszakats:
As for `context-free`, `go2shell`, `kega-fusion`, `papers`, their sin is that they use an HTTPS `url`, but that gets redirected to a HTTP one, _and_ I managed to find an alternate HTTPS that's fully secure (albeit not officially listed), by keeping redirections secure all along. The practice is done by other Casks as well and the redirection is visible throughout the download for transparency, so I can't see a reason why not to upgrade these particular few cases. Possibly upstream projects may be informed about these redirection bugs so that they can fix it.

If nothing else, homepages could be upgraded just fine.

vszakats:
If a working archive URL isn't acceptable, those projects with broken homepages, unlisted apps, or broken download URLs shouldn't be removed instead? These are affected: `biba` (product deprecated and replaced with an Amazon product), `cactus` (inaccessible homepage for 2 years), `iteleport-connect` (broken download URLs), `open-sankore` (inaccessible homepage 1 year), `quickpvr` (app not listed on website anymore, but download works), `soube` (homepage broken for almost 1 year, GitHub page still there, but inactive).

vszakats:
As for the remaining cases of S3 URLs. The two supported URL styles are documented here:
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro

The "path-style" URLs have the advantage to support HTTPS for bucket names with dot (`.`) characters in them. This patch made sure to use it where possible. They point to the exact same content, which is guaranteed by AWS.

Affected Casks: `td-agent`, `inc`, `find-any-file`, `bitpost`.

There is another case with S3, where it's possible to add HTTPS by stripping the region designator `-website-us-east-1` from the S3 domain. This region is the default, so it's not required. Serving single files also don't require the [_website_ hosting feature](https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html) of S3, so it can be stripped as well. Affected Casks: `mylio`, `cactus`.

PR #53183, opened 2018-10-10
Delete inkdrop.rb

vitorgalvao:
Walled build.


suschizu:
@vitorgalvao - you are right - sorry for that! There is no official download link on the website (without subscription) – there is just the download link for the demo-version that is almost the same : 
https://d3ip0rje8grhnl.cloudfront.net/v3.22.3/Inkdrop-demo-3.22.3-Mac.zip
Vs
https://d3ip0rje8grhnl.cloudfront.net/v3.22.3/Inkdrop-3.22.3-Mac.zip

I’m not sure if this could help

PR #53685, opened 2018-10-16
Update itch to 25.0.0

core-code:

### DIFF
--- a/Casks/itch.rb
+++ b/Casks/itch.rb
@@ -8,7 +8,7 @@ cask 'itch' do
   name 'itch'
   homepage 'https://itch.io/app'
 
-  installer manual: 'Install itch.app'
+  installer script: "Install itch/Install itch.app"
 
   uninstall delete: '~/Applications/itch.app'
 end
--- a/Casks/itch.rb
+++ b/Casks/itch.rb
@@ -16,6 +16,15 @@ cask 'itch' do
     set_permissions "#{staged_path}/Install itch.app", '0777'
   end
 
-  uninstall delete: '~/Applications/itch.app',
+  uninstall delete: [
+                      '~/Applications/itch.app',
+                      '~/Library/Application Support/itch-setup/',
+                    ],
             quit:   'io.itch.mac'
+
+  zap trash: [
+               '~/Library/Application Support/itch/',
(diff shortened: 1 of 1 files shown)


core-code:
i think the installer runs unattended anyway so maybe better use that

`
env: /usr/local/Caskroom/itch/25.0.0/Install itch.app: Permission denied
`

anyone know why this happens?

vitorgalvao:
> anyone know why this happens?

Because it’s not a script. You’re calling the `app` bundle, which is a directory.

Casks/itch.rb
```diff
@@ -1,12 +1,20 @@
...
+  # nuts.itch.zone was verified as official when first introduced to the cask
+  url 'http://nuts.itch.zone/download/osx'
   appcast 'https://github.com/itchio/itch/releases.atom'
   name 'itch'
   homepage 'https://itch.io/app'
 
-  app 'itch.app'
+  container nested: 'Install itch.dmg'
+
+  installer script: 'Install itch.app/Contents/MacOS/itch-setup'
```
vitorgalvao: Does this open a GUI?

core-code: yes it does but as far as i remember the installation does run unattended

core-code: 1.) the sha has changed 3 times since i first sent it, so i've changed to version latest
2.) i've verified the installation works locally

Casks/itch.rb
```diff
@@ -1,12 +1,30 @@
 cask 'itch' do
-  version '23.6.3'
-  sha256 '3ed905c4e7e0ad37004dd235a0dcf27325adac10fb829d3bfcd076c25442fda4'
+  version :latest
```
vitorgalvao: Needs to stay versioned and have a `sha256`, because it has an `appcast`.

vitorgalvao (changes requested):

core-code:
just some info so that it isn't lost
the installer checks version by retrieving 'https://broth.itch.ovh/itch/darwin-amd64/LATEST'

if we want to bypass the installer and install directly we could do so by retrieving
'https://broth.itch.ovh/itch/darwin-amd64/#{version}/archive-default'

PR #53687, opened 2018-10-16
topcat: install binaries

PaulPrice:
Topcat provides two command-line scripts that are useful for starting
the program with particular inputs.

### DIFF
--- a/Casks/topcat.rb
+++ b/Casks/topcat.rb
@@ -1,6 +1,6 @@
 cask 'topcat' do
-  version '4.6-1'
-  sha256 'd26dcdec120d1498a4d97140396b93dd1ea7738364fcfa959e66a9e3201cadf2'
+  version '4.6-2'
+  sha256 '010cf7f23a75a1831f26cfe176a6e998007b8a90b10d9b34c058905705822d4e'
 
   url 'http://www.star.bris.ac.uk/~mbt/topcat/topcat-full.dmg'
   appcast 'http://www.star.bris.ac.uk/~mbt/topcat/sun253/versions.html'
--- a/Casks/topcat.rb
+++ b/Casks/topcat.rb
@@ -8,4 +8,6 @@ cask 'topcat' do
   homepage 'http://www.star.bris.ac.uk/~mbt/topcat/'
 
   app 'TOPCAT.app'
+  binary "#{appdir}/TOPCAT.app/Contents/Resources/bin/topcat"
+  binary "#{appdir}/TOPCAT.app/Contents/Resources/bin/stilts"
 end


PaulPrice:
Hang on; hadn't checked that the binaries actually work. They don't find the `.jar`... ):

PaulPrice:
Regarding the binaries, I've filed an issue upstream (it needs to look in `/Applications` for the `.jar` file).
Regarding the failing check, I have no idea what's going on. Travis says the style check failed, but it works for me:
```
price@Pauls-MacBook-Pro:/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask (topcat-20181016=) $ brew cask style topcat

```
Please advise?

PaulPrice:
Well, looks like the check is passing now... magic?
Ready to merge, if you please.

PaulPrice:
That is correct. This will be fixed in the next upstream release (expected in a couple of weeks).
We could hold off merging this until the next upstream release?

PaulPrice:
Sorry, I misread your question.

Currently, the binaries don't work *because* `TOPCAT.app` is installed into `/Applications` and the binaries are under `/usr/local`, and they don't look outside their own directory tree for the `.jar` file in `TOPCAT.app`. That's what will be fixed in the next release.
Based on the changes pending in `TOPCAT.app`, the binaries will not work if `TOPCAT.app` is not installed under `/Applications` (because it hardwires `/Applications` in its search path for OSX). Can the app be installed elsewhere?

reitermarkus:
> Can the app be installed elsewhere?

Yes, if you specify the `--appdir=` flag.

PaulPrice:
Hmm. Can we put the binaries in the installed `TOPCAT.app` (next to the `.jar` file, so it can find it easily) and then use a forwarding script in `/usr/local/bin`? Or `sed` the binaries to look in the appropriate directory?

reitermarkus:
The easiest would be to have the binaries inside the `.app`, then we can symlink them. That needs to happen upstream, however.

PaulPrice:
I've updated this with a pre-release from Upstream that includes the binaries inside the `.app`, and symlinking those into `/usr/local/bin`. This procedure works for me!
This should NOT be merged until the next official release, which I'm expecting within a week or two.

Casks/topcat.rb
```diff
@@ -1,11 +1,18 @@
...
+  url 'ftp://andromeda.star.bris.ac.uk/pub/star/topcat/pre/topcat-full_script.dmg'
   appcast 'http://www.star.bris.ac.uk/~mbt/topcat/sun253/versions.html'
   name 'TOPCAT'
   homepage 'http://www.star.bris.ac.uk/~mbt/topcat/'
 
   app 'TOPCAT.app'
+
+  postflight do
+    File.symlink("#{appdir}/TOPCAT.app/bin/topcat", "#{HOMEBREW_PREFIX}/bin/topcat")
+    File.symlink("#{appdir}/TOPCAT.app/bin/stilts", "#{HOMEBREW_PREFIX}/bin/stilts")
```
reitermarkus: Use the `binary` stanza for this.

mbtaylor:
I (upstream author) have prepared the application with the startup scripts in TOPCAT.app/bin.  As above, this has been tested and seems to work, so will be ready to go once I have made an upstream public release.

However, somebody told me that a more appropriate location for these startup scripts would be TOPCAT.app/Contents/MacOS.  Is that true, and a well-enough established convention that I ought to be following it?  I don't know anything about what people expect to see where in these application bundles.  If there's somebody here who is well-informed on this, and thinks that I ought to move the scripts into the MacOS directory instead, I'll do it, but if not I'll just leave them where they are.  Thanks for any advice.

vitorgalvao:
> somebody told me that a more appropriate location for these startup scripts would be TOPCAT.app/Contents/MacOS.

I’m not a native macOS developer, but I think that is incorrect. From studying several macOS app bundles, it seems that directory is supposed to contain only the binary to start the app. A better place would likely be `TOPCAT.app/Contents/Resources` (and inside here a `bin`, if you’d like).

But having them in `TOPCAT.app/bin` definitely feels off. I’ve never encountered a macOS app bundle with anything but a `Contents` directory at the top level.

PaulPrice:
@mbtaylor has released the version with the bin scripts in the correct places, so I have bumped the version and sha256. I believe this is ready for merge now.

PR #53749, opened 2018-10-18
Update fluid to 2.1

commitay:

### DIFF
--- a/Casks/fluid.rb
+++ b/Casks/fluid.rb
@@ -1,6 +1,6 @@
 cask 'fluid' do
   version '2.1'
-  sha256 '7b27dfcec1b74209496eb1150f38e464c95ae695dc9d95f1eab3d11925189e58'
+  sha256 'dea3e4c93f679f227594fa1cb53f5e78880e879fa796aa6396641dc4a1962195'
 
   url "https://fluidapp.com/dist/Fluid_#{version}.zip"
   appcast "https://fluidapp.com/appcast/fluid#{version.major}.rss"


dullin:
This is getting weird. The site now has only version 1.8.6 available which seems clean.
[VirusTotal of 1.8.6](https://www.virustotal.com/#/url/f65c21837823049fa9fb2eee47d4828834da6466b2aa5c7ed047246b1536c3ba/detection)

itod:
Thank you for the heads up!

First, when I was made aware of this issue this morning via email, I immediately pulled Fluid 2 and reverted my website to the latest version of Fluid 1 out of an abundance of caution. That's just a matter of me doing everything I can to revert to a known safe state, while I investigate. Just trying to address this as quickly as I can.

Second, I believe both issues are actually false positives, but I would appreciate feedback. Thank you!

(One thing to keep in mind: Although I understand the purpose and function of Brew Cask, I've never used it before, and am not familiar with any particulars.)

AFAICT, there are two apparent issues:

===

1. The current Fluid 2.1 checksum does not match the checksum expected by Brew Cask.

Again, I'm no Cask expert, but I suspect this is because I've released a few very minor updates under the name "Fluid 2.1".

To be specific, The "2.1" part is the "CFBundleShortVersionString". However, each update I have released with the "2.1" "CFBundleShortVersionString" has included a *new, incremented* "CFBundleVersion", which is the more granular version number in Apple's version numbering scheme. The latest "CFBundleVersion" for Fluid 2.1 is "2109". You can see these full version number distinctions in the "About" box of Fluid and Fluid Apps you've built.

I suspect Cask tracks app versions by the former (CFBundleShortVersionString), but not the latter (CFBundleVersion)? I'm not sure if that's the best approach for Cask to take, but if that is the issue, I will no longer do that, just to avoid problems with Cask. I will release each Fluid version with a new, incremented "CFBundleShortVersionString" from now on, just to avoid this issue.

AFAICT, that will "solve" this issue going forward. Please correct me if I'm wrong.

===

2. Virustotal.com is reporting "two engines detected." 

I welcome all input on this, but I highly suspect these are false positives.

I've been able to track down the "two engines detected" to two specific items in the Fluid_2.1.zip package.

One of the "engines" detected is no longer detected if you remove a single file from the Fluid_2.1.zip package: a PEM file (fluid2_dsa_pub.pem) which is a text file which only contains the public key used for verifying app updates downloaded via the Sparkle Update Framework. This file clearly does not contain malware. Here's the file's source:

https://gist.github.com/itod/45d39952230e9d962e3355c734b50c12

The other "engine" detected is no longer detected if you remove the "Sparkle.framework" embedded framework - built from Sparkle 1.19.0 (released in Apr 2018, this was the very latest stable release of Sparkle when I began serious work on Fluid 2 – it has since been superseded by a single Sparkle stable release). This embedded Sparkle Framework does include a few minor code changes made by me to support Fluid Apps retaining their custom icon when being Sparkle updated. It seems my minor code changes are triggering this "engine detection". Uploading a version of Fluid to Virustotal.com built with stock Sparkle 1.19.0 embedded removes the second engine detected. But of course I don't understand why my small code change should be flagged as malware. It's not. And that small code change also existed in Fluid 1.x versions (which are not flagged as containing "engines").

So… I welcome all input. Any input on how likely false positives are to occur in the Virus scanning software run by Virustotal.com –– (specifically: Ikarus and Qihoo-360)?

commitay:
> I suspect Cask tracks app versions by the former (CFBundleShortVersionString)

We parse the cask version from the URL or from the appcast if it contains an additional version that gets incremented that we can see.

> I will release each Fluid version with a new, incremented "CFBundleShortVersionString" from now on, just to avoid this issue.

> AFAICT, that will "solve" this issue going forward.

Yes, it will. Thanks!

> Virustotal.com is reporting "two engines detected."

> I welcome all input on this, but I highly suspect these are false positives.

Yes, we occasionally do see similar false positives. 

As the app bundle is codesigned and you're aware of the positives we can merge this if you want or we can revert the cask to `1.8.6` to match the version on the homepage.

commitay:
Updated to `2.1.1` in https://github.com/Homebrew/homebrew-cask/pull/53790, `2.1.1` is the current version on the homepage.

PR #53758, opened 2018-10-18
Update mongoproc to 3.1,3.1.1

core-code:

### DIFF
--- a/Casks/mongoproc.rb
+++ b/Casks/mongoproc.rb
@@ -1,9 +1,9 @@
 cask 'mongoproc' do
-  version '3.1,3.1.1'
+  version '3.1.1'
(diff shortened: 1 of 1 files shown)


Casks/mongoproc.rb
```diff
@@ -1,9 +1,9 @@
 cask 'mongoproc' do
-  version '3.1'
-  sha256 'cc4d2dfc94ae8df014ac0e317663e074e5d128c61df851e19da0fec9912d2a0e'
+  version '3.1,3.1.1'
```
reitermarkus: Why not just use `3.1.1`?

core-code: how would you make sure it uses just 3.1 before the last slash?

reitermarkus: `version.major_minor`

core-code: ah sorry forgot about that option ^^

PR #53761, opened 2018-10-18
Update stattransfer to 14

core-code:


core-code:
its up to version 14.1 now but they update it in-place

reitermarkus:
Since there is no appcast, this should be `latest`.

reitermarkus:
> It needs a version for the pkg install.

I don't think that should be a reason to use `version`. There is no `appcast` and the `url` is unversioned.

vitorgalvao:
> > It needs a version for the pkg install.
> 
> I don't think that should be a reason to use `version`.

We have precedent. Other casks in this situation have it like this one is.

PR #53775, opened 2018-10-18
Update ssh-tunnel-manager to 227.006,1478083232

core-code:


core-code:
well, you shouldn't change version without SHA

but the version / link that was used previously can be seen neither on their homepage nor in the appcast, which is kinda suspicious

PR #53492, opened 2018-10-12
Secure URLs

vszakats:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
---

This was done by checking for in-place protocol upgrades, following permanent redirects to HTTPS and upgrading to known-good secure alternative (for sf.net URL). For Adobe downloads, the URLs were harmonised to use the same domain and root path for all Flash and sister products.

PR #53786, opened 2018-10-19
Secure URLs

vszakats:


vszakats:
Same as https://github.com/Homebrew/homebrew-cask/pull/53492, but with separate commit for each Cask.

PR #53879, opened 2018-10-19
Update setapp to 1.18.1,1539860348

brianbrownton:

### DIFF
--- a/Casks/setapp.rb
+++ b/Casks/setapp.rb
@@ -1,6 +1,6 @@
 cask 'setapp' do
-  version '1.17,1537866884'
-  sha256 '5ce806cd0f6960e10ccc9759587fce49e8dc1ac3c46348d71738434aad6db65c'
+  version '1.18.1,1539860348'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
We still need an `uninstall`.

brianmorton:
Well that didn't go as I thought it would. I thought if I put something nonsensical in the `uninstall` that travis would tell me what actually needed to be installed... instead it passed. What would you like to do now?

commitay:
It's `installer manual`, the script doesn't exist on Travis but it will for users?

brianmorton:
@commitay when I unpacked it manually it didn't exist

commitay:
The script does exist after running the installer.

@Homebrew/cask I think we might need to come up with a way to get casks that will always fail install/uninstall to play nice on CI, maybe blacklisting stuff like this, `virtualbox`, etc so it only checks audit and style.

brianmorton:
So do you want to update this PR (restore it maybe) to what it should be and open a separate issue for blacklisting these type of casks from travis?

PR #53916, opened 2018-10-19
Update textricator to 9.0.44

brianbrownton:


brianmorton:
Changed the `appcast` away from github as the github releases always seemed to be ahead of what was on the site. I am happy to switch it back if we would rather use the latest releases from github instead of the public site.

PR #53926, opened 2018-10-20
Fix cask testing instructions

OndroNR:
Documentation fix. `brew cask install {{token}}` checkouts master branch and updates it, so I can't test already committed in feature branch, but not pushed updated cask. Env variable `HOMEBREW_NO_AUTO_UPDATE` prevents this git update.


reitermarkus:
Thanks!

PR #53930, opened 2018-10-21
 brewcask-ci: skip install/uninstall 

commitay:
```
$ brew cask ci
==> brew cask audit virtualbox ✔
==> brew cask style virtualbox ✔
Error: skipping install/uninstall of virtualbox
```

### DIFF
--- a/cmd/brewcask-ci.rb
+++ b/cmd/brewcask-ci.rb
@@ -48,6 +48,22 @@ module Cask
             Style.run(path)
           end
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
While we’re at it, we should add a skip to `installer manual:` casks.

commitay:
I'll see about checking dependencies so we don't need to list every cask that has a direct or indirect dependency on `virtualbox`, etc.

PR #53973, opened 2018-10-22
Update cloudcompare to 2.9.1

herrriehm:


vitorgalvao:
Since the URL isn’t versioned and this has no `appcast`, having it this way would break the cask on every update.

PR #54058, opened 2018-10-23
Update origin to latest

yuusharo:

### DIFF
--- a/Casks/origin.rb
+++ b/Casks/origin.rb
@@ -4,7 +4,6 @@ cask 'origin' do
 
   # origin-a.akamaihd.net was verified as official when first introduced to the cask
   url 'https://origin-a.akamaihd.net/Origin-Client-Download/origin/mac/live/Origin.dmg'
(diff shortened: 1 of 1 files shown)


yuusharo:
What is the reason for adding a shasum check when the download url doesn't change? I can't install or reinstall Origin whenever the client updates server side due to a failed hash check.

Is this necessary?

yuusharo:
It appears Origin does offer versioned downloads. Closing this PR, submitting a new PR with an updated versioned URL.

yuusharo:
Sorry, the versioned URL is only for updates, not for client downloads. Reopening PR.

core-code:
i don't think the cask is broken this often. its updated every few weeks and i have a script to observe it so it gets repaired quickly.

btw, the version downloads are identically to the full downloads, except they only contains the 'contents' of the app. all that would be necessary to use them is to create a script to create the enclosing 'Origin.app' folder.

PR #54158, opened 2018-10-27
Update vuescan to 9.6.20

lenlo:

### DIFF
--- a/Casks/vuescan.rb
+++ b/Casks/vuescan.rb
@@ -1,8 +1,8 @@
 cask 'vuescan' do
-  version '9.6'
+  version '9.6.20'
   sha256 :no_check # required as upstream package is updated in-place
 
-  url "https://www.hamrick.com/files/vuex64#{version.no_dots}.dmg"
+  url "https://www.hamrick.com/files/vuex64#{version.major_minor.no_dots}.dmg"
   appcast 'https://www.hamrick.com/alternate-versions.html'
   name 'VueScan'
   homepage 'https://www.hamrick.com/'


lenlo:
I understand why the checksum check got dropped a while back as VueScan is usually is updated in-place on the download site, but why was the patch level dropped too? Without that, I don't see how anyone will be able to upgrade vuescan automatically. (So I put it back.)

commitay:
Without a checksum the download will continue to work with `9.6.20` even after it is updated upstream to `9.6.21`, i.e we won't know that the version has actually changed. This would mean that users could install `9.6.21` as `9.6.20` and then be "upgraded" to `9.6.21` when we change the patch number.

lenlo:
I'm sorry, I don't quite understand your comment. Are you concerned that someone with a ``9.6.20`` formula would get the ``9.6.21`` version instead by accident?

My concern is that without a patch level in the formula, ``brew cask upgrade vuescan`` won't perform any upgrades until the minor version changes. This is pretty rare, maybe only once a year or so, while patch level updates may happen every couple of weeks. This means that you no longer can use the ``brew upgrade`` mechanism to keep VueScan up to date. Not even ``brew cask reinstall vuescan`` will do it if you happened to have the previous version downloaded already. You'll have to manually remove the cached download and then reinstall it. This seems really messy to me.

Or am I missing something?

commitay:
My point was it doesn't matter what we set the patch version to, it will always install the current version.

Does Vuescan not provide a way for the app to self update?

lenlo:
Yes, the formula will always install the last patch level regardless of what we set the patch level number to. But how is this a problem?

As I mentioned, my concern is that brew won't be able to upgrade vuescan without including a patch number. For me, this is a real problem.

The app does not do automatic updates itself. It will notice new updates and offer the user to download a new version, but s/he'll have to install it him/herself. But that seems silly since using homebrew to keep my apps up-to-date is the main reasons for me to install them that way.

commitay:
> But how is this a problem?

It forces unnecessary updates for users that already have the current version installed.

> The app does not do automatic updates itself. It will notice new updates and offer the user to download a new version

Right, we'll have to handle it here regardless then.

lenlo:
👍

lenlo:
Sigh, I just noticed that doing ``brew cask upgrade vuescan`` with an updated patch level will reuse any cached copy of the previously downloaded version instead of downloading a new copy since the filename doesn't change. I'll submit a fix/workaround for that too...

PR #54162, opened 2018-10-27
Update vuescan to 9.6.20

lenlo:

### DIFF
--- a/Casks/vuescan.rb
+++ b/Casks/vuescan.rb
@@ -2,7 +2,10 @@ cask 'vuescan' do
   version '9.6.20'
   sha256 :no_check # required as upstream package is updated in-place
 
-  url "https://www.hamrick.com/files/vuex64#{version.major_minor.no_dots}.dmg"
+  # Include a dummy query in the URL to make it unique wrt the patch level.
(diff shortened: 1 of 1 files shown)


lenlo:
> @reitermarkus Shouldn't the downloaded file be unique based on the cask `version`?

That would probably be good. As far as I can tell, it's not today. From `class AbstractFileDownloadStrategy` in `download_strategy.rb`:

```
  def cached_location
    return @cached_location if defined?(@cached_location)

    url_sha256 = Digest::SHA256.hexdigest(url)
    downloads = Pathname.glob(HOMEBREW_CACHE/"downloads/#{url_sha256}--*")
                        .reject { |path| path.extname.end_with?(".incomplete") }

    @cached_location = if downloads.count == 1
      downloads.first
    else
      HOMEBREW_CACHE/"downloads/#{url_sha256}--#{resolved_basename}"
    end
  end
```

I.e. the `cached_location` is computed solely on the URL string.

vitorgalvao:
Yeah, we’re not going to merge this. We’re not going to start adding ugly hacks (and multiple-line comments) to bypass a quirk in the core. Either we fix the core or we don’t, but this is not a solution.

PR #54214, opened 2018-10-29
Fix tableplus 1.0.146 SHA-256

SeanSith:
Per https://tableplus.io/blog/2017/02/changelogs.html the SHA-256 in
the existing cask is incorrect. Also fails to pass the check on download.

### DIFF
--- a/Casks/tableplus.rb
+++ b/Casks/tableplus.rb
@@ -1,6 +1,6 @@
 cask 'tableplus' do
   version '1.0,146'
-  sha256 '7abd9fee3f4c8508c18144e07a1d5528993b4ab4bf6e549885e9ab87e38e2dfb'
+  sha256 '956e92f651f93e2fa9a7cbdf36eec835bc55d887dacdfb6cf42fd009a7f68c00'
 
   # s3.amazonaws.com/tableplus-osx-builds was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


peterjc-123:
Share side request about SHA-256

SeanSith:
I'm not sure what's being asked above, but just in case we're also looking for a VirusTotal scan: https://www.virustotal.com/#/file/956e92f651f93e2fa9a7cbdf36eec835bc55d887dacdfb6cf42fd009a7f68c00/detection

It does not contain a signature section, unfortunately.

vitorgalvao:
> I'm not sure what's being asked above

I’m not either. Not sure if that’s a spam account.

@peterjc-123 This is the second post of yours I’d had to delete. If there’s a third, you’ll be banned from Homebrew.

vitorgalvao (changes requested):
The VirusTotal verification ins’t enough because there’s no signing info. We need developer confirmation, as per the instructions.

SeanSith:
@huyphams Any chance we could get you over here to verify that the SHA checksum mentioned in the commit I made above is correct based on your blog post?

huyphams:
Thanks @SeanSith I have accidentally updated the wrong version of TablePlus (which is for my partner Setapp), but I realize and re-upload it, sorry because the confusing :(
The changes log and the SHA can always be found here: 
https://tableplus.com/osx/changelog

vitorgalvao:
Thank you for the confirmation, @SeanSith.

PR #54266, opened 2018-10-31
Update origin to 10.5.30.15625

core-code:

### DIFF
--- a/Casks/origin.rb
+++ b/Casks/origin.rb
@@ -1,6 +1,6 @@
 cask 'origin' do
-  version '10.5.29.14153'
-  sha256 '2d5adbc292c4a1947a50b47ddf547bbcf19e8ba450824ba99b49f0d6db08f363'
+  version '10.5.30.15625'
+  sha256 'd679a1635957edcc677b4f00f9a6d3fb236b1af8d401306999085b3e1c8cfeab'
 
   # origin-a.akamaihd.net was verified as official when first introduced to the cask
   url 'https://origin-a.akamaihd.net/Origin-Client-Download/origin/mac/live/Origin.dmg'


core-code:
the appcast seems to be out-of-sync with the download, which could explain some of the breakage that @yuusharo alluded to in #54058 
would anyone like to have a try to use the download that is linked in the appcast? this would prevent any breakage of this cask. AFAIK all that is needed to do is to create an enclosing Origin.app folder with a script

reitermarkus:
The download in the appcast seems to indeed be the full version, except for the enclosing `Origin.app` directory.

core-code:
who is the local preflight scripting guru around here that could convert the cask to use the download from the appcast?

javierjulio:
I'm not able to update to this release. I've tried updating as I use brew-bundler. I tried `brew cask reinstall origin` which in the past case has worked but now each time I can't get it to update Origin.

```
==> Upgrading 1 outdated package:
origin 10.5.26.8487 -> 10.5.30.15625
==> Satisfying dependencies
==> Downloading https://origin-a.akamaihd.net/Origin-Client-Download/origin/mac/live/Origin.dmg
Error: Checksum for Cask 'origin' does not match.

Expected: d679a1635957edcc677b4f00f9a6d3fb236b1af8d401306999085b3e1c8cfeab
Actual:   dbe7f76d52c806dc6f3a1187b3f5b015abc4c32d250c093e8b2c9ff5b1a30baf
File:     /Users/javierjulio/Library/Caches/Homebrew/downloads/3246e05932616d37e5ef05e7d94bbd97316c02002682ed130a025ea94592795f--Origin.dmg

==> Verifying SHA-256 checksum for Cask 'origin'.
==> Note: Running `brew update` may fix SHA-256 checksum errors.
==> Purging files for version 10.5.30.15625 of Cask origin
Installing origin has failed!
```

Even with removing the mentioned file above I continue to get the same error.

core-code:
it seems the appcast is out-of-sync with the download again. as noted this would be solved by using the download from the appcast but so far no-one has offered some shell-fu to implement this.

Issue #54117, opened 2018-10-25
Casks/jdownloader.rb symlink instead of app / Spotlight problem


ulab:
### Description of feature/enhancement

After installing the jdownloader cask I noticed that I can not find and open it using Spotlight. Reading through issues I learned that it was a problem with symlinking apps, but other casks worked fine.

Main difference seems to be that other cask are moved to /Applications, but jdownloader is not.

I have no experience with writing Casks, but it seems jdownloader is missing the "app" stanza?

### Justification

Using Spotlight is rather convenient.

### Example use case

`cmd` + `space` `jdownloader` => nothing :-(

alagos:
Same story here. The way I use to run jd2 is going to finder -> applications -> jd2, but it would be way quicker to run it with spotlight.

vitorgalvao:
[Try this](https://github.com/Homebrew/homebrew-cask/pull/54628/files). Does it work?

alagos:
@vitorgalvao I've just tried installing with 
```
$ brew cask uninstall https://raw.githubusercontent.com/Homebrew/homebrew-cask/a0fbb0a7bd1a596e1f2103a1ed2a2f1711422c08/Casks/jdownloader.rb
==> Downloading https://raw.githubusercontent.com/Homebrew/homebrew-cask/a0fbb0a7bd1a596e1f2103a1ed2a2f1711422c08/Casks/jdownloader.rb.
######################################################################## 100.0%
==> Caveats
jdownloader requires Java 8. You can install it with

  brew cask install homebrew/cask-versions/java8

==> Satisfying dependencies
==> Downloading http://installer.jdownloader.org/clean/JD2Setup.dmg
######################################################################## 100.0%
==> No SHA-256 checksum defined for Cask 'jdownloader', skipping verification.
==> Installing Cask jdownloader
🍺  jdownloader was successfully installed!
```
and it works in spotlight. But if I try to uninstall it with:
```
brew cask uninstall https://raw.githubusercontent.com/Homebrew/homebrew-cask/a0fbb0a7bd1a596e1f2103a1ed2a2f1711422c08/Casks/jdownloader.rb
==> Downloading https://raw.githubusercontent.com/Homebrew/homebrew-cask/a0fbb0a7bd1a596e1f2103a1ed2a2f1711422c08/Casks/jdownloader.rb.
######################################################################## 100.0%
==> Uninstalling Cask jdownloader
==> Running uninstall process for jdownloader; your password may be necessary
[...]

PR #54628, opened 2018-11-08
jdownloader.rb: install to appdir

vitorgalvao:
Fixes https://github.com/Homebrew/homebrew-cask/issues/54117.

### DIFF
--- a/Casks/jdownloader.rb
+++ b/Casks/jdownloader.rb
@@ -20,7 +20,7 @@ cask 'jdownloader' do
                    print_stderr: false
   end
 
-  uninstall delete: "#{appdir}/JDownloader2.app"
+  uninstall delete: "#{appdir}/JDownloader 2.0.app"
 
   caveats do
     depends_on_java '8'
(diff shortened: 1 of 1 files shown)


Casks/jdownloader.rb
```diff
@@ -20,7 +20,7 @@
                    print_stderr: false
   end
 
-  uninstall delete: '/Applications/JDownloader2.app'
+  uninstall delete: "#{appdir}/JDownloader 2.0.app"
```
reitermarkus: It needs to remove `#{appdir}/JDownloader 2.0`.

vitorgalvao: @reitermarkus Are you saying it‘s a directory, and not an app bundle?

reitermarkus: Both actually, `#{appdir}/JDownloader 2.0` and a symlink from `#{appdir}/JDownloader2.app` to `#{appdir}/JDownloader 2.0/JDownloader2.app`.

PR #54634, opened 2018-11-08
Clarify docs about testing brand-new casks with relative path

cjcjameson:
docs change only


vitorgalvao:
There’s no need for this. You don’t need the full path for it to work.

cjcjameson:
So, hm... it definitely didn't work for me without properly specifying the path. I tried it a couple times but still might not have been doing quite the right thing.

cjcjameson:
Aha! Maybe that explains it. I copy-pasted from an existing cask rather
than timing the init script.

PR #54655, opened 2018-11-08
Update BBEdit to 12.5

chrisfinazzo:

### DIFF
--- a/Casks/bbedit.rb
+++ b/Casks/bbedit.rb
@@ -1,6 +1,6 @@
 cask 'bbedit' do
-  version '12.1.6'
-  sha256 '23b9fc6ef5c03cbcab041566503c556d5baf56b2ec18f551e6f0e9e6b48dc690'
+  version '12.5'
+  sha256 'fa346e6bc3ab98bcecddcfc86c024d4bf7aca1817290c6b50baf59382e55c59b'
 
   # s3.amazonaws.com/BBSW-download was verified as official when first introduced to the cask
   url "https://s3.amazonaws.com/BBSW-download/BBEdit_#{version}.dmg"
@@ -8,7 +8,7 @@ cask 'bbedit' do
   name 'BBEdit'
   homepage 'https://www.barebones.com/products/bbedit/'
 
-  depends_on macos: '>= :el_capitan'
+  depends_on macos: '>= :sierra'
 
   app 'BBEdit.app'
 
--- a/Casks/bbedit.rb
+++ b/Casks/bbedit.rb
@@ -1,24 +1,25 @@
-cask "bbedit" do
+cask 'bbedit' do
   if MacOS.version <= :el_capitan
-    version "12.1.6"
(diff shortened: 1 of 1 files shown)


Casks/bbedit.rb
```diff
@@ -1,21 +1,24 @@
...
   url "https://s3.amazonaws.com/BBSW-download/BBEdit_#{version}.dmg"
-  appcast 'https://versioncheck.barebones.com/BBEdit.xml'
-  name 'BBEdit'
-  homepage 'https://www.barebones.com/products/bbedit/'
+  appcast "https://versioncheck.barebones.com/BBEdit.xml"
+  name "BBEdit"
+  homepage "https://www.barebones.com/products/bbedit/"
 
-  depends_on macos: '>= :el_capitan'
+  depends_on :macos => ">= :sierra"
```
reitermarkus: @MikeMcQuaid, `brew cask style` is not reporting this anymore, it should be `depends_on macos: '>= :sierra'`.

reitermarkus: It apparently also corrected the all single to double quotes.

chrisfinazzo: @reitermarkus Good to see someone else notice this - honestly would have expected it to be defined in a `.rubocop.yml`.

MikeMcQuaid: @reitermarkus I don't think it's possible with the current setup to enforce contradictory styles for Homebrew/brew and formulae vs. cask styles. I added a TODO to investigate getting them closer in sync but for now to get `brew cask style` passing everywhere Casks have just been excluded from contradictory rules.

reitermarkus: The best way I could think of would be to have a `.rubocop.yml` in each tap with

```yaml
inherit_from: ../../Homebrew/.rubocop_cask.yml
```

or

```yaml
inherit_from: ../../Homebrew/.rubocop_formula.yml
```

This way the Formula rules also don't conflict with the core code.

MikeMcQuaid: @reitermarkus Seems like the only option for now although obviously the duplication is regrettable. Long-term I think we should aim to have the styles more in sync, though.

core-code: is there anything preventing this PR from being merged?

core-code: i think we should just use the style that any and all other casks use? or is there something i don't get?

reitermarkus: > this should probably be using `else` as opposed to `elsif` 

Yes, use `else`.

> Other than that, the only thing is whether Rubocop’s changes should be reverted.

Yes, please revert these.

chrisfinazzo: OK, will update.

chrisfinazzo:
Resolved in #54812. I'll open a separate PR for the if-else statement.

core-code:
thanks @chrisfinazzo i already did so in https://github.com/Homebrew/homebrew-cask/pull/54814

PR #54658, opened 2018-11-08
Update SpiderOakONE

g-a-c:
This corrects the Cask's behaviour to install the SpiderOakONE package appropriate for the user's version of macOS

* Correctly set the minimum supported version to Mavericks
* Correctly format the url conditionals such that High Sierra is the only version with a specific URL. Both older and newer macOS versions use the "default" URL
* Send a SIGTERM before uninstalling with ‘pkgutil:’ (the `org.python.python` package ID is correct for this package - this is, I assume, an upstream bug that homebrew would need to work with)
* add support for ‘brew cask zap’ in case the Cask needs to be permanently removed, or the user wants to reinstall as a brand new device

- [x] The commit message includes the cask’s name and version (unversioned Cask).

### DIFF
--- a/Casks/spideroakone.rb
+++ b/Casks/spideroakone.rb
@@ -4,7 +4,7 @@ cask 'spideroakone' do
 
   if MacOS.version <= :sierra
     url 'https://spideroak.com/release/spideroak/osx'
-  elsif MacOS.version == :high_sierra
+  elsif MacOS.version <= :high_sierra
     url 'https://spideroak.com/release/spideroak/osx_hs'
   else
     url 'https://spideroak.com/release/spideroak/osx'
--- a/Casks/spideroakone.rb
+++ b/Casks/spideroakone.rb
@@ -4,7 +4,7 @@ cask 'spideroakone' do
 
   if MacOS.version <= :sierra
     url 'https://spideroak.com/release/spideroak/osx'
-  elsif MacOS.version <= :high_sierra
+  elsif MacOS.version == :high_sierra
     url 'https://spideroak.com/release/spideroak/osx_hs'
   else
     url 'https://spideroak.com/release/spideroak/osx'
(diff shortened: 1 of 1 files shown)


Casks/spideroakone.rb
```diff
@@ -4,15 +4,22 @@
 
   if MacOS.version <= :sierra
     url 'https://spideroak.com/release/spideroak/osx'
-  else
+  elsif MacOS.version <= :high_sierra
     url 'https://spideroak.com/release/spideroak/osx_hs'
+  else
+    url 'https://spideroak.com/release/spideroak/osx'
```
reitermarkus: This is the same as for `<= :sierra`.

g-a-c: Correct, High Sierra is the only version that needs the `osx_hs` separate download URL - anything older or newer than High Sierra uses the regular `osx` download URL.

reitermarkus: Use `if MacOS.version == :high_sierra` then.

g-a-c: I did - you changed it a few days ago to be `<= :high_sierra`, which I can revert. In the first pull request, this was a single if/else, which seemed to cause confusion. In this pull request, it now reads more logically oldest->newest, at the cost of having some duplication. This has also seemed to cause confusion. Both are equally readable to me, but as a project maintainer, which does the project prefer and I'll format accordingly?

```
if MacOS.version == :high_sierra
    url 'https://spideroak.com/release/spideroak/osx_hs'
else
    url 'https://spideroak.com/release/spideroak/osx'
end
```
or
``` 
if MacOS.version <= :sierra
    url 'https://spideroak.com/release/spideroak/osx'
elsif MacOS.version == :high_sierra
    url 'https://spideroak.com/release/spideroak/osx_hs'
else
    url 'https://spideroak.com/release/spideroak/osx'
end
```

reitermarkus: I think it's confusing either way, so maybe the first one with a comment explaining that indeed this only applies to High Sierra.

vitorgalvao (approved):

Issue #47351, opened 2018-05-17
List of additional (Sparkle) AppCasts


core-code:
as per discussion in pull-request #47289 i'm attaching a text file with many additional appcast URLs that are not obvious (i.e. not stored in the Info.plist => SUFeedURL) of common apps

the format is like this
BUNDLE_IDENTIFIER|SU_FEED_URL_FROM_PLIST|ACTUAL_WORKING_FEED_URL

if you prefer to have the file with caskfile-names instead of bundle-identifierts, i'm sure i could provide an alternative version with some work.
[overridesparklefeedurls.txt](https://github.com/caskroom/homebrew-cask/files/2014478/overridesparklefeedurls.txt)

vitorgalvao:
> if you prefer to have the file with caskfile-names instead of bundle-identifierts

That would be preferred, yes.

Also, you can just paste the info on the post, no need to attach a file.

core-code:
ok here are those with casknames:

```
CNAME:=geniatech-eyetv|(null)|http://updates.geniatech.eu/autoupdate/eyetv3.rss
CNAME:5kplayer|(null)|http://www.5kplayer.com/upgrade/mac/updatecast.xml
CNAME:acorn|(null)|https://flyingmeat.com/download/latest/acorn6update.xml
CNAME:airflow|(null)|https://s3.amazonaws.com/Airflow/Updates/appcast.xml
CNAME:airfoil|(null)|https://rogueamoeba.net/ping/versionCheck.cgi?format=sparkle&bundleid=com.rogueamoeba.Airfoil&system=10134&platform=osx&arch=x86_64&version=5738000
CNAME:airfoil|(null)|https://rogueamoeba.net/ping/versionCheck.cgi?format=sparkle&bundleid=com.rogueamoeba.AirfoilSpeakers&system=10134&platform=osx&arch=x86_64&version=5738000
CNAME:archiver|SET_PROGRAMMATICALLY_BY_APPLICATION_HELPER|https://api.incrediblebee.com/appcasts/archiver-3.xml
CNAME:arq|(null)|https://www.arqbackup.com/download/arq5.xml
CNAME:bookmacster|(null)|https://sheepsystems.com/bookmacster/appcast.xml
CNAME:ceceree|(null)|https://updates.devmate.com/com.appiculous.Ceceree.xml
CNAME:chatmate-for-facebook|(null)|https://updates.devmate.com/net.coldx.mac.Facebook.xml
[...]

PR #54715, opened 2018-11-09
add valid sparkle appcasts for casks with missing appcasts

core-code:
as per "discussion" (well, just me talking;) here ( https://github.com/Homebrew/homebrew-cask/issues/47351 ) there are a lot of casts that have no 'appcast' stanza although there is a valid sparkle appcast for them. i've tried to add them all at once.

### DIFF
--- a/Casks/acronis-true-image.rb
+++ b/Casks/acronis-true-image.rb
@@ -3,6 +3,7 @@ cask 'acronis-true-image' do
   sha256 :no_check # required as upstream package is updated in-place
 
   url "https://dl.acronis.com/u/AcronisTrueImage#{version}.dmg"
+  appcast 'https://dl.acronis.com/u/liveupdate/%7B2CE2087A-349A-4F8A-9919-6AF8B53D8F99%7D/appcast.xml'
   name 'Acronis True Image'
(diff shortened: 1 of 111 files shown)


vitorgalvao:
Many of these require an update to `version` and `sha256` (no longer `:latest` and `:no_check`).

core-code:
well, the appcasts don't hurt even if you don't have a version ;)

core-code:
also, feel free to merge just those that are currently versioned

vitorgalvao:
> well, the appcasts don't hurt even if you don't have a version

True, but I’d prefer to have everything done right from the start. Otherwise we’ll eventually get bad submissions that’ll use some of these we haven’t updated yet as a base.

core-code:
of course i also prefer things to be done right, but, well, sorry but i wont update 60 casks to be versioned.

i'd recommend to either merge all of them (so that the appcasts are there once they are upgraded)
or merge just those that already are versioned to comply with current rules ;)

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/issues/54811

core-code:
thanks ;) i will see if i can do some of  those when i have a few minutes

Issue #36162, opened 2017-07-04
brew cask zap - trash error 


commitay:
#### General troubleshooting steps

- [x] I have checked the instructions for [reporting bugs](https://github.com/caskroom/homebrew-cask#reporting-bugs) (or [making requests](https://github.com/caskroom/homebrew-cask#requests)) before opening the issue.
  - [x] None of the templates was appropriate for my issue, or I’m not sure.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).

#### Description of issue

`brew cask zap` errors when trashing. Repeating the same `zap` succeeds.

#### Output of your command with `--verbose --debug`

```
brew cask zap musicbrainz-picard --verbose --debug
==> Zapping Cask musicbrainz-picard
==> Implied "brew cask uninstall musicbrainz-picard"
==> Un-installing artifacts
==> Determining which artifacts are present in Cask musicbrainz-picard
==> 2 artifact/s defined
#<Hbc::Artifact::App:0x007fcbec8260d0>
#<Hbc::Artifact::Zap:0x007fcbec825fb8>
==> Un-installing artifact of class Hbc::Artifact::App
==> Removing App '/Applications/MusicBrainz Picard.app'.
==> Dispatching zap stanza
==> Running zap process for musicbrainz-picard; your password may be necessary
==> Removing files:
~/Library/Caches/MusicBrainz
==> Executing: ["/usr/bin/sudo", "-E", "--", "/usr/bin/xargs", "-0", "--", "/bin/rm", "-r", "-f", "--"]
==> Trashing files:
~/.config/MusicBrainz
[...]

PR #54739, opened 2018-11-10
Update datadog-agent to 6.6.0-1

suschizu:

### DIFF
--- a/Casks/datadog-agent.rb
+++ b/Casks/datadog-agent.rb
@@ -12,7 +12,7 @@ cask 'datadog-agent' do
 
   preflight do
     require 'etc'
-    File.open('/tmp/datadog-install-user', 'w') { |f| f.write(Etc.getlogin) }
+    File.write('/tmp/datadog-install-user', Etc.getlogin)
   end
 
   uninstall launchctl: 'com.datadoghq.agent',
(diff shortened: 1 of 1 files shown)


suschizu:
I’ve added all suggestions from travis but I still receive an error when uninstalling the app - could somebody perhaps have a look on that? Thank you!

vitorgalvao:
Ping @reitermarkus.

This kind of thing has been happening a lot. Are these things we need to fix in the core, or are they Travis-related?

ran-dall:
@vitorgalvao I had a question/suggestion. Couldn't we use something like what you suggested here: https://github.com/Homebrew/homebrew-cask/issues/36162#issuecomment-312849549

core-code:
so, is there any way to get this un-stuck?

core-code:
i have installed the package locally and can not reproduce the travis error. the files travis complains are missing *are* actually installed. so i believe the failure is nonsense.

in any case, why does travis complain if files that the package installs are not there? its very common that PKGs do that.

for example, this very package claims to install the application to /opt/datadog-agent/ , but actually its moved to /Applications in the postinstall. so, if travis wouldn't trip earlier, it would trip there.

because of the move-around in the postinstall, for this cask to be correct it should have an 
`uninstall delete /Applications/Datadog Agent.app`

core-code:
ok, i've installed the updated cask, and indeed we get the error that travis claimed on uninstall ;(

the trouble is that the package is basically broken. they list files that only differ in case, and of course HFS/APFS is case insensitive. after deleting the first of them, the second one is not there. thats why my analysis was off - the file is there - but it won't be after deleting another one which differs just in case.

is there a way to have 'uninstall pkgutil' succeed even when some files aren't there? i see erroring out in this case as terribly un-useful.

ran-dall:
@core-code Very good work!

PR #54761, opened 2018-11-11
Update parallels to 14.0.1-45154

core-code:

### DIFF
--- a/Casks/parallels.rb
+++ b/Casks/parallels.rb
@@ -3,6 +3,7 @@ cask 'parallels' do
   sha256 '2d3157fa684c9e255927ae4a04f303107427a3ea166eea6ea86f0963cb24e4bb'
 
   url "https://download.parallels.com/desktop/v#{version.major}/#{version}/ParallelsDesktop-#{version}.dmg"
+  appcast 'http://www.corecode.io/cgi-bin/check_url_redirect/check_url_redirect.cgi?url=https://www.parallels.com/directdownload/pd/'
   name 'Parallels Desktop'
   homepage 'https://www.parallels.com/products/desktop/'
 


core-code:
ok this is an experiment. there are many casks where we have no appcast and there is no version information on the homepage or anywhere else at all BUT the version information is basically contained in the HTML redirect they make for their download.
normally you click on
download.dmg
but are redirected to
download.realversionnumber.dmg

i've written a CGI script and placed it on my server that returns the destination of an URL redirect.
entering this as an appcast should always yield the correct version number (unless the download URL changes)

core-code:
i was mainly thinking about this to support @brianmorton scripts to automatically update casks. if this works with, i am all for it.

core-code:
also, i can't get "appcast: referer" to work at least not when using cask-repair:

Error: Cask 'minecraftpe' definition is invalid: 'appcast' stanza failed with: bad argument (expected URI object or URI string)

reitermarkus:
`appcast: :referer` isn't functional, it was merely a suggestion on how this could be officially implemented.

core-code:
so, should we use my CGI script until another solution is actually implemented or will this be implemented soon?

there are a few other casks where the only way to get version information is from the redirect

reitermarkus:
I don't think a dependency on a web server for this is worth it so I'd like to close this in favour of an issue for further discussion or a PR implementing `appcast: :referer` or something similar to express this behaviour.

cc @Homebrew/cask

core-code:
i'm not sure how the dependency could hurt (except by placing load on my server) and it would be easy to replace the dependency with a different system when and if it is implemented, but ok, lets close this issue ;)

PR #54788, opened 2018-11-12
Fix RuboCop config.

reitermarkus:
Depends on https://github.com/Homebrew/brew/pull/5312.


Casks/.rubocop.yml
```diff
@@ -0,0 +1 @@
+inherit_from: ../../../../.rubocop_cask.yml
```
MikeMcQuaid: This definitely warrants a comment.

PR #54814, opened 2018-11-12
Update bbedit to 12.5

core-code:

### DIFF
--- a/Casks/bbedit.rb
+++ b/Casks/bbedit.rb
@@ -1,14 +1,18 @@
 cask 'bbedit' do
-  version '12.5'
-  sha256 'ab40f9f3177a23851b9607cfd6ac427dfec4982330a08a1b22cd7ed438548114'
-
+  if MacOS.version <= :el_capitan
+    version '12.1.6'
+    sha256 '23b9fc6ef5c03cbcab041566503c556d5baf56b2ec18f551e6f0e9e6b48dc690'
+  else
+    version '12.5'
+    sha256 'ab40f9f3177a23851b9607cfd6ac427dfec4982330a08a1b22cd7ed438548114'
+  end
   # s3.amazonaws.com/BBSW-download was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


Casks/bbedit.rb
```diff
@@ -1,14 +1,18 @@
...
+    version '12.5'
+    sha256 'ab40f9f3177a23851b9607cfd6ac427dfec4982330a08a1b22cd7ed438548114'
+  end
   # s3.amazonaws.com/BBSW-download was verified as official when first introduced to the cask
   url "https://s3.amazonaws.com/BBSW-download/BBEdit_#{version}.dmg"
   appcast 'https://versioncheck.barebones.com/BBEdit.xml'
   name 'BBEdit'
   homepage 'https://www.barebones.com/products/bbedit/'
 
-  depends_on macos: '>= :sierra'
```
chrisfinazzo: 12.5 still requires Sierra - can this block be used in `if/else` to clarify the minimum OS required? 🤔

core-code: i thought there can only be a single 'depends_on' per cask? the absolute minimum the cask supports is el_capitan so that should currently be right

chrisfinazzo: Honestly not sure.

I don’t see that in the doc for the `depends_on` stanza. Can someone on 10.11 test this to confirm it works?

core-code: why would it not work if there is an '  depends_on macos: '>= :el_capitan' ' in there?

core-code: i've just looked through all 'depends_on macos' in all of HBC, and there doesn't seem to be a  single cask with multiple depends_on macos

chrisfinazzo: Was concerned you might end up with the wrong version b/c reqs & checksum are different, but if this works, I’m 👍🏻.

core-code: thanks!

PR #54829, opened 2018-11-12
Add google-chrome login item uninstall

dnicolson:

### DIFF
--- a/Casks/google-chrome.rb
+++ b/Casks/google-chrome.rb
@@ -16,10 +16,11 @@ cask 'google-chrome' do
 
   app 'Google Chrome.app'
 
-  uninstall launchctl: [
-                         'com.google.keystone.agent',
-                         'com.google.keystone.daemon',
-                       ]
+  uninstall launchctl:  [
+                          'com.google.keystone.agent',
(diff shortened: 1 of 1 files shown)


Casks/google-chrome.rb
```diff
@@ -16,10 +16,11 @@
...
 
-  uninstall launchctl: [
-                         'com.google.keystone.agent',
-                         'com.google.keystone.daemon',
-                       ]
+  uninstall launchctl:  [
+                          'com.google.keystone.agent',
+                          'com.google.keystone.daemon',
+                        ],
+            login_item: 'Google Chrome'
```
vitorgalvao: I don’t think this is correct. If Chrome is starting with your machine, you might have been the one to do it.

vitorgalvao: > but how is it different to other apps where `login_item` stanzas were added?

The difference is that if we’re going to add this to every cask where a login item *could* be added by the user, then every cask would need it.

> It was merged in [Google Chrome Canary](https://github.com/Homebrew/homebrew-cask-versions/pull/6526).

Shouldn’t have been, for the same reason. I’ll revert that.

PR #54840, opened 2018-11-13
flash 31.0.0.148

andyli:
including:
flash-npapi
flash-ppapi
flash-player-debugger-npapi
flash-player-debugger-ppapi
flash-player
flash-player-debugger

Note that Adobe hasn't updated its appcast.


andyli:
Going to redo in separated PRs...

PR #54891, opened 2018-11-13
Update bettertouchtool to 2.660

core-code:

### DIFF
--- a/Casks/bettertouchtool.rb
+++ b/Casks/bettertouchtool.rb
@@ -23,7 +23,7 @@ cask 'bettertouchtool' do
 
     # bettertouchtool.net/releases was verified as official when first introduced to the cask
     url "https://bettertouchtool.net/releases/btt#{version}.zip"
-    appcast 'https://bettertouchtool.net/releases/'
+    appcast 'https://updates.bettertouchtool.net/highsierra_new.xml?bttversion=967'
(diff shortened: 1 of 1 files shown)


Casks/bettertouchtool.rb
```diff
@@ -23,7 +23,7 @@
 
     # bettertouchtool.net/releases was verified as official when first introduced to the cask
     url "https://bettertouchtool.net/releases/btt#{version}.zip"
-    appcast 'https://bettertouchtool.net/releases/'
+    appcast 'https://updates.bettertouchtool.net/highsierra_new.xml?bttversion=967'
```
vitorgalvao: Won’t the fixed version at the end be a problem?

core-code:
as i understand it, it will be a problem once they decide to charge all users again for an upgrade again.


but when and if that happens we can change the number?

vitorgalvao:
> once they decide to charge all users for an upgrade again

It’s a subscription now. You pay for two years at a time so there’s no moment where everyone has to pay again. The old `appcast` might still be better.

core-code:
i didn't get the part of the subscription - just refused to pay again ;)

i am OK with dropping the appcast.

alternative could be to change the cask to carry both version numbers as many other casks do

PR #54953, opened 2018-11-14
Update textmate to 2.0-rc.19

schwedenmut:

### DIFF
--- a/Casks/textmate.rb
+++ b/Casks/textmate.rb
@@ -1,6 +1,6 @@
 cask 'textmate' do
-  version '2.0-rc.16'
-  sha256 '998c5b5aece79f2e27939149f8ed276e49aa1e5f451b359db977d55be2c1c873'
+  version '2.0-rc.19'
+  sha256 '31c16d92b4f267df0eb87bc3b04864f949f0007bb0661ab282ff89f397ffb41a'
 
   # github.com/textmate/textmate was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


core-code:
i think we should actually revert to rc10 since it is the one linked on their download page AND the "latest release" on github - all others are marked as "prerelease":

schwedenmut:
Hmmm, actually sounds good and conform to stable-rule. Thought there was an exception to the stable-rule.

core-code:
well, lets see what the powers-that-be think ;)

in any case, thanks for your PR!

reitermarkus:
> Thought there was an exception to the stable-rule.

Yeah, I'd say this is an exception because the homepage isn't really consistent with the GitHub releases. The `rc.10` is indeed not marked as a pre-release, but this seem's quite arbitrary. Also the homepage links to `beta.10`, despite there being a `beta.12.33`.

Maybe @sorbits can clear things up, but for now let's continue with using the latest `rc` release regardless.

core-code:
> Yeah, I'd say this is an exception because the homepage 
> isn't really consistent with the GitHub releases. 

it seems fully consistent. both claim rc10 is the latest 'stable'

> The rc.10 is indeed not marked as a pre-release, but this seem's quite arbitrary. 

the fact that this release is also linked from the download page as the latest one makes it seem 'not arbitrary' to me

but i may be interpreting the facts wrong....


in any case the developer seems completely nuts and doesn't seem to understand what an 'rc' actually is.

Issue #51492, opened 2018-08-31
Relax Travis rules regarding `sha256` changes?


vitorgalvao:
Now that quarantining has landed and the rules regarding `sha256`-only changes are going to change, we’re bound to get a bunch of false failures in Travis that’ll difficult review of what matters (cask functionality).

The simplest solution is to remove that check from Travis.

A better alternative would be to check the cask for non-`app`, non-`pkg` install stanzas, and only run the check on those.

PR #54966, opened 2018-11-14
Update macmediakeyforwarder URL and checksum.

telyn:
I'm not the developer of macmediakeyforwarder, and the developer doesn't publish any -SUMS files nor signs the downloads, so I cannot say why the checksum has changed :-1:

The URL changed recently from /hsme/ to /mmkf/ per the broader rename from HighSierraMediaKeyEnabler to MacMediaKeyForwarder (see also !54851, !54793, !54779), breaking the cask.

### DIFF
--- a/Casks/macmediakeyforwarder.rb
+++ b/Casks/macmediakeyforwarder.rb
@@ -1,8 +1,8 @@
 cask 'macmediakeyforwarder' do
   version '2.0'
-  sha256 '4941afa2ed60bf098e25a62d2ef3a42dfadcd38c18c39d4ae59cfd591ed63d1a'
+  sha256 '3eff3459f07ccecfb0bd1c456bf4d4a281af12550166863599d7230567a8ed31'
 
-  url "http://milgra.com/downloads/hsme/MacMediaKeyForwarder#{version}.zip"
+  url "http://milgra.com/downloads/mmkf/MacMediaKeyForwarder#{version}.zip"
   name 'Mac Media Key Forwarder'
   homepage 'http://milgra.com/mac-media-key-forwarder.html'
 


telyn:
Ah - I've read the sha256 policy - will see if I can get the developer to confirm the new checksum.

ran-dall:
I've been updating some casks. I went ahead an updated it with cask-repair (mad credit to @vitorgalvao ) that automagically pulls the sha256. I don't know if it's what you're looking for but I did update the path as well.

`brew cask install https://raw.githubusercontent.com/ran-dall/homebrew-cask/cask_repair_update-macmediakeyforwarder/Casks/macmediakeyforwarder.rb`

https://github.com/Homebrew/homebrew-cask/pull/54980

```
brew cask reinstall https://raw.githubusercontent.com/ran-dall/homebrew-cask/cask_repair_update-macmediakeyforwarder/Casks/macmediakeyforwarder.rb
==> Downloading https://raw.githubusercontent.com/ran-dall/homebrew-cask/cask_repair_update-macmediakeyforwarder/Casks/macmediakeyforwarder.
==> Satisfying dependencies
==> Downloading http://milgra.com/downloads/mmkf/MacMediaKeyForwarder2.0.zip
Already downloaded: /Users/draque/Library/Caches/Homebrew/downloads/6eca8baca40439116eef34bc0385c03f3854f7de6e34711ecf68e4ed1a835b57--MacMediaKeyForwarder2.0.zip
==> Verifying SHA-256 checksum for Cask 'macmediakeyforwarder'.
==> Installing Cask macmediakeyforwarder
==> Moving App 'MacMediaKeyForwarder.app' to '/Applications/MacMediaKeyForwarder.app'.
🍺  macmediakeyforwarder was successfully installed!
```

ran-dall:
@vitorgalvao It appears that's the right sha256

[Verification Link @ VirusTotal](https://www.virustotal.com/#/url/6eca8baca40439116eef34bc0385c03f3854f7de6e34711ecf68e4ed1a835b57/detection)

telyn:
@ran-dall That doesn't show anything about the code signing, which is what's needed per the [sha256 verification policy](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md). I've opened [an issue](https://github.com/milgra/macmediakeyforwarder/issues/61) on the macmediakeyforwarder github

ran-dall:
@telyn I was trying to show that the verification that you put in was correct. `3eff3459f07ccecfb0bd1c456bf4d4a281af12550166863599d7230567a8ed3` was the key you have, which is the same as the output of the `downloaded file` field in the screenshot. 

screen shot 2018-11-15 at 4 18 22 am

That said if I understand 
https://github.com/milgra/macmediakeyforwarder/issues/61 correctly; the developer updated the file `in place` of the old file. If that's the case, then I would recommend using`sha256 :no_check`, as the app comes in a zip and the VirusTotal link can be used as verification. It's on 
[the link.](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md)

Example: https://github.com/Homebrew/homebrew-cask/blob/master/Casks/spotify.rb

I think the reason Travis is failing; is because the file path was changed, but the sha256 sum wasn't. 

> When updating the sha256 stanza of an existing Cask, the version also has to have changed. Otherwise, the new checksum has to be confirmed. This is necessary to help rule out malicious tampering.

If I understand correctly, both the file and the sha256 sum needs to update; otherwise, I think that warning will pop up. Or you could use the `sha256 :no_check` value, as it fits the use case

> If there is no Signature Info section, VirusTotal verification is not enough.
> Maintainers will confirm the VirusTotal submission is legitimate by comparing its sha256 with the one on the updated cask.

vitorgalvao:
> Or you could use the `sha256 :no_check` value, as it fits the use case.

It does not. As explained in [the doc you linked to before](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#special-value-no_check):

> We use a checksum whenever possible.

`:no_check` is for specific cases and to be used sparingly. It should **never** be used merely to bypass Travis. That’d be worse (security-wise, which is the point of requiring verification) than outright merging the new `sha`.

ran-dall:
@vitorgalvao I got that (I see where I was mistaken in a way); but, correct me, if I'm wrong but Travis is pointing out that `only sha256 changed` but I guess it's saying the version has not. So I get that we wouldn't use `sha256 :no_check`, but wouldn't it still fit the use case of using `:latest`? I think this is probably a different build for whatever reason, which is why the checksum is different. I was thinking that using `:latest` (and `sha256 :no_check`) would make it future proof, cause the developer is updating the application in place. This would avoid having to verify the app every time someone updates it; as stated here:
> When updating the sha256 stanza of an existing Cask, the version also has to have changed. _Otherwise, the new checksum has to be confirmed. This is necessary to help rule out malicious tampering._

Or is verifying the cask manually every time no that big of a deal?  I see it as a bottleneck caused by the original developer, not HBC. I don't think it was covered in the documentation what is preferred. I think somewhere (I've re-read all the documents multiple times, I couldn't tell you exactly which one it was) it says something about making it easy to update the casks, but I understand the needs of having the right version number and sha256 sum.

---
Now, as for the verification, I did re-read the link; specifically where it's mentioned what specifically needs to happen to verify a cask. I understand asking the original developer through Twitter or a GitHub issue, which is in progress...but it also said...
> If the Cask is an .app that is codesigned (in a .dmg or .zip container) it can be uploaded and verified using VirusTotal by looking at the “Details” tab.

Which means...

screen shot 2018-11-15 at 11 34 31 am

Am I right, or am I missing something? Cause...

I redownloaded the .zip and [uploaded on VirusTotal](https://www.virustotal.com/#/file/3eff3459f07ccecfb0bd1c456bf4d4a281af12550166863599d7230567a8ed31/details) and verified the Signature. So would that be enough for the verification? Or we (HBC) will wait for the developer to respond?

---

(@vitorgalvao Thanks for helping me clear things up)

(@telyn SORRY!! I'm new at helping out and just trying to learn the right way of doing this. @vitorgalvao Is my much-appreciated defacto HBC mentor that's helping point out all my misunderstandings)

vitorgalvao:
> but wouldn't it still fit the use case of using `:latest`?

No, because there’s a version in the URL.

Unless we can show this cask is really problematic (as in, constantly updating in place) we should follow procedure. If the cask is shown to be problematic, [there’s also a procedure for that](https://github.com/Homebrew/homebrew-cask/blob/4184d50d2d7fa4afb62e6c7eb6761ae5ac417c88/Casks/3cxphone.rb#L3).

> I see it as a bottleneck caused by the original developer, not HBC.

And you are correct.

> I don't think it was covered in the documentation

See https://github.com/Homebrew/homebrew-cask/issues/48862, https://github.com/Homebrew/homebrew-cask/issues/51491, https://github.com/Homebrew/homebrew-cask/issues/51492

> I redownloaded the .zip and [uploaded on VirusTotal](https://www.virustotal.com/#/file/3eff3459f07ccecfb0bd1c456bf4d4a281af12550166863599d7230567a8ed31/details) and verified the Signature. So would that be enough for the verification?

It is indeed.

Thank you both!

milgra:
Hi, I'm the developer of Mac Media Key Enablers, and yes, I updated the zip once without incrementing the version, sorry about that, it won't happen any more.

ran-dall:
@vitorgalvao As always, I'm in your debt! Thank you!

@milgra Thanks for keeping us in mind!

Issue #54999, opened 2018-11-15
Alternative Cask for Skype for China


hanxue:
Currently the Skype cask cannot be installed in China because Microsoft Azure is blocked in China

```
$ brew cask reinstall skype
==> Satisfying dependencies
==> Downloading https://endpoint920510.azureedge.net/s4l/s4l/download/mac/Skype-

curl: (7) Failed to connect to endpoint920510.azureedge.net port 443: Operation timed out
Error: Download failed on Cask 'skype' with message: Download failed: https://endpoint920510.azureedge.net/s4l/s4l/download/mac/Skype-8.33.0.41.dmg

For reference: https://www.comparitech.com/privacy-security-tools/blockedinchina/

![image](https://user-images.githubusercontent.com/760008/48525929-c3bb5880-e8c0-11e8-8fc4-c0fc2f6a474c.png)
```

## Different Version for China

If you access http://skype.com/ within China, you will be redirected to http://skype.gmw.cn

The current version of Skype, `8.33.0.41` will not work in China. There is a Chinese version of Skype, version `7.59.37`, that is available at the url http://imgskype.gmw.cn/resource/uploadimg/mac/Skype_7.59.37.dmg

As far as I understand, Homebrew Cask does not have the `mirror` Cask stanza for alternative URL. Having a condition based on geographical location will also be tricky. 

Is it still possible to add the China download URL to the `skype.rb` cask, or is it better to have a separate Cask, i.e. https://github.com/Homebrew/homebrew-cask-versions/pull/6541/commits/592ba15773a128f8db9e56bb8c7eee3ef02172ff ?

ran-dall:
@hanxue Just a thought, I know `sudo systemsetup -gettimezone` will get the time zone. It only works on with `sudo` thought. I think maybe you could make an if...else type statement so that if the time zone matches China, then it would install the Skype from China. The only downside to this is that it requires `sudo`, as I mentioned, but [there have been some takes about how to work around that.](https://github.com/Homebrew/homebrew-cask/issues/19180#issuecomment-188522310) Just an idea, using time zones which is sort of geographic...

ran-dall:
@hanxue I guess this has already been addressed...
[**Regional, Localized: Any version that isn’t the US English one when that exists.**](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask)
I linked it to the place in the docs where it states that.

[...]

PR #55063, opened 2018-11-16
Localized Skype for China

hanxue:
There is a separate version of Skype for users in China, due to
network requirements (Government firewall). See #54999 for details

### DIFF
--- a/Casks/skype.rb
+++ b/Casks/skype.rb
@@ -1,22 +1,23 @@
 cask 'skype' do
   name 'Skype'
-  language 'en', default: true do
+
+  language 'XW', default: true do
     version '8.34.0.78'
     sha256 '814b60d85381d4a6332944cdf5f67f0e067b01843821989e685e53b49e6edf80'
     # endpoint920510.azureedge.net/s4l/s4l/download/mac was verified as official when first introduced to the cask
     url "https://endpoint920510.azureedge.net/s4l/s4l/download/mac/Skype-#{version}.dmg"
-    'en-US'
+    homepage 'https://www.skype.com/'
+    'XW'
   end
 
-  language 'zh' do
+  language 'CN' do
     version '7.59.37'
     sha256 '1f0ced6a3b50e9c43a684992a1d63d576eeb689e84ff6572a73105e9a92796cd'
-    # Sype in China has a different release version and homepage
+    # Skype in China has a different release version and homepage
     url "http://imgskype.gmw.cn/resource/uploadimg/mac/Skype_#{version}.dmg"
-    'zh-CN'
+    homepage 'http://skype.gmw.cn/'
+    'CN'
   end
-
-  homepage 'https://www.skype.com/'
 
   auto_updates true
 
--- a/Casks/skype.rb
+++ b/Casks/skype.rb
@@ -1,10 +1,21 @@
 cask 'skype' do
-  version '8.34.0.78'
-  sha256 '814b60d85381d4a6332944cdf5f67f0e067b01843821989e685e53b49e6edf80'
-
-  # endpoint920510.azureedge.net/s4l/s4l/download/mac was verified as official when first introduced to the cask
-  url "https://endpoint920510.azureedge.net/s4l/s4l/download/mac/Skype-#{version}.dmg"
   name 'Skype'
+  language 'en', default: true do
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Is this official? Is there a way we can confirm this is really *the* Skype, by Microsoft?

This is an unorthodox solution, but I don’t really have a problem with it. Pinging @Homebrew/cask in case anyone does.

amyspark:
VirusTotal analysis confirms this is a validly signed DMG: https://www.virustotal.com/#/file/1f0ced6a3b50e9c43a684992a1d63d576eeb689e84ff6572a73105e9a92796cd/details

claui:
There may be lots of Macs which are not within reach of the Great Firewall but still set to a `zn` locale. Examples may include [Hong Kong and Macau](https://en.wikipedia.org/wiki/Great_Firewall). I wouldn’t want unrelated Chinese-speaking people to have their Skype unexpectedly downgraded to the 7 branch.

We already have `skype7` on Homebrew/cask-versions, which is on the same version (7.59.37) as this PR suggests. Why not add the `zn` alternative there?

hanxue:
@claui Thanks for checking on the package security and your feedback. 

I am leaning towards creating a `skype-china` Cask, really, simply because Skype in China has a totally different website - visitors to https://skype.com will be redirected to http://skype.gmw.cn/. The latest stable version is 7.xx.x . Whilst most other applications are either totally unavailable in China due to GFW, or has a localized version, Skype seems to be marketed differently. It does not make sense to create a separate Cask for each localized application, but in the case of China, there is a significant number of users accessing the Chinese skype website and using version 7.xx.x

Someone who frequently uses VPN or officially blocked applications would not have any problem getting the international edition of Skype. But most other Chinese users will be unable to download from the 8.xx.x link.

Bear in mind that users in Hong Kong and Macau do not have filtered Internet access - at least not the GFW. I was not even aware of a `skype7` Cask until this thread's discussion, and I only ran into this problem having previously installed Skype out of China, and finding myself being unable to upgrade here in Beijing. 

I have just tried setting the locale to `zh`, and homebrew cask does not automatically pick up the Chinese URL. I still have to set `--language=zh` for Homebrew Cask to use the 7.xx.x Chinese version.

claui:
> Someone who frequently uses VPN or officially blocked applications would not have any problem getting the international edition of Skype.

That’s another reason why I’m still a bit reluctant to merge this PR. Imagine a user who lives in mainland China and uses Homebrew while connected to an out-of-country VPN, bypassing the GFW.

I believe that from that user’s point of view, it would be unexpected if we directed them to gmw\.cn automatically. In extreme cases, it may even put that user at risk because gmw\.cn 1) doesn’t support TLS so the request itself can leak information, and 2) it’s controlled by the national newspaper _Guangming Daily,_, which [appears to have close ties to the Communist Party of China](https://en.wikipedia.org/wiki/Guangming_Daily), the sole governing party.

Your mileage may vary but if I were in mainland China, I would not be happy to find myself on the access logs of Guangming Daily while connected to a foreign VPN.

> I have just tried setting the locale to `zh`, and homebrew cask does not automatically pick up the Chinese URL.

That’s because I used the word _locale_ ambiguously. Sorry for the confusion.

The Unix locale (as set via environment variables such as `LANG` or `LC_ALL`) is ignored by Homebrew.

With no `--language` argument given, Homebrew will use (by order of priority):

1. the locale strings from the `HOMEBREW_LANGUAGES` environment variable, and

2. macOS’s locale as set via the language and regional settings on the _Localization_ system preference pane. The preference is encoded as a language-region string like `en-CN` or `zh-HK`.

You can check your macOS locale strings by running:

```
defaults read -g AppleLanguages
```

Depending on the value of the `language` stanza, [Homebrew Cask will match language codes, region codes, or both.](https://github.com/Homebrew/homebrew-cask/blob/11b48ee20be60af6d6e7f743796965da52741ee2/doc/cask_language_reference/stanzas/language.md) For Homebrew to match `zh` (the language), it’s not enough to have your region set to mainland China. You can verify that by exporting e. g. `HOMEBREW_LANGUAGES=en-CN,zh-CN`. In the case of this PR, Homebrew would then pick up the English version.

So what we actually want here is to match the mainland China region (`CN`), not `zh`. Care to check if this variant works for you?

PR #55018, opened 2018-11-15
Update freefilesync to 10.6

Wayneoween:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #54820, opened 2018-11-12
Update liclipse to 5.1.3

core-code:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #55071, opened 2018-11-16
Remove casks that use mediafire as a source

vitorgalvao:
Closes https://github.com/Homebrew/homebrew-cask/pull/55018 https://github.com/Homebrew/homebrew-cask/pull/54820

See those two issues. Apparently mediafire is making it harder to download and get a correct `sha256`. They’re so few casks, it seems to me the best bet is to give it the Audacity/fosshub treatment.


amyspark:
@vitorgalvao do you have usage stats for these?

vitorgalvao:
> @vitorgalvao do you have usage stats for these?

I don’t. Those would indeed be useful to check.

core-code:
i have usage stats for all apps, therefore for all casks that install at least one app into the /Applications folder. they are taken from the few thousand users of our app, so obviously they are not completely indicative of the HBC project at large.



org.freefilesync.FreeFileSync	Users: 0209	
com.mediafire.desktop  	Users: 0009
LiClipse	Users: 0004	

lets put things into perspective with a few other numbers

org.videolan.vlc	Users: 5696
com.Google.GoogleEarthPro	Users: 2219
com.titanium.OnyX	Users: 1263
com.c-command.DropDMG	Users: 0552


btw, there are many many casks which have zero users (well, the intersection between your users and my users is zero).

amyspark:
@Homebrew/cask: If none has issues, I'll proceed with the Audacity treatment.

reitermarkus:
For what it's worth, I'm using FreeFileSync myself.

core-code:
i think the FreeFileSync license would allow just placing it on a properly hosted mirror.

we've mirrored Audacity here: https://github.com/suschizu/audacity-mirror/

vitorgalvao:
> For what it's worth, I'm using FreeFileSync myself.

Then I’m fine with any alternative you wish for that particular cask. I’m partial to letting casks used by HBC maintainers bend a few rules, since they’re more likely to have the extra oversight to compensate.

trinitronx:
Tried re-installing `freefilesync` just now and ran into this.  According to `brew cask audit`, the SHA-256 has indeed changed as observed by others:

```
$ brew cask audit --download freefilesync
==> Downloading https://www.freefilesync.org/download/FreeFileSync_10.6_macOS.zip
Already downloaded: ~/Library/Caches/Homebrew/downloads/120dd1affdd28ba060e44321733094f2e746f0031f34b9529884ad50c0fd28f4--download.php
==> Verifying SHA-256 checksum for Cask 'freefilesync'.
==> Note: Running `brew update` may fix SHA-256 checksum errors.
audit for freefilesync: failed
 - download not possible: Checksum for Cask 'freefilesync' does not match.

Expected: 5b94686935da5160228bb1965f2dafd22894e23a1c8a0bef1709f222645c26cd
Actual:   00adc3e3fb32a9d1e8d7e8204524460c03624b7238b2b9877d3474670a01bba1
File:     ~/Library/Caches/Homebrew/downloads/120dd1affdd28ba060e44321733094f2e746f0031f34b9529884ad50c0fd28f4--download.php

Error: audit failed for casks: freefilesync
```

If it isn't changing too often maybe new sha will work?
Otherwise, I vote for the alternate solution of using ` sha256 :no_check`

Unfortunate that not all file hosts are curl & checksum friendly, but still seems heavy handed to give all of these Audacity treatment if OSS.  Convenience factor for using Homebrew Cask to install these is really not worth throwing away.  I still really wish Audacity would fix their distribution to use a better file host.

core-code:
if you depend on audacity, try this: 
`
brew cask install https://www.corecode.io/macupdater/casks/audacity.rb
`

vitorgalvao:
> If it isn't changing too often maybe new sha will work?

It’s changing every time, which is what makes it a problem.

> I still really wish Audacity would fix their distribution to use a better file host.

There’s nothing to fix. Their system isn’t broken; it’s working exactly as intended. The way they have it setup is how they make money.

claui:
While I’m totally up for FOSSHubbing any and all casks hosted by shitty services, I believe Mediafire is not at fault here (for once).

My theory is that the `freefilesync` SHA change was legitimate, and done by the upstream author last week.

I also believe that the [Travis glitch](https://github.com/Homebrew/homebrew-cask/pull/54820#issuecomment-438016887) we’ve seen with `liclipse` is rather an error in the Cask. The correct (and stable) URL for liclipse 5.1.3 is `https://www.mediafire.com/file/32gtr23s399ec2x/liclipse_#{version}_macosx.cocoa.x86_64.dmg`, not `https://download2267.mediafire.com/23daeigrqddg/32gtr23s399ec2x/liclipse_#{version}_macosx.cocoa.x86_64.dmg`.

The stable `www.mediafire.com` URL will redirect to a `download****.mediafire.com` URL. But that’s a thing we have with many hosters, and will certainly not lead to SHA mismatches unless the upstream author uploads a new package.

I’m now going to check my theory by fixing #54820.

reitermarkus:
> My theory is that the `freefilesync` SHA change was legitimate, and done by the upstream author last week.

Seems like it, since it already happened once: https://github.com/dtgm/chocolatey-packages/issues/234

claui:
> Seems like it, since it already happened once: dtgm/chocolatey-packages#234

Good find!

I suggest that we keep `freefilesync` for now under the condition that such informal updates don’t occur more than, say, once per year.

@Homebrew/cask How do the fellow maintainers feel about that?

vitorgalvao:
I’m fine with keeping `freefilesync`. Can similar fixes be applied to the other two?

vitorgalvao:
`liclipse` is fixed and looks like `freefilesync` is fixable as well. Any takers?

core-code:
> `liclipse` is fixed and looks like `freefilesync` is fixable as well. Any takers?



> `liclipse` is fixed and looks like `freefilesync` is fixable as well. Any takers?

i don't think there are any changes to the freefileync URL needed. it doesn't contain the 'wrong' mediafire URL that has the SHA problem.

however, the SHA is wrong, because, i think, the developer has re-rolled the release on 14 november after being initially posted on 12 november. sure i can send a cask-repair, but will it actually be accepted because of our SHA-only-change-policy?

PR #55115, opened 2018-11-17
Update gray to 0.9.5

ran-dall:

### DIFF
--- a/Casks/gray.rb
+++ b/Casks/gray.rb
@@ -1,6 +1,10 @@
 cask 'gray' do
-  version '0.9.5'
-  sha256 'f7f7dc488fd3ce39658c382672dc85014d253d0d673cf0b7f9d7c90f1330dae4'
+  if MacOS.version >= :mojave
+    version '0.9.5'
+    sha256 'f7f7dc488fd3ce39658c382672dc85014d253d0d673cf0b7f9d7c90f1330dae4'
+  else
+    version '0.9.3'
+    sha256 '0d9566ec7741603ddfcee4a241370aab97605b874ee168e79d0ea89b587b8154'
 
   url "https://github.com/zenangst/Gray/releases/download/#{version}/Gray.zip"
   appcast 'https://github.com/zenangst/Gray/releases.atom'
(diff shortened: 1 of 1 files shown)


Casks/gray.rb
```diff
@@ -1,13 +1,16 @@
 cask 'gray' do
-  version '0.9.3'
-  sha256 '0d9566ec7741603ddfcee4a241370aab97605b874ee168e79d0ea89b587b8154'
+  if MacOS.version >= :mojave
```
reitermarkus: Switch this around, so the `else` clause contains the newer version.

reitermarkus (changes requested):

Casks/gray.rb
ran-dall: @reitermarkus Thank you for helping me out! Just for clarity, the new version is only compatible Mojave, otherwise, for High Sierra and everything else, you'd want the other version. I just wasn't sure I could make a ```if MacOS.version <= :high_sierra...else...```statement, but that's what wer're looking for right?

suschizu:
@ran-dall - I just took a look in our database and since v 0.2.1 this app solely depends on Mojave - I think the if/else statement doesn't make sense here

ran-dall:
@suschizu I would then say, let's make 0.2.1 available to High Sierra users. There's a significant amount of people that haven't upgraded to Mojave yet because of different incompatibilities with their hardware and/or software. So I don't see this a problem because HBC officially supports the last two versions of a MacOS. 

Also, I was able to open Gray 0.2.3 on High Sierra yesterday. I don't use it, so I don't know much else. If 0.2.1 is the right version for High Sierra, I would say go with that. It doesn't bother me either way you go, just my two cents.

suschizu:
hey @ran-dall - this app was made for Mojave only. I think it allows you to transform the dark-mode- grayscale, so it cannot be used with high-sierra. The headline says: "Tailor your macOS Mojave experience"

ran-dall:
@suschizu I see what you're saying. Sorry, I don't use Gray. So...I'll just update it.

PR #55117, opened 2018-11-17
Add gpg-suite-pinentry 2018.5

mafredri:
Add gpg-suite-pinentry 2018.5. This installs only the pinentry package from gpg-suite.

This cask installs only pinentry from gpg-suite which can be considered a more up-to-date version (1.1.0) of pinentry-mac available in homebrew.

The reason for this cask is that [GPGTools/pinentry-mac](https://github.com/GPGTools/pinentry-mac) (a.k.a. pinentry-mac v0.9.4 from homebrew) is no longer updated and a more up-to-date pinentry is located at [GPGTools/pinentry](https://github.com/GPGTools/pinentry) (v1.1.0). However, there are no releases for the up-to-date pinentry apart from it being bundled in gpg-suite. As such it seemed appropriate to create this cask with this naming. I'm open to alternatives and will update the PR accordingly.

Another reason is that, in its default configuration, gpg-suite is intrusive as it modifies the users `PATH` and sets a high priority for it's own `gpg`-binaries and also installs a Mail plugin, etc. It's more preferable to only install pinentry and use gpg from homebrew.

### DIFF
--- a/Casks/gpg-suite-pinentry.rb
+++ b/Casks/gpg-suite-pinentry.rb
@@ -10,7 +10,7 @@ cask 'gpg-suite-pinentry' do
   auto_updates true
   conflicts_with cask: [
                          'gpg-suite',
-                         'gpg-suite-nightly'
+                         'gpg-suite-nightly',
                        ]
   depends_on macos: '>= :sierra'
 
--- a/Casks/gpg-suite-pinentry.rb
+++ b/Casks/gpg-suite-pinentry.rb
@@ -95,20 +95,13 @@ cask 'gpg-suite-pinentry' do
                  },
                ]
 
-  uninstall pkgutil: 'org.gpgtools.pinentry*',
+  uninstall pkgutil: 'org.gpgtools.pinentry.*',
             quit:    'org.gpgtools.pinentry-mac',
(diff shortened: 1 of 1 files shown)


amyspark:
I've had to hack a pinentry bottle because the existing offering didn't work (for a while) in VS Code. This would be more than welcome.

amyspark (approved):

mafredri:
@vitorgalvao @amyspark thanks for reviewing!

@vitorgalvao and thanks for the formatting fixes :). I have one question though, and it's part of the reason I wrote some `caveats` in the original PR:

Is there some other way than caveats to inform users about quirks in the installed software? And what is the general policy on including caveats?

In this instance I added the note about `Save in Keychain` because by not installing GPG Preferences, there is no way to configure that option unless you know about it. Personally I'm also not a fan of the default (Save in Keychain is on by default) and it's also not configurable via the Pinentry software, you must do it by modifying the plist (e.g. via `defaults`). That said, I don't care, for me, if there's a caveat or not, I just wanted to raise awareness.

The note about setting default pinentry was there so that the user doesn't have to hunt down the path of the Pinentry program, it's installed into a slightly hidden location (`/usr/local/MacGPG2/libexec`).

Issue #142, opened 2013-01-28
Allow for more flexibility with multiple versions of the same app


passcod:
In #114, I hinted at this, and #41 is the original issue. We need some way to support apps that have multiple concurrent builds/versions/channels. Here are the various situations in the Casks library at the moment:

- Firefox has four "channels": [Stable], [Beta], [Aurora], and [Nightly]. Currently we have two casks: [firefox] and [firefox-aurora].

- We have both [Chrome], [Canary] \(which is the equivalent of Aurora for Firefox), and [Chromium] \(which is Chrome without the Google in it).

These are just multiple stable/beta/alpha/nightlies, though. IntelliJ presents a different (but no less worthy) use-case: we currently has a cask each for the [Community] and [Ultimate] editions. It would be interesting to combine these!

[Stable]: www.mozilla.org/firefox
[Beta]: https://www.mozilla.org/firefox/beta/
[Aurora]: https://www.mozilla.org/firefox/aurora/
[Nightly]: http://nightly.mozilla.org/
[firefox]: https://github.com/phinze/homebrew-cask/blob/master/Casks/firefox.rb
[firefox-aurora]: https://github.com/phinze/homebrew-cask/blob/master/Casks/firefox-aurora.rb

[Chrome]: https://github.com/phinze/homebrew-cask/blob/master/Casks/google-chrome.rb
[Canary]: https://github.com/phinze/homebrew-cask/blob/master/Casks/google-chrome-canary.rb
[Chromium]: https://github.com/phinze/homebrew-cask/blob/master/Casks/chromium.rb

[Community]: https://github.com/phinze/homebrew-cask/blob/master/Casks/intellij-community.rb
[Ultimate]: https://github.com/phinze/homebrew-cask/blob/master/Casks/intellij-ultimate.rb

- - - - - - - -

My current proposal would have the formulae look like this:

```ruby
class SomeCask < Cask
  url 'http://example.com/thing-1.0.0.dmg'
  homepage 'http://example.com'
  version '1.0.0'
  sha1 '...'

  # The devel spec has a version but no checksum
  # (useful when the file changes frequently and w/o warning).
  devel do
    url 'http://example.com/thing-1.1.0-beta.dmg'
    version '1.1.0-beta'
  end

  # The edge spec has no version nor checksum
  # (e.g. nightly builds). Will be installed under `EDGE`
  # as version-path.
  edge do
    url 'http://example.com/nightly.dmg'
  end

[...]

PR #55265, opened 2018-11-21
Docker.app 2.0.0 requires 10.12

Amar1729:

Note:
The [docker store requirements](https://store.docker.com/editions/community/docker-ce-desktop-mac) officially say the minimum required OS is macOS 10.10 (yosemite) but attempting to run the downloaded Docker.app on El Capitan hard-exits with the [message that at least 10.12 is required](https://i.imgur.com/3Z4xx5g.png).

Edit: Based on conversation below, this PR updates minimum system requirements for current edge release of Docker (2.0.0) ~and renames old `docker.rb` for version 18.06.1 to `docker@18` for compatibility with Yosemite and El Capitan.~ (`docker@18.rb` moved to #55280)


ran-dall:
@vitorgalvao @reitermarkus Docker 2.0.0. is a edge release (https://docs.docker.com/docker-for-mac/edge-release-notes/) that already exists in HBC Versions (https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/docker-edge.rb), athough Docker has version 27117 for the Edge build and we have 28700, but I think it makes little difference.

The stable release is listed here: https://docs.docker.com/docker-for-mac/release-notes/
So that would be: 18.06.1-ce-mac73 2018-08-29 (which was updated and merged by @brianmorton here: https://github.com/Homebrew/homebrew-cask/pull/51409).

ran-dall:
@LanikSJ I only see it mentioned in the 'appcast', but I don't see it mentioned anywhere else, on the Enterprise Edition. It's true though that (https://download.docker.com/mac/stable/Docker.dmg) downloads 2.0.0mac78.

Amar1729:
Yeah @ran-dall I'm confused by it also - the site and documentation indicate that 10.11 is supported and the app itself actually runs, but all it does is open a "Fatal Error" dialog saying that at least 10.12 is required, with only a button to exit.

On El Captian 10.11.06, [Docker 18.06.0-ce-mac70 (26399)](https://download.docker.com/mac/stable/26399/Docker.dmg
) opens and runs properly (although on startup still informs about not being able to update anymore since El Capitan is deprecated).

Maybe 18.06.0 could be installed for macOS Yosemite - El Capitan, and 2.0.0 for Sierra or newer?

ran-dall:
@Amar1729 Awesome! Thanks for the feedback. 👍 

I think your suggestion makes sense. Technically HBC only supports the last two versions of MacOS. However, I feel Docker is one of those tools that everyone is using now. So I'd vote that we try to keeping it as compatible as possible.

Amar1729:
Keeping in mind the discussion in #142 about multiple versions of an app, I think a good way to do this might be to keep Docker 2.0.0 edge in `docker.rb` and rename the old version to `docker@18.rb` . I'm not sure what problems, if any, this might introduce to users still running Yosemite/El Capitan that already have docker installed.

ran-dall:
@vitorgalvao Maybe you or @reitermarkus have an opinion on this?

@LanikSJ I don't have a problem with @Amar1729 adding the statement. I like that too. You're right that it would be much more comfortable. I really hadn't considered that 🙃. I was thinking 'officially' of what we should do... 😇

LanikSJ:
> @vitorgalvao Maybe you or @reitermarkus have an opinion on this?
Exactly that, once the HBC folks chime in we will know which they would prefer.

reitermarkus:
I think it's fine to add this as an `if` statement since it there isn't really an incompatible change except for the macOS version requirement.

vitorgalvao:
I’m on a phone, so forgive me for the lack of links and direct quotes, but I’ll address some points in [@ran-dall’s comment](https://github.com/Homebrew/homebrew-cask/pull/55265#issuecomment-440901731).

Please don’t *encourage* adding new casks to the `versions` repo. Yes, it exists and should be used, but casks there also tend to be less maintained, so we should add with care.

We (Cask) don’t use the `@` naming scheme.

Old versions of open-source software need a good reason to be added to the repo (see the main README in that repo, as it covers acceptable casks).

`caveats` cannot have the suggested message. Messages should be sparse and installation (not even usage) related, to avoid them being ignored due to too much inconsequential messages (see the `caveats` document).

Regarding the `if`, *always* use `<=`, never `>=`. `else` is always the most recent.

As for this case, I’m fine with adding the conditional for a while.

reitermarkus:
Wow, the timing @vitorgalvao. 😜

ran-dall:
I agree with https://github.com/Homebrew/homebrew-cask/pull/55265#issuecomment-441096619

I talked to some maintainers of Docker early this morning on Discord. The edge Documentation was having some weird issue, but now there should be a 'Deprecation' passage in the Edge documentation.  That should reflect the depreciation of 10.11. That said, from what I gathered from the conversation, "2.0 is still not fully released yet but is considered Stable". I was told that "...documentation will be updated when there's a need to update it." So I don't know what that's supposed to mean, but I'm assuming it implies that updating Documentation is probably in the priority list and **we should keep 2.0.**

After the discussion with Docker, other than the 10.11 depreciation, AUFS is the only breaking change. So maybe a 'version' of this specific cask isn't necessary, as there's only supposed to be a minimal number of users that will be affected.

@vitorgalvao I've seen a lot of different caveats keep the casks up-to-date, so I appreciate the clarification. I've been trying to just not touch them, but when I do, I'll know what to do. Also, thanks for point out the error in the `if...else`; it was a conscious error (if you know what I mean). I'll try to be more careful for it and recommending 'HBC/Versions' in the future. Although, I do think that certain large 'Developer' tools do belong in 'HBC/Versions' if they introduce breaking changes that can interfere with a Developer's workflow, process, etc... Please correct my way of thinking, if there's a better HBC viewpoint to adopt. As always, I'm very grateful for your intervention. 🙏

@Amar1729 Here's the new code:
```ruby
cask 'docker' do
if MacOS.version <= :el_capitan
  version '18.06.1-ce-mac73,26764'
  sha256 '3429eac38cf0d198039ad6e1adce0016f642cdb914a34c67ce40f069cdb047a5'
else
  version '2.0.0.0-ce-mac78,28905'
  sha256 'a452026a1780e17ea2717acf68c56a09bb19a2e46dbc6988e011f889cd28e5b6'
end

 url "https://download.docker.com/mac/stable/#{version.after_comma}/Docker.dmg"
  appcast 'https://download.docker.com/mac/stable/appcast.xml'
  name 'Docker Community Edition'
  name 'Docker CE'
  homepage 'https://www.docker.com/community-edition'

  auto_updates true

  app 'Docker.app'

  uninstall delete:    [
                         '/Library/PrivilegedHelperTools/com.docker.vmnetd',
                         '/private/var/tmp/com.docker.vmnetd.socket',
                         '/usr/local/bin/docker',
                         '/usr/local/bin/docker-compose',
                         '/usr/local/bin/docker-credential-osxkeychain',
                         '/usr/local/bin/docker-machine',
                         '/usr/local/bin/hyperkit',
                         '/usr/local/bin/notary',
                         '/usr/local/bin/vpnkit',
                       ],
            launchctl: [
                         'com.docker.helper',
                         'com.docker.vmnetd',
                       ],
            quit:      'com.docker.docker'

  zap trash: [
               '/usr/local/bin/docker-compose.backup',
               '/usr/local/bin/docker-machine.backup',
               '/usr/local/bin/docker.backup',
               '~/Library/Application Scripts/com.docker.helper',
               '~/Library/Caches/KSCrashReports/Docker',
               '~/Library/Caches/com.docker.docker',
               '~/Library/Caches/com.plausiblelabs.crashreporter.data/com.docker.docker',
               '~/Library/Containers/com.docker.docker',
               '~/Library/Containers/com.docker.helper',
               '~/Library/Group Containers/group.com.docker',
               '~/Library/Preferences/com.docker.docker.plist',
             ],
      rmdir: [
               '~/Library/Caches/KSCrashReports',
               '~/Library/Caches/com.plausiblelabs.crashreporter.data',
             ]
end
```

PR #55302, opened 2018-11-23
Update dashlane to 6.1844.0.14691

ran-dall:


ran-dall:
@vitorgalvao This seems to be an installer app. I haven't encountered something like this before. I would really appreciate some guidance because I'm pretty sure we don't want to put that in the /Applications folder.

PR #55329, opened 2018-11-23
Update sdformatter.rb

suschizu:
I think something was changed in the download-file - instead the cask-repair wouldn't have passed the travis-check (#55276)

screenshot 2018-11-23 at 10 07 20


vitorgalvao:
@suschizu Mind uploading to VirusTotal? That check is as much for as as it is for other users auditing PRs.

vitorgalvao:
Oh, it’s a `pkg`. Hadn’t noticed. Those are fine, though. By design, we don’t install non-signed `pkg`s.

PR #52314, opened 2018-09-22
Update media-center to 24.00.53

nrlquaker:
Changed `latest` back to `stable` in the url.
Because I assume not every latest build is stable.
On [download page](https://www.jriver.com/download.html) macOS links are always stable.
And not every build from forum is posted to download page.
Also when using auto updater in app for stable channel, it updates only to stable builds, skipping forum builds.

After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/Homebrew/homebrew-cask/pulls
[already refused]: https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues
[the correct repo]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask
[version-checksum]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256

PR #55333, opened 2018-11-23
Change channel to stable from latest for media-center

nrlquaker:
Changed latest back to stable in the url. [Again](https://github.com/Homebrew/homebrew-cask/pull/52314).
Because I assume not every latest build is stable.
On download page macOS links are always stable.
And not every build from forum is posted to download page.
Also when using auto updater in app for stable channel, it updates only to stable builds, skipping forum builds.
Homebrew Cask docs [says](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) that 

> Stable versions live in the main repository

Pinging @core-code who did this changes. Please explain why its needed.
Thanks.

### DIFF
--- a/Casks/media-center.rb
+++ b/Casks/media-center.rb
@@ -2,7 +2,7 @@ cask 'media-center' do
   version '24.00.56'
   sha256 'e9fe2e1755f533c898b5d201081feb5b786c9fd0c9cb3c466f57631c0f6c82ef'
 
-  url "https://files.jriver.com/mediacenter/channels/v#{version.major}/latest/MediaCenter#{version.no_dots}.dmg"
+  url "https://files.jriver.com/mediacenter/channels/v#{version.major}/stable/MediaCenter#{version.no_dots}.dmg"
   name 'JRiver Media Center'
   homepage 'https://www.jriver.com/'
 


core-code:
i've changed it to the 'latest' because the 'latest version' was only available on that 'channel' when i submitted the pull request.

'latest' did not in any way sound 'unstable' to me but rather sounded like the 'latest official version', and still does. if we want to track only 'media-center' releases that are on the 'stable' channel, sorry, i'll keep that in mind.

nrlquaker:
Sometimes there is no latest stable version in download section of the site but its available using stable link in the in app auto updater.

core-code:
does it really matter which URL we download from?

nrlquaker:
Better to ping core members.

nrlquaker:
1. Latest stable version is in `stable`.
2. Some new build are only available on 'latest', but they are not stable. Upstream only publish them at forum. Not on the download page. And in app update uses stable builds for updates.
3. Both are available. Latest is ahead of stable (its naturally).

core-code:
personally i find this confusing. latest, implies official and 'stable'. they should probably call the channel 'unstable' not 'latest' then. 

in any case, it seems to make sense to stick the 'stable', +1

vitorgalvao:
> latest, implies official and 'stable'.

They must mean `latest` as `cutting-edge`.

PR #55392, opened 2018-11-25
Update horos to 3.3.4

coreygo:


coreygo:
Looking at https://horosproject.org/download/ it appears v4 is going to be available soon and `<= :el_capitan    version '2.0.2'` probably needs updating to v2.2.0 or v2.4.1 (not sure which, I'm on Mojave).

`~/Library/Application Support/Horos` is also left on the system after uninstalling. So this cask may need an uninstall section which zaps that.

PR #55436, opened 2018-11-26
Update bonitastudiocommunity.rb

core-code:
update to 7.7.5, update download URL


core-code:
sorry, the file is too big for me to actually download ATM so the earliest i can look into this is end-of-week

PR #55499, opened 2018-11-29
Update ogdesign-eagle to 1.8.2-build5

whalesingswee:

### DIFF
--- a/Casks/ogdesign-eagle.rb
+++ b/Casks/ogdesign-eagle.rb
@@ -1,6 +1,6 @@
 cask 'ogdesign-eagle' do
-  version '1.8.1'
-  sha256 '2ea4de81ef4f6dc5951ab682352595516e43948ea5716d31b646d105397c46f6'
+  version '1.8.2-build5'
+  sha256 '3f93ede904a1b88167f32738bacad65d67eaa852f45c47045d92ac9d362b5814'
 
   # s3.amazonaws.com/eagleapp/ was verified as official when first introduced to the cask
   url "https://s3.amazonaws.com/eagleapp/releases/Eagle-#{version}.dmg"


whalesingswee:
> is there an official guideline which one to use? if not could we make one and stick to it?

Hi there, I'm afraid the developer still cannot decide and finalize for the deployment. How do I create one if I would like to contribute for long term maintenance?

vitorgalvao:
@core-code It’s unclear who you’re asking. If you’re asking “does HBC have a guideline?”, the answer is “whatever’s most convenient for interpolation”, so it varies by developer/cask.

core-code:
@whalesingswee thanks for mentioning this to the developer. indeed the reason the format of the 'caskversion' changes all the time is because the upstream URL format changes all the time because the developer hates consistency ;)

i propose to always stick to the format
#{versionumber},#{buildnumber}, and handle changes in the upstream URL format by changing the URL stanza. 

reasons:
1.) separating with a comma is the format most commonly used in HBC
2.) this makes sense to me because this way upstream version changes lead to a changed version stanza and url changes lead to a changed url stanza - this seems consistent

by total coincidence this guideline would also be useful to my project using HBC and i guess others will see things differently ^^

vitorgalvao:
> and handle changes in the upstream URL format by changing the URL stanza.

We specifically don’t want that. A good bump PR is one that changes as little as possible (i.e. just the `version` and `url`. It’s easier to parse and review. Changes to the `url` require more attention to parse.

Also, we’d rather avoid splitting the version (`comma` and `colon`).

core-code:
> A good bump PR is one that changes as little as possible (i.e. just the version and url. It’s easier to parse and review.
> Changes to the url require more attention to parse.

i am not sure i understand that, because the second sentence seems to contradict the first sentence - as both talk about changing the url stanza.

> Also, we’d rather avoid splitting the version (comma and colon).

ok, sorry to reach out a bit. we were just talking about version separators, but its all part of something more fundamental.

i understand that its the position of the HBC project that the 'version' stanza isn't actually a 'version number' in the traditional (or real) sense of the word and is really a 'download token' to be put into the 'url' or “whatever’s most convenient for interpolation” as you put it. 

my app requires the something different, i want the real version number of the app (i.e. the CFBundleVersion or CFShortVersionString) of the app in there, or at least something than can be transformed in any possible way into either or them or at least into something thats close enough. preferably both of them, with a clear separator like the comma.

for about 90% of apps the version stanza is actually the version number of the app and both things are one and the same and the difference is a moot point because it doesn't exist.
for the remaining 10% percent of apps i understand that my needs are of no concern at all to HBC and i'll happily abide by all rules - and wouldn't even want to see them changed to anything that doesn't make perfect sense to the project itself just to accomodate external needs. i am happy enough as-is that i was able to slip in real version numbers to some casks ;)

that said, i think there are a number of arguments to be said for 
1.) carrying real version numbers of the app in the version stanza - even if they are not needed for url interpolation (this is not connected to the ogdesign-eagle cask)
2.) not needlessly changing the version number format just so the url stanza doesn't need to be adjusted (which is connected to the ogdesign-eagle cask and the discussion at hand)

off the top of my (saturday night) head:

1.) if the real version number is (also) in the cask this probably makes more sense to your users installing the app or sending repairs.
2.) if the real version number is in the cask, we don't have problems with SHA-only changes with casks that don't change the download URL for every version. i realize this is mitigated a bit by relaxing the rules for SHA-only changes now however but i guess this is the most important reason currently for many casks to have a version (part) thats not used in the url for interpolation.
3.) if there are real version numbers in the URL stanza, i can 'support' the cask in my system and will be able to help with maintaining it e.g. with the automated heads-up for new versions in PR #55268 but also in general. we generally only update mostly casks that have a 'usable' version. i plan to do even more in that regard in the future (e.g,. for casks with appcast) but can only do this for casks where the actual version number is part of the cask version stanza.
4.) as i mentioned in the post above i don't think it makes sense to change the version stanza format around all the time (like in this case) just because upstream changes the separators in the download-url. it makes sense to me that download-url changes should affect the 'url' and version changes should affect the 'version'. i don't see how it makes interpolation more difficult if we keep separators that change all the time in the 'url' instead of the 'version'. all in all i think this makes bump PRs even easier to read?

in any case, the changing of the version number format is a minor point for me personally and i just mentioned it because i thought more consistency would be a good thing - these days i have systems in place that let me know automatically if the version format of a cask changed unexpectedly.

vitorgalvao:
> i am not sure i understand that, because the second sentence seems to contradict the first sentence - as both talk about changing the url stanza.

I meant “the `version` and `sha256`”. Fixed in the post.

As for the rest, you make good points. I don’t mind having the “true” version number in the cask as well, even if we don’t use it (like you added to a few already), and using the comma to separate it from the interpolation string we need for `url`. As for developers who use separators inconsistently, I’m happy to discuss their case, but preferably only after they’re contacted asking for consistency, so we save unnecessary headaches.

PR #55295, opened 2018-11-22
Delete mtmr.rb

vitorgalvao:
Apparently they’re not making later versions publicly available anymore.

https://github.com/Homebrew/homebrew-cask/issues/55289
https://github.com/Toxblh/MTMR/issues/128

PR #55512, opened 2018-11-29
Revert "Delete mtmr.rb"

Toxblh:
Reverts Homebrew/homebrew-cask#55295
Version still is public. But use auto-update not from github for version up one


vitorgalvao:
This doesn’t change anything. Are you still giving people `0.18.5` and having them upgrade to `0.19.0` from inside the app?

That’s exactly what I said we won’t do. That’s not “publicly available”; that’s quite the opposite. Either we get the direct download URL for the latest version of the app, or it stays out of the main Homebrew Cask repos.

vitorgalvao:
I’ve updated the cask in a way that works for us. If that works for you, we can merge it back in.

PR #55513, opened 2018-11-29
Update macx-dvd-ripper-pro.rb

suschizu:
adding appcast-stanza plus version 6.1.1 to macx-dvd-ripper-pro

### DIFF
--- a/Casks/macx-dvd-ripper-pro.rb
+++ b/Casks/macx-dvd-ripper-pro.rb
@@ -1,8 +1,9 @@
 cask 'macx-dvd-ripper-pro' do
-  version :latest
-  sha256 :no_check
+  version '6.1.1'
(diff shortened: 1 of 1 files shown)


suschizu:
i think the travis check is wrong here - using HTTPS here is not a good idea as it just re-directs to the HTTP anyway:
```

curl "https://www.macxdvd.com/mac-dvd-ripper-pro/upgrade/macx-dvd-ripper-pro"
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="http://www.macxdvd.com/mac-dvd-ripper-pro/upgrade/macx-dvd-ripper-pro">here</a>.</p>
</body></html>
```

PR #55548, opened 2018-11-30
Update trinity to 0.4.3

rajivshah3:

### DIFF
--- a/Casks/trinity.rb
+++ b/Casks/trinity.rb
@@ -1,6 +1,6 @@
 cask 'trinity' do
   version '0.4.3'
-  sha256 '7232ca4e50de765e62410dc74bea70019dad14ecb78f998d7e6e1b43c35ceebd'
+  sha256 '0bcccebdc60d95a6c6c0385e460556703a0d7d67d10ad1698ec9701b58888982'
 
   # github.com/iotaledger/trinity-wallet was verified as official when first introduced to the cask
   url "https://github.com/iotaledger/trinity-wallet/releases/download/#{version}/trinity-desktop-#{version}.dmg"


rajivshah3:
It looks like the CI failed because only the SHA256 hash changed. We had to upload new builds because of a bug but did not increase the version.

ran-dall:
@rajatvig All good! ✊ Just remember...

> When updating the sha256 stanza of an existing Cask, the version also has to have changed. Otherwise, the new checksum has to be confirmed. This is necessary to help rule out malicious tampering.
> 
> The confirmation of the updated sha256 should ideally be publicly available. Specifically:
> 
>     Ask the developer to publicly confirm the new checksum and provide a link to this confirmation in your pull request. As examples, confirmation can be acheived by opening an issue of the project’s GitHub, or asking on the developer’s Twitter. Note that a link to an upstream page with the checksum is never sufficient, as a malicious third-party who could modify the download could also modify checksums on the page.
> 
> If the Cask is an .app that is codesigned (in a .dmg or .zip container) it can be uploaded and verified using VirusTotal by looking at the “Details” tab.
**https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md**

I just downloaded it and uploaded it to VirusTotal for the sake of the verification.

alebcay:
Thanks for the quick verification!

rajivshah3:
> When updating the sha256 stanza of an existing Cask, the version also has to have changed. Otherwise, the new checksum has to be confirmed. This is necessary to help rule out malicious tampering.

Ah, I see. Will remember this for next time 👍 thanks @alebcay @ran-dall !

PR #48874, opened 2018-06-23
Add Yandex.Radio v1.5.7(126)

remarkov:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.

Additionally, if **adding a new cask**:

- [x] Named the cask according to the [token reference].
- [x] `brew cask install {{cask_file}}` worked successfully.
- [x] `brew cask uninstall {{cask_file}}` worked successfully.
- [x] Checked there are no [open pull requests] for the same cask.
- [x] Checked the cask was not already refused in [closed issues].
- [x] Checked the cask is submitted to [the correct repo].

[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/Homebrew/homebrew-cask/pulls
[closed issues]: https://github.com/Homebrew/homebrew-cask/issues?q=is%3Aissue+is%3Aclosed
[the correct repo]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask
[version-checksum]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256

PR #55583, opened 2018-11-30
Delete filezilla.rb

suschizu:
Filezilla seems to be a bit evil now: the official download contains now an installer that contains ad- and malware (https://download.filezilla-project.org/client/FileZilla_3.39.0_macosx-x86_setup_bundled.dmg) and the alternative download-link has a download-protection (https://filezilla-project.org/download.php?show_all=1)  
Here is an article to the 3.39.0 release: 
https://www.bleepingcomputer.com/news/security/filezillas-use-of-bundled-offers-sparks-outrage-from-users/


claui:
Just fired up Hopper to look at the FileZilla installer out of curiosity. Whoa, [how](https://www.virustotal.com/en/file/42e598b406b95a731c1067cfb4968e3d888470ba6b9049cb3028aaac473f0734/analysis/1543605606/) it [ticks](https://www.virustotal.com/#/file/aad28f6bebc16d4d93f846f9fffa7e56027762979fba3381637dc0809268cc91/detection) all the [boxes](https://twitter.com/c_pellegrino/status/1069694363360526337).

image

You know what’s the good thing about a clear written policy? You can tell **exactly** when you have to make an exemption from that policy.

claui (approved):

core-code:
thanks for removing it. the last time ( https://github.com/Homebrew/homebrew-cask/pull/48874#issuecomment-401381110 ) i wanted to remove malware from HBC i was told the official policy is to keep it ( https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/apps_with_malware.md ) ;(

claui:
@core-code You’re welcome, and thanks for bringing the case to our attention!

up9cloud:
@suschizu @claui @core-code 

I might wrong, but

Actually, there are 2 version on their server.

One is bundled version: `https://download.filezilla-project.org/client/FileZilla_3.39.0_macosx-x86_setup_bundled.dmg`

One is clean version: `https://download.filezilla-project.org/client/FileZilla_3.39.0_macosx-x86.app.tar.bz2`

The article just said:

> There is also a stark contrast between the installers in terms of how they are detected by antivirus vendors. For example, the bundled installer has a 8/68 detections on VirusTotal, with most detecting it as an adware installer.  The clean version, on the other hand, has 0/68 detections. 

So, can we add it back to the cask? (and passing param --http1.1 to curl to solve the issue #55692)

Thanks.

core-code:
noone claimed that both versions are infected. one version is infected, the other version we had problems with the download protection.

that said, i honestly am unsure why anyone would want to use software from people (even if this particular version is 'clean') if those people also make adware infected versions. they can just infect your supposedly clean version at will too, don't be naive about the technical possibilities. and the will to do 'harm' is obviously there.

vitorgalvao:
Agreed. Unless they’re bought by someone who vows to clean up their act, FileZilla is permanently soiled. Having a “clean” version does not excuse the “dirty” one, especially since they want users to “fall for” the bad one.

> the last time ( [#48874 (comment)](https://github.com/Homebrew/homebrew-cask/pull/48874#issuecomment-401381110) ) i wanted to remove malware from HBC i was told the official policy is to keep it

And to be fair, in that case that was the right decision because that specific software was not malware (not made by the same company).

vitorgalvao:
Pinging @Homebrew/cask just to let everyone know FileZilla [should not be readded](https://github.com/Homebrew/homebrew-cask/pull/55583#issuecomment-443869673) if someone submits it.

up9cloud:
> noone claimed that both versions are infected. one version is infected, the other version we had problems with the download protection.

There is no download protection, you can try `curl -q --http1.1 --output filezilla.app.tar.bz2 https://download.filezilla-project.org/client/FileZilla_3.39.0_macosx-x86.app.tar.bz2`

or manually download at https://download.filezilla-project.org/client/
> 
> that said, i honestly am unsure why anyone would want to use software from people (even if this particular version is 'clean') if those people also make adware infected versions. they can just infect your supposedly clean version at will too, don't be naive about the technical possibilities. and the will to do 'harm' is obviously there.

Know your concern.

but for now, it's clean.

vitorgalvao:
@up9cloud Your workaround may work for downloading FileZilla, yes, but it may also have adverse effects on other casks (changing the command won’t affect just one cask). You likely haven’t tested for that.

And it doesn’t really matter that they have clean and dirty versions. Their dirty version tries to hide it’s dirty, so it’s quite possible the clean version is also bad or will be in the future.

And since I pinged all maintainers so they’re aware, if we keep up the discussion they’ll get spammed to no end. For that reason, I’m locking this particular issue.

We may revisit this decision in the future if something changes, but for now it’s clear FileZilla should not be trusted. If you want it as a cask, consider hosting your own [tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

PR #55199, opened 2018-11-20
Update managenameext to 1.6.6

ran-dall:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #55586, opened 2018-11-30
Update managenameext 1.6.6 - sha256

ran-dall:
This PR addresses the findings reported by @suschizu here: https://github.com/Homebrew/homebrew-cask/pull/55199#issuecomment-443353008

After verifying with Pacifist that the downloaded `.zip` was indeed not signed; I updated the formula accordingly.

screen shot 2018-11-30 at 2 23 44 pm 1

### DIFF
--- a/Casks/managenameext.rb
+++ b/Casks/managenameext.rb
@@ -1,6 +1,6 @@
 cask 'managenameext' do
   version '1.6.6'
-  sha256 :no_check # required as upstream package is updated in-place
+  sha256 '77868a31f6e2e4e4c56922155ce19dbbe0748858ba32f871132d5b1511a3db3c'
 
   url 'http://throb.pagesperso-orange.fr/prg/Xojo/ManageNameExt_c.zip'
   appcast 'http://throb.pagesperso-orange.fr/prg/Xojo/ManageNameExt_AffV.html'
--- a/Casks/managenameext.rb
+++ b/Casks/managenameext.rb
@@ -1,6 +1,6 @@
 cask 'managenameext' do
   version '1.6.6'
-  sha256 'edd09e9fcdd197727f3db310aaa768f6b086a9e55fb15a101cec894402fdb44a'
+  sha256 :no_check
 
   url 'http://throb.pagesperso-orange.fr/prg/Xojo/ManageNameExt_c.zip'
   appcast 'http://throb.pagesperso-orange.fr/prg/Xojo/ManageNameExt_AffV.html'
--- a/Casks/managenameext.rb
+++ b/Casks/managenameext.rb
@@ -1,6 +1,6 @@
 cask 'managenameext' do
   version '1.6.6'
-  sha256 :no_check
+  sha256 :no_check # required as upstream package is updated in-place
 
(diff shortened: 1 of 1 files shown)


suschizu:
great perfect thanks! but is it allowed to have a version and a :no-check?

ran-dall:
@suschizu From what I read and seen, it's entirely possible. I've never had to do it till now but I guess it would fit the use case. @vitorgalvao Maybe you can correct our thinking or verify we're making the right call?

> The special value sha256 :no_check is used to turn off SHA checking whenever checksumming is impractical due to the upstream configuration.
> 
> version :latest requires sha256 :no_check, and this pairing is common. However, sha256 :no_check does not require version :latest.
> 
> We use a checksum whenever possible.

> https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md

vitorgalvao:
> but I guess this would fit the 'use case'.

I’ve pointed out before that bypassing the need for verification is **never** a valid use case for using `:no_check`. That would make no sense: *we* are the ones that make Travis fail in this situation. We can merge a failing build whenever we want (and we do, after verification is provided). It doesn’t matter if the build is green or red; it matters if the cask is correct. Travis is a checker, not a gatekeeper. Using `:no_check` in this case just to make Travis green makes the cask objectively less safe with no benefit other than a meaningless green checkmark.

This requires a `sha256`, so it requires public confirmation that the shasum was modified.

vitorgalvao:
Alternatively, we do want to scrape this policy (see https://github.com/Homebrew/homebrew-cask/issues/51491).

So if we get the documentation/Travis changes for that, we can follow through with it and stop this verification process that nobody likes and never prevented anything.

ran-dall:
@vitorgalvao **Always very appecaitivite of your valuable input.** 👏

However, I think you misread the situation. @suschizu initially reported to me that the sha256 was incorrect on https://github.com/Homebrew/homebrew-cask/pull/55199#issuecomment-443353008. I investigated the matter with @suschizu, and we discovered that the sha256 kept being different. Initially, @suschizu reported that the zip wasn't signed, so I attributed this to that situation. 

I did try to make this work with a sha256, but it kept changing. That said, I don't' think either PR failed the Travis verification; so these don't have anything to do with Travis. 

I think that when I submitted the original PR, the sha256 was correct. After it was merged and @suschizu downloaded it, the sha256 changed, but the version was the same. After he reported it and I redownloaded it to try to fix, the sha256 had changed yet again.

Therefore, I would think that either we'd have to use :no_check or we'd need to remove it.

vitorgalvao:
@ran-dall Ah, I did misread it. Thank you for the clarification.

Casks/managenameext.rb
```diff
@@ -1,6 +1,6 @@
 cask 'managenameext' do
   version '1.6.6'
-  sha256 'edd09e9fcdd197727f3db310aaa768f6b086a9e55fb15a101cec894402fdb44a'
+  sha256 '77868a31f6e2e4e4c56922155ce19dbbe0748858ba32f871132d5b1511a3db3c'
```
vitorgalvao: Let’s give it one more try.

vitorgalvao (approved):

PR #55690, opened 2018-12-03
Update arq-cloud-backup to 1.2.2 - checksum

ran-dall:

### DIFF
--- a/Casks/arq-cloud-backup.rb
+++ b/Casks/arq-cloud-backup.rb
@@ -1,5 +1,5 @@
 cask 'arq-cloud-backup' do
-  version '1.2.1'
+  version '1.2.2'
   sha256 'f027225a30cdf8ae9faf60ddd170967d99614a0d8d8c617ce84ff0c6a319f5e3'
 
   url 'https://www.arqbackup.com/download/arqcloudbackup/ArqCloudBackup.dmg'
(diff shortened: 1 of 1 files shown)


core-code:
in this case the pull request is wrong. the SHA has changed because now it is version 1.2.2.

so, you should just update the version too as keeping it at 1.2.1 would be wrong anyway.

core-code:
for other cases where the version was not updated and it really was a SHA-only change i guess you could either provide proof of valid signature, or if its unsigned just wait until this is merged and the rules have been relaxed: https://github.com/Homebrew/brew/pull/5371

vitorgalvao:
Technically, this one should be `:latest`/`:no_check`. But if it doesn’t cause problems as is, no harm done.

vitorgalvao (approved):

core-code:
we could use this as 

> Technically, this one should be `:latest`/`:no_check`. But if it doesn’t cause problems as is, no harm done.

we could use this as an appcast:
https://www.arqbackup.com/download/arqcloudbackup/arqcloudbackup_release_notes.html

ran-dall:
> btw, you can use 'Suspicious Package' (or 'Pacifist') to look into packages without actually installing them.

@ core-code Yea, this is what I normally do, I use Pacifist.

PR #55709, opened 2018-12-04
Revert "Update axure-rp to 8.0.0.3381"

vitorgalvao:
Reverts Homebrew/homebrew-cask#55695


Trinistar (commented):
the sha256 value is wrong. it should be 08b0460ad0cd00263cea3cc35321dc2b20bf75c208a90d170ab344d59723c5c8 for version 8.1.0.3379

PR #55797, opened 2018-12-05
Update youdaonote to 3.4.1,6311 / 3.4.1,6313

ran-dall:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #55805, opened 2018-12-06
fix: fix YoudaoNote mojave dmg sha256

out0fmemory:
the YoudaoNote mojave dmg sha256 is wrong, it is impossible to be same as the blew high_sierra version

### DIFF
--- a/Casks/youdaonote.rb
+++ b/Casks/youdaonote.rb
@@ -7,7 +7,7 @@ cask 'youdaonote' do
     url 'https://download.ydstatic.com/notewebsite/downloads/YoudaoNote.dmg'
   else
     version '3.4.1,6313'
-    sha256 'bf389b8bc528aae5e16dc1ffcac49b5ce3f41326272d6ee45c08b92d927ff822'
+    sha256 'bb8dceba9d1be665d473294325434a3308f34768c488494cb9361193976040fb'
 
     # download.ydstatic.com/notewebsite/downloads was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


ran-dall:
@out0fmemory NM You got it handled. Good work! 🤝

Maecenas:
I have the same issue when updating cask with [`cask-repair`](https://github.com/vitorgalvao/tiny-scripts/blob/master/cask-repair). It would only download the only file that met only part of the `if…else` statement to update the sha256 checksum of all routes. In this case (#55797), `cask-repair` would only download `MacOS.version <= :high_sierra` and update sha256 of both. Should this be reported in an Issue or PR of the [`tiny-scripts`](https://github.com/vitorgalvao/tiny-scripts)? @vitorgalvao

ran-dall:
@out0fmemory Good effort, gentlemen! 😎 Please tag me in the new PR if you need any help. 

@Maecenas You can always ask me if you need any general help with cask-repair. VG gets real busy sometimes. Anyways, I think that what you’re experiencing is normal by design. When I’m updating casks like this I simply use -e, so...

`cask-repair <cask> -e`

This will open your editor and just put in both `sha256` for each.

vitorgalvao:
> VG gets real busy sometimes.

Not so busy that I can’t answer `cask-repair` questions. Very few people contributed to the script and understand enough of that monstrosity to be able to comment on the internals. Even I don’t remember all the logic I wrote into it.

> In this case (#55797), `cask-repair` would only download `MacOS.version <= :high_sierra`

That PR wasn’t auto-done by `cask-repair`. It can’t auto-make conditionals, so those had to be manually added.

PR #55896, opened 2018-12-08
Update winds to 3.1.4

MartinForReal:

### DIFF
--- a/Casks/winds.rb
+++ b/Casks/winds.rb
@@ -1,9 +1,9 @@
 cask 'winds' do
-  version '2.1.174'
-  sha256 '53a316dc6a229b290eadc088376925d88f59777f73c8d153f74b282c8a41d906'
+  version '3.1.4'
+  sha256 'ebb116a2a3b7396eff1edfce7d69da31ea6f6d66efc04a98f6c260319eda2ba8'
 
   # s3.amazonaws.com was verified as official when first introduced to the cask
-  url "https://s3.amazonaws.com/winds-#{version.major}.0-releases/releases/Winds-#{version}.dmg"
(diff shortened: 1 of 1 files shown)


Casks/winds.rb
```diff
@@ -1,9 +1,9 @@
 cask 'winds' do
-  version '2.1.174'
-  sha256 '53a316dc6a229b290eadc088376925d88f59777f73c8d153f74b282c8a41d906'
+  version '3.1.4'
+  sha256 'ebb116a2a3b7396eff1edfce7d69da31ea6f6d66efc04a98f6c260319eda2ba8'
 
   # s3.amazonaws.com was verified as official when first introduced to the cask
-  url "https://s3.amazonaws.com/winds-#{version.major}.0-releases/releases/Winds-#{version}.dmg"
+  url 'https://s3.amazonaws.com/winds-2.0-releases/releases/Winds-3.1.4.dmg'
```
ran-dall: @MartinForReal The URL is correct. Please revert it back. Should be ok.👌

ran-dall (changes requested):
@MartinForReal The URL that initially was on the cask was correct. You don't' need to update it. 👍

Casks/winds.rb
vitorgalvao: @ran-dall Technically it wasn’t, because the `winds-2.0-releases` part remained at `2`.

vitorgalvao (approved):

Casks/winds.rb
ran-dall: @vitorgalvao Yes, this would have remained at 2. You are correct. I looking at the `#{version}`.

PR #55931, opened 2018-12-10
Update iridium to 2018.11-1

ran-dall:


core-code:
can we keep the real version number in the cask version please? else its impossible to support the app via our cask-repair system...

ran-dall:
@core-code Noted. I'll tweak my setup. 🔧 

In situations like these, would you prefer the real version before the comma or does it not matter for your side of the system? 

So for example `4.2.1,2018.11-1`

ran-dall:
@core-code **Let's do it the way you suggested.**👌 Pretty much my thoughts exactly. 👍

ran-dall:
@core-code You wanna submit a new PR for this?

core-code:
i think @suschizu  has already done this a few days ago ;)

PR #55980, opened 2018-12-11
 p4v 18.3-1719707: Correctly handle spaces in CLI arguments

annettejanewilson:
Problem: If you attempt to invoke, e.g. p4merge on a filename with spaces in it, the filename will end up being broken into multiple incorrect fragments when the wrapped app is actually invoked.

Solution: Use quotes, i.e. `"$@"` instead of a bare `$@`

I'm not clear on the requirements for the cask name and version, or indeed whether cask versions should change when the underlying software doesn't change. This change is just to fix the broken shim, it doesn't change the version of p4v used. Let me know if you need changes.

### DIFF
--- a/Casks/p4v.rb
+++ b/Casks/p4v.rb
@@ -24,7 +24,7 @@ cask 'p4v' do
       #!/bin/bash
       set -euo pipefail
       COMMAND=$(basename "$0")
-      exec "#{appdir}/${COMMAND}.app/Contents/MacOS/${COMMAND}" $@ 2> /dev/null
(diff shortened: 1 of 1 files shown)


vitorgalvao:
No need to check the boxes that don’t apply. In fact, by doing so you prove you have in fact read the rules. You wouldn’t believe (or maybe you would) the amount of people that just check the boxes without actually doing what’s in them. Of course, we always know because then Travis CI fails.

Thank you for the fix, and welcome.

PR #55983, opened 2018-12-11
uninstall.md: pkgutil runs after script

vitorgalvao:
See https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/artifact/abstract_uninstall.rb#L16L17


reitermarkus:
👍

PR #56054, opened 2018-12-12
Update skype to 8.36.0.52

ran-dall:

### DIFF
--- a/Casks/skype.rb
+++ b/Casks/skype.rb
@@ -1,6 +1,6 @@
 cask 'skype' do
-  version '8.34.0.78'
-  sha256 '814b60d85381d4a6332944cdf5f67f0e067b01843821989e685e53b49e6edf80'
+  version '8.36.0.52'
+  sha256 '14805c571946c46be1d3caa04ef83c07399103415cd9f98b814f2947a3f9dafd'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Reverted. Thank you for noticing, @core-code.

> or it fails `cask-repair`

There’s no way for `cask-repair` to distinguish between a beta and a stable version.

I thought you were manually verifying and then sending the PRs, not automating them from the outdated feeds. There needs to be some manual verification precisely due to these cases. Otherwise we would’ve been able to automate everything a while ago.

This is one of the tasks the proposed eventual automation system would handle, picking different feeds for different versions. For example, for GitHub feeds this is doable. Others are case-by-case.

core-code:
there is an 'appcast' that always has the latest official production version of skype here:

`https://get.skype.com/s4l-update?version=8.30.0.0&os=mac&ring=production&app=s4l`

ran-dall:
@core-code Yea I've been parsing the Appcasts but sometimes it doesn't read them right all the way. It's something I'm working on.

@vitorgalvao Yea I saw several things but mine is a  different as I'm using Machine Learning to try to reduce these types of cases. It's not just automated scripting. 🤪 Obviously, looking at each and every app cast would take ages. I've built in a few things in that will cause it to ask for interaction if it senses something is off.

PR #56114, opened 2018-12-14
Update daylite.rb

suschizu:
just the <= high_sierra version was updated

### DIFF
--- a/Casks/daylite.rb
+++ b/Casks/daylite.rb
@@ -1,7 +1,7 @@
 cask 'daylite' do
   if MacOS.version <= :high_sierra
-    version '6.5.5'
-    sha256 '529477859f49bcca08c0704522338201bc0b64de5d58f650c3011d5e711aa223'
+    version '6.5.11'
+    sha256 'fc49b73a961dc4814fcc6dec7179a41d0eccfaa849dcbabb3ea400fe6588f5eb'
     url "https://download.marketcircle.com/daylite/daylitedma#{version.no_dots}.dmg"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> Error: Some launch jobs were not unloaded, add them to uninstall launchctl:
com.marketcircle.Daylite-Helper
com.marketcircle.persistence.autoupdate

@reitermarkus This is one of those that happens with some frequency. Travis complaining of `launchctl` jobs still being present when they are [referenced in the cask](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/daylite.rb#L25L26). So we’re either not removing them correctly, or not checking for their permanence correctly (e.g. maybe the way we check is looking at cached values, hence thinking they’re there when they’re not).

Can’t check if it’s the former case, because my VM crapped itself.

core-code:
is there any point blocking obviously-right changes like these?

i'd say merge if its right, and if there is a travis issue, open an issue for it so that the issue doesn't block the cask updates being merged.

vitorgalvao:
> is there any point blocking obviously-right changes like these?

In this case, it was so [it could be checked](https://github.com/Homebrew/homebrew-cask/pull/56114#issuecomment-447370543) as this might not be a Travis issue, but an HBC issue.

> i'd say merge if its right, and if there is a travis issue, open an issue for it so that the issue doesn't block the cask updates being merged.

But you are correct. Even if there’s something genuinely wrong, it’ll be in the core, not in the cask. We’re moving off Travis anyway, so hopefully some false negative will stop occurring (and maybe we’ll get others).

PR #56123, opened 2018-12-14
Update syncterm to 1.0

core-code:

### DIFF
--- a/Casks/syncterm.rb
+++ b/Casks/syncterm.rb
@@ -6,7 +6,7 @@ cask 'syncterm' do
   url "https://downloads.sourceforge.net/syncterm/syncterm/syncterm-#{version}/SyncTERM.dmg"
   appcast 'https://sourceforge.net/projects/syncterm/rss'
   name 'SyncTERM'
-  homepage 'https://syncterm.bbsdev.net/'
+  homepage 'http://syncterm.bbsdev.net/'
 
   pkg 'SyncTERM.pkg'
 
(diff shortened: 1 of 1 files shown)


core-code:
previous download url is only available via http anymore so maybe switching to sourceforge is better

victorpopkov (approved):

vitorgalvao:
Is this still maintained, even?

core-code:
well, it doesn't look like it, but if we removed everything that didn't have an update in 3 years, half of HBC would be gone.

vitorgalvao:
> half of HBC would be gone.

You’re not giving me a downside!

Joking aside, I ask because we can add a `discontinued` `caveat` if that’s the case.

vitorgalvao:
> i've been marking only things as 'discontinued' if there is an official confirmation from the developer that it is discontinued.
> is this in line with the HBC stance on this? is there any?

Yes, that is our exact policy.

However, we donLt shy away from deducing in some cases. `discontinued` is just a caveat, i.e. a message, so it does impair installation or usage of the app. However, if it ever gives us trouble in the future (dead links, or whatnot), the `discontinued` is a good filter for “how much time should we spend trying to fix this?” and we can have a lower threshold.

If it is ever updated before then, we can always remove the `caveat`.

> is it time to remove 'vudu'? i've been unable to fetch this or access the homepage for a while.

Fine by me. Apps die everyday. I don’t shy away from removing unknown ones, as they’re unlikely to be missed, and if they’re not, that’s a good metric.

core-code:
sounds all good, i'll send a delete request for 'vudu' momentarily.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/56175

Issue #56124, opened 2018-12-14
kicad cask installs into /Applications/Kicad but the contents are empty


bettabacon:
#### Description of issue

`brew cask uninstall kicad`
`brew cask uninstall --force kicad`
and
`brew cask install kicad`
only install an empty Kicad application. I'm running APFS case-sensitive filesystem, and I see in the below debug references to "kicad", "KiCad", and "Kicad" in paths. Maybe this is a case issue?

#### Command that failed

```
brew cask install kicad
```

#### Output of command with `--force --verbose --debug`

```
$ brew cask install kicad --force --verbose --debug
==> Cask::Installer#install
==> Printing caveats
==> Cask::Installer#fetch
==> Satisfying dependencies
==> Downloading
==> Downloading http://downloads.kicad-pcb.org/osx/stable/kicad-unified-5.0.1.dmg
/usr/bin/curl -q --show-error --user-agent Homebrew/1.8.5\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 10.13.6\)\ curl/7.54.0 --location --silent --head http://downloads.kicad-pcb.org/osx/stable/kicad-unified-5.0.1.dmg
Already downloaded: /Users/redacteduser/Library/Caches/Homebrew/downloads/93ba47216c54ba763ecc0fa5aa4d9a7ef977706817694432f8009e37c2d1f34e--kicad-unified-5.0.1.dmg
==> Checking quarantine support
/usr/bin/xattr
/usr/bin/swift /usr/local/Homebrew/Library/Homebrew/cask/utils/quarantine.swift
==> Quarantine is available.
==> Verifying Gatekeeper status of /Users/redacteduser/Library/Caches/Homebrew/downloads/93ba47216c54ba763ecc0fa5aa4d9a7ef977706817694432f8009e37c2d1f34e--kicad-unified-5.0.1.dmg
/usr/bin/xattr -p com.apple.quarantine /Users/redacteduser/Library/Caches/Homebrew/downloads/93ba47216c54ba763ecc0fa5aa4d9a7ef977706817694432f8009e37c2d1f34e--kicad-unified-5.0.1.dmg
==> /Users/redacteduser/Library/Caches/Homebrew/downloads/93ba47216c54ba763ecc0fa5aa4d9a7ef977706817694432f8009e37c2d1f34e--kicad-unified-5.0.1.dmg is quarantined
[...]

PR #56128, opened 2018-12-14
kicad.rb: fix suite

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/issues/56124


zbeekman:
FYI:

>For the time being there have been some packaging related issues with 5.0.2 for macOS, so please get 5.0.1 from http://downloads.kicad-pcb.org/osx/stable/ or hold back upgrading on macOS.

from http://kicad-pcb.org/download/osx/

zbeekman:
Although, I guess my above comment is not pertinent, since this is still on 5.0.1

PR #56175, opened 2018-12-15
Delete vudu.rb

vitorgalvao:
Dead.


core-code:
ah, have been too late, sorry ;)

i am actually not sure whats going on with them, their twitter is still active  but i am unable to access their site or 'brew cask install' it across different countries

PR #56195, opened 2018-12-15
Update psychopy to 3.0.0b13

ran-dall:


core-code:
i think we should wait with upgrading to 3.0 until it is out-of-beta as per our guidelines.

vitorgalvao:
@core-code You’re right. I was distracted by the Github feed. It shows previous betas as pre-releases and this one as a latest release, so it seemed like they were suggesting this one as the one to upgrade to, but the main website still mentions the old version.

core-code:
i agree that the github feed labeling is confusing ... still think we should downgrade until it is out of beta. i can send a PR

vitorgalvao:
Just downgraded. For some reason the Github website was erroring out when creating a branch for it, so I committed directly.

core-code:
ok thanks ;)

PR #56207, opened 2018-12-16
Update mactracker to 7.7.6

shoichiaizawa:


shoichiaizawa:
The checksum was still broken even though it as fixed in #55961.
This commit should fix the formula. Please double-check if it works properly.

vitorgalvao:
> The checksum was still broken even though it was said to be fixed in #55961.

It was not “still broken”, it broke again. Updates are only merged if the checksum is correct.

vitorgalvao:
Travis is weird, so I’m verifying locally.

PR #56212, opened 2018-12-16
Disallow Karabiner on Sierra and later

poiru:
`karabiner` has been superseded by `karabiner-elements` on macOS Sierra
and later. Lets prevent accidental installation of the old package on
such systems.

### DIFF
--- a/Casks/karabiner.rb
+++ b/Casks/karabiner.rb
@@ -8,6 +8,7 @@ cask 'karabiner' do
   homepage 'https://pqrs.org/osx/karabiner/'
 
   auto_updates true
+  depends_on macos: '<= :el_capitan'
 
   pkg 'Karabiner.sparkle_guided.pkg'
   binary '/Applications/Karabiner.app/Contents/Library/bin/karabiner'


Casks/karabiner.rb
```diff
@@ -8,6 +8,7 @@
   homepage 'https://pqrs.org/osx/karabiner/'
 
   auto_updates true
+  depends_on macos: '<= :el_capitan'
```
ran-dall: Right now, you're saying that it needs to be "equal or greater than El Capitan", if you want to do Sierra, then you'd need to use `depends_on macos: '<= :sierra`

ran-dall (changes requested):

Casks/karabiner.rb
vitorgalvao: This looks correct. The top post is that this has already been substituted in Sierra.

vitorgalvao:
Thank you, but since we don’t officially support El Capitan (and earlier), I’d rather remove the cask.

Casks/karabiner.rb
ran-dall: I'm not understanding your comment. If he wanted to disallow on Sierra and below, wouldn't the command be `depends_macos: '< :sierra'

ran-dall: NM I see my error

PR #56247, opened 2018-12-17
Update datadog-agent to 6.7.0-1

core-code:


core-code:
uninstallation with pkgutil does not work as per previous discussion in https://github.com/Homebrew/homebrew-cask/pull/54739 so i tried to change it to delete

vitorgalvao:
I’d like to keep a comment-out `pkgutil` with an explanation on why it isn’t used.

That’ll make it easier to revisit it once in a while, to see if the situation improved, and nobody else will be tempted to add it if it hasn’t.

vitorgalvao (approved):

PR #56283, opened 2018-12-18
Update policy on apps with malware

vitorgalvao:
Started from [@core-code’s text](https://github.com/Homebrew/homebrew-cask/issues/56223#issuecomment-448006552).


doc/faq/apps_with_malware.md
```diff
@@ -1,13 +1,13 @@
...
 
-* Go to [Apple’s Bug Reporter](https://bugreport.apple.com/) and report the app that bundles malware. Be as precise as possible about how you know it bundles malware, and what steps reproduce your conclusions. Be sure to include relevant URLs, such as the app’s homepage.
-* Make the report public at [Open Radar](http://www.openradar.me/).
-* Submit a pull request with a [`malware` `caveat`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md#caveats-mini-dsl).
+Likelwise, software that provides both “clean” and malware-infested versions might be removed from the repo — even if we could have access to the *good* version — if its developers push for users to install the *bad* version. We do so because in these cases, there’s a higher than normal risk that both versions are (or will soon become) compromised in some manner.
```
colindean: `Likewise`

colindean: Maybe use "malicious" instead of bad and "clean" instead of good.

doc/faq/apps_with_malware.md
```diff
@@ -1,13 +1,13 @@
...
+
+If a cask you depend on was removed due to these rules, fear not. Removal of a cask from the official repositories means we won’t support it, but you can do so by hosting your own [tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).
```
colindean: Is it worth explaining that the cask could be resubmitted once the maliciousness is mitigated?

colindean (changes requested):

doc/faq/apps_with_malware.md
vitorgalvao: I’d like for the words to be somewhat antonyms to drive the point home harder. While “malicious” is explicit, “clean” is a metaphor and was used before.

doc/faq/apps_with_malware.md
vitorgalvao: That’s not something we necessarily want. If someone has malicious apps *on purpose*, I don’t want them to be able to say “oh, never mind, we’re clean now” and have them reinstated like that. Trust lost should be harder to regain.

doc/faq/apps_with_malware.md
colindean: Maybe "friendly" or "user-friendly" or just plain ol' "good" instead of clean, then.

But you've made a point: "bad" and "good" are pretty explicitly opposites.

PR #56557, opened 2018-12-23
Update rescuetime to 2.14.2.3

jinntrance:


reitermarkus:
This doesn't have an appcast and the URL is unversioned, so it needs to keep using `:latest`.

PR #56677, opened 2018-12-26
Update polyphone to 2.0

core-code:


vitorgalvao:
Travis is having trouble with the download:

> curl: (22) The requested URL returned error: 416
Download failed on Cask 'polyphone' with message: Download failed: https://www.polyphone-soundfonts.com/en/download/file/618-polyphone-2-0-app-zip/latest/download

PR #56802, opened 2018-12-30
Delete gcc-arm-embedded.rb

vitorgalvao:
Isn’t this open source (and CLI-only)?

Is there a reason this can’t be moved to Homebrew?


reitermarkus (approved):

alexeden:
@vitorgalvao I was in the process of installing the necessary tools for working with embedded Rust, which requires this cask. [The documentation](https://docs.rust-embedded.org/book/intro/install/macos.html) specifies the command `brew cask install gcc-arm-embedded` which obviously failed, and that led me here.

My knowledge of Homebrew in general is limited... is there an alternative means via `brew` by which I can get this package?

jymoulin:
Why was it deleted ? Just because of gdb ? Why not fall back to previous version ?

vitorgalvao:
> Why was it deleted ?

Because it should be a formula, not a cask, as it is a CLI-only open-source tool. This policy is documented in [faq/rejected_casks](https://github.com/homebrew/homebrew-cask/blob/master/doc/faq/rejected_casks.md):

In an ideal world, we could have opened this PR for removal and have merged it only after the formula is made. In the real world, however, it has been shown to us time and time again that had we done that, this removal PR would’ve been open indefinitely. No matter where we post these notices (here, twitter), people only react after it’s gone.

Occasionally it’s then we see that we should’ve kept it and revert the decision, but removing ends up being the right choice nine times out of ten (if not more). For now, this should be a formula and you should direct your efforts in that regard.

> [The documentation](https://docs.rust-embedded.org/book/intro/install/macos.html) specifies the command `brew cask install gcc-arm-embedded`

That’s unfortunate. Naturally, we can’t know everywhere our commands are used as documentation, and so can sometimes cause unwanted breakage. This might be an opportunity, though. Maybe they (or someone in the community) will want to create the formula.

machinaut:
I ran into this (like many other folks) because I'm trying to install `gcc-arm-embedded` while it is in limbo -- removed from cask but not yet added to core.

The cask installer can take a URL to a script, and you can point it at the commit before this file like so

```
brew cask install https://raw.githubusercontent.com/Homebrew/homebrew-cask/b88346667547cc85f8f2cacb3dfe7b754c8afc8a/Casks/gcc-arm-embedded.rb
```

Hopefully this is useful who came here looking for a temporary solution to `brew cask install gcc-arm-embedded`.

tannewt:
Is anyone working on a core brew formula?

vitorgalvao:
@tannewt I doubt it. If you want to do it, you should go ahead.

fxcoudert:
FYI: this was rejected from Homebrew core, due to a build process that is way to difficult for us to accommodate https://github.com/Homebrew/homebrew-core/pull/45780

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/75091

PR #56884, opened 2019-01-02
Add/tweak probots configuration.

MikeMcQuaid:
This adds the other probots used by the other Homebrew taps and tweaks the
existing configuration to be consistent with the other Homebrew taps.

Feel free to disagree on any of the individual changes, I am very open to
discussion.

Once these are all addressed and this is merged I will open the agreed version
on all the other Cask taps.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -10,6 +10,6 @@ lockComment: false
 # Limit to only `issues` or `pulls`
 # only: issues
 # Add a label when locking. Set to `false` to disable
-lockLabel: false
+lockLabel: outdated
 # Assign `resolved` as the reason for locking. Set to `false` to disable
-setLockReason: false
+setLockReason: resolved
--- a/.github/support.yml
+++ b/.github/support.yml
@@ -1,0 +1,16 @@
+# Configuration for support-requests - https://github.com/dessant/support-requests
+
+# Label used to mark issues as support requests
+supportLabel: support request
+# Comment to post on issues marked as support requests. Add a link
+# to a support page, or set to `false` to disable
+supportComment: >
+  We use the issue tracker exclusively for bug reports.
+  However, this issue appears to be a support request. Please use our
+  [Discourse](https://discourse.brew.sh) or
+  [IRC channel](irc://irc.freenode.net/#machomebrew) to get help with
+  the project.
+# Whether to close issues marked as support requests
+close: true
+# Whether to lock issues marked as support requests
+lock: true
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 21
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 7
+# Issues or Pull Requests with these labels will never be considered stale
+exemptLabels:
+  - gsoc-outreachy
+  - help wanted
+  - in progress
+# Label to use when marking as stale
+staleLabel: stale
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs.


vitorgalvao:
There’s a stray commit for `mkvtoolnix-app`.

.github/lock.yml
```diff
@@ -10,6 +10,6 @@ lockComment: false
 # Limit to only `issues` or `pulls`
 # only: issues
 # Add a label when locking. Set to `false` to disable
-lockLabel: false
+lockLabel: outdated
```
vitorgalvao: Why the label? Seems to me locking as “resolved” should be sufficient.

.github/no-response.yml
```diff
@@ -3,7 +3,7 @@
 # Number of days of inactivity before an Issue is closed for lack of response
 daysUntilClose: 14
 # Label requiring a response
-responseRequiredLabel: awaiting user reply
+responseRequiredLabel: needs response
```
vitorgalvao: I like the `awaiting user reply` label, as it has a nice parallel with `awaiting maintainer feedback`. That said, I’m willing to change them for consistency.

.github/support.yml
```diff
@@ -0,0 +1,16 @@
+# Configuration for support-requests - https://github.com/dessant/support-requests
```
vitorgalvao: Is there a similar bot that supports multiple labels (with a different message for each)?

.github/support.yml
```diff
@@ -0,0 +1,16 @@
...
+
+# Label used to mark issues as support requests
+supportLabel: support-request
```
vitorgalvao: Should be hyphenless, for consistency with other labels.

.github/lock.yml
MikeMcQuaid: @vitorgalvao It makes it easier to quickly query for locked issues which have other labels.

.github/no-response.yml
MikeMcQuaid: Happy to leave them as-is; don't think everything needs to be consistent.

.github/support.yml
MikeMcQuaid: Not that I've seen 😢

.github/lock.yml
reitermarkus: > It makes it easier to quickly query for locked issues which have other labels.

Can't you just search for the other labels then?

MikeMcQuaid: That doesn't allow differentiation between locked/not locked or autolocked/not autolocked from the search interface.

MikeMcQuaid: For the sake of getting this merged can you agree whether this is blocking or not?

vitorgalvao (approved):

vitorgalvao:
@MikeMcQuaid Mind removing https://github.com/Homebrew/homebrew-cask/blob/master/.github/no-response.yml?

`stale` makes that one redundant, and I never liked it much anyway (doesn’t close PRs, only issues).

MikeMcQuaid:
> @MikeMcQuaid Mind removing https://github.com/Homebrew/homebrew-cask/blob/master/.github/no-response.yml?
> 
> `stale` makes that one redundant, and I never liked it much anyway (doesn’t close PRs, only issues).

@vitorgalvao I'd rather it's left to be consistent with other repos. I think it's a useful pattern to be able to adjust that separately.

vitorgalvao:
> I'd rather it's left to be consistent with other repos.

Why do other repos have that, though?

I don’t think anyone in HBC particularly likes that bot. In fact, I’ll go so far as to say it’s outright useless for us, as it *never* acts. Issues never get the label because either:

1. The template was filled, so we already got what we need.
2. The template was not filled, so the issue is closed.

And the only place where we do need it (PRs), it does nothing. The only reason that bot was still active was in the hopes it eventually supported our use case (PRs). Since we’re adding the `stale` bot, `no-response` can go.

.github/lock.yml
```diff
@@ -10,6 +10,6 @@ lockComment: false
...
 # Assign `resolved` as the reason for locking. Set to `false` to disable
-setLockReason: false
+setLockReason: resolved
```
vitorgalvao: This is causing an error in all repos. Looks like it needs to be a boolean (even though they still have that `resolved` comment in their example).

Also, @MikeMcQuaid, note you don’t need to PR every repo to sync these settings, [we have a script for it](https://github.com/Homebrew/homebrew-cask/blob/master/developer/bin/sync_cask_tap_templates).

vitorgalvao: @MikeMcQuaid The bot’s opening issues: https://github.com/Homebrew/homebrew-cask/issues/57088

MikeMcQuaid: @vitorgalvao Sorry about that, will fix.

dessant:
Make sure to put the label name in quotes if it contains spaces, otherwise config parsing will fail:
https://github.com/Homebrew/homebrew-cask/blob/master/.github/support.yml#L4

MikeMcQuaid:
Thanks @dessant!

@vitorgalvao opened https://github.com/Homebrew/homebrew-cask/pull/57222 and https://github.com/Homebrew/homebrew-cask-versions/pull/6821, sorry for the delay.

PR #56920, opened 2019-01-04
Add cinebench R15.038 zap

ytbmulder:

### DIFF
--- a/Casks/cinebench.rb
+++ b/Casks/cinebench.rb
@@ -6,8 +6,6 @@ cask 'cinebench' do
   name 'Cinebench'
   homepage 'https://www.maxon.net/products/cinebench/'
 
-  depends_on macos: '>= :snow_leopard'
-
   app "CINEBENCH #{version}"
 
   zap trash: [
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):
HBC itself won’t work below Snow Leopard, so adding that as a dependency is redundant.

Casks/cinebench.rb
```diff
@@ -7,4 +7,9 @@
   homepage 'https://www.maxon.net/products/cinebench/'
 
   app "CINEBENCH #{version}"
+
+  zap trash: [
+               '~/Library/Caches/net.maxon.cinema4d',
```
vitorgalvao: Are you sure?

ytbmulder: I tested it again to be sure. I removed the directory: 

```
$ ls -l ~/Library/Caches | grep -ri 'net.*'
```

Did a clean install of the cask: 
```
$ brew cask install cinebench
```

Opened the application, ran both benchmarks and checked the path:
```
$ ls -l ~/Library/Caches | grep -ri 'net.*'
drwxr-xr-x Jan  6 11:07 net.maxon.cinema4d
```

PR #57001, opened 2019-01-05
superduper: Add support for mavericks

Amorymeltzer:

### DIFF
--- a/Casks/superduper.rb
+++ b/Casks/superduper.rb
@@ -1,10 +1,37 @@
 cask 'superduper' do
-  version '3.2.4,118'
-  sha256 'a6af47ea7df3903ba7195dfb25eb57d4150b02d8a910c0a52ab7297fb588e3e9'
+  if MacOS.version <= :snow_leopard
+    version '2.7.1'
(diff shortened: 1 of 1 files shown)


vitorgalvao (changes requested):
Thank you, but no. That’s too much. We don’t officially support that far back.

We can keep from `mavericks` on, but no earlier. Also, the appcast needs to be outside the conditionals.

vitorgalvao:
> would a PR clarifying that intent in the docs be welcome?

We have a line in the docs somewhere stating we only officially support the current macOS release and one back.

> Regarding the appcast

Since we changed how they operate, outside might make more sense now.

Amorymeltzer:
Done.

> we only officially support the current macOS release and one back.

Indeed; I meant the "We can keep from `mavericks` on, but no earlier" bit.

vitorgalvao:
> I meant the "We can keep from `mavericks` on, but no earlier" bit.

Oh, that’s unofficial. That’s me saying “in this case it doesn’t hurt, and since you’re a trusted maintainer that went to the trouble of adding these and may require older versions, fine by me to bend the rules a bit (but let’s have a cap on how much)”. I don’t shy from doing so once in a while for good contributors.

PR #57083, opened 2019-01-06
Delete cocos-code-ide.rb

core-code:
as far as i can see (please double check) this has been deprecated for years and is not available anymore


vitorgalvao:
I can’t even access the website.

core-code:
weirdly accessing the website works if you omit the www.

PR #57089, opened 2019-01-06
Update musixmatch to 0.21.13

core-code:

### DIFF
--- a/Casks/musixmatch.rb
+++ b/Casks/musixmatch.rb
@@ -1,8 +1,8 @@
 cask 'musixmatch' do
-  version :latest
-  sha256 :no_check
+  version '0.21.13'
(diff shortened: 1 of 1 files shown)


core-code:
not especially happy to convert this to versioned, but the cask didn't work as it was anymore

vitorgalvao:
Can’t even verify locally. Download is unbearably slow.

core-code:
i had the 'freezing' problem only before the changes i've added. the cask repair only took half a minute

core-code:
not sure what is going on here, the download takes 8 seconds here

vitorgalvao:
Got it locally.

core-code:
we could use this as an appcast but there has been some resistance to adding 'dependencies to external servers': 

http://www.corecode.io/cgi-bin/check_urls/check_url_filename.cgi?url=http://download-app.musixmatch.com/download/osx

PR #57127, opened 2019-01-07
Update mailmate to 5579

core-code:


core-code:
rationale here is that i can automatically check non-html appcasts every day for updates, but am not doing HTML because of the noise issue

PR #57136, opened 2019-01-07
Delete electronic-wechat.rb

suschizu:
https://github.com/geeeeeeeeek/electronic-wechat
active development has discontinued


core-code:
i think we should mark it as discontinued instead

PR #57152, opened 2019-01-07
Update qownnotes to 19.1.0,b4034-122701

core-code:


core-code:
the github feed is quite useless in this case cause they 1.) post a lot of dev releases and 2.) insist on labelling these releases as 'latest release' instead of properly as 'prerelease'

core-code:
note that i may have got them to fix this ( https://github.com/pbek/QOwnNotes/issues/1109 ) so if someone prefers the github feed this should work better now

Issue #56815, opened 2018-12-31
Consider using Azure Pipelines instead of Travis CI


MikeMcQuaid:
### Description of feature/enhancement

Homebrew has migrated to using Azure Pipelines instead of Travis CI. Homebrew Cask should consider doing the same.

### Justification

The Azure Pipelines free macOS workers are much more performant than the Travis CI ones.

### Example use case

PRs are built with Azure Pipelines rather than Travis CI.

vitorgalvao:
Fine by me.

shikhar-scs:
@MikeMcQuaid I guess this lies along the lines of https://github.com/Homebrew/brew/pull/5457/files. If agreed upon, may I take this up ?

reitermarkus:
@shikhar-scs, go for it. 👍

shikhar-scs:
@reitermarkus I'd started with the work on `azure-pipelines`. I wrote the [azure-pipelines file](https://github.com/Homebrew/homebrew-cask/compare/master...shikhar-scs:azure-pipelines?expand=1#diff-fec826feae04e51c0d94076385408bdc) , taking help from the travis.yml file in the directory and then ran a build on the pipelines portal on azure devops. Here's where I'm getting stuck 
<img width="716" alt="screenshot 2019-01-07 at 9 38 09 am" src="https://user-images.githubusercontent.com/25258877/50748643-73c59600-1260-11e9-8e63-a72700c09431.png">

The complete log file is here : [6.txt](https://github.com/Homebrew/homebrew-cask/files/2731382/6.txt)

 Could you take a look please ? Also, is this the correct way ?

vitorgalvao:
@shikhar-scs You’re getting that error because the `brew cask ci` command is [expecting a `CI` environment key](https://github.com/Homebrew/homebrew-cask/blob/7c37f9032d986c3dc3f984a61ff1e183fe9b61bb/cmd/brewcask-ci.rb#L14).

reitermarkus:
Yes, you'll need to adjust [`brewcask-ci.rb`](https://github.com/Homebrew/homebrew-cask/blob/master/cmd/brewcask-ci.rb), which is currently pretty specific to Travis.

Also, please open a PR sooner rather than later, since it's much easier to review that way.

shikhar-scs:
Thanks @reitermarkus & @vitorgalvao . I'll include a sample key as of now and debug further.

[...]

PR #57156, opened 2019-01-08
Azure pipelines for CI instead of Travis #56815

shikhar-scs:
Since this is unrelated to any of the cask files, none of the checkboxes fit here.
Refer #56815

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,10 +10,9 @@ jobs:
         brew update
         # Mirror the repo as a tap.
         HOMEBREW_TAP_DIR="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask"
-        mkdir -p "$HOMEBREW_TAP_DIR"
+        mkdir -p "${HOMEBREW_TAP_DIR}"
         rsync -az --delete "${HOMEBREW_TAP_DIR}/"
         export TRAVIS_BUILD_DIR="${HOMEBREW_TAP_DIR}"
         builtin cd "${TRAVIS_BUILD_DIR}"
         brew cask ci
       displayName: Run brew cask-test-bot
-      
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,7 +11,7 @@ jobs:
         # Mirror the repo as a tap.
         HOMEBREW_TAP_DIR='/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask'
         mkdir -p "${HOMEBREW_TAP_DIR}"
-        rsync -az --delete "${HOMEBREW_TAP_DIR}"
+        rsync --archive --compress --delete "${HOMEBREW_TAP_DIR}"
         export AZURE_BUILD_DIR="${HOMEBREW_TAP_DIR}"
         cd "${AZURE_BUILD_DIR}"
         brew cask ci
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,18 @@
+jobs:
+- job: macOS
+  pool:
+    vmImage: xcode9-macos10.13
+  steps:
+    - bash: |
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        HOMEBREW_TAP_DIR="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask"
+        mkdir -p "$HOMEBREW_TAP_DIR"
+        rsync -az --delete "${HOMEBREW_TAP_DIR}/"
+        export TRAVIS_BUILD_DIR="${HOMEBREW_TAP_DIR}"
+        builtin cd "${TRAVIS_BUILD_DIR}"
+      displayName: Run brew cask-test-bot
+      
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,5 +14,6 @@ jobs:
         rsync -az --delete "${HOMEBREW_TAP_DIR}/"
         export TRAVIS_BUILD_DIR="${HOMEBREW_TAP_DIR}"
         builtin cd "${TRAVIS_BUILD_DIR}"
+        brew cask ci
       displayName: Run brew cask-test-bot
       
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,10 +9,10 @@ jobs:
         export HOMEBREW_NO_AUTO_UPDATE=1
         brew update
         # Mirror the repo as a tap.
-        HOMEBREW_TAP_DIR="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask"
+        HOMEBREW_TAP_DIR='/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask'
         mkdir -p "${HOMEBREW_TAP_DIR}"
-        rsync -az --delete "${HOMEBREW_TAP_DIR}/"
-        export TRAVIS_BUILD_DIR="${HOMEBREW_TAP_DIR}"
-        builtin cd "${TRAVIS_BUILD_DIR}"
+        rsync -az --delete "${HOMEBREW_TAP_DIR}"
+        export AZURE_BUILD_DIR="${HOMEBREW_TAP_DIR}"
+        cd "${AZURE_BUILD_DIR}"
         brew cask ci
       displayName: Run brew cask-test-bot
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,10 +9,9 @@ jobs:
         export HOMEBREW_NO_AUTO_UPDATE=1
         brew update
         # Mirror the repo as a tap.
-        HOMEBREW_TAP_DIR='/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask'
+        HOMEBREW_TAP_DIR=Agent.BuildDirectory
         mkdir -p "${HOMEBREW_TAP_DIR}"
         rsync --archive --compress --delete "${HOMEBREW_TAP_DIR}"
-        export AZURE_BUILD_DIR="${HOMEBREW_TAP_DIR}"
-        cd "${AZURE_BUILD_DIR}"
+        cd "${HOMEBREW_TAP_DIR}"
         brew cask ci
       displayName: Run brew cask-test-bot
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,9 +9,9 @@ jobs:
         export HOMEBREW_NO_AUTO_UPDATE=1
         brew update
         # Mirror the repo as a tap.
+        ln -s "$PWD" $(Build.SourcesDirectory)
         mkdir -p $(Build.SourcesDirectory)
         export AZURE_BUILD_DIR=$(Build.SourcesDirectory)
         cd $(Build.SourcesDirectory)
-        ln -s "$PWD" $(Build.SourcesDirectory)
         brew cask ci
       displayName: Run brew cask-test-bot
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,6 +11,7 @@ jobs:
         # Mirror the repo as a tap.
         mkdir -p $(Build.SourcesDirectory)
         rsync --archive --compress --delete $(Build.SourcesDirectory)
+        export AZURE_BUILD_DIR=$(Build.SourcesDirectory)
         cd $(Build.SourcesDirectory)
         ls $(Build.SourcesDirectory)
         brew cask ci
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,7 +11,7 @@ jobs:
         # Mirror the repo as a tap.
         mkdir -p $(Build.SourcesDirectory)
         rsync --archive --compress --delete $(Build.SourcesDirectory)
+        # cd $(Build.SourcesDirectory)
         ln -s "$PWD" $(Build.SourcesDirectory)
-        cd $(Build.SourcesDirectory)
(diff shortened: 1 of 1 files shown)


shikhar-scs:
@reitermarkus I ran a build today. It gave a green tick (a successful build) on the Azure portal.
screenshot 2019-01-08 at 9 35 16 am


However, there was also this error at the end of the build report 
screenshot 2019-01-08 at 9 33 21 am

I then ran the `brew cask ci` command after getting in the `HOMEBREW_TAP_DIRECTORY` and then did the `rm -rf`. Didnt work.

Also, the changes I'm making in the brewcask-ci.rb file isnt reflecting in the build. I tried using some `puts` statements to debug, also tried removing a few lines of code and stuff, however it still makes the build as if no changes were made, throwing the same errors again and again.

reitermarkus:
What exactly do you mean with `HOMEBREW_TAP_DIRECTORY`? The command expects you to be in e.g. `/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask`.

azure-pipelines.yml
```diff
@@ -0,0 +1,20 @@
...
+  pool:
+    vmImage: xcode9-macos10.13
+  steps:
+    - bash: |
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        HOMEBREW_TAP_DIR="/usr/local/Homebrew/Library/Taps/#{tap}"
```
reitermarkus: `#{tap}` is neither set nor is it Bash syntax.

azure-pipelines.yml
```diff
@@ -0,0 +1,20 @@
...
+        mkdir -p "$HOMEBREW_TAP_DIR"
+        cd "$HOMEBREW_TAP_DIR"
+        brew cask ci
+        rm -rf "$HOMEBREW_TAP_DIR"
+      displayName: Run brew cask-test-bot
+      env:
+        HOMEBREW_GITHUB_API_TOKEN: $(github.publicApiToken)
+        HOMEBREW_COVERALLS_REPO_TOKEN: $(coveralls.homebrewBrewApiToken)
```
reitermarkus: We don't need this for casks.

shikhar-scs: Cool. I'll remove this.

azure-pipelines.yml
shikhar-scs: True. I noticed that now. I'll correct it too.

azure-pipelines.yml
shikhar-scs: @reitermarkus I made the changes, and now I'm getting this. Can it be related to Travis ? 
screenshot 2019-01-08 at 7 25 26 pm

I checked up https://github.com/Homebrew/homebrew-core/issues/31840 and it was mentioned that `.to_str` is an upstream issue. 

Also, how could I force the build to run the `brewcask-ci.rb` file that I'm pushing and not the one in the system. This is essential to allow me to debug.

reitermarkus:
@shikhar-scs, the next commit should actually run on our Azure pipeline, which will probably solve the issue of it not using the proper `brewcask-ci.rb`.

azure-pipelines.yml
```diff
@@ -0,0 +1,17 @@
...
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        HOMEBREW_TAP_DIR="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask"
+        mkdir -p "$HOMEBREW_TAP_DIR"
+        cd "$HOMEBREW_TAP_DIR"
+        brew cask ci
+        rm -rf "$HOMEBREW_TAP_DIR"
+      displayName: Run brew cask-test-bot
```
vitorgalvao: It has a “no newline at end of file” warning. Those are [actually meaningful](http://unix.stackexchange.com/a/18789).

vitorgalvao (changes requested):

azure-pipelines.yml
shikhar-scs: @vitorgalvao I had tried the same but still got the error
screenshot 2019-01-08 at 7 25 26 pm

reitermarkus:
You need to do the same thing the Travis script does: `rsync` the current repo to the correct tap directory.

shikhar-scs:
@reitermarkus I did the changes. Thanks a lot for your help. Its working now. Please have a look at the azure build.

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+  steps:
+    - bash: |
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        HOMEBREW_TAP_DIR="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask"
+        mkdir -p "${HOMEBREW_TAP_DIR}"
+        rsync -az --delete "${HOMEBREW_TAP_DIR}/"
```
vitorgalvao:
```suggestion
        rsync --archive --compress --delete "${HOMEBREW_TAP_DIR}"
```

Always use long flags, as they are clearer.

Also, is the trailing slash required?

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+  pool:
+    vmImage: xcode9-macos10.13
+  steps:
+    - bash: |
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        HOMEBREW_TAP_DIR="/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask"
```
vitorgalvao:
```suggestion
        HOMEBREW_TAP_DIR='/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask'
```

Single quotes when interpolation isn’t needed.

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+        mkdir -p "${HOMEBREW_TAP_DIR}"
+        rsync -az --delete "${HOMEBREW_TAP_DIR}/"
+        export TRAVIS_BUILD_DIR="${HOMEBREW_TAP_DIR}"
```
vitorgalvao: Since Travis is no longer the CI, this variable should be renamed.

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+        builtin cd "${TRAVIS_BUILD_DIR}"
```
vitorgalvao: Is `builtin` really necessary?

shikhar-scs: I tried to replicate the code written in the `travois.yml` file. That's it. I'll remove the Travis Specific Code.

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+        brew cask ci
+      displayName: Run brew cask-test-bot
```
vitorgalvao: Shouldn’t this start with `-` as well? As in:

```suggestion
      - displayName: Run brew cask-test-bot
```

Also, it’s my understanding this is pretty much just a message. So shouldn’t we change this for something more descriptive and that looks less like an outright command?

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+  steps:
+    - bash: |
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        HOMEBREW_TAP_DIR='/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask'
+        mkdir -p "${HOMEBREW_TAP_DIR}"
+        rsync -az --delete "${HOMEBREW_TAP_DIR}"
```
vitorgalvao:
```suggestion
        rsync --archive --compress --delete "${HOMEBREW_TAP_DIR}"
```

Again, we should use long flags whenever they exist.

azure-pipelines.yml
vitorgalvao: Finally, having it at the top might be clear visually.

shikhar-scs: > Shouldn’t this start with - as well?
[Not really](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=vsts&tabs=schema#tabpanel_8ugYiRH0yv-6_schema) it works like this too.

> Finally, having it at the top might be clear visually.

Sure.

azure-pipelines.yml
shikhar-scs: my fault sorry

shikhar-scs: Its still bringing the same error :(

vitorgalvao: > Its still bringing the same error :(

Note the changes I’m suggesting are not to fix the errors (that’s the area @reitermarkus), but to tidy everything up consistently.

shikhar-scs: Oops. Sure. Thanks for the help. 👍

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
+jobs:
+- job: macOS
+  pool:
+    vmImage: xcode9-macos10.13
+  steps:
+    - bash: |
```
reitermarkus: This also needs the other flags Travis uses, like `set -e`, so that any error results in a failure.

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        HOMEBREW_TAP_DIR='/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask'
```
reitermarkus: This should be built from a variable. The current repo should be available from some environment variable provided by Azure.

reitermarkus (changes requested):

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+        mkdir -p "${HOMEBREW_TAP_DIR}"
+        rsync --archive --compress --delete "${HOMEBREW_TAP_DIR}"
+        export AZURE_BUILD_DIR="${HOMEBREW_TAP_DIR}"
```
reitermarkus: Is this actually the variable name Azure uses initially? In the Travis case, this was only re-exported because it is actually a variable initially provided by Travis.

azure-pipelines.yml
shikhar-scs: Oh. I'll look into that.

azure-pipelines.yml
shikhar-scs: I guess it does not.

shikhar-scs:
@reitermarkus . Sorry I was unavailable for a week, got back this weekend only. 😬 

> This should be built from a variable. The current repo should be available from some environment variable provided by Azure.

I looked this up and made them changes accordingly. Have a look at the current `azure-pipelines.yml` file.

However, I'm getting the same old error, that the command `brew cask ci` should be run from inside the tap directory. I double checked that we already were there using `ls` and everything but the error still persists. 
Here's the output of `ls`
screenshot 2019-01-20 at 9 07 29 am

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+    vmImage: xcode9-macos10.13
+  steps:
+    - bash: |
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
+        mkdir -p $(Build.SourcesDirectory)
+        rsync --archive --compress --delete $(Build.SourcesDirectory)
```
reitermarkus: What is this supposed to do? It doesn't sync anywhere.

shikhar-scs: It was used in the `travis.yml` file, so I guess it does have some significance. Maybe, if not, we can remove it ?

reitermarkus: What I mean is, this is missing the target directory parameter, so it isn't doing anything.

shikhar-scs: Oh, Cool. But what about the `This command must be run from inside the Tap Directory` error ?

reitermarkus: What I mean to say is, this *should* be doing something, that is syncing the build directory into the tap directory, which is why the command is not running inside a tap directory.

shikhar-scs: Oh I get it. Correct. Let me look out for the same.

azure-pipelines.yml
```diff
@@ -0,0 +1,17 @@
...
+- job: macOS
+  pool:
+    vmImage: xcode9-macos10.13
+  steps:
+    - bash: |
+        export HOMEBREW_COLOR=1
+        export HOMEBREW_DEVELOPER=1
+        export HOMEBREW_NO_AUTO_UPDATE=1
+        brew update
+        # Mirror the repo as a tap.
```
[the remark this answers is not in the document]
reitermarkus: An advantage of syncing is that we can fix the CI script and it will be used when restarting failed builds.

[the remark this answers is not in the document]
shikhar-scs: symlink isn't working. The error persists.
Also, a combo of rsync and symlink fails too.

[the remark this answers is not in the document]
reitermarkus: You still aren't syncing into the tap directory.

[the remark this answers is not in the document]
shikhar-scs: Then how do I fix that

PR #57199, opened 2019-01-08
Remove Tiger references.

MikeMcQuaid:
Homebrew/brew has never supported these and the symbols have been removed in https://github.com/Homebrew/brew/pull/5477.


MikeMcQuaid:
@reitermarkus 👍d merging this as-is in Slack.

PR #57125, opened 2019-01-07
add qlImageSize.qlgenerator v1.6.0

L1cardo:
After making all changes to the cask:

- [Yes] `brew cask audit --download {{cask_file}}` is error-free.
- [Yes] `brew cask style --fix {{cask_file}}` reports no offenses.
- [Yes] The commit message includes the cask’s name and version.
- [Yes] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #57212, opened 2019-01-09
add qlImageSize.qlgenerator v1.6.0

L1cardo:
Owner no longer provides compiled versions. So the homepage is the origin owner's repo , and the url is my repo, like below

I really hope it can be merged , thanks !

### DIFF
--- a/Casks/qlimagesize.rb
+++ b/Casks/qlimagesize.rb
@@ -4,7 +4,7 @@ cask 'qlimagesize' do
 
   url 'https://github.com/L1cardo/qlImageSize/releases/download/1.6.0/qlImageSize.qlgenerator.zip'
   name 'qlImageSize'
-  homepage 'https://github.com/Nyx0uf/qlImageSize'
+  homepage 'https://github.com/L1cardo/qlImageSize'
 
   qlplugin 'qlImageSize.qlgenerator'
 end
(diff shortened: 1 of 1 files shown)


reitermarkus:
From https://github.com/Homebrew/homebrew-cask/pull/57125:

> It is not possible for me to convince the owner.

What did upstream say?

L1cardo:
> From #57125:
> 
> > It is not possible for me to convince the owner.
> 
> What did upstream say?

He said he won't make any build . He demands people to compile by themselves .

L1cardo:
Since the owner will not host the project , I will host it . 
So , the homepage and url should be my repo

vitorgalvao:
> Since the owner will not host the project , I will host it .
> So , the homepage and url should be my repo

No, we cannot accept that. That would be deceiving to our users, making it seem like this is official, when in fact it is not. We have no guarantee you won’t suddenly turn and include malware in your version.

Your version is quite below the acceptable threshold (50 stars), but given the circumstances I’d be willing to ignore that this time *if the other maintainers think it’s worth it*. My take is that we shouldn’t do it yet.

But even so, for us to accept your version, it has to:

1. Add an [`appcast`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).
2. Be an actual direct fork of the original project (right now it seems like you made a new repo and copied the code).
3. Be [renamed like a fork](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md#token-overlap-with-another-cask) (which it effectively is).

Also **please do not delete templates in place; they exist for a reason**.

PR #57236, opened 2019-01-09
meld.rb: remove binary

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/issues/57154

### DIFF
--- a/Casks/meld.rb
+++ b/Casks/meld.rb
@@ -9,7 +9,6 @@ cask 'meld' do
   homepage 'https://yousseb.github.io/meld/'
 
   app 'Meld.app'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
@reitermarkus Yeah, the removal suggestion was due to https://github.com/Homebrew/homebrew-cask/issues/57154#issuecomment-452789723 (we don’t link those binaries unless upstream suggests that as a course of action, and in this case the original addition didn’t mention it).

But since upstream seems to indeed suggest that, this should be a shim script (https://github.com/Homebrew/homebrew-cask/issues/57154#issuecomment-452921245).

PR #57270, opened 2019-01-10
Update screens to 4.6.6

core-code:

### DIFF
--- a/Casks/screens.rb
+++ b/Casks/screens.rb
@@ -1,6 +1,6 @@
 cask 'screens' do
   version '4.6.6'
-  sha256 'f97f17703c50bb8429a980bf9239ee9e49b35178f67db2efdf84c4b31fdfbceb'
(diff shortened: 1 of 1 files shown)


Casks/screens.rb
```diff
@@ -1,5 +1,5 @@
 cask 'screens' do
-  version '4.6.5'
+  version '4.6.6'
```
vitorgalvao: Not seeing this one in the appcast yet.

core-code:
good that you check, thanks.

it certainly has been in the sparkle appcast yesterday as evidenced here: 
http://macupdater.net/cask-repair/crsparklefeed-2019-01-09.txt

but since they removed it, i'd say hold off with merging until it re-appears in the appcast or website

Casks/screens.rb
core-code: it has re-appeared in the appcast. it has a different SHA now, i guess they just re-rolled

PR #57370, opened 2019-01-11
Delete copay.rb

suschizu:


suschizu:
I think this cask can be deleted - copay and bitpay have been migrated to the App Store since v 4.8.1 - the latest v is 5.4.2 (see https://github.com/bitpay/copay/releases)

PR #57386, opened 2019-01-12
Update musixmatch to 0.21.19

core-code:


vitorgalvao:
So, let’s find a way to make this work.

I understand @reitermarkus’ previous wariness for adding this, as it depends on an external, non-maintainer server to work, which users also have to trust. However, how I’m guessing this works is it accesses the URL given on the second part, and gets only the headers. In this case, it’d be the equivalent of doing:

```
curl --silent --head 'http://download-app.musixmatch.com/download/osx' | grep 'content-disposition' | cut -d= -f2
```

Which returns `Musixmatch-0.21.19.dmg`. is that right? And if so, am I also right in assuming the script performs the same operation every time we access that URL?

Being a simple and harmless operation, I’m fine with including this. What could change here to make your worries dissipate, @reitermarkus? Some options:

1. Have a copy of @core-code’s script, so if the server ever goes down, it’ll be a quick replace.
2. Have a copy of the script and run it from somewhere official, under Homebrew’s hosting.

core-code:
> as it depends on an external, non-maintainer server to work, which users also have to trust. 

i am actually not sure if i understand the wariness. users don't even use the appcast stanze, as i understand it its only used by contributors that want to check if casks are up-to-date.

lets assume the worst case, some very evil person gets control over my server and tries to do harm. i can see no possible way how he could do that since the appcast stanza is only checked by a few scripts to keep casks up-to-date, if they don't find something worthwhile delivered from the CGI, they'll just drop the response

as mentioned previously, i'd actually prefer if someone else could host the CGIs, but unless the load poses a problem i'd be happy to do it.
i've attached  both the  sources to both CGIs.

[check_url_filename.cgi.txt](https://github.com/Homebrew/homebrew-cask/files/2757238/check_url_filename.cgi.txt)
[check_url_redirect.cgi.txt](https://github.com/Homebrew/homebrew-cask/files/2757239/check_url_redirect.cgi.txt)

> Which returns Musixmatch-0.21.19.dmg. is that right? 

yes, one CGI script just gets the REDIRECT from the header, the other one gets the FILENAME. both are often the only sensible ways to get up-to-date vendor version information.

vitorgalvao:
> as mentioned previously, i'd actually prefer if someone else could host the CGIs

I’d like to make solving this my priority, although unfortunately right now that is difficult. What I’d like to do is get rid of all `appcasts` and start work on the previously talked idea of moving the auto-update-submissions to a separate project (not in that order).

I’d guess having that separate project could even help your own more straightforwardly.

core-code:
sounds good to me, i am certainly open to cooperating on that

vitorgalvao:
Merging when Travis passes (or tomorrow, if it takes too long) as I see it as a positive change with no consequence to users.

Pinging @Homebrew/cask, in case anyone wants to disagree and we can revert this. See [my reasoning above](https://github.com/Homebrew/homebrew-cask/pull/57386#issuecomment-454176947). https://github.com/Homebrew/homebrew-cask/pull/57816 is similar.

vitorgalvao:
Working now.

PR #57416, opened 2019-01-13
Add ReiKey 1.1.0

jacobbednarz:
Creates a new cask for [`ReiKey`](https://objective-see.com/products/reikey.html) at version 1.1.0.


vitorgalvao:
Thank you.

PR #57527, opened 2019-01-16
Update freefilesync to 10.8

core-code:


vitorgalvao:
> pkg source file not found: 'FreeFileSync_10.8_macOS.pkg'

core-code:
> pkg source file not found: 'FreeFileSync_10.8_macOS.pkg'

i think thats a false positive and just means travis actually downloaded an error HTML site instead of the archive. unless i am especially confused today.

vitorgalvao:
> travis actually downloaded an error HTML site instead of the archive.

Looks like that’s the case, but it doesn’t seem like a false positive. A local `audit --download` is downloading a PHP file for me. `user_agent: :fake` is no longer enough.

PR #57691, opened 2019-01-19
Update adobe-dng-converter 11.1 URL

danielhoherd:
The https URL is giving 404. I navigated through Adobe.com and found that they are now linking to an http protocol for the same URL.


vitorgalvao:
The Travis error wasn’t a network failure, it was `The URL http://download.adobe.com/pub/adobe/dng/mac/DNGConverter_11_1.dmg should use HTTPS rather than HTTP`.

vitorgalvao:
I can’t reproduce your 404. Either it was already fixed, or the problem is on your side.

danielhoherd:
It must have been an Adobe problem because I got 404's from two different locations using curl and wget. But yay for now working HTTPS and not having to make unnecessary changes.

PR #57717, opened 2019-01-20
Update keepingyouawake zap stanza

miccal:


Casks/keepingyouawake.rb
```diff
@@ -15,9 +15,13 @@
   uninstall login_item: 'KeepingYouAwake'
 
   zap trash: [
+               '~/Library/Application Scripts/info.marcel-dierkes.KeepingYouAwake',
```
reitermarkus: This already exists.

reitermarkus: Nevermind, they are different, I see it now. 😄 

```
~/Library/Application Support/info.marcel-dierkes.KeepingYouAwake
~/Library/Application Scripts/info.marcel-dierkes.KeepingYouAwake
```

reitermarkus:
Thanks!

PR #57768, opened 2019-01-22
Update parallels to 14.1.0-45387

ahousseini:


core-code:
this was previously reverted on purpose because version 14.1.0 does not seem to work with homebrew cask (see travis error).

PR #57776, opened 2019-01-22
Updated Microsoft Teams to 1.2.00.1761 and changed to now-default pkg installer

mylesagray:


reitermarkus:
This needs a rebase.

vitorgalvao:
Why did the version go *down*?

vitorgalvao (approved):

vitorgalvao:
> ==> Uninstalling packages:
com.microsoft.teams
==> Purging files for version 1.2.00.1761 of Cask microsoft-teams
Permission denied @ rb_sysopen - /Library/LaunchDaemons/com.microsoft.teams.TeamsUpdaterDaemon.plist

core-code:
in my opinion we should avoid using pkgs, if app downloads are available too. 
whats the official rule regarding this?

vitorgalvao:
> whats the official rule regarding this?

I think we don’t have anything written down.

> in my opinion we should avoid using pkgs, if app downloads are available too.

But that‘s my view as well.

But in this case, *are* app downloads still available? And will they continue to be? I only see the pkg in the download page, and if they’re phasing out app downloads, might as well change early.

core-code:
> But in this case, _are_ app downloads still available? 

yes, the latest official version is '1.2.00.1761' and the app is still available and the cask is still working

> And will they continue to be? 

well, we never know anything about the future ;)

> I only see the pkg in the download page, and if they’re phasing out app downloads, might as well change early.

why not perform the change when and if the app/dmg goes away?

mylesagray:
The pkg format seems to be the de-facto one going-forward see screenshot from in app with the .app version installed:

https://www.dropbox.com/s/t0rzao8p8fneycq/Screenshot%202019-02-15%2015.44.38.png?dl=0

The download link forces you to the pkg installer and it provides auto-updating built in.

I imagine it it a matter of time before it goes away, and it would be prudent to make the change early, resulting in no loss of functionality than waiting for the app link to break.

core-code:
> resulting in no loss of functionality than waiting for the app link to break.

thats not a problem, i can check the cask daily

> one going-forward see screenshot from in app with the .app version installed

ok thats a problem - if users having the app version installed get weird messages bugging them to change. 

so i guess changing over to the pkg is necessary after all. can you rebase or submit a new PR against the current version of the cask? thanks!

mylesagray:
Can do.

mylesagray:
@core-code - updated cask to latest version and includes original fix to pkg installer.

core-code:
thanks, looks good to me ;) no idea about the travis error, hope the team can figure it out.

core-code:
ping

vitorgalvao:
Ran it in a VM. Worked fine.

PR #57888, opened 2019-01-24
Use Pharo Launcher instead, remove old cask. Resolves pharo-project/p…

chrisfinazzo:
…haro-issues#2297

After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [ ] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

At the time of submission, Pharo Launcher is still @ v1.5.1 and has the same checksum.

PR #57889, opened 2019-01-24
Update and rename pharo-launcher.rb to pharolauncher.rb

vitorgalvao:
Did not follow the token reference.


vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/57888

PR #57915, opened 2019-01-25
Update skype to 8.34.0.78

core-code:


core-code:
ok i think its time to revert skype. while i have confirmed that 8.37 was the version offered for download from the homepage in pr #57572, it  seems not to be available from there anymore.

furthermore, the release notes ( https://support.skype.com/en/faq/FA34778/release-notes-for-skype-for-windows-mac-and-linux ) and the appcast also point to 8.34 as the latest one.

i think in the future we might only want to update if either the release notes or appcast mention the new version...

vitorgalvao (approved):

Issue #57954, opened 2019-01-26
Download local by Flywheel error


Kmutahar:
#### General troubleshooting steps

- [x] I have retried my command with `--force`.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.
- [x] I checked there are no [open issues](https://github.com/Homebrew/homebrew-cask/issues) for the same problem.
- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).

#### Description of issue

Error: Failure while executing; `unzip -qq /Users/kmutahar/Library/Caches/Homebrew/downloads/e3e70c1f2e15d08ccbf217b06a646602853abf9b5e11089867a1ab76e73dd038--local-by-flywheel-3-0-4-mac.zip -d /var/folders/76/rrz7wncx6l59l46brgfv7sz80000gn/T/d20190126-39830-qopm2q` exited with 9.

#### Command that failed

```
brew cask install local-by-flywheel
```

#### Output of command with `--force --verbose --debug`

```
==> Cask::Installer#install
==> Printing caveats
==> Cask::Installer#fetch
==> Satisfying dependencies
All Cask dependencies satisfied.
==> Downloading
==> Downloading https://local-by-flywheel-flywheel.netdna-ssl.com/releases/3-0-4/local-by-flywheel-3-0-4-mac.zip
/usr/bin/curl -q --show-error --user-agent Homebrew/1.9.3-22-g0670282\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 10.14.3\)\ curl/7.54.0 --location --silent --head https://local-by-flywheel-flywheel.netdna-ssl.com/releases/3-0-4/local-by-flywheel-3-0-4-mac.zip
Already downloaded: /Users/kmutahar/Library/Caches/Homebrew/downloads/e3e70c1f2e15d08ccbf217b06a646602853abf9b5e11089867a1ab76e73dd038--local-by-flywheel-3-0-4-mac.zip
[...]

PR #57958, opened 2019-01-26
Delete local-by-flywheel.rb

vitorgalvao:
Walled build.


claui:
The [download link appears to be public](https://local.getflywheel.com/community/t/local-by-flywheel-3-0-4/10622) to me.

@vitorgalvao I thought that by walled build, we mean there’s some sort of registerwall or paywall between the user and the download?

Or is there another meaning I haven’t thought of?

vitorgalvao:
@claui You’re correct. I hadn’t found that link. I went to the homepage and only found some other link that required a registration.

Then [we have an issue to fix on the cask](https://github.com/Homebrew/homebrew-cask/issues/57954).

PR #57978, opened 2019-01-26
Delete seashore.rb

suschizu:
I'm not 100% sure if this app will only be available in the App Store in future
but there is no download for the latest version and the developer refers to the App Store
see: https://github.com/robaho/seashore/releases
thanks to @ran-dall


core-code:
maybe things will stay fine, i opened an issue here: 

https://github.com/robaho/seashore/issues/38

robaho:
I am working on a fix so that I can continue to build the standalone dmg. I can easily do that now, with the 'check for updates' removed, since if installing via brew you don't need that anyway.

vitorgalvao:
DMG is back up. Let’s see if it sticks.

PR #58086, opened 2019-01-29
Update xnconvert to 1.78

tiiiecherle:

### DIFF
--- a/Casks/xnconvert.rb
+++ b/Casks/xnconvert.rb
@@ -1,6 +1,6 @@
 cask 'xnconvert' do
-  version :latest
-  sha256 :no_check
+  version '1.78'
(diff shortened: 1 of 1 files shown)


reitermarkus:
This doesn't have an appcast nor a versioned URL.

tiiiecherle:
ok, I like having version numbers nether the less for better overview and incoming automatic updates. I usually do not update "latest" casks very often.

If this is not in sync with any guidelines or technical problems feel free to close the pull request.

All in all I would like to have versions for as much casks as possible (perhaps with any kind of context addition to make clear that the cask does not have upcast or versions url) for the reasons stated above.

Perhaps there is a chance of implementing something like this. Thanks in advance. Have a great sunday ;)

reitermarkus:
> All in all I would like to have versions for as much casks as possible

That's our goal as well, but we can't have versions if neither the URL is versioned nor there is an appcast.

PR #58108, opened 2019-01-29
Update skype to 8.38.0.138

ahousseini:


core-code:
you have additional commits in here....

apart from that, about skype. i've previously said that we should be careful updating skype because we've had so many problems with beta versions. but 8.38.0.138  looks good to go from me now because 
1.) its the default download on the homepage now and
2.) the appcast serves this version at least SOME of the time. they do seem to do some staging here, only abut 1-2  of 10 requests get the new version

Issue #37123, opened 2017-07-27
Checksum mismatch: imazing 2.3.3


vickio:
[`imazing`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/imazing.rb) 2.3.3 install is failing with a checksum mismatch error:

```
Error: Checksum for Cask 'imazing' does not match.

Expected: ce31c7a8fd766e8587f8e3d236ca4db815a54c6fe1d5c31a588b3155c4c4a846
Actual:   ebc5f79e961aca34818d9bb0bf0565bbe333766ad5e5c68577551184d51d1879
File:     /Users/vickio/Library/Caches/Homebrew/Cask/imazing--2.3.3.dmg
```

commitay:
https://github.com/caskroom/homebrew-cask/pull/37125

PR #58112, opened 2019-01-29
Add json generation during travis ci.

cmbernard333:
Issue reference: [Add Casks to formulae.brew.sh](https://github.com/Homebrew/formulae.brew.sh/issues/13)

Generate the cask json using the new formulae.brew.sh script


.travis.yml
```diff
@@ -48,7 +48,12 @@ install:
     export TRAVIS_BUILD_DIR="${TAP_DIR}"
     builtin cd "${TRAVIS_BUILD_DIR}"
 
-script: brew cask ci
+script: 
+  - brew cask ci
+  - cd formulae.brew.sh && rake
+
+before_script:
```
MikeMcQuaid: Makes sense to put this above `script`, I think.

.travis.yml
```diff
@@ -48,7 +48,12 @@ install:
...
+  - travis_retry git clone https://$GITHUB_TOKEN@github.com/Homebrew/formulae.brew.sh
```
MikeMcQuaid: This should only be run for `master` branch builds.

.travis.yml
```diff
@@ -48,7 +48,12 @@ install:
     export TRAVIS_BUILD_DIR="${TAP_DIR}"
     builtin cd "${TRAVIS_BUILD_DIR}"
 
-script: brew cask ci
+script: 
+  - brew cask ci
+  - cd formulae.brew.sh && rake
```
MikeMcQuaid: It may be worth attempting to print more debugging information to figure out what this is doing.

.travis.yml
cmbernard333: I believe, correct me if I'm wrong here, this job only runs for master with this stanza ```branches:
  only:
    - master```

cmbernard333: Unless this section from `homebrew-core` should be copied in
```
deploy:
  local-dir: formulae.brew.sh
  repo: Homebrew/formulae.brew.sh
  target-branch: master
  verbose: true
  provider: pages
  skip-cleanup: true
  github-token: $GITHUB_TOKEN
  keep-history: true
  on:
    all_branches: true
```

MikeMcQuaid: homebrew-cask is using Travis CI for PRs too: https://github.com/Homebrew/homebrew-cask/pull/58204

Yeh, that section does need to be copied as otherwise the changes won't be set to formulae.brew.sh. Will also need to update homebrew-core too to make sure you don't deploy old versions of those (or separate the generation scripts)

.travis.yml
MikeMcQuaid: As suggested in the other thread: this should perhaps call `rake cask` instead which generates just the Casks.

cmbernard333: PR for scripts for JSON generation [#99](https://github.com/Homebrew/formulae.brew.sh/pull/99)

cmbernard333:
@MikeMcQuaid build ran through. I'm not seeing obvious errors in travis. It's updated to use the new `rake cask` but it says it failed. Any insight here?

cmbernard333:
@reitermarkus Do you know what might be causing the issues with this travis CI build?

MikeMcQuaid:
@cmbernard333 You'll need to figure out a way to get it to print more output from the `rake cask`. At the very least I'm guessing there's no tokens set here so it cannot be run.

cmbernard333:
@MikeMcQuaid I'm a little at a loss here. I adjusted the order of the `rake cask` and added the `--verbose` flag and nothing comes up. Is there anything I can do locally to validate this? The `rake cask` runs just fine when run by itself.

Here is the log for reference: https://api.travis-ci.org/v3/job/493373647/log.txt

MikeMcQuaid:
@cmbernard333 I'm currently working on moving the Homebrew/homebrew-core repository to use GitHub Actions for this publication (https://github.com/Homebrew/homebrew-core/blob/master/.github/main.workflow) so may want to hold off on this until that's done.

MikeMcQuaid:
@cmbernard333 now using GitHub Actions on Homebrew/homebrew-core to push to formulae.brew.sh: https://github.com/Homebrew/homebrew-core/actions/workflow-runs/MDEwOkNoZWNrU3VpdGU2NzUzMjE4MA== https://github.com/Homebrew/homebrew-core/blob/master/.github/main.workflow https://github.com/Homebrew/homebrew-core/blob/master/.github/main.workflow.sh https://github.com/Homebrew/formulae.brew.sh/commit/9b8a4f239ffce78ed9f8492726f7c1ecb36d28d7

The goal would be to do the same thing here. You should be able to mostly copy-paste from there and I can add the necessary tokens/secrets to the repo.

cmbernard333:
Sorry about not getting to this sooner. Been sick for the past week or so. I'll get around to looking at this soon.

cmbernard333:
@MikeMcQuaid as per [#37123](https://github.com/Homebrew/homebrew-core/pull/37123) Travis CI support is being taken. Should this also happen for homebrew-cask?

cmbernard333:
Create a new PR [#59521](https://github.com/Homebrew/homebrew-cask/pull/59521) to address this.

Issue #58046, opened 2019-01-28
Fix casks with `depends_on` that reference pre-Mavericks


vitorgalvao:
If you get an error of the type `Error: Cask 'hex-fiend-beta' definition is invalid: invalid 'depends_on macos' value: ":lion"`, where `hex-fiend-beta` can be any cask name, and `:lion` any macOS release name, run the following command:

```bash
/usr/bin/find "$(brew --prefix)/Caskroom/"*'/.metadata' -type f -name '*.rb' -print0 | /usr/bin/xargs -0 /usr/bin/perl -i -pe 's/depends_on macos: \[.*?\]//gsm;s/depends_on macos: .*//g'
```

This will remove all `depends_on macos` references of *installed* casks (where it doesn’t matter anymore anyway).

We should deal with these cases better in the future.

glhrmv:
I was trying to uninstall the `keka` cask and I couldn't because of this error message. Thank you.

ran-dall:
@vitorgalvao I remember that we had a conversation with @core-code about trying 'us' trying to support as many releases of `macOS` as possible. I now see what you were saying about the high maintenance overhead...

Maybe we should have it a **harder** requirement of **only supporting the last two releases of `macOS.`** _I know it doesn't fix the issue, but it would mitigate it a bit by not having as many tags._ 🏷

We should consider that we're talking [~31% of users; which are on something different from the last two releases of `macOS`](http://gs.statcounter.com/macos-version-market-share/desktop/worldwide).

vitorgalvao:
@rand-all This specific lack of support for pre-Maverick is now enforced in the core, with the release of Homebrew 2.0.0 (today!).

ran-dall:
@vitorgalvao Yea I saw about an hour after I wrote that...haha! 👍

reitermarkus:
I think we can safely remove `depends_on: '>= :mavericks'`.

ran-dall:
@reitermarkus 🧞‍♂️It shall be done, then...

brianmorton:
https://github.com/Homebrew/homebrew-cask/pull/58460

brianmorton:
https://github.com/Homebrew/homebrew-cask/pull/58507

felker:
By the way, if you have edited your `PATH` environment variable so that GNU `sed` is the default, @vitorgalvao's command won't work as written:
```
sed: can't read /depends_on macos:/d: No such file or directory
```

[...]

PR #58460, opened 2019-02-05
remove cask lines referencing mavericks

brianbrownton:
should help with https://github.com/Homebrew/homebrew-cask/issues/58046

### DIFF
--- a/Casks/0-ad.rb
+++ b/Casks/0-ad.rb
@@ -8,7 +8,5 @@ cask '0-ad' do
   name '0 A.D.'
   homepage 'https://play0ad.com/'
 
-  depends_on macos: '>= :mavericks'
-
   app '0 A.D.app'
 end
--- a/Casks/abstract.rb
+++ b/Casks/abstract.rb
@@ -7,7 +7,6 @@ cask 'abstract' do
   homepage 'https://www.goabstract.com/'
 
   auto_updates true
-  depends_on macos: '>= :mavericks'
 
   app 'Abstract.app'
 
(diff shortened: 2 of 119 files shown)


reitermarkus:
The ones with `MacOS.version <= :mavericks` should still stay. Only `depends_on macos: '>= :mavericks'` should be removed, since that's implied now that Homebrew only runs on Mavericks and later.

brianmorton:
>The ones with `MacOS.version <= :mavericks` should still stay. Only `depends_on macos: '>= :mavericks'` should be removed, since that's implied now that Homebrew only runs on Mavericks and later.

While it is possible I've made a mistake, I don't believe I've done anything to the contrary of the above.

reitermarkus:
> are you saying these should instead be changed to `==` instead of `<=`?

I'm saying they should not be deleted. I think we should leave them as `<=`.

vitorgalvao:
> Forgive my ignorance, but if brew is Mavericks+ why do we need `<` (less than) Mavericks?

We don’t *need* it, but it’s nice to have. They both mean the same right now, so having the one that covers more cases is better.

For example, let’s say we decide to revert the decision and add support for pre-Mavericks, or that a project such as [Tigerbrew](https://github.com/mistydemeo/tigerbrew) uses Casks. Not having to redefine those `==` is beneficial.

And from a contributors point of view, I can see people that don’t know that we only support from Mavericks on trying to include older versions, `<=` “eases the mind” into “that’s already done”.

vitorgalvao:
Also, by leaving them as `<=`, this PR will require less changes, as it just means not editing those casks where it was deleted.

brianmorton:
I'll drop the PR and just resub one that removes the depends on lines.

PR #58542, opened 2019-02-07
Update parallels to 14.1.1-45476

core-code:


vitorgalvao:
Isn’t this the version that gave errors even when tried manually?

core-code:
its the next one released yesterday or so

but the error is the same

should i close this?

vitorgalvao:
If we can’t get this to work, might as well keep the version we have for now. Maybe a Parallels user will eventually find a solution. The cask is `#249` over the last 90 days, according to analytics.

But it reminds me. @reitermarkus You’re a Parallels user. Have you not encountered this issue?

reitermarkus:
> Have you not encountered this issue?

No, since I'm still using Parallels 13.

PR #58687, opened 2019-02-10
Update vienna to 3.5.4

barijaona:


core-code:
` - The URL http://www.vienna-rss.com/ should use HTTPS rather than HTTP
`

barijaona:
> - The URL http://www.vienna-rss.com/ should use HTTPS rather than HTTP
Done

PR #58706, opened 2019-02-10
Update SoapUI: clean install4j args and comments, and uninstall stanza

sbibauw:
Following @vitorgalvao 's suggestion, I updated this formula following discussion and suggestions on #58609

Arguments of the `install4j` install script have been limited to essential only, commented inline, and including `#{appdir}` for installation directory. 
Uninstall stanza was changed from `delete` to `trash` (and now refers to `#{appdir}`).


vitorgalvao:
Thank you for doing it so fast, @sbibauw!

PR #58731, opened 2019-02-11
Update musixmatch to 0.21.26

core-code:


vitorgalvao:
I’m thinking about removing this cask. Every time it gets updated it’s a bother to download and verify, and according to the analytics, it’s not insanely popular.

core-code:
thats fine with me.

i tried to get the SHA for this all day long, because it downloads just with 3kb and stops every hour. i hoped travis could do that job better, but no.

i think 'manager' and 'kate' are also much more trouble than they are worth - no users but an update each day. it seems 'musixmatch' is 40 times more popular than both of them combined.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/58737

PR #58963, opened 2019-02-15
Update nvidia-geforce-now to 1.17.2.28

core-code:


core-code:
i am not sure why the SHA changed right after we added this. lets keep this under observation, if things don't work well we can revert back to un-versioned.

PR #58985, opened 2019-02-15
Delete easyfig.rb

vitorgalvao:
So few installs, it doesn’t even register in the analytics. Also, open source, CLI-only, and fewer than 50 stars.

Pinging @sjackman since you added this way back when, and may still be using it.


sjackman:
I'm okay with removing it. Thanks for checking in! And for reminding me that this tool exists. I had forgotten about it!

sjackman (approved):

PR #12493, opened 2015-07-09
Add Pokemon Trading Card Game Online Latest

rhysm94:

PR #59115, opened 2019-02-18
Add latest Pokémon Trading Card Game Online

jasonmp85:
My son and I play this and I've automated my home Mac setup, so I need a cask for this. I found Homebrew/homebrew-cask#12493, but it looks as though the original submitter didn't follow up. Fortunately, I independently discovered and implemented all the feedback on my own (I suppose it could have saved me time to look for the PR first).

The developers post release notes to a [specific forum topic](http://forums.pokemontcg.com/forum/108-patch-notes-known-issues); unfortunately, there's no straightforward way to access this over RSS, or I'd mark it as a "page that changes every release". The reply counts and last updated dates will change between releases, so without some sort of structured post list it's unreliable.

# Requested Feedback

  - [x] _Remove all the comments, except the necessary one above `url`_
  - [x] _Please see how other casks use `installer` and `preflight`. You’ll see your style is off_
  - [x] _Forget `container` and `installer` […] instead extract everything in `preflight`, then use `app` to move the file_
  - [x] _And in doing that, remove `uninstall`_

### DIFF
--- a/Casks/pokemon-trading-card-game-online.rb
+++ b/Casks/pokemon-trading-card-game-online.rb
@@ -1,0 +1,19 @@
+cask 'pokemon-trading-card-game-online' do
+  version :latest
+  sha256 :no_check
+
+  # tcgo-installer.s3.amazonaws.com was verified as official when first introduced to the cask
+  url 'https://tcgo-installer.s3.amazonaws.com/PokemonInstaller_Mac.dmg'
+  name 'Pokemon Trading Card Game Online'
+  homepage 'https://www.pokemon.com/us/pokemon-tcg/play-online/'
+
+  app 'Pokemon Trading Card Game Online.app'
+
+  preflight do
+    FileUtils.cd staged_path do
+      system_command '/usr/bin/tar', args: ['-xf', 'The Pokémon Online Trading Card Game Installer.app/Contents/Resources/packages/pokemon.tar', 'pokemon.tar.gz']
+      system_command '/usr/bin/tar', args: ['-xzf', 'pokemon.tar.gz']
+      FileUtils.rm_rf 'pokemon.tar.gz'
+    end
+  end
+end


vitorgalvao:
Welcome. This is a nice start, but needs several fixes.

1. Remove all the comments, except the necessary one above `url`.
2. Please see how other casks use `installer` and `preflight`. You’ll see your style is off.
3. Forget `container` and `installer`. The `install.sh` script does nothing more than extract the `pokemon.tar.gz` file next to it and move it to `Applications`. So instead extract everything in `preflight`, then use `app` to move the file.
4. And in doing that, remove `uninstall`.

vitorgalvao (changes requested):
Welcome. This is a nice start, but needs several fixes.

1. Remove all the comments, except the necessary one above `url`.
2. Please see how other casks use `installer` and `preflight`. You’ll see your style is off.
3. Forget `container` and `installer`. The `install.sh` script does nothing more than extract the `pokemon.tar.gz` file next to it and move it to `Applications`. So instead extract everything in `preflight`, then use `app` to move the file.
4. And in doing that, remove `uninstall`.

jasonmp85:
> Please see how other casks use `installer` and `preflight`. You’ll see your style is off.

I said it once to Homebrew proper and will say it again here: this is probably something that a properly crafted `rubocop` rule would catch and save everyone (you too) some time…

> Forget `container` and `installer`. The `install.sh` script does nothing more than extract the `pokemon.tar.gz` file next to it and move it to `Applications`. So instead extract everything in `preflight`, then use `app` to move the file.

I thought about this. It differs in a few ways…

  * Things installed by `cask` appear to be `chown`ed to `$USER:staff`, whereas this script explicitly `chown`s to `$USER:admin`. I'm almost certain there's no meaningful reason for this, but still, it's what the installer would have done
  * This seems more future-proof, as subsequent changes may add steps to the script, which would then be missed by a simple extraction seeking to emulate the script as-is
  * Given the script's assumptions about what `$PWD` is set to, I'm sort of concerned the main application may make equally bad assumptions about the path to the install. Is getting rid of enclosing folders (e.g. an app in `/Applications/Foo/Foo.app` being changed in its cask to `/Applications/Foo.app`) something this project does? One of the style guides had a warning about making target changes for "aesthetic reasons", which this seemed to fall into.

I'll address your feedback, but just wanted to provide my thoughts on the approach, as I did consider `app` for quite a while before deciding delegating the install logic to the included script felt more future-proof.

vitorgalvao:
> this is probably something that a properly crafted `rubocop` rule would catch and save everyone (you too) some time…

I’m surprised our rules didn’t catch your cases, to be honest, as they were so far off. That also means something so far off style is *rare*. Incredibly so. We would take a PR to improve the cop’s rule, though.

> Things installed by `cask` appear to be `chown`ed to `$USER:staff`, whereas this script explicitly `chown`s to `$USER:admin`.

[We have methods for that](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/flight.md#flight-mini-dsl).

> I'm almost certain there's no meaningful reason for this, but still, it's what the installer would have done

There doesn’t seem to be. The whole script is badly written, using `eval` and escaping with no apparent reason.

> This seems more future-proof, as subsequent changes may add steps to the script, which would then be missed by a simple extraction seeking to emulate the script as-is

Just as well the script itself may be gone or renamed, or put in a different location. Future-proof in this case is not `installer script:` but `installer manual:`. But since we’re going for a better experience, go with my recommendation first and we’ll scale back to `installer manual:` if need be.

> Given the script's assumptions about what `$PWD` is set to, I'm sort of concerned the main application may make equally bad assumptions about the path to the install.

Even if it did, we could also fix that. But it seems it does not. I unpacked it to the Desktop and ran it, and it opened without a problem.

> Is getting rid of enclosing folders (e.g. an app in `/Applications/Foo/Foo.app` being changed in its cask to `/Applications/Foo.app`) something this project does? One of the style guides had a warning about making target changes for "aesthetic reasons", which this seemed to fall into.

That rule is about renaming of the app bundle. But you can also reproduce the structure the app expects.

jasonmp85:
Alright, I think this hits all your feedback.

Quick question: I used `/usr/bin/tar` similarly to what was in use in `importio` and `qz-tray`…

https://github.com/Homebrew/homebrew-cask/blob/4184d50d2d7fa4afb62e6c7eb6761ae5ac417c88/Casks/importio.rb#L12-L16

https://github.com/Homebrew/homebrew-cask/blob/4184d50d2d7fa4afb62e6c7eb6761ae5ac417c88/Casks/qz-tray.rb#L15-L23

While at least similar to other casks, the direct use of `tar` seemed rare enough I wanted to double-check that this was OK (perhaps these casks were written before styles were set in stone).

In addition, what's the convention around the staging directory? Do I need to remove the intermediate files here?

jasonmp85:
OK, so I realized I could just extract the `tar.gz` from the `tar` and then keeping the stage directory tidy is a single `rm_rf` away. Ready for review.

vitorgalvao:
Looks good. Thank you.

PR #59152, opened 2019-02-19
Update onionshare.rb to v2.0

ThatLurker:
In this version they decided to not but the .0 on the file name so needed additional changes.

we could wait for 2.1 where we can revert back to the old version where we don't need to specify filename separately 

I was not able to run the test commands for some reason.


pahakalle:
Im not 100% sure this works so if someone could test it before merging

PR #59153, opened 2019-02-19
Update onionshare.rb to v2.0

ThatLurker:
In this version they decided to not but the .0 on the file name so needed additional changes.

For some reason i was unable to run the tests


vitorgalvao:
> For some reason i was unable to run the tests

Still, next time please leave the template in place, even if unfilled.

Also, if you share the exact errors, we may be able to help.

vitorgalvao (approved):

PR #59214, opened 2019-02-20
Add -lax argument to list_ids_in_pkg script

ytbmulder:
**Symptoms:**
While writing `uninstall` stanza's for `pkg`s from Apple, the `pkg` ID(s) are filtered out using `egrep`.
This requires manual expansion of the `pkg` and digging through `bom` files for the ID(s).

For example:
```
$ ./list_ids_in_pkg /Volumes/Brother\ Printer\ Drivers/BrotherPrinterDrivers.pkg
``` 
returns no `pkg` IDs at all, while two are present.

**Proposal:**
The `list_ids_in_pkg` script already does the digging through the expanded `pkg`, but applies a filter afterwards. This pull-request simply adds an argument (`-lax`) to turn the filtering of Apple and Sparkle IDs off (similarly to the `list_apps_in_pkg` script).
It also shares the same structure now as the `list_apps_in_pkg` script with stricter checking on the arguments provided to the script.

Using the updated script, the previous command returns the two expected IDs:
```
$ ./list_ids_in_pkg -lax /Volumes/Brother\ Printer\ Drivers/BrotherPrinterDrivers.pkg
com.apple.pkg.BrotherPrinterDrivers
com.apple.pkg.BrotherPrinterDriversPreInstall
```

At the same time, it does not change the already existing behaviour of the script:
```
$ ./list_ids_in_pkg ~/Downloads/WebDriver-378.05.05.25f13.pkg
com.nvidia.nvprefpane
com.nvidia.web-driver
```
```
$ ./list_ids_in_pkg -lax ~/Downloads/WebDriver-378.05.05.25f13.pkg
com.nvidia.nvprefpane
com.nvidia.web-driver
```

### DIFF
--- a/developer/bin/list_ids_in_pkg
+++ b/developer/bin/list_ids_in_pkg
@@ -15,6 +15,8 @@ shopt -s nocasematch  # case-insensitive regular expressions
 ### global variables
 ###
 
+opt_lax=''
+opt_pkg=''
 pkgdir=''
 
 # prefer GNU xargs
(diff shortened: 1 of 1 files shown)


developer/bin/list_ids_in_pkg
```diff
@@ -74,42 +131,14 @@ _list_ids_in_pkg () {
       bundle_id_source_2;
     } |                     \
     merge_sources |         \
-    clean_sources |         \
+    ( [[ ! "$opt_lax" ]] && clean_sources || cat ) | \
```
vitorgalvao: What’s this line doing?

reitermarkus:
For future reference: I'd rather the existing Bash scripts be ported to Ruby instead of extended further.

vitorgalvao:
> For future reference: I'd rather the existing Bash scripts be ported to Ruby instead of extended further.

For what it’s worth, I’d rather clean up the existing scripts than change their language. They require system tools that we’ll have to call via `system` or `Open3` in ruby, and if they’re cleaned up they can stay decent portable scripts.

Right now they work but are ugly, and a cleanup could do wonders for their readability.

reitermarkus:
> I’d rather clean up the existing scripts than change their language.

Why not both? 😄

Sure, a clean Bash script is better than an ugly one, but a clean Ruby script is even nicer, especially if it involves parsing options.

developer/bin/list_ids_in_pkg
ytbmulder: It tests if `opt_lax` is defined. If it is not, the `clean_sources` function will be called. Otherwise, the stream will be piped to `cat`, which in this case does not alter the intermediate result.

ytbmulder:
> Right now they work but are ugly, and a cleanup could do wonders for their readability.

I agree that some scripts are tough to read and improving their readability would definitely help.

> Sure, a clean Bash script is better than an ugly one, but a clean Ruby script is even nicer, especially if it involves parsing options.

I wasn't aware that porting to `ruby` was preferred. Therefore I continued with the `bash` script. 
Is this milestone tracked somewhere?

I am comfortable with `bash` and not so much with `ruby`. However, I am more than happy to start porting the `bash` scripts to `ruby` if this is something you would like to have.

vitorgalvao:
> Why not both?

Because for small scripts that rely on system tools for their main task, bash is more future proof maintainability wise (we barely need to touch these as it is). It’s a testament to how good Homebrew’s (including Cask) DSL is that so many non-rubyists are maintainers.

But I also think it depends on the script. I’ve made both ruby and bash scripts for that directory, picking the language depending on the task (will I need to parse JSON, will I need system tools), so it does depend on the situation.

This one in particular has some obtuse lines that could maybe be cleaner if the script was in ruby, yes.

ytbmulder:
> This one in particular has some obtuse lines that could maybe be cleaner if the script was in ruby, yes.

I do agree that the language chosen depends on the task at hand. 

For the time being, I would prefer this PR to be merged. The functionality is then available in `master`.
Porting the scripts to `ruby`, if desired, would then be the next step.

reitermarkus:
> For the time being, I would prefer this PR to be merged.

Sure, that's why I said “For future reference”.

ytbmulder:
> Sure, that's why I said “For future reference”.

I missed that remark, sorry @reitermarkus :)

reitermarkus:
@ytbmulder, no worries, your work is appreciated in any case. 😉

PR #59274, opened 2019-02-21
Revert "Update metaz to 1.0.beta-45"

suschizu:
Reverts Homebrew/homebrew-cask#59246
@ran-dall … sorry 1.0.beta-45 is a pre release


reitermarkus:
What is the difference if all of them are called beta?

core-code:
all of them may be called beta but
• beta 43 is available from the homepage and marked as 'latest release' in github
• beta 45 is not available from the homepage and marked as 'pre-release' in github

ran-dall:
@suschizu Good catch. I don't think my system understood this one correctly. Good looking 👀 out!

PR #59292, opened 2019-02-22
Add action to sync templates.

reitermarkus:
Replace `sync_cask_tap_templates` script with GitHub Actions.


.github/actions/sync/sync.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -euo pipefail
+
+ls -all
```
vitorgalvao:
```suggestion
ls -al
```

The last `l` is a repetition.

vitorgalvao:
Might be worth it to add a way to sync labels as well. Right now they are kind of a mess.

PR #59330, opened 2019-02-22
Core code appcasts

vitorgalvao:
@core-code It just occurred to me we could simplify some of the `appcast`s.

### DIFF
--- a/Casks/blitz.rb
+++ b/Casks/blitz.rb
@@ -3,7 +3,7 @@ cask 'blitz' do
   sha256 'a255fb27f817cb910487eb1efc6fdad0fdc085c90c887933a0598786c5c21096'
 
   url 'https://dl.blitz.gg/download/mac'
-  appcast 'https://www.corecode.io/cgi-bin/check_urls/check_url_filename.cgi?url=https://dl.blitz.gg/download/mac'
+  appcast "https://www.corecode.io/cgi-bin/check_urls/check_url_filename.cgi?url=#{url}"
   name 'Blitz'
   homepage 'https://blitz.gg/'
 
--- a/Casks/focus.rb
+++ b/Casks/focus.rb
@@ -3,7 +3,7 @@ cask 'focus' do
   sha256 '54753c79c8cc92c1d32faf035fe3ad97e439367b60929926995834f952df9fe6'
 
   url 'https://heyfocus.com/focus.zip'
-  appcast 'https://www.corecode.io/cgi-bin/check_urls/check_url_redirect.cgi?url=https://heyfocus.com/focus.zip'
+  appcast "https://www.corecode.io/cgi-bin/check_urls/check_url_redirect.cgi?url=#{url}"
   name 'Focus'
   homepage 'https://heyfocus.com/'
 
--- a/Casks/marvin.rb
+++ b/Casks/marvin.rb
@@ -4,7 +4,7 @@ cask 'marvin' do
 
   # amazingmarvin.s3-website-us-east-1.amazonaws.com was verified as official when first introduced to the cask
   url 'http://amazingmarvin.s3-website-us-east-1.amazonaws.com/Marvin.dmg'
-  appcast 'https://www.corecode.io/cgi-bin/check_urls/check_url_redirect.cgi?url=http://amazingmarvin.s3-website-us-east-1.amazonaws.com/Marvin.dmg'
+  appcast "https://www.corecode.io/cgi-bin/check_urls/check_url_redirect.cgi?url=#{url}"
   name 'Amazing Marvin'
   homepage 'https://www.amazingmarvin.com/'
 


core-code:
hey vitor,
simplification sounds always good. i‘ll need to check how hard that will be to support in our system that checks the appcasts daily - don‘t expect troubles though.

vitorgalvao:
@core-code Tip: `_stanza` is an internal command that will give you the value of any stanza with the interpolation resolved. If you always use that, you shouldn’t have any trouble. Here’s what I mean:

```bash
$ brew cask cat uninstallpkg | grep url
url "https://www.corecode.io/downloads/uninstallpkg_#{version}.zip" # Interpolation exists in the `url` stanza

$ brew cask _stanza url uninstallpkg
https://www.corecode.io/downloads/uninstallpkg_1.1.5.zip # But `_stanza` gives us the usable result
```

core-code:
thanks :) but thats not helping as the appcast check system is all-native code, so i‘ll have to implement support for replacing this in the appcast myself.

even executing a single small ruby script takes 200ms, so i try to avoid external dependencies when checking 3000+ casks daily.

vitorgalvao:
> so i‘ll have to implement support for replacing this in the appcast myself

Would you rather this was reverted? The change was supposed to make updates easier, but if it’s making them harder, I’m fine with reverting.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/59334

core-code:
thanks for asking :) i‘ll have a look tomorrow if i can implement support for this. i don‘t epect it to be difficult but if it is we can think about options like reverting.

core-code:
btw i was thinking about doing a github repository with the sources to the CGI scripts so that other people could reproduce that. but maybe they should be part of HBC?

btw2 whats up with @BrewTestBot?

vitorgalvao:
> i don‘t expect it to be difficult but if it is we can think about options like reverting.

I’ve marked the revert PR as `do not merge` until you have time to look into it.

> btw i was thinking about doing a github repository with the sources to the CGI scripts so that other people could reproduce that. but maybe they should be part of HBC?

You mean all of the scripts you use to gather the latest versions? That sounds like a good idea. I’m open to having them as part of HBC.

> btw2 whats up with @BrewTestBot?

We’re testing it as a way to merge simple PRs faster.

core-code:
> I’ve marked the revert PR as `do not merge` until you have time to look into it.

thanks, but you can close it. implementing support for #{url} in the appcast was trivial, all should be good now. 

> You mean all of the scripts you use to gather the latest versions?
> That sounds like a good idea. I’m open to having them as part of HBC.

well, actually i was talking about the sources to 'check_url_filename.cgi' and 'check_url_redirect.cgi'
i've shared them previously, but have made changes since. i think having them either in HBC or in a separate github repository readily available would be nice.

sharing the sources to the tool that generates the output here ( http://macupdater.net/cask-repair/ ) would be more difficult. on one hand its part of a 3000loc ObjC tool that does a lot of things. one the other hand, the DB check requires access to our MacUpdater DB which is only available internally here. but if there is demand to have the github/sparklefeed/appcast/htmlappcast checks open-sourced, i'd be willing to do that.

> We’re testing it as a way to merge simple PRs faster.

sounds good in theory, and i'm all in favour! however, since the bot has already merged some beta versions which we had to revert, what is the approach to prevent that in the future?

core-code:
> however, since the bot has already merged some beta versions which we had to revert, what is the approach to prevent that in the future?

constructive idea: in most cases, a stable version will be contained in the appcast. for many casks it would be as simple as looking up if the version stanza (or, #{version.before_comma} is contained in the appcast.

vitorgalvao:
> i think having them either in HBC or in a separate github repository readily available would be nice.

Agreed.

As to your internal tool, I understand the point about keeping that internal. Makes sense.

> since the bot has already merged some beta versions which we had to revert, what is the approach to prevent that in the future?

We don’t have one, but are open to ideas. [That was my concern as well](https://github.com/Homebrew/brew/pull/5740#issuecomment-464049825).

> for many casks it would be as simple as looking up if the version stanza (or, #{version.before_comma} is contained in the appcast.

Not sure if the percentage where that is true outweighs the cases where it is not. For example, Github appcasts contain both versions.

core-code:
> Agreed.

ok let me know if you want them in HBC, in which case i'll just send you the current sources.
else i'll create a repository for this.

> As to your internal tool, I understand the point about keeping that internal. Makes sense.

as mentioned i'd be willing to extract the bits that are useful to HBC and provide the sources.
but since this would require some work, i'd rather only do it when and if there is any demand.
feel free to ping me about this at any time.

> Not sure if the percentage where that is true outweighs the cases where it is not.
> For example, Github appcasts contain both versions.

valid concern. however if we blacklist github (or make it more intelligent so that only 'latest release's from github will be picked up), i think the percentage is pretty high. and some protection is obviously better than none.

however, i had a look today to see whether it is even true that (HTML) appcasts always contain the #{version.before_comma.before_hyphen}. i found about 70 exceptions, not finished inspecting them yet. a lot just did not have the current version in HBC. some other appcasts were just plain invalid. in other cases the version number is only contained in some transformed form (without trailing zero, without points, etc). so, as expected, that idea would not work in all cases at least if implemented trivially.

vitorgalvao:
> ok let me know if you want them in HBC, in which case i'll just send you the current sources.

Maybe as an addition to [`developer/bin`](https://github.com/Homebrew/homebrew-cask/tree/master/developer/bin)?

core-code:
sounds good, i'll send a PR time permitting

core-code:
i've sent a PR, didn't bring it over to me to mess up the developer/bin directory with new files that are a bit different than the rest with the extensions and all

how about adding a developer/cgi directory? https://github.com/Homebrew/homebrew-cask/pull/59688

PR #59334, opened 2019-02-22
Revert corecode appcast interpolation

vitorgalvao:
This makes updates harder for some systems (https://github.com/Homebrew/homebrew-cask/pull/59330#issuecomment-466545542).

### DIFF
--- a/Casks/marvin.rb
+++ b/Casks/marvin.rb
@@ -4,7 +4,7 @@ cask 'marvin' do
 
   # amazingmarvin.s3-website-us-east-1.amazonaws.com was verified as official when first introduced to the cask
   url 'http://amazingmarvin.s3-website-us-east-1.amazonaws.com/Marvin.dmg'
-  appcast "https://www.corecode.io/cgi-bin/check_urls/check_url_redirect.cgi?url=#{url}"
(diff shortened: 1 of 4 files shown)


reitermarkus (approved):
I think it makes more sense to have the unversioned URL in the appcast, and then actually use the URL it redirects to for the `url`.

vitorgalvao:
@reitermarkus That does make more sense. Changed, though I’m not finding the correct versioned URL for `ocenaudio`.

Pinging @core-code since this will end up being merged after all.

Casks/marvin.rb
```diff
@@ -3,8 +3,8 @@
   sha256 'fce9293100fceacabc700e74cac4c6627074f796acf9b716962f16c0ac4b7ff6'
 
   # amazingmarvin.s3-website-us-east-1.amazonaws.com was verified as official when first introduced to the cask
```
reitermarkus: Comment needs to be updated.

core-code:
> @reitermarkus That does make more sense. Changed, though I’m not finding the correct versioned URL for `ocenaudio`.

unable to find it either.

> Pinging @core-code since this will end up being merged after all.


thanks

PR #59342, opened 2019-02-23
Remove Twig v0.0.5

lukakerr:
This just removes my application Twig. I have decided to create my own tap and release updates there.


vitorgalvao:
That is fine. But I ask that in the future, if you have any intentions of doing so, please consider it before submitting the cask. This was added just a week ago and already has almost 100 downloads. By removing, you’re breaking the installation for those people, and I know from experience that if there are going to be complaints, they’re likely to come to us, not you. In addition, this took maintainer time to review, correct, and accept, which could’ve been spent on other casks.

I don’t for a moment think that was your intention, but I do ask you to consider it for future submissions.

PR #59358, opened 2019-02-23
Update busycontacts to 1.3.2

core-code:


core-code:
the new appcast is the one the app uses for update checks. sadly it also contains 'news' , so using the CGI ( http://www.corecode.io/cgi-bin/check_urls/check_url_redirect.cgi?url=https://www.busymac.com/download/BusyContacts.zip ) would probably have even less noise but i guess we should stick to official appcasts where available

PR #59377, opened 2019-02-23
Update autopano-giga to 4.4.2

core-code:

### DIFF
--- a/Casks/autopano-giga.rb
+++ b/Casks/autopano-giga.rb
@@ -1,5 +1,5 @@
 cask 'autopano-giga' do
-  version '4.4.2.400,2018-09-10''
+  version '4.4.2.400,2018-09-10'
   sha256 '0d5808f3904dae5937411ba7efdb275eaef2dea957c89b869c6c637aa29b824a'
 
   url "https://cdn-download.kolor.com/apg/#{version.before_comma}_#{version.after_comma}/AutopanoGiga_Mac_#{version.major_minor_patch.no_dots}_#{version.after_comma}.dmg"
--- a/Casks/autopano-giga.rb
+++ b/Casks/autopano-giga.rb
@@ -1,8 +1,9 @@
 cask 'autopano-giga' do
-  version '4.4.2.400,2018-09-10'
+  version '4.4.2'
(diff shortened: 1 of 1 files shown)


core-code:
i think it's better to have the simpler URL in there akin to autopano-pro, but if anyone disagrees i can change that back.

reitermarkus:
I think it's better to use the versioned URL if there is one, otherwise the `sha256` will break if the version behind the unversioned URL changes.

core-code:
ok, i'll change this back as well as convert autopano-pro to also use this.

Casks/autopano-giga.rb
```diff
@@ -3,6 +3,7 @@
   sha256 '0d5808f3904dae5937411ba7efdb275eaef2dea957c89b869c6c637aa29b824a'
 
   url "https://cdn-download.kolor.com/apg/#{version.before_comma}_#{version.after_comma}/AutopanoGiga_Mac_#{version.major_minor_patch.no_dots}_#{version.after_comma}.dmg"
+  appcast 'https://macupdater.net/cgi-bin/check_urls/check_url_redirect.cgi?url=http://download.kolor.com/apg/stable/mac'
```
reitermarkus: Why not `www.corecode.io` on this one?

core-code: when using the corecode server, i get an error message about an SSLv3 failure instead of the actual, useful output.

it seems the reason is that it still runs python 2.6 which is incompatible with some SSL requirements set forth by some servers like the autopano ojne.

copying the script unmodified to the macupdater server which has python 2.7 seems to fix the issue (but only when passing the URL parameter with  http instead of https).

i am not happy about that workaround but i am not sure if i should bug our hosting-provider to upgrade the python installation just for this - especially since other, mission critical, python stuff also runs on the corecode server.

vitorgalvao: @core-code Would it be the same to you if we changed all the other `corecode.io` appcasts to `macupdater.net`, for consistency?

core-code: yes its all the same to me, feel free to change it over. i don't expect any problems from changing over, did test a few casks and everything continues to work if we change the domain.

PR #59453, opened 2019-02-24
Update powerword to 1.1.1,1.0.1

core-code:

### DIFF
--- a/Casks/powerword.rb
+++ b/Casks/powerword.rb
@@ -5,7 +5,7 @@ cask 'powerword' do
   url "https://download.iciba.com/mac/mac#{version.after_comma}/PowerWord.dmg"
   name 'PowerWord'
   name '金山词霸'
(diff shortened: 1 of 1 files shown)


Casks/powerword.rb
```diff
@@ -1,11 +1,11 @@
 cask 'powerword' do
-  version '1.0.1'
-  sha256 'bacf37532f98879761c098cfa93777ef0250ec9ae8e63392f87c5e34cd74c363'
+  version '1.1.1,1.0.1'
```
vitorgalvao: Why the `1.1.1` addition, if the URL stayed at `1.0.1`?

core-code: hmm because it IS version 1.1.1 now

core-code: but i can remove it, though i find labelling v1.1.1 as 1.0.1 confusing

PR #59521, opened 2019-02-26
Use github actions to generate cask json

cmbernard333:
Issue reference: [Add Casks to formulae.brew.sh] (https://github.com/Homebrew/formulae.brew.sh/issues/13).

Generate cask json using github actions similar to [#37123](https://github.com/Homebrew/homebrew-core/pull/37123) for homebrew-core.


reitermarkus:
Should this include the other cask taps?

.github/main.workflow
```diff
@@ -60,3 +60,14 @@ action "sync_labels" {
...
+
+workflow "Trigger generate formulae.brew.sh on push" {
+  on = "push"
+  resolves = ["Generate formulae.brew.sh"]
+}
+
+action "Generate formulae.brew.sh" {
+  uses = "docker://linuxbrew/brew"
+  runs = ".github/main.workflow.sh"
+  secrets = ["ANALYTICS_JSON_KEY", "FORMULAE_DEPLOY_KEY"]
```
MikeMcQuaid: Note to self: I'll need to add these before this will work.

.github/main.workflow.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+#!/bin/bash
```
MikeMcQuaid: Dupliated.

.github/main.workflow.sh
```diff
@@ -0,0 +1,53 @@
...
+# configure git
+git config --global user.name "BrewTestBot"
+git config --global user.email "homebrew-test-bot@lists.sfconservancy.org"
+
+# create stubs so build dependencies aren't incorrectly flagged as missing
+for i in python svn unzip xz
+do
+  touch /usr/bin/$i
+  chmod +x /usr/bin/$i
+done
```
MikeMcQuaid: These shouldn't be needed here.

.github/main.workflow.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# setup Homebrew/homebrew-cask instead of Linuxbrew's
```
MikeMcQuaid: This comment no longer makes sense.

.github/main.workflow.sh
```diff
@@ -0,0 +1,53 @@
...
+CASK_DIR="$(brew --repo homebrew/cask)"
+mkdir -p "$CASK_DIR"
+rm -rf "$CASK_DIR"
+ln -s "$PWD" "$CASK_DIR"
+
+# setup Homebrew environment
+export HOMEBREW_NO_AUTO_UPDATE=1
+export HOMEBREW_NO_ANALYTICS=1
+export HOMEBREW_FORCE_HOMEBREW_ON_LINUX=1
```
MikeMcQuaid: Are you sure this is needed for Cask generation?

cmbernard333: If we aren't doing any `install` commands we could probably remove the first one. Same with analytics.

I would expect the third one as this is being run inside a docker container.

cmbernard333:
> Should this include the other cask taps?

I'm not sure @reitermarkus - I am intending this only to do `homebrew-cask` right now much in the same way `homebrew-core` is done.

.github/main.workflow.sh
cmbernard333: Removed

.github/main.workflow.sh
cmbernard333: Removed

.github/main.workflow.sh
cmbernard333: Removed

reitermarkus:
This action will be synced to the other cask taps by another action, so either make sure it works with any cask tap or make sure it isn't synced to/enabled on the other taps.

cmbernard333:
In theory it should work for any tap. It's design just to call `to_h` on each cask within a tap and jsonify that. If they are not well formed or do not adhere to the cask DSL then it would probably throw an error.

reitermarkus:
Analytics includes all taps, so it would make sense for this to also apply to all taps.

.github/main.workflow.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# silence bundler complaining about being root
+mkdir ~/.bundle
+echo 'BUNDLE_SILENCE_ROOT_WARNING: "1"' > ~/.bundle/config
+
+# configure git
+git config --global user.name "BrewTestBot"
+git config --global user.email "homebrew-test-bot@lists.sfconservancy.org"
+
+CASK_DIR="$(brew --repo homebrew/cask)"
```
reitermarkus:
```suggestion
CASK_DIR="$(brew --repo "$GITHUB_REPOSITORY")"
```

.github/main.workflow.sh
```diff
@@ -0,0 +1,44 @@
...
+mkdir ~/.bundle
+echo 'BUNDLE_SILENCE_ROOT_WARNING: "1"' > ~/.bundle/config
+
+# configure git
+git config --global user.name "BrewTestBot"
+git config --global user.email "homebrew-test-bot@lists.sfconservancy.org"
+
+CASK_DIR="$(brew --repo "$GITHUB_REPOSITORY")"
+mkdir -p "$CASK_DIR"
+rm -rf "$CASK_DIR"
```
reitermarkus: `rm` after `mkdir` doesn't make sense.

MikeMcQuaid: It's to ensure the parent hierarchy is created

cmbernard333:
@MikeMcQuaid what are the next steps here?

MikeMcQuaid:
@cmbernard333 I've been on vacation. Will get to this ASAP.

.github/main.workflow.sh
MikeMcQuaid: Let's try to remove them all for now (and `unset HOMEBREW_NO_ANALYTICS` below).

MikeMcQuaid:
Let's see how this goes after merge and I'll open up a follow-up as needed. Thanks @cmbernard333!

cmbernard333:
@MikeMcQuaid how did this work out? I'm not sure where to check the status. Is there an issue or enhancement for generating the HTML for casks? That is something I could take on in the next week or so.

MikeMcQuaid:
@cmbernard333 Thanks for checking in! We're getting Cask JSON and analytics data going in https://github.com/Homebrew/formulae.brew.sh/tree/master/_data so, yeh, I think it's a matter of getting the HTML working, linking it (e.g. on the analytics pages and index) and then everything is done!

PR #59581, opened 2019-02-28
Check for version decrease in `automerge` action.

reitermarkus:


.github/actions/automerge/git_diff_extensions.rb
```diff
@@ -14,22 +14,54 @@ def single_cask?
...
+
+    def version_changed?
+      old_version != new_version
+    end
+
+    def version_decreased?
+      return false unless version_changed?
+
+      return false if old_version == :latest
+      return true if new_version == :latest
```
scpeters: should the logic on these two lines be reversed?

~~~
      return true if old_version == :latest
      return false if new_version == :latest
~~~

reitermarkus: I think I should just remove the handling for `:latest` and never auto-merge version bumps involving `:latest`.

MikeMcQuaid (approved):
Looks great! Thanks for the fast turnaround and sorry I couldn't do this myself.

scpeters (approved):

PR #59775, opened 2019-03-04
Update Kicad to Version 5.0.2-5

ibutra:


ibutra:
This currently installs only the KiCad suit.
This is how the cask worked before and KiCad can be run this way.

The recommended install also includes a kicad folder which should go into Application Support.
Is it fine if I add this as well? If yes, any hints as to how to do it?

reitermarkus:
> The recommended install also includes a kicad folder which should go into Application Support.

Is this actually included in the download?

ibutra:
Yes, it is part of the downloaded dmg file. 
The folder contains default schematic symbols and footprints.

reitermarkus:
Ok, add it then. But add each subfolder individually, and also add the `Application Support` directory to `zap trash:`.

ibutra:
Ok, I tried various things:
First I added the folder to the user's `~/Application Support`. Sadly KiCad expects the folder in the system's folder: `/Application Support`.
I suspected, that brew cask doesn't have rights to install anything there as it shouldn't have without root. But I tried it anyways. It doesn't throw any errors but noting is added to the folder (as shouldn't be in my opinion).

I then looked into where the configuration is saved for the paths KiCad is searching for libraries etc.
This configuration is created the first time KiCad is launched and while I could create one to point to the user's `Application Support` I don't know whether this would be the right way.

I am not sure how to proceed from here, as I think the cask is not really that great without the support files (help, default symbols and footprints are missing), but I also don't really see how I can add them.

reitermarkus:
Can you add the changes you tried, even if they don't work? It's easier to review that way. 

From what I see in the DMG, it is case-sensitive, so when it is extracted the contents of `KiCad` and `kicad` are merged I assume.

ibutra:
Sorry for the delay.
I pushed the change where I try to add the kicad folder to '/Application Support'.
At least on my system adding the kicad folder fails silently.
Adding the kicad folder to '~/Application Support' works, but KiCad doesn't pick it up there.

reitermarkus:
Have you checked if anything is actually installed into `/Library/Application Support/kicad`? I think you'll need to add every subdirectory separately, since `KiCad` and `kicad` are merged into the same directory.

kavu:
@reitermarkus yeah, I tried that too, but in fact they are merged into the same dir during `suite` execution? Is there any way to get the path from Caskroom during install, to copy some files somewhere?

reitermarkus:
You could try to separate the contents in a `postflight` block.

kavu:
@reitermarkus Thanks! But I've come with a different way to handle this situation.

```ruby
cask 'kicad' do
  version '5.0.2-5'

  if MacOS.version == :mojave
    sha256 'f2c98036b624dce756ed0c01cac70ee2a096cb74df4d40e3358b4d2613b65d39'
    url "https://kicad-downloads.s3.cern.ch/osx/stable/kicad-unified-#{version}-10_14.dmg"
  else
    sha256 '1fc507987dfbc637805f1e1294480abdf869933d3300c01de691b12d655e84c7'
    url "https://kicad-downloads.s3.cern.ch/osx/stable/kicad-unified-#{version}.dmg"
  end

  appcast 'https://kicad-downloads.s3.cern.ch/index.html?prefix=osx/stable/'
  name 'KiCad'

  artifact 'KiCad/kicad.app', target: '/Applications/KiCad.app'

  artifact 'KiCad/help',     target: '/Library/Application Support/kicad/help'
  artifact 'KiCad/library',  target: '/Library/Application Support/kicad/library'
  artifact 'KiCad/modules',  target: '/Library/Application Support/kicad/modules'
  artifact 'KiCad/share',    target: '/Library/Application Support/kicad/share'
  artifact 'KiCad/template', target: '/Library/Application Support/kicad/template'

  zap trash: '~/Library/Preferences/kicad'

  uninstall trash: [
    '/Library/Application Support/kicad',
    '/Applications/KiCad'
  ]
end

```

What do you think? I can open new PR with this version to run it on the CI.

stiff:
Why is this cask so much different from 5.0.1? Asking just to make sure it won't mess existing libraries after upgrade.

ibutra:
It isn’t. The default libraries apparently just were ignored in the old casks.

kavu:
@ibutra can you test my implementation of formula, will it work for you?

ibutra:
@kavu Yes it works for me.
Have you already opened a pull request? Then I could link it and close this one.

ibutra:
Close in favor of #60128

PR #59844, opened 2019-03-06
Revert "Update postman to 7.0.4"

martinneff:
Reverts Homebrew/homebrew-cask#59815

7.0.4 ist not yet a stable Version. According to https://www.getpostman.com/downloads/release-notes the stable version ist still 6.7.4


vitorgalvao:
Thank you. Already reverted in a different commit, because the `appcast` needed to stay as it was.

PR #59865, opened 2019-03-06
formulae.brew.sh generation tweaks

MikeMcQuaid:
Need to set this file as executable as a minimum but open a PR to be able to further tweak this and test the push hook.

CC @cmbernard333 FYI


MikeMcQuaid:
@cmbernard333 @reitermarkus 

> Warning: Cask 'openoffice' is unreadable: undefined method `languages' for OS::Mac:Module

This will probably need some tweaking to read these casks on Linux. Probably just an exception that needs caught.

Issue #58883, opened 2019-02-13
Improve Android NDK path


Skoti:
### Description of feature/enhancement
It seems like `ANDROID_NDK_HOME` is no longer respected by _Android Studio_ and _sdkmanager_ as now there is only one path and it is `ANDROID_SDK_ROOT` which contains everything. NDK should be in the `ndk-bundle` subfolder.

There are only 2 ways to make it work:

1. Create a symlink in the current _android-sdk_ installation folder i.e.`/usr/local/share/android-sdk/ndk-bundle` pointing to `/usr/local/share/android-ndk`. This will make the current NDK installation to be used as a default for all projects.
2. Add `ndk.dir=/usr/local/share/android-ndk` in the `local.properties` file of an _AndroidStudio_ project. This needs to be done for every project which uses _android-ndk_.

Would it be possible to either change the _android-ndk_ installation path to `/usr/local/share/android-sdk/ndk-bundle` or change the caveat information to set _ndk.dir_ instead of using `ANDROID_NDK_HOME` environment variable?

reitermarkus:
Can you open a PR with your proposed changes? Might be easier to discuss that way.

Skoti:
Hmm, I was hoping to discuss this first before I do anything.
Simply because I am not sure how the final implementation should look and also I am not familiar with _homebrew_ code.

Changing _caveat_ information is easy and at least it could guide people what to do, however, I think we can do better than this. So let's focus on the other idea.

I think it would be best to symlink the `android-ndk` installation into `android-sdk/ndk-bundle`.
I came up with two possible solutions:

- check in _postflight_ of both _casks_ if the other one is installed and if so symlink _ndk_ as described above
- make `android-ndk` dependent on `android-sdk` so it can symlink almost without checks

Almost because in both solutions _ndk_ might already be installed via _sdkmanager_ - this means there is a conflicting folder `ndk-bundle` inside of `android-sdk` installation folder. Should we overwrite it with our symlink? Just print a warning message?

These are two ideas I have right now - which way should we go? If any at all, maybe you or someone else that have a better understanding of _homebrew_ code can come up with a more suitable solution :smile:

reitermarkus:
I have so many questions.

Is `sdkmanager` part of `android-sdk`? What would you do if you installed the NDK manually? When would you need to do so? If you can install the NDK `sdkmanager`, would you ever need to install it manually? Is there a use case for the NDK without the SDK?

[...]

PR #59874, opened 2019-03-06
removing android-ndk cask based on #58883

Skoti:
Removing android-ndk cask based on #58883


vitorgalvao:
It’s still unclear which solution is the best, but this is easily reversible, so for now let’s go through with it knowing we can undo it with a simple revert.

PR #59945, opened 2019-03-08
Delete chat.rb (duplicate)

suschizu:
I think this cask is a duplicate:
https://github.com/Homebrew/homebrew-cask/blob/master/Casks/google-chat.rb


compuguy:
Apologies  @suschizu. When I made #59630, I thought there wasn't an existing cask for hangouts chat.

PR #59946, opened 2019-03-08
Revert "Update sublime-merge to 1106"

suschizu:
Reverts Homebrew/homebrew-cask#59929
sorry @ran-dall - this is a dev-build - 1104 too: 
https://www.sublimemerge.com/dev
can we revert this to 1103?

### DIFF
--- a/Casks/sublime-merge.rb
+++ b/Casks/sublime-merge.rb
@@ -1,6 +1,6 @@
 cask 'sublime-merge' do
-  version '1106'
-  sha256 '2a3a2e4e778b5a5b01681e39fea19d1e213de05e57552727531a4adbca52c757'
+  version '1104'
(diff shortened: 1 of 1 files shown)


ran-dall:
@suschizu No problem. 👍 You should update it on @core-code's end cause the feed it has been listed a few times.

suschizu:
ok I'll manage that thanks :)

core-code:
@ran-dall  thanks for the heads-up. there is no feed involved since this came from the database check ( crdb-processed-2019-03-07.txt ). the database results always need to be checked whether they are real releases, while the feed checks should be accurate in most cases. however, if sublime-merge has dev-releases it should be removed from the database check altogether. i've done this now and instead looked for a real appcast so that we can support this with less troubles in the future.

@suschizu good job finding out that 1104 is beta too, but your revert actually contained this version instead of the stable 1103 ;)

i'll send another PR to switch to stable and add the appcast

PR #59967, opened 2019-03-08
add appcast to yubico-authenticator.rb

suschizu:


suschizu:
the same cask appears in the cask-drivers repo:
https://github.com/Homebrew/homebrew-cask-drivers/blob/master/Casks/yubico-authenticator.rb
should one be deleted perhaps @vitorgalvao?

vitorgalvao:
> should one be deleted perhaps @vitorgalvao?

Yes. This one.

PR #60020, opened 2019-03-09
Update whatsapp to 0.3.2386

ran-dall:

### DIFF
--- a/Casks/whatsapp.rb
+++ b/Casks/whatsapp.rb
@@ -1,6 +1,6 @@
 cask 'whatsapp' do
-  version '0.3.2276'
-  sha256 'feb0b6b883632b08e1ba2064c52d53e4bf512ecce97c77379fa86d7022a3825a'
+  version '0.3.2386'
+  sha256 'aefa68840293eba9233051624fae993d54501791b03b880ab272f10f54cbdd62'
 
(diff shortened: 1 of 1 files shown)


core-code:
this release is neither in the appcast nor in the official download from the homepage. 

if no checking goes on both on the submitting as well in the merging end, our process is broken at the moment...

@vitorgalvao i have worked on checking against beta releases by seeing if the version number is contained in the appcast. however, cleaning up afterwards is not ideal. is the plan to keep the bot going? if so maybe we could add that logic there so it can refuse to merge if the version number is not in the appcast.

core-code:
@ran-dall  how is the verification performed? perhaps i can help here...  hm, i realise that both the human-readable format as well as the fact that we add things are not ideal for automatic parsing of the files. i could try to generate a format that is better for parsing like XML or JSON to make things easier here.

ran-dall:
@core-code I'm trying to think of the best way to explain it. 🤔 A lot of the background research is summarized here (most consise link I could find): https://blog.acolyer.org/2018/11/28/towards-usable-checksums-automating-the-integrity-verification-of-web-downloads-for-the-masses ; which is what made me start to playing around with the idea of augementing their research with Machine Learning on Homebrew Cask.

PR #60128, opened 2019-03-11
KiCad 5.1.0

kavu:

Bump KiCad 5.0.2-5. It is based on existing #59775. 

### DIFF
--- a/Casks/kicad.rb
+++ b/Casks/kicad.rb
@@ -1,13 +1,35 @@
 cask 'kicad' do
-  version '5.0.1'
-  sha256 '3066e7c8da09afea2e27b4dc5d0742c83c5a2e93408d9d76e342e14e594bc7b3'
+  version '5.1.0'
 
-  url "http://downloads.kicad-pcb.org/osx/stable/kicad-unified-#{version}.dmg"
-  appcast 'http://downloads.kicad-pcb.org/osx/stable/'
+  if MacOS.version <= :high_sierra
+    sha256 'be024d121a42c9d1e1ea08fdf9dd2ad24cf77879d7f32214b2de13322cd18bae'
+    url "https://kicad-downloads.s3.cern.ch/osx/stable/kicad-unified-#{version}.dmg"
+  else
+    sha256 '23955e59e0ac1fe7db86156e0470bd2bc8f1c1666ef1c5def5fa3a82932c9907'
+    url "https://kicad-downloads.s3.cern.ch/osx/stable/kicad-unified-#{version}-10_14.dmg"
+  end
+
+  appcast 'https://kicad-downloads.s3.cern.ch/?delimiter=/&prefix=osx/stable/'
   name 'KiCad'
   homepage 'http://kicad-pcb.org/'
 
-  suite 'KiCad'
+  app 'KiCad/kicad.app',            target: 'KiCad/KiCad.app'
+  app 'KiCad/bitmap2component.app', target: 'KiCad/bitmap2component.app'
+  app 'KiCad/eeschema.app',         target: 'KiCad/eeschema.app'
+  app 'KiCad/gerbview.app',         target: 'KiCad/gerbview.app'
+  app 'KiCad/pcb_calculator.app',   target: 'KiCad/pcb_calculator.app'
+  app 'KiCad/pcbnew.app',           target: 'KiCad/pcbnew.app'
+  app 'KiCad/pl_editor.app',        target: 'KiCad/pl_editor.app'
+  artifact 'kicad/help',            target: '/Library/Application Support/kicad/help'
+  artifact 'kicad/library',         target: '/Library/Application Support/kicad/library'
+  artifact 'kicad/modules',         target: '/Library/Application Support/kicad/modules'
+  artifact 'kicad/share',           target: '/Library/Application Support/kicad/share'
+  artifact 'kicad/template',        target: '/Library/Application Support/kicad/template'
+
+  uninstall rmdir: [
+                     '/Library/Application Support/kicad',
+                     '/Applications/KiCad',
+                   ]
 
   zap trash: '~/Library/Preferences/kicad'
 end


Casks/kicad.rb
```diff
@@ -1,13 +1,26 @@
...
+    sha256 'f2c98036b624dce756ed0c01cac70ee2a096cb74df4d40e3358b4d2613b65d39'
+    url "https://kicad-downloads.s3.cern.ch/osx/stable/kicad-unified-#{version}-10_14.dmg"
+  end
+
+  appcast 'https://kicad-downloads.s3.cern.ch/index.html?prefix=osx/stable/'
   name 'KiCad'
   homepage 'http://kicad-pcb.org/'
 
-  suite 'KiCad'
+  app 'KiCad/kicad.app'
```
reitermarkus: Before, this was a suite, is this really the only app contained in this directory?

stiff: Technically in fact kicad.app is just a convenient way to launch pcbnew, eeschema and other apps. In current install (5.0.1) /Applications/Kicad has 7 apps. I quite often run pcb_calculator and pcbnew in standalone mode, because it's faster to just click on .pcb file in Finder to quickly see what's there or do a minor edit, rather than opening project, and opening pcb from there.

stiff: Maybe a silly question, but why do we need to get rid of `suite` in the first place? :)

kavu:
Hmm, still can't beat the CI, uninstall step doesn't complete properly. Even without my tries to return supplemental apps besides KiCad.app.

Casks/kicad.rb
reitermarkus: > Well, then we need to find a way to install them in as a suite.

I already mentioned this in the other PR: Separate the contents into two directories in a `preflight` block.

core-code:
thanks for working on this.

btw, can you also change the appcast? the current one works in a browser, but is useless to process automatically, 
`curl "https://kicad-downloads.s3.cern.ch/index.html?prefix=osx/stable/"
`
yields nothing

this one should work nicely:
`https://kicad-downloads.s3.cern.ch/?delimiter=/&prefix=osx/stable/`

ibutra:
@kavu Are you still working on this? Especially the changes recommended by @reitermarkus?
If not I could pick it up again and try that. (Even though I have no clue how to do that but I am positive I could figure it out)

kavu:
@ibutra Yeah, I am still in. 

By the way, it looks like currently formula working on CI as intended — creating a suite in /Application, copies all support file into Application Support. But CI fails to uninstall properly, while it works on my localhost. @reitermarkus can you give an advice, what can be the problem here? I doubt it can be cause by missing suite stanza, but I am not 100% sure.

core-code:
i'd try changing "uninstall trash" to "uninstall delete" to avoid the "AppleEvent timed out" problem.

kavu:
`uninstall :delete` is semantically different, it will delete all artefacts, possibly deleting important files in KISYSMOD (`Application Support/kicad`), that's why I think it's better to leave it as `:trash`. But moving `Application Support/kicad` is also an option.

What do you think?

fauxpark:
Just FYI, 5.1.0 has just been released: http://kicad-pcb.org/download/osx/

stiff:
Changelog for 5.1 looks promising

fauxpark:
Spoke too soon, there seems to be a problem with the dmg for <= 10.13 - it's identical to the 10.14 one:
https://forum.kicad.info/t/kicad-5-1-0-released/15751/6

kavu:
Update to 5.1.0. Still have a problem with `uninstall :trash`. @reitermarkus Should I change it to `:delete`?

kavu:
@fauxpark Yeah, you are right. It looks like that we are waiting again for a fix for =< 10.13. I have 10.13.3 right now and even I downloaded `kicad-unified-5.1.0.dmg` it still doesn't work for me.

fauxpark:
@kavu looks like it's been fixed now

Casks/kicad.rb
```diff
@@ -1,13 +1,35 @@
...
+  app 'KiCad/pcbnew.app',           target: 'KiCad/pcbnew.app'
+  app 'KiCad/pl_editor.app',        target: 'KiCad/pl_editor.app'
+  artifact 'KiCad/help',            target: '/Library/Application Support/kicad/help'
+  artifact 'KiCad/library',         target: '/Library/Application Support/kicad/library'
+  artifact 'KiCad/modules',         target: '/Library/Application Support/kicad/modules'
+  artifact 'KiCad/share',           target: '/Library/Application Support/kicad/share'
+  artifact 'KiCad/template',        target: '/Library/Application Support/kicad/template'
+
+  uninstall delete: [
+                      '/Library/Application Support/kicad',
```
reitermarkus: This isn't needed now, since all `artifact`s are uninstalled already. You can add this as `rmdir:` though.

Casks/kicad.rb
```diff
@@ -1,13 +1,35 @@
...
 
-  suite 'KiCad'
+  app 'KiCad/kicad.app',            target: 'KiCad/KiCad.app'
+  app 'KiCad/bitmap2component.app', target: 'KiCad/bitmap2component.app'
+  app 'KiCad/eeschema.app',         target: 'KiCad/eeschema.app'
+  app 'KiCad/gerbview.app',         target: 'KiCad/gerbview.app'
+  app 'KiCad/pcb_calculator.app',   target: 'KiCad/pcb_calculator.app'
+  app 'KiCad/pcbnew.app',           target: 'KiCad/pcbnew.app'
+  app 'KiCad/pl_editor.app',        target: 'KiCad/pl_editor.app'
+  artifact 'KiCad/help',            target: '/Library/Application Support/kicad/help'
```
reitermarkus: This should use lowercase `kicad`, so it also works on case-sensitive file systems.

kavu: Fixed and still passing the CI, yay!

Casks/kicad.rb
```diff
@@ -1,13 +1,35 @@
...
+
+  appcast 'https://kicad-downloads.s3.cern.ch/?delimiter=/&prefix=osx/stable/'
   name 'KiCad'
   homepage 'http://kicad-pcb.org/'
 
-  suite 'KiCad'
+  app 'kicad/kicad.app',            target: 'KiCad/KiCad.app'
+  app 'kicad/bitmap2component.app', target: 'KiCad/bitmap2component.app'
+  app 'kicad/eeschema.app',         target: 'KiCad/eeschema.app'
+  app 'kicad/gerbview.app',         target: 'KiCad/gerbview.app'
```
reitermarkus: The `app`s are in `KiCad`, so this should still be capitalized.

kavu: Done! Thank you!

reitermarkus:
Thanks!

PR #60198, opened 2019-03-12
Update camtasia to 2018.0.8

core-code:


core-code:
1.) it seems the hockeyapp feed is for beta versions, so i've reverted to the stable release according to their homepage and built-in sparkle check
2.) they have a sparkle feed but it requires spoofing user-agent for reading the feed and for parsing its download location ( https://www.techsmith.com/snagitmactudi/?version=18.0.0 ) so i don't think there is any point in trying to add this to the cask ATM
3.) given that the file is updated in place, we could also just put "2018" in the version and use no-sha. if you prefer that, just let me know and i'll change the cask over

PR #60208, opened 2019-03-12
Update postman appcast stanza

core-code:
it seems when you pass the current version to the appcast it just returns nothing


jfoy:
@core-code This change seems to have broken Postman installs. When I revert it locally, `brew cask upgrade postman` works as expected.

core-code:
aaah sorry i meant to update the appcast stanza. will send a PR to fix this right now.

PR #60209, opened 2019-03-12
Update lightsoff appcast stanza

core-code:


vitorgalvao:
The main site redirects to the gumroad page, now. And it looks that we need to give an email address to get it. Which means we’ll have to remove the cask.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/60225

core-code:
ok thanks

PR #60242, opened 2019-03-13
Add pointer to HOMEBREW_DEVELOPER setting in Contributing doc.

toothbrush:
I found it counter-intuitive that `brew cask` tried to switch to the Git `master` branch while i was trying to do development on a new Cask.  I think it might be good to include a pointer in the Contributing docs to avoid others having issues similar to mine in #60241.

I'd love to add a link from here to somewhere where the `HOMEBREW_DEVELOPER` variable is explained.  The `man brew` text is very brief and IMHO didn't help me discover my own error:

       HOMEBREW_DEVELOPER
              If set, Homebrew will tweak behaviour to be more relevant for Homebrew developers (active  or  bud-
              ding), e.g. turning warnings into errors.

Comments?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,8 @@ $ git remote add "${github_user}" "https://github.com/${github_user}/homebrew-ca
 
 4: Switch to a new branch (ie. `new-feature`), and work from there: `git checkout -b new-feature`.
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you for the contribution. But as you noted, this is already explained in the documentation for adding a cask. This isn’t a widespread issue, so I donLt think it’s worth it to add the messaging here.

I do think this section could do with some improvements, though. I’ll give it a thought.

PR #60396, opened 2019-03-14
Split versions and compare parts.

reitermarkus:


core-code:
i am not sure if i read the ruby correctly but just wanted to mention that other commonly used separators besides "," are "-" and "_"

PR #60458, opened 2019-03-16
Update tankix.rb / add app version to version stanza 

suschizu:
would it be possible to add the real app version to the version stanza although it's not used inside the url stanza?


vitorgalvao:
Yes, I understand the real version is useful for your purposes.

PR #60503, opened 2019-03-17
Update itch formula to use itch-setup v1.17.0

dgellow:
Multiple issues with the previous version of the formula:
- the version in the formula (25.4.0) was the version of the installed application, not the version of the installer that the formula downloads (1.17.0), and wasn't used in the URL making it a bit meaningless
- the url used (https://nuts.itch.zone/download/osx) can serve content that will change over time. https://broth.itch.ovh has been [suggested to me](https://twitter.com/fasterthanlime/status/1107367222010351622) by @fasterthanlime, and should be stable.
- the shasum doesn't match the downloaded content
- the formula was downloading a dmg then extracting the installer, while the installer is available naked thus would be a better choice

## PR checklist

### DIFF
--- a/Casks/itch.rb
+++ b/Casks/itch.rb
@@ -5,11 +5,8 @@ cask 'itch' do
   # broth.itch.ovh was verified as official when first introduced to the cask
   url "https://broth.itch.ovh/itch-setup/darwin-amd64/#{version}/archive/default"
   appcast 'https://github.com/itchio/itch/releases.atom'
-  name 'itch-setup'
-  name 'the itch.io app installer'
+  name 'itch.io'
   homepage 'https://itch.io/app'
-
-  container type: :zip
 
   installer script: 'itch-setup'
 
--- a/Casks/itch.rb
+++ b/Casks/itch.rb
@@ -1,20 +1,17 @@
 cask 'itch' do
-  version '25.4.0'
-  sha256 'b88dc10ceb69ece6ecab17340e6c4abc6e89e8eb20c7b030207eca5d0c2d8f17'
+  version '1.17.0'
+  sha256 '9053778cbae81ce09129364e0b3ec7f2b9abdd9705334011398d2d66d553c31c'
 
-  # nuts.itch.zone was verified as official when first introduced to the cask
-  url 'https://nuts.itch.zone/download/osx'
+  # broth.itch.ovh was verified as official when first introduced to the cask
+  url "https://broth.itch.ovh/itch-setup/darwin-amd64/#{version}/archive/default"
   appcast 'https://github.com/itchio/itch/releases.atom'
-  name 'itch'
+  name 'itch-setup'
+  name 'the itch.io app installer'
   homepage 'https://itch.io/app'
 
-  container nested: 'Install itch.dmg'
+  container type: :zip
 
-  installer script: 'Install itch.app/Contents/MacOS/itch-setup'
-
-  preflight do
-    set_permissions "#{staged_path}/Install itch.app", '0777'
-  end
+  installer script: 'itch-setup'
 
   uninstall delete: [
                       '~/Applications/itch.app',


Casks/itch.rb
```diff
@@ -1,19 +1,20 @@
 cask 'itch' do
-  version '25.4.0'
-  sha256 'b88dc10ceb69ece6ecab17340e6c4abc6e89e8eb20c7b030207eca5d0c2d8f17'
+  version '1.17.0'
+  sha256 '9053778cbae81ce09129364e0b3ec7f2b9abdd9705334011398d2d66d553c31c'
 
-  # nuts.itch.zone was verified as official when first introduced to the cask
-  url 'https://nuts.itch.zone/download/osx'
+  # broth.itch.ovh was verified as official when first introduced to the cask
+  url "https://broth.itch.ovh/itch-setup/darwin-amd64/#{version}/archive/default"
```
dgellow: URL provided by @fasterthanlime, itch's maintainer. See https://twitter.com/fasterthanlime/status/1107367222010351622 for the actual source.

Casks/itch.rb
```diff
@@ -1,19 +1,20 @@
...
   appcast 'https://github.com/itchio/itch/releases.atom'
-  name 'itch'
+  name 'itch-setup'
+  name 'the itch.io app installer'
```
dgellow: Not sure if that's something common or correct, I thought that "itch-setup" wasn't that clear or explicit thus added a second name.

Casks/itch.rb
```diff
@@ -1,19 +1,20 @@
...
 
-  container nested: 'Install itch.dmg'
+  container type: :zip
 
-  installer script: 'Install itch.app/Contents/MacOS/itch-setup'
+  installer script: 'itch-setup'
 
   preflight do
-    set_permissions "#{staged_path}/Install itch.app", '0777'
+    set_permissions "#{staged_path}/itch-setup", '0755'
```
dgellow: I don't know why it was chmod as 0777 before, but that's not necessary.

Casks/itch.rb
fasterthanlime: Can confirm https://broth.itch.ovh are supported and will be stable in the future, whereas https://nuts.itch.zone will be retired when v23 usage falls beneath a threshold.

Casks/itch.rb
fasterthanlime: Just providing a point of comparison: the AUR (ArchLinux) package uses the name `itch` (unlike the proposed changes), but is versioned after the setup program, like the proposed changes: https://aur.archlinux.org/packages/itch/

dgellow: @fasterthanlime just to clarify and be sure that we are talking about the same thing, I didn't change the cask "identifier" (to not use "name" 🙃 ), it would still be available via `brew cask install itch`. The stanza `name` is defined in https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/name.md as 
> `name` accepts a UTF-8 string defining the full name of the software, and is used to help with searchability and disambiguation. It can be repeated multiple times if there are useful alternative names.

> Its first instance should use the latin alphabet, include the software vendor’s name, and be as verbose as possible while still making sense.

So it's mostly for documentation and search purpose. After reading this I assumed that using `itch-setup` as the name makes it more clear that the version `v1.17.0` is the version of the installer and not the application itself (which would be `v25.4.0`). But I'm not that familiar with homebrew cask to know for sure that I'm doing the correct thing.

Casks/itch.rb
```diff
@@ -1,19 +1,20 @@
 cask 'itch' do
-  version '25.4.0'
-  sha256 'b88dc10ceb69ece6ecab17340e6c4abc6e89e8eb20c7b030207eca5d0c2d8f17'
+  version '1.17.0'
+  sha256 '9053778cbae81ce09129364e0b3ec7f2b9abdd9705334011398d2d66d553c31c'
 
-  # nuts.itch.zone was verified as official when first introduced to the cask
-  url 'https://nuts.itch.zone/download/osx'
+  # broth.itch.ovh was verified as official when first introduced to the cask
```
vitorgalvao: Thank you for the fixes and confirmations @dgellow and @fasterthanlime. However, this line serves for *users* to be able to audit the cask and know this is not a malicious URL. Is there anywhere official (the website) where this can be verified?

fasterthanlime: Well, I suppose I could mention it somewhere in https://github.com/itchio/itch's documentation? (or https://github.com/itchio/itch-setup, in this case)

(I'm an @itchio employee and the main developer of the itch app)

vitorgalvao: > Well, I suppose I could mention it somewhere in https://github.com/itchio/itch's documentation?

If the main website links somewhere to that repo, then that repo has the information, that would work, yes.

> I'm an @itchio employee and the main developer of the itch app

I know, but we still need a way for a user to figure that out without going through the issues. Auditability is important to us.

fasterthanlime: Understood!

https://itch.io/app does link to https://github.com/itchio/itch - the top-right logo is a link, and there's another link further down the page:

image

image

I've just added a note about itch setup and broth here: https://github.com/itchio/itch#about-itch-setup

dgellow:
Hi @vitorgalvao. Do you have more feedback for this pull request? @fasterthanlime updated itch documentation to mention their release servers, I assume that part is good now?

Any comment on the "name" section, or the rest of the formula? I would be happy to do changes if anything is still blocking this pull request.

Casks/itch.rb
reitermarkus: Is it necessary at all now?

Casks/itch.rb
vitorgalvao: @fasterthanlime That’s perfect. Thank you.

core-code:
while we are at this, is there any way to avoid the installer altogether and just download the app?

fasterthanlime:
> while we are at this, is there any way to avoid the installer altogether and just download the app?

There is [it's also on broth](https://broth.itch.ovh/itch/darwin-amd64), but there's no way to disable the auto-updater right now (which is also powered by itch-setup), so it may behave weirdly.

I'd prefer for it to use itch-setup, like the AUR package does for ArchLinux.

core-code:
thanks. lets keep this cask as planned, including the rename. i'll see if i can submit an additional cask for those that want to avoid the installer.

vitorgalvao:
> but there's no way to disable the auto-updater

Neither would we want that. We embrace apps that auto-update (it’s so common in macOS apps, fighting it wouldn’t make sense).

fasterthanlime:
> Neither would we want that. We embrace apps that auto-update (it’s so common in macOS apps, fighting it wouldn’t make sense).

Thanks. I'm the sole maintainer and have spend years fighting with Squirrel.Windows & Squirrel.Mac, and at some point I decided 1) I was tired of looking at stack traces I didn't understand 2) Deployment took a lot longer than needed 3) I already spent quote some time developing cross-platform installing/patching technology, so itch-setup was born, and it works the same way on all 3 platforms (save for the UI and some staging strategies). It downloads directly from our CDN, never requires superuser/root, updates in the background, has small patches and lets us deploy fast, should we need it.

I appreciate third-party formulas/install scripts using itch-setup for those reasons.

Casks/itch.rb
dgellow: Nop, you're right, that's not necessary. I removed that part.

dgellow:
Hey there. This PR is still flagged with "awaiting user reply", but I think that everything has been answered so far (huge thank you @fasterthanlime for providing the additional information 🙏). Is anything still keeping it from being merged?

Casks/itch.rb
```diff
@@ -1,20 +1,17 @@
...
+  url "https://broth.itch.ovh/itch-setup/darwin-amd64/#{version}/archive/default"
   appcast 'https://github.com/itchio/itch/releases.atom'
-  name 'itch'
+  name 'itch-setup'
+  name 'the itch.io app installer'
   homepage 'https://itch.io/app'
 
-  container nested: 'Install itch.dmg'
+  container type: :zip
```
vitorgalvao: Are you sure this is needed? Doesn’t HBC detect the container correctly?

dgellow: Hmm, you're right, following the URL ultimately results in the file `itch-setup-darwin-amd64.zip`, so the `type: :zip` isn't necessary. I will remove it.

vitorgalvao:
Looks ready. Thank you for the work and patience, @dgellow and @fasterthanlime.

PR #60637, opened 2019-03-20
Update eclipse-ide to 4.11.0,2019-03:R

core-code:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #60742, opened 2019-03-22
update_cask_family: add eclipse and unity

vitorgalvao:
As pointed out it https://github.com/Homebrew/homebrew-cask/pull/60637#issuecomment-474922399.

### DIFF
--- a/developer/bin/update_cask_family
+++ b/developer/bin/update_cask_family
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 readonly program="$(basename "${0}")"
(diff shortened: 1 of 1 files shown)


core-code:
thanks! can you also add unity?

```
readonly families=(unity eclipse flash navicat) # Needs to be in sync with the 'get_' functions
	
	
function get_unity {
	  brew search 'unity' | grep '^unity' | grep -v 'hub' | grep -v 'player' | grep -v 'assets'
}
```

PR #60779, opened 2019-03-23
Update signal to 1.23.1

ahousseini:

### DIFF
--- a/Casks/signal.rb
+++ b/Casks/signal.rb
@@ -1,6 +1,6 @@
 cask 'signal' do
-  version '1.22.0'
-  sha256 '777496e4b23f1d088383dfc9694deb646313a2020bd5cd17ab93f68be31d5317'
+  version '1.23.1'
+  sha256 'baf2a609e78c0314c3bbc509c8d7ea8878425ca34e3f43856cda3fe3d911dc1e'
 
   url "https://updates.signal.org/desktop/signal-desktop-mac-#{version}.zip"
   appcast 'https://github.com/signalapp/Signal-Desktop/releases.atom'


core-code:
these damaged zip files are becoming more common. i had the idea of fixing them in a 'preflight' stanza (search for 0x06054b50 and remove the additional bytes), but it seems 'preflight' is run after decompression...

vitorgalvao:
@core-code Maybe switching to something other than `unzip` (like `ditto`) will work.

core-code:
@vitorgalvao i can confirm that using 'ditto -xk' to extract 'signal-desktop-mac-1.23.1.zip' works and produces a launchable app that is bit-for-bit identical with the one that the macOS GUI unarchiver produces.

vitorgalvao:
@reitermarkus How about changing the unzipping strategy to use `ditto`? What stops us, if anything?

reitermarkus:
The problem is some things fail with `ditto` and some fail with `unzip`. We already use `ditto` as a fallback for a specific type of error.

core-code:
it seems upstream doesn't want to fix this, so we need a solution or a workaround in HBC.

i propose a workaround by adding
  
```
container type: :naked
…
preflight do
	    system_command '/usr/bin/ditto', args: ['-xk', staged_path.join(filename), staged_path]
end
```

see  #61140 for my try to get signal working this way.

vitorgalvao:
> it seems upstream doesn't want to fix this, so we need a solution or a workaround in HBC.

https://github.com/Homebrew/brew/pull/5954

core-code:
sorry didn't see that a real solution was already in-progress

PR #60861, opened 2019-03-25
Update jump to 8.2.17

LanikSJ:

### DIFF
--- a/Casks/jump.rb
+++ b/Casks/jump.rb
@@ -1,8 +1,8 @@
 cask 'jump' do
-  version '8.2.16'
-  sha256 '1d9e5e14a98bac66091ccc11d857d6980442b7beb9ff73d700d7de6811249ec7'
+  version '8.2.17'
+  sha256 '2a68bc68c2f2dd6c1b506b1ffc2346510992714b03b34950e4dc0f3f71eecd52'
 
-  url 'https://jumpdesktop.com/downloads/jdmac'
+  url 'https://mirror.jumpdesktop.com/downloads/JumpDesktopMac-8.2.17.zip'
   appcast 'https://jumpdesktop.com/downloads/viewer/jdmac-web-appcast.xml'
   name 'Jump Desktop'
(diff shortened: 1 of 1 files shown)


Casks/jump.rb
```diff
@@ -1,8 +1,8 @@
 cask 'jump' do
-  version '8.2.16'
-  sha256 '1d9e5e14a98bac66091ccc11d857d6980442b7beb9ff73d700d7de6811249ec7'
+  version '8.2.17'
+  sha256 '2a68bc68c2f2dd6c1b506b1ffc2346510992714b03b34950e4dc0f3f71eecd52'
 
-  url 'https://jumpdesktop.com/downloads/jdmac'
+  url 'https://mirror.jumpdesktop.com/downloads/JumpDesktopMac-8.2.17.zip'
```
vitorgalvao: How did you find this?

LanikSJ: Stole it from MacUpdater:
image

vitorgalvao: Ah, nice, it’s in the `appcast`. Nice work. Thank you.

Pinging @core-code because it’s great that a user of both MacUpdater and Homebrew Cask is using one to improve the other.

vitorgalvao (approved):

Casks/jump.rb
core-code: awesome ;)

LanikSJ: I've said it before MacUpdater is worth every penny. 💰

core-code: glad to hear it (again ;)! 
btw, about this pull request, it shouldn't be strictly necessary to update the URL as the old one also downloads 8.2.17, but since we prefer versioned downloads it is good that way.

LanikSJ: Honestly that was my motivation for changing the URL.  I personally dislike apps that use a URL that points to latest version of their release since it's hard to tell when there's a new update unless you download and install it again.

PR #60917, opened 2019-03-26
version.md: removed slashes

vitorgalvao:
To be merged after https://github.com/Homebrew/brew/pull/5893.


reitermarkus:
`*_to_slashes` should still work, only `slashes_to_*` shouldn't, since versions shouldn't include slashes.

vitorgalvao:
`*_to_slashes` is seldom used (if ever) and also needs to be removed, because otherwise it’s another possible attack vector: a malicious actor would find a cask with a `*_to_slashes` and put the `*` in the version to have it converted.

I’m not even convinced allowing letters in `automerge` is worth it, because I feel that may still lead to an attack. If there’s one thing we need to assume, is that malicious actors are smarter than us.

`slashes_to_*` was always useless, because we never have slashes in `versions`, since it makes problematic directory names when installing. So since they’re both seldom used and one is an attack vector, might as well remove both and avoid confusion.

PR #60930, opened 2019-03-27
syncplay 1.6.3: add legacy version

albertosottile:
I saw you correctly limited Syncplay compatibility to Sierra and onwards, but we also have a legacy release that supports from 10.6.8 to El Capitan. This update to the cask includes this release for people on older systems.

**NOTE:** we do not update the legacy version as often as the regular release so, in time, the version numbers will certainly differ. I wanted to have two version stanzas in this cask, but `brew cask style` complained that, at this moment, the two version number are equal. Please keep this in mind when you will update the version number of the regular release. 

Thanks for supporting Syncplay on homebrew-cask! 

-------------------------------------------------

### DIFF
--- a/Casks/syncplay.rb
+++ b/Casks/syncplay.rb
@@ -1,14 +1,18 @@
 cask 'syncplay' do
   version '1.6.3'
-  sha256 '27cc76a66a2a1c79d2e46ad6568ed8678b83f00d80a3ec7db5527774f2223caf'
+
+  if MacOS.version <= :el_capitan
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Welcome @albertosottile, and thank you for the PR.

I’ll refuse it because the legacy version is not worth the support effort. By your own admission it will become a maintenance burden when the versions differ, and the cask isn’t popular:

```
syncplay (added 149 days ago)
30 days: 5 (#2918)
90 days: 14 (#2933)
365 days: 43 (#2657)
```

albertosottile:
Well, the extra maintenance cost will occur only during the next update. After that, the version for the legacy version will be fixed to the current value (probably indefinitely). In any case, I respect your decision.

PR #60964, opened 2019-03-27
Revert sourcetree to 3.1.1_213

core-code:


core-code:
sourcetree is notorious for a lot of insane stuff (appcasts yield randomly staged results, website gives random release, multiple releases just disappear, etc), but 3.1.2 has never been on either the homepage or in any feed as far as i know, so we shouldn't use it for the time being.

PR #61035, opened 2019-03-28
Update wondershare-video-converter-ultimate from 10.3.2.6,735 to 10.3.3.6,735

core-code:


core-code:
i question the wisdom of having the 735 in the version stanza. it never changes. it hasn't changed since the cask was added 3 years ago. for all we know it could just mean 'cool' in chinese and is probably as likely to change as other parts of the filename, e.g. the 'full'

same goes for aimersoft-video-converter-ultimate which is the same thing with a different branding and where the 747 in the filename hasn't changed since it was added in 2014.

PR #61140, opened 2019-03-30
Update signal to 1.23.1

core-code:
with workaround for ZIP-problem ;)


vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/60779#issuecomment-478259085

PR #61147, opened 2019-03-30
Fix permissions for update multilangual casks

core-code:
update_multilangual_casks not marked executable


core-code:
i think something is wrong here.

PR #61241, opened 2019-04-01
Add additional framework ids to list_ids_in_app script

ytbmulder:
As discussed with @vitorgalvao in https://github.com/Homebrew/homebrew-cask-drivers/pull/857, several frameworks are possibly overkill to be added to the `quit` or `signal` stanza. This PR adds these frameworks to the `clean_sources ()` method of the `list_ids_in_app` script.

As a side question, why is an id such as `^com\.3dconnexion\.driver\.client` included?
https://github.com/Homebrew/homebrew-cask/blob/5667013df750383edcfe884b4e10fec3841d23b5/developer/bin/list_ids_in_app#L46
This seems to me like an id only used by one cask: 
https://github.com/Homebrew/homebrew-cask-drivers/blob/master/Casks/3dconnexion.rb


ytbmulder:
> Those scripts are old. If you see something that looks like an outdated rule (like this), feel free to comment on why (like you did) and make the change in the PR. If the change is problematic we’ll know soon; if it is not, the script was improved. Better to have someone try to fix what they find than leaving the abandoned.

I removed the rule in question.

PR #61247, opened 2019-04-01
Update intel-haxm from 7.3.2 to 7.4.1

robertleeblairjr:


robertleeblairjr:
I'm not sure why this would fail. I've installed and uninstalled both versions, 7.3.2 and 7.4.1, without incident on macOS 10.14.4. The files included in the archive and installation method are the same between both versions. The silent installers are the same other than the one change to the version numbering on one line in the script. The installer for both packages require sudo privileges and acknowledging a blocked System Extension which must be enabled during the installation.

reitermarkus:
I think this always fails on Travis.

robertleeblairjr:
Awesome. Thanks for the update and will do.

PR #61623, opened 2019-04-09
Update Trader-Workstation to 9.75

fasmat:


vitorgalvao (approved):

vitorgalvao:
> Is there some way to automatically get a message when the recipe for the cask breaks? 

None that we provide. You could setup your own script to try to install it everyday and warn on failures.

> So when 9.75.4 will be released the sha will change and the script breaks.

Which would also happen on `9.76.x`.

PR #61990, opened 2019-04-16
Update anydesk from latest to 4.3.0

tbrek:

### DIFF
--- a/Casks/anydesk.rb
+++ b/Casks/anydesk.rb
@@ -1,6 +1,6 @@
 cask 'anydesk' do
-  version :latest
-  sha256 :no_check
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Seeing as `url` doesn’t point to a specific version and there’s no [appcast](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md), this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/homebrew/homebrew-cask/blob/37f8d181d1789cbb06c6e73625587fdbe77ca225/Casks/kibako.rb#L2L3).

PR #62036, opened 2019-04-17
Update vscodium: add binary stanza

nicerloop:

### DIFF
--- a/Casks/vscodium.rb
+++ b/Casks/vscodium.rb
@@ -7,7 +7,10 @@ cask 'vscodium' do
   name 'VSCodium'
   homepage 'https://github.com/VSCodium/vscodium'
 
+  conflicts_with cask: 'visual-studio-code'
+
   app 'VSCodium.app'
+  binary "#{appdir}/VSCodium.app/Contents/Resources/app/bin/code"
 
   zap trash: [
                '~/Library/Application Support/VSCodium',


Casks/vscodium.rb
```diff
@@ -7,7 +7,10 @@
   name 'VSCodium'
   homepage 'https://github.com/VSCodium/vscodium'
 
+  conflicts_with cask: 'visual-studio-code'
```
vitorgalvao: By setting this, you’re saying both apps can’t coexist in the same system. That doesn’t seem to be the case.

vitorgalvao: Ah, it’s because of the `binary`. Then yes, this was correct. Please revert to how it was.

nicerloop: I just renamed the binary to avoid the conflict. Travis build is in progress...

vitorgalvao: > I just renamed the binary to avoid the conflict.

And what I’m saying is “don’t rename the binary. Instead, make this as it was”.

nicerloop: I initially did not check if the binary could be (re)named from its very definition, in order to avoid conflicts, thus I blindly copied the visual-studio-code one and added the conflict.

Is there a policy about such binary definitions? It is quite easier to avoid the conflict with a 'codium' binary name.

vitorgalvao: Yes, there is policy: do not rename things just because it’s convenient, because that causes surprises and we don’t want to impose opinions on users.

nicerloop: Ok, then I will revert to the original binary name and conflict.

nicerloop: Reverted to simple binary + conflict

PR #62076, opened 2019-04-17
Delete bluos-controller.rb

vitorgalvao:
Not popular:

```
bluos-controller (added 235 days ago)
30 days: 4 (#3114)
90 days: 6 (#3930)
365 days: 16 (#4038)
```


micdah:
Why remove something, even if unpopular? Seems odd

vitorgalvao:
> Why remove something, even if unpopular?

Because more casks means more maintenance burden, which means less time dedicated to fix what people use most.

I found this cask while making a change in a bunch of them (of which this one was part).

PR #62079, opened 2019-04-17
Delete powermate.rb

vitorgalvao:
Not popular:

```
powermate (added 1343 days ago)
30 days: 1 (#4847)
90 days: 4 (#4608)
365 days: 20 (#3826)
```


phaseOne:
@vitorgalvao Why remove unpopular casks? Even though the publisher hasn't updated the software for years, it still works on 10.15.

PR #62201, opened 2019-04-20
Update adoptopenjdk from 11.0.2,9 to 12,33

tiiiecherle:

### DIFF
--- a/Casks/adoptopenjdk.rb
+++ b/Casks/adoptopenjdk.rb
@@ -1,14 +1,14 @@
 cask 'adoptopenjdk' do
-  version '12,33'
+  version '12.33'
   sha256 '985036459d4ef0867a3fe83b0bf87877d8e66a121c7b9c145bb97bd921aaf3f1'
 
   # github.com/AdoptOpenJDK was verified as official when first introduced to the cask
-  url "https://github.com/AdoptOpenJDK/openjdk#{version.before_comma}-binaries/releases/download/jdk-#{version.before_comma}+#{version.after_comma}/OpenJDK#{version.before_comma}U-jdk_x64_mac_hotspot_#{version.before_comma}_#{version.after_comma}.tar.gz"
-  appcast "https://github.com/adoptopenjdk/openjdk#{version.before_comma}-binaries/releases.atom"
+  url "https://github.com/AdoptOpenJDK/openjdk#{version.major}-binaries/releases/download/jdk-#{version.major}+#{version.minor}/OpenJDK#{version.major}U-jdk_x64_mac_hotspot_#{version.major}_#{version.minor}.tar.gz"
+  appcast "https://github.com/adoptopenjdk/openjdk#{version.major}-binaries/releases.atom"
   name 'AdoptOpenJDK Java Development Kit'
   homepage 'https://adoptopenjdk.net/'
 
-  artifact "jdk-#{version.before_comma}+#{version.after_comma}", target: "/Library/Java/JavaVirtualMachines/adoptopenjdk-#{version.before_comma}.jdk"
+  artifact "jdk-#{version.major}+#{version.minor}", target: "/Library/Java/JavaVirtualMachines/adoptopenjdk-#{version.major}.jdk"
   
   uninstall rmdir: '/Library/Java/JavaVirtualMachines'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Why change this from an `artifact` to a `pkgutil`? Assuming the end result is the same, that is a net loss.

tiiiecherle:
@vitorgalvao no specific reason for changing artifact to a pkgutil. I adopted the changes form [here](https://github.com/AdoptOpenJDK/homebrew-openjdk/blob/master/Casks/adoptopenjdk12.rb).

Feel free to change it back if you see any disadvantages. Thanks

vitorgalvao:
Please do change it back to using an `artifact`. A `pkg` can do anything, an `artifact` is just a moved file; way easier for the user to track.

vitorgalvao (changes requested):
Please do change it back to using an `artifact`. A `pkg` can do anything, an `artifact` is just a moved file; way easier for the user to track.

Casks/adoptopenjdk.rb
```diff
@@ -8,8 +8,8 @@
   name 'AdoptOpenJDK Java Development Kit'
   homepage 'https://adoptopenjdk.net/'
 
-  pkg "OpenJDK#{version.before_comma}U-jdk_x64_mac_hotspot_#{version.before_comma}_#{version.after_comma.before_colon}#{version.after_colon}.pkg"
-
+  artifact "jdk-#{version.before_comma}+#{version.after_comma}", target: "/Library/Java/JavaVirtualMachines/adoptopenjdk-#{version.before_comma}.jdk"
```
vitorgalvao: Just this won’t work. You’ve made other changes. The download can’t be a `pkg`, for example, it needs to use their new `tar.gz` (or whatever) if available.

vitorgalvao (changes requested):

tiiiecherle:
I just installed the cask locally and it is working. 

How do we make sure all "old" versions are uninstalled correctly?
After the installation I have
/Library/Java/JavaVirtualMachines/adoptopenjdk-11.0.2.jdk
/Library/Java/JavaVirtualMachines/adoptopenjdk-12.jdk
both installed.
And the old unused one is 312,6 MB.

```
java --version
openjdk 12 2019-03-19
OpenJDK Runtime Environment AdoptOpenJDK (build 12+33)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 12+33, mixed mode, sharing)
```

Casks/adoptopenjdk.rb
```diff
@@ -1,16 +1,16 @@
...
-  url "https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-#{version.before_comma}%2B#{version.after_comma}/OpenJDK11U-jdk_x64_mac_hotspot_#{version.before_comma}_#{version.after_comma}.tar.gz"
-  appcast 'https://github.com/adoptopenjdk/openjdk11-binaries/releases.atom'
+  url "https://github.com/AdoptOpenJDK/openjdk#{version.before_comma}-binaries/releases/download/jdk-#{version.before_comma}+#{version.after_comma}/OpenJDK#{version.before_comma}U-jdk_x64_mac_hotspot_#{version.before_comma}_#{version.after_comma}.tar.gz"
+  appcast "https://github.com/adoptopenjdk/openjdk#{version.before_comma}-binaries/releases.atom"
   name 'AdoptOpenJDK Java Development Kit'
   homepage 'https://adoptopenjdk.net/'
 
   artifact "jdk-#{version.before_comma}+#{version.after_comma}", target: "/Library/Java/JavaVirtualMachines/adoptopenjdk-#{version.before_comma}.jdk"
-
-  uninstall rmdir: '/Library/Java/JavaVirtualMachines'
```
vitorgalvao: This also has to be restored, and the `uninstall pkgutil` line removed.

vitorgalvao (changes requested):

Casks/adoptopenjdk.rb
tiiiecherle: o.k., done, would be nice if you could take the time and have a look at my question regarding the old versions. thanks

vitorgalvao (approved):

vitorgalvao:
If you’ve installed from this cask (`homebrew/cask/adoptopenjdk`), then the previous version will upgrade to this one. If you installed from another tap, they should be considered different casks, and thus will not.

If there’s any install confusion, try uninstalling them all or installing again, or just removing the the file `artifact` points to (an advantage of doing it this way).

tiiiecherle:
o.k., thanks

PR #62295, opened 2019-04-22
Update url for VLC player download

anteloio:
I couldn't download VLC using cask today so I checked the URL on the official website.

- Site link: https://get.videolan.org/vlc/3.0.6/macosx/vlc-3.0.6.dmg
- Download: https://ftp.free.org/mirrors/videolan/vlc/3.0.6/macosx/vlc-3.0.6.dmg

It's the first time I try to contribute. Can someone help me with making the tests pass?

### DIFF
--- a/Casks/vlc.rb
+++ b/Casks/vlc.rb
@@ -2,7 +2,7 @@ cask 'vlc' do
   version '3.0.6'
   sha256 '0c807a2e352fd6fe7e591fe01a7a53794b7d817e1e1eda3f673df06f26db27d5'
 
-  url "https://get.videolan.org/vlc/#{version}/macosx/vlc-#{version}.dmg"
+  url "https://ftp.free.org/mirrors/videolan/vlc/#{version}/macosx/vlc-#{version}.dmg"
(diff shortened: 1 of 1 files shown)


fwuensche:
It's working again. Closing it.

fwuensche:
@suschizu i've had a couple of friends reporting the same issue this week. so two questions:

1. should we update the download url for vlc?

2. if i were to do that, what are the steps i can follow to pass the tests?

core-code:
i am not sure why you would want to replace the official download URL of VLC, which is also used on their website, with a random mirror?

VLC downloads and installs through Homebrew-Cask for me without problem with the current (official!) URL just fine, if it doesn't for you then i'd suggest filing an issue...

suschizu:
@fwuensche - the official download link works too for me … I can't reproduce the problem and would also suggest to keep the official url

vitorgalvao:
We use the official download link. If the reliability of that URL is finicky, I’d recommend you contact *them* about it.

fwuensche:
@vitorgalvao fully agreed. that makes sense

PR #62413, opened 2019-04-25
Update videoproc.rb

core-code:
after 9 months being called 'videoproc' this app has been renamed back *grrr*


vitorgalvao:
For a moment I thought you meant the new name of the app was `grrr` and was confused!

PR #62416, opened 2019-04-25
Add Wasabi Wallet.app v1.1.4

zavan:

### DIFF
--- a/Casks/wasabi-wallet.rb
+++ b/Casks/wasabi-wallet.rb
@@ -1,0 +1,12 @@
+cask 'wasabi-wallet' do
+  version '1.1.4'
+  sha256 '4f2aeeec21e321099542d09504dd202d7679b209dd8244a49eb880343ea3f8b3'
+
+  # github.com/zkSNACKs/WalletWasabi was verified as official when first introduced to the cask
+  url "https://github.com/zkSNACKs/WalletWasabi/releases/download/v#{version}/Wasabi-#{version}.dmg"
+  appcast 'https://github.com/zkSNACKs/WalletWasabi/releases.atom'
+  name 'Wasabi Wallet'
+  homepage 'https://www.wasabiwallet.io/'
+
+  app 'Wasabi Wallet.app'
+end


zavan:
Pinging @nopara73 just so he knows about this.

nopara73:
Thanks for the heads up. So. Unless I don't break this: `"https://github.com/zkSNACKs/WalletWasabi/releases/download/v#{version}/Wasabi-#{version}.dmg"` we're good?

However we're going to break that at one point in the future, because I think `.pkg` is a more user friendly way to release a software for OSX, it's just we're creating the `.dmg` on Windows and Linux and that way only `.dmg` can be created, but planning to move our packaging to OSX because apple code signing is not possible on other platforms.

So, while I greatly appreciate your efforts here, I'm afraid this is not future proof :/

zavan:
@nopara73 Correct. Don't worry, It's ok if the way it's downloaded/installed is changed, it's pretty simple to adapt the formula to work with a pkg, I can do it myself once it's changed.

P.S.: Thank you for Wasabi!

nopara73:
Yes, it's not about the difficulty, it's about that these things tend to get forgotten:)

Anyway, I'll write it in my calendar to check back this issue one year from now, just in case.

vitorgalvao:
> because I think `.pkg` is a more user friendly way to release a software for OSX

I disagree. There are times when one doesn’t have a choice, but if your software can be distributed as an `app` and you make it a `pkg`, I’ll instantly question your motives. By converting it to a `pkg`, you’re going to make supporting it as a cask more of a pain.

nopara73:
You think so? I haven't researched the issue so I don't really understand what you are referencing by "questionable motives with pkg" and it being a maintenance issue, but I may very well come to the same conclusion when I try to actually do it.

Could you elaborate on them?

vitorgalvao:
> You think so?

As someone who has been dealing daily with installers for over half a decade in the context of this project, yes. But that experience might also make me an outlier.

> what you are referencing by "questionable motives with pkg"

A package is installed with administrator privileges, meaning it can do anything to the user’s system. So if your app doesn’t *need* elevated permissions, I’ll question why you’re asking for them (and if I see no reason, skip your app altogether).

> and it being a maintenance issue

`pkg` casks are more complex. They always require an `uninstall`, and necessary changes down the line are harder to detect.

Issue #62412, opened 2019-04-25
Merge menumeters and yujitach-menumeters


sbliven:
### Description of feature/enhancement

There are currently two recipes providing the MenuMeters app:
- [menumeters](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/menumeters.rb) provides version 1.8.1 from ragingmenace.com and works up to yosemite
- [yujitach-menumeters](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/yujitach-menumeters.rb) provides 1.9.3 through 1.9.7 and works on el-capitan and newer.

These packages provide the same software, but the current version is much harder to find and install due to the prefix. I propose that they be merged into a single 'menumeters' recipe, where the download location is selected based on the OS version.

### Justification

The original ragingmenace branch has been defunct since 2015, while the yujitach fork continues to be maintained and regularly released. They provide the same software and are mutually incompatible due to the non-overlapping OS restrictions. 

### Example use case

I would expect the following to work on a modern system:
```
brew cask install menumeters
```

vitorgalvao:
Agreed. https://github.com/Homebrew/homebrew-cask/pull/62422.

PR #62422, opened 2019-04-25
Make yujitach-menumeters the new menumeters

vitorgalvao:
Refs https://github.com/Homebrew/homebrew-cask/issues/62412.

The old software only works until macOS 10.10, so it’s four major releases behind. While technically a fork, this has earned the right to be the main:

```
yujitach-menumeters (added 1302 days ago)
30 days: 101 (#597)
90 days: 384 (#507)
365 days: 1,097 (#476)
```

Nobody will miss the old cask:

```
menumeters (added 1993 days ago)
30 days: n/a
90 days: n/a
365 days: 2 (#6446)
```

Even as a cask it’s superior (versioned, has appcast).


sbliven:
Nice!

Will `brew cask install yujitach-menumeters` now fail? Would that be a problem for anyone?

vitorgalvao:
> Will `brew cask install yujitach-menumeters` now fail?

Yes.

> Would that be a problem for anyone?

Yes. People who already installed it will have to uninstall the old one and reinstall the new one. We don’t yet support cask renaming.

PR #62446, opened 2019-04-25
Add Colornerd.app.

jpederson:
Adds a small Mac app called Colornerd.app (which I created) to Casks.

### DIFF
--- a/Casks/colornerd.rb
+++ b/Casks/colornerd.rb
@@ -1,0 +1,10 @@
+cask 'colornerd' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://github.com/jpederson/colornerd.mac/raw/master/Colornerd.app.zip'
+  name 'Colornerd'
+  homepage 'https://github.com/jpederson/colornerd.mac'
+
+  app 'Colornerd.app'
+end


vitorgalvao:
Welcome, and thank you, but we’ll be rejecting this on account of [not being notable enough](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/rejected_casks.md):

> The Cask is too obscure (for example, a self-submitted app from a GitHub repository that is [not notable enough](https://github.com/Homebrew/homebrew-cask/pull/28103) (under 50 stars)).

vitorgalvao:
Also, I’ll ask that next time you go through the steps in the submission template.

jpederson:
I did go through the steps, I just made the mistake of removing the checklist from the pull request (my apologies). But yeah, all the checks passed and my tests worked fine.

So I have to make this popular (somehow) before I can submit it? I've spent a lot of time on this - the original repo for the project is https://github.com/jpederson/colornerd (it's a node project, which outputs the file that this app uses as a webview - and while there's still not a lot of stars, you can see that there's been a ton of work put into it. No exceptions?

vitorgalvao:
> So I have to make this popular (somehow) before I can submit it?

Same was as every project. [Homebrew Cask is not a discoverability service](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md).

Every cask added is an extra maintenance burden; we need to make sure it’s worth, i.e., that people want it. If you’re the only one that uses this, there is no point to having it in Homebrew Cask in the first place, and having it in Homebrew Cask won’t make more people use it if they don’t want it.

It is *especially* common that new apps with few stars get abandoned, meaning we have an extra zombie cask to be another source of issues.

> I've spent a lot of time on this

Presumably so the app exists and solves the needs you built it for, I don’t see a case where you work on an app for the sole purpose of having it added to Homebrew Cask. Those would’t be a fit anyway.

> and while there's still not a lot of stars, you can see that there's been a ton of work put into it.

Again, the point is not how much work you put into it, is how much the users of Homebrew Cask would want it included.

> No exceptions?

Prolific contributors and maintainers. I.e. people we know are invested in the project and are less likely to abandon maintenance of cask they use.

jpederson:
Alright - it is new, and I understand where you're coming from. I'll continue maintaining and encouraging others to use the app. Partly, why there aren't any stars, is that I just link directly to the zip file to download the app, so when people download it, they do so directly and aren't hitting the Github page to star it. But yes, I'll keep it going and keep updating it for more months. I don't really post things without knowing I'm going to support them, but I get where you're coming from.

Thanks,
James

PR #62474, opened 2019-04-26
Fix 1password-cli 0.5.5 checksum

alext:
The current checksum doesn't match the checksum of the file published on the 1Password site. I've downloaded the file and verified the contents using their GPG key as described here: https://support.1password.com/command-line-getting-started/#set-up-the-command-line-tool.

### DIFF
--- a/Casks/1password-cli.rb
+++ b/Casks/1password-cli.rb
@@ -1,6 +1,6 @@
 cask '1password-cli' do
   version '0.5.5'
-  sha256 'e4ea329debcf991434d90728fa3cba531bce5449a08883d3530dfeb796fc3a3b'
+  sha256 '5e47ee0f1801d178818056eae9752d332ee69fbd0d96d49b3087a098b49a5db0'
 
   # cache.agilebits.com/dist/1P/op/pkg was verified as official when first introduced to the cask
   url "https://cache.agilebits.com/dist/1P/op/pkg/v#{version}/op_darwin_amd64_v#{version}.zip"


alext:
I'm not sure why Travis is seeing the old checksum. I've tested the checksum of the download from several locations (my home connection, and an EC2 instance in both us-east-1 and eu-west-2) and have consistently got the updated checksum:

```
ubuntu@ip-172-31-27-128:~$ curl -L https://cache.agilebits.com/dist/1P/op/pkg/v0.5.5/op_darwin_amd64_v0.5.5.zip | shasum -a 256 -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 3533k  100 3533k    0     0  9347k      0 --:--:-- --:--:-- --:--:-- 9347k
5e47ee0f1801d178818056eae9752d332ee69fbd0d96d49b3087a098b49a5db0  -
```

vitorgalvao:
I get the same as Travis, which is the same I get from their website.

alext:
@vitorgalvao Could you share how you verified this. I'd like to understand why I'm getting a different result (even when testing from several locations).

I'm wondering whether the inconsistency arises because homebrew has cached the download, and therefore isn't picking up the changed archive from the agilebits server.  Do you get the same checksum if you test with curl? eg:
```
 curl -s https://cache.agilebits.com/dist/1P/op/pkg/v0.5.5/op_darwin_amd64_v0.5.5.zip | shasum -a 256 -
```

vitorgalvao:
> Do you get the same checksum if you test with curl?

Yes. Exact command in your post.

alext:
That's very odd, I have no idea why it would be different for me. I wonder whether there are several servers behind that hostname, and it depends which one you get. Do you mind sharing the verbose output you get to help me diagnose why it's different.

For reference, I get the following: https://gist.github.com/alext/01ad13c963c84cda1263956f3a3d880c

alext:
@vitorgalvao I've just pushed a test project to Travis to see what the checksum is there, and it's seeing this updated checksum as well: https://travis-ci.org/alext/1password-cli-travis-sha-test/builds/525886828

Given this, and the fact that 3 other people have opened identical PRs (#62516, #62525 and #62531), suggests that the checksum has changed. Do you think you could take another look - I'm struggling to find any way to get the file with the checksum as currently recorded here.

alext:
Interestingly, this would not be the first time they've updated the checksum of an existing published binary: #48145

vitorgalvao:
> Given this, and the fact that 3 other people have opened identical PRs (#62516, #62525 and #62531), suggests that the checksum has changed.

And we might just as well get 30 people reverting it if we merge this now. This cask is popular.

```
1password-cli (added 599 days ago)
30 days: 272 (#298)
90 days: 961 (#269)
365 days: 2,582 (#271)
```

Clearly this is a problem on their end. Either the CDN’s cache hasn’t been updated all over (either on purpose or not), or a mistake is pushing different versions to different CDN location (it has happened), or one of the server has been compromised (less likely).

Either way, both checksums are both right and wrong, depending on who you ask. Given that, and that neither of us has any data on what the reason is or what change affects more people, this stays as is until Travis sees the new checksum. I still don’t.

vitorgalvao:
The way to fix this is not to keep insisting we make the change, but to clarify the situation with them.

It’s a common occurrence that when upstream (of any project) has a problem, people try to build workarounds in Homebrew Cask instead of contacting the source and fixing it for *everybody*. That behaviour is one of the reasons the world’s software feels like it’s held together by duct tape — workarounds on top of workarounds.

We prefer to fix things right. Contact the source, make them aware of the problem, let them fix it.

vitorgalvao:
I’m going to reopen this and use it as the PR to link others to. Also the one to merge when the problem is fixed on their side.

vitorgalvao:
I’m finally getting the new checksum. Let’s see Travis.

vitorgalvao:
I’m finally getting the new checksum. Let’s see Travis.

Thank you for your patience.

mickaelperrin:
II can confirm that I am unable to install this cask on a new setup and that the new sha256 is the one provided in this pull request.

image


[According to their team](https://discussions.agilebits.com/discussion/comment/503748/)

> Yes I can confirm the latest version was re-built in error when. our build servers were upgraded. The version currently being served on app-updates.agilebits.com is the "new" version and we will unfortunately be keeping it. There are steps being done to prevent this from happening in the future

vitorgalvao:
@mickaelperrin Nobody is questioning the validity of the checksum. What matters now is the right moment to update the cask, as some users (including Travis CI) are still seeing the old one.

vitorgalvao:
By the way, one of the reasons merging this now would be problematic is that if another user submits the fix with the old checksum, Travis will pass it and then automerge, meaning we’re back at square one.

mickaelperrin:
Understood, thanks for the details.

Issue #62503, opened 2019-04-27
Download failed on Cask 'java'


dogudo:
#### General troubleshooting steps

- [x] I have retried my command with `--force`.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.
- [x] I checked there are no [open issues](https://github.com/Homebrew/homebrew-cask/issues) for the same problem.
- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).

#### Description of issue

Download of Java JDK 12.0.1 using `brew cask install java` is failing in a clean installation of Mojave.

#### Command that failed

```
brew cask install java
```

#### Output of command with `--force --verbose --debug`

```
==> Cask::Installer#install
==> Printing caveats
==> Cask::Installer#fetch
==> Satisfying dependencies
==> Downloading
/usr/bin/curl -q --show-error --user-agent Homebrew/2.1.1-78-g21bb9f6\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 10.14.4\)\ curl/7.54.0 --location --silent --head --request GET https://download.java.net/java/GA/jdk12.0.1/69cfe15208a647278a19ef0990eea691/12/GPL/openjdk-12.0.1_osx-x64_bin.tar.gz
==> Downloading https://download.java.net/java/GA/jdk12.0.1/69cfe15208a647278a19ef0990eea691/12/GPL/openjdk-12.0.1_osx-x64_bin.tar.gz
/usr/bin/curl -q --show-error --user-agent Homebrew/2.1.1-78-g21bb9f6\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 10.14.4\)\ curl/7.54.0 --location --range 0-1 --dump-header - --write-out \%\{http_code\} --output /dev/null https://download.java.net/java/GA/jdk12.0.1/69cfe15208a647278a19ef0990eea691/12/GPL/openjdk-12.0.1_osx-x64_bin.tar.gz
[...]

PR #62505, opened 2019-04-27
Update java

lemeb:

### DIFF
--- a/Casks/java.rb
+++ b/Casks/java.rb
@@ -2,7 +2,6 @@ cask 'java' do
   version '12.0.1,69cfe15208a647278a19ef0990eea691'
   sha256 'cba6f42f82496f62c51fb544e243d440984d442bdc906550a30428d8be6189e5'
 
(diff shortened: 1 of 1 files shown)


lemeb:
resolves #62503

vitorgalvao:
Thank you for the fast turnaround and fixes, @lemeb!

Unfortunately, they were a bit too fast and need some modifications, pointed out in the other PRs. Specifically, they need either an upstream fix (preferred), or a new homepage that directly links to the download URL. As it is, the comment above `url` is not accurate to its rules, and we need to take that seriously, as it’s a security measure.

Thank you again. I’ll close this PR, but another one that fixes the problem with the requested solution is acceptable.

vitorgalvao (approved):

vitorgalvao:
Working now. Thank you again for the fast turnaround.

If you find a homepage for the beta, we can (re)change that cask. Virtualbox will be trickier.

lemeb:
thanks for everything, @vitorgalvao! didn’t want to break the standards of cask in any way :)

It’s very puzzling to see oracle do such a change on a saturday morning... especially since the normal webpage didn’t direct to the right url either.

PR #62562, opened 2019-04-29
Update raw-photo-processor from 1876Beta to 4.8.0,1703

suschizu:

### DIFF
--- a/Casks/raw-photo-processor.rb
+++ b/Casks/raw-photo-processor.rb
@@ -2,7 +2,7 @@ cask 'raw-photo-processor' do
   version '4.8.0,1703'
   sha256 '578b2ad21ba0a6a1704a6b80083be5096aba18a50ea394711cf9c12f8b46a538'
 
-  url 'http://www.raw-photo-processor.com/RPP/RPP64_Upd.zip'
+  url 'https://www.raw-photo-processor.com/RPP/RPP64_Upd.zip'
   appcast 'https://www.raw-photo-processor.com/rpp_updates_64.xml'
   name 'Raw Photo Processor'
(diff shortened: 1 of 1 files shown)


frihamn (commented):
Can this be reverted? It is in fact a downgrade. As already discussed, the betas are release quality and the developer makes no distinction as clarified here: https://groups.google.com/forum/#!topic/raw-photo-processor/PJyyP2JwIwI

suschizu:
oh, ok sorry for that! So 1876Beta is still the latest version?

apropostech:
Yes, I believe 1876Beta is the latest (it was published 21 March).

apropostech:
Another thought, is it allowed to add a comment in the cask regarding this to prevent it from being downgraded by error again in the future?

vitorgalvao:
> Another thought, is it allowed to add a comment in the cask regarding this to prevent it from being downgraded by error again in the future?

No.

suschizu:
thanks @vitorgalvao 👍  
@apropostech - do you know if there is a new website or appcast that we could use for this cask? This would perhaps be a good idea to prevent this cask from up/downgrading to version 4.8.0

apropostech:
Unfortunately not. I subtly raised the issue in the [group discussion](https://groups.google.com/forum/#!topic/raw-photo-processor/PJyyP2JwIwI) linked to above and mentioned homebrew. The developer has moved away from publishing updates to both the official website and appcast (possibly due to convenience) and has been releasing through the google forum (as well as http://raw-rpp.livejournal.com/ ) since 2014.

Maybe the appcast stanza should be removed though to avoid confusion on that part since it is linking to what is now 5 year old versions.

PR #62564, opened 2019-04-29
add mine shim for rubymine

glensc:

----

this adds launcher script which is typically created from `Tools` menu, but it contains version specific variables, so it's better if cask automates this.

image

did not know how to create such launchers, i.e how to add cask specific external file. so created it using heredoc with some replacements.

perhaps should use the original script somehow:
- i.e all intellij platform tools create the launcher based on something like this: https://github.com/JetBrains/intellij-community/blob/webstorm/191.6707.60/platform/platform-resources/src/launcher.py

### DIFF
--- a/Casks/rubymine.rb
+++ b/Casks/rubymine.rb
@@ -11,6 +11,125 @@ cask 'rubymine' do
 
   app 'RubyMine.app'
 
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/mine.py"
+  binary shimscript, target: 'mine'
+
+  preflight do
+    IO.write shimscript, <<~EOS
+#!/usr/bin/env python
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you, but this is too complex.

We purposefully don’t do this for Intellij casks (there’s an old closed issue on this somewhere).

glensc:
see my other questions, perhaps it's possible to make it simpler by not using `EOS`, for example remote resource?

vitorgalvao:
> perhaps it's possible to make it simpler by not using `EOS`, for example remote resource?

That doesn’t make it simpler, it just moves the complexity somewhere else.

Also, we’re not going to add a second hidden download as part of the cask. Not only would that be highly unusual, it’s just asking to be exploited.

Finally, that would require us to host the scripts, and we’re not going to do that.

If you want it that bad in the cask, consider hosting your own tap.

PR #62760, opened 2019-05-03
Update nvidia-geforce-now from 2.0.6.86 to 2.0.6.87

suschizu:

### DIFF
--- a/Casks/nvidia-geforce-now.rb
+++ b/Casks/nvidia-geforce-now.rb
@@ -1,5 +1,5 @@
 cask 'nvidia-geforce-now' do
-  version '2.0.6.86'
+  version '2.0.6.87'
   sha256 '2aa9c198646f2c32c8d12eb4bc12970613349120895e3673d3d62a7982140474'
 
   url 'https://ota-downloads.nvidia.com/ota/GeForceNOW-release_E943AA.dmg'


core-code:
@suschizu this is wrong, you've updated the version but the SHA didn't change. this still contains  2.0.6.86.

@vitorgalvao any idea why this was auto-merged? changing a version without changing the SHA  is almost always wrong - except when correcting version that was wrong before. i don't think such changes should be auto-merged, since they are likely to be incorrect and need manual inspecting to see whether they are right...

suschizu:
sorry for that - I've updated the url and the sha 
2.0.6.87 is the latest version 
https://github.com/Homebrew/homebrew-cask/pull/62765

vitorgalvao:
> changing a version without changing the SHA is almost always wrong - except when correcting version that was wrong before. i don't think such changes should be auto-merged

Agreed. cc @reitermarkus 


> additionally, i don't think 'cask-repair -b' should go ahead in this case.

`cask-repair` makes no checks of its own, except finicky methods to detect casks it knows it can’t update (`language`, `url do`) and abort early to not screw up. Checks that later on are offloaded to `audit` and `style`.

core-code:
ok good enough ;)

PR #62820, opened 2019-05-04
Update trader-workstation from 975 to 972

core-code:


core-code:
downgrading version because i've switched the cask from 'latest' to 'stable', which matches the download from the homepage, and is, i think, what we want

PR #63033, opened 2019-05-09
Update stremio

core-code:


core-code:
we were unable to use the previous appcast as it both
1.) doesn't contain the actual version number (contains-check)
2.) is too noisy to observe for changes (change-check)

PR #63080, opened 2019-05-10
Update hashbackup from 2295 to 2298

MSNexploder:

Notes:

- Download URLs changed and currently include `i386` even in the `x86_64` builds.

### DIFF
--- a/Casks/hashbackup.rb
+++ b/Casks/hashbackup.rb
@@ -1,10 +1,10 @@
 cask 'hashbackup' do
-  version '2295'
-  sha256 'e3930f5f49b742faeb5295b88fed4b42cdbe0cb2c7ddb0ce274101bfdc3203e1'
+  version '2298'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
I still see the old download link. Where are you getting this one from?

MSNexploder:
> I still see the old download link. Where are you getting this one from?

The download link on http://www.hashbackup.com/download is just a "boot binary" which downloads the actual executable on the first run. This is described in more detail in the changelog (http://www.hashbackup.com/changelog).

> previously, new releases of HB were posted to the hashbackup.com
  Download page and also to the upgrade server.  To make automated
  deployments easier, an installer or "boot binary" that does not
  change from release to release is now posted on the Download page.
  The boot binary is run just like the regular HB command, but the
  initial run will do an hb upgrade, replacing the boot binary, then
  execute the original command.  The boot binary has the public key
  built in and does RSA 4096 verification of the latest version
  downloaded from the upgrade server, just like a regular hb upgrade.

The link to the version specific download is taken from the boot binary.

PR #63342, opened 2019-05-15
Update brewcask-ci.rb => Enable checking of appcast inside the 'brew audit' check of Travis

core-code:
this nable checking of appcast inside the 'brew audit' check of Travis, which was discussed here ( https://github.com/Homebrew/brew/pull/5888 ) and which has been made possible now since this ( https://github.com/Homebrew/brew/pull/5972 ) has been merged.

### DIFF
--- a/cmd/brewcask-ci.rb
+++ b/cmd/brewcask-ci.rb
@@ -40,6 +40,7 @@ module Cask
 
           overall_success &= step "brew cask audit #{cask.token}", "audit" do
             Auditor.audit(cask, audit_download: true,
+                                audit_appcast: true,
                                 check_token_conflicts: added_cask_files.include?(path),
                                 commit_range: ENV["TRAVIS_COMMIT_RANGE"])
           end


core-code:
i hope @BrewTestBot doesn't automerge this without discussion ^^

reitermarkus:
Thanks, @core-code!

core-code:
thanks too!  lets keep under observation how well this works in practice now (cc @vitorgalvao )

core-code:
hmm i see the first failure:

` - appcast at URL 'https://www.5kplayer.com/upgrade/mac/updatecast.xml' does not contain the version number: '5.8.0'`

thats not true

core-code:
i've debugged this case. it seems the appcast of that cask really only contains'
```
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="http://www.5kplayer.com/upgrade/mac/updatecast.xml">here</a>.</p>
</body></html>
```

i guess switching the appcast to http (following the redirect) would just yield another audit error.

as a general solution, is there any way to make the `curl_output` call that is used to download the appcasts follow redirects? i think the method i used initially for the appcast download did that just fine.

core-code:
seems this should work:

`      appcast_contents, = curl_output("-L", "--max-time", "5", appcast_stanza)
`

core-code:
can someone check in the fix for the redirects or should i send a PR?

PR #63537, opened 2019-05-20
Rename adoptopenjdk.rb to openjdk.rb

gdams:
As discussed with several people the plan is move the adoptopenjdk casks over to cask-versions and rename them to be openjdk<version>. In turn we will deprecate the casks on our side with a message pointing to now download them from cask-versions.

TODO: Add a blank `adoptopenjdk.rb` cask once this has landed which caveats users towards this renamed version.


joewiz:
Could you point me to the discussion thread in question, or at least summarize the thinking? I ask because `brew cask install java` now installs OpenJDK 12:

```bash
$ java -version
openjdk version "12.0.1" 2019-04-16
OpenJDK Runtime Environment (build 12.0.1+12)
OpenJDK 64-Bit Server VM (build 12.0.1+12, mixed mode, sharing)
```

Now, with this PR, `brew cask install openjdk`, I would get AdoptOpenJDK. 

Doesn't it make sense to keep the cask for `adoptopenjdk` as that, instead of something else? 

I have no particular affection for any of these names - I'm just thinking about what would be clearest to users looking to install these projects.

gdams:
so the main reason I am suggesting that we rename the casks to `openjdk` is that it means that our existing users won't be broken. Currently, anyone who has tapped `cask-versions` and the adopt tap is getting errors because a duplicate cask named `adoptopenjdk8.rb` was landed into `cask-versions` I'm all onboard for bringing the casks over to cask versions but I don't want anymore breaking changes, hence the decision to rename to `openjdk`. AdoptOpenJDK is an openjdk vendor so I don't see any issues with the naming. Let me know your thoughts, I'm open to any solution, I just want to not make any further breaking changes if at all possible

tiiiecherle:
I don`t have anything to do with these casks or their development. But as an admin or user I would rather take a small effort of uninstalling and old one and installing a new one other than having confusing names for a long time. In my opinion anyone looking for these casks (no matter if for the first time or for maintenance) would be expecting:

java = oracle java
openjdk = openjdk
adoptopenjdk = adoptopenjdk

Everything else would be confusing or missleading.

reitermarkus:
> java = oracle java
> openjdk = openjdk
> adoptopenjdk = adoptopenjdk

I agree with that, but I'll leave the final decision here to @claui.

claui:
@reitermarkus @gdams Will take a look at it today.

gdams:
>java = oracle java
openjdk = openjdk
adoptopenjdk = adoptopenjdk

adoptopenjdk is an **openjdk** vendor. Of course homebrew can choose which openjdk vendor they wish to use for an official `openjdk` cask but previous conversations have indicated that adoptopenjdk was the logical vendor to pick.

tiiiecherle:
> > java = oracle java
> > openjdk = openjdk
> > adoptopenjdk = adoptopenjdk
> 
> adoptopenjdk is an **openjdk** vendor. Of course homebrew can choose which openjdk vendor they wish to use for an official `openjdk` cask but previous conversations have indicated that adoptopenjdk was the logical vendor to pick.

As a user I still think keeping the names of the casks as close to the "original" names as possible makes the most sense. Names were given for a reason and are found by search sites or used when talked about these topics. Name the original as the original and the vendor like the vendor. As stated above anything else than

java = oracle java
openjdk = openjdk
adoptopenjdk = adoptopenjdk

would be confusing, missleading and would even lead to wrong assumptions after an installation.
Let`s assume someone wants (the original) openjdk and just installs it by this cask-name (as not all users do a research before they install a cask, they follow a clear naming logic) and ends up with anything else. That doesn`t feel right for me.

gibfahn:
>java = oracle java
>openjdk = openjdk
>adoptopenjdk = adoptopenjdk

@tiiiecherle What does `openjdk = openjdk` mean here? There is no upstream "openjdk" AFAIK, there are Oracle OpenJDK, Adopt OpenJDK, and various others.

Right now `brew cask install java` installs `https://download.oracle.com/java/GA/jdk12.0.1/69cfe15208a647278a19ef0990eea691/12/GPL/openjdk-12.0.1_osx-x64_bin.tar.gz`.

So saying that Oracle OpenJDK == "java" seems equally confusing.

As a user I'd like `brew cask install java` to give me a permissively licensed JDK that is supported for as long as possible. With the [changes to Oracle OpenJDK licensing and releases](https://upperedge.com/oracle/using-java-heres-how-oracles-new-2019-java-se-licensing-affects-you/), I think AdoptOpenJDK sounds like a better bet.

tiiiecherle:
I´m sorry if my comment was not clear enough or missleading.
I just think finding the java version a user is looking for should be intuitive and no matter of a longer research what version is behind which cask. If a user uses a search engine he ends up with the following:

java mac = [https://www.java.com/de/download/](https://www.java.com/de/download/)
openjdk = [https://openjdk.java.net](https://openjdk.java.net)
adoptopenjdk = [https://adoptopenjdk.net](https://adoptopenjdk.net)

and I think this should be what a user is ending up with when installing the respective cask.

gibfahn:
>java mac = java.com/de/download

I don't find this intuitive, why is Oracle OpenJDK more suited to being `java` than the other distributions? Why is Oracle OpenJDK more suited to being `openjdk` than the other distributions?

>If, for some reason (licensing or anything else), java or openjdk are no longer a good choice to go with I would rather delete the casks and only keep the adoptopenjdk cask

This feels like a terrible tradeoff for me (as a random user). The most important thing for me is that `brew cask install java` gives me some usable Java. Ideally `brew cask install java@8` would give me an older version of the same Java (possible with Adopt but not with Oracle due to licensing). If I know enough to ask the question "which java do I want", then I can go find the relevant StackOverflow post. But for the average person, which Java you get is irrelevant, you just want one that works.

joewiz:
The article that @gibfahn cites as evidence that the Oracle OpenJDK has non-permissive licensing appears to actually be about Java SE, not OpenJDK. In fact, it states: 

> Businesses in this situation will either need to purchase Java SE subscription licenses or move off Java SE completely and onto an alternative platform such as Oracle OpenJDK.

So do we really have evidence that Oracle OpenJDK is not permissive? I think we have just the same evidence that Java SE is not permissive—which prompted the homebrew community to switch the `java` cask to the version with the permissive license, Oracle OpenJDK, in the first place.

In the absence of a positive reason to leave Oracle OpenJDK, I think **most users would assume, given the lineage of Java, that when they ask for `java`, they're asking for an Oracle-issued product**. That's not to detract from the quality and presumed advantages of other official OpenJDK projects. 

If, however, the homebrew community feels that the Oracle OpenJDK is non-permissive and user-hostile, then it would make sense to switch `java` to that. 

In the meantime, I'd suggest that homebrew keep:

- `java` = OpenJDK Java Development Kit
- `adoptopenjdk` = AdoptOpenJDK Java Development Kit
- `corretto` = Amazon Corretto
- `zulu` = Azul Zulu Java Standard Edition Development Kit
- `oracle-jdk` = Oracle Java Standard Edition Development Kit

... aligned with the current stable versions.

tiiiecherle:
> * `java` = OpenJDK Java Development Kit
> * `adoptopenjdk` = AdoptOpenJDK Java Development Kit
> * `corretto` = Amazon Corretto
> * `zulu` = Azul Zulu Java Standard Edition Development Kit
> * `oracle-jdk` = Oracle Java Standard Edition Development Kit

Would be ok for me. But I would not (as suggested before) mix adoptopenjdk in java or openjdk.

joewiz:
@tiiiecherle Great. And that's right, my proposal didn't conflate `adoptopenjdk` with `java` or `openjdk`:

- Requesting `adoptopenjdk` gets you AdoptOpenJDK. 
- Requesting `java` gets you Oracle OpenJDK.
- I propose that `openjdk` is undefined.

gibfahn:
>The article that @gibfahn cites as evidence that the Oracle OpenJDK has non-permissive licensing appears to actually be about Java SE, not OpenJDK. In fact, it states:

Sorry, you're right. Java SE has [a non-OSS license](https://www.oracle.com/technetwork/java/javase/overview/oracle-jdk-faqs.html), and Oracle OpenJDK only supports the latest release of Java, which is largely useless for me as most projects are migrating from Java 8 to Java 11 ([LTS releases](homebrew/cask-versions)), not on Java 12. I assume very few developers will be using the non-LTS releases and upgrading every 6 months.

gdams:
I 100% agree with @gibfahn. Nearly all Java developers work on LTS releases only (8 and gradually 11). Making the default java cask install java 12 or 13 seems like strange behaviour and is not what 90% or more of users probably want.

gdams:
@cclauss any thoughts on the above?

brodybits:
This PR to be blocking Homebrew/homebrew-cask-versions#7375. Is there any more news about this PR?

And I think a couple things should be cleared up in this discussion:

As @joewiz said, the referenced article ([[1]](https://upperedge.com/oracle/using-java-heres-how-oracles-new-2019-java-se-licensing-affects-you/)) seems to be about Oriacle Java, not OpenJDK. I think using "OpenJDK" in the link text could be a little misleading.

I think the OpenJDK license should be considered an "open source license" ([[2]](https://en.wikipedia.org/wiki/Open-source_license), [[3]](https://opensource.org/licenses)), *not* a "permissive license". OpenJDK seems to have some kind of a "copyleft" license ([GPL v2 with some kind of classpath exception ([4])](https://openjdk.java.net/legal/gplv2+ce.html)).

[1] https://upperedge.com/oracle/using-java-heres-how-oracles-new-2019-java-se-licensing-affects-you/
[2] https://en.wikipedia.org/wiki/Open-source_license
[3] https://opensource.org/licenses
[4] https://openjdk.java.net/legal/gplv2+ce.html

claui:
Thanks y’all for putting up with the delay. I’m at a conference right now; will take care of the PRs when I’m back.

robertleeblairjr:
I realize this has been a topic discussed for a while and I've not read over all of them. But, I would think it would be easier to differentiate between cask names for jdk distributions. Perhaps it would be helpful to use cask names referring to the other distributions as just builds of openjdk.

'{distribution identifier}-openjdk{# for any development, non-latest stable}'
Examples: zulu-openjdk11, oracle-openjdk, corretto-openjdk8, etc.

The homebrew/cask tap could just host the latest, stable which is version 12 for oracle-jdk, zulu-openjdk, adopt-openjdk, and so on to include those future development versions when transitioned to latest,stable. Any other dependencies for versions other than latest,stable (LTS, development, beta) could be informed in the caveats to the homebrew/cask-versions tap or to the distributions tap, if they have one, such as adopt-openjdk.

joewiz:
@robertleeblairjr’s proposal sounds reasonable - and particularly because no preference is shown for any particular distribution. Agreed also that the latest stable should be the default, with others (LTS still active, etc.) could be handled through versions. The only question is about the fate of the `java` cask - so we pick a sane default like Adopt, or delete the cask entirely?

gdams:
> The only question is about the fate of the java cask - so we pick a sane default like Adopt, or delete the cask entirely?

There is surely some value to having a `Java` cask.  I haven't personally seen the download stats but I would assume that a pretty large number of homebrew users wanting "Java" will try a `brew cask install java` before anything else. 

I'm personally pro using AdoptOpenJDK as the sane default here, SDKMAN has chosen Adopt as their default (https://twitter.com/sdkman_/status/1150103494097100811)

robertleeblairjr:
> @robertleeblairjr’s proposal sounds reasonable - and particularly because no preference is shown for any particular distribution. Agreed also that the latest stable should be the default, with others (LTS still active, etc.) could be handled through versions. The only question is about the fate of the `java` cask - so we pick a sane default like Adopt, or delete the cask entirely?

Since oracle commercial/non-commercial are the most commonly used then just having 'java' -> 'oracle-openjdk' because the non-GPL is 'oracle-jdk'. The 'java' cask could be renamed but I understand @gdams argument that homebrew users have the natural affinity to associate java to generally refer to the JRE, JDK, or both and to install without referring to the cask info. Plus, I'm sure there are an overwhelming number of casks referring to the 'java' requirements for installation. I would think they should be able to have a crawler to go through all casks and change the references of 'java' requirements strings appropriately for name changes. I would think that those who require a JDK for development purposes will probably know to search for any names with a string 'jdk'. Dependencies requiring anything other than latest,stable could have the caveat and referral to the other homebrew taps which host them. Avoiding the conflict with multiple taps whether from the distributions, individual, etc. is probably the most difficult challenge. So, may be removing any casks whereby the distribution for openjdk has a tap which I believe is just adopt-openjdk right now. Other project taps that contain those distributions may not likely have the same naming scheme but perhaps homebrew maintainers could have a policy of precedence for naming casks starting with a higher order for the primary repositories of homebrew and a non-official-homebrew/tap that has a conflict would be a PR for the maintainer of that tap to rename because of conflicts. It would be nice to have a tool with the feature of scanning a tap that are created/added/existing for name conflicts. Although, this may appear to be controlling the project; sometimes it's necessary when the size of it has overwhelmingly grown to be extremely difficult to manage. Especially with the naming of thousands in one tap, nevermind, many other taps added by users later.

ran-dall:
> @robertleeblairjr’s proposal sounds reasonable - and particularly because no preference is shown for any particular distribution. 
@joewiz 👉I concur with you and @robertleeblairjr; I also feel HBC should try to avoid showing preference, whenever possible.

robertleeblairjr:
> > @robertleeblairjr’s proposal sounds reasonable - and particularly because no preference is shown for any particular distribution.
> 
> @joewiz 👉I concur with you and @robertleeblairjr; I also feel HBC should try to avoid showing preference, whenever possible.

Thanks. Any depends :java requirements could be referenced to a different distributor or binary in the future if you allow non-preferential treatment that would reduce/eliminate breaking the majority of casks. A more uniform naming scheme of these casks is probably needed for the official taps. Would love to see a tool implemented by homebrew which could scan taps for the naming conflict which could have any maintainer wishing to fork/create a tap to do the same which would give them the opportunity to address it. Maybe an order of precedence for the naming schemes could be implemented, as well. I'm not just referring to jdk related casks. Not sure how doable any of these ideas are though. I'm quite new to teaching myself sysadmin, programming, and other computer related tasks since I only have a small amount of free time to indulge in it.

joewiz:
So let me summarize what I think we're converging on here:

## Naming

With the profusion of OpenJDK distributions, a common naming convention is needed to bring some order and help users find OpenJDK distros available via homebrew. 

Every OpenJDK distribution should be named according to the formula:

    {distribution identifier}-openjdk{# for any development, non-latest stable}

Thus: 

| Distribution | Current Name | New Name (Current Stable Version) | Alternate Versions |
| - | - | - | - |
| Oracle | `java` | `oracle-openjdk` (12) | none |
| AdoptOpenJDK | `adoptopenjdk` | `adopt-openjdk` (12) | `8` |
| Corretto | `corretto` | `corretto-openjdk` (11) | `8` |
| Zulu | `zulu` | `zulu-openjdk` (12) | `7, 8, 11` |

## A sane default for `java`

Many casks already declare a dependency on `java`. It's safe to assume that applications with such a declaration and most users looking to install Java **assume an LTS build**—thus ruling out Oracle's OpenJDK, which is *only* available in the non-LTS version 12. Thus, the homebrew community must select a different default for its users. 

The best choice of a sane default is AdoptOpenJDK. 

The best choice of version is the latest, LTS build. 

Thus, `java` => `adopt-openjdk8`.

## Locations

For the distribution-specific casks, the homebrew/cask tap should just host the latest, stable version:  version 12 for oracle-openjdk, zulu-openjdk, and adopt-openjdk, and version 11 for corretto. 

As future development versions achieve "latest, stable" status, they will replace the version in homebrew/cask. 

For versions other than latest, stable (such as LTS, development, beta), these should go in the homebrew/cask-versions tap. 

(Above, @robertleeblairjr noted: "dependencies for versions other than latest, stable (LTS, development, beta) could be informed in the caveats to the homebrew/cask-versions tap or to the distributions tap, if they have one, such as adopt-openjdk." I mention this because I'm not entirely familiar with caveats, but this sounds like an important point.)

## Non-OpenJDK JDK

See the `oracle-jdk` cask, which installs Java SE 12 and assumes you have agreed to Oracle's license agreement. Presumably this will be used only by casks/users for whom an OpenJDK distribution will not suffice.

gdams:
This looks great but to clarify the latest LTS version is JDK11 so the java cask should probably point to adopt-openjdk11

davidghiurco:
Hi everybody. Coming here from google/stackoverflow

> Many casks already declare a dependency on `java`. It's safe to assume that applications with such a declaration and most users looking to install Java **assume an LTS build**—thus ruling out Oracle's OpenJDK, which is _only_ available in the non-LTS version 12. Thus, the homebrew community must select a different default for its users.

I'm trying to wrap my head around this whole Oracle/OpenJDK/AdoptOpenJDK mess. I was using Oracle JDK8 until brew failed to upgrade it because the tap got deleted. I read about it, and understood that Oracle changed its licensing after April 2019.

However, I'm confused why OpenJDK8 is not available. Can you provide source that indicates OpenJDK is not available except for the latest major release of Java? That's quite unexpected news for me.

I'm trying to understand if I should stay away from "openjdk" completely and just go for adoptopenjdk but there's a lot of mangled confusion between those 2 terms as far as homebrew goes.

vitorgalvao:
I’ll ask that unless you are already familiar with the situation and have something important to add, do not comment further. As you can see, this issue is delayed. @claui is in charge of it but hasn’t been able to make the time; the last thing we need is more messages for her to have to parse and deal with.

This is not the issue for asking for clarifications regarding the java mess, it’s an issue to decide how that mess will be interpreted in homebrew-cask.

claui:
Seconding all points made by @vitorgalvao. Thanks to all users and maintainers for putting up with the delay and the inconvenience that arises. I’m perfectly aware of it, and I’m sorry.

I’m still committed (and trying my best) to understand and sort out the mess very soon.

In the meantime, I’d like to remind everyone of a quick workaround for the `brew cleanup` conflict: remove the cached Java packages in `~/Library/Caches/Homebrew/downloads/` by hand whenever the error message occurs.

pron:
@joewiz writes:

> The best choice of a sane default is AdoptOpenJDK.

It may well become the best default at some point in the future, but at the moment, Adopt is one of the only builds that do not run the Java TCK, and therefore their builds are not certified as Java, and are less tested than other builds. In fact, because of this, Adopt is not even allowed to use the name Java or call their builds a JDK, and having the name `java` used for the cask, let alone the default one, is problematic.

FranklinYu:
> Oracle OpenJDK only supports the latest release of Java

@gibfahn Is there any source for this claim? I only found [this Reddit post](https://www.reddit.com/r/java/comments/bs1rr7/openjdk_11_download_i_only_find_oracle_jdk/) suggesting that OpenJDK 11 has been dropped, but I can't find an exact statement from Oracle saying “we won’d release future OpenJDK 11 build, so go finding AdoptOpenJDK or other vendors”.

avarun42:
Any update on this? The proposed naming scheme by @joewiz makes perfect sense to me, and I'd love to see it implemented so I don't have to tap `AdoptOpenJdk/openjdk` just to install Java 11.

pron:
> The proposed naming scheme by @joewiz makes perfect sense to me

Whether it makes sense to you or not, AdoptOpenJDK's binaries are not allowed at this time to be referred to as "Java" since they currently don't pass compliance.

avarun42:
We get it, you work at Oracle. The community doesn't really care what Oracle thinks about OpenJDK or the "standards" you're attempting to enforce — it's clearly standardized around AdoptOpenJDK. I'm sure Oracle doesn't appreciate you brigading this thread on their behalf.

cclauss:
Is it possible resolve the conflicting file?

pron:
@avarun42 I think that the community cares very much what Oracle thinks about OpenJDK given that Oracle leads the project and contributes ~90% of the work, but that's neither here nor there. I am merely pointing out that Adopt cannot, to the best of my knowledge, use the name Java (feel free to contact them and ask them for yourself), I am not doing this on anyone's behalf, and I am not "brigading" on this thread any more than you are (and I'm actually an OpenJDK contributor).

vitorgalvao:
A few notes:

1. I have no knowledge of what’s currently holding up @claui, but I trust her to have her reasons and that she’ll get to this when she’s able to. It’s a decision that should remain in effect for quite a while, so it requires some time.
2. The more comments you make, the more she has to read, thus the later she can take action. So if you must comment, make it good, short, and relevant to the immediate problem. Before posting, think: “will this comment help @claui make the best decision regarding HBC’s Java kerfuffle?”
3. It hasn’t reached that point (thank you), but if an argument breaks out I will lock the issue.

OranguTech:
My preference would be that `brew cask install java` give an informational "caveat" message listing out the different packages, with a URL pointing to more detailed info.

reitermarkus:
`openjdk@11`, `openjdk@12` and `openjdk@13`/`openjdk` are now available as formulae. 

We will remove the `java11` and `java` casks in the near future in favour of these and start migrating other formulae to depend on these.

tiiiecherle:
> `openjdk@11`, `openjdk@12` and `openjdk@13`/`openjdk` are now available as formulae.
> 
> We will remove the `java11` and `java` casks in the near future in favour of these and start migrating other formulae to depend on these.

@reitermarkus Thanks for the heads-up. What about adoptopenjdk as cask? Will it also be discontinued? Thanks and have a nice weekend

joewiz:
@claui @vitorgalvao @reitermarkus Yes, sounds like this PR should be closed, then, right?

For others looking for more context, see the discussion at https://github.com/Homebrew/homebrew-core/pull/45935.

reitermarkus:
We will keep the `adoptopenjdk` cask for now. Ideally I think it should also eventually be converted to a formula since it is open source and CLI-only.

PR #63582, opened 2019-05-21
Update tor-browser to 8.5

miccal:
https://dist.torproject.org/torbrowser/8.5/sha256sums-signed-build.txt


core-code:
i can not find any mention of this here
https://www.torproject.org/download/
or here
https://blog.torproject.org/category/tags/tor-browser

can we hold off merging these things until they are really, officially released?

vitorgalvao:
> i can not find any mention of this here
> https://www.torproject.org/download/

That’s where I checked. It shows up there, for me.

core-code:
actually i checked about 10 times today but it was stuck at 8.0.9, but i see the 8.5 there now - sorry for the noise then.

PR #63628, opened 2019-05-21
add "real" version number to prudent.rb

suschizu:
this app has a version - could we add it to "initial"?


core-code:
wasn't there a rule that only github projects with 50 stars are included? i see just one...

vitorgalvao:
> i see just one...

I saw the same before merging, but that repo has no code, they’re using it just for the hosting, so I decided to go with the “let’s see and maybe purge” method.

core-code:
sounds fair ;-)

PR #63752, opened 2019-05-23
Update changed github project name from datproject/ to  dat-land/ in dat.rb

suschizu:


vitorgalvao:
Great work!

Did you develop a new script to catch these redirects?

suschizu:
thanks! 
no script - just regex 
thanks to @core-code :)

PR #63915, opened 2019-05-27
Update j from 806 to 807

00dani:

### DIFF
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -13,7 +13,7 @@ cask 'j' do
 
   installer script: {
                       executable: "j64-#{version}/bin/jconsole",
-                      args: ['-js', "exit install'qtide'"]
+                      args:       ['-js', "exit install'qtide'"],
                     }
 
   # target names according to readme.txt
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -1,6 +1,6 @@
 cask 'j' do
-  version '806'
-  sha256 '29df4c491b50f5cfef73a2f4a6e89d18bc514a45851a2766d617746f615b934c'
+  version '807'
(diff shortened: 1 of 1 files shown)


00dani:
`cask-repair` *seemed* to work entirely automatically, but the `updatejqt.sh` install script from the previous version was removed in this version for some reason. So I've patched that issue manually. It installs fine now. 🐱

vitorgalvao:
> `cask-repair` _seemed_ to work entirely automatically, but the `updatejqt.sh` install script from the previous version was removed in this version for some reason.

That could only be detected by performing an installation (and even then, there’s only so much a bash script can do). If I made `cask-repair` install the cask every time it updates it, I’d get a lot of confused and angry people. That would definitely not be worth it for the minority of cases a tweak is needed (which I’d bet is close to or even below 1%).

Casks/j.rb
```diff
@@ -11,7 +11,10 @@
     app "j64-#{version}/#{a}.app"
   end
 
-  installer script: "j64-#{version}/updatejqt.sh"
+  installer script: {
+                      executable: "j64-#{version}/bin/jconsole",
+                      args: ['-js', "exit install'qtide'"]
```
vitorgalvao: Does it work instead as:

```suggestion
                      args: ['-js', 'exit install"qtide"']
```

00dani: Unfortuntely no. Strings in J can only use single quotes, since `"` is instead an operator.

00dani:
> That could only be detected by performing an installation (and even then, there’s only so much a bash script can do). 

Oh, sure, absolutely! It's quite reasonable for `cask-repair` to work the way it does. I'm mostly just confused that they removed that script from the upstream package, even though you still need to do the things it does to install Jqt properly.

vitorgalvao (approved):

PR #64012, opened 2019-05-28
Add depends on sierra to apple-juice.rb

suschizu:


suschizu:
The minimum system version for this app is 10.12 - in this case a depends_on stanza seems to be suitable

core-code:
> Please provide a better issue/pull request title and/or description!

who are you? ;-> i think the title is quite good ;)

PR #64050, opened 2019-05-29
Update skype from 8.45.0.43 to 8.46.0.60

core-code:


core-code:
autoupdate does not have it yet (they do staged releases), so the appcast check will expectedly fail

however, this version is the default download from the homepage right now

PR #64062, opened 2019-05-29
Update eudic from 3.8.9 to 3.9.0

suschizu:


suschizu:
the new version isn't yet listed in the appcast - but the appcast looks well-maintained - I think they will update it in the near future …

PR #64080, opened 2019-05-29
Update musescore.rb from 3.0.5.21343 to 3.1.0.22425

suschizu:
+ update url stanza

### DIFF
--- a/Casks/musescore.rb
+++ b/Casks/musescore.rb
@@ -2,8 +2,8 @@ cask 'musescore' do
   version '3.1.0.22425'
   sha256 '5c5722a429223f320507008aa119f48551e712f1f26cfdf700772b8f3acb58b1'
 
-  # musescore.com was verified as official when first introduced to the cask
-  url "https://download.musescore.com/releases/MuseScore-#{version.major_minor}/MuseScore-#{version.major_minor_patch}.dmg"
+  # https://github.com/musescore/MuseScore was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


Casks/musescore.rb
```diff
@@ -1,9 +1,9 @@
 cask 'musescore' do
-  version '3.0.5.21343'
-  sha256 'be052c6f7fe782cc7a24cb319be9b9b6c5f8661aac328f36908ce97d50ecafac'
+  version '3.1.0.22425'
+  sha256 '5c5722a429223f320507008aa119f48551e712f1f26cfdf700772b8f3acb58b1'
 
-  # musescore.com was verified as official when first introduced to the cask
-  url "https://download.musescore.com/releases/MuseScore-#{version.major_minor_patch}/MuseScore-#{version.major_minor_patch}.dmg"
+  # github.com/musescore/MuseScore was verified as official when first introduced to the cask
+  url "https://github.com/musescore/MuseScore/releases/download/v#{version.major_minor}/MuseScore-#{version.major_minor_patch}.dmg"
```
vitorgalvao: I’m not finding the link to this, from the homepage.

suschizu: I took the link from the appcast because it contains the version-number: 
https://sparkle.musescore.org/stable/3/macos/appcast.xml

the download link from the website doesn't contain the version-number - would you prefer this one instead: 
https://musescore.org/en/download/musescore.dmg

vitorgalvao: Link is fine as is; we just need a way to get to it from the official domain so users can audit it.

PR #64097, opened 2019-05-29
Update frhelper from 3.8.9,2019-01-15 to 3.9.0

suschizu:


suschizu:
frhelper and eudic (https://github.com/Homebrew/homebrew-cask/pull/64062) seem to be the same app with just different names (and BundleIDs) - both appcasts seem to contain the same content:
https://www.francochinois.com/update/frhelper_mac.xml
https://www.eudic.net/update/eudic_mac.xml

so also here is the appcast behind but looks well-maintained. (see: https://github.com/Homebrew/homebrew-cask/pull/64062)) 

The download here contains version 3.9.0

PR #64154, opened 2019-05-30
add depends_on sierra stanza to enpass.rb

suschizu:


suschizu:
> appcast at URL 'https://dl.sinew.in/mac/package/appcast.xml' does not contain the version number: '6.1.0'
that's true but hopefully the new version will be added soon …

PR #64156, opened 2019-05-30
Update dashlane from 6.1914.0.19481 to 6.1918.1.20296

core-code:

### DIFF
--- a/Casks/dashlane.rb
+++ b/Casks/dashlane.rb
@@ -1,9 +1,9 @@
 cask 'dashlane' do
-  version '6.1918.1.20296'
-  sha256 'c80a1f6a1934aad9792ef63f8c8d7249fda9781e736cba087041f19693d6662b'
+  version '6.1922.0.20633'
+  sha256 'af0a59b059d2ec8f56bf4ffd036002189fae938fc80913d6eabad33a90f884cb'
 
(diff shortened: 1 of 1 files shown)


core-code:
actually not sure about that warning, would have to debug it

Casks/dashlane.rb
```diff
@@ -1,8 +1,8 @@
 cask 'dashlane' do
-  version '6.1914.0.19481'
```
vitorgalvao: I see this version in the appcast.

core-code: thats weird, i am certainly seeing '6.1918.1.20296' in the appcast every time i check. are you using the URL from the cask?

core-code: btw the download version from the homepage is already even newer, as can be seen in this config file:

https://ws1.dashlane.com/5/binaries/query?format=json&os=OS_X_10_14_5&currentSoftwareVersion=6.1850.0.16084&partner=&target=archive&platform=launcher_macosx

vitorgalvao: > are you using the URL from the cask?

Yes.



> btw the download version from the homepage is already even newer, as can be seen in this config file

Any objection to replacing the appcast with that? The link can be further shortened: `https://ws1.dashlane.com/5/binaries/query?format=json&os=OS_X_10_14_5&target=archive&platform=launcher_macosx`.

core-code: no objections from my side, though i still find it weird we are seeing different results from the appcast. must be some form of staging.

PR #64212, opened 2019-05-31
Add Localdev by Pantheon

ran-dall:
👉 Pantheon's new tool; which is in pre-release but I figure it's popular enough so I'd submit it here before adding it to my repo. 👉 Feel free to decline.

### DIFF
--- a/Casks/localdev.rb
+++ b/Casks/localdev.rb
@@ -2,7 +2,7 @@ cask 'localdev' do
   version '0.4.0'
   sha256 '5ce8b74ab6af8214abb2cd8a4d56d3ed431f76cfa9083ba1606aa25e947af96a'
 
-  # pantheon-localdev.s3.amazonaws.com/ was verified as official when first introduced to the cask
+  # pantheon-localdev.s3.amazonaws.com was verified as official when first introduced to the cask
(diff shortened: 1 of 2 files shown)


vitorgalvao (approved):

ran-dall:
@vitorgalvao Thanks for cleaning that up and point that stuff out; espically with the `appcast`, wasn't too sure to add it! 🙏

Question, wouldn't this be better as `pantheon-localdev` cause `localdev` just doesn't sit right...a little too ambigious IMO.

vitorgalvao:
> Question, wouldn't this be better as `pantheon-localdev` cause `localdev` just doesn't sit right...a little too ambigious IMO.

If we don’t follow the token reference now, we’ll have a duplicate when the next person comes along and does. If it ever becomes ambiguous, we also have rules for that (adding a prefix on the second one that comes along, unless it’s insanely more popular).

ran-dall:
> If we don’t follow the token reference now, we’ll have a duplicate when the next person comes along and does. If it ever becomes ambiguous, we also have rules for that (adding a prefix on the second one that comes along, unless it’s insanely more popular).

@vitorgalvao Make sense, didn't think of it that way. ✌️ Thanks!

PR #64233, opened 2019-06-01
Update amethyst from 0.13.0 to 0.14.0

suschizu:


core-code:
its on github.
https://github.com/ianyh/Amethyst/releases
 we could consider using the github atom feed as appcast

suschizu:
ok - perhaps the better solution for the appcast here because it's generated automatically

vitorgalvao:
> perhaps the better solution for the appcast here because it's generated automatically

That the reason we prefer github appcasts if they’re available.

PR #64297, opened 2019-06-02
Update Suspicious Package to 3.5

Frizlab:

⚠️ Important:
- I removed the `auto_updates` property as the app has a mechanism to check for new version, but not the mechanism to automatically install the new version;
- Maybe the cask should be marked as without a version (`:latest`) and no check for the shasum as the URL for downloading the dmg does not depend on the version of the app. What do you think?


vitorgalvao:
> I removed the `auto_updates` property as the app has a mechanism to check for new version, but not the mechanism to automatically install the new version;

Thank you. That is the correct interpretation.

> Maybe the cask should be marked as without a version (`:latest`) and no check for the shasum as the URL for downloading the dmg does not depend on the version of the app. What do you think?

But it does have an appcast, thus checking for a new version is automated, thus it remains versioned.

PR #64320, opened 2019-06-03
Update chromium from 665421 to 665546

ahousseini:


Casks/chromium.rb
```diff
@@ -1,6 +1,6 @@
 cask 'chromium' do
-  version '665421'
-  sha256 '959d9c13142e9a8ef2d373ecc19393a7e22620614eb30e852afc9f8ee1ea8c1c'
+  version '665546'
```
vitorgalvao: Appcast shows `665556`.

ahousseini: It feels like this snapshots are updated all 10 minutes :(

PR #64375, opened 2019-06-04
Update expressvpn from 7.2.0.5711 to 7.3.0.5989

ran-dall:

### DIFF
--- a/Casks/expressvpn.rb
+++ b/Casks/expressvpn.rb
@@ -1,6 +1,6 @@
 cask 'expressvpn' do
-  version '7.2.0.5711'
-  sha256 'e8a2eb4370d5c220d4f7a52b01f7d047ed835c521f7b7f75495bf02cf24b71f3'
+  version '7.3.0.5989'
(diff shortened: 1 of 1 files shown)


Casks/expressvpn.rb
```diff
@@ -1,6 +1,6 @@
 cask 'expressvpn' do
-  version '7.2.0.5711'
```
vitorgalvao: Homepage still shows this version.

suschizu: I also just found version 7.2.0.5711 to download
but perhaps we could change the homepage stanza to :
https://www.expressvpn.xyz/vpn-software/vpn-mac/ 
😬

ran-dall: Let's just keep it as it is cause I also can't find this particular release but it would seem that the system did pull this correctly according to the logs.

suschizu: ok 👍  but strange because the download-link contains the version … 
I've updated the product site -> https://github.com/Homebrew/homebrew-cask/pull/64437

ran-dall: >  strange because the download-link contains the version

@suschizu 👉 Yea it is, I'm going to look into it.

PR #64394, opened 2019-06-05
Update graphicconverter from 10.7.1,3460 to 11.0,4003

core-code:


core-code:
the new feed for version 11 is not yet fully operational and the versioned ZIP files are not yet online, therefore i had to switch the URL from the feed to the homepage download.

will switch it back as soon as it is possible

PR #64427, opened 2019-06-05
Update a-better-finder-rename from 10.40 to 10.42

ran-dall:

### DIFF
--- a/Casks/a-better-finder-rename.rb
+++ b/Casks/a-better-finder-rename.rb
@@ -1,6 +1,6 @@
 cask 'a-better-finder-rename' do
-  version '10.40'
-  sha256 :no_check # required as upstream randomly uses two different downloads for a/b testing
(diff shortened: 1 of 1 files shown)


Casks/a-better-finder-rename.rb
```diff
@@ -1,6 +1,6 @@
 cask 'a-better-finder-rename' do
-  version '10.40'
-  sha256 :no_check # required as upstream randomly uses two different downloads for a/b testing
```
vitorgalvao: Does this no longer happen?

ran-dall: @vitorgalvao 👉 I only updated it. I thought it was weird too when I saw it.

vitorgalvao: Let’s see what happens.

PR #63979, opened 2019-05-28
Delete couleurs.rb

core-code:
they have been offline for 2 months (expired SSL). whats our official threshold how long a project needs to be offline to be removed?

PR #64484, opened 2019-06-06
Revert "Delete couleurs.rb"

core-code:
Reverts Homebrew/homebrew-cask#63979

Project is back up, not sure if we want to re-add it, based on popularity


vitorgalvao:
Since someone asked for it and it’s back up, let’s give it another chance.

PR #64597, opened 2019-06-09
Update soulver.rb from 2.6.9-6055 to 3.0.0-19

suschizu:
soulver 2 still exists on the website - but 3 is the latest and out of Beta:
https://soulver.app/

### DIFF
--- a/Casks/soulver.rb
+++ b/Casks/soulver.rb
@@ -1,16 +1,25 @@
 cask 'soulver' do
-  version '2.6.9-6055'
-  sha256 'e8f96895e43d28177f458404bfc082ebc940c66ce6f425cdceeedab9389272bf'
+  if MacOS.version <= :mojave
+    version '2.6.9-6055'
+    sha256 'e8f96895e43d28177f458404bfc082ebc940c66ce6f425cdceeedab9389272bf'
 
-  url "https://www.acqualia.com/files/sparkle/soulver_#{version}.zip"
-  appcast "https://www.acqualia.com/soulver/appcast/soulver#{version.major}.xml"
+    url "https://www.acqualia.com/files/sparkle/soulver_#{version}.zip"
+    appcast "https://www.acqualia.com/soulver/appcast/soulver#{version.major}.xml"
+    homepage 'https://www.acqualia.com/soulver/'
+  else
+    version '3.0.0-19'
+    sha256 'ecaca54c6ec8fb148301a194237e043b439646eeeb48a87d4a17c869c8c9c8b2'
+  end
+
+  url "https://soulver.app/mac/sparkle/soulver-#{version}.zip"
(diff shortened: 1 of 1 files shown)


Casks/soulver.rb
```diff
@@ -1,16 +1,25 @@
...
+  if MacOS.version <= :mojave
+    version '2.6.9-6055'
+    sha256 'e8f96895e43d28177f458404bfc082ebc940c66ce6f425cdceeedab9389272bf'
 
-  url "https://www.acqualia.com/files/sparkle/soulver_#{version}.zip"
-  appcast "https://www.acqualia.com/soulver/appcast/soulver#{version.major}.xml"
+    url "https://www.acqualia.com/files/sparkle/soulver_#{version}.zip"
+    appcast "https://www.acqualia.com/soulver/appcast/soulver#{version.major}.xml"
+    homepage 'https://www.acqualia.com/soulver/'
+  else
```
vitorgalvao: Considering the conditional above, this `else` will only trigger on Catalina.

vitorgalvao: Website suggest this works on 10.14.4 (High Sierra) and above.

suschizu: uff sorry - yes that was a mistake 
I just noticed that version 2 supports mojave too … should we  keep version 2 then at all?
https://www.acqualia.com/soulver/

vitorgalvao: Since bother versions overlap, we should remove the conditional and keep only version 3. Version 2 will be added to [homebrew/cask-versions](https://github.com/Homebrew/homebrew-cask-versions) by someone who wants and submits it.

suschizu: ok thanks :) 
just the Mojave-error left 
`Satisfying dependencies
Cask soulver depends on macOS release >= 10.14, but you are running release 10.13.`

suschizu:
@vitorgalvao - is there another problem I have overseen in soulver?

vitorgalvao:
> is there another problem I have overseen in soulver?

Nope. All good.

Kmutahar:
> > is there another problem I have overseen in soulver?
> 
> Nope. All good.

So two things 
1. The zap trash will need to be changed since this is a completely new app 
    ```
    ~/Library/Application Support/Soulver 3
    ~/Library/Application Support/app.soulver.mac
    ~/Library/Preferences/app.soulver.mac.plist
    ```
1. @vitorgalvao  I'm not sure should a 'Soulver 2' be added to homebrew-cask-versions or should I just use my own tap?

vitorgalvao:
> I'm not sure should a 'Soulver 2' be added to homebrew-cask-versions

Yes.

suschizu:
@Kmutahar - you're right - I forgot it
https://github.com/Homebrew/homebrew-cask/pull/64700

PR #64639, opened 2019-06-10
Update luminance-hdr.rb from 2.5.1_1 to 2.6.0

suschizu:
+ update url stanza

### DIFF
--- a/Casks/luminance-hdr.rb
+++ b/Casks/luminance-hdr.rb
@@ -1,9 +1,12 @@
 cask 'luminance-hdr' do
-  version '2.5.1_1'
-  sha256 'be4ae6c1110bbac4fd57358e1a0489871ba143f4f295a6d07f95be9aaecb4d18'
+  version '2.6.0'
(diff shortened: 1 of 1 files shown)


suschizu:
`audit for luminance-hdr: warning
 - SourceForge URL format incorrect. See https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#sourceforgeosdn-urls`

the download-url is not sourceforge-url-conform that's true but I can't find another one
website download: http://qtpfsgui.sourceforge.net/?page_id=10
appcast download : https://sourceforge.net/projects/qtpfsgui/rss
sourcforge project download: https://sourceforge.net/projects/qtpfsgui/files/luminance/2.6.0/
it's everywhere the same url

core-code:
the URL needs to be

`  url "https://downloads.sourceforge.net/qtpfsgui/luminance/LuminanceHDR_#{version}-Qt5.13.dmg"
`

suschizu:
🤦🏻‍♀️ 
thanks 😙

PR #64697, opened 2019-06-11
Update skype from 8.46.0.60 to 8.47.0.59

core-code:


core-code:
travis will fail since its not yet in the appcast but it IS the default download from the homepage:

https://macupdater.net/cgi-bin/check_urls/check_url_redirect.cgi?url=https://go.skype.com/skype.download&user_agent=Mozilla/5.0%20(Macintosh;%20Intel%20Mac%20OS%20X%2010_14_4)

ideally we'd have two appcasts, and only either of them would need to match ;)

PR #64710, opened 2019-06-12
Update yinxiangbiji.rb from 9.0.5_458014 to 9.0.7_458129

suschizu:
cask-repair wasn't working


core-code:
i can not reproduce the cask error, '9.0.7_458129' is indeed in the appcast

suschizu:
@core-code: me too 
curl doesn't have a problem with that download - just travis …

PR #64754, opened 2019-06-12
Update musicbrainz-picard appcast stanza

core-code:
Update musicbrainz-picard appcast stanza


core-code:
rationale behind a few appcast changes today:

to keep casks with appcasts that are unstructured (non-sparkle, non-github) up-to-date we have two checks:
• change check, where we look through the difference each time the appcast changes and
• contains check, if the version number is not contained any longer in the appcast, it means there must be a new version

if either of these checks works, we are good. however, we have to disable the 'change' check for noisy HTML sizes that very often have large changes without any new version being released. if the 'contains' check doesn't work either, because we've linked a release-notes page that also contains old versions, we are basically fucked as one check is disabled and the other one wont work. 

so, if we have to use noisy html appcasts, better use a page that only contains the current, but not previous versions....

vitorgalvao:
> for all casks where it is possible to use some filtering on our end to get the 'contains' check working, i'll do this instead to minimize noise/work here.

Don’t mind the noise. These changes are better.

PR #64765, opened 2019-06-12
Update mblock from version 3.4.11 to version 5.0.1

core-code:
Version 3.x is still available on their download page, but this should probably go to cask-versions if anyone wants that


core-code:
they seem to have issues with their server, lets wait a while

core-code:
their server is a mess. i verified the SHA locally

core-code:
the only remaining error is about the appcast - can't reproduce that ;/

PR #64801, opened 2019-06-13
add depends on sierra to krisp.rb

suschizu:
That's because losers look stuff up while the rest of us are carp'en all them 'diems


vitorgalvao:
@suschizu Is there a context to that message? I’m guessing you may have tried to use a text snippet and it expanded your clipboard instead?

suschizu:
@vitorgalvao - sorry no that was just nonsense

vitorgalvao:
As long as you’re fine, no worries! Feel free to post whatever message you’d like.

PR #64970, opened 2019-06-18
Update amazon-music from 7.3.2,20190506:200920c097 to 7.4.0.1811,20190605:18024163ef

suschizu:

### DIFF
--- a/Casks/amazon-music.rb
+++ b/Casks/amazon-music.rb
@@ -4,8 +4,7 @@ cask 'amazon-music' do
 
   # ssl-images-amazon.com/images was verified as official when first introduced to the cask
   url "https://images-na.ssl-images-amazon.com/images/G/01/digital/music/morpho/installers/#{version.after_comma.before_colon}/#{version.after_colon}/AmazonMusicInstaller.dmg"
-  appcast 'https://www.amazon.com/gp/dmusic/desktop/downloadPlayer',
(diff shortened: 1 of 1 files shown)


suschizu:
I've added a "configuration: :no_check" here because the appcast links directly to the app download (https://www.amazon.com/gp/dmusic/desktop/downloadPlayer)
I'm not sure if it's appropriate in this case @core-code, @vitorgalvao (?)

core-code:
it doesn't really link to the app download. it links to the html site. the reason you see 'gibberish' when downloading the appcast with curl is that its gzip compressed, which makes it pretty useless unless we implement some way to handle that.

i'm fine with 'configuration: :no_check' as it makes sense here unless someone implements a unzip method for the appcast check

vitorgalvao:
https://github.com/Homebrew/brew/pull/6240 should do it.

core-code:
great job on the compressed appcast support :) i remember a few others that also need that, though i am not sure which ones it were.

unfortunately, it won't help in this case, unless we fake the user-agent too ^^

vitorgalvao:
> unless we fake the user-agent too

Done.

core-code:
> > unless we fake the user-agent too
> 
> Done.

how, where?

vitorgalvao:
> how, where?

Same PR.

PR #65052, opened 2019-06-20
Update reaper from 5.97.800 to 5.97.900

ran-dall:

### DIFF
--- a/Casks/reaper.rb
+++ b/Casks/reaper.rb
@@ -1,6 +1,6 @@
 cask 'reaper' do
-  version '5.97.800'
-  sha256 'a6b0805427ae3323d5df01f101eced3a34bcecc18091d7a067bdb8e30f5d4f0c'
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao the changed versionnumber has caused major havoc on our side ;/ any possibility to revert this?

vitorgalvao:
@core-code Do you need the trailing zeros, or would this suffice as `5.97.9`?

core-code:
thanks for your quick reply. i've had a look but sadly i see no easy way to make due without those zeroes (they are in the CFBundleVersion and trying to remove things won't work on older versions of that app).
if its a problem on your end, i can always fork the cask.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/commit/004b18f3a191d966d179bb58af9e0e0b6a8485c5

PR #65054, opened 2019-06-20
Update perforce from 19.1 to 19.1-1813586

someposer:

### DIFF
--- a/Casks/perforce.rb
+++ b/Casks/perforce.rb
@@ -1,8 +1,8 @@
 cask 'perforce' do
-  version '19.1'
-  sha256 :no_check # required as upstream package is updated in-place
+  version '19.1-1813586'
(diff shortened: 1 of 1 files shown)


Casks/perforce.rb
```diff
@@ -1,8 +1,8 @@
 cask 'perforce' do
-  version '19.1'
-  sha256 :no_check # required as upstream package is updated in-place
+  version '19.1-1813586'
```
vitorgalvao: Seeing as the extra part doesn’t shown in the URL, adding it makes the cask harder to maintain, not easier.

How often is this updated?

someposer: Perforce usually updates their binaries every quarter. This quarter they updated but didn't change the minor version. 19.1-17xxxxx => 19.1-1813586.
This change also also follows the versioning precedent set in the p4v cask.

PR #65111, opened 2019-06-21
Updated TablePlus version to 2.5 build 233.

huyphams:

- [x] `brew cask audit --download Casks/tableplus.rb` is error-free.
- [x] `brew cask style --fix Casks/tableplus.rb` reports no offenses.

### DIFF
--- a/Casks/tableplus.rb
+++ b/Casks/tableplus.rb
@@ -1,5 +1,5 @@
 cask 'tableplus' do
-  version '2.5,233'
+  version '2.4,233'
   sha256 '45c4f67a4471e7ec482b9b85669eb1694f3cb30caf81d738ccaf9fbe47709236'
 
   # tableplus-osx-builds.s3.amazonaws.com was verified as official when first introduced to the cask
--- a/Casks/tableplus.rb
+++ b/Casks/tableplus.rb
@@ -1,10 +1,11 @@
 cask 'tableplus' do
-  version '2.4,233'
+  version '2.4.233'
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao sorry to be a pest but can you explain the rationale for the changes here?

1.) replacing version 2.5 with version 2.4 is IMO wrong. according to the CFBundleShortVersionString it is version 2.5. yes, the sparkle appcast mistakenly says '2.4'...no need to repeat that mistake ;)

2.) i don't see any reason to change ',' to '.' either. the 2.5 and 233 are seperate things, version number (short version string) and build number (bundle version), so i guess they should best be kept separate? as far as i can see, neither the URL interpolation nor the appcast configuration works better by mashing them together. we can use use .after_comma as usual instead of .patch....

vitorgalvao:
The `2.4` was indeed taken from the appcast, but I have no strong feelings on this and the `233`. I’m fine with changing them back.

core-code:
ok thanks. not that strong feelings either, but if you don't mind i'll send a PR

huyphams:
Hi, sorry for the confusion caused. I update the app cast to 2.5 (2.4 is a mistake) let me know if you want a PR.

vitorgalvao:
> Hi, sorry for the confusion caused.

Don’t worry, it wasn’t you! 

> I updated the app cast to 2.5 (2.4 is a mistake)

It was my mistake putting it at `2.4`; I trusted the appcast.

> let me know if you want a PR.

[No need](https://github.com/Homebrew/homebrew-cask/pull/65298). Thank you.

PR #65116, opened 2019-06-21
Update acquia-dev from 2-2019-02-08 to 2-2019-06-06

suschizu:


suschizu:
@vitorgalvao - could you perhaps have a look on the url stanza … I think there is a better solution to solve it :)

vitorgalvao:
@suschizu Would a version of `2.2019-06.06` work for you, or would you rather keep the `-`?

suschizu:
@vitorgalvao - I just noticed that this version stanza has nothing to do with the apps version - it's just the installers version number so your proposal is indeed ok!  Thanks and sorry for the fuss :)

suschizu:
thanks :) have a nice weekend 🌞

vitorgalvao:
Thank you, you too!

PR #65131, opened 2019-06-21
change version number from 5.2.3.757 to 3.1.9 upwork.rb

suschizu:
@yurikoles: my installation still shows 3.1.9 as latest version, the download too and I can't find a newer version - 5_2_3_757 is just part of the download-token :) 
-> see #63512

### DIFF
--- a/Casks/upwork.rb
+++ b/Casks/upwork.rb
@@ -1,5 +1,5 @@
 cask 'upwork' do
-  version '5.2.3.757,5_2_3_757_3idn3cdxmlq3m6d5'
+  version '3.1.9,5_2_3_757_3idn3cdxmlq3m6d5'
   sha256 'd7da1baf90469e23cb3f23da0151a8b897126205998d67502f7f9440386d6800'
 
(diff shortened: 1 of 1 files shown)


yurikoles:
@suschizu 
Знімок екрана 2019-06-21 о 18 41 40

suschizu:
ok you are right - it looks like they forgot to update the info.plist so the system shows 3.1.9 :(
it looks like they generally have a mess with their version numeration …
I'm not sure - what do you think @core-code , @vitorgalvao?

Bildschirmfoto 2019-06-21 um 17 48 32

core-code:
i can confirm that the version number in the Finder is 3.1.9 while the version number in the app itself is 5.2.3

the cask should probably also have auto_updates

vitorgalvao:
Let’s keep as is, since both sides of the comma will make sense. Also, presumably they’ll notice and fix the `Info.plist`, not the other way around.

core-code:
they had mismatched versions for a few months now and given that this just a wrapper for a website, i doubt they have much clue, but i'll try to contact them about this

PR #65203, opened 2019-06-23
Link dotnet binary in dotnet cask

j1mr10rd4n:
This PR improves the installation of the `dotnet` cask so that the `dotnet` command works after the install completes.

### DIFF
--- a/Casks/dotnet.rb
+++ b/Casks/dotnet.rb
@@ -15,6 +15,7 @@ cask 'dotnet' do
   depends_on macos: '>= :sierra'
 
   pkg "dotnet-runtime-#{version}-osx-x64.pkg"
+  binary '/usr/local/share/dotnet/dotnet'
 
   uninstall pkgutil: 'com.microsoft.dotnet.*',
             delete:  '/etc/paths.d/dotnet'


vitorgalvao:
Welcome, and thank you for the contribution.

However, this is needlessly complex. A single `binary '/usr/local/share/dotnet/dotnet'` should do in one line the same thing as all the other changes. Furthermore, I’m wary of merging this. They install via a `pkg`, and a `pkg` can do whatever it wants. If that binary is supposed to be in `/usr/local/bin`, why didn’t they do so in the first place?

j1mr10rd4n:
Thank you for the feedback - I did try with `binary` first but had it inside the `postflight` block so it didn't work - of course it should be in the normal `cask` block!

As for why the package doesn't link the binary itself: this would be the ideal solution; but it has been the source of previous previous SO questions ([1](https://stackoverflow.com/questions/41693683/ive-installed-dot-net-core-on-mac-but-didnt-find-dotnet-command), [2](https://stackoverflow.com/questions/53030531/dotnet-command-not-found-in-mac)) and I thought it would be nice if it just worked with `brew cask`.

j1mr10rd4n:
I've updated this branch so that it only changes a single cask - with the `binary` directive instead of the `postflight` block. If you are amenable to merging it, I'll create another PR for the `dotnet-sdk` cask.

vitorgalvao:
Thank you for doing and showing your research. I’d prefer an official source if you have one, but I’m inclined to merge. You can proceed with the other casks.

Thank you again.

Issue #25040, opened 2016-09-30
Bug report: VeraCrypt installation issue on macOS Sierra


drawcard:
Remember to follow the [pre bug report](https://github.com/caskroom/homebrew-cask/blob/master/doc/reporting_bugs/pre_bug_report.md) guide beforehand. Failure to do so might get your issue closed.

#### Description of issue

I'm having some trouble trying to install Veracrypt (on macOS Sierra) - I think it might be a permissions problem with /usr/local which is apparently read only in Sierra.

Output below.
```
$ brew cask install veracrypt
==> Caveats
veracrypt requires osxfuse with "MacFUSE Compatibility Layer" enabled.
If you did not enable that option, veracrypt will fail to install.

To enable the option, you need to install osxfuse manually (double-click the installer).
The installer should be located under /usr/local/Caskroom/osxfuse

==> Satisfying dependencies
==> Installing Cask dependencies: osxfuse
osxfuse ...
already installed
complete
==> Downloading https://download-codeplex.sec.s-msft.com/Download/Release?Projec
Already downloaded: /Users/drawcard/Library/Caches/Homebrew/Cask/veracrypt--1.18a
==> Verifying checksum for Cask veracrypt
==> Running installer for veracrypt; your password may be necessary.
==> Package installers may write to any location; options such as --appdir are i
Password:
==> installer: Error -  
Error:   Command failed to execute!

  ==> Failed command:
  ["/usr/bin/sudo", "-E", "--", "/usr/sbin/installer", "-pkg", "#<Pathname:/usr/local/Caskroom/veracrypt/1.18a/VeraCrypt_Installer.pkg>", "-target", "/"]

  ==> Standard Output of failed command:
  

  ==> Standard Error of failed command:
  installer: Error -  


  ==> Exit status of failed command:
  #<Process::Status: pid 59917 exit 1>
```

<details><summary>Output of `brew cask <command> --verbose`</summary>

```
[drawcard:~] 3s % brew cask install veracrypt --verbose
==> Caveats
veracrypt requires osxfuse with "MacFUSE Compatibility Layer" enabled.
If you did not enable that option, veracrypt will fail to install.

To enable the option, you need to install osxfuse manually (double-click the installer).
The installer should be located under /usr/local/Caskroom/osxfuse

==> Satisfying dependencies
==> Installing Cask dependencies: osxfuse
osxfuse ...
already installed
complete
[...]

PR #65213, opened 2019-06-23
OSXFUSE: Remove MacFUSE compatibility layer

russelldavis:
This was discussed at #26965 but the change didn't get made.

The compatibility layer was [originally added](https://github.com/Homebrew/homebrew-cask/pull/25044) to the formula as a workaround for a [Veracrypt issue](https://github.com/Homebrew/homebrew-cask/issues/25040), however, Veracrypt no [longer needs it](#https://github.com/veracrypt/VeraCrypt/issues/11).

There was concern that fossil might need it, however if you look at its [formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/fossil.rb), it is now being installed with `--disable-fusefs`, so I don't think that's a problem anymore.

On the flip side, installing the compatibility layer is actively harmful. It causes build errors in [python-llfuse](https://github.com/python-llfuse/python-llfuse) (`error: FUSE version too old, 2.9.0 or newer required`) (see https://github.com/borgbackup/borg/issues/4639).

The author of osxfuse [said](https://github.com/osxfuse/osxfuse/issues/114):

> I've noticed that some people are linking file systems against the compatibility layer. This is a really bad idea. Other formulas should link against libosxfuse not libfuse. There should really be no need for the compatibility layer in homebrew.

So, this PR removes installation of the compatibility layer.

### DIFF
--- a/Casks/osxfuse.rb
+++ b/Casks/osxfuse.rb
@@ -11,12 +11,7 @@ cask 'osxfuse' do
   auto_updates true
   conflicts_with cask: 'osxfuse-dev'
 
-  pkg "Extras/FUSE for macOS #{version}.pkg",
-      choices: [
-                 'choiceIdentifier' => 'com.github.osxfuse.pkg.MacFUSE',
-                 'choiceAttribute'  => 'selected',
-                 'attributeSetting' => 1,
-               ]
+  pkg "Extras/FUSE for macOS #{version}.pkg"
 
   postflight do
     set_ownership ['/usr/local/include', '/usr/local/lib']


vitorgalvao:
Welcome, and thank you for showing your research. Makes our job easier, and I very much appreciate it.

Easy merge. Thank you again for the informative PR.

leonklingele:
TrueCrypt no longer works now:

```
Dyld Error Message:
  Library not loaded: /usr/local/lib/libfuse.2.dylib
  Referenced from: /Applications/TrueCrypt.app/Contents/MacOS/TrueCrypt
  Reason: image not found
```

Can this change please be reverted?

russelldavis:
@leonklingele if we reverted this, it would cause other things to break, like `python-llfuse` and `borgbackup` (see the details in the PR description). The author of osxfuse has said linking against the compatibility layer is a bad idea -- I'd recommend filing a bug with TrueCrypt and/or updating its formula (possibly doing what Veracrypt did, linked above).

razvanphp:
TrueCrypt is no longer maintained AFAIK.

Can we at least install an older version of the cask? What was the latest version with compatibility layer included?

Or can we manually install the later?

Thank you!

PR #65220, opened 2019-06-24
Update lockrattler appcast

core-code:
Experimental idea: Lockrattler is one of these with a problematic appcast. Its too noisy to observe for changes (wordpress!) and also contains a lot of other version numbers that make a 'contains' check less than ideal.

The good news is that this app contains a 'check for updates' function that has been recently released.
The bad news is that the configuration file ( https://raw.githubusercontent.com/hoakleyelc/updates/master/eclecticapps.plist ) is used for a few dozen apps and therefore just as noisy and contains just as many unrelated versions.

I've been toying with the idea of using a CGI to extract just the portions we need from the file.

The CGI ( https://gist.github.com/core-code/0def150fa245b333fb0541f9f999b040#file-extract_text_split-cgi ) takes a URL to process as parameter as well as a JSON array containing the 'split points'. Configured in the right way it just spits out the version number we need. 
I think CGIs like these could also be useful if we need special headers ( ubiquiti-unifi-controller  ), cookies or a POST request to get the information we need.
Ideally all this (including our current CGIs that get the filename or redirect) would be implemented as parameters on the appcast stanza, defining what to do, instead of defining a special way to do this (by calling our CGI), but i don't see that happening anytime soon.

Feel free to reject this PR - i don't think its strictly necessary in this case, and this might also not be the most popular cask anyway. Also, I have no problem implementing a list of 'override' appcasts on our end to help keep casks like this one up-to-date if you feel like using this 'extract_text' CGI in the appcast of the caskfile is not a good idea.

### DIFF
--- a/Casks/lockrattler.rb
+++ b/Casks/lockrattler.rb
@@ -4,7 +4,7 @@ cask 'lockrattler' do
 
   # eclecticlightdotcom.files.wordpress.com was verified as official when first introduced to the cask
   url "https://eclecticlightdotcom.files.wordpress.com/#{version.after_comma.major}/#{version.after_comma.minor}/lockrattler#{version.before_comma.no_dots}.zip"
-  appcast 'https://eclecticlight.co/lockrattler-systhist/'
+  appcast 'https://www.macupdater.net/cgi-bin/extract_text/extract_text_split.cgi?splitters=[{"s":"LockRattler","i":1},{"s":"Version","i":1},{"s":"dict","i":0}]&url=https://raw.githubusercontent.com/hoakleyelc/updates/master/eclecticapps.plist'
   name 'Lock Rattler'
   homepage 'https://eclecticlight.co/'
 
--- a/Casks/lockrattler.rb
+++ b/Casks/lockrattler.rb
@@ -4,7 +4,7 @@ cask 'lockrattler' do
 
   # eclecticlightdotcom.files.wordpress.com was verified as official when first introduced to the cask
   url "https://eclecticlightdotcom.files.wordpress.com/#{version.after_comma.major}/#{version.after_comma.minor}/lockrattler#{version.before_comma.no_dots}.zip"
-  appcast 'https://www.macupdater.net/cgi-bin/extract_text/extract_text_split.cgi?splitters=[{"s":"LockRattler","i":1},{"s":"Version","i":1},{"s":"dict","i":0}]&url=https://raw.githubusercontent.com/hoakleyelc/updates/master/eclecticapps.plist'
+  appcast 'https://www.macupdater.net/cgi-bin/extract_text/extract_text_split.cgi?splitters=\[{"s":"LockRattler","i":1},{"s":"Version","i":1},{"s":"dict","i":0}\]&url=https://raw.githubusercontent.com/hoakleyelc/updates/master/eclecticapps.plist'
   name 'Lock Rattler'
   homepage 'https://eclecticlight.co/'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> generic exception: Traceback (most recent call last):
  File "extract_text_split.cgi", line 25, in &lt;module&gt;
    response = urllib2.urlopen(req)
  File "/usr/lib/python2.7/urllib2.py", line 154, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python2.7/urllib2.py", line 435, in open
    response = meth(req, response)
  File "/usr/lib/python2.7/urllib2.py", line 548, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib/python2.7/urllib2.py", line 473, in error
    return self._call_chain(*args)
  File "/usr/lib/python2.7/urllib2.py", line 407, in _call_chain
    result = func(*args)
  File "/usr/lib/python2.7/urllib2.py", line 556, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
HTTPError: HTTP Error 404: Not Found

core-code:
no idea whats going on here, i can only suspect the fact that 'curl' is used and that it likes to use 'globbing' for everything with square brackets unless you pass '-g' gets in the way here. i'll try to work around it

vitorgalvao:
> i can only suspect the fact that 'curl' is used

I got that output from opening the appcast in a web browser, not via `curl`.

core-code:
thats really weird, i can open it without problems in both Safari and Chrome

Screenshot 2019-06-24 at 16 58 30

vitorgalvao:
> i can open it without problems in both Safari and Chrome

I get the same error in both browsers.

Casks/lockrattler.rb
```diff
@@ -4,7 +4,7 @@
 
   # eclecticlightdotcom.files.wordpress.com was verified as official when first introduced to the cask
   url "https://eclecticlightdotcom.files.wordpress.com/#{version.after_comma.major}/#{version.after_comma.minor}/lockrattler#{version.before_comma.no_dots}.zip"
-  appcast 'https://www.macupdater.net/cgi-bin/extract_text/extract_text_split.cgi?splitters=\[{"s":"LockRattler","i":1},{"s":"Version","i":1},{"s":"dict","i":0}\]&url=https://raw.githubusercontent.com/hoakleyelc/updates/master/eclecticapps.plist'
+  appcast 'https://www.macupdater.net/cgi-bin/extract_text/extract_text_split.cgi?splitters=\[\{%22s%22:%22LockRattler%22,%22i%22:1\},\{%22s%22:%22Version%22,%22i%22:1\},\{%22s%22:%22dict%22,%22i%22:0\}\]&url=https://raw.githubusercontent.com/hoakleyelc/updates/master/eclecticapps.plist'
```
vitorgalvao: No difference. Same error.

core-code: hmmm no idea, i even tried Opera now, but it works too.

does this one work for you?

`https://www.macupdater.net/cgi-bin/extract_text/extract_text_split.cgi?splitters=[]&url=https://raw.githubusercontent.com/hoakleyelc/updates/master/eclecticapps.plist`

core-code: > No difference. Same error.

Travis disagrees ^^ #131240 passed

vitorgalvao: > does this one work for you?

Yes.

core-code:
;) i wonder if we shouldn't disable 'globbing' in homebrew's 'curl' calls instead of massaging URLs to work around it

vitorgalvao:
https://github.com/Homebrew/brew/pull/6252

PR #65435, opened 2019-06-27
Update capture-one

core-code:


core-code:
we'd need a configuration that removes the trailing dot if the version number ends in a trailing dot...

core-code:
i am not sure about major_minor. it is correct now, but will not be when we have a version number that has really 3 components

vitorgalvao:
> but will not be when we have a version number that has really 3 components

It will still be correct, just not as complete. `patch` number shouldn’t be problematic (i.e. updating that should always be beneficial).

PR #65451, opened 2019-06-28
Update pdfelement from 6.7.12,2991 to 7.0.4,5237

core-code:


core-code:
it seems pdfelement and pdfelement-pro are the same download now. should the pdfelement-pro cask be removed or stay and be identical?

vitorgalvao:
If it’s the same download and the same app, let’s remove one of them.

core-code:
this one is more popular, so i guess we should remove the pro one. i'll send a pr

core-code:
anything still missing here?

PR #65456, opened 2019-06-28
Update qiyimedia from 5.9.11 to 5.10.10

suschizu:

### DIFF
--- a/Casks/qiyimedia.rb
+++ b/Casks/qiyimedia.rb
@@ -1,9 +1,9 @@
 cask 'qiyimedia' do
-  version '5.10.10'
+  version '20190628,5.10.10'
   sha256 '38be0152e83714239ca9fd4295fc061b811a66b200907e4f5d6a5f1f0f805762'
 
   url 'https://mbdapp.iqiyi.com/j/ot/iQIYIMedia_000.dmg'
-  appcast 'https://app.iqiyi.com/mac/player/index.html'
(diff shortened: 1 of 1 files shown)


suschizu:
`appcast at URL 'https://app.iqiyi.com/mac/player/index.html' does not contain the version number: '5.10.10'`
unfortunately true - they haven't updated the version on the website until now but the downloaded version is 5.10.10 - I also can't find a better appcast … should we add a 'configuration: no-check' here? @vitorgalvao @core-code?

core-code:
i am not sure. lets see what vitor prefers

vitorgalvao:
The download redirects to a path that doesn’t have the version number, but it does have a date: `http://59.52.16.9/cdn/pcclient/20190628/16/33/iQIYIMedia_000.dmg`. We could add `20190628` as `after_comma`, and make the appcast your `cgi`.

Presumably you’re downloading it anyway to confirm the version, so maybe the redirected URL can be considered in that process?

core-code:
good idea! we can use
version '20190628,5.10.10'
appcast 'https://macupdater.net/cgi-bin/check_urls/check_url_redirect.cgi?url=https://mbdapp.iqiyi.com/j/ot/iQIYIMedia_000.dmg'

suschizu:
thanks :)

core-code:
ugh, a minute ago it was working. chinese servers...

vitorgalvao:
It’s all the same to me, but wouldn’t you rather have the real version before the comma?

core-code:
thanks for considering. its the same for us too, just a different 'mapping number' so it doesn't matter. by having the date first we can avoid a 'configuration' for the appcast, thats why i suggested it this way.

PR #65513, opened 2019-06-29
Update intel-power-gadget from 3.5.5,778485 to 3.5.5,828382

core-code:


seognil:
3.5.5,778485 works fine to me

3.5.5,828382 seems won't work on 10.14.5 (I tried that)

https://software.intel.com/en-us/articles/intel-power-gadget#comment-1941590

core-code:
as far as i know, we don't do quality assurance here at Homebrew Cask. if there are problems with the latest version of any software, please report it to the vendor.

vitorgalvao:
@seognil @core-code is correct. We only install things; if they don’t work that’s the vendor’s problem, complain to them. That dialog you posted even tells you as much. It happens because the app is 32-bit.

PR #65568, opened 2019-07-01
add depends on mojave to maccy.rb

suschizu:
app depends on 10.14


suschizu:
`Cask maccy depends on macOS release >= 10.14, but you are running release 10.13.`

PR #65597, opened 2019-07-01
Update capture-one

core-code:


core-code:
i am not sure if the #{version} in the appcast will work completely as hoped, since the actual network request that the app sends uses a more detailed version number than the cask (or the BundleVersons / ShortVersionString) actually has. but lets keep it this way and wait and see ;)

vitorgalvao:
> since the actual network request that the app sends uses a more detailed version number than the cask

I tested it with the shorter version and it works. Let’s see if it continues to.

PR #65613, opened 2019-07-02
Update anka-flow from 1.4.4.105 to 2.0.0.108

core-code:


vitorgalvao:
What’s the software this installs now? The cask token may need a rewrite.

core-code:
hm i don't think much has changed:

• the homepage ( https://veertu.com/download-anka-run/ ) still calls it 'Anka Flow'
• the only app installed by the package is still named 'AnkaView', so there is no connection to the cask name, just like before
• what has changed is the name of the PKG. if you think this warrants re-name of the cask to 'anka', i can do that

PR #65642, opened 2019-07-02
Update jalbum from 18.3.0 to 18.3.1

suschizu:


suschizu:
the download from the website contains version 18.3.1 but the app cast just lists 18.3 - I don't think it's a good idea to add a major_minor here because sometimes they use the patch too …

vitorgalvao:
> I don't think it's a good idea to add a major_minor here because sometimes they use the patch too

The patch number is typically safe to update because it’s just bug fixes, it’s not used for betas.

PR #65646, opened 2019-07-02
Update betterzip from 4.2.3 to 4.2.4

arpitnamdeo:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #65665, opened 2019-07-03
betterzip.rb: appcast with patch

vitorgalvao:
Pinging @core-code. Continuation of https://github.com/Homebrew/homebrew-cask/pull/65646#discussion_r299948178.


core-code:
thanks for the heads up there, but i think this one is better ;)

#65671

PR #65689, opened 2019-07-03
Update lockrattler.rb appcast stanza

suschizu:
appcast stanza: delete \
the download link wasn't working with the backslashes so I deleted it


core-code:
seem to be needed for some reason. i can adjust this on our side

PR #65748, opened 2019-07-04
Update teamdrive from 4.6.7.2230 to 4.6.7.2328

ran-dall:

### DIFF
--- a/Casks/teamdrive.rb
+++ b/Casks/teamdrive.rb
@@ -1,6 +1,6 @@
 cask 'teamdrive' do
-  version '4.6.7.2230'
-  sha256 '87e00a27fbcb88dc5fb44f72ed8d81a4a7914a1bbb1bace71f50a843816e9adc'
+  version '4.6.7.2328'
+  sha256 'f34d3615d60efc7c037972684397858065b6bc6e7bdddcbea722d3b31e6d61bc'
 
   # s3-eu-west-1.amazonaws.com/s3download.teamdrive.net was verified as official when first introduced to the cask
   url "https://s3-eu-west-1.amazonaws.com/s3download.teamdrive.net/#{version.major_minor}.#{version.split('.').last}/TMDR/mac-10.14.2/Install-TeamDrive-#{version}_TMDR.dmg"


ran-dall:
@core-code 👉 Can you please look into what can be done about this `appcast`, it's pretty terrible. I also think we're completely out of sync with what's offered on the site; if I understand correctly.

core-code:
i am not sure i understand. the "appcast" is the download site, how can it be out of sync?

ran-dall:
@core-code Let me clarify, the `version` we currently have in the HBC tap is about few `versions` different from what's on the download site. Outside of that, the download site appears to be `https://synqion.com/downloads`. What I was talking about before, come back to our discussion a few months ago about certain sites not being very good `appcasts` and really only hinder the system. I meant to point out that this might be a good fit for your new `cgi`.

core-code:
sorry, i still don't get it. the teamdrive version available on the website and in the cask is (now) identical, its 4.6.7.2328. i don't see anything out-of-sync here.
i agree that HTML site appcasts are bad in general, especially when they contain lots of version, but when only doing a 'contains' check it should be fine.
i also don't understand why you mention 'synquion'. this seems to be a different company and product.

ran-dall:
> sorry, i still don't get it. the teamdrive version available on the website and in the cask is (now) identical, its 4.6.7.2328. i don't see anything out-of-sync here.

When I go to the link; I get a page that indicates  `Synqion 4.6.6 (Build: 2196)` is the latest - linking the user to download located @ `https://s3-eu-west-1.amazonaws.com/download.synqion.com/4.6.2196/SYNQ/mac-10.13.4/Install-Synqion-4.6.6.2196_SYNQ.dmg`

> only doing a 'contains' check it should be fine

Unfortunately, it's pretty hard to have the ML run this and understand what needs to be done. It's getting better but there is still a long way to go. `appcasts` like this only confuse it more.

>  i also don't understand why you mention 'synquion'. this seems to be a different company and product.

I guess neither do I. When I visit the links on the `cask` that's where it takes me. Perhaps, its some sort of regional redirect.

core-code:
i guess the your guess with the 'regional redirect' is true:
https://sourceforge.net/p/keepass/discussion/329220/thread/fdf48313f7/

> Unfortunately, it's pretty hard to have the ML run this and understand what needs to be done. It's
> getting better but there is still a long way to go. appcasts like this only confuse it more.

i am not sure if its possible but you could rely on the 'appcast_containsversion' file. if the version number is not in the appcast, it will be listed there.

in any case, i tried to find out the file the app uses for an update check so that we have a better appcast, but was not successful there. seems like you need an account with them or something to fully launch the app.

PR #65636, opened 2019-07-02
Update fl-studio from 20.1.2.210 to 20.5.0.1142

ran-dall:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #65783, opened 2019-07-05
Update fl-studio from 20.5.0.1142 to 20.5.0.476

core-code:

### DIFF
--- a/Casks/fl-studio.rb
+++ b/Casks/fl-studio.rb
@@ -1,9 +1,9 @@
 cask 'fl-studio' do
-  version '20.5.0.1142'
+  version '20.5.0.476'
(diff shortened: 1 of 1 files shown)


core-code:
this fixes the incorrect PR #65636

1.) follow the redirect for the download URL so that incorrect version numbers aren't possible anymore
2.) adjust the appcast so that the version number of the Windows version doesn't show up anymore (the other option besides sniffing the redirect would be to use our new CGI to cut the existing appcast so that the windows portion is omitted - if you prefer that just let me know)

ran-dall:
@core-code 👉 I like this new way you have it. It's a lot easier on the ML.

Issue #65309, opened 2019-06-26
What is the difference between `wine` and `cask/wine-stable`?


gaochenyi:
There are `wine` and `cask/wine-stable`. They both seem to be the stable branch of wine and have the same version number. What is the difference?

vitorgalvao:
I think @yurikoles is the one that pushed for the `wine` casks. He may be able to explain in detail.

yurikoles:
The main difference is that `wine` formula is built from source and `wine-stable` is official binary pkg from WineHQ.

yurikoles:
But `wine` can't be build on Mojave and higher due to lack of 32 bit dev libs. And can't run on Catalina and higher because Apple had deprecated 32 bit apps.

vitorgalvao:
I though there were bigger differences. In that case, I see no point in having the cask. It should be just the formula.

PR #65831, opened 2019-07-06
deprecate wine-stable in flavor of wine formula

yurikoles:
See https://github.com/Homebrew/homebrew-cask/issues/65309#issuecomment-508922390


vitorgalvao:
What about the other wine versions?

yurikoles:
`-devel` is unstable version, `-staging` also adds experimental patches on top of it.

yurikoles:
There was a long battle on bringing those in form of formula(-s).

vitorgalvao:
> There was a long battle on bringing those in form of formula(-s).

And do they exist as formulae?

yurikoles:
No, they don't (yet).

vitorgalvao:
> No, they don't (yet).

Then let’s wait a bit, or this will be confusing.

PR #65866, opened 2019-07-06
Create user exceptions for request-info bot

vitorgalvao:
@core-code @suschizu This will hopefully get that damn bot of our backs on the faster PRs!

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,0 +1,4 @@
+requestInfoUserstoExclude:
+  - code-code
+  - suschizu
+  - vitorgalvao
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,4 +1,5 @@
 requestInfoUserstoExclude:
   - code-code
+  - ran-dall
   - suschizu
   - vitorgalvao


ran-dall:
@vitorgalvao Can I ask why I wasn't included? I'm a little offended to TBH

ran-dall:
@vitorgalvao Honestly I don't want to be included, it's just BS

vitorgalvao:
> Can I ask why I wasn't included? I'm a little offended to TBH

Apologies! Added now. I did think of including you (and the other cask maintainers, which you’ll notice are also absent) but I don’t remember you ever being pestered by the bot, as your PRs always include the template, don’t they?

The inclusion in the PR was made on the basis of “who submits PRs that will be good, but sometimes doesn’t include the template and that’s fine?”. Your PRs are good, but you always include the template (that I’ve noticed), hence the initial exclusion.

vitorgalvao:
In sum, the though process was “who **deserves and needs** this?”. You (and the rest of the cask maintainers) deserve it, but from what I’ve seen don’t need it, *which is a positive thing*.

ran-dall:
@vitorgalvao Lol. FYI I have a sarcastic sense of humor. I wasn't serious. 🤪

ran-dall:
> Apologies! Added now. I did think of including you (and the other cask maintainers, which you’ll notice are also absent) but I don’t remember you ever being pestered by the bot, as your PRs always include the template, don’t they?

BUT you made me think up a good point. 
What would happen if we use the ML to check casks updates instead of Travis? It might take some working but it shouldn't be too difficult, and there are a lot of upsides cause the ML can do a lot more things Travis can't.

vitorgalvao:
What Travis does, any user can do locally. That’s an important feature.

ran-dall:
> What Travis does, any user can do locally. That’s an important feature.

Of course, the idea would be to have something that is more resilient than this or Travis in checking a PR for errors.

core-code:
thanks vitor!

PR #65889, opened 2019-07-07
Update file-juicer

ideologysec:

- [x] `brew cask audit --download --appcast --token-conflicts {{cask_file}}` is error-free.

### DIFF
--- a/Casks/file-juicer.rb
+++ b/Casks/file-juicer.rb
@@ -1,6 +1,6 @@
 cask 'file-juicer' do
   version '4.77'
-  sha256 :no_check # required as upstream package is updated in-place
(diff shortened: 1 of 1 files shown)


core-code:
do you have justification for this change?

ideologysec:
sha256 sum is better than not? ie, ideally Cask is verifying downloaded apps as much as possible. The original comment said that no checksum was done because upstream did replace in place, but I don't see any justification for that - there is no autoupdate upstream, it's required to download a new copy of the app each time it's updated.

vitorgalvao:
> there is no autoupdate upstream, it's required to download a new copy of the app each time it's updated.

That’s not what the comment means. It means the file is replaced upstream but URL stays the same. The comment is added to apps that do that too often, breaking the cask frequently.

PR #65914, opened 2019-07-07
Update mobile-mouse-server

core-code:


core-code:
hm,  the app website works only when using a mozilla user agent, but not in curl, which travis uses.

a CGI to fake the user agent?

vitorgalvao:
`audit` uses a browser user agent, but it’s stripped down.

PR #65925, opened 2019-07-08
Update double-commander

core-code:

### DIFF
--- a/Casks/double-commander.rb
+++ b/Casks/double-commander.rb
@@ -1,9 +1,9 @@
 cask 'double-commander' do
   version '0.9.5-8947'
-  sha256 '924e9a25b872ccbb91eacb44bc4a7a4f99a6136c718d2d94a87768deb68cfa85'
+  sha256 '4147954fa8dc47d0e4f1393452091a4900ba091edcf8c90252ccb40d0e7a78a3'
 
   # downloads.sourceforge.net/doublecmd was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


core-code:
we should use the preferred 'Cocoa' version instead of the 'QT' version that is only for systems older than El Capitan and will probably disappear soon

kaznovac:
@core-code can you also introduce the `qt` version. this `cocoa` build is quite unstable on my mohave installation?

vitorgalvao:
> this `cocoa` build is quite unstable on my mohave installation?

Please take that up with upstream.

core-code:
[double-commander-qt.rb.txt](https://github.com/Homebrew/homebrew-cask/files/3493505/double-commander-qt.rb.txt)
i've attached the latest cask for the QT version for your convenience. can you send a PR to get it added through the normal channels? 
https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#adding-a-cask

core-code:
sorry didn't see that comment earlier.

kaznovac:
Thanks core-code.

@vitorgalvao `Qt` is a variation is more mature than `cocoa` GUI build.
Vendor provides both variations - it would be great if `brew cask` provides them both also (IMHO more mature and stable version unsufixed, and alternatives clearly differentiated).

Do you think `Qt` variation is viable candidate for `cask-versions`?

vitorgalvao:
@kaznovac No, due to its dependency on systems older than what we officially support. I recommend you use a personal tap if you need the older cask.

PR #65931, opened 2019-07-08
Revert "removing android-ndk cask based on #58883"

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/issues/58883#issuecomment-509028688

### DIFF
--- a/Casks/android-ndk.rb
+++ b/Casks/android-ndk.rb
@@ -1,0 +1,40 @@
+cask 'android-ndk' do
+  version '19'
+  sha256 '04159ade2fc5c7d055248cf65664039b8596f4b9ee3fbc44a9bf2ce2ee28d95d'
+
+  # dl.google.com/android/repository/android-ndk was verified as official when first introduced to the cask
+  url "https://dl.google.com/android/repository/android-ndk-r#{version}-darwin-x86_64.zip"
+  name 'Android NDK'
+  homepage 'https://developer.android.com/ndk/index.html'
+
+  conflicts_with cask: 'crystax-ndk'
+
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/ndk_exec.sh"
+  preflight do
+    FileUtils.ln_sf("#{staged_path}/android-ndk-r#{version}", "#{HOMEBREW_PREFIX}/share/android-ndk")
+
+    IO.write shimscript, <<~EOS
+      #!/bin/bash
+      readonly executable="#{staged_path}/android-ndk-r#{version}/$(basename ${0})"
+      test -f "${executable}" && exec "${executable}" "${@}"
+    EOS
+  end
+
+  [
+    'ndk-build',
+    'ndk-depends',
+    'ndk-gdb',
+    'ndk-stack',
+    'ndk-which',
+  ].each { |link_name| binary shimscript, target: link_name }
+
+  uninstall_postflight do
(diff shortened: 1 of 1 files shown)


sesam:
Since in https://github.com/Homebrew/homebrew-cask/issues/58883 it says up top that the only ENV variable still relevant seems to be ANDROID_HOME but this code has a `caveat` suggesting to add an ENV variable that likely no longer works.

What is the proper process to get this fixed?

I guess I could maybe open a new Issue and try to tie together the existing several locations where this is discussed, but that might complicate things even further. And I'm not actively using this, just have a strong memory of the hours lost - working with Android is lots of not useful outputs, and near none useful install/post-install warnings. I rant, sorry.

vitorgalvao:
Proper way to fix it would be a PR changing the message, if that’s the problem.

Also pinging @rcombs, since they use this.

sesam:
Try this:
1. install cask android-ndk
2. Open /usr/local/Caskroom/android-ndk/19/android-ndk-r19/prebuilt/darwin-x86_64/bin/ndk-which
3. Lines 18-20 says this tool is DEPRECATED
4. The script uses its path on disk and steps up/out to find the ndk within which it is installed. The ANDROID_NDK_DIR is not used by ndk-which.

A question remains if there are some other tools that use the ANDROID_NDK_DIR.

@rcombs Likely `ndk-which` being marked deprecated isn't a showstopper for you. But can you see if `ndk-which` still does its job in your use-case where you don't want to have to insatll the android sdk?

vitorgalvao:
> 3\. Lines 18-20 says this tool is DEPRECATED

That’s an important consideration that wasn’t made in the other issue. Does that mean `android-ndk` as a whole is deprecated, or just `ndk-which`? If the former, I’ll mark the cask as deprecated and remove in (let’s say) six months.

rcombs:
It's just the `ndk-which` tool, which I use just as a portable way to locate the NDK install. Specifically, I do `ndk-which ld` and back up a few dirs from there. It wouldn't be a huge deal to just use `/usr/local/share/android-ndk` explicitly on macOS instead, if that's eventually removed.

sesam:
@rcombs Specifically in https://github.com/Homebrew/homebrew-cask/issues/58883#issuecomment-464232674 it was suggested that the ndk can not be used independently. What parts of the android-ndk are you using? The ndk seems to contain a mix of common unix tools like make, python and some more, didn't look it all through and now I'm curious what you do after finding the actual dir? Does the not deprecated variant `$ANDROID_HOME/ndk-bundle` work equally well as your current ndk-which + going some dirs up?

rcombs:
I don't have an `$ANDROID_HOME` env var, nor dir named `ndk-bundle` anywhere I can find. The essential components are the sysroot (`/usr/local/share/android-ndk/sysroot`, containing the Android platform headers and static and shared libraries) and the toolchains (`/usr/local/share/android-ndk/toolchains`, containing clang (AIUI with some modifications to automatically use the correct gcc components) and the binutils tools such as ld, nm, objcopy, etc. required to link for the relevant Android/Linux targets). My understanding is that the SDK provides a way to install these, but it's significantly more complex than using the cask and pointing at `/usr/local/share/android-ndk`.

sesam:
Excellent. This clearly shows the android-ndk cask has value, as well as contradicts the loose suggestion that the ndk isn't useful on it's own.

I'm thinking maybe removing the caveat string could help limiting the amount of brew cask users that install this cask, then install the sdk, and them the two don't find each other and you land in a situation where ndk commands on the command line modify things that the sdk ignores so the sdk then goes on to "help" creating a second duplicate ndk-bundle folder.

PR #65963, opened 2019-07-08
Update skype from 8.48.0.51 to 8.49.0.49

core-code:

### DIFF
--- a/Casks/skype.rb
+++ b/Casks/skype.rb
@@ -1,6 +1,6 @@
 cask 'skype' do
-  version '8.48.0.51'
-  sha256 '78c5e8084b4281cbed4e02246dbf511ddc414daf4b0279e1043a5c6f473d77d5'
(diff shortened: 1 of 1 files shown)


core-code:
their update-check config file is notoriously lagging the download from the homepage

i wonder if we should use this instead
`
https://macupdater.net/cgi-bin/check_urls/check_url_redirect.cgi?url=https://go.skype.com/skype.download&user_agent=Mozilla/5.0%20(Macintosh;%20Intel%20Mac%20OS%20X%2010_14_4)`

vitorgalvao:
Seems to be up-to-date now.

core-code:
if its just a day behind i don’t think there is much point in changing the appcast

vitorgalvao:
Especially since the build will always fail anyway (`string contains null byte`).

PR #66023, opened 2019-07-09
Update blue-jeans from 2.13.1.17 to 2.14.0.487

core-code:


core-code:
for some reason the appcast check receives an empty output when downloading that URL. i am not sure why, perhaps its a user-agent issue, though normal 'curl' works fine.

PR #66216, opened 2019-07-14
Update gitkraken appcast stanza

core-code:


core-code:
why is the bot still talking

PR #66317, opened 2019-07-16
Update ocenaudio from 3.7.0 to 3.7.1

core-code:


404NetworkError:
@core-code Mind updating the `sha256` of `MacOS.version <= :high_sierra` also? I believe this is why your PRs keep failing CI.

core-code:
@404NetworkError aaah you are completely right. i was under the wrong impression they hadn't updated the sierra file since the mojave version shipped. which is of course nonsense and you are right, this has caused the CI failures. thanks for the heads-up.

PR #66457, opened 2019-07-19
Update adoptopenjdk from 12.0.1,12 to 12.0.2,10

steinybot:

### DIFF
--- a/Casks/adoptopenjdk.rb
+++ b/Casks/adoptopenjdk.rb
@@ -4,7 +4,7 @@ cask 'adoptopenjdk' do
 
   # github.com/AdoptOpenJDK was verified as official when first introduced to the cask
   url "https://github.com/AdoptOpenJDK/openjdk#{version.major}-binaries/releases/download/jdk-#{version.before_comma}%2B#{version.after_comma}/OpenJDK#{version.major}U-jdk_x64_mac_hotspot_#{version.before_comma}_#{version.after_comma}.tar.gz"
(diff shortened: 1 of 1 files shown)


core-code:
we could try using
https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/latest
to exclude those nightlies

claudionieder:
Well the current atom app cast URL also incorporates #{version.major} so I don't see an why not have an URL which incorporates also #{version.before_comma} and #{version.after_comma} if that is what it takes to make it work.

core-code:
> Well the current atom app cast URL also incorporates #{version.major} so I don't see an why not have an URL which incorporates also #{version.before_comma} and #{version.after_comma} if that is what it takes to make it work.

i am not sure what you are suggesting. which URL would you use for an appcast?

claudionieder:
My comment was thought as an answer to steinybot's question "It could use https://api.github.com/repos/AdoptOpenJDK/openjdk12-binaries/releases/tags/jdk-12.0.2+10 **but that is version specific**. I was thinking that anyhow the major version is already in todays used, not working, appcast URL so there is no harm in using also minor version too.

core-code:
@steinyboy can you try changing the appcast stanza to use 'https://github.com/AdoptOpenJDK/openjdk12-binaries/releases/latest' ?

core-code:
thanks looks good to me

core-code (approved):

PR #66471, opened 2019-07-19
Update privatetunnel to add appcast

core-code:


Casks/privatetunnel.rb
```diff
@@ -4,6 +4,7 @@
 
   # swupdate.openvpn.org/privatetunnel was verified as official when first introduced to the cask
   url "https://swupdate.openvpn.org/privatetunnel/client/privatetunnel-mac-#{version}.dmg"
+  appcast 'https://www.privatetunnel.com/apps/#'
```
reitermarkus: Why the `#` at the end?

core-code: well, thats the URL you get when navigating their site. but i guess it has zero impact on anything and can safely be removed

PR #66502, opened 2019-07-20
Update jami from 20190602.0216 to 201907191319

ran-dall:

### DIFF
--- a/Casks/jami.rb
+++ b/Casks/jami.rb
@@ -3,8 +3,7 @@ cask 'jami' do
   sha256 'bd40a94d31371756a4d7ecdeca49e0592f0d20a786ce60589e97be7091293fb6'
 
   url "https://dl.ring.cx/mac_osx/jami-#{version}.dmg"
-  appcast 'https://dl.ring.cx/mac_osx/sparkle-ring.xml',
-          configuration: version.major
+  appcast 'https://dl.ring.cx/mac_osx/sparkle-ring.xml'
   name 'Jami'
   name 'Savoir-faire Linux Ring'
   homepage 'https://ring.cx/'
--- a/Casks/jami.rb
+++ b/Casks/jami.rb
@@ -1,8 +1,8 @@
 cask 'jami' do
-  version '20190602.0216'
-  sha256 'fc813559c65295796fd5b9b69da31cec4a5e98fea5213e792f7fcfa9c2dbcf9c'
+  version '201907191319'
+  sha256 'bd40a94d31371756a4d7ecdeca49e0592f0d20a786ce60589e97be7091293fb6'
 
-  url "https://dl.ring.cx/mac_osx/ring-#{version.no_dots}.dmg"
+  url "https://dl.ring.cx/mac_osx/jami-#{version}.dmg"
   appcast 'https://dl.ring.cx/mac_osx/sparkle-ring.xml',
           configuration: version.major
   name 'Jami'


ran-dall:
@core-code  `configuration: version.major` this isn't defined for the ML; so it causes it to bypass the the `appcast` and do a `download analysis` What should it be doing?

vitorgalvao:
> What should it be doing?

`configuration` on `appcast` tells it “look for this value instead of `version`”. `version.major` is [an HBC method](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md#version-methods) that says “return only the `major` part of a version”.

If a `version` is `1.2.3` but the appcast’s contents only show `1.2`, the check for “is `version` in the appcast feed” will return a fail, because `1.2.3` is indeed not there. With `configuration`, we’re telling the check “look for this string instead of `version`”. In the example, `configuration: version.major_minor` is saying “look for `1.2`”, making the check succeed.

vitorgalvao:
Took the opportunity to finally [document this](https://github.com/Homebrew/homebrew-cask/pull/66524).

core-code:
i've sent pr #66527 to re-add the dot and #{version.no_dots} since the trailing "1319" isn't actually part of the version number in the app and its impossible to know automatically how much to strip away otherwise

ran-dall:
🙌 Thank you both of you! @core-code and @vitorgalvao

Can I ask that when something like get added like this gets added next time; someone would shoot me a mention, please 🙏? 

What really happened was the ML went bananas and gave me all sorts of weird, contradicting info and I didn't really know what it all meant.

core-code:
yes, will do! 

i think one thing that might be missing is a place for all important HBC contributors to exchange information or news. 

(i've been meaning to mention a HomebrewCask iPhone app i wrote, but i wouldn't know an appropriate place to share it)

ran-dall:
@core-code @vitorgalvao 👉 Might I suggest a mailing list?

vitorgalvao:
> i've been meaning to mention a HomebrewCask iPhone app i wrote, but i wouldn't know an appropriate place to share it

Please do tell! I’m interested in knowing what it is. I’m also curious to see the icon — do you have one?

Opening a new issue for that is fine. Details!

> Might I suggest a mailing list?

I’ve been thinking of the mailing list idea for a long while. But I doubt it’d have enough subscribers to justify the work — look at the HBC twitter account followers number (though it’s been rising, I think).

It’s one of those things I’d definitely like to do if this was a paid job. As it is, the time sink would take me away from more important things, even inside HBC itself. The `configuration` documentation is something I’d have done sooner, in the past when the project was smaller.

core-code:
> I’ve been thinking of the mailing list idea for a long while

i see that the overhead might be too much for what we want. if something comes up again where 'everyone' should be informed i guess we can just open a new issue and ping 'everyone'

ran-dall:
> i see that the overhead might be too much for what we want. if something comes up again where 'everyone' should be informed i guess we can just open a new issue and ping 'everyone'

I would agree with this statement. I'm looking over the app, and it's really great work; but I worry about the logistics of putting up an app if we're bugging about a mailing list.

> I’ve been thinking of the mailing list idea for a long while. But I doubt it’d have enough subscribers to justify the work —...

Couldn't we simply set up a Google Group, like the way Vagrant (and HashiCorp) do it? I would concur with anyone that thinks it's a true mailing list, but I think it's a low effort way we (contributors and maintainers) can have conversations and keep everyone involved in the loop (which I think at least solves the communication problem). Frankly, it's something that would take about 10m to setup. I can set it up and open an issue with the instructions to how to request access. Once it's set up it's not that hard to manage either.

vitorgalvao:
> I worry about the logistics of putting up an app if we're bugging about a mailing list.

A mailing list requires work from the maintainers, while the app does not.

> Frankly, it's something that would take about 10m to setup. Once it's set up it's not that hard to manage either.

But it is something else to *use*. I don’t think that’s necessary. I’m with the idea of simply pinging people.

ran-dall:
> A mailing list requires work from the maintainers, while the app does not.

👉 I'm a bit confused as to how? It's a distributed mailing list. Once it's set up, it doesn't require more work. Are you making an argument that maintaining an Apple Developer account for HBC is less work?

> But it is something else to use. I don’t think that’s necessary. I’m with the idea of simply pinging people.

👉 I'm a bit concerned about the idea of having an `app`, as it would require someone to maintain the `certificate` required and the associated `account`. Not to mention, it's a pretty big expectation to require someone to install something 'else' on their phone to partake in developer conversations. There's also a number of other issues, such as; the archiving of these conversations, the transparency of these conversations, and the public accessibility of the conversation (just to name a few.) 

---
@vitorgalvao  👉 Now, let me clarify my position. _I'm only respectfully disagreeing that coordination of HBC development should occur through an `app`._ I think the core functionality of the `app` is interesting and can be useful. However, I'm of the position that `dev` talks should be kept as publicly accessible as required to other contributors and maintainers, that may have not been present when certain decisions were taken.

vitorgalvao:
> I'm only respectfully disagreeing that coordination of HBC development should occur through an `app`.

I don’t think anyone is advocating that. [The app is for contributions](https://github.com/Homebrew/homebrew-cask/issues/66601), not conversation.

ran-dall:
> I don’t think anyone is advocating that. The app is for contributions, not conversation.

@vitorgalvao 👉 I'm glad we're on the same page. 😅 However, I'd like to point out the that originally this conversation started becasue I wasn't a part of https://github.com/Homebrew/brew/pull/6155 or https://github.com/Homebrew/brew/pull/5972 (which is partially my fault for not paying attention close enough). My initial request was simply to shoot me a `@` shout so that I can keep up the ML with these changes; cause as it was standing before, this just added 'noise' for the formula that it didn't understand. Any 'noise' of this type causes the security features of the ML to go off because it doesn't 'fit' any of the logic it's be taught.

core-code:
> I don’t think anyone is advocating that. The app is for contributions, not conversation.

right. and i wasn't advocating it to become something 'official' for HBC. its just something i whipped up in an afternoon because i thought being able to quickly check versionsnumbers in HBC and send a PR for updates might be useful in some "emergencies" when no computer is around.

PR #66581, opened 2019-07-22
Update go-agent from 19.1.0-8469 to 19.6.0-9515

ran-dall:

### DIFF
--- a/Casks/go-agent.rb
+++ b/Casks/go-agent.rb
@@ -1,6 +1,6 @@
 cask 'go-agent' do
-  version '19.1.0-8469'
-  sha256 '113b21b10590c7a066bdd0597825b672bfed8b33b9cda756b9d6fea42a7c7c3e'
+  version '19.6.0-9515'
+  sha256 '1237b6c300e4f8f02967c5ec3abb37efefd6e539469385565f308a5ec4bd566e'
 
   # download.gocd.io/binaries was verified as official when first introduced to the cask
   url "https://download.gocd.io/binaries/#{version}/osx/go-agent-#{version}-osx.zip"


ran-dall:
`.app` no longer available

core-code:
i am not sure if this a reason to drop this update. can you re-open this and the corresponding go-server PR? i think @vitorgalvao should decide how we want to proceed with these casks.

ran-dall:
> i am not sure if this a reason to drop this update. can you re-open this and the corresponding go-server PR? i think @vitorgalvao should decide how we want to proceed with these casks.

@core-code I closed it because I think it would be a better fit for HB as opposed to HBC at this point. I really hate to open these with 32 PRs outstanding, but I'll do it just for you. 😉

core-code:
thanks! if its a problem just let me know and i'll open a similar PR. 

i agree it might be a better fit for HB now. lets see what the maintainers think.

vitorgalvao:
Is this now a CLI-only open-source tool that can be built from source and doesn’t depend on signed binaries? If so, HB.

ran-dall:
> thanks! if its a problem just let me know and i'll open a similar PR.

I didn't mean it like that. I meant to point out that I hated the feeling of further clogging the PR log, not that it matters. lol.

> Is this now a CLI-only open-source tool that can be built from source and doesn’t depend on signed binaries? If so, HB.

I don't think it requires to be built but I do believe it's a CLI-only tool. https://docs.gocd.org/current/installation/install/agent/osx.html

vitorgalvao:
> I don't think it requires

Not *require*, but *can* it be done easily? Otherwise it won’t fit in HB. I suggest you still make the HB PR, and we’ll continue from there.

ran-dall:
> Not require, but can it be done easily? Otherwise it won’t fit in HB. I suggest you still make the HB PR, and we’ll continue from there.

You got it!

PR #66783, opened 2019-07-26
Add powerpanel v1.1.0 and v1.0.9

scribblemaniac:
I encountered a small problem when making this cask. Version 1.0.9 of the app is for macOS versions 10.12.1 and 10.11, while version 1.1.0 of the app is for macOS versions 10.12.3+. But it doesn't seem that MacOS.version has the level of finesse needed to distinguish between 10.12.1 and 10.12.3. And don't even ask me what to do if you're on 10.12.2. I went with using the lower version for all macOS 10.12 versions because I figure it's more likely that mac versions are backwards compatible than the app being backwards compatible. However if we are assuming that our users on the latest minor version, then sierra users should use 1.1.0.

### DIFF
--- a/Casks/powerpanel.rb
+++ b/Casks/powerpanel.rb
@@ -1,11 +1,6 @@
 cask 'powerpanel' do
-  if MacOS.version <= :sierra
-    version '1.0.9'
-    sha256 '7b021f9cf182afd04fa25522cee115b271e462f73fa9c76571681e6b4ebda36d'
-  else
-    version '1.1.0'
-    sha256 '9ef3160ad516cf246647a11bdbeaa756a98b2c0d9915ccabbf0d6cc72bebdf51'
-  end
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Removed the conditional. No need to complicate something we don’t even know if HBC users want.

vitorgalvao (approved):

scribblemaniac:
@vitorgalvao What do you mean? Homebrew is supposed to support mavericks+. MacOS.version is used in plenty of casks. Now the depends_on is wrong too.

vitorgalvao:
> I was under the impression that Homebrew Cask is supposed to support mavericks+.

We *can* support it. Officially we only support the current version and one back.

There’s no sense in increasing the maintenance burden from the start without knowing if users want it. If someone needs it, we can add it then.

PR #67184, opened 2019-08-05
Update protonvpn from 1.5.4 to 1.5.5

suschizu:


suschizu:
`MissedUpdateML`

ran-dall:
Actually, we need ignoring `ProtonVPN` because it's causing problems and it's taken too much time out of my day to maintain. Please don't try to override the ignore, when I already reported as so in the thread of the ML.

suschizu:
Hey @ran-dall - I can't find any problems with version 1.5.5 - everything works fine for us -  I've updated the appcast that contains now the right version and download … perhaps that helps.
I'm sorry - we can't blacklist casks from the sparkle-output for future releases but I will keep an eye on it

PR #67190, opened 2019-08-06
Downgrade textmate from 2.0-rc.29 to 2.0-rc.10

core-code:

### DIFF
--- a/Casks/textmate.rb
+++ b/Casks/textmate.rb
@@ -4,7 +4,8 @@ cask 'textmate' do
 
   # github.com/textmate/textmate was verified as official when first introduced to the cask
   url "https://github.com/textmate/textmate/releases/download/v#{version}/TextMate_#{version}.tbz"
-  appcast 'https://api.textmate.org/releases/release?os=10.14.6'
+  appcast 'https://api.textmate.org/releases/release?os=10.14.6',
+          configuration: version
   name 'TextMate'
   homepage 'https://macromates.com/'
 
--- a/Casks/textmate.rb
+++ b/Casks/textmate.rb
@@ -1,10 +1,10 @@
 cask 'textmate' do
-  version '2.0-rc.29'
-  sha256 '49c14380728d8bf1af1e09298fa616a09ae535f5f106967361b84f6ec81d6249'
+  version '2.0-rc.10'
+  sha256 'b8058fd562701b3bb7aab6bb7b1f3a5ac752f153fe258121aa54033cb154d20b'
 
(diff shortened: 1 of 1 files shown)


core-code:
i know we've had this before.

there is a mountain of evidence that rc-10 should be considered the last 'stable' version and all later versions should be considered 'unstable'

1.) the download link on the homepage leads to rc10 ( https://macromates.com )
2.) the download latest link on the downloads page leads to rc10 ( https://macromates.com/download )
3.) the update check built into the app picks up rc10 but does NOT pick up later releases ( https://api.textmate.org/releases/release?os=10.14.6 )
4.) on the github release page rc10 is listed as 'Latest Release' while all newer versions are listed as "Pre-Release" ( ( https://github.com/textmate/textmate/releases  )

so we have 4 points of evidence that show the latest version in homebrew cask should be rc10. i suggest reverting the version to rc10 and adding 'textmate-latest' to the versions-repository since it seems many people are interested in tracking unstable versions here.

core-code:
@tiiiecherle @adruiza @ran-dall @schwedenmut pinging the previous contributors of this cask for discussion

core-code:
if we merge this, i'll probably want to update the appcast configuration. as it is, it will look for 2.0, which lets any RC release pass.

ran-dall:
I have no opinion really, as I don't use TextMate (and really don't have any involvement with them, at all); although it does seem these releases are stable and widely-found elsewhere, but I don't have a problem reverting. 

The logic on the `ML` checks out on this one because they have released `RC`s and `beta`s in the past and we've been merging `Pre-releases` since before the `ML`. Given that the `outputs` put it up and the `appcast` check out; it will submit. I can add this to the `ML` exception list if you (@core-code) deem it necessary.

adruiza:
I believe your approach is correct, as it is consistent with TextMate's repository. Maybe it is not as convenient for some users, but this way we stick to what its developers consider 'stable' and avoid problems in the future.

schwedenmut:
The way you suggested seems pretty compliant for me. Also providing the opportunity to use `textmate-latest` for unstable-like release candidates.

core-code:
thanks for your feedback all. i think this looks ready to merge unless @vitorgalvao or @reitermarkus object

core-code:
i've sent a request to add the latest version of TextMate to Homebrew Cask Versions here:

https://github.com/Homebrew/homebrew-cask-versions/pull/7722

is anyone of the previous contributors interested in keeping this cask up to date with the latest 'unstable' version over there?

PR #67286, opened 2019-08-07
Update nvidia-geforce-now from 2.0.8.85,3AB74C to 2.0.10.71,350E76

core-code:


core-code:
we would basically need a way to insert the previous cask version into the appcast.

vitorgalvao:
We don’t need the previous version, only the current one. While there isn’t a new one, the result of the appcast is `[]`.

core-code:
> We don’t need the previous version, only the current one. 
> While there isn’t a new one, the result of the appcast is [].

isn't that the point? if we insert the current one the result is [] (no update) which makes the CI check fail. and if we add configuration: no_check, *anything* will pass....

hmmmm maybe if we add 'configuration: "[]" it will look for the expected result (no update) and this is exactly what we need? ;)

core-code:
it was meant more as a joke than as a serious suggestion, but if it works, cool
;)

PR #67418, opened 2019-08-10
Revert malwarebytes from 3.9.23.2814 to 3.8.17.2526 [Update pulled by Vendor!]

core-code:


core-code:
actually malwarebytes has "pulled" the  3.9.23.2814 update because it was causing kernel panics. 
it is not available as download in the app or from the homepage anymore.
i think we should downgrade too.

core-code:
here is a URL:
https://forums.malwarebytes.com/topic/250138-just-updated-to-3923-and-have-issues/?do=findComment&comment=1327989

PR #65722, opened 2019-07-04
Analytics cask purge

vitorgalvao:

PR #67450, opened 2019-08-11
Readd instasizer.rb

winkelsdorf:
Reverts single removal from PR #64037.


winkelsdorf:
Readding mentioned in discussion of #65722.

PR #54020, opened 2018-10-23
Update background-music to 0.2.0

brianbrownton:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #67524, opened 2019-08-13
Update background-music from 0.3.0 to 0.3.1

muzimuzhi:

### DIFF
--- a/Casks/background-music.rb
+++ b/Casks/background-music.rb
@@ -14,5 +14,8 @@ cask 'background-music' do
   uninstall launchctl: 'com.bearisdriving.BGM.XPCHelper',
             pkgutil:   'com.bearisdriving.BGM',
             quit:      'com.bearisdriving.BGM.App',
-            script:    '/Applications/Background Music.app/Contents/Resources/_uninstall-non-interactive.sh'
+            script:    {
+                         executable: '/Applications/Background Music.app/Contents/Resources/_uninstall-non-interactive.sh',
+                         sudo: true,
+                       }
 end
--- a/Casks/background-music.rb
+++ b/Casks/background-music.rb
@@ -16,6 +16,6 @@ cask 'background-music' do
             quit:      'com.bearisdriving.BGM.App',
             script:    {
                          executable: '/Applications/Background Music.app/Contents/Resources/_uninstall-non-interactive.sh',
-                         sudo: true,
+                         sudo:       true,
                        }
 end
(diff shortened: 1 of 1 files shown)


muzimuzhi:
background-music 0.3.0 has a problem that it can not launch, see https://github.com/kyleneideck/BackgroundMusic/issues/233.
0.3.1 fixes this.

kyleneideck:
I'm not sure why this update would make the CI build fail. It looks like it failed at `brew cask uninstall background-music` because it timed out waiting for a password, but the uninstall script used in the cask (`_uninstall-non-interactive.sh`) hasn't been changed since 2017.

muzimuzhi:
- The merged pr #54020, which updated background-music from 0.2.0 to 0.3.0, also failed the ci.  
 - When pr #55413, which updated from 0.1.1 to 0.2.0, is merged, ci is not deployed for this repo.

Maybe there is some problem between background-music and the uninstall script.

reitermarkus:
What's inside the `_uninstall-non-interactive.sh` script?

kyleneideck:
https://github.com/kyleneideck/BackgroundMusic/blob/master/BGMApp/BGMApp/_uninstall-non-interactive.sh

It deletes 

- `/Applications/Background Music.app` The app
- `/Library/Audio/Plug-Ins/HAL/Background Music Device.driver` The virtual audio driver
- `/usr/local/libexec/BGMXPCHelper.xpc` and/or `Library/Application Support/Background Music/BGMXPCHelper.xpc` A helper for communication between the app and the driver.
- `/Library/LaunchDaemons/com.bearisdriving.BGM.XPCHelper.plist` The launchd service for the helper (and unloads it)
- The `_BGMXPCHelper` user and group

And restarts `coreaudiod`, the Core Audio daemon.

Actually, I just noticed that the first thing it does is this:

```
# Ensure that the user can use sudo. (But not if this is a Travis CI build, because then it would fail.)
if ([[ -z ${TRAVIS:-} ]] || [[ "${TRAVIS}" != true ]]) && ! sudo -v; then
  echo "ERROR: This script must be run by a user with administrator (sudo) privileges." >&2
  exit 1
fi
```

But I think that's probably not working when the Travis build uses `brew cask uninstall background-music` because the `TRAVIS` env var might not be set in the environment `_uninstall-non-interactive.sh` runs in. The `sudo -v` command would cause a password prompt on Travis: https://github.com/travis-ci/travis-ci/issues/9608#issuecomment-438452070

I'll change the script so it uses `sudo true` instead of `sudo -v`, since apparently that doesn't cause the same problem on Travis.

reitermarkus:
@kyleneideck, you can revert your fix. We weren't running the script with `sudo`.

PR #67528, opened 2019-08-13
Update flash-npapi from 32.0.0.223 to 32.0.0.238

andyli:


andyli:
Adobe's appcast has been known to lagging everytime they push a new version of Flash. It is now updated so the build should pass now. But I wonder if the CI can be more forgiving to the Flash family...

andyli:
Note that there are a few related PRs: #67529 #67530 #67531 #67532 #67533

PR #67653, opened 2019-08-15
Update microsoft-word from 16.27.19071500 to 16.28.19081202

joh-klein:

### DIFF
--- a/Casks/microsoft-word.rb
+++ b/Casks/microsoft-word.rb
@@ -1,6 +1,6 @@
 cask 'microsoft-word' do
   version '16.28.19081202'
-  sha256 '0EB8D554C22047A543E3ED0433AEF41B6E5B9C32EECD5ACBDD12D30564CB3410'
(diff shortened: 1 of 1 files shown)


reitermarkus:
These should have a `conflicts_with cask: 'microsoft-office'`.

joh-klein:
Do they actually conflict?

reitermarkus:
Do they not install Word to the same location?

joh-klein:
Yes, true – hmm, is that already considered a conflict even though they are the same apps?

joh-klein:
This should probably also be added to the microsoft-office cask, right?

reitermarkus:
Yes, `conflicts_with` has to be added to both casks which conflict with each other.

reitermarkus:
> is that already considered a conflict even though they are the same apps?

Doesn't matter whether or not it is the same app, if it installs to the same location, it is definitely a conflict.

PR #67726, opened 2019-08-16
Delete kdiff3.rb

vitorgalvao:
Software is five years old, and development has now [moved to Github](https://github.com/KDE/kdiff3) with no compiled binaries available. Unless compiled versions are provided, we’ll have to remove the cask.

> so users don’t get stuck using an old version, wrongly thinking they’re using the most up-to-date one (which, amongst other things, might be a security risk).


designorant:
Could we consider bringing it back?

Running any software is a security risk, regardless whether it's up to date or not, and the old version, still on SourceForge, is still very much usable.

Related:
```
❯ brew search kdiff3
It was migrated from homebrew/core to homebrew/cask.
You can access it again by running:
  brew tap homebrew/cask
And then you can install it by running:
  brew cask install kdiff3

❯ brew tap homebrew/cask

❯ brew cask install kdiff3
Error: Cask 'kdiff3' is unavailable: No Cask with this name exists.
```

vitorgalvao:
> Running any software is a security risk, regardless whether it's up to date or not

But there are different degrees of risk, and this is where we drew the line. It’s been serving us well and I see no reason to make an exception in this case. If you want it, make a [tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

designorant:
> But there are different degrees of risk, and this is where we drew the line. It’s been serving us well and I see no reason to make an exception in this case. 

Thanks @vitorgalvao, makes sense. I reached out to the authors to see whether there's a chance for binaries for macOS so we can keep the cask alive. I'll open a PR when/if that happens.

vitorgalvao:
> I reached out to the authors to see whether there's a chance for binaries for macOS so we can keep the cask alive. I'll open a PR when/if that happens.

That works.

vitorgalvao:
> I reached out to the authors to see whether there's a chance for binaries for macOS so we can keep the cask alive. I'll open a PR when/if that happens.

That works.

In the meantime, the erroneous message telling users to `brew cask install kdiff3` has been removed. Thank you for letting us know.

designorant:
> In the meantime, the erroneous message telling users to brew cask install kdiff3 has been removed. Thank you for letting us know.

I did noticed, thanks!

cbandy:
@designorant is there a link to you reaching out (a pull request, mailing list thread, etc?) I'd like to follow if possible.

designorant:
@cbandy Sadly not. I reached out to Joachim Eibl over email, and was told the new fork is only for KDE. He pointed me to Michael Reeves for further questions but I didn't get back from him, I'm afraid.

PR #67822, opened 2019-08-19
Update middleclick from 1.0 to Mojave

ran-dall:

### DIFF
--- a/Casks/middleclick.rb
+++ b/Casks/middleclick.rb
@@ -1,9 +1,9 @@
 cask 'middleclick' do
-  version '1.0'
-  sha256 '74533304ec68f8e6c716a4b7c3ad6ed34d985b88c66eee7e43cce22883dee08e'
+  version 'Mojave'
+  sha256 'a83a329493b9140cb73b418ddfa2e921555fc7706a11d8ec2ba13c4016392aca'
 
   # github.com was verified as official when first introduced to the cask
-  url "https://github.com/cl3m/MiddleClick/releases/download/#{version}/MiddleClick.zip"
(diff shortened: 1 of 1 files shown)


Casks/middleclick.rb
```diff
@@ -1,9 +1,9 @@
 cask 'middleclick' do
-  version '1.0'
-  sha256 '74533304ec68f8e6c716a4b7c3ad6ed34d985b88c66eee7e43cce22883dee08e'
+  version 'Mojave'
```
reitermarkus: Not sure if we should consider this a version.

suschizu: the bundle version still is 1.0 and the download url should be:
https://github.com/cl3m/MiddleClick/releases/download/mojave/MiddleClick_Mojave.zip

ran-dall: `1.0,Mojave:mojave`? Seems impractical though, IMO.

core-code: 1.0,mojave should suffice since its not case-sensitive.

vitorgalvao: We have other casks that have the OS version in the URL. The problem with adding `mojave` to the version is that we have no idea when or if it will change. Unless we can gleam a pattern (too soon for that), there’s no point in adding `mojave` to the version.

vitorgalvao (approved):

vitorgalvao:
Not worth it to spend too much time on this. The app is no longer maintained (see the README). Presumably the “Mojave” update was a simple bump to make it work on the latest macOS.

core-code:
is there really a need to wait for mantainer feedback here? with the latest updates the cask looks good, or?

Issue #67806, opened 2019-08-18
Make rmdir: recursive?


danielbayley:
### Description of feature/enhancement

Make [`uninstall`/`zap`] `rmdir:` recursive, so that it will remove a tree of empty directories that exist under the given path… this should probably also include any `.DS_Store` files under this revised definition of _empty_.

### Justification

This would simplify [`uninstall`/`zap`] `rmdir:` in Casks in [the many] cases where an App leaves behind a tree of empty directories.

### Example use case

~~~ rb
zap rmdir: '~/Library/Application Support/Vendor'
~~~
instead of:
~~~ rb
zap rmdir: [
             '~/Library/Application Support/Vendor/Empty Folder',
             '~/Library/Application Support/Vendor',
           ]
~~~

reitermarkus:
Do we not already support globbing for `rmdir`?

reitermarkus:
Indeed we do:

```ruby
zap rmdir: '~/Library/Application Support/Vendor{,/Empty Folder}'
```

danielbayley:
Ah yes brace expansion does work, but it can get tricky to work out for a tree of nested empty folders beyond one or two… and also this doesn't work if there are any `.DS_Store` files. Surely it would be easier for maintenance if `rmdir` ignored them, and worked recursively from the given single path…?

reitermarkus:
> this doesn't work if there are any `.DS_Store` files

It does:

https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/artifact/abstract_uninstall.rb#L348

danielbayley:
> It does:

Interesting… `rmdir:` often seems to fail; I always assumed it was because of `.DS_Store` files. Still, it would be nice if it was recursive, to save figuring out the exact brace expansion for any/all subfolders… Also @reitermarkus what do you think of handling folder `Icon?`s in the same way as `.DS_Store` files? As these can also prevent `rmdir` from working… Maybe something like:

~~~ rb
if (os_file = Dir[resolved_path.join "{.DS_Store,Icon?}"]).any?
  command.run!("/bin/rm", args: ["-f", "--", os_file], sudo: true, print_stderr: false)
end
~~~

reitermarkus:
> what do you think of handling folder `Icon?`

I don't think we should treat it the same as `.DS_Store`; icons need a deliberate action to be generated in the first place, so they should also require the same for removing it.

danielbayley:
[...]

PR #68039, opened 2019-08-23
Improve Adobe Creative Cloud uninstall

danielbayley:
Following on from #67745, which was merged prematurely… this PR cleans up all of the leftover crap from Adobe CC.

### DIFF
--- a/Casks/adobe-creative-cloud.rb
+++ b/Casks/adobe-creative-cloud.rb
@@ -7,21 +7,106 @@ cask 'adobe-creative-cloud' do
   homepage 'https://creative.adobe.com/products/creative-cloud'
 
   installer script: {
-                      executable: "#{staged_path}/Creative Cloud Installer.app/Contents/MacOS/Install",
-                      args:       ['--mode=silent'],
-                      sudo:       true,
+                      executable:   "#{staged_path}/Creative Cloud Installer.app/Contents/MacOS/Install",
+                      args:         ['--mode=silent'],
+                      sudo:         true,
+                      print_stderr: false,
                     }
 
-  uninstall launchctl: [
-                         'com.adobe.AdobeCreativeCloud',
-                         'com.adobe.acc.installer',
-                         'com.adobe.agsservice',
-                         'Adobe_Genuine_Software_Integrity_Service',
-                       ],
-            delete:    [
-                         '/Applications/Adobe Creative Cloud/Adobe Creative Cloud',
-                         '/Applications/Utilities/Adobe Creative Cloud',
-                         '/Applications/Utilities/Adobe Application Manager',
-                       ],
-            rmdir:     '/Applications/Utilities/Adobe Installers'
+  uninstall_postflight do
+    stdout, * = system_command '/bin/launchctl', args: ['print', "gui/#{Process.uid}"]
+    ccx_processes = stdout.lines.grep(%r{com\.adobe\.CCXProcess\.\d{5}}) { $& }.uniq
+    ccx_processes.each { |id| system '/bin/launchctl', 'bootout', "gui/#{Process.uid}/#{id}" }
+  end
+
+  uninstall early_script: {
+                            executable:   '/usr/bin/pluginkit',
+                            args:         ['-r', '/Applications/Utilities/Adobe Sync/CoreSync/Core Sync.app/Contents/PlugIns/ACCFinderSync.appex'],
+                            must_succeed: false,
+                            print_stderr: false,
+                          },
+            launchctl:    [
+                            'Adobe_Genuine_Software_Integrity_Service',
+                            'com.adobe.AdobeCreativeCloud',
+                            'com.adobe.acc.installer',
+                            'com.adobe.acc.installer.v2',
+                            'com.adobe.ccxprocess',
+                          ],
+            quit:         'com.adobe.acc.AdobeCreativeCloud',
+            signal:       [['QUIT', 'com.adobe.accmac']],
+            script:       {
+                            executable:   '/usr/bin/pkill',
+                            args:         ['Adobe Desktop Service', 'AdobeIPCBroker', 'AdobeCRDaemon'],
+                            must_succeed: false,
+                          },
+            delete:       [
+                            "#{appdir}/Adobe Creative Cloud/*Adobe Creative Cloud",
+                            "#{appdir}/Adobe Creative Cloud/Icon?",
+                            "#{appdir}/Adobe Creative Cloud/.Uninstall*",
+                            '/Applications/Utilities/Adobe Creative Cloud*',
+                            '/Applications/Utilities/Adobe Application Manager',
+                            '/Applications/Utilities/Adobe Sync',
+                            '/Applications/Utilities/Adobe Installers/Uninstall Adobe Creative Cloud',
+                            '/Applications/Utilities/Adobe Installers/.Uninstall*',
+                            '/Library/Application Support/Adobe/CEP/extensions/CC_*',
+                            '/Library/Application Support/Adobe/CEP/extensions/com.adobe.ccx.*',
+                            '/Library/Application Support/Adobe/Adobe Desktop Common',
+                            '/Library/Application Support/Adobe/AdobeApplicationManager',
+                            '/Library/Application Support/Adobe/AdobeGC*',
+                            '/Library/Application Support/Adobe/*[Ii]nstall*',
+                            '/Library/Application Support/Adobe/ADCRefs',
+                            '/Library/Application Support/Adobe/caps',
+                            '/Library/Application Support/Adobe/OOBE',
+                            '/Library/Application Support/Adobe/PCF',
+                            '/Library/Application Support/Adobe/SL*',
+                            '/Library/Application Support/Adobe/Vulcan',
+                            '/Library/Application Support/regid.*.com.adobe',
+                            '~/Library/Application Support/Adobe/AAMUpdater',
+                            '~/Library/Application Support/Adobe/ExtensibilityLibrary',
+                            '~/Library/Application Support/Adobe/FloodGate',
+                            '~/Library/Application Support/Adobe/.adobelicnotification',
+                            '~/Library/Application Scripts/com.adobe.accmac.ACCFinderSync',
+                            '~/Library/*/Adobe/CoreSync',
+                            '/Library/*/com.adobe.*.plist',
+                            '~/Library/*/com.adobe.*.plist',
+                            '~/Library/Preferences/Adobe/.[A-Z0-9]???????????',
+                            '/Library/PrivilegedHelperTools/com.adobe.acc.installer*',
+                            '~/Library/Group Containers/Adobe-Hub-App',
+                            '~/Library/*Containers/com.adobe.accmac.*',
+                            '/Library/Internet Plug-Ins/AdobeAAMDetect.plugin',
+                            '~/Creative Cloud Files/Icon?',
+                            '/Users/Shared/Adobe/Installer',
+                            '/Users/Shared/Adobe/OOBE',
+                          ],
+            rmdir:        [
+                            "#{appdir}/Adobe Creative Cloud",
+                            '/Applications/Utilities/Adobe Installers',
+                            '/Library/Application Support/Adobe{/CEP{/extensions,},}',
+                            '/Library/*/Adobe',
+                            '~/Library/*/Adobe',
+                            '~/Creative Cloud Files',
+                            '/Users/Shared/Adobe',
+                          ]
+
+  zap trash: [
+               '/Library/Application Support/Adobe/Extension Manager CC',
+               '~/Library/Application Support/Adobe/Extension Manager CC',
+               '~/Library/Application Support/Adobe/OOBE',
+               '/Library/Application Support/Adobe/Creative Cloud Libraries',
+               '/Library/Application Support/Adobe/Adobe PCD',
+               '/Library/Logs/CreativeCloud',
+               '~/Library/Logs/CreativeCloud',
+               '~/Library/Logs/ACC*.log',
+               '~/Library/Logs/PDApp*.log',
+               '~/Library/Logs/AdobeDownload.log',
+               '~/Library/Logs/AdobeIPCBroker*.log',
+               '~/Library/Logs/CoreSyncInstall.log',
+               '~/Creative Cloud Files',
+             ],
+      rmdir: [
+               '/Library/Application Support/Adobe',
+               '~/Library/Application Support/Adobe',
+               '/Library/Logs/Adobe',
+               '~/Library/Logs/Adobe',
+             ]
 end


danielbayley:
I don't understand why this is [failing](https://github.com/Homebrew/homebrew-cask/pull/68039/checks?check_run_id=201657095#step:3:67) when I have [`must_succeed: false`](https://github.com/Homebrew/homebrew-cask/blob/2f5a40755feae02cae5ba303390dc64603da443b/Casks/adobe-creative-cloud.rb#L22)? It doesn't help that I can't seem to run `brew cask ci` offline…

reitermarkus:
You can run it with `env CI=1 brew cask ci`, if you really need to. `brew cask uninstall` should fail with the same error though.

Casks/adobe-creative-cloud.rb
```diff
@@ -12,16 +12,60 @@
...
+            script:    {
+                         executable:   '/usr/bin/pkill',
+                         args:         ['Adobe Desktop Service', 'AdobeIPCBroker', 'AdobeCRDaemon', 'ACCFinderSync'],
+                         must_succeed: false,
+                       },
             delete:    [
-                         '/Applications/Adobe Creative Cloud/Adobe Creative Cloud',
-                         '/Applications/Utilities/Adobe Creative Cloud',
-                         '/Applications/Utilities/Adobe Application Manager',
+                         "#{appdir}/#{name.first}/{*#{name.first},Icon?}",
```
vitorgalvao: Don’t hook things to `name`. That’s freeform info that can change at any time. Be explicit.

vitorgalvao: Don’t use `{}` as shorthand for multiple paths. In all cases you’re making it work for just two cases, which is not worth the extra complexity of the line.

Casks/adobe-creative-cloud.rb
```diff
@@ -12,16 +12,60 @@
...
+                         "/Applications/Utilities/Adobe Installers/Uninstall #{name.first}",
+                         '/Applications/Utilities/Adobe Installers/.Uninstall*',
+                         '/Library/Application Support/Adobe/CEP/extensions/{CC_*,com.adobe.ccx.*}',
+                         '/Library/Application Support/Adobe/Adobe{ Desktop Common,ApplicationManager,GC*}',
+                         '/Library/Application Support/Adobe/{ADCRefs,caps,*[Ii]nstall*,OOBE,PCF,SL*,Vulcan}',
+                         '/Library/Application Support/regid.*.com.adobe',
+                         '~/Library/Application Support/Adobe/{AAMUpdater,ExtensibilityLibrary,FloodGate}',
+                         '~/Library/Application Support/Adobe/.adobelicnotification',
+                         '~/Library/*/Adobe/CoreSync',
+                         "/{#{Dir.home}/,}Library/*/com.adobe.*.plist",
```
vitorgalvao: As it is, it will translate to `//Users/name` (two forward slashes at the start). This is why the aforementioned complexity isn’t worth it for two paths — mistakes are that mush easier to make.

vitorgalvao: `~`, not `Dir.home`.

Casks/adobe-creative-cloud.rb
```diff
@@ -12,16 +12,60 @@
...
+                         '~/Library/Preferences/Adobe/.[A-Z0-9]???????????',
```
vitorgalvao: Why not just get rid of `~/Library/Preferences/Adobe` altogether?

vitorgalvao (changes requested):

Casks/adobe-creative-cloud.rb
danielbayley: @vitorgalvao Not in all cases… some of them cover 5 or more paths. Are those ones not worth basic brace expansion?

Casks/adobe-creative-cloud.rb
danielbayley: @vitorgalvao Because other Adobe apps store things there… a cask should only remove it's own files.

Casks/adobe-creative-cloud.rb
```diff
@@ -12,16 +12,87 @@
                       sudo:       true,
                     }
 
+  uninstall_postflight do
+    prefix = Regexp.quote 'com.adobe.CCXProcess.'
+    `/bin/launchctl list`.lines.grep(%r{#{prefix}\d{5}}) { $& }.map { |id| `/bin/launchctl remove #{id}` }
```
vitorgalvao: Use `system_command` if that outputs `stdout`, or `Open3.capture2` otherwise, or run the whole command in `system_command` with system tools. We don’t use backticks.

Casks/adobe-creative-cloud.rb
```diff
@@ -12,16 +12,87 @@
                       sudo:       true,
                     }
 
+  uninstall_postflight do
+    prefix = Regexp.quote 'com.adobe.CCXProcess.'
```
vitorgalvao: There’s no point to setting this here as a variable, it could be included in the command itself, and it doesn’t need to be a regex.

Casks/adobe-creative-cloud.rb
vitorgalvao: `remove` is a deprecated `launchctl` command.

vitorgalvao (changes requested):

Casks/adobe-creative-cloud.rb
vitorgalvao: What files does this remove?

Casks/adobe-creative-cloud.rb
vitorgalvao: > some of them cover 5 or more paths. Are those ones not worth basic brace expansion?

Not really, no. There’s already a metric crapton of referenced paths, hiding more behind braces just makes the cask harder to maintain and those paths harder to spot. The braces are useful in cases with few lines, not with many.

danielbayley:
> metric crapton of referenced paths

That’s Adobe for you 🙄

> or run the whole command in system_command with system tools.

@vitorgalvao What do you mean? I don’t see how this is possible given that you can’t use `|` pipes in `system_command`; the second call to `bin/launchctl` works from the `stdout` of the first…

> remove is a deprecated launchctl command.

Looking at the `man launchctl` page… would `kill` be the replacement?

vitorgalvao:
> That’s Adobe for you 🙄

I hear you!

> What do you mean? I don’t see how this is possible given that you can’t use `|` pipes in `system_command`

Yes, you’re right.

> Looking at the `man launchctl` page… would `kill` be the replacement?

`bootout` should do it. But now that I think of it, why aren’t you using `uninstall launchctl` for that?

danielbayley (commented):
> why aren’t you using `uninstall launchctl`

@vitorgalvao Because for the processes I haven’t already included in that stanza, it doesn’t work…
~~~ sh
> brew cask uninstall adobe-creative-cloud
# …
> launchctl print system | grep -i adobe

"com.adobe.GC.Scheduler-1.0" => false
"com.adobe.fpsaud" => true
~~~
and
~~~ sh
> launchctl print gui/$UID | grep -i adobe

		       0      - 	com.adobe.CCXProcess.61720
		       0      - 	com.adobe.CCXProcess.61380
		       0      - 	com.adobe.CCXProcess.61636
		   14864      - 	com.adobe.CCXProcess.61716
		       0      - 	com.adobe.CCXProcess.61532
				"com.adobe.accmac.explinder.ACCFinderInnerExtensionHost2.14846" = {
				"com.adobe.accmac.explinder.ACCFinderInnerExtensionHost2.14845" = {
		 0xcbe3b    U   A   com.adobe.accmac.explinder.ACCFinderInnerExtensionHost2.14845
		 0xed07b    U   A   com.adobe.accmac.explinder.ACCFinderInnerExtensionHost2.14846
	"com.adobe.AdobeCreativeCloud" => true
	"com.adobe.GC.Scheduler-1.0" => false
~~~

Apparently they generate a random 5 digit suffix for each process… just to be even more of a pain in the ass 🙄 Fuck Adobe backwards.

Unfortunately glob patterns aren’t available with `uninstall launchctl:`. Also `signal:` with _any_ of the `SIG`s doesn’t work either… including even `KILL`.

danielbayley:
@vitorgalvao I think this is as good as we’re going to get now with this…

reitermarkus:
> Apparently they generate a random 5 digit suffix for each process

They aren't, this is done by `launchctl` I think, and this should already be handled by:

https://github.com/Homebrew/brew/blob/d5cc46f931c4db200b0ca9e119411dd7a6721da8/Library/Homebrew/cask/artifact/abstract_uninstall.rb#L123

Casks/adobe-creative-cloud.rb
```diff
@@ -7,21 +7,106 @@
...
+                            args:         ['-r', '/Applications/Utilities/Adobe Sync/CoreSync/Core Sync.app/Contents/PlugIns/ACCFinderSync.appex'],
+                            must_succeed: false,
+                            print_stderr: false,
+                          },
+            launchctl:    [
+                            'Adobe_Genuine_Software_Integrity_Service',
+                            'com.adobe.AdobeCreativeCloud',
+                            'com.adobe.acc.installer',
+                            'com.adobe.acc.installer.v2',
+                            'com.adobe.ccxprocess',
```
reitermarkus: Why is this lowercase and `CCXProcess` above?

danielbayley: Because _both_ cases are listed as separate processes.

danielbayley:
> this is done by `launchctl` I think, and this should already be handled

@reitermarkus Interesting. Looks like it _should_, but it isn't… even if I put both
~~~ rb
'com.adobe.CCXProcess',
'com.adobe.ccxprocess',
~~~
in the `launchctl: [`array`]`, without the `uninstall_postflight` code, `com.adobe.CCXProcess` are still left running after `brew cask uninstall`, according to `launchctl list | grep -i adobe`. Could this be a bug with that part of the core code?

This current iteration is the only way I managed to find that both to passes the CI and cosistently removes every process/piece of leftover Adobe crap I could find relating directly to CC.

reitermarkus:
> Could this be a bug with that part of the core code?

The core code is not using `launchctl bootout`.

danielbayley:
> The core code is not using `launchctl bootout`.

@reitermarkus Right, but as mentioned by @vitorgalvao `remove` is a deprecated `launchctl` command… so what to do?

reitermarkus:
> `remove` is a deprecated `launchctl` 

Since when?

danielbayley:
> Since when?

It's listed as a _legacy subcommand_ on the `man` page… but since other code seems to use `list` (another apparently deprecated subcommand) I don't see how this is a particular blocker to this formula. @vitorgalvao brought it up… seems like a separate issue from this PR.

@reitermarkus I’m not sure what the blocker/s is/are now?

reitermarkus:
So is `bootstrap/bootout` a direct replacement for `load/unload`?

>  I’m not sure what the blocker/s is/are now?

That depends on whether we should change the core code to use `bootout` instead of `unload`.

Casks/adobe-creative-cloud.rb
```diff
@@ -7,21 +7,106 @@
...
-            delete:    [
-                         '/Applications/Adobe Creative Cloud/Adobe Creative Cloud',
-                         '/Applications/Utilities/Adobe Creative Cloud',
-                         '/Applications/Utilities/Adobe Application Manager',
-                       ],
-            rmdir:     '/Applications/Utilities/Adobe Installers'
+  uninstall_postflight do
+    stdout, * = system_command '/bin/launchctl', args: ['print', "gui/#{Process.uid}"]
+    ccx_processes = stdout.lines.grep(%r{com\.adobe\.CCXProcess\.\d{5}}) { $& }.uniq
+    ccx_processes.map { |id| system '/bin/launchctl', 'bootout', "gui/#{Process.uid}/#{id}" }
```
reitermarkus: Use `each` here.

Casks/adobe-creative-cloud.rb
```diff
@@ -7,21 +7,106 @@
   homepage 'https://creative.adobe.com/products/creative-cloud'
 
   installer script: {
-                      executable: "#{staged_path}/Creative Cloud Installer.app/Contents/MacOS/Install",
-                      args:       ['--mode=silent'],
-                      sudo:       true,
+                      executable:   "#{staged_path}/Creative Cloud Installer.app/Contents/MacOS/Install",
+                      args:         ['--mode=silent'],
+                      sudo:         true,
+                      print_stderr: false,
```
reitermarkus: What is printed on `stderr`?

danielbayley: > What is printed on `stderr`?

~~~ sh
objc[79807]: Class HTTPHeader is implemented in both /usr/local/Caskroom/adobe-creative-cloud/latest/Creative Cloud Installer.app/Contents/MacOS/Install (0x10efc96c8) and /tmp/A92AD873-9AA3-4227-A553-AEABCD702C76/CCLBS/PDIM/AdobePIM.dylib (0x112cdcb70). One of the two will be used. Which one is undefined.
objc[79807]: Class ProxyManager is implemented in both /usr/local/Caskroom/adobe-creative-cloud/latest/Creative Cloud Installer.app/Contents/MacOS/Install (0x10efc9718) and /tmp/A92AD873-9AA3-4227-A553-AEABCD702C76/CCLBS/PDIM/AdobePIM.dylib (0x112cdcbc0). One of the two will be used. Which one is undefined.
~~~

danielbayley:
> So is `bootstrap/bootout` a direct replacement for `load/unload`?

I'm not sure on the details of this… you'll have to ask @vitorgalvao who raised it.
 
> That depends on whether we should change the core code to use `bootout` instead of `unload`.

Surely that's a seperate issue to this PR… why now all of a sudden is this an issue? @vitorgalvao

reitermarkus:
> why now all of a sudden is this an issue?

Because then we wouldn't need the workaround in `uninstall_postflight` in the first place.

danielbayley:
> Because then we wouldn't need the workaround in `uninstall_postflight` in the first place.

I wouldn't be so sure about that… cleaning up these Adobe processes feels like playing whack-a-mole at best! @reitermarkus Are you saying that you think because the underlying code behind the `launchctl:` stanza is [presumably] using `unload`, that if this was updated to use `bootout` then `launchctl:` would be sufficient to unload all of the given bundle IDs?

reitermarkus:
> I wouldn't be so sure about that…

Well, I'm not. 

> if this was updated to use `bootout` then `launchctl:` would be sufficient

Probably, but that's the open question.

> Have there been other casks

I don't remember any, probably only Adobe manages to pull that shit off.

reitermarkus (approved):

danielbayley:
> I don't remember any, probably only Adobe manages to pull that shit off.

😅 Probably 🙄

> Probably, but that's the open question.

So this PR is basically now reliant on a new separate issue, which is to update all of the Homebrew internals to remove apparently deprecated `launchctl` commands?

vitorgalvao (approved):

Casks/adobe-creative-cloud.rb
```diff
@@ -7,21 +7,106 @@
...
+                            '~/Library/*Containers/com.adobe.accmac.*',
+                            '/Library/Internet Plug-Ins/AdobeAAMDetect.plugin',
+                            '~/Creative Cloud Files/Icon?',
+                            '/Users/Shared/Adobe/Installer',
+                            '/Users/Shared/Adobe/OOBE',
+                          ],
+            rmdir:        [
+                            "#{appdir}/Adobe Creative Cloud",
+                            '/Applications/Utilities/Adobe Installers',
+                            '/Library/Application Support/Adobe{/CEP{/extensions,},}',
```
vitorgalvao: I still think some of these are a bit hard to parse, but fine, it’s Adobe, that you went through all this trouble is already great. Thank you. Let’s get this merged.

danielbayley: > I still think some of these are a bit hard to parse

@vitorgalvao Yeah brace expansion can be tricky… this line is an example of why I opened https://github.com/Homebrew/homebrew-cask/issues/67806.

PR #11351, opened 2015-05-24
Added pdftk 2.02

winkelsdorf:
Cask for pdftk 2.02 package with uninstall and zap methods

PR #68107, opened 2019-08-24
Add cask for PDFTK Server

crishoj:

#11351 was closed with the message:

> Let's wait if someone is able to rewrite this formula to a cask, that should really do the trick.

So here's a cask for the app, including a `post_flight` block which restores group write permission on `/usr/local/bin`, since the PDF installer seems to change it to `0755`. I'm not using `set_permissions ` from the mini-DSL, since that method works recursively.


vitorgalvao:
Thank you, but this doesn’t really solve the issue, it works around it, and the problem it causes is not a small one. I’d rather not include this until the developer fixes that problem.

PR #68159, opened 2019-08-26
Update meshmixer to add appcast stanza

core-code:


core-code:
homepage not reachable error is not really connected to adding the appcast

PR #68160, opened 2019-08-26
Update intel-power-gadget to add appcast stanza

core-code:


core-code:
travis error
`installer: Error - Intel® Power Gadget does not currently support systems with more than one CPU package.
`
does not seem connected to adding an appcast

PR #68183, opened 2019-08-26
Update appcast.md

core-code:
We have two methods to use unstructured (HTML) appcasts to keep apps up-to-date

1.) we can watch the URL for changes in its content and
2.) we can see if the current version number is contained in the URL

method 1.) only works if the URL isn't noisy. sadly, nearly all HTML pages are "too" noisy to observe for changes, although we have a custom regexp list of 200 or so patterns to ignore.

method 2.) only works if version strings of old versions aren't contained in the URL in addition to the current version. 

obviously we have a problem if both methods don't work, then the appcast is basically useless to keep the app up-to-date. thats why i have submitted a number of PRs to replace appcasts of "ChangeLog" HTML pages with other pages that do not contain old version strings.

however, this endeavour is pretty useless if our documentation still recommends adding links to ChangeLog HTML pages - more and more will be added.

i suggest
1.) changing the documentation to suggest best practices to avoid using ChangeLog HTML pages as appcast. i am not sure about the exact wording here and just wanted to send a PR to get things rolling
2.) also, i'd suggest adding a "rule" that new casks that are added, should always contain an appcast if they have a version number (i.e. aren't ':latest'). i guess the first step here would also be to modify the default cask template to contain an appcast stanza? should i send a separate PR for this?


vitorgalvao:
> i'd suggest adding a "rule" that new casks that are added, should always contain an appcast if they have a version number (i.e. aren't ':latest').

Apps that only provide an unversioned download link that redirects to a versioned one might be problematic for that, unless we add your server scripts as another instruction step in these rules. Unsure if that’s desirable (could any abuse come from that?)

> i guess the first step here would also be to modify the default cask template to contain an appcast stanza?

I’ve never used that command, so it makes little difference to me, but I feel like it should only include the bare minimum stanzas for a working cask.

However, we may add “searched for an appcast” to the “adding a new cask” section of the [pull request template](https://github.com/Homebrew/homebrew-cask/blob/master/.github/PULL_REQUEST_TEMPLATE.md). I’m wary of making it mandatory, but I’m fine with nudging users in that direction.

core-code:
> Apps that only provide an unversioned download link that redirects to a versioned one might be problematic for that, 
> unless we add your server scripts as another instruction step in these rules. Unsure if that’s desirable (could any abuse come from that?)

i couldn't imagine much abuse, apart from crawling bots hammering the server. but i agree that we may not want to add these instructions at this point.
"rule" is probably a strong word anyway. all i wanted to say is that it would be great if a new versioned cask is added, it should also have an appcast, if thats possible.
the reason there is, in a few weeks i'll be finished adding appcasts for all the casks on the list that i've generated a few months ago. however i didn't really intend on revisiting this periodically check new casks that have been added. 

> I’ve never used that command, so it makes little difference to me,
> but I feel like it should only include the bare minimum stanzas for a working cask.

ok, agreed

> However, we may add “searched for an appcast” to the “adding a new cask” section of the pull request template.
> I’m wary of making it mandatory, but I’m fine with nudging users in that direction.

that sounds like a good idea!

PR #68215, opened 2019-08-27
Update snipaste to add appcast stanza

core-code:


core-code:
for reference, this should also work` https://macupdater.net/cgi-bin/check_urls/check_url_redirect.cgi?url=https://dl.snipaste.com/mac-beta`

PR #68328, opened 2019-08-29
Update uniconverter from 11.1.3.1,735 to 11.1.1.2

core-code:


core-code:
1.) the download has been downgraded to we need to go back to version 11.1.1.2
2.) i've looked at the apps from wondershare and other chinese companies which are basically re-branding the same things. they all have numbers in the URL, which we've put into the version stanza. however, those numbers (735 in this case) have NEVER changed since those casks have been first submitted between 2014 and 2016. since they are obviously no version numbers and have never changed so far, there is no reason to keep them in the version stanza just because they are numerical

PR #68333, opened 2019-08-29
Update thunderbird.rb from 60.8.0 to 68.0

suschizu:
language stanza update


ffloimair:
I have already filed the same https://github.com/Homebrew/homebrew-cask/pull/68330

PR #68674, opened 2019-09-05
Update jamovi to update appcast

core-code:
not really happy about this change, but this cask has been updated to the beta version a lot and i don't really see another way to prevent that


vitorgalvao:
Doesn’t really help they name versions `solid` and `current`.

PR #68915, opened 2019-09-11
Add VirtualBox 5.2.32

maxmiller:
Add VirtualBox 5.2.32.
Vagrant works better with this version.

### DIFF
--- a/Casks/virtualbox5
+++ b/Casks/virtualbox5
@@ -1,4 +1,4 @@
-cask 'virtualbox@5.2' do
+cask 'virtualbox5' do
   version '5.2.32,132073'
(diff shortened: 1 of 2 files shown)


vitorgalvao:
Welcome and thank you, but we won’t accept this.

1. File is lacking an extension.
2. Older versions belong in [homebrew/cask-versions](https://github.com/Homebrew/homebrew-cask-versions).
3. [But](https://github.com/Homebrew/homebrew-cask-versions#acceptable-casks): “is not intended to be used for all and any old versions you personally require. Casks submitted here should be expected to be used by a reasonable number of people and supported by contributors long-term”.

PR #69182, opened 2019-09-16
Delete i4tools.rb

vitorgalvao:
Contains malware. See https://github.com/Homebrew/homebrew-cask/issues/56223#issuecomment-527985439 and https://github.com/Homebrew/homebrew-cask/issues/56223#issuecomment-531824750.

Leaving it open for a short while in case someone has relevant information we’re missing.


ran-dall (approved):

ran-dall:
@vitorgalvao @core-code I did more investigating 🧐, while I was waiting for the ML to finish the updates tonight. It would appear the main focus of this particular jailbreak utility is allowing the user to access pirated iOS applications (I think it might have also been mentioned that that article you posted @core-code)(but, I don't live or know much about iOS in China, so it may have other legitimate uses, such as bypassing some government regulation on a specific app?); all of which, I don't think this is something that we don't deal with; so I'm for the deletion. 

It's also old, especially with on the verge of iOS 13; this was for iOS 10.

PR #69345, opened 2019-09-19
Updated inkscape.rb to remove X11 dependency

vikrantrathore:
Inkscape just released 1.0 beta version 1 which is a native mac os x application and remove the need for xquartz.

Updated the cask to include 1.0 beta version.

### DIFF
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -1,8 +1,8 @@
 cask 'inkscape' do
   version '1.0beta1'
-  sha256 '6939e99761009befa23a85b8f0d117c1df83934bb1d5fcb8084b4040e122aa66'
+  sha256 '68831989b3919e3137d5acfb130a844933706748addc8b4cd7f957348c1c60a3'
 
-  url "https://inkscape.org/gallery/item/14919/Inkscape-#{version}.dmg"
+  url "https://inkscape.org/gallery/item/14919/Inkscape-#{version}_OEMhoXK.dmg"
   name 'Inkscape'
   homepage 'https://inkscape.org/'
 
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -8,6 +8,5 @@ cask 'inkscape' do
 
   app 'Inkscape.app'
 
-  zap trash: '~/Library/Application Support/Inkscape',
-      rmdir: '~/Library/Application Support/Inkscape'
+  zap trash: '~/Library/Application Support/Inkscape'
 end
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -8,7 +8,7 @@ cask 'inkscape' do
 
   app 'Inkscape.app'
 
-   zap trash: [
+  zap trash: [
                '~/Library/Application Support/Inkscape',
              ],
       rmdir: [
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -1,12 +1,10 @@
 cask 'inkscape' do
-  version '0.92.2-1'
-  sha256 'faece7a9a5fa9db7724b0c761f7f2014676d00ef8b90a0ef506fa39d09209fea'
+  version '1.0beta1'
+  sha256 '6939e99761009befa23a85b8f0d117c1df83934bb1d5fcb8084b4040e122aa66'
 
-  url "https://media.inkscape.org/dl/resources/file/Inkscape-#{version}-x11-10.7-x86_64.dmg"
+  url "https://inkscape.org/gallery/item/14919/Inkscape-#{version}.dmg"
   name 'Inkscape'
   homepage 'https://inkscape.org/'
-
-  depends_on x11: true
 
   app 'Inkscape.app'
   binary "#{appdir}/Inkscape.app/Contents/Resources/bin/inkscape"
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -8,10 +8,6 @@ cask 'inkscape' do
 
   app 'Inkscape.app'
 
-  zap trash: [
-               '~/Library/Application Support/Inkscape',
-             ],
-      rmdir: [
-               '~/Library/Application Support/Inkscape',
-             ]
+  zap trash: '~/Library/Application Support/Inkscape',
+      rmdir: '~/Library/Application Support/Inkscape'
 end
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -7,7 +7,11 @@ cask 'inkscape' do
   homepage 'https://inkscape.org/'
 
   app 'Inkscape.app'
-  binary "#{appdir}/Inkscape.app/Contents/Resources/bin/inkscape"
 
-  zap trash: '~/.inkscape-etc'
+   zap trash: [
+               '~/Library/Application Support/Inkscape',
+             ],
+      rmdir: [
+               '~/Library/Application Support/Inkscape',
+             ]
 end
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -1,8 +1,18 @@
 cask 'inkscape' do
-  version '1.0beta1'
-  sha256 '68831989b3919e3137d5acfb130a844933706748addc8b4cd7f957348c1c60a3'
+  if MacOS.version <= :mavericks
+    version '0.92.2-1'
+    sha256 'faece7a9a5fa9db7724b0c761f7f2014676d00ef8b90a0ef506fa39d09209fea'
 
-  url "https://inkscape.org/gallery/item/14919/Inkscape-#{version}_OEMhoXK.dmg"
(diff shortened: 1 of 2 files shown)


amyspark:
Hello @vikrantrathore !

Only one Cask per PR, please! Since this PR deals with Inkscape, please remove all changes involving `scribus` and push again.

vikrantrathore:
@amyspark 
Removed the scribus.rb changes now it containes only inkscape.rb.

amyspark:
Reviewing the little URL change. Please stand by.

amyspark (approved):

jemcclin:
Noticed when testing this out that it no longer installs the CLI to `/usr/local/bin/inkscape`. This broke a couple of our workflows locally while we were testing upgrades to Inkscape, as we were still expecting the binary to be present.

Would it be worth adding a `caveat` to alert any users installing this cask of the change?

That said: seems the CLI was moved to `/Applications/Inkscape.app/Contents/MacOS/Inkscape` as a wrapper around `/Applications/Inkscape.app/Contents/MacOS/Inkscape-bin`. Seems like almost how it used to be, although just copying over the `inkscape` file leads to errors about `inkscape-bin` that didn't appear in 0.92.2, despite the similar setup:
```/usr/local/bin/inkscape: line 196: /usr/Contents/MacOS/inkscape-bin: No such file or directory
/usr/local/bin/inkscape: line 196: exec: /usr/Contents/MacOS/inkscape-bin: cannot execute: No such file or directory```

I suspect it's due to how it's being called, but haven't had a chance to dig deeper.

**Comment by vitorgalvao** [comment]
> Would it be worth adding a `caveat` to alert any users installing this cask of the change?

[No](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md#caveats).

**Comment by reitermarkus** [comment]
Since this is a beta it should be submitted to the `cask-versions` tap.

**Comment by vikrantrathore** [comment]
> > Since this is a beta it should be submitted to the cask-versions tap.
> 
> No need, we’ll be [getting a better one from upstream](https://github.com/Homebrew/homebrew-cask-versions/pull/7765#issuecomment-537149542).

@vitorgalvao I personally think beta published on inkscape website is still better than nightly builds, but if you think nightly builds are better than published build on inkscape by all means go ahead.

I just thought that inkscape without X11 dependency is better than having one with X11 dependency which people are reluctant to use.

Once you confirm you have taken care of this cask updates I will close this PR.

**Comment by vladimyr** [comment]
> > > Since this is a beta it should be submitted to the cask-versions tap.
> > 
> > 
> > No need, we’ll be [getting a better one from upstream](https://github.com/Homebrew/homebrew-cask-versions/pull/7765#issuecomment-537149542).
> 
> @vitorgalvao I personally think beta published on inkscape website is still better than nightly builds, but if you think nightly builds are better than published build on inkscape by all means go ahead.
> 
> I just thought that inkscape without X11 dependency is better than having one with X11 dependency which people are reluctant to use.

As a longtime Inkscape user, here are my 2 cents...

There are two things that could act as driving force for Inkscape update:
1. As @vikrantrathore mentioned people don't find using X11/Quartz pleasant experience
2. Once you upgrade to macOS Catalina you don't have choice anymore because stable version of Inkscape is 32bit app and thus you can't run it anymore

That being said there is upcoming 1.0 release channel (consisting of multiple releases): https://inkscape.org/release/inkscape-1.0/?latest=1 that has reached `beta1` level: https://inkscape.org/release/inkscape-1.0beta1/

Those releases are certainly more stable than CI/CD artifacts (apart from other technical issues) which @dehesselle also pointed out and suggested creating something between `beta` and `nightly` channel in form of manually curated `inkscape-preview`.

While I sincerely applaud Inkscape team for their effort and clearly see appeal of running bleeding edge software on advanced user's machine this isn't something that average users should install in order to solve neither 1. or 2. Instead there should be (additional) `inkscape-beta` cask, basically what @vikrantrathore proposed here. Unaware of this PR I did exactly the same here: https://github.com/Homebrew/homebrew-cask-versions/pull/8025

**Review by reitermarkus — APPROVED** [review]
Given that this cask is essentially useless as is on Catalina, I think we should switch to the beta.

**Comment by vladimyr** [comment]
>Given that this cask is essentially useless as is on Catalina, I think we should switch to the beta.

Am I reading this correctly, you are proposing to update stable cask to point to 1.0beta1? If so keep in mind that while it kinda solves issue for all Catalina users:
1. It delivers false promise because that version has never been officially marked as stable
2. Users that haven't upgraded yet loose ability to install (preferred?) stable and working version

File: Casks/inkscape.rb [hunk 1]
```diff
@@ -1,15 +1,12 @@
 cask 'inkscape' do
-  version '0.92.2-1'
-  sha256 'faece7a9a5fa9db7724b0c761f7f2014676d00ef8b90a0ef506fa39d09209fea'
+  version '1.0beta1'
+  sha256 '68831989b3919e3137d5acfb130a844933706748addc8b4cd7f957348c1c60a3'
 
-  url "https://media.inkscape.org/dl/resources/file/Inkscape-#{version}-x11-10.7-x86_64.dmg"
+  url "https://inkscape.org/gallery/item/14919/Inkscape-#{version}_OEMhoXK.dmg"
```
vladimyr: You can use direct link to avoid http jumps:
`https://media.inkscape.org/dl/resources/file/Inkscape-#{version}_OEMhoXK.dmg`

vitorgalvao:
Here’s a compromise:

```ruby
cask 'inkscape' do
  if MacOS.version <= :mavericks
    version '0.92.2-1'
    sha256 'faece7a9a5fa9db7724b0c761f7f2014676d00ef8b90a0ef506fa39d09209fea'

    url "https://media.inkscape.org/dl/resources/file/Inkscape-#{version}-x11-10.7-x86_64.dmg"

    depends_on x11: true
  else
    version '1.0beta1'
    sha256 '68831989b3919e3137d5acfb130a844933706748addc8b4cd7f957348c1c60a3'

    url "https://inkscape.org/gallery/item/14919/Inkscape-#{version}_OEMhoXK.dmg"
  end

  name 'Inkscape'
  homepage 'https://inkscape.org/'

  app 'Inkscape.app'

  zap trash: '~/Library/Application Support/Inkscape'
end
```

That way we’re doing [what upstream recommends](https://inkscape.org/release/inkscape-0.92.4/mac-os-x/dmg/dl/).

image

vladimyr:
FWIW I see it as really good compromise 🥇👍

leafac (approved):

vitorgalvao:
No objections were voiced, so I’ve changed the cask to the compromise version.

Casks/inkscape.rb
```diff
@@ -1,15 +1,22 @@
 cask 'inkscape' do
-  version '0.92.2-1'
-  sha256 'faece7a9a5fa9db7724b0c761f7f2014676d00ef8b90a0ef506fa39d09209fea'
+  if MacOS.version <= :mavericks
```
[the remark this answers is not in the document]
vitorgalvao: It was a mistake; thought of one, wrote the other.

vitorgalvao:
Merging. Thank you for kickstarting this, @vikrantrathore.

Casks/inkscape.rb
```diff
@@ -1,15 +1,26 @@
...
+    depends_on x11: true
+
+    binary "#{appdir}/Inkscape.app/Contents/Resources/bin/inkscape"
+
+    zap trash: '~/.inkscape-etc'
+  else
+    version '1.0beta1'
+    sha256 '68831989b3919e3137d5acfb130a844933706748addc8b4cd7f957348c1c60a3'
+
+    url "https://inkscape.org/gallery/item/14919/Inkscape-#{version}_OEMhoXK.dmg"
```
vladimyr: This still applies: https://github.com/Homebrew/homebrew-cask/pull/69345#discussion_r337721193

core-code:
i agree with @vladimyr 
 according to the homepage, the current version of inkscape is still 0.92.4, while 1.0beta1 is only labeled as an "upcoming" release. per our policy we should have 0.92.4 until upstream declares 1.0 as their official version.

Casks/inkscape.rb
reitermarkus: Can you submit a PR?

reitermarkus:
@core-code, we are only using 1.0 on Catalina since 0.92 is 32-bit.

vitorgalvao:
> we are only using 1.0 on Catalina since 0.92 is 32-bit.

More specifically, we’re using the specific versions upstream recommends on the macOS versions they recommend (https://github.com/Homebrew/homebrew-cask/pull/69345#issuecomment-546357258).

core-code:
ah thanks there, and sorry for the noise. currently at dialup speeds here.

Casks/inkscape.rb
vladimyr: Done #71852

PR #69368, opened 2019-09-19
Update oracle-jdk-javadoc from 12.0.2,10:e482c34c86bd4bf8b56c0b35558996b9 to 13,33:5b8a42f3905b406298b72d750b6919f6

bondolo:

### DIFF
--- a/Casks/oracle-jdk-javadoc.rb
+++ b/Casks/oracle-jdk-javadoc.rb
@@ -7,7 +7,7 @@ cask 'oracle-jdk-javadoc' do
                  'oraclelicense' => 'accept-securebackup-cookie',
                }
   name 'Oracle Java Standard Edition Development Kit Documentation'
-  homepage "https://www.oracle.com/technetwork/java/javase/documentation/jdk#{version.major}-doc-downloads-4417029.html"
+  homepage "https://www.oracle.com/technetwork/java/javase/documentation/jdk13-doc-downloads-5672541.html"
 
   artifact 'docs', target: "/Library/Java/JavaVirtualMachines/jdk-#{version.before_comma}.jdk/Contents/Home/docs"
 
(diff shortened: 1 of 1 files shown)


Casks/oracle-jdk-javadoc.rb
```diff
@@ -1,13 +1,13 @@
...
+  version '13,33:5b8a42f3905b406298b72d750b6919f6'
+  sha256 '8e9848af68ee227438661cfb73740de94e11755234432ceae04f86a04b1c7e15'
 
   url "https://download.oracle.com/otn-pub/java/jdk/#{version.before_comma}+#{version.after_comma.before_colon}/#{version.after_colon}/jdk-#{version.before_comma}_doc-all.zip",
       cookies: {
                  'oraclelicense' => 'accept-securebackup-cookie',
                }
   name 'Oracle Java Standard Edition Development Kit Documentation'
-  homepage "https://www.oracle.com/technetwork/java/javase/documentation/jdk#{version.major}-doc-downloads-4417029.html"
+  homepage 'https://www.oracle.com/technetwork/java/javase/documentation/jdk13-doc-downloads-5672541.html'
```
vitorgalvao: Don’t remove the interpolation. That makes the cask harder to maintain.

vitorgalvao (changes requested):

Casks/oracle-jdk-javadoc.rb
bondolo: per the comment on the second commit the interpolation is not valid. The final segment of the url "5672541" is not predictable. As a result the only way to get the url is to visit the page and get the url literal. The URL can't be generated via interpolation. Since cut-and-paste is the only way to provide the URL I chose to remove interpolation entirely for this case.

vitorgalvao: Why not use https://docs.oracle.com/en/java/javase/13/?

bondolo: That could work and [https://www.oracle.com/technetwork/java/javase/documentation/](https://www.oracle.com/technetwork/java/javase/documentation/) seems reasonable as well. Either would be more useful than original homepage for this cask.

vitorgalvao: Then please make the change to whichever of the two is more appropriate.

bondolo: Done. Thank you

vitorgalvao (approved):
Thank you.

PR #69445, opened 2019-09-20
Update mathtype from 7.4.2 to 7.4.3

ran-dall:

### DIFF
--- a/Casks/mathtype.rb
+++ b/Casks/mathtype.rb
@@ -1,6 +1,6 @@
 cask 'mathtype' do
-  version '7.4.2'
-  sha256 'd2b2bfcd37f212edd1d7241d11ecc345cafe6adffcc2fc7a0e1565afb2759e89'
+  version '7.4.3'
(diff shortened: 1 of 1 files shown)


Casks/mathtype.rb
```diff
@@ -1,6 +1,6 @@
 cask 'mathtype' do
-  version '7.4.2'
-  sha256 'd2b2bfcd37f212edd1d7241d11ecc345cafe6adffcc2fc7a0e1565afb2759e89'
+  version '7.4.3'
```
vitorgalvao: Why not `7.11.0`?

ran-dall: Cause I think it verified it with the output request.

ran-dall: I'll fix it shortly.

suschizu: hey @ran-dall - you were absolutely right with version 7.4.3 (#69554)

ran-dall: @suschizu  It's the ML, but it's good to know that it's working correctly. Every day getting closer to a public release. ⭐️

PR #69508, opened 2019-09-22
Update osbuddy.rb

core-code:
try to fix download


core-code:
homepage not reachable - not connected to change

PR #69509, opened 2019-09-22
Delete marp.rb

core-code:
the download is no longer available, it seems to have been replaced with something called 'marp-next' but i didn't find mac binaries for that


colindean:
Just in case someone else finds this in search of where the Marp cask went…

Marp Next is the new re-imagining of Marp but it doesn't have a desktop version nor does it work like the old Marp did. It's now mostly a CLI tool with a Desktop version planned but not started.

Here's what I had to do to get a presentation up and running with Marp Next: https://github.com/colindean/talks/commit/16a4b1803554cdb63d37889bbd38cd044fa3be61

Issue #68730, opened 2019-09-06
/usr/local/lib permission discrepancies


tb94:
#### General troubleshooting steps

- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [x] I have retried my command with `--force`.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

#### Description of issue

I'd like to preface that i've looked for an open issue to comment on with further information, but all of the relevant issues were closed. I'm not sure if this is an issue with virtualbox or with homebrew, but i haven't seen this observation anywhere else and wanted to provide it where it can be easily found and potentially used to solve the issue. if it should be moved somewhere else, i would gladly do so.

I've seen more than a few reports of people running into an issue resembling [this post on stack overflow](https://stackoverflow.com/questions/27784545/brew-error-could-not-symlink-path-is-not-writable)
mainly where a directory in /usr/local is owned by root and cannot be written to by homebrew.

This has been plaguing my development environment setup script for days until i realized that this only happens when a set of instructions are performed in a specific order. I noticed that the first thing my script did after installing homebrew was install virtualbox, yet when i installed git instead of virtualbox `brew install git` directly after installing homebrew , these directories are created with the correct permissions.

this is on a fresh installation of Mojave (10.14.6)
with a fresh installation of homebrew (2.1.11)

#### Command that failed
Starting after macOS finishes installing Mojave:
```
$ ls -al /usr/local
total 0
drwxr-xr-x  3 root wheel  96 Sep 6 10:16 .
drwxr-xr-x@ 9 root wheel 288 Sep 6 10:16 ..
-rw-r--r--  1 root wheel   0 Sep 6 10:16 .com.apple.installer.keep

$ ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
...
$ ls -al /usr/local/lib
total 0
drwxrwxr-x   2 developer admin  64 Sep 6 10:21 .
[...]

PR #69601, opened 2019-09-24
virtualbox.rb: fix /usr/local/lib/pkgconfig permissions

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/issues/68730#issuecomment-534363026


amyspark (approved):
LGTM (CI is expected to fail because [it needs authorization to load the kexts](https://developer.apple.com/library/archive/technotes/tn2459/_index.html).)

PR #69610, opened 2019-09-24
Update mendeley.rb from 1.19.4 to 2.12.0

suschizu:
this app was renamed from mendeley desktop to mendeley reference manager.
quite hard to find this out because the website doesn't offer proper links 
but there is a description on the new apps website to upgrade:
https://www.mendeley.com/download-reference-manager/macOS
release notes https://www.mendeley.com/release-notes-reference-manager


reitermarkus:
The cask itself should also be renamed.

core-code (approved):

PR #69650, opened 2019-09-25
Add Samsung DeX 1.0.0.54

kirb:

- [X] Attempted to find an [appcast].

[appcast]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md

----

This is the DeX on PC client that provides a VNC-like Samsung DeX session from the phone over USB.

Some notes:

* The pkg forces a reboot. I tried manually `kextload`ing `ssuddrv.kext` after allowing it in the Security preferences, but it still didn’t work till I rebooted. Manual `kextload` won’t work before the user manually allows the kext publisher anyway.
* The uninstaller app (requires user input to continue, and won’t exit with failure on error, so Homebrew will think it’s uninstalled when it isn’t) can very likely be replaced with Cask uninstall stanzas. Aside from uninstalling all the pkgs, killing the processes, and unloading the kext, I don’t think it does anything special.
* Plenty of references to DevMate, but no DevMate appcast that I can see. Closest I can see is from the Info.plist, a URL that returns an auth error:

```xml
<key>LiveUpdate_Mode</key>
<string>A</string>
<key>LiveUpdate_URL</key>
<string>https://dopc-api.linuxondex.com</string>
```

### DIFF
--- a/Casks/samsung-dex.rb
+++ b/Casks/samsung-dex.rb
@@ -1,0 +1,20 @@
+cask 'samsung-dex' do
+  version '1.0.0.54'
+  sha256 'e6541ed3dc7aea4eb5534b92207812c1ec7b1bf4e312a504be134a30e60dbb00'
+
+  url 'http://downloadcenter.samsung.com/content/SW/201909/20190911142048616/SamsungDeXSetup.dmg'
+  name 'Samsung DeX'
+  homepage 'https://www.samsung.com/global/galaxy/apps/samsung-dex/'
+
+  pkg 'Install Samsung DeX.pkg'
+
+  uninstall script: {
+                      executable: "#{staged_path}/Uninstall.app/Contents/MacOS/Uninstall",
+                      sudo:       true,
+                    }
+
+  caveats do
+    kext
+    reboot
+  end
+end


kirb:
Hah, I put the file in the root dir instead of in Casks/. Amended with that fixed.

Casks/samsung-dex.rb
```diff
@@ -0,0 +1,20 @@
...
+  sha256 'e6541ed3dc7aea4eb5534b92207812c1ec7b1bf4e312a504be134a30e60dbb00'
+
+  url 'http://downloadcenter.samsung.com/content/SW/201909/20190911142048616/SamsungDeXSetup.dmg'
+  name 'Samsung DeX'
+  homepage 'https://www.samsung.com/global/galaxy/apps/samsung-dex/'
+
+  pkg 'Install Samsung DeX.pkg'
+
+  uninstall script: {
+                      executable: "#{staged_path}/Uninstall.app/Contents/MacOS/Uninstall",
```
vitorgalvao: This is not a script, it’s just calling the app through the command-line. Please use the other uninstall options (`uninstall pkgutil:` and whatever else is necessary).

vitorgalvao (changes requested):

Casks/samsung-dex.rb
kirb: Yes, I mentioned this. I’ll work on it then.

vitorgalvao:
Closing for lack of progress.

kirb:
After some time I realised the USB DeX client is awful to use, compared to attaching keyboard/mouse/monitor directly to the phone. Even using USB-C over a Thunderbolt cable, it uses H.264 fixed at 1920x1080 on the video with unusable latency, by contrast a directly attached monitor runs at native resolution with no compression. I don’t think it’s right to have this in Cask, it feels like just a proof of concept/padding out a feature list. So I agree with closing this.

PR #69664, opened 2019-09-25
Update navicat-for-postgresql

nfourtythree:


core-code:
this is incorrect as the download will not currently download version 12.1.25

nfourtythree:
I've just realised that is because the latest version is `12.1.24`

core-code:
it was 12.1.25, but they have reverted back to 12.1.24 for some reason, causing the SHA mismatch

PR #69864, opened 2019-09-28
fix for issue #70 - showing the application window and not only the icon for meld version 3.19.2-r6,osx-15

electriquo:
fix for showing the application window and not only the icon.
see [issue #70](https://github.com/yousseb/meld/issues/70) for more information.

[appcast]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md

### DIFF
--- a/Casks/meld.rb
+++ b/Casks/meld.rb
@@ -16,6 +16,7 @@ cask 'meld' do
   preflight do
     IO.write shimscript, <<~EOS
       #!/bin/sh
+      rm -rf ~/Library/Saved\ Application\ State/org.gnome.meld.savedState
       exec '#{appdir}/Meld.app/Contents/MacOS/Meld' "$@"
     EOS
   end


vitorgalvao:
Thank you, but this is an upstream bug and should be fixed upstream. We don’t like adding these types of workarounds that’ll be forgotten.

dalboris:
@vitorgalvao I was just bitten by this bug, and I think it affects quite a lot of people. May I kindly ask if you may reconsider? It is quite a standard practice in software development to implement workarounds for bugs which are not yet fixed upstream. Python folks would say "practicality beats purity".

But the question really is: do you want your users to have a working installation of "meld" after typing "brew cask install meld"? Because currently, they don't, and they each need to spend dozens of minutes to figure out what's wrong, find the workaround, and apply it. In turn, they get frustrated at Meld but also Homebrew for failing to provide an easy way to install Meld. And you are just one click away to save them hundreds of people from all this trouble (based on the number of likes in the workaround provided in https://github.com/yousseb/meld/issues/70).

vitorgalvao:
> the question really is: do you want your users to have a working installation of "meld" after typing "brew cask install meld"?

Do the meld maintainers? Because that’s a more interesting question. If they don’t (as is their right) and their software becomes unusable, that’s one less broken piece of software we have to maintain as a cask.

Implementing too many workarounds has [landed us in trouble in the past](https://github.com/Homebrew/homebrew-cask/issues/13201) and brought the project to a halt, almost killing it. A lot more people have made arguments similar to yours in past cases, and having the bug fixed upstream proved to be the most beneficial solution. We do implement workarounds for problems we create, but avoid them for upstream issues. Duct taping software instead of fixing it is why everything is buggy. It’s easy to come to HBC in passing, submit a drive-by fix and then go off never to return, leaving us with the mess.

Also, presumably this problem happens both when the app is launched in a GUI and when started via CLI, but this PR only addresses the latter (the former is not fixable in a cask), making it an inadequate solution. If, however, the issue does indeed only happen when the software is started via CLI, we might as well remove the `shimscript` from the cask. If `meld` users are happy with that, I’m also happy doing it.

The linked bug is a year old and ostensibly easy to work around, yet I don’t see any PR on their repo. Why?

vitorgalvao:
@dalboris @y0y0z But please do note I found your requests polite and reasonable. I’m willing to have my mind changed or reach another solution, it’s just that your the current ones don’t fit.

dalboris:
I'd say, although this is just a gut feeling, that most people don't run meld directly, but have their version control system call meld for them. I personally have a git alias and type "git meld", which opens meld with the changes known by git.

And no, this isn't at all "ostensibly easy" to work around. I've personally spent at least 1h to find the right Google keywords, try different workarounds, finally find one that seems to work, but then realise that it worked only once, then try to write my own meld wrapper in `~/bin/` but this didn't worked either, then finally edit the file in `/usr/local/Caskroom/meld/3.19.2-r6,osx-15/meld.wrapper.sh` which finally worked. But I'm not even sure this will last after an update. Then I opened a new document somewhere to document this fix for the next time I need to make a macOS installation, then made a comment here to try to make it easier for other users. All of this in an operating system which is very new to me, as I usually use Linux, which makes every single operation somewhat quite complicated.

Of course, It'd be better if the maintainers of meld fixed it upstream, but maybe they're not familiar with macOS either. Maybe that's a quirk of macOS specifically, and having it fixed within the package manager (i.e., Homebrew) actually makes sense. I don't know without looking at the problem in more details, which I don't really have time right now. All I know is that fixing the problem here and right now, at least partially, doesn't prevent to implement a better and proper fix later.

I hope my point of view only comes as constructive. Deferring responsibility is always the easiest thing to do, but doesn't really help users. I'm not pointing fingers at anyone, just saying that you have the opportunity to fix this in one click (at least partially) and it seems a pity not to do it. Is this one additional line such a maintenance burden that between "adding the line" and "frustrating hundreds of users", you prefer the latter? I just find that surprising.

dalboris:
@vitorgalvao Oh, and yes, thanks for your time and consideration :) I do understand that you don't want homebrew to become a duct tape mess. We all run into similar issues in our own projects, and it's sometimes hard to find where to draw the line. I appreciate that you're willing to look into this for a better fix.

vitorgalvao:
> I'd say, although this is just a gut feeling, that most people don't run meld directly, but have their version control system call meld for them.

So the bug *does* happen when its run via a GUI?

> And no, this isn't at all "ostensibly easy" to work around.

It is *upstream*, even if for now it just means that *they* remove the problematic file *on their side* even if as a temporary workaround; but it’s their workaround and as such has a higher chance of being revisited and fixed.

We want bugs to be fixed for everybody, not worked around only if you install and use the software in a specific way.

> I've personally spent at least 1h to find the right Google keywords, try different workarounds, finally find one that seems to work, but then realise that it worked only once

The solution may have been hard to *find*, but it’s not hard to *perform* (again, upstream).

> but maybe they're not familiar with macOS either.

But there are people on their user base who are. I’ve implemented fixes to casks that were *completely* community driven, suggested by the users of the software that I’m unfamiliar with it.

> Maybe that's a quirk of macOS specifically, and having it fixed within the package manager (i.e., Homebrew) actually makes sense.

I’d agree if we were an official package manager to either Meld or macOS.

> I don't know without looking at the problem in more details, **which I don't really have time right now.**

[quote emphasis mine]

Which is fair and I empathise with, but is also part of the problem.

> All I know is that fixing the problem here and right now, at least partially, doesn't prevent to implement a better and proper fix later.

Because over half a decade of maintaining HBC tells me that the fix has a good chance of lingering on our side.

> I hope my point of view only comes as constructive.

We posted at the same time, but from my previous comment I hope you know that yes, I’m seeing your interaction as positive

> The linked bug is a year old and ostensibly easy to work around, yet I don’t see any PR on their repo. Why?

That’s still a question I’d like to have an answer to before deciding on a workaround from our side.

dalboris:
Oh, I see, you meant ostensibly easy to work around for Meld developers, not end users.

There seems to be some work going on to resolve this issue at least "one step upstream":

https://github.com/yousseb/meld/issues/93
https://github.com/yousseb/meld/commit/dc6b82f781438e0e310eac9a3f1574e2b9699822

y0y0z:
@dalboris: i totally agreen with @vitorgalvao [comment](https://github.com/Homebrew/homebrew-cask/pull/69864#issuecomment-536206169)

PR #69901, opened 2019-09-29
Update phonerescue

core-code:


core-code:
the new URL is the one the installer download from the homepage uses. it points to v3.9 while the old DMG is at v3.8.

PR #70040, opened 2019-10-02
GitHub Actions: updates for data generation

EricFromCanada:
1. Perform a shallow clone when fetching Homebrew/formulae.brew.sh, as is currently done in [homebrew-core](https://github.com/Homebrew/homebrew-core/blob/master/.github/workflows/generate_formulae.brew.sh_data.yml#L25).
2. Ensure we're running off the original homebrew-cask repo and not a fork, as suggested in Homebrew/homebrew-core#44809.


reitermarkus:
Thanks!

PR #70140, opened 2019-10-04
Update ipass

core-code:

### DIFF
--- a/Casks/ipass.rb
+++ b/Casks/ipass.rb
@@ -2,8 +2,8 @@ cask 'ipass' do
   version :latest
   sha256 :no_check
 
-  # d18uyicgt51j26.cloudfront.net/packages was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


core-code:
can't reproduce travis error

vitorgalvao (approved):

vitorgalvao:
@reitermarkus @amyspark CI gives a warning:

> Warning: Skipping quitting application ID 'com.iPass.iPass'. Enable Automation Access for “Terminal > System Events” in “System Preferences > Security > Privacy > Automation”.

Which a few lines down translates to:

> Error: Some applications are still running, add them to uninstall quit:
> com.iPass.iPass

This might become a problem. If we can no longer automate closing an app, checking if an app is still running is useless and will always result in a failure.

vitorgalvao:
I’ll merge because the problem is general, not related to this update, but we need to address the cause.

PR #70099, opened 2019-10-03
Update BBEdit to 13.0

chrisfinazzo:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #70435, opened 2019-10-09
Update if/else, depend_on stanzas to match BBEdit's Cask requirements

chrisfinazzo:
Complement to #70099 and #70102. No change to the version number.

### DIFF
--- a/Casks/bbedit.rb
+++ b/Casks/bbedit.rb
@@ -1,8 +1,5 @@
 cask 'bbedit' do
-  if MacOS.version <= :el_capitan
-    version '12.1.6'
-    sha256 '23b9fc6ef5c03cbcab041566503c556d5baf56b2ec18f551e6f0e9e6b48dc690'
-  elsif MacOS.version <= :high_sierra
+  if MacOS.version <= :high_sierra
     version '12.6.7'
     sha256 'd0647c864268b187343bd95bfcf490d6a2388579b1f8fce64a289c65341b1144'
   else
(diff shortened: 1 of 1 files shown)


core-code:
as far as i can see you aren't fixing anything, just removing support for 10.11 and 10.12

chrisfinazzo:
> as far as i can see you aren't fixing anything, just removing support for 10.11 and 10.12

🙄 Anyways...

See @MikeMcQuaid’s comment re Cask’s policy on compatibility.

core-code:
i see that we aren't trying to add support for such old systems, but why remove it? bbedit is quite popular and 10.12 is still in use.

chrisfinazzo:
Yes, however Mike’s comment in the original PR is self explanatory. In addition to Apple dropping support for these versions, Catalina has shipped, hence change.

core-code:
as far as i know, the "we support" means we don't actively add support via if/else for old systems unless there are good reasons in a particular case. 

i don't think it means that we should actively go out and make popular casks worse. but thats just my opinion and how i understood the "rules" so far. 

@vitorgalvao is there consensus that we should really change all casks to actively drop support for sierra and lower?

MikeMcQuaid:
> as far as i know, the "we support" means we don't actively add support via if/else for old systems unless there are good reasons in a particular case.

I'm not a Homebrew Cask maintainer but: yes, that's what I think too. It means we'll close bug reports for those systems and remove dead code but we won't remove code that's working fine. 

You will note Homebrew/homebrew-core still has code that's only run on Mavericks.

> 🙄 Anyways...

@chrisfinazzo Not sure the eye rolling emoji was really necessary or helpful there. Let's try to all stay civil, please.

vitorgalvao:
What @core-code and @MikeMcQuaid said is correct. Thank you anyway, @chrisfinazzo.

PR #70520, opened 2019-10-10
Update pages-data-merge.rb

core-code:
change to the 'mojave' url. WARNING: this is officially declared as 'beta' on the homepage, but it might still be better as the current version won't run on 2/3 of our officially supported macOS versions.

### DIFF
--- a/Casks/pages-data-merge.rb
+++ b/Casks/pages-data-merge.rb
@@ -2,7 +2,7 @@ cask 'pages-data-merge' do
   version :latest
   sha256 :no_check
 
-  url 'https://iworkautomation.com/pages/PagesDataMergeApp.zip',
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> WARNING: this is officially declared as 'beta' on the homepage

[We’ve got that covered](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version) (emphasis added):

vitorgalvao (approved):

core-code:
i'd go with 'general standard' here ;)

vitorgalvao:
> i'd go with 'general standard' here ;)

That’s my thought as well. I wrote “(emphasis added)” but then forgot to add the emphasis! Fixed now.

PR #70543, opened 2019-10-10
Revert bluegriffon.rb from 3.1 to 3.0.1

suschizu:
min version is 10.13


vitorgalvao:
> min version is 10.13

There’s a disparity between this comment and the change. Did you mean for a `depends_on` instead? Why the version downgrade?

suschizu:
oh sorry - I forgot to add the comment:
the download on the website still contains version 3.0.1 (that depends on 10.6)- 
I think 3.1 is a beta version that depends on 10.13 
http://bluegriffon.org/#download

PR #70745, opened 2019-10-13
Update cocktail from 12.4.2 to 13.0.0.1

core-code:


core-code:
it is labeled as 'beta 1' on the homepage but since it only applies to the catalina-path i think it might be ok?

reitermarkus:
> i think it might be ok?

Yes, if there isn't a stable Catalina version yet, that's fine.

Casks/cocktail.rb
```diff
@@ -29,12 +29,18 @@
 
     url "https://www.maintain.se/downloads/sparkle/highsierra/Cocktail_#{version}.zip"
     appcast 'https://www.maintain.se/downloads/sparkle/highsierra/highsierra.xml'
-  else
+  elsif MacOS.version <= :high_sierra
```
reitermarkus: I think this should have been `:mojave`.

PR #70777, opened 2019-10-14
Improve installation for prince

cprecioso:
Previously, this cask delegated to a bundled installation script that does the following:
- Ask for a prefix location
- Copy its files wholesale to that location
- Create a shim script that calls the binary with an added `--prefix` option
- Set permissions on the files

As Cask can manage all of that, the installation script is skipped and Cask is instructed on having a similar outcome.

There is however a problem in the Catalina update, Gatekeeper flat out refuses to launch the executable -even from the terminal- as it has the quarantine bit set and is not signed. Should I remove the quarantine bit in a `postflight` script, or leave it as-is?

---

> It doesn't include the version as there's no change to it

### DIFF
--- a/Casks/prince.rb
+++ b/Casks/prince.rb
@@ -3,17 +3,18 @@ cask 'prince' do
   sha256 'd7940c2f60b1e9657db1deb1144b2496cc34c728f650c36b24d6885b964e9aed'
 
   url "https://www.princexml.com/download/prince-#{version}-macosx.tar.gz"
+  appcast 'https://www.princexml.com/download/'
   name 'Prince'
   homepage 'https://www.princexml.com/'
 
-  installer script: "prince-#{version}-macosx/install.sh"
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/prince-#{version}-macosx/prince.wrapper.sh"
+  binary shimscript, target: 'prince'
 
-  uninstall delete: [
-                      '/usr/local/bin/prince',
-                      '/usr/local/lib/prince',
-                    ]
-
-  caveats do
-    files_in_usr_local
+  preflight do
+    IO.write shimscript, <<~EOS
+      #!/bin/sh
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> There is however a problem in the Catalina update

What *isn’t* a problem in that blasted OS?

> Gatekeeper flat out refuses to launch the executable -even from the terminal- as it has the quarantine bit set and is not signed.

Scripts also need to be signed and notarised now? Who’s going to pay $99/year to sign shell scripts?

> Should I remove the quarantine bit in a `postflight` script, or leave it as-is?

Removing the quarantine bit wouldn’t make sense, as we’re the ones who set it (to great pains). If removing the quarantine bit is enough, Catalina is bigger security theatre than I thought. I was under the impression a notarisation check would trigger regardless of quarantining status.

But it isn’t clear to me why this change is being made. Why are you skipping the install script? Is it the install script that gets stopped by Gatekeeper?

cprecioso:
Sorry it wasn’t unclear. I’m skipping the bundled installation script as it installed things in /usr/local that were not managed by Homebrew. 

Their install script copies binaries and supporting files in /usr/local/lib/prince/, and creates a shim script at /usr/local/bin/prince. Their shim script runs fine, the executable it calls (the actual binary of the program) to is the one that trips up Gatekeeper. 

Removing the quarantine bit of the binary does indeed allow it to pass through Gatekeeper. IIRC the notarizarion check does still run but will only kick in if the app is marked as harmful by Apple.

vitorgalvao:
> IIRC the notarizarion check does still run but will only kick in if the app is marked as harmful by Apple.

Which only happens if the app is submitted to them in the first place, making it useless. I’ve since [confirmed](https://lapcatsoftware.com/articles/notarization.html) notarisation is indeed like DRM in games in that it punishes the wrong people.

I do like your changes, but it worried me a bit that by not using the install script we might be missing something when the software updates. On the other hand, we have better control over what it does, and the cask is updated sporadically enough that it shouldn’t be a problem.

On the whole I like it. Thank you for the submission and the explanation.

PR #70877, opened 2019-10-15
Add mysqlworkbench version 6.3.10 for High Sierra users

ottnorml:

### DIFF
--- a/Casks/mysqlworkbench.rb
+++ b/Casks/mysqlworkbench.rb
@@ -1,13 +1,18 @@
 cask 'mysqlworkbench' do
-  version '8.0.18'
-  sha256 '965f85163d1723be26c1f0c74c5b1cd908fac79e02c00fa371c217c9a3bf09ae'
+  if MacOS.version <= :high_sierra
+    version '6.3.10'
+    sha256 '29857bf84bebb7c4442ce147e44602d00f8c001e3c09b3a6e3af356767e08d2c'
+  else
+    version '8.0.18'
+    sha256 '965f85163d1723be26c1f0c74c5b1cd908fac79e02c00fa371c217c9a3bf09ae'
+  end
 
   url "https://cdn.mysql.com/Downloads/MySQLGUITools/mysql-workbench-community-#{version}-macos-x86_64.dmg"
   appcast 'https://dev.mysql.com/downloads/workbench/'
   name 'MySQL Workbench'
   homepage 'https://www.mysql.com/products/workbench/'
 
-  depends_on macos: '>= :mojave'
+  depends_on macos: '>= :high_sierra'
 
   app 'MySQLWorkbench.app'
 


Casks/mysqlworkbench.rb
```diff
@@ -1,13 +1,18 @@
 cask 'mysqlworkbench' do
-  version '8.0.18'
-  sha256 '965f85163d1723be26c1f0c74c5b1cd908fac79e02c00fa371c217c9a3bf09ae'
+  if MacOS.version <= :high_sierra
+    version '6.3.10'
+    sha256 :no_check
```
ottnorml: The checksum should be `29857bf84bebb7c4442ce147e44602d00f8c001e3c09b3a6e3af356767e08d2c`, but `brew cask style --fix` replaced it. How can I fix that?

vitorgalvao: This still needs to be set.

vitorgalvao (changes requested):

Casks/mysqlworkbench.rb
ottnorml: I added the checksum in 6836936.

vitorgalvao:
Looking good. Thank you.

vitorgalvao (approved):

ottnorml:
@vitorgalvao Do you have an idea, why the `xcode10.1` Travis build ([143135.1](https://travis-ci.com/Homebrew/homebrew-cask/jobs/245980573)) is failing?

reitermarkus:
The `appcast` doesn't apply for the High Sierra version.

ottnorml:
@reitermarkus Is this a problem? `find_appcast` did not find anything and I did not change it.

vitorgalvao:
> The `appcast` doesn't apply for the High Sierra version.

Technically it does, because it links back to the `6.3` versions. But presumably version 6 won’t get further updates, so it shouldn’t be a problem (if it is, we can just stick the `appcast` in the conditional).

John-Dennert:
Per https://dev.mysql.com/downloads/workbench/, MySQL Workbench 8.0.18 works with High Sierra. "Packages for Mojave (10.14) are compatible with High Sierra (10.13)"

ottnorml:
> Per https://dev.mysql.com/downloads/workbench/, MySQL Workbench 8.0.18 works with High Sierra. "Packages for Mojave (10.14) are compatible with High Sierra (10.13)"
In this case, it must be new in version 8.0.18, because the program in version 8.0.17 with High Sierra refused to run (see #69926).

ottnorml:
Screenshot
As you can see from the screenshot, version `8.0.18` is not compatible with High Sierra.

ottnorml:
@vitorgalvao  & @reitermarkus How do we continue?

PR #70886, opened 2019-10-15
pr_template: remove appcast item

vitorgalvao:
@core-code It was worth a try, but it didn’t work. Even experienced contributors are failing to add the most obvious appcasts. I’m not sure where the problem lies so I don’t have a fix.

I usually catch the new submissions and I’m taking care to always find an appcast, so might as well remove the item from the template.


core-code:
ok no problem, thanks for taking care of new submissions. 

i have a feeling the "appcast" item isn't the only template item that has been ignored...

PR #69180, opened 2019-09-16
Update amazon-music from 7.7.0,20190826:220957d923 to 7.8.0,20190914:051004b120

suschizu:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #70973, opened 2019-10-16
Update amazon-music from 7.8.3,20190926:081652a3d0 to 7.8.4,20191014:194547a7a5

suschizu:


suschizu:
always the same issue with amazon-music:
https://github.com/Homebrew/homebrew-cask/pull/69180
`Some applications are still running, add them to uninstall quit:
com.amazon.music`
is already added

core-code:
> Some applications are still running, add them to uninstall quit

i've seen a lot of those recently.

@reitermarkus does it use Apple Events to quit the app? i've noticed that recently some apps stopped responding them, e.g. Dropbox will no longer quit when asked to do so via Apple Events.

reitermarkus:
> does it use Apple Events

It uses AppleScript, so I guess so.

PR #70980, opened 2019-10-16
Update zulu from 13.27.9-ca to 13.0.1,13.28.11-ca

joewiz:


joewiz:
Please note that the revised `url` formula was taken from https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/zulu8.rb. It was needed because the target URL doesn't simply repeat the major version, but uses an entirely unique and complete version number.

PR #70984, opened 2019-10-16
Update baidunetdisk from 2.2.3 to 2.2.4

CoolTomatos:

### DIFF
--- a/Casks/baidunetdisk.rb
+++ b/Casks/baidunetdisk.rb
@@ -1,6 +1,6 @@
 cask 'baidunetdisk' do
-  version '2.2.3'
-  sha256 'b656ffb4e188077734f3f08e18cdb25025d16d36cb2f16ea3aa9ac4988853a80'
+  version '2.2.4'
+  sha256 '711fac5fde9af66136ec3c30d0d2ac00e57d6ad9bfaed11ff1aa2dd66666df9c'
 
   # baidupcs.com/issue/netdisk/MACguanjia was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


iceskyzcl:
Can't open baidunetdisk 2.2.4 in macOS Catalina because mac can't check if this app is safe. I find that the newest version on baidunetdisk homepage is still 2.2.3 now? Version 2.2.3 is ok to open.

CoolTomatos:
@iceskyzcl
Two days ago they updated the website to 2.2.4 now somehow they rolled it back. 

With either 2.2.3 or 2.2.4 you have to go the the security panel to allow the access. That’s the case for many third party Mac apps that are not from the App Store.

iceskyzcl:
> @iceskyzcl
> Two days ago they updated the website to 2.2.4 now somehow they rolled it back.
> 
> With either 2.2.3 or 2.2.4 you have to go the the security panel to allow the access. That’s the case for many third party Mac apps that are not from the App Store.

I have allowed the access of installing the third party apps, and i can use many apps installed from brew casks, maybe version 2.2.4 has some problems?

CoolTomatos:
@iceskyzcl
“Allow third party app” only allows apps from notarized developers by Apple. For developers not notarized, in this case baidu, you need to allow access for each single app. 
It’s had been the case since Mojave since Apple removed the option to “allow any third party app”.

iceskyzcl:
@CoolTomatos 
Thanks. I have rolled back to 2.2.3, there's no additional access for this version in both Mojave and Catalina.

PR #71082, opened 2019-10-18
Update fiji

matkozak:
The download is unversioned, but the .dmg package has not been updated in two years. It ships with old Java which cannot be updated from within Fiji.


vitorgalvao:
Thank you, and welcome.

PR #71087, opened 2019-10-18
Update omniplan.rb to 3.13.1 + add cases for sierra and high_sierra

suschizu:
the app is quite popular - so i think this would be ok 
https://update.omnigroup.com/appcast/com.omnigroup.OmniPlan3

### DIFF
--- a/Casks/omniplan.rb
+++ b/Casks/omniplan.rb
@@ -1,10 +1,5 @@
 cask 'omniplan' do
-  if MacOS.version <= :sierra
-    version '3.10.4'
-    sha256 'a30728e72ae970dbf37b2ef9942a6b54267aa3456288dcc1815f20b44667e9e5'
-
-    url "https://downloads.omnigroup.com/software/MacOSX/10.12/OmniPlan-#{version}.dmg"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
We technically don’t support Sierra, and I think neither does Apple, so that’s too far back. I’m fine with High Sierra.

Casks/omniplan.rb
```diff
@@ -1,13 +1,25 @@
...
+    version '3.13'
+    sha256 '82e0d7db2626d751f93f97d80dc032e4bc01bba1e05ea52c553e4771c8cfeec5'
+
+    url "https://downloads.omnigroup.com/software/MacOSX/10.13/OmniPlan-#{version}.dmg"
+  else
+    version '3.13.1'
+    sha256 '631e4ef1a740b5bbc914fb67da007cbabef94c56a1ee21b53af1cc30ac07c72f'
+  end
+
+  url "https://downloads.omnigroup.com/software/MacOSX/10.14/OmniPlan-#{version}.dmg"
```
vitorgalvao: This one needs to be inside the conditional as well.

PR #71194, opened 2019-10-20
Update depends on stanza from el capitan to sierra telegram.rb

suschizu:
depends on 10.12


suschizu:
looks like they have downgraded back to 5.7,181196
should we revert it or just wait a bit?

PR #71245, opened 2019-10-21
Update wiznote from 2.7.8,2019-10-17 to 2.7.9,2019-10-21

ran-dall:

### DIFF
--- a/Casks/wiznote.rb
+++ b/Casks/wiznote.rb
@@ -1,6 +1,6 @@
 cask 'wiznote' do
-  version '2.7.8,2019-10-17'
-  sha256 'e4af518847c6400416cd367833f1add767a0bc997d924acbda8094cc6854abcc'
+  version '2.7.9,2019-10-21'
+  sha256 'd9f089534ff54bce5af36f6caea347a15c94f78d320fffd5fe438ed2da5ad5d4'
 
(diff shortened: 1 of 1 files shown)


ran-dall:
@core-code I think the ML is making a jump with the `version`, cause I can't find it anywhere. `2.7.9` part, the `2019-10-21` part is ok.

core-code:
it seems to contain version 2.7.9 (3407) so it looks ok, can't see a problem

ran-dall:
> it seems to contain version 2.7.9 (3407) so it looks ok, can't see a problem

@core-code The problem is that I think the ML is guessing on these version numbers. It really doesn't have a reason to assume `2.7.9` is `latest`. Which means that eventually this could be a problem. I was just letting you know, because maybe you could find a different `appcast` or change the `appcast` configuration.

core-code:
ah thanks. guessing is indeed bad ;)

we could use 'https://www.wiz.cn/wiznote-mac.html' in this case. it does contain the real version number (before the comma). on the other hand, it doesn't contain the filename (after comma). i don't see a real solution here ;(

ran-dall:
> i don't see a real solution here ;(

Agreed, let just leave it the way it is.

PR #71251, opened 2019-10-22
Test style annotations.

reitermarkus:


MikeMcQuaid:
This is nice. Would be lovely to have similar on Homebrew/brew!

PR #71257, opened 2019-10-22
Update archiver from 3.0.7 to 3.0.8

core-code:


core-code:
the appcast does not have it yet, but i can confirm that the official download from the homepage leads to 3.0.8 for 4 days now

PR #71415, opened 2019-10-24
Add high_sierra case to sourcetree.rb

suschizu:
I think we should keep the  10.13 version


core-code:
also, as the cask is currently it will try to install the 4.0 version on 10.13, which just does not work.

PR #71439, opened 2019-10-25
Update waterfox from 56.2.14 to 2019.10

ran-dall:

### DIFF
--- a/Casks/waterfox.rb
+++ b/Casks/waterfox.rb
@@ -1,9 +1,9 @@
 cask 'waterfox' do
-  version '56.2.14'
-  sha256 'c9df511b7a9474a123f2eba456332744e9bdd8ef0a9fae464441932a6f655bd0'
+  version '2019.10'
+  sha256 '69ff0aca4cd0e58fe583c3807e002f3e5fc460e6de8a5a41413bb50785cfb8b9'
 
   # storage-waterfox.netdna-ssl.com was verified as official when first introduced to the cask
-  url "https://storage-waterfox.netdna-ssl.com/releases/osx64/installer/Waterfox%20#{version}%20Setup.dmg"
+  url "https://storage-waterfox.netdna-ssl.com/releases/osx64/installer/Waterfox%20Current%20#{version}%20Setup.dmg"
   appcast 'https://www.waterfox.net/releases/'
   name 'Waterfox'
   homepage 'https://www.waterfox.net/'
--- a/Casks/waterfox.rb
+++ b/Casks/waterfox.rb
@@ -3,7 +3,7 @@ cask 'waterfox' do
   sha256 '69ff0aca4cd0e58fe583c3807e002f3e5fc460e6de8a5a41413bb50785cfb8b9'
 
   # storage-waterfox.netdna-ssl.com was verified as official when first introduced to the cask
-  url "https://storage-waterfox.netdna-ssl.com/releases/osx64/installer/Waterfox%20Current%20#{version}%20Setup.dmg"
+  url "https://storage-waterfox.netdna-ssl.com/releases/osx64/installer/Waterfox%20Classic%20#{version}%20Setup.dmg"
   appcast 'https://www.waterfox.net/releases/'
   name 'Waterfox'
   homepage 'https://www.waterfox.net/'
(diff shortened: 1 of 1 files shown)


ran-dall:
@core-code @suschizu This needs an `appcast` configuration.

core-code:
the appcast doesn't seem the problem but the app stanza should be 

'Waterfox Current.app'

however, since there are two editions now "Classic" versus "Current", i wonder whether we should have two casks or rename the cask, instead of implicitly switching to "Current". @vitorgalvao @nickl-  any opinions here?

ran-dall:
I personally feel we should keep the `waterfox` because I doubt people will know it's now `waterfox-classic`; unless we also have `waterfox-current` which I think would be redundant. JMO 😁

I also agree with @core-code 's comments about `current` but I think `classic` is the default link on the site, so I changed it to `classic`. 

@vitorgalvao feel free to update it to whatever you think is best. 👌

core-code:
i agree this is the best solution until they re-word the description of 'current' to make it sound less "experimental"

reitermarkus:
I wonder how that conversation went down.

> How should we call the next generation Waterfox? 

> How about Waterfox NG?

> No, let's call it Current.

> Isn't that kind of confusing?

> You're right, we will need to rename the current one to Waterfox Classic.

> Yep, that sounds like the mot straightforward solution to me.

vitorgalvao:
The cask should still be renamed. For one, we always do and I don’t see anything special about this cask to warrant an exception; but also because if we don’t it will be more confusing, as a user won’t know which of the versions it is without inspecting the cask.

grahamperrin:
FYI https://redd.it/drain3 raises awareness of: 

- https://formulae.brew.sh/cask/waterfox-classic
- this merge #71439
- #71601

### Distinguishing between Waterfox Classic 2019.10 and Waterfox Current 2019.10

Re: https://github.com/Homebrew/homebrew-cask/blob/master/.github/support.yml I refrained from raising this as an issue … should encourage replies/discussion [in Discourse](https://discourse.brew.sh/) but for starters I'm taking this opportunity to ping @MrAlex94.

From https://github.com/Homebrew/homebrew-cask/pull/71439#issuecomment-546463054 above: 

> … unless we also have `waterfox-current` which I think would be redundant. JMO :grin:
> 
> I also agree with @core-code's comments about `current` but I think `classic` is the default link on the site, so I changed it to `classic`.

@ran-dall please, do you foresee `waterfox-current` becoming redundant because some word other than _Current_ might be used technically – or colloquially – to describe [the modern, feature updated branch](https://www.waterfox.net/blog/waterfox-2019.10-release-download/)?

@MrAlex94 re: https://github.com/MrAlex94/Waterfox/releases

- https://github.com/MrAlex94/Waterfox/releases/tag/68.0a2 was _Pre-release_
- https://github.com/MrAlex94/Waterfox/releases/tag/2019.10-current-1 is not (but probably should have been) _Pre-release_ and is currently categorised as a release _for testing_ at https://www.waterfox.net/releases/ and of course I expect some number of users to be word blind to the phrase _for testing_ `;-)`

Partly an orientation/documentation issue; https://github.com/MrAlex94/Waterfox/issues/582#issuecomment-546688254

Thanks everyone

ran-dall:
@grahamperrin I'm not sure I understand what you're asking me. I made this comment because the main button the site links to the `classic` release of the download, not `current`.

As if there's a better word then `current`, I don't think it really depends on me but rather it depends on `upstream`. IMO it's a poor choice of vocabulary. That said...

> There are also two versions of Waterfox to choose from. One focused on a legacy branch maintaining security and bug fixes and another branch focusing on keeping up to date with the modern web and a fully customisable web browser to go with it. 

...I'm not sure I really understand this reasoning, but if I understand correctly it's two versions of the same software. Just one has more features than the other?

Regardless, I'm not sure what the HBC stance on something like this would be. Maybe @vitorgalvao could add some insight into what he thinks the HBC stance would be.

grahamperrin:
Thanks, 

>> I should encourage replies/discussion in Discourse 

Here: [Waterfox Project - Homebrew …](https://discourse.brew.sh/t/waterfox-project/6139)

PR #71453, opened 2019-10-25
Upgrade fre:ac from 1.1-alpha-20190423 to 1.1-alpha-20190423a

photomoose:
Fixes compatibility issue with macOS Catalina.

### DIFF
--- a/Casks/freac.rb
+++ b/Casks/freac.rb
@@ -3,7 +3,7 @@ cask 'freac' do
   sha256 'f6cbe820d0c8c114064b1282ea45e256a95df0c4433483043477cfe77c3627f5'
 
   # github.com/enzo1982/freac was verified as official when first introduced to the cask
-  url "https://github.com/enzo1982/freac/releases/download/v#{version.sub(/a$/, '')}/freac-#{version}-macosx.dmg"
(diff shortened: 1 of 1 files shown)


Casks/freac.rb
```diff
@@ -1,9 +1,9 @@
 cask 'freac' do
-  version '1.1-alpha-20190423'
-  sha256 'ce10882ca82198ec84397ab3a2d1fb3292dd9272ab59db2e22226cf09cab2739'
+  version '1.1-alpha-20190423a'
+  sha256 'f6cbe820d0c8c114064b1282ea45e256a95df0c4433483043477cfe77c3627f5'
 
   # github.com/enzo1982/freac was verified as official when first introduced to the cask
-  url "https://github.com/enzo1982/freac/releases/download/v#{version}/freac-#{version}-macosx.dmg"
```
reitermarkus: This still works.

photomoose: @reitermarkus No it doesn't; there's no trailing 'a' in the version after /download.

reitermarkus: Oh, right. Maybe use `.sub(/a$/, '')` on the first interpolation then.

photomoose: @reitermarkus I've replaced the interpolation as suggested.

PR #71514, opened 2019-10-26
Update openrct2 from 0.0.7 to 0.2.3

core-code:


core-code:
note: this is not a straightforward update. what the cask 'openrct2' previously contained was NOT the official 'openrct2' project, but some old and unofficial 'launcher'. i don't think this launcher is needed anymore, and if it is, it should be moved to a cask with a name clarifying what it actually is. please double-check

Issue #70798, opened 2019-10-14
Should we temporarily disable quarantining


vitorgalvao:
Quarantining seems to be the cause of a ton of issues these days, be it problems with `xattr` or with Catalina. From what I can gather so far, Catalina and notarisation requirements are little more than security theatre, as removing (or in our case, not adding) the quarantine attribute bypasses the whole security checks system. [A blog post by Jeff Johnson](https://lapcatsoftware.com/articles/notarization.html) seems to confirm this (scroll down to the last section: “False Benefit: Malware Scan”).

Manually adding the quarantine attribute is a somewhat opaque process (@reitermarkus did research on this early on) and Apple doesn’t provide a tool to make it simple, so we had to devise our own (implemented in Swift by @amyspark), bending over backwards to accommodate Apple’s Gatekeeper security model.

However, it has become abundantly clear that something changed on Apple’s side, and our solution is creating issues for our users — too many for us to keep up with until we get everything straight.

Thus, as a proponent *for* macOS security, I sadly propose we disable the quarantining feature for now, until we can either recover our footing or Catalina’s system changes. We may wish to include a message on every `install`, informing users quarantining is no longer in place, lest they get a false sense of security.

Opening this post to gather opinions from maintainers and users alike.

Pinging @Homebrew/cask.

amyspark:
I do agree that Catalina has made macOS security a mess. The real problem is that **Apple loves changing tools without detailing them in changelogs**. The cornerstone of Gatekeeper is extended attributes, and the only tool to handle them in macOS is `xattr`.

`xattr` has two main issues:

- Apple loves taking away functionality. When Mojave was released, Apple took away its `-R` (recursive) flag, and with Catalina, it started following symlinks blindly. And in neither case there was a changelog.
- It depends on system Python, which is *very* vulnerable with SIP disabled. All issues of "distribution not found", "xattr==0.6.4", "pkg_resources", etc. come up after someone did so and tried installing Python from a pkg.

in re security -- I do not have any problem with Apple introducing whatever changes they deem necessary, but I just want some headsup. 

As for the rest, FTR, I think it's high time we started moving away from this tool in particular. If we keep getting this kind of bug reports, I want to be able to pull the trigger and remove `xattr` usage from Homebrew altogether. I have got a proof of concept in Swift to deal with it.

vitorgalvao:
> It depends on system Python

[...]

PR #71588, opened 2019-10-28
Update yandex-cloud-cli 0.42.0 to install docker-credential-yc

andreiborisov:
Invoke installer during cask install to add second binary.

### DIFF
--- a/Casks/yandex-cloud-cli.rb
+++ b/Casks/yandex-cloud-cli.rb
@@ -7,5 +7,12 @@ cask 'yandex-cloud-cli' do
   name 'Yandex Cloud CLI'
   homepage 'https://cloud.yandex.com/docs/cli/'
 
+  installer script: {
+                      executable: "#{staged_path}/yc",
+                      args:       ['components', 'post-update'],
+                    }
+  binary 'docker-credential-yc'
   binary 'yc'
+
+  uninstall delete: "#{staged_path}/docker-credential-yc"
 end
--- a/Casks/yandex-cloud-cli.rb
+++ b/Casks/yandex-cloud-cli.rb
@@ -13,6 +13,4 @@ cask 'yandex-cloud-cli' do
                     }
   binary 'docker-credential-yc'
   binary 'yc'
-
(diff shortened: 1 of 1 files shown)


schrodincat:
This is not working with Gatekeeper active on macOS Catalina due to refusal to run just downloaded installer script. How can I fix it?

vitorgalvao:
> How can I fix it?

You can’t. Users would need to install this with `--no-quarantine`, or [we’d need to stop quarantining](https://github.com/Homebrew/homebrew-cask/issues/70798), or upstream would have to notarize the tool (which, as far as I’m aware, is even harder for CLI tools), or Apple would need to understand the train wreck they’ve created and remedy the situation (I wouldn’t hold out for this one).

It may be a better option to leave the cask as was. Why is running the script important?

schrodincat:
> Why is running the script important?

It's installing additional binary `docker-credential-yc` which important for some features.

I'd prefer to merge this in some shape or form, I guess I just add this to caveats  for now, so user can install additional binary manually until proper solution is ready.

@vitorgalvao what do you think?

vitorgalvao:
> It's installing additional binary `docker-credential-yc` which is important for some features.

Are those features mandatory? Couldn’t someone presumably use this tool without running that command or installing those tools?

> just add this to caveats for now, so the user can install additional binary manually until a proper solution is ready.

That’s what they have to do now, so this change complicates the cask with diminishing benefits (as more users upgrade, less will take advantage of it). There’s no “proper solution” in the works, so waiting for one isn’t wise.

We’ll pass for now.

vitorgalvao:
> It's installing additional binary `docker-credential-yc` which is important for some features.

Are those features mandatory? Couldn’t someone presumably use this tool without running that command or installing those tools?

> just add this to caveats for now, so the user can install additional binary manually until a proper solution is ready.

That’s what they have to do now, so this change complicates the cask with diminishing benefits (as more users upgrade, less will take advantage of it). There’s no “proper solution” in the works, so waiting for one isn’t wise.

We’ll pass for now. Thank you anyway, and for bringing this to our attention.

PR #71601, opened 2019-10-28
Update and rename waterfox.rb to waterfox-classic.rb

core-code:
https://github.com/Homebrew/homebrew-cask/pull/71439#issuecomment-547049133


vitorgalvao:
That was fast! Thank you.

For further context, this won’t be a disruptive change:

```
waterfox (added 1733 days ago)
30 days: 56 (#1016)
90 days: 189 (#879)
365 days: 805 (#790)
```

PR #71656, opened 2019-10-29
Update veracrypt from 1.24 to 1.24-hotfix1

suschizu:


suschizu:
not sure if this is the best way to update to this version by changing the URL - but the problem is that the download contains 1x hotfix h > lower case and 1x H > upper case … hopefully the next update won't be a hotfix

PR #71773, opened 2019-10-31
Fix installation filename extension

stefanosala:
I just installed HMA and the installation file has a different extension.


vitorgalvao:
Closing for lack of reply.

PR #71800, opened 2019-10-31
Update osu

kazigk:

I got a new version from [there](https://osu.ppy.sh/community/forums/topics/682197?start=7140911). `zap trash` was removed because I couldn't find those files.


vitorgalvao:
Officially, we don’t like adding apps from random forum posts, because we can’t be sure of their authenticity and official status. But the current version seems to be really old and this cask isn’t that popular. We may remove it instead.

Thank you anyway.

PR #71853, opened 2019-11-01
Update intel-haxm from 7.5.2 to 7.5.4

robertleeblairjr:

### DIFF
--- a/Casks/intel-haxm.rb
+++ b/Casks/intel-haxm.rb
@@ -1,6 +1,6 @@
 cask 'intel-haxm' do
-  version '7.5.2'
-  sha256 '20914a1dfac8f4ca7a6a9bf77998f7aa4075384fa88c883e3eb6666661821b81'
+  version '7.5.4'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> uninstall script silent_install.sh does not exist.

Looks like uninstallation will need further revisions.

robertleeblairjr:
> > uninstall script silent_install.sh does not exist.
> 
> Looks like uninstallation will need further revisions.

Don't know why this would occur. The previous silent_install.sh file in version 7.5.2 is the same, other than versioning difference in the reference to the *.dmg file, as version 7.5.4.

> > uninstall script silent_install.sh does not exist.
> 
> Looks like uninstallation will need further revisions.

I don't know why this would occur if there wasn't a problem with the current 7.5.2. Running a 'diff' on the silent_install.sh files for each version shows they are the same other than the version reference to each *.dmg file. I believe this previously required you to push it on version 7.5.2, but I may be mistaken.

PR #71880, opened 2019-11-01
Update veracrypt from 1.24 to 1.24,hotfix1

tiiiecherle:

### DIFF
--- a/Casks/veracrypt.rb
+++ b/Casks/veracrypt.rb
@@ -1,9 +1,9 @@
 cask 'veracrypt' do
-  version '1.24'
+  version '1.24,hotfix1'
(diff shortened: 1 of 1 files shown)


reitermarkus:
Change the interpolation. The URL already contains `hotfix1`.

tiiiecherle:
The problem is that the first -hotfix1 in the url is written in lower letters, the second one -Hotfix1 with a capital H.
I did not find a documentation for {version.before_hyphen}.
https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md

And with leaving the cask as before it would not be detected as an update.

Is this solution better now? I hope this is ok.

PR #71887, opened 2019-11-01
Update depends on stanza from yosemite to sierra couleurs.rb

suschizu:
min version is 10.12


suschizu:
magic error
`Error: undefined method `magic_number' for nil:NilClass`

PR #72010, opened 2019-11-03
Update mpv from 0.29.1 to 0.30.0

luzpaz:


chrisfinazzo:
Weird side effect: Even though I have the Cask installed, I was prompted to `brew install` dependencies, which appears to do nothing...because the Cask is already installed 😕.

Screen Shot 2019-11-05 at 3 18 43 PM

Does someone have the output from `brew missing` for this case?

PR #72164, opened 2019-11-06
Update syncmate from 7.4.452 to 8.0.469

suschizu:


core-code:
>  # cdn.electronic.us/products was verified as official when first introduced to the cask

this is incorrect as this is most likely not enough to identify the download location. there doesn't seem to be a connection from electronic.us to eltima, so it should read

 # cdn.electronic.us/products/syncmate/ was verified as official when first introduced to the cask

PR #72202, opened 2019-11-07
Update folx from 5.11(13873).13873 to 5.12(13890).13890

ran-dall:

### DIFF
--- a/Casks/folx.rb
+++ b/Casks/folx.rb
@@ -1,6 +1,6 @@
 cask 'folx' do
-  version '5.11(13873).13873'
-  sha256 'ba1598b86e6dd0d783e0457aa3808eab5aac560905d8196ef26a5752a621815e'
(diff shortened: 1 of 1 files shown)


core-code:
why the weird version number here?

ran-dall:
> why the weird version number here?

@core-code What do you mean?

core-code:
well, it was changed from a simple version number to something more complicated in #67693

the URL doesn't need a version interpolation at all, so we can basically use what we like and what makes sense


we have
short version string: 5.12
bundle version: 13890

so, either of those, or a simple combination makes sense... however "5.12(13890).13890" seems overly complicated and .. weird

ran-dall:
@core-code Agreed, could you make that PR or should I? Good spot BTW.

core-code:
i can do it no problem ;-)

PR #72260, opened 2019-11-07
Update macupdater

i0ntempest:


vitorgalvao:
Seems correct according to the homepage, but we should check anyway. Is this correct, @core-code, or do you have a reason to prefer the zip?

core-code:
the PR is correct. we've replaced the ZIP file with a DMG file because Catalina contains bugs that can crash 'translocated' apps, and using the signed disk image allows us to prevent 'translocation' from occurring in the first place.

PR #72286, opened 2019-11-08
Update boinc from 7.16.3 to 7.14.2

core-code:


core-code:
ok the 7.16.x versions are all development versions (see here https://boinc.berkeley.edu/download_all.php ) so we should downgrade to the officially stable version.

@randall i've changed the appcast so that beta versions shouldn't be picked up here anymore

@suschizu can you remove 'boinc' from the database check too, to prevent beta versions being picked up again?

suschizu:
all right - will be ignored in the future

PR #72365, opened 2019-11-09
Update launchpad-manager

core-code:


core-code:
this has been messed up in PR 72076. i know its confusing that the lower version number is the newer build, but please refer to the website.

i've erased the if/else, as the "yosemite" build supports everything from 10.10 onwards.  if someone really wants a path for 10.7/10.8/10.9, we can add that, but AFAIK we haven't supported that in a while.

PR #72480, opened 2019-11-12
Updated version and sha256 for cask NoSQL Workbench for Amazon DynamoDB

NoSQLKnowHow:
All checks for existing casks passed.


synchrophoto:
Let me fix the comment and get this correct.

PR #72485, opened 2019-11-12
Update vimediamanager.rb from 0.7a19 to 0.7a20 + url change

suschizu:
hey @vitorgalvao - do you perhaps have an idea how to solve the new url notation?
`url "https://github.com/vidalvanbergen/ViMediaManager/releases/download/v#{version}/ViMediaManager-v720.dmg"`


vitorgalvao:
Did it by removing letters from the string. Let’s see how that scales when the next versions comes around.

suschizu:
great - good to know - thanks :)

PR #72496, opened 2019-11-12
Update version in macs-fan-control.rb from 1.5.1 to 1.5.1.7

suschizu:
would it be possible to take the full version notation for this cask?
there often just a change in the last number


ModdedGamers:
On the github page, it does not appear that there is a 1.5.1.7 version available. If you could point me to this, I would be glad to help you!

suschizu:
hey @ModdedGamers - when you download the app you can see that the app contains version 1.5.1.7 - they haven't changed the download link address but the version changed - there have been 5 versions of 1.5.0 for example and there are already 2 versions of 1.5.1 - so perhaps its better to have the app version in the cask (not just the changed sha)

PR #72716, opened 2019-11-16
nextcloud 2.6.1.20191104-legacy: Install legacy build on 10.10 and 10.11

keeganquinn:
This updates the nextcloud cask to fetch the officially-provided legacy build as needed on OS X 10.10 and 10.11.


Casks/nextcloud.rb
```diff
@@ -1,13 +1,18 @@
 cask 'nextcloud' do
-  version '2.6.1.20191105'
-  sha256 '922f8fbdcef34acf851dc629cc99066962208c32565990bc03fe73c8b1581c92'
+  if MacOS.version >= :sierra
```
reitermarkus: Reverse the order, the `else` should be the latest version.

reitermarkus:
Thanks!

PR #72771, opened 2019-11-18
Add version (2.3) to rightzoom.rb

suschizu:
download has version


core-code:
`Error: Cask 'rightzoom' definition is invalid: 'url' stanza failed with: bad URI(is not URI?): "https://www.blazingtools.com/mac/RightZoom\#{version.no_dots}.zip"`

PR #72754, opened 2019-11-17
Update sqlworkbenchj from 124 to 125

mzemp:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #72958, opened 2019-11-21
Revert "Update sqlworkbenchj from 124 to 125"

suschizu:
Reverts Homebrew/homebrew-cask#72754


suschizu:
see https://github.com/Homebrew/homebrew-cask/pull/72754#issuecomment-556938453

PR #73088, opened 2019-11-23
Update gifox from 010601.03 to 2.0.0,020000.07

core-code:

### DIFF
--- a/Casks/gifox.rb
+++ b/Casks/gifox.rb
@@ -4,7 +4,7 @@ cask 'gifox' do
 
   # d3si16icyi9iar.cloudfront.net/gifox was verified as official when first introduced to the cask
   url "https://d3si16icyi9iar.cloudfront.net/gifox/#{version.after_comma}.dmg"
-  appcast 'https://api.gifox.io/appcast?prereleases=false'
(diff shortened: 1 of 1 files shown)


core-code:
```
 - appcast at URL 'https://api.gifox.io/appcast?prereleases=false' does not contain the version number: '2.0.0'
```

not sure whats going on here, the message doesn't seem to be true

core-code:
not happy about changing to the feed URL that also contains beta versions, but the real one didn't work inside travis

reitermarkus:
Maybe we should output the appcast contents on CI if it fails so we can actually see what's the problem.

core-code:
i did modify my local copy to do just that. it seems it downloads an empty feed (balid sparkle cml but zero entries). i guess it must be an user-agent check. off the top of my head i am unsure though why it uses a different user agent than curl on the cli.

core-code:
i am sorry but can would it be possible merge this first and investigate how to get a perfect appcast later?

core-code:
thanks, i've opened https://github.com/Homebrew/homebrew-cask/pull/73155 to try to add the real appcast

PR #72491, opened 2019-11-12
Delete impactor.rb

vitorgalvao:
[Suspected of malware](https://github.com/Homebrew/homebrew-cask/issues/56223#issuecomment-551698790).

```
impactor (added 1169 days ago)
30 days: 49 (#1083)
90 days: 184 (#907)
365 days: 734 (#853)
```

PR #73127, opened 2019-11-24
Revert "Delete impactor.rb (#72491)"

TheNoim:
This reverts commit 76e8b4119666a7bd80f30d3cfb790f4b076ed7b1.

Revert the following pull request: #72491 

> This is a false positive. And not just that but whomever made that comment i'll quote in a moment is very misinformed
> 
> > is an old jailbreaking tool to sideload things into your iOS device
> 
> Impactor is only a sideloading tool, not a jailbreaking tool. Also, it's not that old, it came out like 3 or 4 years ago.
> 
> > I don't know if it's even maintained anymore
> 
> All you need to do is some simple research, it's constantly updated, saurik literally tweeted about it a few days ago.
> 
> > as Cydia kinda went defunct
> 
> No it didn't. Cydia still gets updates, although it's not on the hands of saurik anymore.
> 
> > BUT, I would argue that it probably doesn't have any legitimate purposes anymore
> 
> That's a stupid thing to say considering a ton of people use Impactor all the time (well maybe not right now as a server-side change by Apple broke it, but you get the point)

~https://github.com/Homebrew/homebrew-cask/pull/72491#issuecomment-557908163

### DIFF
--- a/Casks/impactor.rb
+++ b/Casks/impactor.rb
@@ -1,0 +1,12 @@
+cask 'impactor' do
+  version '0.9.52'
+  sha256 '9db548074424473c5804d1118d27cd4f052db8b53b3e7c3261c1a903f521cbf1'
+
+  # cache.saurik.com/impactor was verified as official when first introduced to the cask
+  url "https://cache.saurik.com/impactor/mac/Impactor_#{version}.dmg"
+  appcast 'https://cydia.saurik.com/api/appcast/1'
+  name 'Impactor'
+  homepage 'http://www.cydiaimpactor.com/'
+
+  app 'Impactor.app'
+end


claui (commented):
I don’t mind the cask to be re-added.

(I consider it good practice not to override fellow maintainers’ actions so I’d love one of @vitorgalvao and @reitermarkus to chime in before this gets merged.)

vitorgalvao:
I don’t mind re-adding.

Though I do mind the manner in which the situation was brought up on the other issue. If something like that happens again, I’ll question the value this cask brings to the project.

vitorgalvao (commented):
I haven’t checked the `was verified as official when first introduced to the cask` case, but I’m fine with merging if that is correct.

ran-dall:
🙏 Just wanted to take a moment to thank @vitorgalvao, @core-code, and @claui. This may not get said enough, but if no one else, I appreciate you guys and all the hard work that everyone does.

haykam821:
If casks are going to be removed for being jailbreaking tools, then checkra1n, pangu, and evasi0n should be removed too.

vitorgalvao:
> If casks are going to be removed for being jailbreaking tools, then checkra1n, pangu, and evasi0n should be removed too.

You’re commenting this **on the PR where the cask was readded**. Let it go, you’re only making it worse.

haykam821:
@vitorgalvao I’m commenting this because I don’t want this to be standard for Homebrew Cask.

ran-dall:
@haykam821 I don't think HBC has anything against jailbreaking tools. You can read the rest here: https://github.com/Homebrew/homebrew-cask/pull/72491#issuecomment-557981182. But as @vitorgalvao said...
> You’re commenting this on the PR where the cask was readded. Let it go, you’re only making it worse.

PR #73155, opened 2019-11-25
Update gifox

core-code:

### DIFF
--- a/Casks/gifox.rb
+++ b/Casks/gifox.rb
@@ -4,7 +4,7 @@ cask 'gifox' do
 
   # d3si16icyi9iar.cloudfront.net/gifox was verified as official when first introduced to the cask
   url "https://d3si16icyi9iar.cloudfront.net/gifox/#{version.after_comma}.dmg"
-  appcast 'https://d3si16icyi9iar.cloudfront.net/gifox/appcast.xml'
+  appcast 'https://gifox.io/sparkle?prereleases=false'
   name 'gifox'
   homepage 'https://gifox.io/'
 
--- a/Casks/gifox.rb
+++ b/Casks/gifox.rb
@@ -4,7 +4,7 @@ cask 'gifox' do
 
   # d3si16icyi9iar.cloudfront.net/gifox was verified as official when first introduced to the cask
   url "https://d3si16icyi9iar.cloudfront.net/gifox/#{version.after_comma}.dmg"
-  appcast 'https://gifox.io/sparkle?prereleases=false'
+  appcast 'https://www.macupdater.net/cgi-bin/extract_text/download_with_useragent.cgi?url=https://api.gifox.io/appcast?prereleases=false'
   name 'gifox'
   homepage 'https://gifox.io/'
 


core-code:
this is not supposed to be merged ATM. for some reason the travis check gets an empty appcast. https://github.com/Homebrew/brew/pull/6771 is supposed to help in cases like this to show what gets actually downloaded.

i suspect there is some sort of user-agent issue going on but i don't have any evidence yet and i am not sure how we can work around this for appcasts.

core-code:
ok, as expected it is an user-agent issue.

this works
`curl -A "Test" -v "https://api.gifox.io/appcast?prereleases=false"
`
this doesn't (the default homebrew user agent:
`curl -A "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/602.4.8 (KHTML, like Gecko) Version/10.0.3 Safari/602.4.8" -v "https://api.gifox.io/appcast?prereleases=false"
`

i haven't been able to see a clear pattern when it works and when it doesn't.

i don't think we have a facility to use a different user_agent for the appcast check so i am not sure how to fix this.


remark #1: we could use a CGI to fake the user agent like so
https://www.macupdater.net/cgi-bin/extract_text/extract_text_split_easy.cgi?url=https://api.gifox.io/appcast?prereleases=false

remark #2: the default homebrew user agent is still from macOS 10.12.3 which seems quite outdated. should i send a PR to update this or is it on purpose?

vitorgalvao:
> the default homebrew user agent is still from macOS 10.12.3 which seems quite outdated. should i send a PR to update this or is it on purpose?

I would say it was set once and there was never a need to change it, which is why it’s that old. If there’s an advantage to updating it, I don’t see a reason to not do it.

As for this particular cask, it also works if there’s no user agent, and [looking at the relevant code](https://github.com/Homebrew/brew/blob/96e0686d783236ca15c6a718c60202d23d567037/Library/Homebrew/cask/audit.rb#L309) I wonder why we even have a user agent set. Appcasts are not supposed to be checked by browsers, so faking a browser user agent shouldn’t make a difference.

core-code:
> If there’s an advantage to updating it, I don’t see a reason to not do it.

i am not sure if there is any benefit apart from an updated string feeling more like the right string. maybe we should wait until we see the current string actually failing in a case

> Appcasts are not supposed to be checked by browsers, so faking a browser user agent shouldn’t make a difference.

well, in some cases it does make a difference. i know a few apps that require a browser user-agent, though none of them have casks. Transmit and the TechSmith products also require a faked user-agent so that they are similar to what the apps themselves send. so, its probably not possible to find a string that works in every case.

if you want to switch to the default curl string, i can run an analysis on all casks to see whether the retrieved appcast still stays valid.

vitorgalvao:
> if you want to switch to the default curl string, i can run an analysis on all casks to see whether the retrieved appcast still stays valid/identical.

Adding a new option to `appcast` to specify the user-agent on a per-cask basis would be too much, but apart from that I have no strong preference. I’m fine with whatever works best for you, be it removing the user agent, updating it, or changing the appcast to one of yours.

core-code:
i fear that removing the user-agent will help in this case, but make things worse in other cases.

updating the user-agent to a newer string will not actually help in this case.

and i agree that adding an option to specify the user-agent is probably overkill [1] unless we see more of these cases.

so, i'd go with a CGI to modify the user-agent. i see that it mangles the html entities, i'll see if something can be done about that.

@reitermarkus any opinion here?

[1] i wonder why adding a new option to the caskfile format is not a 5 minute job like it ought to be. or perhaps it actually is for someone who is into the ruby and the codebase?

core-code:
i've made a new CGI that does nothing but download a URL with a given user agent

```
#!/usr/bin/python
import cgi
import urllib2

form = cgi.FieldStorage()
url = form.getvalue("url", "")
ua = form.getvalue("user_agent", "Mozilla")
	
headers = { 'User-Agent' : ua }
	
try: 
	req = urllib2.Request(url, None, headers)
	response = urllib2.urlopen(req)
	text = response.read()

except Exception:
	import traceback
	text = 'generic exception: ' + traceback.format_exc()

print "Content-type: text/plain"
print
print text.encode("utf8")
```

vitorgalvao:
> i've made a new CGI

Thank you. Let’s add it to [`developer/cgi`](https://github.com/Homebrew/homebrew-cask/tree/master/developer/cgi). Does it also have server and local versions?

core-code:
yes i have a local version too.
i've uploaded it here, feel free to cherry-pick:

https://github.com/core-code/MiscApps/tree/master/HomebrewCGI

PR #73304, opened 2019-11-27
Update vuescan from 9.7.08 to 9.7.09

core-code:


core-code:
@vitorgalvao i am not sure the "no_check" is still required since we have the full version number in there now?

vitorgalvao:
Let’s try it. Technically every cask with `:no_check # required as upstream package is updated in-place` could have a shasum if its `appcast` is good enough and checked often.

PR #73310, opened 2019-11-27
Update foobar2000

core-code:

### DIFF
--- a/Casks/foobar2000.rb
+++ b/Casks/foobar2000.rb
@@ -2,7 +2,7 @@ cask 'foobar2000' do
   version '2.1.55'
   sha256 'c3841ab4fbdb8ab36c60bde3607387cb74311e70070660401183fee47e0cf6f3'
 
-  url "https://www.foobar2000.org/files/b238d40de415d3afc284eee48712521d/foobar2000-v#{version}.dmg"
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao the cask is broken again. it is definitely download-protection. do you see a way around this? if not, i'll send a PR to delete it.

vitorgalvao:
We could use a [`url do`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#using-a-block-to-defer-code-execution) block, but I question if it’d be worth it. In my experience, if a vendor is adamant about download protection, it’s not worth our time to fight it because they can just change their method. This cask is still young and has few users, so removal wouldn’t be problematic.

If you think it’s worth a try (you might have better numbers on how widespread this software is), we may try, in order (1 and 2 can be switched):

1. Add `url do`.
2. Ask them if they would provide a stable URL.
3. Remove the cask.

core-code:
personally i'd just remove it. its unpopular and i agree that fighting the developer's wishes doesn't make sense in most cases.

maybe previous contributors may want to give the 'url block' method a try or contact the vendor for a stable url. ping @icpz @kiendang @mjsteinbaugh

kiendang:
I'm ok with removing it.

mjsteinbaugh:
I’m OK with removal if this continues to be a problem

icpz:
I'm ok with removal, too.

PR #73391, opened 2019-11-29
Update superduper from 3.3,120 to 3.3.1,121

chohner:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #73412, opened 2019-11-30
Update sha in superduper.rb

suschizu:
sha was wrong


ran-dall:
@suschizu Aren't we already updating `superduper` in https://github.com/Homebrew/homebrew-cask/pull/73391

suschizu:
sure - thanks!

PR #73493, opened 2019-12-01
Update get-iplayer-automator from 1.16.1.b20191008001 to 1.17,20191130002

suschizu:

### DIFF
--- a/Casks/get-iplayer-automator.rb
+++ b/Casks/get-iplayer-automator.rb
@@ -1,10 +1,10 @@
 cask 'get-iplayer-automator' do
-  version '1.17.b20191130002'
+  version '1.17,b20191130002'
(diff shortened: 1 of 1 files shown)


Casks/get-iplayer-automator.rb
```diff
@@ -1,10 +1,10 @@
 cask 'get-iplayer-automator' do
-  version '1.16.1.b20191008001'
-  sha256 '616702b1830f30ff34dea6b0d5adf1fee4f06f676ee0326597f587dec32112bd'
+  version '1.17,20191130002'
```
vitorgalvao: Why not `1.17.b20191130002`, as they have in the releases feed? It would also simplify the interpolation and eliminate the need for `configuration` in `appcast`.

PR #73495, opened 2019-12-02
Add Sorbet 0.4.5085.20191130175112-e9dddf67d

fzakaria:
Added Sorbet https://sorbet.org Cask.
I had attempted to create a homebrew formula https://github.com/Homebrew/homebrew-core/pull/47173
however was unable to get the formula to build on all OSX versions.

Opted instead to create a cask which distributes the prebuilt binaries
by the sorbet team.

```brew cask audit sorbet --download
==> Downloading https://github.com/sorbet/sorbet/releases/download/0.4.5085.20191130175112-e9dddf67d/mac.sorbet
Already downloaded: /Users/faridzakaria/Library/Caches/Homebrew/downloads/59f9f3451a944dafa6613811e8821d99dccc13acdf16696d3e3f0a2f0f8ad0f5--mac.sorbet
==> Verifying SHA-256 checksum for Cask 'sorbet'.
audit for sorbet: passed
```
```
$ brew cask style Casks/sorbet.rb     
```

### DIFF
--- a/Casks/sorbet.rb
+++ b/Casks/sorbet.rb
@@ -1,0 +1,13 @@
+cask 'sorbet' do
+  version '0.4.5085.20191130175112-e9dddf67d'
+  sha256 '26c37639190ba53aa817fb0ac43792a1c8005ac38ef5b1dd911ef5b0472426b7'
+
+  # github.com/sorbet/sorbet was verified as official when first introduced to the cask
+  url "https://github.com/sorbet/sorbet/releases/download/#{version}/mac.sorbet"
+  appcast 'https://github.com/sorbet/sorbet/releases.atom'
+  name 'Sorbet'
+  name 'sorbet'
+  homepage 'https://sorbet.org/'
+
+  binary 'mac.sorbet', target: 'sorbet'
+end


DarkDimius:
Maintainer of upstream sorbet here.
Is it possible to call it `sorbet-static`?

The reason being that this only includes the internal static binary and doesn't include the runtime component nor does it include the `srb` wrapper, thus making it very special and pretty much not covered by docs.

Calling it `sorbet-static` will also match the name that this binary has in rubygems: https://rubygems.org/gems/sorbet-static

vitorgalvao:
@fzakaria Welcome, and thank you for the submission, but this isn’t how we work. Unless there’s an impossibility of adding this to homebrew-core we won’t accept it here, and so far it seems to be possible, even it hasn’t gotten there yet.

@DarkDimius 

> Is it possible to call it `sorbet-static`?

That complicates things, because from what you’re saying it seems like including these instead of compiling from source could lead to users not getting exactly what they want (cue bug reports on our side). Adding the `-static` would be confusing with our naming schemes.

Seems like the best bet for now would be if you both could work on getting the formula (i.e. https://github.com/Homebrew/homebrew-core/pull/47173) working. Having an upstream maintainer looking into it as well is bound to make it easier.

DarkDimius:
> That complicates things...

> Seems like the best bet for now would be if you both could work on getting the formula (i.e. Homebrew/homebrew-core#47173) working.

Note that that pull request has the same issue: it only compiles the static library(and thus excludes `srb`, and thus most documentation examples won't work) and goes against common way to manage versions of dependencies in ruby(via Gemfile).

ooc: what's the motivation to include it in homebrew in the first place compared to using the documented gemfile installation?

fzakaria:
Okay :/ 

<ooc>

@DarkDimius I described the problem statement in https://github.com/sorbet/sorbet/issues/2213

tl;dr; If you are using a non-standard ruby interpreter i.e. JRuby -- you cannot install sorbet-static via gem due to the platform restriction. The platform restriction is also non-traditional since sorbet-static delivers a separate utility vs. a library to be `required` in. 

I was hoping to make `sorbet` (sorbet-static) installable through other mens other than `gem` as I work on integration with our JRuby codebase. The binary is offered on the _release_ page which is great -- I was aiming for a more ergonomic installation.

</ooc>

DarkDimius:
@fzakaria I see, would it be possible for you to join Sorbet slack to discuss options? https://sorbet.org/slack

PR #73527, opened 2019-12-02
Update hedgewars

suschizu:


Casks/hedgewars.rb
```diff
@@ -3,7 +3,7 @@
   sha256 '5a0bdd9bc4cb7beea03b95a2549c1cf994ea8646f6621f5353dd22d79c757404'
 
   url "https://www.hedgewars.org/download/releases/Hedgewars-#{version}.dmg"
-  appcast 'https://www.hedgewars.org/download/appcast.xml'
```
vitorgalvao: The `1.0.0` app still points to this appcast, so there’s a chance it’ll come back. Looks like they might have forgotten to update it.

suschizu: yes true - but I have this mismatch since 2 month in my output so I'm not sure if they will change it … I've opened a topic in the forum - perhaps someone takes care of it

PR #73645, opened 2019-12-04
Update optimage

ericbn:

### DIFF
--- a/Casks/optimage.rb
+++ b/Casks/optimage.rb
@@ -10,4 +10,5 @@ cask 'optimage' do
   auto_updates true
 
   app 'Optimage.app'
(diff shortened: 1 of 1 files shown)


ericbn:
Add the `optimage` cli binary.

Actually, this app provides other binaries too. Not sure if we should link them all:
```
% ls /Applications/Optimage.app/Contents/MacOS/cli
cwebp  ffmpeg  gifsicle  gs  icnsutil  icotool  node  optimage  sips  svgo.sh
```

vitorgalvao:
> Actually, this app provides other binaries too. Not sure if we should link them all

Definitely not. Those other tools are not made by the author of optimage. Several of them have formulae, and `sips` even comes with macOS.

PR #73682, opened 2019-12-04
Rename x-lite to bria-solo and update from 5.8.1_102009 to 6.0.1_102001

suschizu:

### DIFF
--- a/Casks/bria.rb
+++ b/Casks/bria.rb
@@ -1,4 +1,4 @@
-cask 'bria-solo' do
+cask 'bria' do
   version '6.0.1_102001'
   sha256 'd327f8518f8c4aa16e67f3c557fe4985c158d448f889f87f0bf1858a6ae95ea0'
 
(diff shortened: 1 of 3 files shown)


Casks/bria.rb
```diff
@@ -0,0 +1,11 @@
+cask 'bria' do
```
vitorgalvao: Renamed to conform to the token reference (we could always add the suffix to the other apps, if they’re ever submitted) and because I poked around on the website and it seems like the other versions are different plans but use the same app.

vitorgalvao (approved):

Casks/bria.rb
suschizu: hey @vitorgalvao - sorry I forgot to commit the comment yesterday :(

x-lite is now bria solo free according to their description (https://www.counterpath.com/x-lite-download/)- bria is now grouped into solo-team-enterprise - just the solo has a free optional version

PR #73749, opened 2019-12-05
Update neteasemusic.rb from 2.3.0_814,b2a5:20191015153044 to ?? (2.3.1_824,605533604/4f60:201911523186)

suschizu:
hey @vitorgalvao - could you please have a look on this?
The problem is that the feed download has now 4 changing parts instead of 3
this is the new download:
http://d1.music.126.net/dmusic/obj/w5zCg8OCw6fCn2vDicOl/605533604/4f60/201911523186/NeteaseMusic_2.3.1_824_web.dmg
(from https://music.163.com/api/mac/appcast.xml)
thanks

### DIFF
--- a/Casks/neteasemusic.rb
+++ b/Casks/neteasemusic.rb
@@ -1,9 +1,9 @@
 cask 'neteasemusic' do
-  version '2.3.1_824,605533604/4f60:201911523186'
+  version '2.3.1_824,605533604.4f60.201911523186'
   sha256 '4860316e46904594ed2ccb96fc451bb4d49ee7de01be7c64069980d50edac989'
 
   # d1.music.126.net was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


vitorgalvao:
`major`, `minor`, and `patch` are not strictly true in this case, but I’m guessing only the `before_comma` matters to the database as the real version, so this solution makes it easy to maintain without being disruptive.

vitorgalvao:
The solution didn’t work because of the `f` in `minor`, but the restriction that the sections of `version` must be strictly numeric has given problems more than once and I don’t recall it ever having been beneficial, so I’m submitting [a PR to change it](https://github.com/Homebrew/brew/pull/6820).

Changed this to a mostly hardcoded version so we can merge, but once it’s in I’ll submit a new PR to change it to the previous solution, which will have to wait for the PR to the core to be merged and a new Homebrew version to be released.

suschizu:
thanks for the explanation

PR #73836, opened 2019-12-07
Update icq from 3.0.24385 to 3.0.24905

suschizu:

### DIFF
--- a/Casks/icq.rb
+++ b/Casks/icq.rb
@@ -1,5 +1,5 @@
 cask 'icq' do
-  version '3.0.24385'
+  version '3.0.24905'
   sha256 '0cd6532b87c38b082711891f8f1908f51c3031d6606ab9bc59c3f08e43ab9254'
 
   # mra.mail.ru/icq_mac3_update was verified as official when first introduced to the cask


core-code:
@sushizu: i can not reproduce this at all. the file downloaded is version 3.0.24385. where does the number 3.0.24905 come from? you've just changed the version number but not the SHA, so things couldn't have been correct anyway (unless they were incorrect before). as i see it, the feed is correct and this PR should be reverted.

vitorgalvao:
> the file downloaded is version 3.0.24385.

True for the download from the appcast.

> where does the number 3.0.24905 come from?

That’s the version you get if you download from the website.

> you've just changed the version number but not the SHA, so things couldn't have been correct anyway (unless they were incorrect before)

Since I verified the version as being correct from the website before merging, I assumed the cask was indeed incorrect before (with unversioned URLs that’s easier to happen). I’ve rechecked, and the URL itself has changed (new one is `https://hb.bizmrg.com/icq-www/mac/x64/icq.dmg`). I wasn’t able to find a working appcast yet.

https://github.com/Homebrew/homebrew-cask/pull/73912

core-code:
> Since I verified the version as being correct from the website before merging, I assumed the cask was indeed incorrect before (with unversioned URLs that’s easier to happen).
 
ah that explains things

> I’ve rechecked, and the URL itself has changed (new one is `https://hb.bizmrg.com/icq-www/mac/x64/icq.dmg`). I wasn’t able to find a working appcast yet.
> 
> #73912

ah ok thanks there. i'll look into finding a new appcast.

PR #73867, opened 2019-12-08
Fix Native Access Cask

danielbayley:

- [x]  The commit message includes the cask’s name and version.

### DIFF
--- a/Casks/native-access.rb
+++ b/Casks/native-access.rb
@@ -1,10 +1,40 @@
 cask 'native-access' do
-  version :latest
-  sha256 :no_check
+  version '1.12.1.129'
+  sha256 '21d7f6cde1a173c12392e323bf8d23e51460f389e38552c367bb8ab41d61f20e'
 
-  url 'https://www.native-instruments.com/fileadmin/downloads/Native_Access_Installer.dmg'
+  url 'https://native-instruments.com/fileadmin/downloads/Native_Access_Installer.dmg'
+  appcast "https://native-instruments.com/forum/threads/official-update-status-native-access-current-version-#{version.major_minor_patch.dots_to_hyphens}.313999",
+          configuration: version.major_minor_patch
   name 'Native Access'
-  homepage 'https://www.native-instruments.com/en/specials/native-access/'
+  homepage 'https://native-instruments.com/specials/native-access'
+
+  depends_on macos: '>= :sierra'
 
   app 'Native Access.app'
+
+  uninstall quit:      'com.native-instruments.Native Access',
+            launchctl: 'com.native-instruments.NativeAccess.Helper2',
+            delete:    [
+                         '/Library/PrivilegedHelperTools/com.native-instruments.NativeAccess.Helper2',
+                         '/Library/Application Support/Native Instruments/Helper/NIHelperAuth',
+                       ],
+            rmdir:     [
+                         '/Library/Application Support/Native Instruments/*',
+                         '/Library/Application Support/Native Instruments',
+                       ]
+
+  zap trash: [
+               '/Library/Application Support/Native Instruments/Service Center/NativeAccess.xml',
+               '~/Library/Application Support/Native Instruments/Native Access*',
+               '~/Library/Application Support/Native Instruments/Service Center',
+               '~/Library/Application Support/CrashReporter/Native Access_*.plist',
+               '~/Library/Caches/Native Instruments/Native Access*',
+               '~/Library/Preferences/com.native-instruments.Native Access.plist',
+               '~/Library/Preferences/com.native-instruments.Shared.plist',
+             ],
+      rmdir: [
+               '/Library/Application Support/Native Instruments/*',
+               '/Library/Application Support/Native Instruments',
+               '~/Library/Application Support/Native Instruments',
+             ]
 end


Casks/native-access.rb
```diff
@@ -1,10 +1,36 @@
 cask 'native-access' do
-  version :latest
-  sha256 :no_check
+  version '1.12.1.129'
+  sha256 '21d7f6cde1a173c12392e323bf8d23e51460f389e38552c367bb8ab41d61f20e'
 
-  url 'https://www.native-instruments.com/fileadmin/downloads/Native_Access_Installer.dmg'
+  url 'https://native-instruments.com/fileadmin/downloads/Native_Access_Installer.dmg'
+  appcast "https://native-instruments.com/forum/threads/official-update-status-native-access-current-version-#{version.major_minor_patch.dots_to_hyphens}.313999",
```
vitorgalvao: Doesn’t the `.313999` change?

Casks/native-access.rb
```diff
@@ -1,10 +1,36 @@
...
 
   app 'Native Access.app'
+
+  uninstall quit:      'com.native-instruments.Native Access',
+            launchctl: 'com.native-instruments.NativeAccess.Helper2',
+            delete:    [
+                         '/Library/PrivilegedHelperTools/com.native-instruments.NativeAccess.Helper2',
+                         '/Library/Application Support/Native Instruments/Helper/NIHelperAuth',
+                       ],
+            rmdir:     '/Library/Application Support{/Native Instruments{/*,},}'
```
vitorgalvao: These are getting confusing. It’s one thing to do them in huge casks, but on smaller casks this will only confuse other users. Please don’t use anything more complicated than `*`.

Casks/native-access.rb
```diff
@@ -1,10 +1,36 @@
...
+               '/Library/Application Support/Native Instruments/Service Center/NativeAccess.xml',
+               '~/Library/Application Support/Native Instruments/Native Access*',
+               '~/Library/Application Support/Native Instruments/Service Center',
+               '~/Library/Application Support/CrashReporter/Native Access_*.plist',
+               '~/Library/Caches/Native Instruments/Native Access*',
+               '~/Library/Preferences/com.native-instruments.Native Access.plist',
+               '~/Library/Preferences/com.native-instruments.Shared.plist',
+             ],
+      rmdir: [
+               '/Library/Application Support{/Native Instruments{/*,},}',
```
vitorgalvao: Same as the comment above

vitorgalvao (changes requested):

vitorgalvao:
@danielbayley Pinging to know if you’re still interested in doing this (with the changes).

danielbayley:
> @danielbayley Pinging to know if you’re still interested in doing this (with the changes).

@vitorgalvao Yeah sure, I'll get around to it soon…

Casks/native-access.rb
danielbayley: Actually no, I don't think so, as it's a pinned forum post with official release notes.

Casks/native-access.rb
danielbayley: @vitorgalvao I agree. But this is exactly why I raised issue #67806, which you closed… it's a recurring problem I come across all the time when adding `rmdir:`. This kind of brace expansion was the only solution suggested [here](https://github.com/Homebrew/homebrew-cask/issues/67806#issuecomment-522380747) by @reitermarkus.

As I stated [here](https://github.com/Homebrew/homebrew-cask/issues/67806#issuecomment-546115001):
> I still think `rmdir:` should work recursively on empty folders though… save using brace expansion.

But the point was never acknowledged…

vitorgalvao: > But the point was never acknowledged…

To clarify, I agree with having `rmdir:` be recursive. But I don’t see anyone with enough interest to *implement* the feature, hence the closing.

vitorgalvao: > But the point was never acknowledged…

To clarify, I agree with having rmdir: be recursive. But I don’t see anyone with enough interest to implement the feature, hence the closing.

vitorgalvao:
Thank you.

ErikMinekus:
1. This cask now fails to upgrade with the message:
```
Error: Directory not empty @ dir_s_rmdir - /Applications/Native Access.app
```

2. Native Access auto-updates, so I don't understand why the cask now has a `version`, `sha256` and `appcast`. It should just have the `auto_updates` stanza.

3.
```
rmdir:     [
                         '/Library/Application Support/Native Instruments/*',
                         '/Library/Application Support/Native Instruments',
                       ]
```
```
      rmdir: [
               '/Library/Application Support/Native Instruments/*',
               '/Library/Application Support/Native Instruments',
               '~/Library/Application Support/Native Instruments',
```
Other Native Instruments apps are installed in this directory. They will now break when you uninstall Native Access.

4.
```
  zap trash: [
               '/Library/Application Support/Native Instruments/Service Center/NativeAccess.xml',
...
               '~/Library/Application Support/Native Instruments/Service Center',
```
Service Center is a different app, it shouldn't be removed.

danielbayley:
> Native Access auto-updates, so I don't understand why the cask now has a version, sha256 and appcast. It should just have the auto_updates stanza.

I did forget to add `auto_updates true`… but other casks have all 4.

> Other Native Instruments apps are installed in this directory. They will now break when you uninstall Native Access.

Surely not, because `rmdir:` only removes empty folders… So if other NI apps remain installed and have data stored here, nothing would happen… or am I missing something @ErikMinekus?

> Service Center is a different app, it shouldn't be removed.

Is it not the precursor to Native Access? I could be wrong… but I seem to remember Service Centre replacing itself with Native Access (through it's own auto update)… apparently leaving behind some files. Also `NativeAccess.xml` clearly is related…

vitorgalvao:
> Native Access auto-updates, so I don't understand why the cask now has a `version`, `sha256` and `appcast`. It should just have the `auto_updates` stanza.

Those are not interdependent. Having `auto_updates` does not preclude the other three, and is the least important of the four.

ErikMinekus:
> Surely not, because `rmdir:` only removes empty folders… So if other NI apps remain installed and have data stored here, nothing would happen… or am I missing something @ErikMinekus?

I still wouldn't want a Cask to touch folders for other apps, even if they're empty. And what if the implementation for `rmdir` changes in the future and it does remove folders with files?

> Is it not the precursor to Native Access? I could be wrong… but I seem to remember Service Centre replacing itself with Native Access (through it's own auto update)… apparently leaving behind some files. Also `NativeAccess.xml` clearly is related…

Service Center has indeed been replaced by Native Access, but I still have both installed. I can still launch Service Center (on macOS Mojave at least) and it doesn't replace itself. And again, I don't think Casks should touch files and folders from other apps.

vitorgalvao:
> I still wouldn't want a Cask to touch folders for other apps

In a `zap`, that’s fair game and [covered by the documentation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/zap.md). `uninstall` is more conservative. There’s a reason we have both.

> And what if the implementation for `rmdir` changes in the future and it does remove folders with files?

Then it would no longer be `rmdir`. The chance of that happening is essentially nil. We already have `delete`; `rmdir` only exists because it only removes empty directories.

PR #73944, opened 2019-12-10
Add HX Edit v2.82

kevinhwang:

### DIFF
--- a/Casks/hx-edit.rb
+++ b/Casks/hx-edit.rb
@@ -1,0 +1,16 @@
+cask 'hx-edit' do
+  version '2.82'
+  sha256 'e8ef13d04bffcdb9a367fa24781d21ba81546c5528149277b9e552118bfdfff6'
+
+  # l6c-acdn2.line6.net was verified as official when first introduced to the cask
+  url "https://l6c-acdn2.line6.net/media/software/9314/HXEdit#{version}.dmg"
+  name 'HX Edit'
+  homepage 'https://line6.com/software'
+
+  pkg 'HX Edit.pkg'
+
+  uninstall pkgutil: [
+                       'com.line6.application.hxedit.package',
+                       'com.line6.application.line6updater.package',
+                     ]
+end


vitorgalvao:
Thank you, but this looks to be a [walled build](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) so [we will not accept it](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#unofficial-vendorless-and-walled-builds).

kevinhwang:
> Thank you, but this looks to be a [walled build](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) so [we will not accept it](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#unofficial-vendorless-and-walled-builds).

@vitorgalvao, Why is this a walled build? `l6c-acdn2.line6.net` looks like a reasonable CDN for https://line6.com. Just because the download host is different in the TLD (`.net` vs `.com`) shouldn't automatically disqualify a cask—lots of sites host their download on separate hosts dedicated to content.

vitorgalvao:
> `l6c-acdn2.line6.net` looks like a reasonable CDN for https://line6.com.

“Looks like” is not good enough. This policy is to protect from malicious actors, and malicious actors would also use something that looks like it’s legitimate. We need verifiability.

> Just because the download host is different in the TLD (.net vs .com) shouldn't automatically disqualify a cask

It doesn’t. It only does so if the authors prevent us from reaching the download link through a login wall or similar.

> lots of sites host their download on separate hosts dedicated to content.

They do. And most of them don’t prevent us from verifying the URL is legitimate. The ones that do, we don’t allow.

> I'd ask you please reconsider.

No. This policy has been in place for a long time and has served us well. There’s nothing about this cask that would warrant an exception. If you need it, I recommend you host your own [tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

kevinhwang:
> [...] We need verifiability.

Okay, it sounds like you're saying that if we could verify that host as a legitimate CDN serving content on behalf of `line6.com`, this cask would pass that rule. That sounds like fair criteria.

If you go to https://line6.com and inspect the network requests being made, you'll see stylesheets, JavaScript, and other assets are being served from https://l6c-acdn2.line6.net. It's actually there in the HTML source. Is that sufficient verification?

vitorgalvao:
> Is that sufficient verification?

No. Verification needs to be simple, in a way that a regular user can do it with little effort, and directly connected to the download URL.

This cask doesn’t even seem to be submitted to the correct repo, as it appears to be a case for [homebrew/cask-drivers](https://github.com/Homebrew/homebrew-cask-drivers). I don’t expect this to be a popular cask.

kevinhwang:
> No. Verification needs to be simple, in a way that a regular user can do it with little effort, and directly connected to the download URL.

Very well. I'll circle back if they ever make the direct download link obviously visible on their site.

> This cask doesn’t even seem to be submitted to the correct repo, as it appears to be a case for [homebrew/cask-drivers](https://github.com/Homebrew/homebrew-cask-drivers).

It's not a driver. It's an app to edit patches (presets / snapshots) on Line 6's line of multi-effects units. It's an editor for electric guitar pedals.

> I don’t expect this to be a popular cask.

I don't know enough to dispute that with any confidence, but I'm pretty sure the number of musicians, and in particular, guitarists who use multi-effects pedals like Line 6s' is larger than that of some of the  really random stuff that's hosted by Homebrew Cask.

vitorgalvao:
> It's not a driver. (…) It's an editor for electric guitar pedals.

Which makes it useless without extra hardware, which makes it a case for `homebrew/cask-drivers`. That’s covered in the links I sent you previously.

> but I'm pretty sure the number of musicians, and in particular, guitarists who use multi-effects pedals like Line 6s' is larger than that of some of the really random stuff that's hosted by Homebrew Cask.

That is only relevant to us if said musicians also use Homebrew Cask, which is where my doubt lies.

PR #73974, opened 2019-12-10
Update visual-paradigm

core-code:


core-code:
i think using the target of the redirect for the appcast is a bad idea because by the time there is a new release the file is already gone.

core-code:
i am not sure why this doesn't pick up the redirect in travis correctly. as far as i can remember the CURL call for the appcast should follow the redirect

core-code:
we could just use something like this
`https://www.macupdater.net/cgi-bin/check_urls/check_url_redirect.cgi?url=https://www.visual-paradigm.com/downloads/vp/checksum.html`

PR #74032, opened 2019-12-11
Update bettertouchtool from 3.209 to 3.213

ran-dall:


wagnerand:
Could we consider not updating BTT before the author of the application has made it available through their built-in updater? There is usually a reason some builds are not shipped as an update to users. Thanks!

PR #74177, opened 2019-12-13
Fix ghidra binary path for ghidra-9.1_PUBLIC_20191023

ixs:
The assumed path was based on the version, not taking into account
the difference between download URL and actual path included in the
release package.

The beta package does it right, backport that change to the stable
ghidra package.

An example error report can be seen at
https://discourse.brew.sh/t/brew-cask-install-ghidra-fails/6444


vitorgalvao:
Thank you, and welcome.

PR #74238, opened 2019-12-14
Update sherlock.rb from 1.7.3 to 1.7.4 + update appcast and url 

suschizu:
I'm not sure about this update because the website download still contains v 1.7.3 - but the downloaded version contains this appcast as SUFeedURL


vitorgalvao:
Presumably, 1.7.4 will be released soon on the website. Let’s add the appcast but keep the cask at 1.7.3 for now. If by the time they release the new version, the appcast is ahead again, let’s switch the cask to the URL and version they use on the appcast.

PR #74463, opened 2019-12-19
Update joplin from 1.0.175 to 1.0.176

ran-dall:


DanielNetoP:
I think that this update shouldn't have been made, the version 1.0.176 still is a "Pre-release".
I believe that this should've been updated on releases only and not pre-releases.

ran-dall:
@DanielNetoP This passed the ML's update scheme for some reason however it can be reverted.

DanielNetoP:
No worries, It was just that I follow this app, and didn't know about an update, and then I checked and it was indeed a pre release version.
It's already in the correct version, thanks for the new PR ;)

Issue #74479, opened 2019-12-19
Failure while executing; `/usr/bin/sudo -E -- /usr/sbin/dseditgroup -o delete access_bpf` exited with 64. Here's the output: Group not found


dione2017:
brew cask reinstall --force  wireshark
==> Downloading https://1.na.dl.wireshark.org/osx/all-versions/Wireshark%203.2.0%20Intel%2064.dmg
Already downloaded: /Users/lijialin/Library/Caches/Homebrew/downloads/d0e402a9e8202174b53fa3f39b8d56637021df493cea55974a1cef025290e011--Wireshark 3.2.0 Intel 64.dmg
==> Verifying SHA-256 checksum for Cask 'wireshark'.
==> Uninstalling Cask wireshark
Group not found.
Error: Failure while executing; `/usr/bin/sudo -E -- /usr/sbin/dseditgroup -o delete access_bpf` exited with 64. Here's the output:
Group not found.

Follow the instructions here:
  https://github.com/Homebrew/homebrew-cask#reporting-bugs
/usr/local/Homebrew/Library/Homebrew/system_command.rb:176:in `assert_success!'
/usr/local/Homebrew/Library/Homebrew/system_command.rb:53:in `run!'
/usr/local/Homebrew/Library/Homebrew/system_command.rb:29:in `run'
/usr/local/Homebrew/Library/Homebrew/system_command.rb:33:in `run!'
/usr/local/Homebrew/Library/Homebrew/cask/dsl/base.rb:16:in `system_command'
/usr/local/Caskroom/wireshark/.metadata/3.0.7/20191218035135.374/Casks/wireshark.rb:19:in `block (2 levels) in load'
/usr/local/Homebrew/Library/Homebrew/cask/artifact/abstract_flight_block.rb:41:in `instance_eval'
/usr/local/Homebrew/Library/Homebrew/cask/artifact/abstract_flight_block.rb:41:in `abstract_phase'
/usr/local/Homebrew/Library/Homebrew/cask/artifact/abstract_flight_block.rb:28:in `uninstall_phase'
/usr/local/Homebrew/Library/Homebrew/cask/installer.rb:454:in `block in uninstall_artifacts'
/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/set.rb:777:in `each'
/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/set.rb:777:in `each'
[...]

PR #74502, opened 2019-12-19
wireshark.rb: fix permissions of /etc/paths.d/Wireshark

vitorgalvao:
As reported in https://github.com/Homebrew/homebrew-cask/issues/74479.

### DIFF
--- a/Casks/wireshark.rb
+++ b/Casks/wireshark.rb
@@ -15,6 +15,7 @@ cask 'wireshark' do
   pkg 'Add Wireshark to the system path.pkg'
 
   uninstall_preflight do
+    set_permissions '/etc/manpaths.d/Wireshark', '0644'
     set_permissions '/etc/paths.d/Wireshark', '0644'
     set_ownership '/Library/Application Support/Wireshark'
     system_command '/usr/sbin/dseditgroup', args: ['-o', 'delete', 'access_bpf'], sudo: true
--- a/Casks/wireshark.rb
+++ b/Casks/wireshark.rb
@@ -15,6 +15,7 @@ cask 'wireshark' do
   pkg 'Add Wireshark to the system path.pkg'
 
   uninstall_preflight do
+    set_permissions '/etc/paths.d/Wireshark' '0644'
     set_ownership '/Library/Application Support/Wireshark'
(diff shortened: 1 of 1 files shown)


jabenninghoff:
Should the `set_permssions` statement be part of the `postflight` stanza? I think that will fix the problem reported by @Tenzer and myself in #74479. I wasn't able to replicate the original poster's issue.

Or... as this seems to be a bug in the actual original Wireshark distribution - the `Add Wireshark to the system path.pkg` doesn't set the correct permissions (I confirmed this with testing) - is this something that should be fixed upstream?

vitorgalvao:
> is this something that should be fixed upstream?

Yes. Would you mind informing them?

jabenninghoff:
I don't mind - I'll file a bug report with Wireshark.

jabenninghoff:
To be clear - I don't think this proposed PR will actually fix issue #74479. I wasn't able to replicate that issue.

vitorgalvao:
[There have been more reports](https://github.com/Homebrew/homebrew-cask/issues/74548). This needs to be reported upstream.

Stellarspace:
I have the same problem.

vitorgalvao:
> I have the same problem.

Have you reported the issue upstream, [as mentioned](https://github.com/Homebrew/homebrew-cask/pull/74502#issuecomment-568228072)?

Stellarspace:
> > I have the same problem.
> 
> Have you reported the issue upstream, [as mentioned](https://github.com/Homebrew/homebrew-cask/pull/74502#issuecomment-568228072)?

No, I don't know what that means, but I resolved it by typing the following:
`sudo chmod 644 /etc/manpaths.d/Wireshark /etc/paths.d/Wireshark`

marques-work:
@jabenninghoff @vitorgalvao 

> Should the `set_permssions` statement be part of the `postflight` stanza?

It ought to be both `postflight` and `uninstall_preflight` to fix both cases until this is resolved upstream:

1. After installing Wireshark, the user sees a permission error from `path_helper` for every shell login session (evaluating `/etc/paths.d`
2. Fixing Wireshark uninstall.

vitorgalvao:
[Fixed upstream](https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=16284).

PR #74631, opened 2019-12-21
Correction of the version notation in command-tab-plus.rb from 359,359… to 1.109,359…

suschizu:
there was a mistake in the notation of the version number - should be: short version string,bundle version:downloadtoken


reitermarkus:
Thanks!

PR #74797, opened 2019-12-24
Add depends on catalina to hiddenbar.rb

suschizu:
min version is 10.15


hellokangning:
Per the [Requirements](https://github.com/dwarvesf/hidden#requirements), the min vesion might be 10.13?

suschizu:
hey @hellokangning - thanks and sorry for this! 
you are right - the 10.15 was for an unofficial version

PR #74880, opened 2019-12-27
Update plover 3.1.1

mtklr:

new homepage


vitorgalvao:
Thank you, and welcome.

mtklr:
thanks - curious why the build seemed to fail, i checked it on my end and it was ok, if i should have done anything else please let me know.
-matt

vitorgalvao:
> curious why the build seemed to fail

Not your fault. You’ll notice the cask has an `appcast`; the build checks that page for the number in `version` and fails if it isn’t there, which is a safeguard against updating to non-stable versions. In this case, the feed has so many pre-releases that the latest stable is no longer present, making it fail. It’s a false negative, which is why I merged anyway.

PR #74923, opened 2019-12-30
Delete sampler (added to homebrew-core instead)

sqshq:
Hello. I'd like to delete `sampler`, since it was added to Homebrew-core recently:
https://github.com/Homebrew/homebrew-core/pull/48262

The possibility of switching from cask to core was discussed half a year ago, when I originally added it to cask: https://github.com/Homebrew/homebrew-cask/pull/67090

Thank you.


vitorgalvao:
Thank you for following-up.

PR #74933, opened 2019-12-30
Update gnucash from 3.7,1 to 3.8b,1

bd12:

### DIFF
--- a/Casks/gnucash.rb
+++ b/Casks/gnucash.rb
@@ -1,9 +1,9 @@
 cask 'gnucash' do
-  version '3.7,1'
-  sha256 'b687fb0a17b15e735a29be83354409fb8059aebf0bd97079b74eac3dfc4cbeac'
+  version '3.8b,1'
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao the 'delete_suffix' method used here is only available in ruby 2.5+, and therefore only available on 'Catalina'. any idea why 'brew cask fetch gnucash' still works on older macOS versions? does homebrew use its own custom ruby internally?

vitorgalvao:
> does homebrew use its own custom ruby internally?

[Yes](https://github.com/Homebrew/homebrew-portable-ruby).

core-code:
thanks for the explanation. i see that 'chomp' should have the effect as 'delete_suffix' with more backwards compatibility (which is of course not really interesting if you have your own ruby version). however, since 'chomp' is already used in 5 casks while 'delete_suffix' is only used in this one, would a PR changing it to 'chomp' be accepted?

vitorgalvao:
> however, since 'chomp' is already used in 5 casks while 'delete_suffix' is only used in this one, would a PR changing it to 'chomp' be accepted?

Yes.

PR #75003, opened 2020-01-02
Update gnucash.rb

core-code:
in reference to https://github.com/Homebrew/homebrew-cask/pull/74933#issuecomment-570103100


suschizu:
wouldn't this be an easier option - ok the "b" is a problem but it would be deleted with the next update:
```
version '3.8-2'

url "https://github.com/Gnucash/gnucash/releases/download/#{version.before_colon}b/Gnucash-Intel-#{version}.dmg"
```

PR #75004, opened 2020-01-02
Revert bettertouchtool from 3.218 (alpha) to 3.214 (stable)

core-code:

### DIFF
--- a/Casks/bettertouchtool.rb
+++ b/Casks/bettertouchtool.rb
@@ -6,8 +6,8 @@ cask 'bettertouchtool' do
     # bettertouchtool.net/releases was verified as official when first introduced to the cask
     url "https://bettertouchtool.net/releases/btt#{version}_final_10_9.zip"
   else
-    version '3.218'
-    sha256 'cac25f3755e48bccdcbbe5568e5db3b2c1ace6beae9a9a897f69f27022e35df1'
+    version '3.214'
+    sha256 'c87ebe47d8783087e1aed26d152b9c6f130bfc69aeb7ad5db7f7f5a45cbe2526'
 
     # bettertouchtool.net/releases was verified as official when first introduced to the cask
     url "https://bettertouchtool.net/releases/btt#{version}.zip"


core-code:
i am unhappy with the state of this cask, as we've been carrying alpha versions of this for a while.

the update-check is at version 3.209:
`http://updates.bettertouchtool.net/highsierra_new.xml?bttversion=1000`

the official download from the website is at version 3.214:
`https://bettertouchtool.net/releases/BetterTouchTool.zip`

any newer version than 3.214 is therefore not an official release. in fact the 3.218 version we currently have is only offered for download if you do "check for alpha versions" in the app:
`https://updates.bettertouchtool.net/highsierraalpha_new.xml?bttversion=1000`


the problem is of course having this as an appcast:
`https://bettertouchtool.net/releases/`
this contains EVERY release, even versions that are still 'alpha'.

i only see two workable solutions:
• revert the appcast to the stable sparkle feed and stick to versions from the sparkle feed even if they may lag several weeks behind the website as they often do. this will allow auto-merging of any updates. or
• revert the appcast to the stable sparkle feed and manually check new updates that are released from the website. this necessitates manual checks whether submitted versions are *really* available from the official download link

i see no way to make an appcast that tracks only versions from the homepage - there is no version number anywhere. (their release-notes are also notoriously outdated for many many weeks).

core-code:
there would probably be a way to observe changes (but not version numbers) from the website.

```
curl -vvvv https://bettertouchtool.net/releases/BetterTouchTool.zip
[…]
< last-modified: Fri, 13 Dec 2019 22:08:34 GMT
< etag: "14e17a7-5999d1a562cac"
```

core-code:
@suschizu @ran-dall pinging persons of interest ;-)

suschizu:
ok - I can check the download if there is an update listed in the output - no problem
@ran-dall - could you put this cask perhaps on your "skipped" list so that I have a reminder when I miss it please 😬

ran-dall:
I’ll add it to skipped. 👍

vitorgalvao:
> revert the appcast to the stable sparkle feed and manually check new updates that are released from the website. this necessitates manual checks whether submitted versions are _really_ available from the official download link

By reverting the appcast, that’s what’s likely to happen (rather than it lagging for a few weeks) because BTT is a popular app and users will continue to submit alpha cask updates.

Only solution I see for this—though not very practical—would be to periodically download the last version on the website (they have an unversioned link) and check the bundle version. If anyone wants to do so, I can whip up the script to do it.

vitorgalvao:
Case in point, less than a day and we already got another PR: https://github.com/Homebrew/homebrew-cask/pull/75045.

core-code:
if we are interested in a technical solution, i can put a script on our server that periodically downloads the website version of BetterTouchTool and saves its version number into a TXT file. we could use this as an appcast for this cask. does that sound like an acceptable solution?

vitorgalvao:
```bash
#!/bin/bash

readonly tmp_dir="$(mktemp -d)"
readonly btt_url='https://bettertouchtool.net/releases/BetterTouchTool.zip'
readonly btt_plist="${tmp_dir}/BetterTouchTool.app/Contents/Info.plist"

curl "${btt_url}" --output "${tmp_dir}/btt.zip"
unzip "${tmp_dir}/btt.zip" -d "${tmp_dir}"

if [[ ! -f "${btt_plist}" ]]; then
  echo 'Could not find BetterTouchTool. Maybe the download failed or the zip structure has changed.'
  exit 1
fi

readonly btt_version="$(defaults read "${btt_plist}" 'CFBundleShortVersionString')"
```

Then it’s only a matter of redirecting `"${btt_version}"` to a file, *and/or*, while we’re at it, invoke `cask-repair`.

vitorgalvao:
Edited the previous script. Untested, but it has quite a bit more error detection.

Issue #66596, opened 2019-07-22
Wrong version being installed from personal branch .rb


jajajaime:
#### General troubleshooting steps

- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [x] I have retried my command with `--force`.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

#### Description of issue

Using the `brew cask reinstall` command to install a recently upgraded cask from my personal branch produces the installation of 2 versions behind instead.
In this case going from 0.4.9 -> 0.4.10, but installs 0.4.8 instead.
The personal branch .rb used is:
```
cask 'ripcord' do
  version '0.4.10'
  sha256 '4dc3c01227f8e5fbf2e71a1b7f8da7273b9870e432e35856217f839eac2d0a3e'

  url "https://cancel.fm/dl/Ripcord_Mac_#{version}.zip"
  appcast 'https://cancel.fm/ripcord/updates/v1'
  name 'Ripcord'
  homepage 'https://cancel.fm/ripcord/'

  depends_on macos: '>= :sierra'

  app 'Ripcord.app'

  zap trash: [
               '~/Library/Application Support/Ripcord',
               '~/Library/Preferences/com.cancelfm.Ripcord.plist',
               '~/Library/Saved Application State/com.cancelfm.Ripcord.savedState',
             ]
end
```

#### Command that failed

```
brew cask reinstall https://raw.githubusercontent.com/jajajaime/homebrew-cask/cask_repair_update-ripcord/Casks/ripcord.rb
```

#### Output of command with `--force --verbose --debug`

```
==> Downloading https://raw.githubusercontent.com/jajajaime/homebrew-cask/cask_repair_update-ripcord/Casks/ripcord.rb.
[...]

PR #75036, opened 2020-01-03
Update macdown from 0.7.2 to 0.7.3

ShivaHuang:

### DIFF
--- a/Casks/macdown.rb
+++ b/Casks/macdown.rb
@@ -1,6 +1,6 @@
 cask 'macdown' do
-  version '0.7.2'
-  sha256 '271f11eb64c19fccee2615e092067cdecc29adf0c2ed0703dae9acda8fa0a672'
+  version '0.7.3'
+  sha256 '57a0568354ee7af694e825f19f9f03b9844c6301c70ea04fb324b6efdf7fdcd0'
 
   # github.com/MacDownApp/macdown was verified as official when first introduced to the cask
   url "https://github.com/MacDownApp/macdown/releases/download/v#{version}/MacDown.app.zip"


vitorgalvao:
Can you install this version locally?

ShivaHuang:
No, unfortunately.
I run
```
brew cask reinstall https://raw.githubusercontent.com/ShivaHuang/homebrew-cask/cask_repair_update-macdown/Casks/macdown.rb
```
and got
```
==> Downloading https://raw.githubusercontent.com/ShivaHuang/homebrew-cask/cask_repair_update-macdown/Casks/macdown.rb.
==> Downloading https://github.com/MacDownApp/macdown/releases/download/v0.7.2/MacDown.app.zip
```
The downloaded file's version doesn't match the one in spec, thus checksum failed.

Any idea what may cause this?

ShivaHuang:
I think the situation is the same as https://github.com/Homebrew/homebrew-cask/issues/66596. But currently there are no findings.

vitorgalvao:
That’s irrelevant. The issue here tripping CI is a permission denied problem.

varenc:
The issue with this is the `/MacDown.app/./Contents/_CodeSignature` file.  It has `000` permissions.  (not readable/writeable by anyone).  I believe those permissions are specified in the zip file and `unzip` which homebrew uses seems to preserve those permissions.  

When unzipped with macOS's built-in "Archive Utility" that file gets `.rwxr-xr-x` permissions instead. It probably overrides the permissions included in the zip in cases like this?

Also, the `_CodeSignature` file is empty so probably isn't doing anything at all.  There was no file at that path in the previous release. Though obscure, I think this is a problem with the MacDown release and something that they should fix. Unless homebrew wants to start fixing permissions in cases like this.

vitorgalvao:
> I think this is a problem with the MacDown release and something that they should fix.

It is. Someone should contact them.

> Unless homebrew wants to start fixing permissions in cases like this.

We prefer not to, [but it is doable](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/flight.md#flight-mini-dsl).

varenc:
Interesting! `set_permissions` seems like exactly what we'd need.

If I made a PR for this cask utilizing that in the preflight to fix this issue, would it be accepted? I opened an issue with MacDown but haven't waited long enough to get a response yet.

(In my earlier comment, I was thinking more like homebrew casks could, though it shouldn't, generally look for really nonsense pathological permissions cases like this that'll always cause errors and fix them generally.  Sort of like how the macOS Archive Utility seems to override the permissions the .zip file specifies and just uses sensible default permissions instead.  I don't think that's a good practice to get into though and glad to hear confirmation)

ShivaHuang:
I tried `set_permissions` with `preflight` as follows

```ruby
preflight do
    set_permissions "#{staged_path}/MacDown.app/Contents/_CodeSignature", '0755'
end

postflight do
    set_permissions "#{appdir}/MacDown.app/Contents/_CodeSignature", '0000'
end
```

but no luck. It seems the processes is something like
1. Download zip file
2. unzip file in temp folder
3. move unzipped files to `staged_path`
4. `preflight` -> `set_permissions`

But the process breaks on stage 3, so `preflight` doesn't execute at all.

I think it should be fixed from macdown.

vitorgalvao:
> But the process breaks on stage 3, so `preflight` doesn't execute at all.

Ah, right.

There are more things we can try (I seem to remember broken DMGs we could find some workarounds for), but at this point I’d rather wait for a response from upstream.

jnozsc:
upstream ticket https://github.com/MacDownApp/macdown/issues/1144

FranklinYu:
Please try updating the testing branch of MacDown. I applied a fix there.

If the testing branch is updated successfully, I would backport the fix to stable branch, and we can proceed this PR.

FranklinYu:
Good news! We finally released 0.8.0d71 on https://macdown.uranusjr.com/history/testing/. Please try the binary there; we should have removed the troublesome code signature in the binary.

PR #75056, opened 2020-01-03
Update bettertouchtool.rb

core-code:
as discussed in #75004


core-code:
a script updates the text file with the version numbers of the download of bettertouchtool once a day at midnight CET

PR #75184, opened 2020-01-06
Update gcc-arm-embedded to 9-2019q4-major from 8-2018q4-major.

tannewt:
Thank you for adding this cask back! It is really helpful for us.

Here is an update to the version of it.


tannewt:
🎉 Thanks!

vitorgalvao:
Thank *you*.

PR #75247, opened 2020-01-08
Update upic from v0.16.3 to v0.16.4

gee1k:
Emergency repair

### DIFF
--- a/Casks/upic.rb
+++ b/Casks/upic.rb
@@ -1,6 +1,6 @@
 cask 'upic' do
-  version '0.16.3'
-  sha256 '73be6028763fa251ff315fa198735bbc84ebeffd57ee877dddcb5e795ca08ea8'
+  version '0.16.4'
+  sha256 '972da76944ed78591a450eaf9dd00642b3d4ca2cece0b6700c72c2d36ce49c4b'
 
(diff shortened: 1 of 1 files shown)


Casks/upic.rb
```diff
@@ -10,9 +10,4 @@
   depends_on macos: '>= :sierra'
 
   app 'uPic.app'
-
-  zap trash: [
```
vitorgalvao: Why remove these?

gee1k: These files do not need to be deleted but are kept

vitorgalvao: > These files do not need to be deleted but are kept

I don’t understand what you mean. Aren’t these valid configuration files? They’ll only be removed if the user `zap`s the cask, not if they `uninstall` or `upgrade`.

gee1k: Oh, it seems I understand it wrong. I will readjust

Sorry for your trouble

vitorgalvao (approved):

Casks/upic.rb
gee1k: I have readjusted, and I am sorry again for the trouble

vitorgalvao: No need to apologise. It was a small and understandable mistake. Thank you for the quick fix.

PR #75255, opened 2020-01-08
Add tableau-log-viewer v1.2

aogail:

### DIFF
--- a/Casks/tableau-log-viewer.rb
+++ b/Casks/tableau-log-viewer.rb
@@ -2,7 +2,7 @@ cask 'tableau-log-viewer' do
   version '1.2'
   sha256 'd203b4d7d3d4d02d2e4051d3f08e2a506462e543d233a1d952207b830e5a3a4f'
 
-  url 'https://github.com/tableau/tableau-log-viewer/releases/download/1.2/tlv-osx-1.2.zip'
+  url "https://github.com/tableau/tableau-log-viewer/releases/download/#{version}/tlv-osx-#{version}.zip"
   appcast 'https://github.com/tableau/tableau-log-viewer/releases.atom'
   name 'Tableau Log Viewer'
   homepage 'https://github.com/tableau/tableau-log-viewer'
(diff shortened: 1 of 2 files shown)


Casks/tableau-log-viewer.rb
```diff
@@ -0,0 +1,11 @@
+cask 'tableau-log-viewer' do
+  version '1.2'
+  sha256 'd203b4d7d3d4d02d2e4051d3f08e2a506462e543d233a1d952207b830e5a3a4f'
+
+  url 'https://github.com/tableau/tableau-log-viewer/releases/download/1.2/tlv-osx-1.2.zip'
```
schwedenmut: What about using the defined version in the url. Would look like this:
url 'https://github.com/tableau/tableau-log-viewer/releases/download/#{version}/tlv-osx-#{version}.zip'

Casks/tableau-log-viewer.rb
```diff
@@ -0,0 +1,11 @@
+cask 'tableau-log-viewer' do
```
vitorgalvao: Needs to follow the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md), meaning it needs to be `tlv`.

vitorgalvao (changes requested):

Casks/tableau-log-viewer.rb
aogail: Is there a preference for tokenizing the file name instead of the full name from docs?

vitorgalvao: It’s not a preference, it’s a requirement. The token reference goes into detail as to why.

vitorgalvao: It’s not a tokenization, it follows from the bundle’s name. The token reference explains this in detail, as well as the reasons. Have you not read the document? You have ticked the box:

> - [x] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).

aogail: Yes, I read that, I just missed that there was a discrepancy between examples I read and that doc as far as the string to start with when tokenizing. I think in this type of case, where the app file name is less descriptive than the real application name, starting with the file name results in a less useful token.

In any case, I've renamed the cask to `tlv`.

vitorgalvao: > I think in this type of case, where the app file name is less descriptive than the real application name, starting with the file name results in a less useful token.

Not doing so results in duplicates, which is a far greater problem.

PR #75333, opened 2020-01-09
Update wavebox from 4.11.11 to 4.11.12

ran-dall:


core-code:
@vitorgalvao it seems like this has been re-named to "wavebox-classic" and there is a much newer version on their homepage. should i send a PR to update this cask or should we make 2 casks because some users might prefer to stay on the 'classic' version?

vitorgalvao:
Analytics for this are poor. It can just be updated to the newest version.

PR #75483, opened 2020-01-13
Update frhelper from 3.9.4,2019-12-09 to 3.9.4,2019-12-13

suschizu:

### DIFF
--- a/Casks/frhelper.rb
+++ b/Casks/frhelper.rb
@@ -1,9 +1,9 @@
 cask 'frhelper' do
-  version '3.9.4,2019-12-13'
+  version '3.9.4,1037:2019-12-13'
   sha256 '30d03c22f3e3fd0533ef8f223f6db546f922722b4dde9eb91d53c1ce7d2f7660'
 
   # static.frdic.com was verified as official when first introduced to the cask
-  url "https://static.frdic.com/pkg/fhmac.dmg?v=#{version.after_comma}"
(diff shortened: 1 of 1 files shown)


suschizu:
perhaps its better to add the bundle version too because they often just change the BV and keep the SVS (and sometimes also just change the download-token  … like here)
  
and:
`- appcast at URL 'https://www.francochinois.com/update/frhelper_mac.xml' does not contain the version number '3.9.4':`
yes true - this appcast is the SUFeedURL so normally it should be updated … but we are 2 versions behind and I can't find another feed. The release history unfortunately doesn't show the mac version updates - the only way to track the version I think would be to download the app and check the svs - not sure if the effort is worth it (not very popular) @core-code, @vitorgalvao

core-code:
how about using 'https://www.eudic.net/v4/fr/app/download' as the appcast and 'configuration :version.after_colon' so that it only looks for the download-token (which presumably changes every time)?

suschizu:
the download-token changes more often than the version … so sometimes (like this time) they update the download token but not the app itself - but would be oke for me (?)

core-code:
i am not sure. lets see what vitor thinks ;)

vitorgalvao:
Using `https://www.eudic.net/v4/fr/app/download` and the date seems best. But in that case we might as well remove the bundle version: it’s one less value to track (we wouldn’t be using it anyway) and we avoid `colon`.

PR #75564, opened 2020-01-14
Update cisdem-document-reader from 5.0.0 to 5.1.0

ran-dall:

### DIFF
--- a/Casks/cisdem-document-reader.rb
+++ b/Casks/cisdem-document-reader.rb
@@ -1,6 +1,6 @@
 cask 'cisdem-document-reader' do
   version '5.1.0'
-  sha256 '18ccddde919915d46ff60e6fcc23de880ed52d9e1036fd15def9b4148b58515c'
+  sha256 'e621f103edcd8816ca72a1b73cdebaea00e0300f9cf506bcae00178d5fb5965f'
 
   url 'http://download.cisdem.com/cisdem-documentreader.dmg'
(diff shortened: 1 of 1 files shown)


core-code:
sorry @vitorgalvao  but this was wrong. before this PR the cask had 
version 5.0.0 and checksum e621f103edcd8816ca72a1b73cdebaea00e0300f9cf506bcae00178d5fb5965f

which was correct.

after this PR it had version 5.1.0 but still checksum
e621f103edcd8816ca72a1b73cdebaea00e0300f9cf506bcae00178d5fb5965f

i presume the cisdem people have withdrawn the 5.1.0 release

so, we now have the wrong version number in here. i'll send a PR to fix

vitorgalvao:
Something is wrong, likely a caching issue.

I verified this before merging, and the `url` on the cask (unversioned) downloads `5.1.0` *and* it’s on [their release notes](https://www.cisdem.com/document-reader-mac/release-notes.html) (appcast).

I trusted Travis for the checksum, and since it was the same as before, I assumed the mistake was made earlier. Rechecking everything again, I get the same checksum @ran-dall got on the initial submission, and `5.1.0` still.

Seems like the correct fix here is to update the checksum, not revert the version. Being an unversioned URL with a cache issue, both solutions will be problematic to someone, but it makes more sense to bump it up than down.

core-code:
ok, thats weird because i get the old checksum (and version).  but i am happy with any solution as long as the checksum and version actually match.

PR #75584, opened 2020-01-14
Update backblaze from 7.0.0.392 to 7.0.0.402

cameronmalek:

### DIFF
--- a/Casks/backblaze.rb
+++ b/Casks/backblaze.rb
@@ -1,5 +1,5 @@
 cask 'backblaze' do
-  version '7.0.0.392'
+  version '7.0.0.402'
   sha256 'a0967d345e4245b13f4c1a6223e322f9f4a8f03f26dc5e2ad331a311494070ca'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Not sure where you got that version, but the version the `url` downloads and the one on the appcast are both the one currently on the cask.

Also, updating the version without updating the checksum typically does nothing.

cameronmalek:
> Not sure where you got that version, but the version the `url` downloads and the one on the appcast are both the one currently on the cask.
> 
> Also, updating the version without updating the checksum typically does nothing.

I suspect they released then recalled an updated version. This morning when upgrading Backblaze I got a checksum mismatch error and just ran `cask-repair` assuming there was an update. I got the version number from checking my manually updated version and checking for updates, and that is the version number it reported.

I reinstalled just now and it's reporting 7.0.0.392, which is what the cask was already on.

I also went too fast and didn't notice the checksum didn't even change when pushing the update. Sorry about that.

PR #75663, opened 2020-01-16
Update disk-sensei from 1.6.3 to sensei 1.0.0,9

core-code:


core-code:
whole new generation of the app, see homepage

core-code:
new generation needs 10.15, if there is enough demand we could carry the old version in if/else

PR #75664, opened 2020-01-16
Update appcast redirect in ridibooks

suschizu:


suschizu:
the new version is still beta - I'm not sure if we should update this cask to a beta version so I've changed the redirect to the stable version meanwhile

vitorgalvao:
> I'm not sure if we should update this cask to a beta version

If the beta is the officially recommended version (seems like it, as it’s the one on the homepage), [yes](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version) (emphasis added):

> When an App is only available as beta, development, or unstable versions, **or in cases where such a version is the general standard**, then said version can go into the main repo.

PR #75735, opened 2020-01-17
Update lynx from 6.6.2.0 to 6.6.3.9

suschizu:

### DIFF
--- a/Casks/lynx.rb
+++ b/Casks/lynx.rb
@@ -1,5 +1,5 @@
 cask 'lynx' do
-  version '6.6.3.9'
+  version '6.6.3.0'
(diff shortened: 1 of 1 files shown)


Casks/lynx.rb
```diff
@@ -1,6 +1,6 @@
 cask 'lynx' do
-  version '6.6.2.0'
-  sha256 'd63291a6092437fe5a267bf7bdbda36d586854652a0148a08e80c8e48ed376d2'
+  version '6.6.3.9'
```
vitorgalvao: How did you get to the `.9` precision? On the page and appcast I only see until `6.6.3.0` and the app only reports `6.6.3`.

PR #75825, opened 2020-01-19
Update inkscape's shimscript

vladimyr:

### DIFF
--- a/Casks/inkscape.rb
+++ b/Casks/inkscape.rb
@@ -24,7 +24,7 @@ cask 'inkscape' do
     preflight do
       IO.write shimscript, <<~EOS
         #!/bin/sh
-        exec '#{staged_path}/Inkscape.app/Contents/MacOS/Inkscape' "$@"
+        open -a inkscape "$@"
       EOS
     end
 


Casks/inkscape.rb
```diff
@@ -24,7 +24,7 @@
     preflight do
       IO.write shimscript, <<~EOS
         #!/bin/sh
-        exec '#{staged_path}/Inkscape.app/Contents/MacOS/Inkscape' "$@"
+        open -a inkscape "$@";
```
vitorgalvao: The `;` is unnecessary.

I understand the difference between both commands, but *why* are you making the change?

vladimyr: > The `;` is unnecessary.

Shell scripting is definitely not something I'm good at 🙃I'll remove it.

> I understand the difference between both commands, but _why_ are you making the change?

Because invoking binary with `exec` will start inkscape process as child process of given terminal session. It will keep running in foreground, effectively _stealing_ terminal session and converting it to gtk debug log stream. Also if you terminate that terminal session (by closing your console), inkscape goes down too.
OTOH `open` will launch inkscape in background and return stdio handles to caller giving you the same effect as clicking on the app icon which is probably what users expect when typing:
```
$ inkscape /path/to/image.svg
```

vladimyr: > Because invoking binary with `exec` will start inkscape process as child process of given terminal session. It will keep running in foreground, effectively _stealing_ terminal session and converting it to gtk debug log stream. Also if you terminate that terminal session (by closing your console), inkscape goes down too.

After doing more research (consulting official sources: https://wiki.inkscape.org/wiki/index.php/Using_the_Command_Line) seems like this is irritating, albeit expected behavior and shouldn't be altered by brew so I'm closing this PR.

PR #75829, opened 2020-01-19
Change version from 4.6 to 4.6.18 progressive-downloader.rb

suschizu:
would it be possible to use the whole version number here? There are already 5 different versions of 4.6.x


suschizu:
hey @vitorgalvao - I'm sorry - I screwed this up :( the real version is still 4.6 CFBundleShortVersionString:
4.6
CFBundleVersion:
4.6.9532
all other versions are beta versions :/

Sorry for that!

PR #75910, opened 2020-01-21
Update marginnote.rb from 3.5.9007,1576409180 to 3.6.1002,1451f4f692144d46b91cd5898c7e42cb:6 

suschizu:
the appcast and download url have changed

### DIFF
--- a/Casks/marginnote.rb
+++ b/Casks/marginnote.rb
@@ -1,10 +1,10 @@
 cask 'marginnote' do
-  version '3.5.9007,1576409180'
+  version '3.6.1002,1451f4f692144d46b91cd5898c7e42cb:6'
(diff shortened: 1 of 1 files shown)


Casks/marginnote.rb
```diff
@@ -1,10 +1,10 @@
 cask 'marginnote' do
-  version '3.5.9007,1576409180'
-  sha256 '4132d81bd9b88ea1cbf05385e64431210ea03c8bba152a1a799651264eae4245'
+  version '3.6.1002,1451f4f692144d46b91cd5898c7e42cb:6'
```
vitorgalvao: Isn’t the `1451f4f692144d46b91cd5898c7e42cb` fixed? Both releases in the appcast share it.

suschizu: you are right - looks better now :)

PR #76072, opened 2020-01-24
transocks v2.4.2: tweak uninstall and zap stanza

i0ntempest:
Paths in uninstall stanza contains binaries that should be removed on uninstall.

### DIFF
--- a/Casks/transocks.rb
+++ b/Casks/transocks.rb
@@ -10,9 +10,7 @@ cask 'transocks' do
   pkg "Transocks_Mac_#{version}.pkg"
 
   uninstall pkgutil: 'com.穿梭Transocks',
-            trash:   [
-                       '~/.transocks_vendor',
-                     ]
+            trash:   '~/.transocks_vendor'
 
   zap trash: [
                '/Library/Logs/DiagnosticReports/穿梭Transocks*',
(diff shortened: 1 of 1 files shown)


reitermarkus:
Why do they need to be removed on uninstall?

i0ntempest:
This app works by first extracting electron app into the user directory and then run that with args. The app in /Applications is basically a launcher, what’s actually running is the electron app inside the user dir (in ~/.transocks_vendor/). So I think it should be deleted on uninstall.

reitermarkus:
At least the `.tmp` and `.log` should still be in `zap` in that  case.

i0ntempest:
Okay I’ll change it when I have my Mac (or you can edit my PR if you want)

Casks/transocks.rb
```diff
@@ -9,14 +9,16 @@
 
   pkg "Transocks_Mac_#{version}.pkg"
 
-  uninstall pkgutil: 'com.穿梭Transocks'
+  uninstall pkgutil: 'com.穿梭Transocks',
+            trash:   [
+                       '/Users/*/.transocks_vendor',
```
vitorgalvao: Why are you using `/Users/*` instead of `~/`?

i0ntempest: Because it seems `~/`only applies to the current user running the command? I could be wrong but I think all installed files in all users' home should be removed.

vitorgalvao: While that’s accurate, the user that’s uninstalling has no business messing with the home directories of other users.

PR #76094, opened 2020-01-24
Suppress checksum verification in Vuescan

yanokwa:

- [X] The commit message includes the cask’s name ~~and version~~.


core-code:
in this case we should probably also change the version to 9.7

yanokwa:
Thanks for the feedback, @core-code. I've made the change, squashed the extra commits, and force-pushed.

vitorgalvao:
It’s still useful to have the `.20` because changing that will allow the cask to be updated on `upgrade`, considering this does not have `auto_updates`.

PR #76235, opened 2020-01-27
Add Matterhorn 50200.6.0

dAnjou:

One of the project members gave their blessing for this PR in https://github.com/matterhorn-chat/matterhorn/issues/482#issuecomment-578916122.

### DIFF
--- a/Casks/matterhorn.rb
+++ b/Casks/matterhorn.rb
@@ -1,0 +1,11 @@
+cask 'matterhorn' do
+  version '50200.6.0'
+  sha256 '773fed52878e901bfac2998048336c51cb2f406b62c7e6ea7f1f5d30616b0e33'
+
+  url "https://github.com/matterhorn-chat/matterhorn/releases/download/#{version}/matterhorn-#{version}-Darwin-x86_64.tar.bz2"
+  appcast 'https://github.com/matterhorn-chat/matterhorn/releases.atom'
+  name 'Matterhorn'
+  homepage 'https://github.com/matterhorn-chat/matterhorn'
+
+  binary "matterhorn-#{version}-Darwin-x86_64/matterhorn"
+end


vitorgalvao:
This seems to be a CLI-only open-source tool. Please submit it first to [Homebrew/core](https://github.com/homebrew/homebrew-core) as a formula. Return here if not accepted (linking back to the pull request where it was refused).

This policy is documented in [faq/rejected_casks](https://github.com/homebrew/homebrew-cask/blob/master/doc/faq/rejected_casks.md):

jtdaugherty:
Greetings! I'm the maintainer for the [Matterhorn](https://github.com/matterhorn-chat/matterhorn) project that was rejected in this submission. If it's helpful, I want to comment on why building this project from source is probably not a good idea for Homebrew users, and definitely not a good idea from my perspective as the project maintainer.

Matterhorn is a Haskell project, and when Haskell projects are built from source, they fetch their dependencies in source form from public Haskell source code hosts, most commonly `hackage.haskell.org`. Those dependencies will not be in Homebrew in binary form (since it doesn't make sense to distribute binary Haskell packages in Homebrew) so a source build is the only approach possible in that case, unless Matterhorn can be distributed in binary form to avoid a source build. However, doing a source build of those dependencies means that getting a deterministic result could be tricky, because the Haskell build toolchain can select different versions of dependencies for different environments, and that would make supporting Matterhorn under those conditions onerous.

On the other hand, if we ship binaries via Homebrew, I imagine that can get tricky because Matterhorn will have runtime linking dependencies that may or may not be available in Homebrew, depending on the version of the Haskell toolchain that was used to build the program, and depending on the version of a few shared libraries that it uses.

I'm not yet entirely sure it makes sense to distribute Matterhorn via homebrew in binary form, but I am not sure that distributing it in source form will not buy Homebrew users very much. Furthermore, the development team publishes stable binary releases on a select few platforms (including Darwin hosts) but does not support an arbitrary array of build toolchain versions, which is another reason I am hesitant about making a claim that we can support source builds on Homebrew.

I'd love others to share their thoughts on the situation.

jtdaugherty:
To summarize my previous comment, in case it isn't clear: to me, submitting to `homebrew/core` as a source package is not something I think is worth doing for the reasons I outlined, especially if the outcome (that I would want!) is for that to get rejected there.

vitorgalvao:
I’m convinced by your arguments. Thank you for taking the time to lay them out.

PR #76313, opened 2020-01-29
Update amazon-workspaces from 2.5.11 to 3.0.2

jlholtus:


vitorgalvao:
Thank you, but this doesn’t look correct.

For one, if the version is different, so does the shasum needs to be. There’s also no mention of this version in the appcast.

jlholtus:
You're right. The download link on their own site still points to the old location. I ended up using their updater to install the latest version instead.

It would be great to have the latest version here as well but I couldn't readily get the needed info.

PR #76382, opened 2020-01-30
Update readme.md

core-code:
since the numeric comparison isn't even in a single cask i suggest we don't document that it should be done

### DIFF
--- a/doc/cask_language_reference/readme.md
+++ b/doc/cask_language_reference/readme.md
@@ -39,18 +39,13 @@ Tests on the following values are known to be acceptable:
 
 ### Version Comparisons
 
-Tests against `MacOS.version` may use either symbolic names or version
-strings with numeric comparison operators:
+Tests against `MacOS.version` may use the symbolic names with numeric comparison operators:
 
 ```ruby
 if MacOS.version <= :mavericks     # symbolic name
 ```
 
-```ruby
-if MacOS.version <= '10.9'         # version string
-```
-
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Since the code to do it exists, I’d rather either remove that code first or keep it documented *but* have a warning that only the symbolic names are allowed in the official repos.

core-code:
breaking 3rd party taps sounds like a bad idea, so option 2 sounds better?

vitorgalvao:
> so option 2 sounds better?

Agreed. Less work, too.

doc/cask_language_reference/readme.md
```diff
@@ -39,18 +39,13 @@ Tests on the following values are known to be acceptable:
...
 ```ruby
 if MacOS.version <= :mavericks     # symbolic name
 ```
 
-```ruby
-if MacOS.version <= '10.9'         # version string
-```
-
-The available symbols for macOS versions are: `:mavericks`, `:yosemite`, `:el_capitan`, `:sierra`, `:high_sierra`, and `:mojave`. The corresponding numeric version strings should be given as major releases containing a single dot.
+The available symbols for macOS versions are: `:mavericks`, `:yosemite`, `:el_capitan`, `:sierra`, `:high_sierra`, and `:mojave`.
```
miccal: Remove `:mavericks` :

https://github.com/Homebrew/homebrew-cask/issues/58046#issuecomment-460004001

core-code: thanks there. i've added catalina, but wasn't sure about removing. officially we only support sierra and up, so only removing mavericks seemed kinda arbritrary...

PR #76393, opened 2020-01-30
Update sketchup-pro.rb from 19.3.252,2019 to 20.0.362,2020

suschizu:
this time it worked :)


suschizu:
the .dmg contains an installer now :/

PR #76486, opened 2020-02-01
Update finereader

core-code:


core-code:
the old URL seems to be no longer in use as it only has an outdated version.

if the 'secure' part in the URL changes or becomes invalid, i guess we need to remove this cask.

PR #76526, opened 2020-02-02
Add Stack 2.7.3

danielbayley:

### DIFF
--- a/Casks/stack-stack.rb
+++ b/Casks/stack-stack.rb
@@ -1,0 +1,23 @@
+cask 'stack-stack' do
+  version '2.7.3'
+  sha256 '4e52945fcdc1b7f3ff5e72245493074dcebd47cd00b6cc8431f62f68c9cfdade'
+
+  # s3.eu-central-1.amazonaws.com/stack-v1 was verified as official when first introduced to the cask
+  url "https://s3.eu-central-1.amazonaws.com/stack-v1/builds/prod/Stack-#{version}.dmg"
+  appcast 'https://s3.eu-central-1.amazonaws.com/stack-v1'
+  name 'Stack'
+  homepage 'https://stackers.app/'
+
+  auto_updates true
+  depends_on macos: '>= :yosemite'
+
+  app 'Stack.app'
+
+  uninstall quit:       'com.stack.io',
+            login_item: 'Stack'
+
+  zap trash: [
+               '~/Library/Application Support/Stack',
+               '~/Library/Preferences/com.stack.io.plist',
+             ]
+end


danielbayley:
Not sure what the policy is for replacing a formula like this? This app can only really have `stack` as a token… and the 'vendor' seems to be just the same name. Whereas the existing `stack.rb` Cask can easily be prefixed with the vendor as `transip-stack`… So what to do in this situation?

vitorgalvao:
The existing cask isn’t insanely popular, but it might still pose a problem to rename without a proper mechanism in place (which we don’t have)

```
stack (added 1355 days ago)
30 days: 26 (#1529)
90 days: 86 (#1410)
365 days: 301 (#1472)
```

You can call the new one `stack-stack` (vendor-appname). I think we have precedent.

danielbayley:
> You can call the new one stack-stack

Seems a bit daft, but ok…

vitorgalvao:
> Seems a bit daft, but ok…

I agree it looks weird, but unless you want to implement the equivalent of `formula_renames` and submit a PR before getting the cask in, there’s not really a good option.

danielbayley:
> implement the equivalent of `formula_renames`

@vitorgalvao What do you mean by this exactly? A new stanza?

vitorgalvao:
> @vitorgalvao What do you mean by this exactly? A new stanza?

No, it’s a JSON file. See [the documentation](https://docs.brew.sh/Rename-A-Formula) and [a working example](https://github.com/Homebrew/homebrew-core/blob/master/formula_renames.json).

danielbayley:
> No, it’s a JSON file. See [the documentation](https://docs.brew.sh/Rename-A-Formula) and [a working example](https://github.com/Homebrew/homebrew-core/blob/master/formula_renames.json).

@vitorgalvao Right, so the process is exactly the same for a Cask?

vitorgalvao:
> @vitorgalvao Right, so the process is exactly the same for a Cask?

No, the process doesn’t work for casks (but we’d like it to). Therein lies the problem.

danielbayley:
@vitorgalvao Even if it did, would adding a new formula/Cask with the same name as the previously renamed one work reliably?

vitorgalvao:
> Even if it did, would adding a new formula/Cask with the same name as the previously renamed one even work reliably?

It think so, because it’s supposed to affect current installs, not newer ones.

core-code:
since the requested rename has been done, is this good to merge?

vitorgalvao:
> since the requested rename has been done, is this good to merge?

Yes.

PR #76632, opened 2020-02-04
Corrected Checksum

jstader:
I encountered the following error when upgrading:
'Error: Checksum for Cask 'backblaze' does not match.
Expected: b990ebf41bd6703ec46f5cfc298169c26194d2f10c8e74e64c0cd340ed7a2439
Actual: a0967d345e4245b13f4c1a6223e322f9f4a8f03f26dc5e2ad331a311494070ca'

### DIFF
--- a/Casks/backblaze.rb
+++ b/Casks/backblaze.rb
@@ -1,6 +1,6 @@
 cask 'backblaze' do
   version '7.0.0.412'
-  sha256 'a0967d345e4245b13f4c1a6223e322f9f4a8f03f26dc5e2ad331a311494070ca'
+  sha256 ':no_check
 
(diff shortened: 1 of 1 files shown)


jstader:
There seem to be multiple checksums. The CI seems to be getting the correct checksum as listed in the master. It might be worth updating the sha sum to :no_check

vitorgalvao:
> There seem to be multiple checksums.

If there are multiple checksums, then there are multiple different packages, and that has to be reported upstream (i.e. to backblaze). At worst, that means they were breached.

Please don’t ask for specific reviewers so soon after opening a PR. We’re all volunteers, so we’re not always available. We get to issues as we have the chance.

PR #76634, opened 2020-02-04
Only run workflows on main repo

EricFromCanada:
Prevent the automerge and sync workflows from running on forks of Homebrew/homebrew-cask.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   automerge:
+    if: startsWith( github.repository, 'Homebrew/' )
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


vitorgalvao:
I think the actions are also auto-synced between repos, so this would disable it on the other taps (`Homebrew/homebrew-cask-versions`, `Homebrew/homebrew-cask-drivers`, `Homebrew/homebrew-cask-fonts`).

Is there a simple solution to it (e.g. making the line match a regex)?

EricFromCanada:
Using [`startsWith`](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#startswith) should fix it.

PR #76718, opened 2020-02-06
Update fsnotes from 4.0.20 to 4.0.21

ulab:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #76755, opened 2020-02-07
Revert "Update fsnotes from 4.0.20 to 4.0.21"

core-code:
Reverts Homebrew/homebrew-cask#76718


core-code:
ideally we'd improve our appcast parsing to not let updates to "pre-release" versions through automatically.

PR #76756, opened 2020-02-07
Update postman

sz4b0lcs:

### DIFF
--- a/Casks/postman.rb
+++ b/Casks/postman.rb
@@ -1,6 +1,6 @@
 cask 'postman' do
   version '7.17.0'
-  sha256 '23650d8bc898b5654701cc2c20db24601dbe4f7cd0a99524d647d2b0bc9fa910'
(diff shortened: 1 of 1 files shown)


sz4b0lcs:
At first I thought that only the checksum is the problem but after submitting this, I tried to install it myself by changing the checksum in the cask and it fails with:
```
Error: It seems the App source '/usr/local/Caskroom/postman/7.17.0/Postman.app' is not there.
```

After investigating some more it turn out, that it downloads a canary version (with the app name PostmanCanary.app) from here [https://dl.pstmn.io/download/version/7.17.0/osx64](https://dl.pstmn.io/download/version/7.17.0/osx64).
When I try to download it by hand it downloads the proper version.

Can this be some caching issue? I don't quite understand it.

Thanks for your help.

core-code:
i think this was a temporary fuckup on part of upstream and is fixed now

vitorgalvao:
> i think this was a temporary fuckup on part of upstream and is fixed now

Looks like it.

Thank you anyway, @sz4b0lcs, and welcome.

PR #76810, opened 2020-02-08
Update appcast in  segger-embedded-studio-for-arm.rb

suschizu:
the other one had a ssl problem


vitorgalvao:
Why’s the `macupdater.net` one failing? Can you shed some light, @core-code? I’d rather have all of your CGI appcasts on the same domain, for easy grepping when necessary.

core-code:
well, it is failing because strictly speaking their SSL is broken:

"https://www.ssllabs.com/ssltest/analyze.html?d=www.segger.com"

python 3 seems to be picky here and refuse connection. python2 on the other server doesn't mind.

unfortunately we had the other situation too where only the macupdater.net server works for some SSLv3 (?) sites...

not sure about a good solution here, i can definitely understand not wanting to rely on 2 different CGI servers.

core-code:
i see that they have release notes here:

https://studio.segger.com/index.htm?https://studio.segger.com/arm_segger_studio_release_notes.htm

suschizu:
does not contain the version unfortunately (curl)

core-code:
this seems to work fine for me

` curl https://studio.segger.com/arm_segger_studio_release_notes.htm`

vitorgalvao:
Thank you.

PR #75324, opened 2020-01-09
Update alacritty from 0.4.0 to 0.4.1

ran-dall:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #76879, opened 2020-02-11
Revert "Update alacritty from 0.4.0 to 0.4.1 (#75324)"

Julian:
This reverts commit 3e733d8663a6e165d1e648bc166def9328fe2fd8.

alacritty/alacritty#3188 is a fairly significant bug.

### DIFF
--- a/Casks/alacritty.rb
+++ b/Casks/alacritty.rb
@@ -1,6 +1,6 @@
 cask 'alacritty' do
-  version '0.4.1'
-  sha256 '9718c3b0b5cd430cac162a15f5b7a1e84adb7eba3974c6a391bce96e9d2d2e38'
+  version '0.4.0'
+  sha256 'db9f75dc3221709c86362e6de0e3671d328a8bd8eb306122b901ca11e142757c'
 
   url "https://github.com/jwilm/alacritty/releases/download/v#{version}/Alacritty-v#{version}.dmg"
   appcast 'https://github.com/jwilm/alacritty/releases.atom'


core-code:
if the bug is so significant, then upstream should pull the release. if they do, we can follow.

Julian:
Hm, what does "pulling a release" mean for a generic project? The release was done, it's out. There's not much benefit in deleting the artifacts, they're moving forward, and will put out a new release when their (transitive) dep is fixed.

Up to you guys though.

vitorgalvao:
> Hm, what does "pulling a release" mean for a generic project? The release was done, it's out.

It means taking back the release, which is standard procedure in these cases. If a severe bug happens, saying “oh well, we’ve released it now so let’s leave it out for new people to be affected” isn’t good practice.

> There's not much benefit in deleting the artifacts

There is *a ton* of benefit. Most people don’t follow bug trackers, they download the software and get on with their lives. The longer this stays out, the more people will get the buggy release and the worse it will be for the reputation of their software.

That version has been out in cask form for a month and is highly popular:

```
alacritty (added 454 days ago)
30 days: 2,687 (#47)
90 days: 7,074 (#56)
365 days: 28,369 (#54)
```

Maybe this is affecting a minority of people, or maybe upstream is choosing to keep that version up for other reasons. Either way, I agree we should follow their lead on this one, they’re the experts on their software. If they do an *official* recommendation that people should stay on `0.4.0`, I’ll be all for it.

Thank you anyway for the PR and letting us know.

Julian:
> saying “oh well, we’ve released it now so let’s leave it out for new people to be affected” isn’t good practice.

I won't belabor the point, since I don't really want to argue here, it's your repo, I've already fixed this for myself (so presumably others who want this will chime in if they do, otherwise hopefully they just put a release out), but alacritty is cross-platform software.

The recommendation they have *is* to go back to 0.4.0, as linked in that issue, they're not going to delete the release for a subset of their users (which I agree with, deleting stuff that's out is not great practice, but it sounds like you disagree).

If you already saw that (that their recommendation is to go back to 0.4.0) and still want to close fair enough, but just making sure to point that out.

core-code:
they recommend to go back to 0.4.0 if (and only if) you are affected. 

if they thought /everyone/ should go back to 0.4.0, the 0.4.1 version wouldn't be available for download anymore.

with any complex project, there are always be people that think the latest release is unusable because of a showstopper bug. the only sane thing for us to do here is to follow upstream - if they pull a release, we should follow. i don't see any other procedure as working. who should we trust about what bugs are critical and what are not? if you think it though we can only trust the upstream vendor and they have decided to keep the release.

vitorgalvao:
> I won't belabor the point, since I don't really want to argue here

Agreed, neither do I. Whichever way it goes, we’re still grateful for your submission.

> The recommendation they have _is_ to go back to 0.4.0

I haven’t read it all, but I’ve only found a regular user (not repo member) making that suggestion and a contributor (also not member) recommending building from source. Are you able to point me to the specific recommendation made by a maintainer?

> they're not going to delete the release for a subset of their users (which I agree with, deleting stuff that's out is not great practice, but it sounds like you disagree).

They wouldn’t need to delete whole release, simply remove the downloads for the affected OS. That’s not uncommon; often multi-platform software has releases for a subset of supported OSs.

> If you already saw that (that their recommendation is to go back to 0.4.0) and still want to close fair enough, but just making sure to point that out.

That is appreciated. If you can point to a recommendation made by a project maintainer, I’ll gladly take another look.

Julian:
> Agreed, neither do I. Whichever way it goes, we’re still grateful for your submission.

Thanks, and same, I'm sure telling fly-by-contributors the same thing over and over again isn't fun.

> I haven’t read it all, but I’ve only found a regular user (not repo member) making that suggestion and a contributor (also not member) recommending building from source. Are you able to point me to the specific recommendation made by a maintainer?

No sorry, it's what you saw from kchibisov (who I just saw had significant numbers of contributions, they were like #3 or something in the list).

Anyways, fair enough on just leaving it until they clarify what they recommend or hopefully just push out a fix. Maybe I'll leave a comment on the ticket and see what they say.

OJFord:
This doesn't affect _all_ Alacritty users though, but it is a very significant issue for the macOS portion.

Since 100% of homebrew-cask downloading Alacritty users are on macOS, it makes (even) more sense for the homebrew-cask version to revert than it does for Alacritty as a whole.

vitorgalvao:
[SoftU2F is facing a similar situation, and they chose a good way to go about it](https://github.com/github/SoftU2F/issues/73#issuecomment-591481475).

PR #76935, opened 2020-02-12
Update tor-browser from 9.0.4 to 9.0.5

miccal:
`/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/developer/bin/update_multilangual_casks /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Casks/tor-browser.rb 9.0.5`

### DIFF
--- a/Casks/tor-browser.rb
+++ b/Casks/tor-browser.rb
@@ -1,163 +1,163 @@
 cask 'tor-browser' do
-  version '9.0.4'
+  version '9.0.5'
 
   language 'ar' do
-    sha256 '435143ccf5012de8d7bfbd9ab00b81e850f145f1c19c27f16af2917e05230d13'
+    sha256 'f2688ac10736d4e8eb88b362b9bbac5e9569e7379f0414141e58d6e1863612cf'
     'ar'
   end
 
   language 'ca' do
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao can we switch to a more sane appcast here? the 9.0.5 version is available neither on their official download website nor via the built-in update check.

vitorgalvao:
@core-code Sure, I have no issue with changing appcasts for the better. That said, I do see `9.0.5` on the website and I think this situation has happened before, so if that appcast is early (in relation to the website and the built-in updater) it seems to be by less than a day, which doesn’t seem problematic.

core-code:
> That said, I do see 9.0.5 on the website and I think this situation has happened before, so if that appcast is early (in relation to the website and the built-in updater) it seems to be by less than a day, which doesn’t seem problematic.

yes it seems to be have been around a day. in principle, we should avoid jumping the gun and post things just because found them on the download server.  in practice, i agree, it may not make much difference in /most/ cases and certainly didn't here.

i didn't find anything better than 'https://www.torproject.org/download/' to use. should i send a PR?

vitorgalvao:
> i didn't find anything better than 'https://www.torproject.org/download/' to use. should i send a PR?

I don’t have strong feelings on it, so I’m fine with the change.

PR #77028, opened 2020-02-13
Update raw-photo-processor from 1876Beta to 1896Beta

frihamn:


vitorgalvao:
Why’s this at a beta? It should be on the stable version.

apropostech:
It's not really a beta. These are release quality according to the developer who makes no distinction. It's a recurring issue and most recently discussed here: https://github.com/Homebrew/homebrew-cask/pull/62562

I proposed adding a comment in the cask to avoid repetition of the issue but understand that it goes against the syntax. But is there anywhere else we can add a word to clarify this to avoid coming back to it?

apropostech:
Despite there not being any difference, in the view of the developer, I added a beta to the versions tap: https://github.com/Homebrew/homebrew-cask-versions/pull/8587

That should avoid this issue going forward and leaves it for the user to decide which one to use.

vitorgalvao:
The issue isn’t that it’s a beta, but that the beta is not the version provided on the website, which is what we want to follow. A user should be able to visit the website and know that the beta is what they want. Currently that is not the case.

There could be a case for it if the non-beta was terribly outdated, but that doesn’t seem to be the case at all.

If betas really are release quality, why the distinction?

apropostech:
They were significantly outdated until the developer updated it once recently so I think that's a bit of a hasty conclusion. Before then, it was a version from 2014.

I don't think we can count on the developer starting to publish the releases in a more normal fashion just because he has updated it once since 2014, but I agree it would make more sense if he did so. In the end, that's not up to homebrew but the developer though...

I asked the developer before adding it to homebrew (see link above) and he clarified that there is no difference. I didn't bother to ask him why make it complicated as it's his decision on how he labels and publishes it.

It's also impossible to check when the version on the website has been updated without actually downloading it.

But I understand your reasoning. Just a pity that you won't allow adding a beta to the versions tap then. Even if there's only been one download, I suspect I'm not the only one using homebrew to get the latest updates of this.

vitorgalvao:
I’m willing to make the beta version the normal again. But for that I’d first like to understand how a regular user can know of new beta releases in the first place, to see if we can construct an appcast.

apropostech:
That would be great. The developer posts the latest updates in [the support forum](https://groups.google.com/forum/#!forum/raw-photo-processor). Mostly in reply to user issues. E.g., the latest fix in reply to an issue with mac os 10.15: https://groups.google.com/d/msg/raw-photo-processor/vd2MeN9CEYg/uRb9VXqUDQAJ

It doesn't auto-update so I don't think there's any appcast to use unfortunately. Also, the current appcast doesn't work in any context as it's still linking to an old version from 2014.

It's quite messy, but all the more reason to for users to use homebrew as we can quickly update it when noticing a new version posted to the forum by the developer.

vitorgalvao (approved):

PR #77039, opened 2020-02-13
Update capto from 1.2.16 to 1.2.17

suschizu:


core-code:
the app itself now does a weird POST request to that URL, not sure if we can hack that

suschizu:
I can't find another appcast unfortunately :(
`- appcast at URL 'https://apicapto.globaldelight.net/capto-check-update.php' does not contain the version number '1.2.17':
==> Downloading https://d3l6g06uqih57x.cloudfront.net/Captomac/webstore/Capto.dm
 `

core-code:
maybe we need to make the cask unversioned? it has a static download URL anyway

PR #77040, opened 2020-02-13
Update mamp from 5.6 to 5.5

suschizu:

### DIFF
--- a/Casks/mamp.rb
+++ b/Casks/mamp.rb
@@ -1,6 +1,6 @@
 cask 'mamp' do
-  version '5.6'
-  sha256 '0b5f8157d2a6f4cef39d3cf89563bab6503d2d1087f436f8ca9ee33731293453'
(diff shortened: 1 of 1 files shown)


suschizu:
mamp has been downgraded to 5.5

core-code:
hm...maybe try restarting travis?

suschizu:
`installer: The install failed (The Installer encountered an error that caused the installation to fail. Contact the software manufacturer for assistance.)`

core-code:
@vitorgalvao any ideas here? since version 5.6 has been pulled by upstream it is probably quite buggy and we should follow suit...

vitorgalvao:
Agreed. As for the error, since it’s a downgrade I’ll assume for now that CI is the one causing issues (as it happens often with PKGs) and will revisit if we get user complaints.

core-code:
thanks!

PR #77130, opened 2020-02-14
Update remote-desktop-manager from 2020.1.2.0 to 2020.1.3.0

ran-dall:


vitorgalvao:
I see that this version is already on the release notes, but the version on the download page (both the one they reference and the one the download points to) is still the one in the cask.

PR #77254, opened 2020-02-18
Update popsql from latest to 0.9.4

core-code:

### DIFF
--- a/Casks/popsql.rb
+++ b/Casks/popsql.rb
@@ -1,6 +1,6 @@
 cask 'popsql' do
-  version '0.9.3'
-  sha256 'ecc6e0871bb6299452bdea53a509345d52ac746b91c5e4c9cfbe02bbbd6fc65d'
+  version '0.9.4'
+  sha256 '4fd9581f7933acf5ac04bf2c3883028d9ef2f92b7623b05dfce598479fd0808a'
 
   # popsql-releases.s3.amazonaws.com was verified as official when first introduced to the cask
   url "https://popsql-releases.s3.amazonaws.com/mac/PopSQL-#{version}-mac.zip"
--- a/Casks/popsql.rb
+++ b/Casks/popsql.rb
@@ -1,6 +1,6 @@
 cask 'popsql' do
-  version '0.9.2'
-  sha256 '9e97c094d3217a4a5f4ab93751615da5793d0a1c545d871f36f77d14e5ac9920'
+  version '0.9.3'
+  sha256 'ecc6e0871bb6299452bdea53a509345d52ac746b91c5e4c9cfbe02bbbd6fc65d'
 
   # popsql-releases.s3.amazonaws.com was verified as official when first introduced to the cask
   url "https://popsql-releases.s3.amazonaws.com/mac/PopSQL-#{version}-mac.zip"
--- a/Casks/popsql.rb
+++ b/Casks/popsql.rb
@@ -1,8 +1,10 @@
 cask 'popsql' do
-  version :latest
-  sha256 :no_check
+  version '0.9.2'
+  sha256 '9e97c094d3217a4a5f4ab93751615da5793d0a1c545d871f36f77d14e5ac9920'
 
-  url 'https://get.popsql.io/download/mac'
+  # popsql-releases.s3.amazonaws.com was verified as official when first introduced to the cask
+  url "https://popsql-releases.s3.amazonaws.com/mac/PopSQL-#{version}-mac.zip"
+  appcast 'https://popsql-releases.s3.amazonaws.com/mac/latest-mac.yml'
   name 'PopSQL'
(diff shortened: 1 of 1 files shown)


core-code:
reason for PR: website download only has version 0.9.2

@vitorgalvao FYI, the find_appcast didn't find the electron feed on this one

vitorgalvao:
> FYI, the find_appcast didn't find the electron feed on this one

I’ll rewrite that script. I’ve been wanting to consolidate it for a while. I’ll use the chance to remove support for devmate and hockeyapp appcasts, as those were never thorough and both services have been discontinued by now.

This appcast case complicates things, as it’s in conflict with another case already in the script, meaning that if we detect this one, we won’t detect the case already there (for reference, the script should have produced an appcast, but seeing it was invalid it didn’t output it).

I’ve looked at the documentation for `electron-builder`’s auto-updater and they mention `app-update.yml` is auto-generated. Unfortunately, the source it’s auto-generated from shouldn’t be available in the app itself so there may not be a good way to extract the exact URL. I think the solution will be to internally generate every valid `electron-builder` appcast format we know of and check them all, outputting whichever is valid. If none is but `app-update.yml` was found anyway, output a warning so we can inspect it and determine yet another URL format.

---

On to the cask itself. Looks like this is now a walled build. And it’s one of those that can’t be bypassed with a dummy email address. We’ll need to either remove the cask or revise our walled build rule (e.g. by explaining how to confirm it). In this case we can get to the walled URL by downloading the non-walled `https://get.popsql.com/` and inspecting its appcast.

Another note. The download they provide on the page is for `0.9.0`, despite the appcast being already at `0.9.2`. Are you OK with the discrepancy?

core-code:
rewriting the script and removing the discontinued services sounds like a good idea.

i am not sure i follow on the problems with extracting the url. at least in this case it sounds easy, all the details are in the app-update-yml ( provider: s3|bucket: popsql-releases|region: us-east-1|path: mac| ). also, there must be a way to get the URL,  else how would electron/squirrel determine the URL internally? in any case, i am short on technical details at the moment and i trust you've done the analysis of the situation correctly.

i am fine with removing the cask if it violates our rules - it is extremely unpopular anyway. but relaxing the rules might be interesting for other cases too anyway...

> Another note. The download they provide on the page is for 0.9.0, despite the appcast being already at 0.9.2. Are you OK with the discrepancy?

well, thats why i've sent the PR ;) if you'd rather stick to the version available directly on the website we could use this as an appcast: 'https://macupdater.net/cgi-bin/check_urls/check_url_filename.cgi?url=https://get.popsql.com/&user_agent=Mac%20OS%20X'

vitorgalvao:
> all the details are in the app-update-yml ( provider: s3|bucket: popsql-releases|region: us-east-1|path: mac| )

Correct. However it’s not explicit how those fit together. [We’re aware of some patterns](https://github.com/Homebrew/homebrew-cask/blob/832ad3a4d3facb2fe421e516001beb47e8f9f709/developer/bin/parse_electron_builder_appcast#L27L29) but then the question becomes “when to use each?”. We deduced some cases—if `channel` exists use one pattern; if `region` exists use another; if neither, use a third one—but this case doesn’t fit either of them. It has `region`, but the pattern produced by it (which worked for another case) is an invalid URL and there’s nothing else unique about it.

So the issue isn’t the lack of information, but that the same amount of information can be used in different URL patterns, meaning we need to test them all as opposed to the current method of “use the first that fits”.

> also, there must be a way to get the URL, else how would electron/squirrel determine the URL internally?

True, but it’s like those apps where inspecting traffic ends up being the viable alternative—they also must have the URL stored somewhere, but we don’t know where to get it.

> in any case, i am short on technical details at the moment and i trust you've done the analysis of the situation correctly.

I could still try to read `electron-builder`’s source or ask their developers, though it remains to see if that’ll be worth it.

> if you'd rather stick to the version available directly on the website

Not at all, just wanted to make sure this is what you wanted.

core-code:
> So the issue isn’t the lack of information, but that the same amount of information can be used in different URL patterns, meaning we need to test them all as opposed to the current method of “use the first that fits”.

ok thanks for the explanation there. i agree that testing the patterns looks like a sensible way for improvements here.

> True, but it’s like those apps where inspecting traffic ends up being the viable alternative—they also must have the URL stored somewhere, but we don’t know where to get it.

ah ok. i was under the impression that the 'app-update-yml' is that 'somewhere'  but even just looking at the Squirrel documentation ( https://github.com/Squirrel/Squirrel.Mac ) confirms that it is not.

> I could still try to read `electron-builder`’s source or ask their developers, though it remains to see if that’ll be worth it.

probably not.

> Not at all, just wanted to make sure this is what you wanted.

thanks there. looks good to me unless you want to remove it because of the 'wall'

core-code:
existing download seems to be current again now. lets re-visit this later on.

PR #77320, opened 2020-02-18
Remove spotify-notifications

gtmgianni:
Download URL no longer works.


vitorgalvao:
Thank you for the PR, but let’s wait a few more days [to see if it’s resolved](https://github.com/citruspi/Spotify-Notifications/issues/117). In the last 30 days, the URL was working.

felker:
Just came here to submit the same PR. I too noticed that the URL is down, but I am also not confident that the software works anymore on Mojave.

MTCoster:
Looks like the download link is working now

PR #77404, opened 2020-02-20
Update xamarin-mac

robertleeblairjr:

### DIFF
--- a/Casks/xamarin-mac.rb
+++ b/Casks/xamarin-mac.rb
@@ -1,6 +1,6 @@
 cask 'xamarin-mac' do
-  version '6.8.1.17'
-  sha256 '06b714e24237dd29e9bc61a557fa236f9f6a3885808794aaed211a2bbb97f10e'
+  version '6.14.1.30'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Cask could benefit from an update as well. It’s outdated by two minor versions.

robertleeblairjr:
> Cask could benefit from an update as well. It’s outdated by two minor versions.

Sorry, for the late reply. Thanks. At the time of the PR, I had attempted to update the version to the current stable 6.10 but the link is broken. The current 6.8 was preview release and I could update to the current 6.14 preview release. However, this is still a preview release channel branch. But, it wouldn't be any different than this current referenced version being preview release and would be a newer version.

robertleeblairjr:
[https://docs.microsoft.com/en-us/xamarin/mac/release-notes/6/6.14](url)

Appcast fails audit for version 6.14 doesn't exist but that's incorrect.

robertleeblairjr:
Xamarin 6.x requires Xcode 11.x. Travis is failing because of building with Xcode 10.x.

PR #77543, opened 2020-02-22
Update xamarin-ios from 13.8.1.17 to 13.14.1.30

robertleeblairjr:

### DIFF
--- a/Casks/xamarin-ios.rb
+++ b/Casks/xamarin-ios.rb
@@ -1,6 +1,6 @@
 cask 'xamarin-ios' do
-  version '13.8.1.17'
-  sha256 'a6e67a3d79e61553c3907f11ef88a355bf2af6d7a3fcb82ce66317db623fcad8'
+  version '13.14.1.30'
+  sha256 'bfe310c83b844e77f3c02e767cd31e8fd3660bdb8c259d9c3af8b1eb8044a884'
 
   url "https://dl.xamarin.com/MonoTouch/Mac/xamarin.ios-#{version}.pkg"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Page shows 13.10.

robertleeblairjr:
> Page shows 13.10.

Yep. I had tried to update with stable version but download links were broken. This current cask version is preview release channel, as well. Funnily, it seems like the stable versions download links are quite hidden and are only offered in the VS studio app release. I can look for stable channel versions. But, since the preview release versions in the cask were quite old, then I just went with it's merely updating a preview channel version with a newer preview channel version.

vitorgalvao:
> Funnily, it seems like the stable versions download links are quite hidden and are only offered in the VS studio app release.

How is that process? Does the app bundle those versions? Do they link you to a page to download? Our goal here should be to figure out a working system.

robertleeblairjr:
> > Funnily, it seems like the stable versions download links are quite hidden and are only offered in the VS studio app release.
> 
> How is that process? Does the app bundle those versions? Do they link you to a page to download? Our goal here should be to figure out a working system.

This like the other Xamarin software bundles are downloaded from Microsoft's servers for the update. Visual Studio downloads the updates and I'll try to capture the servers its pulling from to see if they're the same for all Xamarin. It seems like MS doesn't update the webpage to reflect updates but I'll look into something more uniform across the Xamarin builds to submit a PR. I agree it's a little bit all over the place with versions/channels and I realize these should be stable channel. Thanks for replying.

robertleeblairjr:
Have some free time and will look for the Xamarin stable channel versions to update this one from 13.8 preview release.

PR #77544, opened 2020-02-22
Update xamarin-profiler from 1.6.12-26 to 1.6.13-11

robertleeblairjr:

### DIFF
--- a/Casks/xamarin-profiler.rb
+++ b/Casks/xamarin-profiler.rb
@@ -1,6 +1,6 @@
 cask 'xamarin-profiler' do
-  version '1.6.12-26'
-  sha256 '8968ee2cd555efcd9058ee313683d09c731aec5bf6c0e01b65020fe4917b0c87'
+  version '1.6.13-11'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Page still shows version in cask.

robertleeblairjr:
> Page still shows version in cask.

[https://docs.microsoft.com/en-us/xamarin/tools/profiler/release-notes/1.6/1.6.13](url)
I pulled it from the release notes and download page. This 1.6.13-11 was released in Oct. 2019. So, I assumed it was current and Microsoft team overlooked ever updating to reflect the new version, which is common.

vitorgalvao:
> I pulled it from the release notes and download page.

Can you be more specific on the steps?

I’m fine with updating if we can get a better appcast, but otherwise I’m wary of changing it for a version they don’t openly advertise on their page.

> So, I assumed it was current and Microsoft team overlooked ever updating to reflect the new version, which is common.

That might be the case, but that should be brought up with them. In either case I don’t want to assume they forget, I want to be certain they don’t have a reason to not publish those releases.

PR #77610, opened 2020-02-24
Update maintenance.rb from 2.6.4 to 2.6.5 

suschizu:
update maintenance


suschizu:
`appcast at URL 'https://www.titanium-software.fr/en/release_maintenance.html' does not contain the version number '2.5.6':`
travis looks for the mojave version

PR #77650, opened 2020-02-25
update the cause of permission_denied_error

hawkzhou3:
I solved this problem with `chown`, and I see that other people on the Internet have encountered the same problem and solved it in this way.


vitorgalvao:
Thank you for the report. Your instructions required fixes, so making them in https://github.com/Homebrew/homebrew-cask/pull/77917.

PR #77667, opened 2020-02-25
Update retro-virtual-machine from 1.1.8 to 2.0.beta-1.r7,6783

core-code:


core-code:
yes, "beta". this might be an exception to our rule - this is the version they primarily list in their download section and the old version isn't even online anymore

vitorgalvao:
Not an exception; [we consider this case in the documentation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Closing in favour of https://github.com/Homebrew/homebrew-cask/pull/77690, which I saw and fixed first.

PR #77754, opened 2020-02-26
Delete appcast in recents.rb

suschizu:
the appcast doesn't work anymore but the download  still works - I can't find another appcast
the download from the website would be:   'https://recentsapp.com/releases/Recents_latest.zip' 
should we use this one instead and change the version staza to :latest ?

### DIFF
--- a/Casks/recents.rb
+++ b/Casks/recents.rb
@@ -4,7 +4,6 @@ cask 'recents' do
 
   # rink.hockeyapp.net/api/2/apps/74f5ee9ebf2d4be3b92a3e8766433b8b was verified as official when first introduced to the cask
   url "https://rink.hockeyapp.net/api/2/apps/74f5ee9ebf2d4be3b92a3e8766433b8b/app_versions/#{version.after_comma}?format=zip"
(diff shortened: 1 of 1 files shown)


cixtor:
“Recents_latest.zip” still points to the ~old~ current appcast, so I wouldn’t change it unless they release a new version with a different Feed URL. If you want to confirm yourself, you can download the [Zip archive you mentioned in your post](https://recentsapp.com/releases/Recents_latest.zip), extract, and execute the following command:

```
plutil -p Recents.app/Contents/Info.plist | grep "SUFeedURL"
```

vitorgalvao:
> If you want to confirm yourself, you can download the [Zip archive you mentioned in your post](https://recentsapp.com/releases/Recents_latest.zip), extract, and execute the following command:
> 
> ```
> plutil -p Recents.app/Contents/Info.plist | grep "SUFeedURL"
> ```

@suschizu has more experience finding appcasts than most users, and with a little work could give us tons (and I do mean a truckload) of examples where that method isn’t sufficient. Developers leave invalid or outdated feeds in `SUFeedURL` _all the time_. It’s not even funny how poorly maintained a lot of feeds are.

suschizu:
@vitorgalvao thanks :)

PR #77755, opened 2020-02-26
Update sococo.rb from 6.5.8,12-09-2019 to :latest / delete appcast

suschizu:
the latest version is 6.8.0 - I can't find a version information anywhere - just found this general downloadlink


adidalal:
Looks ike they switched to an Electron app, and block direct requests to `https://s.sococo.com/rs/client/update.json`, so this is the only option.

PR #77819, opened 2020-02-27
Update libreoffice from 6.4.0 to 6.4.1

core-code:


core-code:
the previous appcast has effectively been prevent us from finding updates since it will - by definition - only list the version that is already in the cask but not the newest version

PR #78082, opened 2020-03-04
Update time-sink from 2.0.1 to 2.1.0

suschizu:


suschizu:
`appcast at URL 'https://manytricks.com/timesink/appcast' does not contain the version number '2.1.0'`
true - the trailing zero is just contained in the download url - I think this is a special case so I don't think its necessary to add a `configuration: version. …`

PR #78230, opened 2020-03-06
Add SHA256 sum to keyboard-cleaner 1.2 cask

nkcmr:
also use https for download

methodology:
```
$ wget http://jan.prima.de/u/Keyboard-Cleaner-64bit.zip
...
$ sha256sum Keyboard-Cleaner-64bit.zip
75af7b4126b75e727045f4d9533b8d52a0ce4f5dc1610280f5bfc6b09814fa48  Keyboard-Cleaner-64bit.zip
```

### DIFF
--- a/Casks/keyboard-cleaner.rb
+++ b/Casks/keyboard-cleaner.rb
@@ -1,10 +1,10 @@
 cask 'keyboard-cleaner' do
-  version :latest
-  sha256 :no_check
+  version '1.2'
+  sha256 '75af7b4126b75e727045f4d9533b8d52a0ce4f5dc1610280f5bfc6b09814fa48'
 
-  url 'http://jan.prima.de/u/Keyboard-Cleaner-64bit.zip'
+  url 'https://jan.prima.de/u/Keyboard-Cleaner-64bit.zip'
   name 'Keyboard Cleaner'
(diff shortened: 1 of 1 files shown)


core-code:
unless we have a reliable appcast, chaning to a 'versioned' cask may not be what we want:

https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md

vitorgalvao:
> unless we have a reliable appcast, changing to a 'versioned' cask may not be what we want:

This is correct. To clarify, @nkcmr, if we don’t have a versioned URL, the cask will break with a shasum error every time upstream updates the software. However, with an appcast we can check for updates more easily/often, which mitigates the issue.

That said, I’ve noticed this on the page:

image

Six updates in fourteen years. That’s less than one update over more than each two years, so in this case we should be safe. In fact in this case we might be *better* off, because if the cask breaks we know it’s time for an update.

Let’s try it.

core-code:
agreed there.

btw, slightly offtopic: i 'brew fetch' *most* casks to find broken downloads or outdated checksums roughly every two weeks. this is quite wasteful on bandwidth though, so i was thinking that some HTTP headers like modification date or filesize could be used optimize this ...

PR #78291, opened 2020-03-07
alternative name ›0ad‹

zeilen:
add alternative name ›0ad‹


vitorgalvao:
Thank you, but `brew search 0ad` already finds the cask, so there’s nothing to gain by adding that `name`.

PR #78318, opened 2020-03-08
Update simply-fortran.rb from 3.9.3177 to 3.10.3203 (moj) and 3.10.3202 (cat)

suschizu:
update simply-fortran

### DIFF
--- a/Casks/simply-fortran.rb
+++ b/Casks/simply-fortran.rb
@@ -1,11 +1,11 @@
 cask 'simply-fortran' do
-  version '3.9.3177'
-
   if MacOS.version <= :mojave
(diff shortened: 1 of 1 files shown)


Casks/simply-fortran.rb
```diff
@@ -1,11 +1,11 @@
 cask 'simply-fortran' do
-  version '3.9.3177'
-
   if MacOS.version <= :mojave
-    sha256 'f21bbc26d2dc9d5e1a9bf427aadf2d9adc28a02ce5a4f51142879d32578d263b'
+    version '3.10.3203'
```
vitorgalvao: Shouldn’t this be reversed? The newest version is on the oldest OS.

ArmstrongJ: Build 3203 contains some legacy-specific fixes that weren't necessary for catalina, which is why it is using 3.10.3202.

Issue #78337, opened 2020-03-08
Global prefpane uninstallation


qaisjp:
#### General troubleshooting steps

- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [x] I have retried my command with `--force`.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

#### Description of issue

If a prefpane was installed for all users — if the prefpane is inside `/Library/PreferencePanes/` instead of `~/Library/PreferencePanes/` — then the `prefpane` stanza does not do anything.

https://github.com/Homebrew/homebrew-cask/blob/202d6019f83bbcea29ca04b41e17a3db515b0f59/lib/hbc/locations.rb#L31-L33

The above snippet needs support for `/Library/PreferencePanes`.

vitorgalvao:
the `prefpane` stanza does not install globally, so it has no obligation to check for whatever the user did with a manual install. If a user uses a custom install method, they’re expected to know what they’re doing and how to undo it.

qaisjp:
Okay. Thanks!

PR #78340, opened 2020-03-08
Uninstall witch better

qaisjp:

So this only fixes uninstallations when you select 'Install for this user only' during (manual) installation:

image

I've created #78337 to log future support for global prefpane removal.

Also https://manytricks.com/osticket/kb/faq.php?id=28 says that the user needs to restart System Preferences (to get rid of the pane) and this is correct. Should `brew cask uninstall` be doing this?

### DIFF
--- a/Casks/witch.rb
+++ b/Casks/witch.rb
@@ -11,5 +11,11 @@ cask 'witch' do
 
   prefpane 'Witch.prefPane'
 
-  zap trash: '~/Library/Preferences/com.manytricks.Witch.plist'
+  uninstall quit:       'com.manytricks.witchdaemon',
+            login_item: 'witchdaemon'
+
+  zap trash: [
+               '~/Library/Preferences/com.manytricks.Witch.plist',
+               '~/Library/Application Support/Witch',
+             ]
 end


Casks/witch.rb
```diff
@@ -11,5 +11,11 @@
 
   prefpane 'Witch.prefPane'
 
-  zap trash: '~/Library/Preferences/com.manytricks.Witch.plist'
+  uninstall quit:       'com.manytricks.witchdaemon',
+            login_item: 'witchdaemon'
+
+  zap trash: [
+    '~/Library/Preferences/com.manytricks.Witch.plist',
+    '~/Library/Application\ Support/Witch'
```
vitorgalvao: The `\` is literal.

vitorgalvao (changes requested):
Please run the commands in the template before checking them. There are style errors.

vitorgalvao:
> So this only fixes uninstallations when you select 'Install for this user only' during (manual) installation

We’re not concerned with manual installation. If a user installed in a way we’re not expecting, outside of HBC, it’s unreasonable to expect us to support that.

> says that the user needs to restart System Preferences (to get rid of the pane) and this is correct. Should `brew cask uninstall` be doing this?

Depends on what exactly this would be achieved.

Casks/witch.rb
qaisjp: My bad. I will remove the backslash.

Please also read the below screenshot. I did run the style checking command.

image

qaisjp:
> We’re not concerned with manual installation. If a user installed in a way we’re not expecting, outside of HBC, it’s unreasonable to expect us to support that.

Makes sense.

> Depends on what exactly this would be achieved.

The pane is only actually removed after a restart (it seems). Or if the active pane is the witch prefpane, then buttons on that pane will crash System Preferences.

Casks/witch.rb
vitorgalvao: > The style checking command does not catch `\`.

Because that’s not a stylistic issue. The backslash is valid, just wrong.

vitorgalvao:
> The pane is only actually removed after a restart

I understood what needs to be done, the question was *how* it would be done. If it’s a complex solution that requires AppleScript, for example, it’s not worth it. But if we can do it with a simple command like with the QuickLook server, we might consider it.

Issue #74132, opened 2019-12-12
Casks with broken homepages


vitorgalvao:
These are casks that, one way or another, have a broken or inaccessible homepage that needs to be fixed. This is a good, low-barrier issue for contributors.

homebrew-cask:

- [x] [airunlock](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/airunlock.rb)
- [x] [alcatraz](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/alcatraz.rb)
- [x] [arrayfire](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/arrayfire.rb)
- [x] [autofirma](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/autofirma.rb)
- [x] [avogadro](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/avogadro.rb)
- [x] [blueservice](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/blueservice.rb)
- [x] [browser-chooserx](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/browser-chooserx.rb)
- [x] [calcservice](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/calcservice.rb)
- [x] [captur](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/captur.rb)
- [x] [cathode](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/cathode.rb)
- [x] [celestia](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/celestia.rb)
- [x] [chameleon-ssd-optimizer](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/chameleon-ssd-optimizer.rb)
- [x] [cliqz](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/cliqz.rb)
- [x] [colorpicker-skalacolor](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/colorpicker-skalacolor.rb)
- [x] [coyim](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/coyim.rb)
- [x] [cyberghost-vpn](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/cyberghost-vpn.rb)
- [x] [data-integration](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/data-integration.rb)
[...]

PR #78588, opened 2020-03-13
Delete cathode.rb

vitorgalvao:
Website has been dead for months. Can’t verify is download is still valid.

https://github.com/Homebrew/homebrew-cask/issues/74132


suschizu:
I think it's just available inside the App Store: https://apps.apple.com/us/app/cathode/id499233976?mt=12&ign-mpt=uo%3D4

PR #78618, opened 2020-03-13
Update sherlock

core-code:

### DIFF
--- a/Casks/sherlock.rb
+++ b/Casks/sherlock.rb
@@ -2,8 +2,8 @@ cask 'sherlock' do
   version '1.7.4'
   sha256 '2eb61363bbd22ccbbf93f188a33fcc897a16311b9112b84563c7217bf6cbd958'
 
-  # rink.hockeyapp.net/api was verified as official when first introduced to the cask
-  url 'https://rink.hockeyapp.net/api/2/apps/d2096a451f9849fd84ddb59fc0a30efe/app_versions/2?format=zip'
+  # appcenter-filemanagement-distrib2ede6f06e.azureedge.net was verified as official when first introduced to the cask
+  url 'https://appcenter-filemanagement-distrib2ede6f06e.azureedge.net/10761dd5-e291-4944-a4f6-40411ad796c9/Sherlock_1.7.4.dmg?sv=2018-03-28&sr=c&sig=jvpVUPPMTbmSRYGgl0JNAL0mG6TLn5BmKGxTMyz8nNY%3D&se=2020-03-14T11%3A05%3A24Z&sp=r'
   appcast 'http://sparkle.sherlock.inspiredcode.io'
   name 'Sherlock'
   homepage 'https://sherlock.inspiredcode.io/'


core-code:
what shall we do with all the appcenter tokens?

i believe it is easiest and fastest just using the URL from the feed as is, both when done manually and also when done automatically

vitorgalvao:
These are going to be a major headache. Can we even trust them to work consistently? There are at least two betas in [homebrew/cask-versions](https://github.com/Homebrew/homebrew-cask-versions) that we had to remove because the URLs broke in less than a day.

core-code:
i am also concerned about the tokens, as i had to update 'marginnote' just to replace the suffix token

if these downloads really expire after a few days i only see 3 options:
• change the affected casks to a (static) download from the homepage instead of the feed version if available
• remove the affected casks
• or add some sparkle support to HBC to that we can omit the URL and it will fetch things straight from the appcast

vitorgalvao:
> change the affected casks to a (static) download from the homepage instead of the feed version if available

Best option.

> remove the affected casks

When the first option doesn’t work, and if the cask is relatively unpopular, that’s the next best thing.

> add some sparkle support to HBC to that we can omit the URL and it will fetch things straight from the appcast

I’d prefer to avoid adding support to any specific framework, [especially now](https://github.com/sparkle-project/Sparkle/issues/1556). We can already kind of do this agnostically with [`url do`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#using-a-block-to-defer-code-execution), but that means we also lose cask versioning so it’s not an ideal solution.

core-code:
> Best option.

ok, i'll try to keep it in mind if the token problem resurfaces

> I’d prefer to avoid adding support to any specific framework, especially now.
> We can already kind of do this agnostically with url do, but that means we also lose cask versioning so it’s not an ideal solution.
 
well, i can see arguments both sides. just for the sake of argument:
• direct support for sparkle would mean HBC could support more than 3000 apps with minimal overhead. no more sending and merging pull requests for each and every new version. no more broken downloads for users due to casks that haven't been updated in time. i think it would work better for end-users, with less maintenance work for contributors and maintainers
• sparkle won't go away even if they change maintainers. they already did that before. its used by thousands of apps and won't disappear anytime soon
• it wouldn't have to be 'any one' blessed appcast-mechanism. with a pluggable parser it would be possible to support electron/squirrel feeds too, adding 500-1000 more apps
• loosing support for versioning is not technically required as those too could be read from the appcast

in any case, it would mean a substantial amount of work. for me it would make HBC quite useless anyway, so of course i am strongly opposed to this idea ;p

vitorgalvao:
> sparkle won't go away even if they change maintainers.

My concern was that with a new maintainer coming in the middle of work for a new major version, we’d be working with a changing implementation.

> loosing support for versioning is not technically required as those too could be read from the appcast

But that would require major changes to allow everything in a cask to be made dynamic. For example, when artifacts have the version interpolated they’d need a network call just to show it. And we’d need to decide what to do in cases like the network being down or an appcast being unavailable (what version do we show?).

> in any case, it would mean a substantial amount of work.

One of my concerns as well. This wouldn’t be a major issue in a world where we could count on some developer consistency, but as you’re well aware there are tons of edge-cases we’d have to (think how to) work around.

Right now I think there are smaller things that need to work well first.

> for me it would make HBC quite useless anyway, so of course i am strongly opposed to this idea ;p

I also thought of that!

core-code:
> My concern was that with a new maintainer coming in the middle of work for a new major version, we’d be working with a changing implementation.

i wouldn't advocate using anything from their implementation, just parsing the feeds. those have had a pretty stable file format over the past 10 years and AFAIK their 2.0 branch uses the exact same RSS format

> But that would require major changes to allow everything in a cask to be made dynamic. For example, when artifacts have the version interpolated they’d need a network call just to show it. > And we’d need to decide what to do in cases like the network being down or an appcast being unavailable (what version do we show?).
> One of my concerns as well. This wouldn’t be a major issue in a world where we could count on some developer consistency, but as you’re well aware there are tons of edge-cases we’d have to (think how to) work around.

agreed on all points.

PR #78803, opened 2020-03-17
Replace AltDeploy homepage with the official one

pixelomer:

The commit message does not include the Cask version because I wasn't aware that it was necessary. Only the homepage is changed though, so I don't think the version is necessary.

The person who added this cask specified an unofficial as the homepage. As the developer of this utility (AltDeploy), I want to replace it with the official one.

### DIFF
--- a/Casks/altdeploy.rb
+++ b/Casks/altdeploy.rb
@@ -2,11 +2,10 @@ cask 'altdeploy' do
   version '1.1'
   sha256 'c9f1c27e3c83022fe0fae7abe3ea825792bc799390d3fb5492f8cfb97a9d7f56'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you for the fix.

I will, however, ask that you include some information about this app on the README for the repo. We generally do not accept apps with no description on the homepage, and that may be a factor in removing this cask when it comes time to review it again.

That’s to say you don’t need to do it today, but it should be on your mid-term list if you wish to keep the cask.

pixelomer:
@vitorgalvao I added a [README.md](https://github.com/pixelomer/AltDeploy/blob/master/README.md) file to the project. Is this enough?

vitorgalvao:
> Is this enough?

Yes. Thank you.

PR #78804, opened 2020-03-17
Update megasync from 4.3.0.8 to 4.3.1.0

ran-dall:

### DIFF
--- a/Casks/megasync.rb
+++ b/Casks/megasync.rb
@@ -1,5 +1,5 @@
 cask 'megasync' do
-  version '4.3.0.8'
+  version '4.3.1.0'
   sha256 '137da3481d80c26b444a4c5c53ab6c7e9d4424e48d7c30e8fafabef11447aed8'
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Not the version they have on the app bundle, but it’s the one they have on the appcast. Seems like they might be making a mess. We’ll fix the shasum when the time comes.

core-code:
i'll send a bug report to them immediately

core-code:
hm there is no bug report to send because the download is obviously still at '4.3.0.8'

therefore i think updating the cask to the wrong version number is just plainly wrong. can we please not do that?

ran-dall:
@core-code I just checked, I do have a download for this update but you're right it's still `4.3.0.8`. So I'm thinking the ML is just massively confused; I'll delete all the ML's data on this cask for it to restart the analyzation, this might fix the problem.

core-code:
thanks there @randall ! meanwhile i'll send a PR to revert

PR #78897, opened 2020-03-18
Convert find-appcast script to ruby

vitorgalvao:
Updated the `find-appcast` script to use ruby; test several possibilities on `electron-builder` appcast; warn if an `app-update.yml` was found but we couldn’t parse it yet; remove devmate and hockeyapp (both dead services).

Note the script’s name is now `find-appcast`, not `find_appcast` (changed the underscore to an hyphen)

Pinging @core-code and @suschizu as you might be frequent users (are you, @ran-dall?).

@core-code The `https://#{bucket}.s3.amazon-aws.yml/#{channel}/latest-mac.yml` URL seems weird to me (`s3.amazon-aws.yml`) and I can’t find it in casks. Is that one you’re aware of?


core-code:
thanks for working on this and for the heads-up. i've never seen the 's3.amazon-aws.yml' thing. which app does it occur with?

i have 1500 'app-update.yml' file contents to test with if you are interested

ran-dall:
@vitorgalvao YUP! Not often, but I do. Thanks for keeping me in the loop. 👏

vitorgalvao:
> which app does it occur with?

None that I can find. Maybe I’ll remove it in a later PR. Do you remember, @amyspark?

> i have 1500 'app-update.yml' file contents to test with if you are interested

If they also have the correct accompanying appcast, it would be useful to gather all patterns.

core-code:
> > i have 1500 'app-update.yml' file contents to test with if you are interested

> If they also have the correct accompanying appcast, it would be useful to gather all patterns.

no, its just the contents of the file on-disk ;(

vitorgalvao:
Let’s add them as they come, then, and release this version so it can have real-world, testing.

core-code:
i believe there is a regression in the new script related to discovering Sparkle feeds. it doesn't work with this one: https://github.com/headkaze/Hackintool/releases/download/3.3.4/Hackintool.zip

although the feed URL  is there in the Info.plist

vitorgalvao:
The script did find the URL, but because the status code is a redirect rather than a success, it was failing. Rather than adding more exceptions or whatnot that may not cover every case, [I’m changing it to use `curl`.](https://github.com/Homebrew/homebrew-cask/pull/79585)

PR #78943, opened 2020-03-20
Family Tree Builder 8.2.0.8516 Does not work on Catalina (32-bit)

hamiltont:
See https://faq.myheritage.com/en/article/is-family-tree-builder-compatible-with-the-new-macos-catalina for:

> MyHeritage's Family Tree Builder is currently incompatible with macOS Catalina. We're working hard to make the necessary changes to Family Tree Builder to ensure its compatibility with macOS Catalina.
>
> We expect this work to be completed soon in 2020.

Please note - I'm not sure what `depends_on` is supposed to do for a MacOS requirement - but it still installs this cask on my system even after the patch. However, I see other Casks with the same `depends on catalina` requirement also continue to install (bonjour-browser), so I'm not quite sure what adding this dependency is supposed to do?

### DIFF
--- a/Casks/family-tree-builder.rb
+++ b/Casks/family-tree-builder.rb
@@ -8,5 +8,7 @@ cask 'family-tree-builder' do
   name 'Family Tree Builder'
   homepage 'https://www.myheritage.com/'
 
+  depends_on macos: '<= :catalina'
+
   app 'Family Tree Builder.app'
 end
--- a/Casks/family-tree-builder.rb
+++ b/Casks/family-tree-builder.rb
@@ -8,7 +8,7 @@ cask 'family-tree-builder' do
   name 'Family Tree Builder'
   homepage 'https://www.myheritage.com/'
 
-  depends_on macos: '<= :catalina'
+  depends_on macos: '< :catalina'
 
   app 'Family Tree Builder.app'
 end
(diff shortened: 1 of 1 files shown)


vitorgalvao:
It should be working. Could it be that our check only works for `>=` and not `<=`? I’d find that bizarre, but the former is the only case covered by the documentation. Are you able to confirm, @reitermarkus?

hamiltont:
> Could it be that our check only works for >= and not <=? 

Doubt it. `/usr/local/Homebrew/Library/Homebrew/requirements/macos_requirement.rb` looks correct. 

My guess is it's much simpler - I see a few formula's using `<= catalina` which means they work if version == catalina. Probably need ot update those to `<= mojave`

Will update this PR soon and see if that fixes it

hamiltont:
Works as expected with `<` operator

vitorgalvao:
> Probably need to update those to `<= mojave`

Oh, of course!

Mind doing the other ones you found too?

hamiltont:
> Mind doing the other ones you found too?

Done. Per a quick grep it was only that one, probably just a typo that I unfortunately copied

vitorgalvao:
> Done (in second PR). Per a quick grep it was only that one, probably just a typo that I unfortunately copied

Thank you.

PR #79077, opened 2020-03-22
Fix Parsec.app installation on case-sensitive fs

gttotev:
The Parsec.app bundle contains a 'Macos' directory holding its binary.
This prevents it from opening on case-sensitive filesystems which expect
the binary in a 'MacOS' directory. Now after the package installation a
check for case-sensitivity will be performed and if it succeeds a
symbolic link 'MacOS' directory will be created in the installed bundle
pointing to the 'Macos' directory.

### DIFF
--- a/Casks/parsec.rb
+++ b/Casks/parsec.rb
@@ -11,6 +11,8 @@ cask 'parsec' do
 
   postflight do
     set_ownership '~/.parsec'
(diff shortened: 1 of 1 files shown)


gttotev:
Hmm, why is it saying it's still running? I tried adding it to `uninstall quit:` but same error, so I removed it.

vitorgalvao:
Thank you, but this is an upstream issue, we don’t want to hack around it. You should report this issue at the source so they can fix it for everyone.

PR #79102, opened 2020-03-22
Add ghdl v0.37

gpanders:

This PR re-adds `ghdl` to Homebrew Cask.

Even though this is an open-source CLI tool, I am submitting it as a cask instead of a "normal" Homebrew formula because building `ghdl` from source requires an Ada compiler, which is not currently supported by Homebrew and would be likely not be worth the trouble to add (see [this comment](https://github.com/Homebrew/homebrew-cask/pull/46031#issuecomment-388602940)). However, `ghdl` does supply pre-compiled binaries for macOS, so it is easy to simply provide a Cask formula to install binaries directly.

This new formula also addresses the issues raised [here](https://github.com/Homebrew/homebrew-cask/pull/46031#issuecomment-388602940) about installing `lib` and `include` files in a Cask formula. This formula does not install these files: if users need to link against `ghdl` in their own applications, they can add `/usr/local/Caskroom/ghdl/vX.Y/{include,lib}` to their include/library search paths.

### DIFF
--- a/Casks/ghdl.rb
+++ b/Casks/ghdl.rb
@@ -1,0 +1,11 @@
+cask 'ghdl' do
+  version '0.37'
+  sha256 'cb870085dd55167eda162b2d8b0020b30574b47fa8f8dfc7a1fd6aa3eb32ee91'
+
+  url "https://github.com/ghdl/ghdl/releases/download/v#{version}/ghdl-#{version}-macosx-mcode.tgz"
+  appcast 'https://github.com/ghdl/ghdl/releases.atom'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> a cask instead of a "normal" Homebrew formula
> (…)
> a Cask formula

It’s just “formula” and ”cask”. They have different names because they are different.

Thank you for the giving the reasoning with the submission. It’s a good PR.

Not totally convinced we should readd it, but I think it’s worth to do it and see how it plays out.

Thank you again.

PR #79137, opened 2020-03-23
Update stack from 2.4.1.70454 to 2.6.2.20200306

suschizu:

### DIFF
--- a/Casks/stack.rb
+++ b/Casks/stack.rb
@@ -1,6 +1,6 @@
 cask 'stack' do
-  version '2.4.1.70454'
-  sha256 '41daf141ffe51fe2c6357e6fe9a289303323c4b5421da89673878e0bfe475bed'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Looks like we might need to replace the appcast.

suschizu:
I can't find another appcast for this app - unfortunately 
should we delete the appcast here?

vitorgalvao:
> should we delete the appcast here?

Unsure. Maybe we should wait until the next update, as it might be a temporary blip. I tried looking for the download page to see if it might be usable, but I’m not finding it.

suschizu:
the download contains just -latest … here is the download site:
https://www.transip.nl/cp/stack/download/

vitorgalvao:
Merging for now, let’s revisit the appcast thing later.

PR #79266, opened 2020-03-25
Add Screen v1.0.16

dzhlobo:
https://screen.so

### DIFF
--- a/Casks/screen.rb
+++ b/Casks/screen.rb
@@ -1,0 +1,10 @@
+cask 'screen' do
+  version '1.0.16'
+  sha256 '9be3cb84f9e6dad6bb4a8b7a1afb2fac3723542fff4f659491a7cb33b58da759'
+
+  url "https://download.screen.so/desktop-app/darwin/#{version}/Screen.dmg"
+  name 'Screen'
+  homepage 'https://screen.so/'
+
+  app 'Screen.app'
+end


dzhlobo:
I don't understand the CI error 🤔.

Screen Shot 2020-03-25 at 13 36 09

jlucktay:
Looking at [the Travis output](https://travis-ci.com/github/Homebrew/homebrew-cask/jobs/302477356) it seems like `brew cask audit screen` on L322 is the issue, not `uninstall`.

dzhlobo:
I don't mind to rename it. Does it mean that we should never have such duplications between homebrew-cask and homebrew-core? Should we have the same warning when we run `brew cask audit screen` locally?

jlucktay:
> Does it mean that we should never have such duplications between homebrew-cask and homebrew-core?

That seems to be the point of this audit failure, yeah, to not end up with cores and casks having overlapping names.

> Should we have the same warning when we run `brew cask audit screen` locally?

Agree; it absolutely should, yes.

jgarber623:
On the Cask's name, prior art might be [the Zoom Cask](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/zoomus.rb) using the format `zoomus`.

I'd also hastily put together a Cask for this before seeing the PR:

```rb
cask 'screenso' do
  version '1.0.16'
  sha256 '9be3cb84f9e6dad6bb4a8b7a1afb2fac3723542fff4f659491a7cb33b58da759'

  url "https://download.screen.so/desktop-app/darwin/#{version}/Screen.dmg"
  name 'Screen'
  homepage 'https://screen.so'

  app 'Screen.app'

  zap trash: [
    '~/Library/Application Support/Screen',
    '~/Library/Preferences/so.screen.screen.app.plist'
  ]
end
```

vitorgalvao:
> I don't mind to rename it.

We do. Following the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) is paramount to avoid duplicates (huge problem in the past).

> Does it mean that we should never have such duplications between homebrew-cask and homebrew-core?

No.

> That seems to be the point of this audit failure, yeah, to not end up with cores and casks having overlapping names.

The point is that when a collision happens, we’ll be more alert and verify *if it’s the same software* because we don’t want to duplicate efforts. This policy is documented in [faq/rejected_casks](https://github.com/homebrew/homebrew-cask/blob/master/doc/faq/rejected_casks.md):

> The app is both open-source and CLI-only (i.e. it only uses the `binary` artifact). In that case, and [in the spirit of deduplication](https://github.com/Homebrew/homebrew-cask/issues/15603), (…)

There are plenty of name overlaps between casks and formulae.

> > Should we have the same warning when we run `brew cask audit screen` locally?
> 
> Agree; it absolutely should, yes.

There’s already a flag to do it, which CI uses. It’s not on by default precisely because we don’t want to encourage wrong renames or dissuade people from contributing.

> On the Cask's name, prior art might be [the Zoom Cask](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/zoomus.rb) using the format `zoomus`.

That’s because *the app bundle* installed by the package is called Zoom.us. That’s also covered in the token reference linked above.

If this were to be renamed due to a conflict with a cask, the correct way to go about it would be to prepped the vendor’s name to the token (again, covered in the reference).

@dzhlobo Since you’ve opened it first, do you want to add the enhancements in https://github.com/Homebrew/homebrew-cask/pull/79400/, or do you prefer I ask @joschi to rename theirs?

@core-code This a good reference: an uncommon case for which we already have rules in place.

vitorgalvao:
Closing in favour of https://github.com/Homebrew/homebrew-cask/pull/79400#pullrequestreview-383349986, which was already fixed.

dzhlobo:
Thank you everyone for quick responses. Glad it's in the repo now ❤️.

PR #79267, opened 2020-03-25
Update keepassxc.rb

slauber:
Updated to 2.5.3-2

After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [ ] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/Homebrew/homebrew-cask/pulls
[already refused]: https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues
[the correct repo]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask

PR #79282, opened 2020-03-25
keepassxc 2.5.3: emergency fix for macOS 10.15.4

maxim-belkin:

Open PR with incorrect URL: #79267

### DIFF
--- a/Casks/keepassxc.rb
+++ b/Casks/keepassxc.rb
@@ -1,9 +1,9 @@
 cask 'keepassxc' do
   version '2.5.3'
-  sha256 'fb9af9c2eab788e710f93165f063ed26281140266181d3904b4a6b2d62c5c9d8'
+  sha256 'e278f02466e3fb294c5c676d7b0ba3c60302d0b08273b8088ff39ca2e97e79af'
 
   # github.com/keepassxreboot/keepassxc was verified as official when first introduced to the cask
-  url "https://github.com/keepassxreboot/keepassxc/releases/download/#{version}/KeePassXC-#{version}.dmg"
+  url "https://github.com/keepassxreboot/keepassxc/releases/download/#{version}/KeePassXC-#{version}-2.dmg"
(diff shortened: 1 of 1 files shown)


skratchdot:
why not just set the version to `version '2.5.3-2'` (instead of adding the `-2` to the download url?

skratchdot:
this fix worked for me locally. thanks!

Casks/keepassxc.rb
```diff
@@ -1,9 +1,9 @@
 cask 'keepassxc' do
   version '2.5.3'
```
schwedenmut: what about changing this to 2.5.3-2 and leaving the url in Line 6 as it was before?

maxim-belkin: `url` has the following part `/#{version}/KeePassXC-#{version}-2.dmg"`, and only the second "version" has to change.

schwedenmut: just found out the "hard way" - thanks for the quick reply 👍

vitorgalvao:
@maxim-belkin Is this an “emergency fix” in the sense that it solves a critical security issue? If so, feel free to ping me in those cases; I could’ve looked at and merged this earlier in the day.

alaz:
Since the version stays the same, `brew cask upgrade` will not upgrade the package automatically, right? At least it does not for me.

Casks/keepassxc.rb
alaz: Since the version stays the same, `brew cask upgrade` will not upgrade the package automatically, right? At least it does not for me.

kesara: > Since the version stays the same, `brew cask upgrade` will not upgrade the package automatically, right? At least it does not for me.

@alaz Yeah, same for me. So you can do `brew cask reinstall keepassxc`.

phoerious:
Not a security issue, but macOS decided to change the way they validate entitlements in signed applications, causing a crash at startup.

vitorgalvao:
Thank you for clarifying, @phoerious.

PR #79327, opened 2020-03-26
Update papers

### LINKED ISSUE #79073: Papers app now has an app installer
[The cask](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/papers.rb) needs to be updated accordingly. The app installer is in a zip at https://www.papersapp.com/download/.

vladimyr:
Download & run app installer.
This fixes #79073

### DIFF
--- a/Casks/papers.rb
+++ b/Casks/papers.rb
@@ -6,11 +6,7 @@ cask 'papers' do
   name 'ReadCube Papers'
   homepage 'https://www.readcube.com/home'
 
-  installer script: {
-                      executable:   'Papers Installer.app/Contents/MacOS/Papers Installer',
-                      sudo:         true,
-                      print_stderr: false,
-                    }
+  installer manual: 'Papers Installer.app'
 
   uninstall quit:   'com.ReadCube.Papers',
             delete: '/Applications/Papers.app'
--- a/Casks/papers.rb
+++ b/Casks/papers.rb
@@ -1,20 +1,27 @@
 cask 'papers' do
-  version '4.0.0'
-  sha256 '8c63a193ced25c92db168e415b3a537efb828c86bd62b7615bc95508d3532801'
+  version :latest
+  sha256 :no_check
 
-  url 'https://download.readcube.com/app/Install%20Papers.pkg'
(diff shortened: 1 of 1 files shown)


Casks/papers.rb
```diff
@@ -1,20 +1,27 @@
...
-  appcast 'https://s3.amazonaws.com/update.readcube.com/desktop/updates/latest-mac.yml'
+  url 'https://download.readcube.com/app/Install%20Papers.zip'
   name 'ReadCube Papers'
   homepage 'https://www.readcube.com/home'
 
-  pkg 'Install Papers.pkg'
+  installer script: {
+                      executable:   'Papers Installer.app/Contents/MacOS/Papers Installer',
+                      sudo:         true,
+                      print_stderr: false,
```
vladimyr: It seems like there is no way to run it in unattended mode. I've checked binary using hopper and haven't found any traces of it 😞

`print_stderr` is required due to the following being written to stderr by the installer:
```
2020-03-26 04:07:57.781 Papers Installer[41815:16530467] P4Installer -> <NSAppleEventDescriptor: 'utxt'("ok")>
2020-03-26 04:07:57.986 Papers Installer[41815:16530467] “P4ScriptIsRunning.sh” couldn’t be removed.
```

vitorgalvao: > It seems like there is no way to run it in unattended mode.

Does this invoke a GUI? If so, use [`installer :manual`](https://github.com/homebrew/homebrew-cask/blob/c129758c8086d86b3d33cc8c2045b4a26bee2a83/doc/cask_language_reference/stanzas/installer.md#installer-manual) instead.

vladimyr: Thank you for the pointer 👍 

It invokes GUI that does literally nothing except asking you to type in your credentials after which it runs a bunch of shell scripts in the background. I could leave it as is which pretty much mimics silent mode but it may break in the future if they decide to add actual interactive steps 🤔

vitorgalvao: > asking you to type in your credentials

Then it needs to be `installer_manual`.

vladimyr: On it 👨‍💻

PR #79392, opened 2020-03-27
Update obs

vladimyr:
Updating `zap` stanza.


vladimyr:
It fails because appcast does not contain macOS release version and that is an existing issue 🤷‍♂

PR #79415, opened 2020-03-27
Add depends on catalina to macvim.rb

suschizu:
min version is 10.15

### DIFF
--- a/Casks/macvim.rb
+++ b/Casks/macvim.rb
@@ -9,7 +9,7 @@ cask 'macvim' do
 
   auto_updates true
   conflicts_with formula: 'macvim'
-  depends_on macos :catalina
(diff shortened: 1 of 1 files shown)


core-code:
=>
`  depends_on macos: '>= :catalina'
`

ariananggada:
i think its run perfectly in mojave

core-code:
@ariananggada how can you launch the macvim 8.2 release on mojave? since the 10.15 requirement is embedded into the app it should refuse to launch on any earlier system

suschizu:
> i think its run perfectly in mojave

you are right - sorry this was my mistake  -> https://github.com/Homebrew/homebrew-cask/pull/79592

PR #79497, opened 2020-03-28
Update proxyman from 1.20.0 to 1.21.0

vladimyr:
Update `zap` stanza & uninstall privileged helper tool.

### DIFF
--- a/Casks/proxyman.rb
+++ b/Casks/proxyman.rb
@@ -1,6 +1,6 @@
 cask 'proxyman' do
-  version '1.20.0'
-  sha256 'd50cf91c1d0bba1f3ae8039cb189dd4572a880579c15df3f15c935890f92342a'
+  version '1.21.0'
+  sha256 '3ea55bb49b017f622ade32a8f252d3af9dba93e0d8ca5589c63a9f85039984b2'
 
   # github.com/ProxymanApp/Proxyman was verified as official when first introduced to the cask
   url "https://github.com/ProxymanApp/Proxyman/releases/download/#{version}/Proxyman_#{version}.dmg"
--- a/Casks/proxyman.rb
+++ b/Casks/proxyman.rb
@@ -13,7 +13,14 @@ cask 'proxyman' do
   app 'Proxyman.app'
 
   uninstall delete:    '/Library/PrivilegedHelperTools/com.proxyman.NSProxy.HelperTool',
-            launchctl: 'com.proxyman.NSProxy.HelperTool'
+            launchctl: 'com.proxyman.NSProxy.HelperTool',
+            script:    {
+                         executable:   '/usr/bin/security',
+                         args:         ['delete-certificate', '-c', 'Proxyman CA'],
+                         must_succeed: false,
+                         print_stderr: false,
+                         sudo:         true,
+                       }
 
   zap trash: [
                '~/Library/Application Support/com.proxyman',
--- a/Casks/proxyman.rb
+++ b/Casks/proxyman.rb
@@ -24,8 +24,9 @@ cask 'proxyman' do
     end
   end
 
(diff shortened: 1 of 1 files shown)


Casks/proxyman.rb
```diff
@@ -12,6 +12,17 @@
 
   app 'Proxyman.app'
 
+  uninstall_postflight do
+    Kernel.system_command '/usr/bin/security',
```
vladimyr: @vitorgalvao I had to use `Kernel.system_command` because `system_command` ignores `must_succed` option and certificate may not be installed: https://github.com/Homebrew/brew/blob/2.2.11/Library/Homebrew/cask/dsl/base.rb#L15-L17

Is this intentional or maybe a bug in `brew` itself?
Is it ok for me to use `Kernel` directly inside the cask?

vitorgalvao: > Is this intentional or maybe a bug in `brew` itself?

That’s a question for @reitermarkus.

But isn’t this change you’re doing something that should be handled by upstream?

vladimyr: I'm removing root cert after proxyman gets removed. It will be there only if the user installs it (to decrypt https traffic). What do you envision by _handled by upstream_?

vladimyr:
I've created another PR #79545 that incorporates proposed changes without _controversial_ root cert bit so they don't get blocked by this :bowtie:

Casks/proxyman.rb
vitorgalvao: My mistake, I read this as an *install* postflight.

vladimyr: I took inspiration for this change from https://github.com/Homebrew/homebrew-cask/blob/304c561/Casks/autofirma.rb but after thinking it through I'm not sure that this is a correct way to handle it. Removal of a certificate, that is deliberately installed by the user, should belong to the same group as removal of user-created files i.e. it should be part of `zap` stanza. Because there is no such thing as `zap` pre/postflight I'm required to use `script` key of `zap` stanza. That sounds both hacky and ugly and would probably require inlining it using heredoc. What do you think? 🤔

vitorgalvao: > Because there is no such thing as `zap` pre/postflight I'm required to use `script` key of `zap` stanza. That sounds both hacky and ugly and would probably require inlining it using heredoc.

Agreed that we should avoid that route.

> Removal of a certificate, that is deliberately installed by the user, should belong to the same group as removal of user-created files

Even if it was installed by the user, is there any reason they’d want it to remain after removing the software?

vladimyr: > Even if it was installed by the user, is there any reason they’d want it to remain after removing the software?

Actually there is. If you do reinstall it will first execute uninstall step and as a consequence, you'll lose certificate as part of reinstallation.

NghiaTranUIT: Totally agree that we should remove the Proxyman cert if they user uninstall the app. As the uninstallation also remove all caching state (NSUserDefault), so if the app is re-installed, it doesn't recognize the old certificate since Proxyman search the certificate by full name, which is "Proxyman + {Generated Date String}"

ran-dall:
@vladimyr @vitorgalvao #79545 got merged, should this be closed?

Casks/proxyman.rb
reitermarkus: It's intentional that `must_succeed` cannot be overriden. You could check the output of `security find-certificate -a` to see if it is installed.

vladimyr: > It's intentional that `must_succeed` cannot be overriden.

That is weird because it can be easily circumvented by using `uninstall` `script` with `must_succeed` set to false: https://github.com/Homebrew/homebrew-cask/pull/79497/commits/5809d82

vladimyr:
@vitorgalvao @ran-dall I think we are good here. I've applied @reitermarkus's advice and got it working 🎉This is ready for review/merge.

Issue #63913, opened 2019-05-27
Clarify which `uninstall` keys support wildcards or regex


vitorgalvao:
For example, from looking at [the `uninstall` summary of keys](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#summary-of-keys), you’d think wildcards weren’t supported in `delete`, [but they are](https://github.com/Homebrew/homebrew-cask-drivers/pull/941/files#r281294282).

We should check the code and update the documentation accordingly.

cc @reitermarkus.

MikeMcQuaid:
Closing this after a long time with no updates or progress.

PR #79620, opened 2020-03-30
Update sound-control: the uninstall `launchctl` key does not accept wildcards. Fixes this cask

varenc:

### DIFF
--- a/Casks/sound-control.rb
+++ b/Casks/sound-control.rb
@@ -12,6 +12,10 @@ cask 'sound-control' do
 
   app 'Sound Control.app'
 
-  uninstall launchctl: 'com.staticz.soundcontrol.*',
+  uninstall launchctl: [
+                         'com.staticz.soundsiphon.bridgedaemon',
+                         'com.staticz.audio.soundsiphon.playeragent',
+                         'com.static.soundsiphon.inputagent',
+                       ],
             quit:      'com.staticz.SoundControl'
 end


varenc:
[per the docs](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md), the uninstall `launchctl` key excepts an id or an array of ids and I don't think wildcards work.  Looking at the --debug logs seems to confirm this. 

This PR fixes this by enumerating the launchctl ids explicitly instead. 

I believe all of these casks make the same mistake:
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/avast-secureline-vpn.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/boom.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/box-drive.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/paragon-extfs.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/paragon-ntfs.rb

Unfortunately fixing all those requires a bit of individual work to find the launchctl ids for each one.  Alternatively, should `launchctl` be fixed so that it excepts wildcards like `pkgutil`? I don't think it should.  The explicit approach is safer.  In which case just adding a lint check to detect this occasional mistake could be wise.

vitorgalvao:
> the uninstall `launchctl` key excepts an id or an array of ids and I don't think wildcards work

Are you able to confirm, @reitermarkus? [Our documentation needs fixing in that regard](https://github.com/Homebrew/homebrew-cask/issues/63913).

> Alternatively, should `launchctl` be fixed so that it excepts wildcards like `pkgutil`? I don't think it should.

It definitely should.

> The explicit approach is safer.

`launchctl` takes bundle IDs; there’s little risk of catching something wrong. For the incorrect apps without a proper bundle ID, we can be explicit.

varenc:
So pretty sure they don't work as is. From `brew cask --debug uninstall`:

```
...
==> Removing launchctl service com.staticz.soundcontrol.*
/bin/launchctl list com.staticz.soundcontrol.\*
/usr/bin/sudo -E -- /bin/launchctl list com.staticz.soundcontrol.\*
```

So `launchctl list` is ran with the string directly and the `*` escaped.   I think when that command exits with a non-zero status homebrew just assumes that job isn't installed and moves along. And unlike pkgutil, the launchctl tool doesn't support regexes so even without the escaping `launchctl list` still won't work. To implement this this you'd have to list all the jobs (with sudo and without sudo to get everything), and then match your regex against the results.  

There's also two styles of regexs in use, the standard regex `.*` style ones used by `pkgutil:`, and the undocumented shell glob style `*` ones used by `delete:`. (Which I assume is implemented just as bash expansion?).  Since launchctl doesn't seem to have any native regex support, you'd have to choose what style to use.  Seems less dirty to use glob style and not reinterpret all the `.` characters in launchd jobs as "match any character" but that's probably fine in practice.  

Not that it matters much, but a few casks have `launchctl:` strings without bundle ids in them. That seems like the fault of the app developers though. [these](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/adobe-acrobat-pro.rb) [for](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/idrive.rb) [example].(https://github.com/Homebrew/homebrew-cask/blob/master/Casks/vyprvpn.rb#L17)

vitorgalvao:
> Not that it matters much, but a few casks have `launchctl:` strings without bundle ids in them.

Sure, but in those we can just not use the wildcard/regex (when/if it’s implemented).

Thank you for fixing this one. Even if you don’t want to take a look at the others, would you mind opening an issue tracking the ones you found, so somebody else can?

PR #79769, opened 2020-04-01
Update torguard from 3.98.2 to 3.98.4

suschizu:


suschizu:
`installer: Cannot install on volume / because it is disabled.
installer: The version of macOS is too new.`
this is new … too new

vitorgalvao:
Reverting, as it looks like this error seems to happen in several macOS versions, also on the manual install. Looks like a mistake on their side.

PR #79773, opened 2020-04-01
Add depends_on java6 to shinhan-ezplus

dahlia:
As the shinhan-ezplus depends on the legacy Java support, I added `depends_on cask: 'java6'` to the caskfile.


vitorgalvao:
That’s not the way to do it for java. It needs to be [a `caveat`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md#caveats-mini-dsl) instead.

dahlia:
@vitorgalvao Thanks! I replaced `depends_on` with `caveat`.

PR #79798, opened 2020-04-02
Update wechat from 2.3.31.17 to 2.3.31.18

huxuan:

### DIFF
--- a/Casks/wechat.rb
+++ b/Casks/wechat.rb
@@ -1,6 +1,6 @@
 cask 'wechat' do
-  version '2.3.31.17'
-  sha256 'd7f1d0b9f5b385b9fb3f63470ebb3e100b97279bd2b3bfbc9f926290423c2169'
+  version '2.3.31.18'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Duplicate.

huxuan:
Hi @vitorgalvao, may I have a small suggestion on the duplication?

I noticed that there are many duplicate pull requests with same changes, but maybe it is better the merge the earliest one rather than the latest one. Actually, every time before I send a pull request, I will double check whether there is already a similar one exists. We should not encourage this kind of duplication work.

vitorgalvao:
> maybe it is better the merge the earliest one rather than the latest one.

I agree. But Github’s Notifications and Pull Request screens don’t make that easy, by putting the latest first.

We get inundated with so many requests, they’re hard to parse. I merged the earliest one *I could see* and then found more in the rubble.

Sometimes contributors notice the duplicates and comment with a link back to the original. That helps merge the earliest one.

huxuan:
Hi @vitorgalvao, thanks for your reply. I totally understand and really appreciate your efforts to maintain the awesome homebrew cask.

PR #79886, opened 2020-04-02
Update teamwire from 2.0.2 to 2.1.0

igorkulman:


RoiEXLab:
Maybe it's just me, but it looks like the checksum does no longer match

igorkulman:
> Maybe it's just me, but it looks like the checksum does no longer match

You are right, fixing.

Issue #79661, opened 2020-03-31
`launchctl:` does not support wildcards, but a few casks use it anyway: avast-secureline-vpn, boom, box-drive, paragon-extfs, paragon-ntfs


varenc:
#### General troubleshooting steps

- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [x] I have retried my command with `--force`.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

#### Description of issue

The `lanchtctl:` key in the `uninstall` stanza does not support wildcards, but a few Casks attempt to use this anyway.  This ends up making it a no-op since that this string doesn't match any launchd job ids.

PR #79620 fixes this for the sound-control cask, but these casks all have the same problem:

- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/avast-secureline-vpn.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/boom.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/box-drive.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/paragon-extfs.rb
- https://github.com/Homebrew/homebrew-cask/blob/master/Casks/paragon-ntfs.rb

From `brew cask --debug uninstall` you can see that the `*` is escaped, but even if it wasn't, launchctl doesn't have any built in support for regexes so it wouldn't work in either case.  See #79620 for a little more discussion on this issue.

```
...
==> Removing launchctl service com.staticz.soundcontrol.*
/bin/launchctl list com.staticz.soundcontrol.\*
/usr/bin/sudo -E -- /bin/launchctl list com.staticz.soundcontrol.\*
```

To fix this you'll need to install each cask, identify the full names of the launchd jobs it installs, and then enumerate all the job ids in the cask. Or improve `launchctl:` so that it accepts wildcards.

vitorgalvao:
[...]

PR #79904, opened 2020-04-03
Update avast-secureline-vpn: remove wildcards from `launchctl:`

varenc:

### DIFF
--- a/Casks/avast-secureline-vpn.rb
+++ b/Casks/avast-secureline-vpn.rb
@@ -10,8 +10,17 @@ cask 'avast-secureline-vpn' do
 
   uninstall pkgutil:   'com.avast.secureline',
             launchctl: [
-                         'com.avast.secureline*',
-                         '*.com.avast.osx.secureline.avastsecurelinehelper',
+                         'com.avast.secureline.update',
+                         'com.avast.secureline.uninstall',
+                         'com.avast.secureline.service',
+                         'com.avast.secureline.init',
+                         'com.avast.secureline.burger',
+                         'com.avast.secureline.userinit',
+                         'com.avast.secureline.update-agent',
+                         # '6H4HRTU5E3' seems like a random string, but cursory investigation shows
+                         # this launchd job has the same prefix when installed in other environments.
+                         # Also multiple results for this in Google
+                         '6H4HRTU5E3.com.avast.osx.secureline.avastsecurelinehelper',
                          'com.avast.secureline.home.userinit',
                        ],
             delete:    '/Applications/AvastSecureLine.app'


varenc:
This enumerates the launchd job Avast installs since wildcards aren't working right now. (#79661).

This thing installs a lot of services. It might setup more if I actually use it, but that requires providing a credit card for a free trial.

The job `6H4HRTU5E3.com.avast.osx.secureline.avastsecurelinehelper` is an odd one.  Not sure where the id `6H4HRTU5E3` comes from but I saw it show up on two different machines and it's in a bunch of Google search results.  Though if it does vary between installs that's a very good reason for why `launchctl:` needs to support wildcards.

varenc:
Wow. After running `cask uninstall` with my new updated Cask, Avast detected that and run `/bin/bash /Library/Application Support/AvastSecureLine/hub/autouninstall.sh` as root (somehow??). Eventually the launchd jobs I enumerated where removed, but now three new ones were installed! `com.avast.hub`, `.xpc`, and `.schedule` Presumably by this autouninstall.sh script.

Not sure what the policy is, but it's acting a bit like malware, so maybe just remove the Cask.

varenc:
This reason for the build fail is odd:
```
Error: Some launch jobs were not unloaded, add them to uninstall launchctl:
6H4HRTU5E3.com.avast.osx.secureline.avastsecurelinehelper
com.avast.secureline.userinit
```

because those jobs are already in `uninstall launchctl:`. I suspect that the same build process for the cask currently on master would also fail in the same way?

Homebrew does run `launchctl remove <job>` and in my experience that also makes the process for the job quit, but Avast must have some unusual behavior that's preventing that.  Presumably this build passed before, but this Cask has no sha256 hash so my guess is this is new behavior. More points for 'acting a bit like malware'.

vitorgalvao:
Nice work!

I agree their behaviour is a bit shady, but I think we should do a bit more investigating beforehand.

varenc:
Out of curiosity I did a bit more poking around and logged all the avast related processes that start after homebrew's uninstall is triggered.

```
PID  PPID      USER     STARTED     COMMAND
33668 21239    chris     3:16PM    00:00 /bin/bash /usr/local/Homebrew/Library/Homebrew/brew.sh cask --debug uninstall avast-secureline-vpn
```
^^ this is me running cask uninstall

```
34035     1    root      3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/autouninstall.sh
```
^^ I think this is actually being triggered by launchd itself since the parent PID is 1. The job definition for `com.avast.secureline.uninstall` is set to keep alive no matter what as long as `/Applications/AvastSecureLine.app` still exists, so after that path stops existing, launchd startups up this job again?  This job is then responsible for all the user executions below.

```
34078 34035    root      3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/autouninstall.sh
34080 34078    root      3:16PM    00:00 grep /Library/Application Support/AvastSecureLine/components/update/update.sh
34554 34035    root      3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/uninstall.sh
34593 34554    root      3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/uninstall.sh
34596 34593    root      3:16PM    00:00 grep /Library/Application Support/AvastSecureLine/components/update/update.sh
34637 34554    root      3:16PM    00:01 nc -U /Library/Application Support/AvastSecureLine/run/update/502.sock
34638 23606    chris     3:16PM    00:01 /bin/bash /Library/Application Support/AvastSecureLine/hub/userinit.sh uninstall
34681 34638    chris     3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/userinit.sh uninstall
34736 34638    chris     3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/userinit.sh uninstall
34737 34736    chris     3:16PM    00:00 curl -s http://mac.sl.u.avcdn.net/macosx/avast/secureline/remoteconfig
```
^^^ This `curl` for file called `remoteconfig` file looks pretty sketchy at first...but it seems benign and is just HockeyApp related.  But even so, remote network calls triggered when a user attempts to remove software still pretty sketchy.

```
34883 34554    root      3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/modules/020_service.sh stop
34912 34554    root      3:16PM    00:00 /bin/bash /Library/Application Support/AvastSecureLine/hub/modules/010_burger.sh stop
34957 34554    root      3:16PM    00:01 /bin/bash /Library/Application Support/AvastSecureLine/hub/modules/020_service.sh uninstall
34959 34957    root      3:16PM    00:01 /bin/bash /Library/Application Support/AvastSecureLine/hub/modules/020_service.sh uninstall
34961 34959    root      3:16PM    00:01 grep M-VPN-AVAST
```

According to some Avast forums the only correct way to uninstall Avast is with the provided uninstaller and any attempts to do it manually will result in bad behavior.  Though the forums also claimed that AvastHUB will self-delete in a few hours after a manual uninstall...

PR #79926, opened 2020-04-03
Update qlvideo

vladimyr:
Fix `version`, update `zap` stanza.

### DIFF
--- a/Casks/qlvideo.rb
+++ b/Casks/qlvideo.rb
@@ -1,5 +1,5 @@
 cask 'qlvideo' do
-  version '193'
+  version '1.93'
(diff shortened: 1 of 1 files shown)


Casks/qlvideo.rb
```diff
@@ -1,5 +1,5 @@
 cask 'qlvideo' do
-  version '193'
+  version '1.9.3'
```
vladimyr, line 2: This is the correct version according to:
https://github.com/Marginal/QLVideo/blob/rel-193/app/Info.plist#L19-L22

vladimyr: Appcast uses `1.93` format so I changed it accordingly. What is preferred here, keeping parity with `CFBundleVersion` or appcast? 🤔

PR #79973, opened 2020-04-03
Update macupdater

vladimyr:
Update `zap` stanza.

### DIFF
--- a/Casks/macupdater.rb
+++ b/Casks/macupdater.rb
@@ -1,6 +1,6 @@
 cask 'macupdater' do
   version '1.5.4'
-  sha256 '5ec1063813bf2502e8a3c1f1ee0770bba3bbee1478cf680cff742dbb21e40a27'
+  sha256 '4ae6910fda61fae7ff5ba1a685eaa2114c908d44fa4b5234b7e838f0fcb6fd3f'
 
   url "https://www.corecode.io/downloads/macupdater_#{version}.dmg"
   appcast 'https://www.corecode.io/macupdater/macupdater.xml'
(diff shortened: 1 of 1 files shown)


core-code:
the SHA did change to 4ae6910fda61fae7ff5ba1a685eaa2114c908d44fa4b5234b7e838f0fcb6fd3f

vladimyr:
> the SHA did change to 4ae6910fda61fae7ff5ba1a685eaa2114c908d44fa4b5234b7e838f0fcb6fd3f

Thanks, updated. 👍

vladimyr:
@core-code Is there a way to quit MacUpdater before uninstalling? I've tried with `quit` stanza and sending both sigkill and sigterm but whatever I do it gets stuck on exit dialog.

core-code:
@vladimyr whether MacUpdater will pop up a dialog on quit depends on a number of factors. we try to both prevent accidental quits while also trying not to be annoying. if MacUpdater is set to be displayed in the dock and the preferences-key 'DidAlreadyAskUserAboutReallyQuittingOnce' is set, it will not put up the dialog. i guess the second condition would be easy to ensure in the uninstall-stanza but i am unsure about the first one. 

i cannot really reproduce (or explain) any failures to properly exit after a sigkill or sigterm however. we don't block the signals. i just did 'killall MacUpdater' and it exited as expected.

vitorgalvao:
> i cannot really reproduce (or explain) any failures to properly exit after a sigkill or sigterm however. we don't block the signals. i just did 'killall MacUpdater' and it exited as expected.

I never got the chance to look into it, but I suspect the CI script should add a delay before sending signals / commands and checking for remnants of the applications. I suspect that a few hardcoded seconds to get things the time to exit might fix some issues.

PR #79981, opened 2020-04-04
Update android-sdk for deprecation and conflict with android-clt

robertleeblairjr:

### DIFF
--- a/Casks/android-sdk.rb
+++ b/Casks/android-sdk.rb
@@ -5,7 +5,7 @@ cask 'android-sdk' do
   # dl.google.com/android/repository was verified as official when first introduced to the cask
   url "https://dl.google.com/android/repository/commandlinetools-mac-#{version}_latest.zip"
   name 'android-sdk'
-  homepage 'https://developer.android.com/index.html'
+  homepage 'https://developer.android.com/'
 
   binary "#{staged_path}/tools/android"
   binary "#{staged_path}/tools/bin/archquery"
--- a/Casks/android-sdk.rb
+++ b/Casks/android-sdk.rb
@@ -1,9 +1,9 @@
 cask 'android-sdk' do
-  version '4333796'
-  sha256 'ecb29358bc0f13d7c2fa0f9290135a5b608e38434aad9bf7067d0252c160853e'
+  version '6200805'
+  sha256 '23f0626336a98d70aff7311a73292026af31bc577c6f06b509cd4ad33752313e'
 
   # dl.google.com/android/repository was verified as official when first introduced to the cask
-  url "https://dl.google.com/android/repository/sdk-tools-darwin-#{version}.zip"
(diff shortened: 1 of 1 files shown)


robertleeblairjr:
This version of SDK Tools and the packaging have been deprecated. These have been replaced with newer versions and a change to the naming of the package which is referred to as Command Line Tools. However, the package still contains the 'sdkmanager' along with other binaries which were used in the SDK Tools package.

[https://developer.android.com/studio/releases/sdk-tools](url)

[https://developer.android.com/studio#cmdline-tools](url)

Casks/android-sdk.rb
```diff
@@ -1,9 +1,9 @@
...
-  version '4333796'
-  sha256 'ecb29358bc0f13d7c2fa0f9290135a5b608e38434aad9bf7067d0252c160853e'
+  version '6200805'
+  sha256 '23f0626336a98d70aff7311a73292026af31bc577c6f06b509cd4ad33752313e'
 
   # dl.google.com/android/repository was verified as official when first introduced to the cask
-  url "https://dl.google.com/android/repository/sdk-tools-darwin-#{version}.zip"
+  url "https://dl.google.com/android/repository/commandlinetools-mac-#{version}_latest.zip"
   name 'android-sdk'
   homepage 'https://developer.android.com/index.html'
```
vladimyr, line 8: It is a really cosmetic suggestion, but now it is a good time to do it:
```suggestion
  homepage 'https://developer.android.com/'
```

robertleeblairjr:
@vladimyr Thanks for the update. Feel free to provide any more. Currently, the binaries in the command line tools package aren't in the same directories or aren't used and removed in comparison to the outdated SDK tools package. In my free time I'll try to look through each and determine what was removed or needs a different symlink in this cask PR.

robertleeblairjr:
Looking through each archive, there may be a significant enough difference in the structures and binaries, that since this android-sdk tools was deprecated and replaced by command line tools that it could be easier to just create a new cask for the command line tools. Perhaps, moving sdk tools to homebrew-cask-versions tap since it's been deprecated. Unsure of how many casks use this cask as a dependency check though.

robertleeblairjr:
@vitorgalvao When you get the chance if you could check and determine if this is a more viable solution? Deprecating this cask, updating conflict_with, maybe moving 'android-sdk' to homebrew/homebrew-cask-versions, and approving the creation of the 'android-clt' pull #80305 request. Thanks.

ran-dall:
> Deprecating this cask, updating conflict_with, maybe moving 'android-sdk' to homebrew/homebrew-cask-versions, and approving the creation of the 'android-clt' pull #80305 request.

@robertleeblairjr I like this idea as well. Regardless let's wait till @vitorgalvao gives us his opinion.

vitorgalvao:
This has too many users to be changed to something else from under user’s feet. A better approach is to mark as discontinued first: https://github.com/Homebrew/homebrew-cask/pull/82014

PR #79992, opened 2020-04-04
Update zalo from 20.3.1a to 20.3.3

suschizu:


vitorgalvao:
Looks like a bad DMG.

suschizu:
> Looks like a bad DMG.

ok - so there is nothing we can do? 
… perhaps it will be fixed with the next update

PR #80035, opened 2020-04-05
Update menubar-colors

vladimyr:
Update appcast.


vitorgalvao:
When the download is from github releases, we prefer the appcast to be as well. Those are auto-generated, so won’t be outdated.

PR #80207, opened 2020-04-08
Update wechat to 2.3.31.19

TomorrowWu:

### DIFF
--- a/Casks/wechat.rb
+++ b/Casks/wechat.rb
@@ -1,6 +1,6 @@
 cask 'wechat' do
-  version '2.3.31.18'
-  sha256 '51bd35f9621cff431934003ba91aa4eaa03768962c90f1888ece54122b8947c3'
+  version '2.3.31.19'
+  sha256 '91dbbdfcb5b9eacdbe5b813e868f0bba4ed3435b85533dd0af0ba179ba480408'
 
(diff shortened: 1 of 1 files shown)


ran-dall:
@TomorrowWu Travis is reporting the `sha` to be `51bd35f9621cff431934003ba91aa4eaa03768962c90f1888ece54122b8947c3` which is the the `sha` for `2.3.31.18`. This makes me suspect that this update isn't valid or isn't availiable.

ran-dall:
@TomorrowWu @ElfSundae @taylorsfan I've downloaded the `.dmg`, checked the `appcast`, and checked the `.plist` of the app. I'm getting `2.3.31.18`. Can one of you show me where you got `2.3.31.19` from? 

```plist
<key>MMBundleVersion</key>
<string>2.3.31.18</string>
```

```xml
sparkle:version="13288" sparkle:shortVersionString="2.3.31.18"
```

@core-code @suschizu @vitorgalvao Have we encountered something like this before? The only thing I can think of would be that its CDN related.

core-code:
i think we've seen this before. it seems to consistently give different versions to users based on their location (?)

maybe we could work around the problem with a language stanza?

ran-dall:
> maybe we could work around the problem with a language stanza?

@core-code That seems like a fine idea.

core-code:
all seems to be fine again now but lets try that if the problem appears again

PR #80208, opened 2020-04-08
Update wpsoffice.rb to add zh_CN package version 2.1.2,3417

ctyytc:

### DIFF
--- a/Casks/wpsoffice.rb
+++ b/Casks/wpsoffice.rb
@@ -1,20 +1,45 @@
 cask 'wpsoffice' do
-  version '2.1.0,3464'
-  sha256 'f46ec300e8129c6c1972eee79b0e2208f5b93b7048d6fd43dca8b35752a09c65'
+  language 'en', default: true do
+    version '2.1.0,3464'
+    sha256 'f46ec300e8129c6c1972eee79b0e2208f5b93b7048d6fd43dca8b35752a09c65'
 
-  # wdl1.pcfg.cache.wpscdn.com was verified as official when first introduced to the cask
-  url "https://wdl1.pcfg.cache.wpscdn.com/wpsdl/macwpsoffice/download/#{version.before_comma}.#{version.after_comma}/WPSOffice_#{version.before_comma}(#{version.after_comma}).dmg"
-  name 'WPS Office'
-  homepage 'https://www.wps.com/mac/'
+    # wdl1.pcfg.cache.wpscdn.com was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


core-code:
thanks for your contribution. can you re-work this so only the version, sha, url and homepage stanzas are duplicated? everything else is the same anyway.

ctyytc:
> thanks for your contribution. can you re-work this so only the version, sha, url and homepage stanzas are duplicated? everything else is the same anyway.

Only "name, depends_on, app" are same, everything else is defferent. Still need to update?

core-code:
@vitorgalvao what do you think?

vitorgalvao:
Even the versions are different? We’ll need to have a discussion about chinese versions of software; they’re a major hassle and should maybe be removed and sent to another (unofficial) tap.

> Only "name, depends_on, app" are same, everything else is defferent. Still need to update?

Yes. `zap` can also be unified: use the `*` glob.

Remove the `'en_US'` and the other from the end, they’re not doing anything.

core-code:
> We’ll need to have a discussion about chinese versions of software; they’re a major hassle and should maybe be removed and sent to another (unofficial) tap.

i agree. all macOS apps should contain all languages in a unified download and app-bundle. i don't think we should invest much time to support apps from vendors that refuse to accept this and other basic macOS development rules. 
the only way how these vendors will fix their act is if enough of their users complain about the broken stuff.
however, if we spend much time to work around their problems this will only lead to them getting less complaints. this is the opposite from trying to improve the status quo.

vitorgalvao:
> all macOS apps should contain all languages in a unified download and app-bundle.

While I agree, in the case of Chinese versions it’s likely they need to provide different downloads due to The Great Firewall, and the software might also be different, modified in accordance to governmental regulations.

For things like PokerStars it’s due to regional gambling laws, though I see little reason as to why they need a different app per legislation. Accounts are already associated with real IDs, so they could impose the restrictions in the same app.

For apps like Firefox, I have no idea. But from conversations I had with their developers, they don’t seem to understand macOS development very well (despite most of them ostensibly using Macs). I tried talking to one of them *at a developer conference* about their lack of AppleScript support and he had no idea what it was (they’re the only major browser without support for it, hence the worst browser for automation), despite an open bug report so old that if it was a person it could vote. From other reports I’ve read, their Keychain support is also broken, as are other macOS features.

ctyytc:
@vitorgalvao @core-code Tks for reply, I fully agree that all macOS apps should contain all languages in a unified download and app-bundle. But unfortunately WPS is not yet able to do it for the time being, the Chinese version has a huge difference from the English version in terms of user login method, cloud storage form, privacy agreement, etc.

vitorgalvao:
@ctyytc It still needs the other requested modifications.

vitorgalvao:
Closing because this now has conflicts. @ctyytc If you’re still interested in getting this merged, please resubmit with the fixes.

PR #80234, opened 2020-04-08
Update coconutbattery

core-code:


core-code:
appcast failure is expected. i guess if they still don't update it in a week we should look for alternatives.

vladimyr:
> appcast failure is expected. i guess if they still don't update it in a week we should look for alternatives.

I believe they've updated appcast so this is ready to be merged 🚀

PR #80257, opened 2020-04-08
Update balsamiq-wireframes from 4.0.23 to 4.0.24

vkarampinis:

### DIFF
--- a/Casks/balsamiq-wireframes.rb
+++ b/Casks/balsamiq-wireframes.rb
@@ -1,6 +1,6 @@
 cask 'balsamiq-wireframes' do
-  version '4.0.23'
-  sha256 '931db241ae178202a96f9cea501869e73bcaf5433af24a3bdc8440805b54962e'
+  version '4.0.24'
+  sha256 '47211b671137170a6e8d401330eb02aeca47ebdcaf6934bce92e169992bac12c'
 
   url "https://builds.balsamiq.com/bwd/Balsamiq%20Wireframes%20#{version}.dmg"
   appcast 'https://builds.balsamiq.com/bwd/mac.jsonp'


ran-dall:
@vkarampinis I'm sorry but we're going to run into the same problem as yesterday if you keep doing this. Was something not clear yesterday? All the updates you've done in the last 20m are listed on those outputs that I indicated yesterday. 

Furthermore, the timing of these updates makes me think that something wasn't clear about not allowing scripted updates.

vkarampinis:
@ran-dall I manual confirmed each update I send.
Also, I checked the output files and only updated those marked as `no auto cask-repair: technically impossible`.

Am I doing something wrong with those casks?

vkarampinis:
@ran-dall probably I don't understand your whole process.
I don't want to create problems here and ruin the work that you doing for years. So maybe the best is to stop here and quit the whole process.
Sorry for any disruption.

ran-dall:
> I manual confirmed each update I send.

Alright, that's good. I'll take your word for it. My mistake. My apologies.

> Also, I checked the output files and only updated those marked as no auto cask-repair: technically impossible.

Those still get processed; however, they do take more power to process. So technically you're not doing anything bad thing here. There is a hiccup though when the ML is processing updates and you're updating it at the same time. This causes an error it's logical processing, that usually requires me to just delete all the analytics we have on the cask; and it needs to start all over again. The best thing though would be to wait for my output for the day to get posted with the errors. You'll notice that there are specific ones like `logos` that the ML just can't update. If you go through the previous days of output, look for the casks that are `i` or `ignored` cause those are casks that the ML prefers to ignore because it's had troubles in the past.

ran-dall:
@vkarampinis For the record, I'm not mad or anything. I really do appreciate your enthusiasm and effort. 

...and you're not really disrupting anything, that much, yet. So it's not a bad thing, we got ourselves sorted. I'd say start with the `ignored` casks; if you want to get some updates done. Then when I post the output shortly, look for the `utu` or `unable to update` casks.

I apologize if I sounded a bit harsh in my initial message. As many around here will tell you, I'm a pragmatist; not many things bother me.  It's just you updated 2 casks today (`tuple` and `qownnotes`) that are hard for the ML to keep its logic up-to-date. So I was trying to be direct (not harsh) and catch it in time to not have to do more work later.

PR #80263, opened 2020-04-08
Update vscodium from 1.43.2 to 1.44.0

nathanwhit:

### DIFF
--- a/Casks/vscodium.rb
+++ b/Casks/vscodium.rb
@@ -1,6 +1,6 @@
 cask 'vscodium' do
-  version '1.43.2'
-  sha256 '3a81ed31c30cd758b16601a764d289561ad57b4da02bfaa01ebdbff9a577215e'
+  version '1.44.0'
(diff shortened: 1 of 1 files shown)


core-code (commented):
>  tag = '1.43.2'

whats this?

nathanwhit:
> > tag = '1.43.2'
> 
> whats this?
The maintainer didn't create a new tag on Github for the 1.44.0 release, so the package is (most likely mistakenly) filed under the `1.43.2` tag. The original download URL assumed that the tag and version match, but in this case they don't, so I manually specified the correct tag in order to make the correct download URL.

vitorgalvao:
We don‘t like having random variables in the middle of casks. Please hardcode the tag version into the URL instead.

nathanwhit:
Understood. I've applied that feedback and pushed the revised version.

reitermarkus:
Seems like there is a 1.44.0 tag now.

nathanwhit:
Indeed there is, it looks like I was a little early on this one.
I've updated the PR once more to use the correct tag.

PR #80297, opened 2020-04-09
Add AirtestIDE 1.2.3

zengxs:

-------

When brew download the cask dmg file, may raise error:
```
curl: (18) transfer closed with 263019584 bytes remaining to read
```

But if use `wget` to download it or try a few more times, download will be succeed.

Download via Safari is no problem also.

### DIFF
--- a/Casks/airtest-ide.rb
+++ b/Casks/airtest-ide.rb
@@ -4,7 +4,7 @@ cask 'airtest-ide' do
 
   url 'https://airtestproject.s3.netease.com/downloads/AirtestIDE/AirtestIDE_2020-01-20_py3_Mac10-12.dmg'
   name 'AirtestIDE'
-  homepage 'https://airtest.netease.com'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> But if use `wget` to download

We can’t use `wget` to download. Homebrew uses `curl`.

> Download via Safari is no problem also.

We also can’t download via Safari.

> or try a few more times

Which means we’ll be getting bug reports for something we can’t fix. I see that CI already failed with that error. This is something that needs to be fixed on their end. Until the download can be reliable, I’d rather not include the software.

Thank you anyway.

Casks/airtest-ide.rb
```diff
@@ -0,0 +1,10 @@
+cask 'airtest-ide' do
```
vitorgalvao, line 1: Does not follow the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).

PR #80305, opened 2020-04-09
Create android-clt.rb

robertleeblairjr:
The Android SDK Tools package has been deprecated and replaced by the Command Line Tools. Because of conflicts/removals/additions to each package for specific binaries, then this may be a better alternative to creating a cask for the command line tools. Perhaps, moving the cask 'android-sdk' to the homebrew-cask-versions tap because of deprecation would be better.

### DIFF
--- a/Casks/android-clt.rb
+++ b/Casks/android-clt.rb
@@ -16,11 +16,11 @@ cask 'android-clt' do
   binary "#{staged_path}/tools/bin/sdkmanager"
 
   postflight do
-    FileUtils.ln_sf(staged_path.to_s, "#{HOMEBREW_PREFIX}/share/android-sdk")
+    FileUtils.ln_sf(staged_path.to_s, "#{HOMEBREW_PREFIX}/share/android-clt")
   end
 
   uninstall_postflight do
-    FileUtils.rm_f("#{HOMEBREW_PREFIX}/share/android-sdk")
+    FileUtils.rm_f("#{HOMEBREW_PREFIX}/share/android-clt")
   end
 
   caveats do
@@ -28,7 +28,7 @@ cask 'android-clt' do
     <<~EOS
       You can control android sdk packages via the sdkmanager command.
       You may want to add to your profile:
-        'export ANDROID_SDK_ROOT="#{HOMEBREW_PREFIX}/share/android-sdk"'
+        'export ANDROID_SDK_ROOT="#{HOMEBREW_PREFIX}/share/android-clt"'
     EOS
   end
 end
--- a/Casks/android-clt.rb
+++ b/Casks/android-clt.rb
@@ -1,0 +1,34 @@
+cask 'android-clt' do
+  version '6200805'
+  sha256 '23f0626336a98d70aff7311a73292026af31bc577c6f06b509cd4ad33752313e'
+
+  # dl.google.com/android/repository was verified as official when first introduced to the cask
+  url "https://dl.google.com/android/repository/commandlinetools-mac-#{version}_latest.zip"
+  name 'android-clt'
+  homepage 'https://developer.android.com/'
+  
+  conflicts_with cask: 'android-sdk'
+
+  binary "#{staged_path}/tools/bin/apkanalyzer"
+  binary "#{staged_path}/tools/bin/avdmanager"
+  binary "#{staged_path}/tools/bin/lint"
+  binary "#{staged_path}/tools/bin/screenshot2"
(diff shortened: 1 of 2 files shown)


vladimyr:
Unfortunately, this won't work. If you try to run `sdkmanager` CLI you'll run into https://stackoverflow.com/questions/60440509/android-command-line-tools-sdkmanager-always-shows-warning-could-not-create-se#comment107440489_60598900 due to specific directory layout that is expected. To make it work, you need to create the following directory structure: `<android-sdk-root>/cmdline-tools/latest` and move extracted `tools` there. The first segment is then assumed to be `ANDROID_SDK_ROOT` and downloaded SDK parts will be installed inside it, alongside command-line tools. To answer https://stackoverflow.com/questions/60440509/android-command-line-tools-sdkmanager-always-shows-warning-could-not-create-se#comment107440489_60598900 it comes from source code itself 🤷

robertleeblairjr:
@vladimyr Thanks. I'll look at it. I just took it that the binaries, although updated, would function from the same directory structure as the SDK tools. Not sure, why it would make a difference but I'll check your SOF links for suggestions. But, from your suggestion, then I would just need to change the cask directory instead of the "{HOMEBREW_PREFIX}/android-sdk/tools" to "{HOMEBREW_PREFIX}/android-clt/cmdline-tools/latest".

vladimyr:
>But, from your suggestion, then I would just need to change the cask directory instead of the "{HOMEBREW_PREFIX}/android-sdk/tools" to "{HOMEBREW_PREFIX}/android-clt/cmdline-tools/latest".

That should work although it is a bit weird name because you can install visual tools inside it too.

Casks/android-clt.rb
```diff
@@ -0,0 +1,34 @@
+cask 'android-clt' do
+  version '6200805'
+  sha256 '23f0626336a98d70aff7311a73292026af31bc577c6f06b509cd4ad33752313e'
+
+  # dl.google.com/android/repository was verified as official when first introduced to the cask
+  url "https://dl.google.com/android/repository/commandlinetools-mac-#{version}_latest.zip"
```
vitorgalvao, line 6: How do we get here from the homepage?

vitorgalvao:
I’m not an Android developer, so I’ll need to rely on you to reach a conclusion regarding what works or not. In terms of cask constructions, this (so far) is acceptable to go in, but it needs to be something that works and is useful (from the conversation, unsure if we’re at that step now).

robertleeblairjr (commented):
> I’m not an Android developer, so I’ll need to rely on you to reach a conclusion regarding what works or not. In terms of cask constructions, this (so far) is acceptable to go in, but it needs to be something that works and is useful (from the conversation, unsure if we’re at that step now).

Yep. I'm not either. I just used it a few times for APKs. Only knew it was deprecated from the page, which was some time ago, and replaced.

Casks/android-commandlinetools.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  conflicts_with cask: 'android-sdk'
+
+  binary "#{staged_path}/tools/bin/apkanalyzer"
+  binary "#{staged_path}/tools/bin/avdmanager"
+  binary "#{staged_path}/tools/bin/lint"
+  binary "#{staged_path}/tools/bin/screenshot2"
+  binary "#{staged_path}/tools/bin/sdkmanager"
+
+  postflight do
```
vitorgalvao, line 19: Are these really needed, still?

vitorgalvao:
What’s the state of this? Is this working? We can’t deal with https://github.com/Homebrew/homebrew-cask/pull/79981/ until this is in a state that it can be merged and works.

tir38:
I tested this.

1. I can confirm those `binary` lines need to be there so that those individual tools are symlinked
2. The bug that @vladimyr reported may be fixed in next version of `sdkmanager` / `commandlinetools`

https://issuetracker.google.com/issues/150942306

I'll keep an eye on https://developer.android.com/studio?hl=en-419#downloads and see when next version of commandlinetools ships. Since that's a bug with commandlinetools and not brew or this cask I'm not sure if that should actually hold up this PR.

vitorgalvao:
> Since that's a bug with commandlinetools and not brew or this cask I'm not sure if that should actually hold up this PR.

Unfortunately, too many users don’t understand that distinction. They’ll still complain to us and increase the maintenance burden. We’ll merge the cask when it works.

sgarfinkel:
Why was this PR closed?

PR #80374, opened 2020-04-10
Update keepassxc

vladimyr:
Add sierra download URL & checksum.
Continuation of #80316

### DIFF
--- a/Casks/keepassxc.rb
+++ b/Casks/keepassxc.rb
@@ -1,7 +1,7 @@
 cask 'keepassxc' do
   version '2.5.4'
 
-  if MacOS.version == :sierra
+  if MacOS.version <= :sierra
     # github.com/keepassxreboot/keepassxc was verified as official when first introduced to the cask
     url "https://github.com/keepassxreboot/keepassxc/releases/download/#{version}/KeePassXC-#{version}-Sierra.dmg"
     sha256 '7cd8dc34022091c240e538f7a9889afd7dc8f9f3957a66bca9d70c067045ade4'
--- a/Casks/keepassxc.rb
+++ b/Casks/keepassxc.rb
@@ -1,14 +1,21 @@
 cask 'keepassxc' do
   version '2.5.4'
-  sha256 '3d56ebbcb3471a2f6116abe884d1e7b662347b80e18cc0a12ca9fdf2c9a14d7a'
 
-  # github.com/keepassxreboot/keepassxc was verified as official when first introduced to the cask
-  url "https://github.com/keepassxreboot/keepassxc/releases/download/#{version}/KeePassXC-#{version}.dmg"
+  if MacOS.version == :sierra
+    # github.com/keepassxreboot/keepassxc was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


core-code (approved):

Casks/keepassxc.rb
```diff
@@ -1,14 +1,21 @@
 cask 'keepassxc' do
   version '2.5.4'
-  sha256 '3d56ebbcb3471a2f6116abe884d1e7b662347b80e18cc0a12ca9fdf2c9a14d7a'
 
-  # github.com/keepassxreboot/keepassxc was verified as official when first introduced to the cask
-  url "https://github.com/keepassxreboot/keepassxc/releases/download/#{version}/KeePassXC-#{version}.dmg"
+  if MacOS.version <= :sierra
```
vitorgalvao, line 4: We prefer these to always be `<=`, because it gives us more flexibility on later changes.

vladimyr: Just to make sure I understand it correctly. This is preferred regardless of `depends_on` that explicitly advertises sierra as a minimum supported version?

vitorgalvao: > Just to make sure I understand it correctly. This is preferred regardless of `depends_on` that explicitly advertises sierra as a minimum supported version?

Yes. At the moment they are equivalent, but if we ever have to change this `if` or change the `depends_on`, having the `<=` prevents some mistakes.

vladimyr: I get it but it should be probably enforced with an audit. 💡

vitorgalvao: > I get it but it should be probably enforced with an audit.

[Some casks](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/onyx.rb) do use `==` because there’s a specific release for every major macOS version.

vladimyr: > I get it but it should be probably enforced with an audit. 💡

...if it is a two-branch conditional with unconditional else branch.

vitorgalvao: > if it is a two-branch conditional with unconditional else branch.

There are conditionals with more branches that require `<=` as well. Maybe something like “if the `else` would trigger on a version under the lowest branch, fail”. That way it still allows `==`, but works in tandem with the `depends_on`. It wouldn’t fix the preference of having `<=` anyway (because it ends up being safer), but it would prevent some mistakes.

I don’t think this is problematic enough to warrant an extra audit (part of it is style), but if you find it relevant enough to warrant a PR, I wouldn’t oppose it either.

vladimyr: > “if the `else` would trigger on a version under the lowest branch, fail”. That way it still allows `==`, but works in tandem with the `depends_on`.

Nicely done!

> I don’t think this is problematic enough to warrant an extra audit (part of it is style), but if you find it relevant enough to warrant a PR, I wouldn’t oppose it either.

Yes it is a convention, and ofc it can be found in the docs
https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/readme.md#always-fall-through-to-the-newest-case 🤦 

I solemnly swear I'll read those markdown files like a damn novel before I create another mess 🤦

PR #80399, opened 2020-04-10
find-appcast: fix path & url slash trimming

vladimyr:
Add additional digitalocean appcast URL pattern - in response to https://github.com/Homebrew/homebrew-cask/pull/80025#issuecomment-609116213

### DIFF
--- a/developer/bin/find-appcast
+++ b/developer/bin/find-appcast
@@ -48,10 +48,10 @@ def find_electron_builder(app)
   channel = data['channel']
   name = data['name']
   owner = data['owner']
-  path = data['path'].sub(%r{^/(.*)/$}, '\1') rescue nil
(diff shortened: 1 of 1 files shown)


vladimyr:
@vitorgalvao https://github.com/Homebrew/homebrew-cask/commit/2cc0f864732590ceaeb2690126d4cf09f020f52a won't work because path might have only leading slash like in Tinkerwall case so you'll end up with: 
```rb
path = "/tinkerell"
url = "https://tinkerwell.fra1.digitaloceanspaces.com//tinkerwell/latest-mac.yml"
```
And fetching that URL ends up in 403 response.

vitorgalvao:
Thank you. Better idea, though: https://github.com/Homebrew/homebrew-cask/pull/80450

PR #80360, opened 2020-04-10
Update vlc from 3.0.8 to 3.0.9.2

miccal:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #57356, opened 2019-01-11
Update vlc to 3.0.6

yurikoles:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #80511, opened 2020-04-12
Update vlc appcast

core-code:

### DIFF
--- a/Casks/vlc.rb
+++ b/Casks/vlc.rb
@@ -3,7 +3,7 @@ cask 'vlc' do
   sha256 'a30ca5f664c27c64f0df8ed64e85844a3422ba8dcf9b962e02acfbdd63a17190'
 
   url "https://get.videolan.org/vlc/#{version}/macosx/vlc-#{version}.dmg"
-  appcast 'https://download.videolan.org/pub/videolan/vlc/'
+  appcast 'https://www.videolan.org/vlc/download-macosx.html'
   name 'VLC media player'
   homepage 'https://www.videolan.org/vlc/'
 


core-code:
this change is aimed at preventing auto-merges of unofficial releases in the future ( https://github.com/Homebrew/homebrew-cask/pull/80360#issuecomment-612598071 )

Casks/vlc.rb
```diff
@@ -3,7 +3,7 @@
   sha256 'a30ca5f664c27c64f0df8ed64e85844a3422ba8dcf9b962e02acfbdd63a17190'
 
   url "https://get.videolan.org/vlc/#{version}/macosx/vlc-#{version}.dmg"
-  appcast 'https://download.videolan.org/pub/videolan/vlc/'
```
vladimyr, line 6: @core-code This URL originates from your earlier suggestion https://github.com/Homebrew/homebrew-cask/pull/57356#issuecomment-453482629 but both trac tickets are now resolved. Maybe it is time to reconsider it and go back to official [appcast](https://update.videolan.org/vlc/sparkle/vlc-intel64.xml)? 🤔

core-code: they are not really resolved. the VLC developers *want* the appcast to lag behind the official releases for up to a few weeks and they are just closing all bug-reports related to this. 

(yes the sparkle feed it is up-to-date now but the last release has been more than 6 months ago and i bet it will not have the next release in a timely manner either(

i think using the official sparkle feed is bad because it lags by a few weeks and will not let official releases through.

i think using the FTP is bad too because it will let unreleased stuff like "3.0.9" through which is STILL not officially released.

what we want to track are official releases from the homepage, so using this as the appcast makes most sense.

core-code: this could work too "http://www.videolan.org/vlc/releases/"

vladimyr: > the VLC developers _want_ the appcast to lag behind the official releases for up to a few weeks and **they are just closing all bug-reports related to this.**

I was missing that crucial part and I figured out you'll have more details 👍  

> this could work too "http://www.videolan.org/vlc/releases"

This looks great, just let's go with secure version: https://www.videolan.org/vlc/releases/ 👍

core-code: thinking about it i think its better to use "https://www.videolan.org/vlc/download-macosx.html" because if the app-cast contains old versions in addition to the current one we have troubles detecting updates automatically.

suschizu: the vlc cask is still on 3.0.9.2 - I've reverted it and added the new appcast:
https://github.com/Homebrew/homebrew-cask/pull/80654/files
or should I close it?

core-code: thanks, i think reverting is the right thing

PR #80521, opened 2020-04-12
Update arq from 5.17.3 to 6.1.6

core-code:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #80531, opened 2020-04-12
Update firecamp from 1.2.0 to 1.2.1

suschizu:


core-code:
`Error: It seems the App source '/usr/local/Caskroom/firecamp/1.2.2/Firecamp.app' is not there.
`

ran-dall:
@suschizu This is technically a duplicate of #80389, but maybe you can fix it first?

core-code:
@reitermarkus we've seen a lot of 'app source is not there' errors in the past days that were false-positives and go away after restarting travis - like in this case. also there are weird failures like in #80521 

any idea whats up with travis?

PR #80573, opened 2020-04-13
Delete bfxr

miccal:
`url` appears dead, cannot download via the `homepage`, and the Github site https://github.com/increpare/bfxr does not host any releases.


aaronliu0130:
@miccal The url does not appear dead at all to me.
image
(Note the download at the bottom)

aaronliu0130:
I'm in china, and the download speed is quite slow here though.
image

miccal:
@aaronliu0130 thank you -- it is definitely not working for me, but it must be a glitch for me only.

PR #80593, opened 2020-04-13
Update betwixt

vladimyr:
Remove cert on uninstall.

### DIFF
--- a/Casks/betwixt.rb
+++ b/Casks/betwixt.rb
@@ -9,6 +9,29 @@ cask 'betwixt' do
 
   app 'Betwixt-darwin-x64/Betwixt.app'
 
+  uninstall_postflight do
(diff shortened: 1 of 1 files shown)


core-code:
```
==> Removing App '/Applications/Betwixt.app'.
undefined method `split' for nil:NilClass
Error: Build failed.
```

vladimyr:
@core-code Yeah I'm aware of that. Thing is I can't reproduce it locally and it is probably caused by the unexpected output of `certtool d <cert>` invocation. I'll try with printf style debugging on CI server 🙈

vladimyr:
I've figured it out. The certificate is created after the first run: https://github.com/joeferner/node-http-mitm-proxy#ssl so removal must be done conditionally.

core-code:
good job! i don't feel qualified to merge this though so you'll have to wait for one of the PROs ;)

vitorgalvao:
This should definitely be a new stanza.

PR #80597, opened 2020-04-13
Fix alacritty un/install

darwingr:

       **NO VERSION NUMBER. This isn't a version update.**

### DIFF
--- a/Casks/alacritty.rb
+++ b/Casks/alacritty.rb
@@ -8,6 +8,7 @@ cask 'alacritty' do
   homepage 'https://github.com/alacritty/alacritty/'
 
   app 'Alacritty.app'
+  binary "#{appdir}/Alacritty.app/Contents/MacOS/alacritty"
 
   zap delete: [
                 '~/Library/Saved Application State/io.alacritty.savedState',
--- a/Casks/alacritty.rb
+++ b/Casks/alacritty.rb
@@ -11,6 +11,5 @@ cask 'alacritty' do
 
   zap delete: [
                 '~/Library/Saved Application State/io.alacritty.savedState',
-                '~/.config/alacritty',
               ]
 end


Casks/alacritty.rb
```diff
@@ -8,9 +8,9 @@
   homepage 'https://github.com/alacritty/alacritty/'
 
   app 'Alacritty.app'
+  binary "#{appdir}/Alacritty.app/Contents/MacOS/alacritty"
 
   zap delete: [
                 '~/Library/Saved Application State/io.alacritty.savedState',
-                '~/.config/alacritty',
```
vitorgalvao, line 14: Why are you removing this?

Casks/alacritty.rb
```diff
@@ -8,9 +8,9 @@
   homepage 'https://github.com/alacritty/alacritty/'
 
   app 'Alacritty.app'
+  binary "#{appdir}/Alacritty.app/Contents/MacOS/alacritty"
```
vitorgalvao, line 11: Unless there’s official documentation recommending this, we don’t do it.

vitorgalvao (changes requested):

Casks/alacritty.rb
darwingr: The is okay to execute directly on MacOS and other platforms.
For example, here's the output you actually get with `-h`:
image


Here's the documentation on how to use alacritty on the command line to make it easier to configure your keyboard mappings: https://github.com/alacritty/alacritty/wiki/Keyboard-mappings

Casks/alacritty.rb
darwingr: Sorry, I put it in the commit message.
This file is never created by the app. The only way it is ever created is intentionally/deliberately by the user. If that's the case they likely would not want it deleted along with the other trash.

The same approach would go for MacVim, Wireshark, Emacs, GnuCash, etc,  to delete their respective user configs found in `~/.config/` and `~/`.

PR #80657, opened 2020-04-14
Update url and hompage in eqmac.rb

suschizu:
the official website just offers the beta version as download - version 2.2 is only available on github - so I think the website should also link to this site …


suschizu:
`Error: Failure while executing; `/usr/bin/sudo -E -- /usr/bin/env`

not sure what I can do here

vitorgalvao:
Travis passed.

PR #80658, opened 2020-04-14
Update appcast in gimp.rb

suschizu:
old url isn't working anymore - the problem here is that the windows/linux version isn't in sync with the mac version … but they just post stable releases


core-code:
why not 
`https://download.gimp.org/pub/gimp/v2.10/osx/`

michaelblyons:
When this got merged, `brew update` flagged Gimp as "changed," and I got all excited, thinking 2.10.18 was out for MacOS. 😞

suschizu:
> When this got merged, `brew update` flagged Gimp as "changed," and I got all excited, thinking 2.10.18 was out for MacOS. 😞
sorry for making false hopes 🍫

PR #80691, opened 2020-04-14
Update free-download-manager from 5.1 to 5.1.38

suschizu:


suschizu:
perhaps it's better to have the full version in the cask(?) there have been already about 20 versions of the free-download-manager

PR #80902, opened 2020-04-18
Update foldit

vladimyr:
Add macOS version constraint & `zap` stanza.

### DIFF
--- a/Casks/foldit.rb
+++ b/Casks/foldit.rb
@@ -4,7 +4,11 @@ cask 'foldit' do
 
   url 'https://fold.it/portal/files/app/Foldit-macos_x64.dmg'
   name 'Foldit'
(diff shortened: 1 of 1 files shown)


core-code:
while their homepage says that 10.12 is required, there is no LSMinimumSystemRequirements in the app. someone should send them some feedback :)

vladimyr:
> while their homepage says that 10.12 is required, there is no LSMinimumSystemRequirements in the app. someone should send them some feedback :)

You are right but that won't be me. I'm kinda put off by the _forum driven_ development and quality of the product itself 😉 

@ran-dall What is the question? 🙃

ran-dall:
@vladimyr No question, just waiting for your reply when you just did.

PR #80959, opened 2020-04-19
Update hook from 1.4 to 1.5.1

thebengeu:


vitorgalvao:
Unfortunately, we’ll need to remove this cask. Appcenter links break on their own all the time, and as such do not work in casks for long.

PR #80960, opened 2020-04-19
Delete hook.rb

vitorgalvao:
Appcenter URL, so it breaks on its own.


core-code:
just an idea: it would be possible to write a CGI that gets a appcenter/sparkle URL as input and would automatically redirect to the download URL of the latest feed entry. the url of this CGI could be used in the URL stanza - also  "version :latest" would be advisable

not really a serious suggestion but it could be a technical possibility if some really important/popular apps had this problem

ran-dall:
> ...it would be possible to write a CGI that gets a appcenter/sparkle URL as input and would automatically redirect to the download URL of the latest feed entry. the url of this CGI could be used in the URL stanza - also "version :latest" would be advisable

@core-code This ML essentially does this on already to verify certain updates; however certain things (usually special characters) make it fail sometimes. It's one of those things thats on my to-do list. It can also parse a page for the download link on a page and use that; which is what I've been talking about when I say that there could be a point that the ML would no longer need an `appcast`, although that wouldn't be soon.

vitorgalvao:
> it would be possible to write a CGI that gets a appcenter/sparkle URL as input and would automatically redirect to the download URL of the latest feed entry.

That would work akin to a [`url do` block](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#writing-the-block), which we’d like to avoid for stable (main repo) casks, in part because (as you mentioned) they would require `version :latest`.

core-code:
ok, i see that. isn't a 'version :latest' cask better than no cask at all though?

vitorgalvao:
> isn't a 'version :latest' cask better than no cask at all though?

Yes, unless it’s an unpopular cask (this one is) which is hard to maintain (`url do` makes it harder).

While it’s fine to have one, my experience with adding an exception to a cask is that suddenly a bunch of them start to pop up with the same workaround. And when we allow one, users start complaining about not allowing others and we waste a bunch of time explaining (and arguing for) the difference.

What increases complexity is that `url do` can take arbitrary ruby code. If we abstracted a new `url` option for appcenter urls, the casks themselves would remain simple.

[Maybe the URLs themselves will get better.](https://github.com/Homebrew/homebrew-cask-versions/pull/8955#issuecomment-616001904)

Closing as this now has conflicts.

PR #80965, opened 2020-04-19
Update stand from 1.0.4 to 2.0

core-code:

### DIFF
--- a/Casks/stand.rb
+++ b/Casks/stand.rb
@@ -1,11 +1,10 @@
 cask 'stand' do
-  version '1.0.4'
-  sha256 '8919e43c9c591657d8d6961b25e8dc5f77d706d71eb246839be22522a82bb0ec'
+  version '2.0'
+  sha256 'eb0555f6376757105e1faea1fd1f9e4afae21eef73fdd5c7d0fb0c76a8cd9702'
 
-  # get-stand-app.s3.amazonaws.com was verified as official when first introduced to the cask
-  url "https://get-stand-app.s3.amazonaws.com/#{version.patch}/Stand.zip"
-  appcast 'https://standapp-sparkle-updater.herokuapp.com/',
-          configuration: version.major_minor
+  # f001.backblazeb2.com/file/stand-app/ was verified as official when first introduced to the cask
+  url "https://f001.backblazeb2.com/file/stand-app/#{version}/Stand.zip"
+  appcast 'https://standapp-sparkle-updater.herokuapp.com/'
   name 'Stand'
   homepage 'https://getstandapp.com/'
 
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao they ask for money via gumroad on their product site now. i don't know if this was the case previously. while it can be downloaded for free when selecting 0$ as an option, i am not sure if circumventing their paywall is the right thing here

ran-dall:
@core-code FWIW I hate it when companies exhibit this type of behavior.

core-code:
```

   # f001.backblazeb2.com/file/stand-app/ was verified as official when first introduced to the cask
   # f001.backblazeb2.com/file/stand-app was verified as official when first introduced to the cask
```

isn't that unsafe because 'f001.backblazeb2.com/file/stand-app-fake' would also match?

vitorgalvao:
> isn't that unsafe because 'f001.backblazeb2.com/file/stand-app-fake' would also match?

That’s a bloody good point. And since that’s the standard, it also means we’ll have to correct every cask that uses this comment (2038)!

I’ll make a script later.

vitorgalvao:
> they ask for money via gumroad on their product site now. i don't know if this was the case previously.

[Their pinned tweet](https://twitter.com/getstandapp/status/1250777720470016000) makes me think so (emphasis added):

> (…) a new version of Stand is available! Stand 2 (…) Existing users can check for updates in app, and it's also **available at the same 'pay what you like' price point** (…)

On the website, they mention (emphasis added):

> We built Stand in our spare time, and if you’d like to download it, **you’re welcome to pay whatever you think it’s worth**

And [since](https://github.com/homebrew/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md):

> our users are expected to have reasonable knowledge about the apps they’re installing through us before doing so.

I’d say this might be OK. By accepting $0, they’re essentially asking for donations.

core-code:
sounds good, i agree! maybe a 'caveats' would be nice in these cases too, although i realise that we try to minimise that.

core-code:
> That’s a bloody good point. And since that’s the standard, it also means we’ll have to correct every cask that uses this comment (2038)!


i really thought the standard was the other way round exactly for this reason. unless the implementation adds the trailing slash automatically?

vitorgalvao:
No, the implementation uses what we give it. That’s necessary for cases when a version is part of the comment, since those don’t take the interpolation.

Somewhere in the beginning we decided on that and never revisited, I guess (I no longer remember).

vitorgalvao:
I’m fixing the comments now, and think I may remember why we decided on not having the trailing slash at the time.

A lot of these cases (perhaps the majority) only need the differing domain (no subdirectory) in the comment. Seeing as those end in a TLD, they don’t pose a problem (or should I say *didn’t*, since TLD names are now the Wild West). This check is in `style`, not `audit`, so we may have decided that all else being equal, we’d go for what looks best.

I do remember writing that the comment should contain “the smallest possible portion of the URL that uniquely identifies the app or vendor”, back when were still under the Caskroom org, but by that time it’s likely we didn’t reconsider the stylistic part of the choice.

PR #80988, opened 2020-04-20
Add trailing slash to verified comments that benefit from it

vitorgalvao:


michaelblyons:
Can you add some brief mention of why this was important to do?

vitorgalvao:
@michaelblyons It makes the verification more secure. See conversation in https://github.com/Homebrew/homebrew-cask/pull/80965#issuecomment-616232313.

Truth be told, the overwhelmingly majority didn’t need this change because the domain alone was enough. On the other hand, it doesn’t hurt, and even if it improves security just a notch without downsides, it’s worth doing.

michaelblyons:
Thanks! The link explains it very well. I was just really surprised when `brew update` said a huge stack of casks were updated, leading me to this PR without any context.

Thanks again.

PR #5342, opened 2014-07-14
Removing Hydra per Hydra installation instructions

sdegutis:
See https://github.com/sdegutis/hydra/#install.

PR #81154, opened 2020-04-22
Remove Tuple from Casks

robamaton:
This is a straight removal, is there anything else I need to check?


vladimyr:
I'm curious, what's the motivation behind it?

ran-dall (approved):
@robamaton ...is one of `tuple`'s main developers 👍 If you don't mind me asking, @robamaton, why is `tuple` being deleted?

robamaton:
We haven't been pushing people towards homebrew for a long time and have updated our onboarding to assume people were downloading the app directly through the website. We were having people guess that it was on homebrew and having issues when installing. I'm not 100% sure what the underlying issue is and might add back later but for now want to keep things clean unless we decide to actively support/test homebrew installation.

vladimyr:
Disclaimer, I'm definitely not a decision-maker here and thoughts are strict of my own. I fail to understand what makes this cask much different than https://github.com/tupleapp/tuple-install/blob/master/bin/tuple-install After all whole process boils down to fetch, extract, and move in the right place.
Also, this is clearly not an official distribution channel so receiving installation-related complaints from brew users sounds weird. Furthermore, you aren't and should not be required to support it because it is a community effort and you haven't added it in the first place?
All in all, I don't see a good reason for the removal instead those who use it should be aware that it might sometimes fail or lag behind the official distribution channel. However, that doesn't sound like a good justification for preventing devs, who in fact do acknowledge that, from happily using it?

vladimyr:
FWIW [author's request](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/rejected_casks.md) is enough for removal to happen:
>+ The author has [specifically asked us not to include it](https://github.com/Homebrew/homebrew-cask/pull/5342).

No matter how it ends up in the end, I think you created a great product and wish you all the best in future development & monetization!

robamaton:
I think we'll hold off on this for now, we're going to reconsider if this is necessary and how easy it would be to clarify our messaging to users. As far I we can tell the cask installation does work right now.

PR #81193, opened 2020-04-22
Update chromium from 761197 to 762311

miccal:


miccal:
The `binary` I added previously exits with `Abort trap: 6`, so I have removed it and replaced it with a `shimscript`. This is required to [run Chromium with flags](https://www.chromium.org/developers/how-tos/run-chromium-with-flags).

PR #81081, opened 2020-04-21
update wireshark

lulor:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #81205, opened 2020-04-23
Update intel-power-gadget

lulor:

Similarly to #81081, `intel-power-gadget` comes with an `Uninstaller.pkg` provided by Intel itself.
Maybe it could be preferable to use that instead of manually replicate its function?
Correct me if I'm wrong, so that I can adjust the PR to just remove the config files in the `zap` stanza

### DIFF
--- a/Casks/intel-power-gadget.rb
+++ b/Casks/intel-power-gadget.rb
@@ -15,4 +15,9 @@ cask 'intel-power-gadget' do
 
   uninstall pkgutil: 'com.intel.pkg.PowerGadget.*',
             kext:    'EnergyDriver'
+
+  zap trash: [
+               '~/Library/Caches/com.intel.PowerGadget',
+               '~/Library/Preferences/com.intel.PowerGadget.plist',
+             ]
 end
--- a/Casks/intel-power-gadget.rb
+++ b/Casks/intel-power-gadget.rb
@@ -3,8 +3,7 @@ cask 'intel-power-gadget' do
   sha256 '22ed3fe050c3b965841ccc5590a3a03bb9498f8620e01ba4dea5557dcd24fe43'
 
   url "https://software.intel.com/file/#{version.after_comma}/download"
-  appcast 'https://software.intel.com/en-us/articles/intel-power-gadget',
-          configuration: version.after_comma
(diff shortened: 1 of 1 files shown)


core-code:
can you use this as an 'appcast'?

`https://software.intel.com/sites/landingpage/powergadget/?app=IntelPowerGadgetMac`

lulor:
> can you use this as an 'appcast'?
> 
> `https://software.intel.com/sites/landingpage/powergadget/?app=IntelPowerGadgetMac`

Sure! Should I leave the "configuration" field as it is? To be honest, it's not entirely clear to me what its purpose is

core-code:
the appcast is used to prevent beta versions being submitted and auto-merged. if the version number isn't contained in the appcast, travis will fail. the 'configuration' tells travis which part of the version number is expected to be contained in the appcast. 

you can delete the configuration since the full version number should be contained in the new appcast

core-code:
thanks!

lulor:
> thanks!

@core-code  Thank you for the explanation!

Anyway, the tests are not going to succeed because of some COU r

lulor:
We

lulor:
@core-code thank you for the explanation!

Anyway, the tests are not going to succeed because of some CPU requirements from Intel

Casks/intel-power-gadget.rb
```diff
@@ -13,6 +12,16 @@
 
   pkg 'Install Intel Power Gadget.pkg'
 
-  uninstall pkgutil: 'com.intel.pkg.PowerGadget.*',
```
vitorgalvao, line 16: We’d rather keep this like it is, if tests succeed.

vitorgalvao (changes requested):

Casks/intel-power-gadget.rb
lulor: Thanks!
I don't think the tests are going to succeed, since the application fails to install because of some CPU requirements from Intel that are not satisfied (only 1 CPU package).

Should I revert to using the `kext` key anyway?

vitorgalvao: I meant if they succeed in that area, if it still works.

If there’s a situation where they’ll fail either way, ignore that when evaluating.

lulor: I understand. I went back to the previous approach, it should be good now

PR #81226, opened 2020-04-23
Delete rightzoom.rb

suschizu:
URL is offline - not sure since when


hautonjt:
The URL seems to be back? I was able to access it just now.

PR #81252, opened 2020-04-23
Update blockblock from 0.9.9.4 to 1.0.1

benmontour-wf:

### DIFF
--- a/Casks/blockblock.rb
+++ b/Casks/blockblock.rb
@@ -1,12 +1,14 @@
 cask 'blockblock' do
-  version '0.9.9.4'
-  sha256 '6ab3a8224e8bc77b9abe8d41492c161454c6b0266e60e61b06931fed4b431282'
+  version '1.0.1'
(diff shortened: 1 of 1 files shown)


benmontour-wf:
With this update, since this version requires MacOS 10.15 Catalina. Will the older version 0.9.9.4 stay available as the latest download for those running < 10.15?

core-code:
only if you create the necessary if/else code to support this

benmontour-wf:
> only if you create the necessary if/else code to support this

I will work on this. Looks like after testing locally I need to update the executable path that's contained within the .app file as well. Withdrawing this PR and fixing.

core-code:
it might be difficult to adapt the 'installer script' stanza to work both with the old and the new version :(

benmontour-wf:
> it might be difficult to adapt the 'installer script' stanza to work both with the old and the new version :(

Could I put my first IF first thing in the stanza based on MacOS.version <= :mojave
and then reference the current installations. Then an ELSE below that with the new stanza?

core-code:
@vitorgalvao any ideas here?

PR #81265, opened 2020-04-24
rejected_casks.md: update notability requirements

vitorgalvao:
Continuation of https://github.com/Homebrew/brew/pull/7427.


core-code:
why the "30 forks"? not sure if they will be correlated with popularity in each and every case (e.g. software that is hosted on github but is proprietary and/or has a closed or sponsored development).

vitorgalvao:
> why the "30 forks"? not sure if they will be correlated with popularity in each and every case

I agree, though I can’t answer that as it wasn’t my decision. That’s how Homebrew has been doing it for a while, and since we’re using their code, it makes sense to have the same rules. As for their reasoning, we’d have to ask.

> (e.g. software that is hosted on github but is proprietary and/or has a closed or sponsored development).

Those typically have a homepage which differs from github, so if the notability check uses `homepage` instead of `url`, it shouldn’t be a big issue. I’ve asked.

core-code:
ah ok :)

ran-dall:
@vitorgalvao I recently was asked why the notability requirements are not on the PR checklist. (https://github.com/Homebrew/homebrew-cask/pull/81328#issuecomment-619341937) 

I think it would be smart to add it. Could we do add that to this PR as we're updating notability requirements already?

vitorgalvao:
@ran-dall The notability requirements [will be part of `audit`](https://github.com/Homebrew/brew/pull/7427) soon, so there’s no need to add a separate checklist item (`audit` is already there).

claui (approved):

PR #81271, opened 2020-04-24
Delete go2shell

miccal:
`homepage` and `url` are dead, and the last update was in early 2019.


vilhelmen:
The site and download url appear to be working fine for me, I think this deletion may have been a bit premature. Given what it does, I'm sure it doesn't need a lot of updating.

Issue #81292, opened 2020-04-24
wineskin-winery needs to be updated to run on 64-bit Mac OSX (Catalina+)


deftdawg:
The current `wineskin-winery` cask points at a binary built in 2012.   

While it starts fine, it is not actually capable of producing 64-bit Wine wrappers required for Windows apps to be able to run on OSX following the move to 64-bit only support with Catalina.

@Gcenx has patched the original wineskin-winery to handle both 32-bit and 64-bit wrappers, my suggestion would be replace this cask with his release of the software found at https://github.com/Gcenx/WineskinServer/releases

miccal:
@deftdawg perhaps it might be best to create a new Cask for this fork of `wineskin-winery`, maybe called `gcenx-wineskin-winery` as per the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md):

> If the result conflicts with the name of an existing Cask, make yours unique by prepending the name of the vendor or developer, followed by a hyphen.

ran-dall:
> @deftdawg perhaps it might be best to create a new Cask for this fork of wineskin-winery, maybe called gcenx-wineskin-winery as per the token reference:
> 
>     If the result conflicts with the name of an existing Cask, make yours unique by prepending the name of the vendor or developer, followed by a hyphen.

@miccal @deftdawg This is correct. Also, `wineskin-winery` has been downloaded just over 300 times in the last 30 days; which overall for HBC isn't a lot; however, it still wouldn't be proper to replace it with an entirely different project. You can look at the case of `chromium` as an example; there are several casks of `chromium` (at least 2 or 3 to my recollection), all with specific names.

Gcenx:
I have a brew tap for installing my project along with PortingKit and wine-crossover.

`brew tap gcenx/wine`
With a cask name of `unofficial-wineskin`

For reference Wineskin Winery from doh123 is no longer updated nor maintained, the last available wine versions are Wine-2.22 and Wine-Staging-2.21.

It also has numerous bugs with GPU detection starting with High Sierra.

tjnycum:
> Also, `wineskin-winery` has been downloaded just over 300 times in the last 30 days; which overall for HBC isn't a lot; however, it still wouldn't be proper to replace it with an entirely different project. You can look at the case of `chromium` as an example; there are several casks of `chromium` (at least 2 or 3 to my recollection), all with specific names.

[...]

PR #81336, opened 2020-04-25
Delete wineskin-winery

miccal:
See discussion in https://github.com/Homebrew/homebrew-cask/issues/81292


ran-dall (approved):
> For reference Wineskin Winery from doh123 is no longer updated nor maintained, the last available wine versions are Wine-2.22 and Wine-Staging-2.21.

Ref: https://github.com/Homebrew/homebrew-cask/issues/81292#issuecomment-619310521

PR #81398, opened 2020-04-26
Update tor-browser.rb

core-code:
mistakenly spelling the countrycode in lowercase creates an interesting bug here. trying to install on spanish systems will install the argentianian version and trying to install in taiwan will install the chinese version:

```
% brew cask fetch  --language=zh-TW tor-browser

==> Downloading external files for Cask tor-browser
==> Downloading https://dist.torproject.org/torbrowser/9.0.9/TorBrowser-9.0.9-osx64_zh-CN.dmg

% brew cask fetch  --language=es-ES tor-browser

==> Downloading external files for Cask tor-browser
==> Downloading https://dist.torproject.org/torbrowser/9.0.9/TorBrowser-9.0.9-osx64_es-AR.dmg
```

i think what is going on here is that if the country-code is in lowercase, it won't match and then it will fallback to the uppermost version where at least the language matches.

### DIFF
--- a/Casks/tor-browser.rb
+++ b/Casks/tor-browser.rb
@@ -36,12 +36,12 @@ cask 'tor-browser' do
     'en-US'
   end
 
-  language 'es-ar' do
+  language 'es-AR' do
     sha256 '69c6050f2b5ac00e3ce5a981e99d72b3479b0d90f9c606b415b1b66da6fa1658'
     'es-AR'
   end
 
-  language 'es-es' do
+  language 'es-ES' do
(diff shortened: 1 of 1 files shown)


core-code:
@reitermarkus i wonder if failing to match the country-code if it is spelled in lowercase in the caskfile should be fixed inside Homebrew itself?

core-code:
well, or 'audit' could check for the country-code being uppercase...

vitorgalvao:
Created https://github.com/Homebrew/homebrew-cask/issues/81802 to track the problem, so we can merge this.

PR #81450, opened 2020-04-27
Delete iphoney.rb

suschizu:
I'm not sure - but I think this app doesn't exist any more - url and appcast are not reachable anymore and I just found this quite old (and beta) project with the same BID:
https://sourceforge.net/projects/iphonesimulator/files/


vitorgalvao:
`brew cask fetch iphoney` works for me, and it seems to be the app. But a quick search hasn’t turn it up on the website and it isn’t very popular. Likely discontinued.

```
iphoney
30 days: 4 (#4349)
90 days: 9 (#4553)
365 days: 27 (#4860)
Age: 2374 days (added 2013, November 01)
```

PR #81491, opened 2020-04-27
Update gitfiend from 0.22.8 to 0.23.1

vladimyr:
Add `gitfiend` binary & shimscript.

### DIFF
--- a/Casks/gitfiend.rb
+++ b/Casks/gitfiend.rb
@@ -1,6 +1,6 @@
 cask 'gitfiend' do
-  version '0.22.8'
-  sha256 '8d0eaaf47074c28d97f8fca7780e11f868adb8ab97f18c8e196f54817a86237d'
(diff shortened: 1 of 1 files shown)


vladimyr:
In response to https://github.com/Homebrew/homebrew-cask/pull/81146#issuecomment-620324182

I've altered the original script to run electron as node because upcoming versions of macOS aren't gonna ship perl/python by default.

vitorgalvao:
> I've altered the original script to run electron as node because upcoming versions of macOS aren't gonna ship perl/python by default.

Why is that relevant in this case, if it’s a nodejs app?

Also, while this one is fine, changes in that regard are premature, as we still have no clue what that deprecation will entail.

vladimyr:
Release v0.23 fixed CLI invocation so no more hacks needed. 🤝

PR #81493, opened 2020-04-27
Update jd-gui

vladimyr:
### ⚠️ **DO NOT MERGE**

Add `jd-gui` binary & shimscript.

Depends on https://github.com/java-decompiler/jd-gui/pull/304

### DIFF
--- a/Casks/jd-gui.rb
+++ b/Casks/jd-gui.rb
@@ -9,6 +9,16 @@ cask 'jd-gui' do
   homepage 'http://jd.benow.ca/'
 
   app "jd-gui-osx-#{version}/JD-GUI.app"
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/jd-gui.wrapper.sh"
+  binary shimscript, target: 'jd-gui'
+
+  preflight do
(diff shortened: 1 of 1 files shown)


core-code:
whats the plan here? :)

vladimyr:
> whats the plan here? :)

Why did you assume there is a plan behind my stuff? 🤣

I was trying to achieve the following. I'm using JD-GUI frequently and what I'm forced to do every time is open it and then navigate to jar archive or drag n drop it inside it. Instead, I wanted to simply invoke it inside the terminal and pass target jar but even though there is support for command-line invocation parameters aren't correctly passed. So I started the chain of events:
1. Raised PR on upstream to fix parameter passing
2. Prepared HBC PR that adds binary/shim script

As you can see former is still without any reaction from upstream so the latter also hangs becoming _sticky_ part of our PR list...
Frankly, I can either close this ATM or we can continue waiting together for upstream to consider my PR. So no plans, but let's hear your opinion/whishes 😉

vladimyr:
Also, note that the last commit to the upstream master is 6mo old. :shrug:

core-code:
ah ok, thanks for the explanation here :)

i'd vote closing this now - we can reopen it anyway iff upstream reacts.

PR #81545, opened 2020-04-28
FAQ: why some apps are not included in `upgrade`

vitorgalvao:


core-code:
> If you then `upgrade`d through Homebrew Cask and we have a lower version on the cask, you’d get a downgrade.

isn't that pretty hypothetical given how good the state of HBC is these days thanks to yourself and all the contributors?

ran-dall:
While we’re on the upgrade subject, the ML has been increasing accurate on downgrades. You know the type, updates that are done one day, only to be pulled back the next. Right now, it just spits it back to me, and I report it; would it be helpful for it to submits the downgrades automatically?

doc/faq/app_not_upgrading.md
```diff
@@ -0,0 +1,27 @@
...
+
+That could cause conflicts when used in tandem with Homebrew Cask’s `upgrade` mechanism.
+
+If you upgrade software through it’s built-in mechanism, that happens without Homebrew Cask’s knowledge so both versions get out of sync. If you then upgraded through Homebrew Cask and we have a lower version on the software on record, you’d get a downgrade.
+
+There are a few ideas to fix this problem:
+
+* Try to prevent the software’s automated updates. That won’t be a universal solution and may cause it to break. Most software on Homebrew Cask is closed-source, so we’d be guessing. This is also why pinning casks to a version isn’t available.
+* Try to extract the installed software’s version and compare it to the cask, deciding what to do at that time. That’s a complicated solution that breaks other parts of our methodology, such as using versions to interpolate in `url`s (a definite win for maintainability). That solution also isn’t universal, as many software developers are inconsistent in their versioning schemes (and app bundles are meant to have two) and it doesn’t work for all types of software we support.
+
```
miccal, line 17: `(and app bundles are meant to have two)` -- app bundles are mean to have two what, exactly?

vitorgalvao:
> isn't that pretty hypothetical given how good the state of HBC is these days

The sheer amount of users we have makes it a possibility, I’d say. There will always be someone to get caught in the update window. Though I get the point that these days said window may be so small as to be mostly a non-issue.

However, some apps allow switching the upgrade channel (stable, pre-release, beta) *only* from inside the app. Alfred does it, and I think Jetbrains apps do as well. Since the main repo is for stable versions only, users of those (insanely popular) apps would be seesawing between stable and not all the time.

@ran-dall As long as what we have on the main repo is the official stable version, however you choose to do it is up to you. As long as you’re reasonably sure that a certain version is correct, that should trump other checks (e.g. if we see an appcast is wrong according to the website, we follow the website).

doc/faq/app_not_upgrading.md
vitorgalvao: > app bundles are mean to have two what, exactly?

Version strings. [`CFBundleVersion`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) and [`CFBundleShortVersionString`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring).

In the wild, sometime they’re the same, other times they’re different, other times one is a shorter version of the other, other times only one is present. It’s a mess.

miccal: May I suggest the following change:

(and app bundles are meant to have two version strings; namely a [CFBundleVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) and a [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring))

vitorgalvao:
Merging for now, but it’s always open to changes.

PR #55213, opened 2018-11-20
Adding drjava latest

darkmastermindz:
**adding a new cask**:

- [ ] Named the cask according to the [token reference].
- [x] `brew cask install drjava` worked successfully.
- [x] `brew cask uninstall drjava` worked successfully.
- [x] Checked there are no [open pull requests] for the same cask.
- [x] Checked the cask was not [already refused].
- [x] Checked the cask is submitted to {{correct repo}}.

The token reference said drjavajar but it doesn't make sense since drjava is the official name of the IDE.

Issue #62715, opened 2019-05-02
A review of `binary`-only casks


vitorgalvao:
This is a continuation of https://github.com/Homebrew/brew/pull/5724/files#r256898140.

In that thread, to advance the goal of lessening the confusion users have between `brew` and `brew cask`, the following rule was suggested:

> “if [you] want the CLI, `brew`, if you want the GUI, `brew cask` but it would also bring the CLI that typically comes in the package”

And

> To push the point further, maybe CLI-only casks (closed-source) should be removed as well.

For that second point it was suggested we look at the analytics, which is what this issue is about. The rule for inclusion was casks that include `binary` but not `pkg`, `app`, or `suite`. Some of these are open-source but can’t be built (or would be too complex) as formulae.

They are ordered by number of downloads in the last 365 days. Of the casks in the table, only 2 are not in the main repo: [nordic-nrf5x-command-line-tools](https://github.com/Homebrew/homebrew-cask-drivers/blob/master/Casks/nordic-nrf5x-command-line-tools.rb) and [chromedriver-beta](https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/chromedriver-beta.rb).

It’s clear from these results we can’t remove `binary`-only casks without upsetting thousands of users, so that’s a no-go for the time being. However, we should continue to think of ways to make the `brew`/`brew cask` divide clearer.

| cask | 30d | 90d | 365d |
|------|:---:|:---:|:----:|
| [minikube](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/minikube.rb) | 13,958 | 43,475 | 107,029 |
| [android-platform-tools](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/android-platform-tools.rb) | 13,251 | 36,055 | 94,566 |
| [chromedriver](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/chromedriver.rb) | 8,499 | 24,223 | 63,225 |
| [android-sdk](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/android-sdk.rb) | 5,061 | 16,072 | 44,099 |
| [google-cloud-sdk](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/google-cloud-sdk.rb) | 3,893 | 11,199 | 28,305 |
| [ngrok](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/ngrok.rb) | 3,279 | 10,082 | 26,369 |
[...]

PR #81550, opened 2020-04-28
Add DrJava 20190813-220051 (beta)

vladimyr:
This is a continuation of #55213

### DIFF
--- a/Casks/drjava.rb
+++ b/Casks/drjava.rb
@@ -1,0 +1,31 @@
+cask 'drjava' do
+  version '20190813-220051,beta'
+  sha256 '69bdccf632ca8396876bd61cd362027d9ec90ea0a4e8b4edadec438d5b253ff9'
+
+  # downloads.sourceforge.net/drjava/ was verified as official when first introduced to the cask
+  url "https://downloads.sourceforge.net/drjava/drjava-#{version.after_comma}-#{version.before_comma}.jar"
+  appcast 'https://sourceforge.net/projects/drjava/rss?path=/1.%20DrJava%20Stable%20Releases'
+  name 'DrJava'
+  homepage 'http://www.drjava.org/'
+
+  container type: :naked
+
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/drjava.wrapper.sh"
+  binary shimscript, target: 'drjava'
+
+  preflight do
+    IO.write shimscript, <<~EOS
+      #!/bin/sh
+      cd "$(dirname "$(readlink -n "$0")")" && \
+        nohup java "$@" -jar drjava-#{version.after_comma}-#{version.before_comma}.jar \
+        >/dev/null 2>&1 &
+    EOS
+  end
+
+  zap trash: '~/.drjava'
+
+  caveats do
+    depends_on_java '8'
+  end
+end


ran-dall:
@vladimyr Good way of fixing this. This recently got updated, however, I still don't know if it's worth the being added. Do you personally use this?

vladimyr:
No, I don't it is really just an educational tool not something you would use professionally. And that is the primary reason why I proposed it. Serving the academic community is something I enjoy doing. We don't have to include it here, closed PR is also a good thing because those who need it can simply c/p into their personal taps or install using direct URL, it is a win-win situation 😉

ran-dall:
@vladimyr As much as we all like to help (and participate in) the academic communities, I don't know that it makes any difference if it's an educational tool or not. The main concern I have is adding something that later breaks and turns into a burden. This is why it's prefered you only add casks that you personally use.

@vitorgalvao What are your feelings on this cask? Are we still on the position that it's not a good fit for HBC? It's been recently updated and the newest version was downloaded a few thousand times from SourceForge. Maybe we can provisionally merge it? Then if it doesn't get any traction, we can remove it in one of our analytics sweeps?

vitorgalvao:
> closed PR is also a good thing because those who need it can simply c/p into their personal taps or install using direct URL, it is a win-win situation

Thumbs up to that.

> @vitorgalvao What are your feelings on this cask?

Negative, because it’s java-related and those casks are some of the biggest headaches we have. That said, it’s not a development kit, so not that big of a deal.

But it is essentially shim-only, which is uncommon (do we have *any* other cask like that?). Considering the shim technique is itself a bit of a hack (ideally we’d get a proper interface for it), it doesn’t feel appropriate.

And it’s a GUI that we need to call via the CLI? That feels weird and incomplete.

I’m for refusal.

All that said,  you’ve handled https://github.com/Homebrew/homebrew-cask/pull/55213 ([my interaction with that issue](https://github.com/Homebrew/homebrew-cask/pull/55213#issuecomment-445501828) followed your suggestion), so I’ll defer to you, as you’re more familiar with it.

Keep in mind that the analytics sweep—by design—only considers casks that have been present for at least 365 days, and I’m considering changing the removal threshold so that less casks get caught in it.

ran-dall:
> Negative, because it’s java-related and those casks are some of the biggest headaches we have. That said, it’s not a development kit, so not that big of a deal.

Agreed. Java on HBC has been a headache.

> But it is essentially shim-only, which is uncommon (do we have any other cask like that?).

I haven't seen anything like this to date.

> Considering the shim technique is itself a bit of a hack (ideally we’d get a proper interface for it), it doesn’t feel appropriate.

Agreed.

> And it’s a GUI that we need to call via the CLI? That feels weird and incomplete.

This threw me for a loop when I initially encountered it.

> I’m for refusal.

Considering everything that has happened around this cask, Java, and the fact that there aren't any maintainers that use this, we're in agreement for refusal.

> Keep in mind that the analytics sweep—by design—only considers casks that have been present for at least 365 days, and I’m considering changing the removal threshold so that less casks get caught in it.

@vitorgalvao So does this mean removing the 365-day threshold? or what threshold are we talking about?

---

@vladimyr I think we have a consensus to keep this cask out of HBC. However, you can still put this on a personal tap (or I can put it on mine; which I've been working with the ML to auto-maintain); therefore it's still available for the folks that need it in whatever situation they find themselves in.

vitorgalvao:
> So does this mean removing the 365-day threshold? or what threshold are we talking about?

Rereading, I see I should’ve been more specific. I mean the threshold for the minimum number of downloads for removal, so that less casks get removed per sweep.

ran-dall:
@vitorgalvao Thanks for the clarification. If you don't mind me asking, what's the current threshold now and what are we moving to?

vitorgalvao:
> what's the current threshold now and what are we moving to?

We use more than one source for it, so it’s possible (and it does happen, I’ve seen it) that a cask with zero downloads on HBC is not deleted because it has more on the other source.

I’ll send you the exact numbers privately. I don’t want to share them publicly until we settle on something (I still consider these tests), to avoid someone gaming the system.

darkmastermindz:
@ran-dall @vitorgalvao Would be happy to help host / maintain this cask and any other future educational tools for programming you may come across. Dr. Java is popularly used as an educational interactive java development environment in AP CS courses and higher education found in a lot of intro to java textbooks. It's is a standalone portable jar so I don't foresee issues with breakages.

ran-dall:
@darkmastermindz As I stated before, I don't think that this being an education tool really makes a difference as to what gets merged into HBC's main tap. 

That said, I'll go ahead and merge this into my tap, which I'm working to auto-maintain itself with the ML.

darkmastermindz:
@ran-dall Thanks for providing a solution and what I was trying to get at is that education tools / SDKs like these are commonly used and downloaded frequently metrics-wise. Having them available provides exposure to using / contributing to HB / HBC in academic communities.

I understand the refusal @vitorgalvao to move to a cleaner divide between cask and non-cask as a similar concern is listed in: #62715. As far as I can tell, this package is similar to the case with [mars](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/mars.rb) which is typically also used as an education tool for learning [MIPS asm language](http://courses.missouristate.edu/KenVollmar/MARS/) which is currently in homebrew-cask.

ran-dall:
@darkmastermindz The problem isn't simply notability. There are other valid concerns which is why it was refused. I think keeping this on my tap is the best option ATM.

PR #81618, opened 2020-04-29
Add chef-workstation @ 0.17.5

marcparadise:
Chef Workstation is the successor to ChefDK, which was previously in homebrew-cask.  It was removed because it was open source, cli-only [1]  Since that time, we have added a GUI component "Chef Workstation App" to Chef Workstation. 

[1] #47905

Signed-off-by: Marc A. Paradise <marc@chef.io>

---

- [x] Checked the cask was not [already refused]. (It was previously removed, see note above)

### DIFF
--- a/Casks/chef-workstation.rb
+++ b/Casks/chef-workstation.rb
@@ -1,0 +1,45 @@
+cask 'chef-workstation' do
+  version '0.17.5'
+  sha256 'e9b9165a80222900a1e421f7ad72d2f8252bc0f46175162e43b8cf271b2d2207'
+
+  url "https://packages.chef.io/files/stable/chef-workstation/#{version}/mac_os_x/10.13/chef-workstation-#{version}-1.dmg"
+  name 'Chef Workstation'
+  homepage 'https://docs.chef.io/workstation/'
+
+  depends_on macos: '>= :high_sierra'
+
+  pkg "chef-workstation-#{version}-1.pkg"
+
+  uninstall quit:      'sh.chef.chef-workstation',
+            pkgutil:   'com.getchef.pkg.chef-workstation',
+            launchctl: 'io.chef.chef-workstation.app',
+            delete:    [
+                         '/Applications/Chef Workstation App.app',
+                         '/opt/chef-workstation/',
+                         '/usr/local/bin/berks',
+                         '/usr/local/bin/chef',
+                         '/usr/local/bin/chef-cli',
+                         '/usr/local/bin/chef-analyze',
+                         '/usr/local/bin/chef-apply',
+                         '/usr/local/bin/chef-client',
+                         '/usr/local/bin/chef-run',
+                         '/usr/local/bin/chef-shell',
+                         '/usr/local/bin/chef-solo',
+                         '/usr/local/bin/chef-vault',
+                         '/usr/local/bin/chefx',
+                         '/usr/local/bin/cookstyle',
+                         '/usr/local/bin/dco',
+                         '/usr/local/bin/delivery',
+                         '/usr/local/bin/foodcritic',
+                         '/usr/local/bin/inspec',
+                         '/usr/local/bin/kitchen',
+                         '/usr/local/bin/knife',
+                         '/usr/local/bin/ohai',
+                         '/usr/local/bin/push-apply',
+                         '/usr/local/bin/pushy-client',
+                         '/usr/local/bin/pushy-service-manager',
+                         '/usr/local/bin/uninstall_chef_workstation',
+                       ]
+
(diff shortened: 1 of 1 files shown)


Casks/chef-workstation.rb
```diff
@@ -0,0 +1,45 @@
...
+  homepage 'https://docs.chef.io/workstation/'
+
+  depends_on macos: '>= :high_sierra'
+
+  pkg "chef-workstation-#{version}-1.pkg"
+
+  uninstall quit:      'sh.chef.chef-workstation',
+            pkgutil:   'com.getchef.pkg.chef-workstation',
+            launchctl: 'io.chef.chef-workstation.app',
+            delete:    [
```
vitorgalvao, line 16: Do you really need all this? Doesn’t the the `pkgutil` take care of it?

vitorgalvao:
Please add an [`appcast`](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

Casks/chef-workstation.rb
marcparadise: Looking into this, it was carried over from the definition we had in our own tap.

Casks/chef-workstation.rb
```diff
@@ -0,0 +1,45 @@
+cask 'chef-workstation' do
+  version '0.17.5'
+  sha256 'e9b9165a80222900a1e421f7ad72d2f8252bc0f46175162e43b8cf271b2d2207'
+
+  url "https://packages.chef.io/files/stable/chef-workstation/#{version}/mac_os_x/10.13/chef-workstation-#{version}-1.dmg"
```
vitorgalvao, line 5: Make the `-1` part of the `version`.

Casks/chef-workstation.rb
tyler-ball: Our package setup does not currently delete these files. I can open an issue on our side to change that, but in the meantime, can we leave this `delete` stanza here? Will it fail if it cannot find these files/folders/symlinks to delete?

Casks/chef-workstation.rb
```diff
@@ -0,0 +1,45 @@
...
+  name 'Chef Workstation'
```
tyler-ball, line 6:
```suggestion
  url "https://packages.chef.io/files/stable/chef-workstation/#{version}/mac_os_x/10.13/chef-workstation-#{version}-1.dmg"
  appcast 'https://omnitruck.chef.io/stable/chef-workstation/metadata?p=mac_os_x&pv=10.14&m=x86_64&v=latest'
  name 'Chef Workstation'
```

Fetching the metadata from omnitruck with `&v=latest` will change whenever we do a release. The contents look like:
```
sha1	b5e1cbc4ee79190a4266316d195e06df0e4b9196
sha256	e9b9165a80222900a1e421f7ad72d2f8252bc0f46175162e43b8cf271b2d2207
url	https://packages.chef.io/files/stable/chef-workstation/0.17.5/mac_os_x/10.14/chef-workstation-0.17.5-1.dmg
version	0.17.5
```

Casks/chef-workstation.rb
marcparadise: Including the `-1` in the version would cause cause the URL to be invalid, because of `chef-workstation/#{version}` , which needs to resolve without the -1.

Casks/chef-workstation.rb
vitorgalvao: > Our package setup does not currently delete these files.

Nor should it. I think there’s some confusion regarding what `uninstall pkgutil:` does. It’s faster this way: remove the `delete` stanza and see if CI still passes.

Casks/chef-workstation.rb
vitorgalvao: Please make the change, @marcparadise.

Casks/chef-workstation.rb
marcparadise: I think instead of `delete`, we could include a `script` stanza: 
```
  uninstall script:    {
     executable: '/usr/local/bin/uninstall_chef_workstation',
     sudo: true
 }
```
This is a script that we include in the installed package, its content can be found [here](https://github.com/chef/chef-workstation/blob/master/omnibus/files/uninstall-scripts/uninstall_chef_workstation).

It does the same cleanup as this, though it does so by removing all symlinks from /usr/ocal/bin that point to /opt/chef-workstation. (Everything the installer puts in /usr/local/bin is a symlink)

marcparadise:
Pushed up changes to add the `appcast` and change `uninstall delete` to `uninstall script`; it does the same cleanup  using a cleanup script we ship with the application itself.

PR #81801, opened 2020-05-02
Update ipsecuritas.rb from 4.9.5 to 4.9.6

suschizu:
update


suschizu:
```
audit for ipsecuritas: warning
 - appcast at URL 'https://www.lobotomo.com/products/IPSecuritas/' does not contain the version number '4.9.1':
```

not true

core-code:
well it is true, it doesn't contain 4.9.1....since 4.9.6 is the latest version

PR #81921, opened 2020-05-05
Update inkscape from 1.0.0rc1 to 1.0.0

ashwinreddy:


vitorgalvao (changes requested):
According to the page, it seems this version is supported on older macOS versions, so please remove the conditional and leave the cask only at `1.0.0`.

PR #81930, opened 2020-05-05
Update pycharm `zap` stanza

singularitti:


singularitti:
The file in [`webstorm.rb`](https://github.com/MineralsCloud/qha/blob/3605a77omebrew-cask/Casks/webstorm.rb#L23-L24) is wrong either. Its number needs to be changed on every version.

PR #81965, opened 2020-05-05
Update eudic.rb from '2020-01-27,3.9.6' to '2020-04-05,3.9.6'

suschizu:
update


core-code:
i think we might need an user_agent or possibly referer.

core-code:
try

```
  url "https://static.frdic.com/pkg/eudicmac.dmg?v=#{version.before_comma}",
      user_agent: :fake,
```

PR #81922, opened 2020-05-05
inkscape.rb: keep only 1.0.0

vitorgalvao:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

According to the website, version `1.0.0` is supported from macOS `10.11`.

PR #81973, opened 2020-05-05
Update inkscape.rb

core-code:
the sole purpose of the if/else has been not to update to a beta version for those users (non catalina) where a stable one exists.

since 1.0.0 is officially released in a stable form there is no need for that anymore


miccal:
https://github.com/Homebrew/homebrew-cask/pull/81922

core-code:
ouch, thanks ;)

PR #82077, opened 2020-05-06
Add Codespace Latest

Crysicia:

### DIFF
--- a/Casks/codespace.rb
+++ b/Casks/codespace.rb
@@ -3,7 +3,7 @@ cask 'codespace' do
   sha256 '34a25da2b6034571a93ae278e41ccb4a94e6cf1c523567a4ba61dbba1e9ff001'
 
   url 'https://codespace.app/download/latest'
-  appcast 'https://codespace.app/releases'
+  appcast 'https://macupdater.net/cgi-bin/check_urls/check_url_filename.cgi?url=https://codespace.app/download/latest'
   name 'Codespace'
   homepage 'https://codespace.app/'
 
--- a/Casks/codespace.rb
+++ b/Casks/codespace.rb
@@ -1,6 +1,6 @@
 cask 'codespace' do
-  version :latest
-  sha256 :no_check
+  version '1.1.2'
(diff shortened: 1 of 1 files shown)


Casks/codespace.rb
```diff
@@ -0,0 +1,10 @@
+cask 'codespace' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://codespace.app/download/latest'
```
vitorgalvao, line 5: Please add an [`appcast`](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md). `https://macupdater.net/cgi-bin/check_urls/check_url_filename.cgi?url=https://codespace.app/download/latest` if you can’t find anything else.

Casks/codespace.rb
```diff
@@ -0,0 +1,10 @@
+cask 'codespace' do
+  version :latest
```
vitorgalvao, line 2: Add a `version` and `sha256`.

vitorgalvao (changes requested):

Casks/codespace.rb
Crysicia: Done!
I've used the [releases](https://codespace.app/releases) page, is that ok ?

Casks/codespace.rb
Crysicia: Unfortunately only the latest version of Codespace is available for download, I'm afraid I cannot give a version number.

Casks/codespace.rb
vitorgalvao: Use the one I gave. It’s better because it only has one number.

Casks/codespace.rb
vitorgalvao: You can. Make it `1.1.2`. There’s an appcast, so the `url` doesn’t need to have a version in it (though that would be preferable).

vitorgalvao (approved):
Thank you for the fast turnaround. Welcome!

Crysicia:
Thanks a lot for your help, first submission here :)

PR #82192, opened 2020-05-08
Update flume

core-code:


core-code:
i don't usually advocate using the CGI for the appcast if there is *any* other choice but the previous appcast is over one megabyte

PR #82283, opened 2020-05-09
MacTex 2020.0407 - Making it clear you need to restart your terminal to use

cpaika:
MacTex registers itself in the PATH via writing to  /etc/paths.d .  This file is only read on terminal startup, so once MaxTex is installed, I'm adding a message telling users to restart their terminal so they can use the CLI tools.

### DIFF
--- a/Casks/mactex.rb
+++ b/Casks/mactex.rb
@@ -77,6 +77,7 @@ cask 'mactex' do
              ]
 
   caveats do
-    'You must restart your terminal window for the installation of MacTex CLI tools to take effect.'
+    'You must restart your terminal window for the installation of MacTex CLI tools to take effect.  Bash or zsh users can run the command:
+      eval "$(/usr/libexec/path_helper)"'
   end
 end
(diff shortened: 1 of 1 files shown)


cpaika:
Pushed a fix for the style problem

vitorgalvao:
You don’t need to restart your terminal. For bash and zsh (and perhaps others) `eval "$(/usr/libexec/path_helper)"` suffices.

cpaika:
@vitorgalvao Nice Is there a pattern for adding that to the cask install automatically, or would you rather I modify the instructions to include that instead of a restart?

vitorgalvao:
Include both, and specify the command is for bash/zsh users.

cpaika:
@vitorgalvao Thanks, fixed let me know what you think

Casks/mactex.rb
```diff
@@ -75,4 +75,11 @@
                '/usr/local/texlive',
                '~/Library/texlive',
              ]
+
+  caveats <<~EOS
+    You must restart your terminal window for the installation of MacTex CLI tools to take effect.
```
vitorgalvao, line 80: Separated into two paragraphs anyway, but we don’t double-space.

PR #82313, opened 2020-05-10
Update deeper.rb from 2.5.5 to 2.5.6

suschizu:
update


suschizu:
` - appcast at URL 'https://www.titanium-software.fr/en/release_deeper.html' does not contain the version number '2.4.8':
`
but contains 2.5.6

PR #82314, opened 2020-05-10
Update maintenance.rb from 2.6.6 to 2.6.7

suschizu:
update


suschizu:
`- appcast at URL 'https://www.titanium-software.fr/en/release_maintenance.html' does not contain the version number '2.5.6'`
but it contains 2.6.7

PR #82341, opened 2020-05-10
Update gloomhaven-helper from 8.3.3 to 8.3.6

samuelmurray:


samuelmurray:
Seems like the current release of 8.3.6 is not stable, as the app crashes by a NPE during startup. Also, the previous version (8.3.3) is removed from the website. I guess we could either downgrade the version to 8.3.1 (the second latest available release), or wait for a new functioning version.

PR #82367, opened 2020-05-11
Update lightproxy from 1.1.13 to 1.1.14

suschizu:


core-code:
that appcast doesn't look too good. since its the GIT version i fear it might have unreleased versions too. but the version number lines up with the actual download so all is good now

PR #82372, opened 2020-05-11
Update bettertouchtool.rb

joaovbmdias:
Proposed release version does not exist in BTT repository.


core-code:
thanks there. but just updating the version without the checksum is wrong. i've sent https://github.com/Homebrew/homebrew-cask/pull/82375

PR #82418, opened 2020-05-12
Update loginputmac from 2.3.5 to 2.3.5_1

R0uter:

### DIFF
--- a/Casks/loginputmac.rb
+++ b/Casks/loginputmac.rb
@@ -1,5 +1,5 @@
 cask 'loginputmac' do
-  version '2.3.5_1'
+  version '2.3.5'
   sha256 '91c5e011792f3ab2ee7f3e69ff7ea3a45c156f02b574cba2c57dbcfec82dd4a6'
 
   # loginput-mac2.content-delivery.top/ was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


Casks/loginputmac.rb
```diff
@@ -1,6 +1,6 @@
 cask 'loginputmac' do
-  version '2.3.5'
-  sha256 '7c6cdd1276f328b049666e9f3e27cca39c2524959ba72c7ffcfcead4bf67ab20'
+  version '2.3.5_1'
```
vitorgalvao, line 2: Where are you getting the `_1` from?

R0uter: Cuz this is an extra update, for an emergency fix that can not wait to next version 2.3.6.

vitorgalvao: Then leave the version be. They’ve updated it in place instead of increasing the version, that’s their choice.

Casks/loginputmac.rb
```diff
@@ -1,5 +1,5 @@
 cask 'loginputmac' do
-  version '2.3.5_1'
+  version '2.3.5'
```
R0uter, line 2: so we can actually release same version again?

Casks/loginputmac.rb
R0uter: Thx man, didnt know that, now learnt

Casks/loginputmac.rb
vitorgalvao: Yes. We don’t need to make up new numbers. If they update in place, so should we.

PR #82432, opened 2020-05-12
PULL_REQUEST_TEMPLATE.md: update with important note

vitorgalvao:
I’m thinking of other ways to improve the template, but this is a first step. I’ll comment on the lines themselves.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,22 +1,16 @@
-<!-- If there’s a checkbox you can’t complete for any reason, that's okay, just explain in detail why you weren’t able to do so. -->
+**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.
 
-After making all changes to the cask:
+After making all changes to a cask, verify:
 
 - [ ] `brew cask audit --download {{cask_file}}` is error-free.
 - [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
-- [ ] The commit message includes the cask’s name and version.
+- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
 - [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).
 
-Additionally, if **adding a new cask**:
+Additionally, **if adding a new cask**:
 
-- [ ] Named the cask according to the [token reference].
+- [ ] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
 - [ ] `brew cask install {{cask_file}}` worked successfully.
 - [ ] `brew cask uninstall {{cask_file}}` worked successfully.
-- [ ] Checked there are no [open pull requests] for the same cask.
-- [ ] Checked the cask was not [already refused].
-- [ ] Checked the cask is submitted to [the correct repo].
-
-[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
-[open pull requests]: https://github.com/Homebrew/homebrew-cask/pulls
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,22 +1,16 @@
-<!-- If there’s a checkbox you can’t complete for any reason, that's okay, just explain in detail why you weren’t able to do so. -->
+**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.
```
vitorgalvao, line 1: It’s clear some people do not heed the comment. By making this a top item, making a direct appeal to honesty, and removing the need for explaining why an item wasn’t done, I expect compliance to improve.

I also expect this to require some revision, after we see some results (or lack thereof).

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,22 +1,16 @@
...
 
-After making all changes to the cask:
+After making all changes to a cask, verify:
 
 - [ ] `brew cask audit --download {{cask_file}}` is error-free.
 - [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
-- [ ] The commit message includes the cask’s name and version.
```
vitorgalvao, line 7: Is this necessary? I expect the name and version to be included naturally as part of the regular PR process, but even when it doesn’t, it won’t be a big deal.

In addition, with so many submissions being done directly with `cask-repair`, this is mostly covered.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,22 +1,16 @@
...
+- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
```
vitorgalvao, line 7: This used to be down on the section of adding a new cask. Realistically, multiple PRs for the same new cask are a rarity, and I doubt the cause was this checkbox.

But multiple concurrent updates to the same cask and version are common, so we want to avoid those.

I understand this can still happen, particularly when maintainers use automated submission systems, but this miht work for manual submissions o contributors.

In this case, `cask-repair`’s popularity might prove problematic (how many people check before the submission?) but no harm in testing.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,22 +1,16 @@
...
 - [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).
```
miccal, line 8: Suggest:
```
[a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions)
```
since `https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask` is referenced below in the last checkbox.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,22 +1,16 @@
-<!-- If there’s a checkbox you can’t complete for any reason, that's okay, just explain in detail why you weren’t able to do so. -->
+**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.
 
-After making all changes to the cask:
+After making all changes to a cask, verify:
 
 - [ ] `brew cask audit --download {{cask_file}}` is error-free.
```
miccal, line 5: Why `{{cask_file}}`, and not just `{cask_file}`?

.github/PULL_REQUEST_TEMPLATE.md
vitorgalvao: Agreed. Changed.

.github/PULL_REQUEST_TEMPLATE.md
vitorgalvao: Because only one set is too close to `${cask_file}`, which we use in other places in the documentation. I’m also thinking of `<cask file>`, as it nicely maps to CLI instructions.

However, this is the method we use in all documentation right now, so changing that is a matter for another PR.

miccal: Thanks for the explanation :)

PR #82490, opened 2020-05-13
Update zoho-docs

core-code:


core-code:
we have a old download URL which is stuck at 1.9.50 while they are up to 19.54 now. 

the new URL looks awful though if it expires or changes constantly we may have to remove this cask

PR #82659, opened 2020-05-15
Update mmex

suschizu:

### DIFF
--- a/Casks/mmex.rb
+++ b/Casks/mmex.rb
@@ -1,9 +1,9 @@
 cask 'mmex' do
   version '1.3.4'
-  sha256 '3d094573241527ab8c578c4dd879373df5ff662d528cb329ab74f36f4836efe2'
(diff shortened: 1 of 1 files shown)


core-code:
@guanlisheng can you explain the difference between the 'Darwin' and 'macos10.9-wx3.0.4' version? thanks!

guanlisheng:
hey @core-code, 
'Darwin' is just a default value to represent it's mac os while macos10.9-wx3.0.4 is more meanful.

- 10.9 means its minimal mac os version would be running
- wx3.0.4 means it's based on wxWidgets v3.0.4. also we have some distributios based on wxWidgets 3.0.5 or 3.1.x

core-code:
thanks there! so, which is the default version for most users that should be available in Homebrew Cask? :)

core-code:
merging now because the old URL is dead, we can always change to the other version later...

PR #82697, opened 2020-05-16
Update wechat from 2.4.1.17 to 2.4.1.18

b0ttle:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] The commit message includes the cask’s name and version.

PR #82711, opened 2020-05-16
Update wechat from 2.4.1.17 to 2.4.1.18

YorhaL:

### DIFF
--- a/Casks/wechat.rb
+++ b/Casks/wechat.rb
@@ -1,6 +1,6 @@
 cask 'wechat' do
-  version '2.4.1.17'
-  sha256 '4b23e9e78e959f279af0328686ca6c0a9b47f01be6f713b21ae63d111fe44f96'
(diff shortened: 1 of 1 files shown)


miccal:
https://github.com/Homebrew/homebrew-cask/pull/82697

alswl:
It seems CDN update latency.

core-code:
i've thought about using language/region stanzas to solve this issue, but Travis would still fail.

since this is a static download URL maybe using version :latest is the only workable thing here?

vitorgalvao:
> since this is a static download URL maybe using version :latest is the only workable thing here?

That would require removing the appcast as well. But this problem always happens with this cask and it’s updated a lot, so I’m not necessarily opposed.

Closing in favour of https://github.com/Homebrew/homebrew-cask/pull/82697.

PR #82816, opened 2020-05-18
Added Glance.app v1.1

crunchtime-ali:
Adds [Glance](https://github.com/samuelmeuli/glance) which is an All-In-One Quick Look plugin that uses the new Quick Look API. Since a lot of the Quick Look plugins use the outdated API and are not sandboxed `Glance` is gaining quite some traction.

The caveat that you have to start the app once manually to set up the Quick look plugin is a request by the author:

@samuelmeuli writes in https://github.com/samuelmeuli/glance/issues/11#issuecomment-630335561 
> I think the user should open the app by themselves. Automatically executing a binary that was downloaded from the internet seems a bit problematic to me. And the user might want to verify that the app is sandboxed before opening it. Maybe a note in caveats is enough?

### DIFF
--- a/Casks/glance.rb
+++ b/Casks/glance.rb
@@ -1,6 +1,6 @@
 cask 'glance' do
-  version '1.0.0'
-  sha256 '8ef7ba4a5130955e213fdfcc2bf43cba47c742d5f02e0e9c62405acaf19626b5'
+  version '1.1.0'
+  sha256 'bd14584dd7f81fa13217e02f6d6cbdc24344f553059abc7dc2d037e54680fbdb'
 
   url "https://github.com/samuelmeuli/glance/releases/download/v#{version}/Glance.dmg"
   appcast 'https://github.com/samuelmeuli/glance/releases.atom'
--- a/Casks/glance.rb
+++ b/Casks/glance.rb
@@ -1,0 +1,25 @@
+cask 'glance' do
+  version '1.0.0'
+  sha256 '8ef7ba4a5130955e213fdfcc2bf43cba47c742d5f02e0e9c62405acaf19626b5'
+
+  url "https://github.com/samuelmeuli/glance/releases/download/v#{version}/Glance.dmg"
+  appcast 'https://github.com/samuelmeuli/glance/releases.atom'
+  name 'Glance'
+  homepage 'https://github.com/samuelmeuli/glance'
+
+  depends_on macos: '>= :catalina'
+
+  app 'Glance.app'
+
+  zap trash: [
+               '~/Library/Application\ Scripts/com.samuelmeuli.Glance/',
+               '~/Library/Application\ Scripts/com.samuelmeuli.Glance.QLPlugin/',
+               '~/Library/Containers/com.samuelmeuli.Glance/',
+               '~/Library/Containers/com.samuelmeuli.Glance.QLPlugin/',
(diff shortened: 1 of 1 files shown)


Casks/glance.rb
```diff
@@ -0,0 +1,25 @@
...
+  appcast 'https://github.com/samuelmeuli/glance/releases.atom'
+  name 'Glance'
+  homepage 'https://github.com/samuelmeuli/glance'
+
+  depends_on macos: '>= :catalina'
+
+  app 'Glance.app'
+
+  zap trash: [
+               '~/Library/Application\ Scripts/com.samuelmeuli.Glance/',
```
vitorgalvao, line 15: Remove the backslashes. They’re inside quotes, meaning they’re interpreted literally.

vitorgalvao: Also, remove the trailing slashes.

Casks/glance.rb
```diff
@@ -0,0 +1,25 @@
...
+               '~/Library/Application\ Scripts/com.samuelmeuli.Glance.QLPlugin/',
+               '~/Library/Containers/com.samuelmeuli.Glance/',
+               '~/Library/Containers/com.samuelmeuli.Glance.QLPlugin/',
+               '~/Library/Group\ Containers/group.com.samuelmeuli.glance/',
+             ]
+
+  caveats <<~EOS
+    You must start #{appdir}/Glance.app once manually to setup the QuickLook plugin.
```
vitorgalvao, line 23: Where is the quicklook plugin after one opens the app? It should be removed on `uninstall`.

vitorgalvao (changes requested):

Casks/glance.rb
crunchtime-ali: Done @vitorgalvao

Casks/glance.rb
crunchtime-ali: As I understand it the Quick Look plugin resides in the app bundle at: /Applications/Glance.app/Contents/PlugIns/QLPlugin.appex

vitorgalvao: And does it stay there after it’s registered by the app?

vitorgalvao:
@samuelmeuli Can’t there be a way to install the quicklook plugin without users having to manually install the app?

Homebrew Cask users expect to not have to do manual tasks to finish the installation after installing a cask. I’ve read [your concerns](https://github.com/samuelmeuli/glance/issues/11#issuecomment-630335561), but they don’t apply—we already install `pkg`s without interaction, so this would be no different.

Casks/glance.rb
samuelmeuli: Yes, removing the app should be enough.

samuelmeuli:
> Can’t there be a way to install the quicklook plugin without users having to manually install the app?

I don't think so. The container app is required [according to this Stack Overflow answer](https://stackoverflow.com/a/39628355/6767508).

> Homebrew Cask users expect to not have to do manual tasks to finish the installation after installing a cask. I’ve read your concerns, but they don’t apply—we already install pkgs without interaction, so this would be no different.

Fair enough :)

Casks/glance.rb
crunchtime-ali: @vitorgalvao Yes, it does.

vitorgalvao:
Thank you for the answers.

PR #82819, opened 2020-05-18
Update amazon-workspaces from 2.5.11 to 3.0.6

core-code:


core-code:
it seems 3.0 is now finally stable+official, see here: https://docs.aws.amazon.com/workspaces/latest/userguide/amazon-workspaces-osx-client.html

for some strange reason the main domain download site still refers to v2 only

PR #82926, opened 2020-05-20
Update scilab

i0ntempest:

### DIFF
--- a/Casks/scilab.rb
+++ b/Casks/scilab.rb
@@ -1,20 +1,27 @@
 cask 'scilab' do
-  version '6.0.2'
-  sha256 'aec8b67c29fb93c7fbae85e8b42d96f908e893ca680fb2d3e4fbb7012d8c8720'
+  version '6.1.0'
+  sha256 '8076c09047d5039291388b5ceadc13d9f98e943ef885fd16f7f46c46bc62df85'
 
   # utc.fr/~mottelet/scilab/ was verified as official when first introduced to the cask
-  url "https://www.utc.fr/~mottelet/scilab/download/#{version}/scilab-#{version}-x86_64.dmg"
+  url "https://www.utc.fr/~mottelet/scilab/download/branch-#{version.major_minor}/scilab-branch-#{version.major_minor}-x86_64.dmg"
   appcast 'https://macupdater.net/cgi-bin/check_urls/check_url_redirect.cgi?url=https://www.scilab.org/download/'
   name 'Scilab'
   homepage 'https://www.scilab.org/'
 
-  depends_on macos: '>= :sierra'
+  depends_on macos: '>= :high_sierra'
 
-  app "scilab-#{version}.app"
-  binary "#{appdir}/Scilab-#{version}.app/Contents/MacOS/bin/scilab"
-  binary "#{appdir}/Scilab-#{version}.app/Contents/MacOS/bin/scilab-cli"
+  app "scilab-branch-#{version.major_minor}.app"
+  binary "#{appdir}/Scilab-branch-#{version.major_minor}.app/Contents/MacOS/bin/scilab"
+  binary "#{appdir}/Scilab-branch-#{version.major_minor}.app/Contents/MacOS/bin/scilab-cli"
 
   caveats do
     depends_on_java '8'
+
+    <<~EOS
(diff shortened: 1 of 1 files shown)


Casks/scilab.rb
```diff
@@ -1,20 +1,21 @@
...
-  binary "#{appdir}/Scilab-#{version}.app/Contents/MacOS/bin/scilab"
-  binary "#{appdir}/Scilab-#{version}.app/Contents/MacOS/bin/scilab-cli"
+  app "scilab-branch-#{version.major_minor}.app"
+  binary "#{appdir}/Scilab-branch-#{version.major_minor}.app/Contents/MacOS/bin/scilab"
+  binary "#{appdir}/Scilab-branch-#{version.major_minor}.app/Contents/MacOS/bin/scilab-cli"
 
   caveats do
     depends_on_java '8'
   end
+  caveats "If propmted to install legacy Java 6 after installing JDK,\nuse this utility provided by the developer to enable\ninstalled JDK to use with Scilab:\nhttps://www.utc.fr/~mottelet/java/enableJDK.dmg"
```
vitorgalvao, line 20: https://github.com/Homebrew/homebrew-cask/blob/935837c3268e9467f82211ecf72c5a171a929381/Casks/udig.rb#L12L20

vitorgalvao (changes requested):

Casks/scilab.rb
i0ntempest: Do you want me to change the syntax to that or?

vitorgalvao: > Do you want me to change the syntax to that or?

Yes, please. Looks like I overwrote the text. Edited the comment.

miccal:
Here is a copy of the `README` in the `enableJDK.dmg`, which may be relevant:
```
A long time ago, Apple announced that El Capitan would be the end of the line for the Java 6 runtime and tools provided by Apple, with the clear statement that developers should be moving on to Oracle’s Java tools. However, it was still possible to install the legacy Java 6 runtiume... until Catalina !

To completely replace Apple’s Java 6 tools, Oracle’s Java JDK (Java SE Development Kit) will need to be installed. This is because the Oracle Java JRE (Java Runtime Environment) on OS X is a browser plug-in for running Java via a web browser and does not include capabilities for running Java desktop apps or command line tools.

By default though, the Oracle JDK does not set several options to advertise the capabilities provided by the JDK to Java apps, which may cause applications that need those capabilities to fail to launch. The capabilities are actually present in the JDK, but those options need to be set before applications will recognize them as available.

To fix this, we need to add the following options to Oracle’s Java JDK:

    BundledApp
    JNI

In turn, enabling these options means they need to be added to the list of JVMCapabilities stored in the following plist file:
	
/Library/Java/JavaVirtualMachines/jdk_version_info_goes_here.jdk/Contents/Info.plist

This is what this small AppleScript application does.

Stéphane Mottelet, Tue Nov 26 19:37:52 CET 2019
```

Casks/scilab.rb
```diff
@@ -1,20 +1,26 @@
...
+  binary "#{appdir}/Scilab-branch-#{version.major_minor}.app/Contents/MacOS/bin/scilab"
+  binary "#{appdir}/Scilab-branch-#{version.major_minor}.app/Contents/MacOS/bin/scilab-cli"
 
   caveats do
     depends_on_java '8'
+
+    <<~EOS
+      When launching the app, if prompted to install legacy Java 6 after installing JDK, use the utility provided by the developer to enable installed JDK to use with Scilab:
+
+        https://www.utc.fr/~mottelet/java/enableJDK.dmg
```
vitorgalvao, line 23: Is there a page with instructions rather than a direct link to a DMG?

i0ntempest: Here
https://www.utc.fr/~mottelet/scilab_for_macOS.html

i0ntempest: "point #{token} to your installed JDK" isn't really accurate because the tool doesn't change the app Info.plist (like in udig), it changes the JDK Info.plist to add capabilities.

vitorgalvao: @i0ntempest Can you correct the message with a PR (but keeping it short)?

PR #83083, opened 2020-05-22
Update qiyimedia

suschizu:


core-code:
i don't think this is necessary. the https works fine, else it didn't work in the URL stanza either.

its just that their server is barely working at all, at least here in europe

PR #83119, opened 2020-05-23
Update stack from 2.6.2.20200306 to 2.6.4.20200506

brianvanburken:

### DIFF
--- a/Casks/stack.rb
+++ b/Casks/stack.rb
@@ -1,6 +1,6 @@
 cask 'stack' do
-  version '2.6.2.20200306'
-  sha256 'f3b03c6a3a6f4bdafdceb2a7c626ed473cdb127e8254a517c4c998d5c286eabd'
+  version '2.6.4.20200506'
+  sha256 '676241a70d26832f8910bdeb64e5c44f8ff265a27b1267c2daf15beefe4e4a00'
 
   # transip.net/stack/ was verified as official when first introduced to the cask
   url "https://mirror.transip.net/stack/software/osx/stack-#{version}.pkg"
--- a/Casks/stack.rb
+++ b/Casks/stack.rb
@@ -4,7 +4,7 @@ cask 'stack' do
 
   # transip.net/stack/ was verified as official when first introduced to the cask
   url "https://mirror.transip.net/stack/software/osx/stack-#{version}.pkg"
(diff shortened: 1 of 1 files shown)


brianvanburken:
Is it failling for the same reason as #79137? It seems the download is there, but the appcast show only the previous version.

ran-dall:
@brianvanburken I can't seem to get the download page. Is this walled? It seems that I need to login. Granted, I'm translating the page but maybe you can point me to the download page to verify? Then I merge this manually for you.

brianvanburken:
The homepage of this cask points towards the homepage of the service. All downloads are available on https://www.transip.nl/cp/stack/download/ which is walled behind a login. The official app currently displays the following popup asking to upgrade: 
Screenshot 2020-05-23 at 21 12 37

ran-dall:
@brianvanburken Thanks for that, however, we'd still need a way to verify the version is what's intended for release by the developer...

@vitorgalvao I'm pretty sure it's the same problem we ran into before on #83079. I'm not sure how you would like to handle it.

brianvanburken:
I understand. The download page, sadly, does not provide any hashes to verify the downloaded file. The page only shows a link alias to the latest (https://mirror.transip.net/stack/software/Stack-osx-latest.pkg), so no version is supplied.

It does seem that the appcast gets updates; only much later. In the conversation of #79137 it gets mentioned that the appcast didn't show the latest version. It does show that version now.

After debugging using a proxy I found out that the client is making a request to: https://mirror.transip.net/stack/update/?version=2.6.2.20200306&platform=macos&oem=stack&versionsuffix=&updatesegment=18&sparkle=true. Here, as you can see, it will show the latest version. The difference with the appcast feed in this cask is that a version is set and `&updatesegment=18` is missing. After adding the missing segment to the appcast it shows the latest version: https://mirror.transip.net/stack/update/?version=0.0.0&platform=macos&oem=stack&versionsuffix=&updatesegment=18&sparkle=true. Interesting enough the number `18` can be anything but empty or `0`. If I put in zero or leave it empty it shows the previous version; version 2.6.2.20200306. Conclusion: adding the segment fixes the appcast, but it is unknown what the effect of the number has on future updates. I will update this PR and add a new commit with the fixed appcast URL

vitorgalvao:
Nice solution with the appcast. Thank you.

PR #83120, opened 2020-05-23
Update sha in houdahspot.rb

suschizu:
this is quite strange - the download url redirects to the latest download:

```
==> Downloading https://www.houdah.com/houdahSpot/download_assets/HoudahSpot5.1.1.zip
==> Downloading from https://www.houdah.com/houdahSpot/download_assets/HoudahSpot_latest.zip
```
and downloads version 5.1
but if you use the original url directly you get the right version (5.1.1)

The new sha belongs to version 5.1.1 (which you can download from the website)

@vitorgalvao - do you have an idea why this is happening?

### DIFF
--- a/Casks/houdahspot.rb
+++ b/Casks/houdahspot.rb
@@ -2,7 +2,7 @@ cask 'houdahspot' do
   version '5.1.1'
   sha256 '38587dee264c7bae0b5370a60c6175d2471e24de4804f592c12864965bfbf32e'
 
-  url "https://www.houdah.com/houdahSpot/download_assets/HoudahSpot#{version}.zip"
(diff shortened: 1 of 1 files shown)


suschizu:
this was expected:
```
audit for houdahspot: failed
 - download not possible: Checksum for Cask 'houdahspot' does not match.
Expected: 38587dee264c7bae0b5370a60c6175d2471e24de4804f592c12864965bfbf32e
  Actual: e353651fbcdb972432798102edce6b6395d1e401abc2183fb69be5a9302bc43f
```

vitorgalvao:
I’ve tried a few more modifications on the URL and can confirm it always does the redirection you mentioned. Weird. But the appcast has a new URL for this version which we can use and works.

vitorgalvao (approved):

suschizu:
great! thanks 😊

PR #83152, opened 2020-05-24
Updated accurics-cli from 0.0.3 to 0.0.4

ws-accurics:
- Spiced up DMG graphics
- Renamed binary from accurics-cli to accurics
- Now using versioned folders for download and appcast

### DIFF
--- a/Casks/accurics-cli.rb
+++ b/Casks/accurics-cli.rb
@@ -1,14 +1,14 @@
 cask 'accurics-cli' do
-  version '0.0.3'
-  sha256 'b0d1ec5975ae3717f48657a57c07cbf0f606cccf00b879fb9236567f6f63213d'
+  version '0.0.4'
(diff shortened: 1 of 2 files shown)


ws-accurics:
Hi @vitorgalvao,

First off--huge thanks for the initial merge and updates. Some questions: in general, should the binary and cask name always be an exact match, and would you prefer if I updated the appcast/download archive to match in future releases as well?

Thanks again,
-Willie

vitorgalvao:
> in general, should the binary and cask name always be an exact match

No need to be exact, but it helps when we can make it so.

> would you prefer if I updated the appcast/download archive to match in future releases as well?

That one is up to you. We’d just like that the `binary` suffered as few renames as possible.

ws-accurics:
Gotcha. Yep, sorry about that; that one was my fault--the branding folks came back almost immediately after I informed the team the installer was live and made it clear the binary name must be just plain "accurics".

PR #83229, opened 2020-05-26
Update aviatrix-vpn-client from latest to 2.10.8

vjeetje:

### DIFF
--- a/Casks/aviatrix-vpn-client.rb
+++ b/Casks/aviatrix-vpn-client.rb
@@ -4,7 +4,7 @@ cask 'aviatrix-vpn-client' do
 
   # aviatrix-download.s3.amazonaws.com/ was verified as official when first introduced to the cask
   url 'https://aviatrix-download.s3.amazonaws.com/AviatrixVPNClient/AVPNC_mac.pkg'
-  appcast 'https://docs.aviatrix.com/HowTos/changelog.html'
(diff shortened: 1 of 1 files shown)


core-code:
if we are going to have a version number here, we also need a reliable appcast

https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md

core-code:
thanks there.  i think this one might be better though "https://docs.aviatrix.com/Downloads/samlclient.html"

vjeetje:
The reason why I chose the changelog over the downloads page is that the changelog page is less likely to change due to non release changes.  If pointing to the downloads page is more conventional, I'm happy to change the PR.

core-code:
thanks for thinking about this, i see your reasoning. in reality both of these are going to be "noisy" as they are HTML pages which will have many unrelated changes. 

the reason i consider "changelog" pages to be the worst choice is because they *also* contain the old version numbers. this means we cannot do an automated check if the current cask version is no longer contained in the appcast, and infer that there has to be an update. this is a check we run each day to find cask updates ( http://macupdater.net/cask-repair/appcast_containsversion-2020-05-25.txt ). which change-log pages this is impossible and we have to resort to other, less reliable checks.

core-code:
if you are going to make sure this cask is always up-to-date yourself then it doesn't matter ;)

PR #83238, opened 2020-05-26
fix(audit): update renamed token_conflicts argument

jasisk:
Attempting to fix `cask ci` command to reflect change introduced by Homebrew/brew#7427.

New arguments were introduced by the above PR which I have not added, only addressing the renamed argument. Happy to add the additional checks if necessary.


ran-dall (approved):
@jasisk I was just going to make a commit when I saw you already make the change required. Thank you, it LGM, but I'll ping @vitorgalvao and @reitermarkus, and leave it open for a bit, just in case.

PR #83287, opened 2020-05-27
Update touch-bar-simulator.rb from 4.1.0 to 4.2.0

suschizu:
update mojave version


suschizu:
`Error: It seems the App source '/usr/local/Caskroom/touch-bar-simulator/3.2.0/Touch Bar Simulator.app' is not there.`
travis is looking for the <mojave version here

PR #83357, opened 2020-05-28
notable: Zap config file

raphlcx:

### DIFF
--- a/Casks/notable.rb
+++ b/Casks/notable.rb
@@ -16,5 +16,6 @@ cask 'notable' do
                '~/Library/Saved Application State/com.fabiospampinato.notable.savedState',
                '~/Library/Preferences/com.fabiospampinato.notable.plist',
                '~/Library/Application Support/Notable',
+               '~/.notable.json',
              ]
 end


ran-dall:
@Raphx Why not `~/.notable` as well?

Raphx:
Hmm Notable only create `~/.notable.json` in the user home directory, there wasn't a `~/.notable`.

More of the application's related files are in `~/Library/Application Support/Notable`, which has already been zapped.

I might not have used Notable that extensively yet that it creates a `~/.notable` file or directory, but if this is actually present, I will amend this PR.

ran-dall:
@Raphx Would you mind double-checking? I created a few test notes and I got the `.notable` directory, in addition to `.notable.json`...which the reason I ask.

Screen Shot 2020-05-28 at 12 02 50 AM

```shell
$ ls -a ~/ | grep 'notable'
.notable
.notable.json
```

Also, using @nrlquaker's [`createzap`](https://github.com/nrlquaker/homebrew-createzap), I get it...

```shell
$ brew cask createzap notable
zap trash: [
             '~/Library/Application Support/Notable',
             '~/Library/Preferences/com.fabiospampinato.notable.plist',
             '~/Library/Saved Application State/com.fabiospampinato.notable.savedState',
             '~/.notable.json',
             '~/.notable',
           ]
```

Raphx:
I see. Thanks for verifying this in-depth.

I'm speculating `~/.notable` is the default data directory that Notable uses to store notes. Currently I'm using a custom directory as the data directory, hence `~/.notable` doesn't show up on my local machine.

Could you help to verify if the `~/.notable` directory contains the test notes that you have created? If it does contain, then it is indeed the default data directory.

Speaking ahead, just in the case `~/.notable` is the data directory, do we want to zap this directory as well since notes are user-created files, and not application configuration or cache files?

ran-dall:
> I see. Thanks for verifying this in-depth.

@Raphx No problem.

> I'm speculating ~/.notable is the default data directory that Notable uses to store notes. 

> Could you help to verify if the ~/.notable directory contains the test notes that you have created? If it does contain, then it is indeed the default data directory.

I believe it is.

> Speaking ahead, just in the case ~/.notable is the data directory, do we want to zap this directory as well since notes are user-created files, and not application configuration or cache files?

This is a good point, as we definitely don't want to ever `zap` the user's data unless that's what's intended. Seeing that the 'data' is just markdown files, so I'd also argue they are usable without the app. **So I'm okay with leaving out.** 

IMO it was a poor decision on the developer's part to default the data directory to a hidden file.

PR #83435, opened 2020-05-29
created zap stanza for Mountain Duck 3.4.0.15624

guydebros:
There's a folder created by Mountain Duck that I _think_ may be shared with Cyberduck if it's installed. That folder shouldn't be zapped, unless there's a way to check if Cyberduck is or isn't installed. Or maybe there could be a message saying "you'll have to delete this yourself, but this is where to find it," or maybe a prompt for confirmation. But I have no idea how to do that, and I'm not sure how this is typically handled.

### DIFF
--- a/Casks/mountain-duck.rb
+++ b/Casks/mountain-duck.rb
@@ -14,6 +14,6 @@ cask 'mountain-duck' do
   zap trash: [
                '~/Library/Application Scripts/io.mountainduck.findersync',
                '~/Library/Containers/io.mountainduck.findersync',
-               # '~/Library/Group Containers/*.duck',
+               '~/Library/Group Containers/G69SCX94XU.duck',
              ]
 end
--- a/Casks/mountain-duck.rb
+++ b/Casks/mountain-duck.rb
@@ -10,4 +10,10 @@ cask 'mountain-duck' do
   auto_updates true
 
   app 'Mountain Duck.app'
+
+  zap trash: [
+               '~/Library/Application Scripts/io.mountainduck.findersync',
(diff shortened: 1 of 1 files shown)


ran-dall:
> There's a folder created by Mountain Duck that I think may be shared with Cyberduck if it's installed.

@guydebros How can we determine for sure? Also, I'm pretty sure that we'd prefer an absolute path when possible. Is there a reason we're globbing instead?

vitorgalvao:
> That folder shouldn't be zapped, unless there's a way to check if Cyberduck is or isn't installed.

[That’s incorrect](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/zap.md):

> * Shared resources such as application updaters. Since shared resources may be removed, other applications may be affected by `brew cask zap`. Understanding that is the responsibility of the end user.

But as @ran-dall pointed out, make it an absolute path.

guydebros:
Got it. I suggested the glob because the folder name starts with a random string of numbers and letters that I assumed was unique to my install. I've never paid close attention to those group containers. Do they have the same prefix each time on every machine? In that case the absolute path would work.

ran-dall:
> Do they have the same prefix each time on every machine?

@guydebros I believe that is the `application-group-id`, which I believe is unique to each developer.

guydebros:
Thanks, @ran-dall, [you're right](https://developer.apple.com/documentation/foundation/filemanager/1412643-containerurl). (I'm a novice so I have to learn these things one by one.) I'll fix this, and I'll search through my old PRs because I'm certain I did the same thing for a couple other casks.

vitorgalvao (approved):

PR #83472, opened 2020-05-30
Update monero-wallet from 0.15.0.4 to 0.16.0.0

core-code:


core-code:
i think we prefer github appcast feeds but it it took over a week from posting this release on github to actually making it available for download. this means the github feed just doesn't work for our purposes.

vitorgalvao:
> i think we prefer github appcast feeds

Github releases `appcast` feeds should always be used when using Github releases `url`s, because those will be up-to-date. When the `url` is from somewhere else, there’s flexibility.

PR #83668, opened 2020-06-03
Update axe-edit-iii from 1.05.11 to 1.06.00

tamerz:

### DIFF
--- a/Casks/axe-edit-iii.rb
+++ b/Casks/axe-edit-iii.rb
@@ -9,5 +9,7 @@ cask 'axe-edit-iii' do
 
   app 'Axe-Edit III.app'
 
+  uninstall delete: "#{staged_path}/#{token}"
+
   zap trash: '~/Library/Application Support/Fractal Audio/Axe-Edit III'
 end
--- a/Casks/axe-edit-iii.rb
+++ b/Casks/axe-edit-iii.rb
@@ -1,6 +1,6 @@
 cask 'axe-edit-iii' do
-  version '1.05.11'
-  sha256 'b10e1d03d52f87de3d6664e9a46ac0979c2c1e2a86dac98342fd56e4e2d2160d'
+  version '1.05.13'
+  sha256 '9d5d17f1b82eab1eefabed513b3211b2d93fd555cbeefd4b7fbefd8067fab715'
 
   url "https://www.fractalaudio.com/downloads/Axe-Edit-III/Axe-Edit-III-OSX-v#{version.tr('.', 'p')}.dmg"
   appcast 'https://www.fractalaudio.com/axe-fx-iii-edit/'
--- a/Casks/axe-edit-iii.rb
+++ b/Casks/axe-edit-iii.rb
@@ -9,7 +9,5 @@ cask 'axe-edit-iii' do
 
   app 'Axe-Edit III.app'
 
-  uninstall delete: "#{staged_path}/#{token}"
-
   zap trash: '~/Library/Application Support/Fractal Audio/Axe-Edit III'
 end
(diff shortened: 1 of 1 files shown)


miccal:
This Cask may require a fake `uninstall` to pass the CI tests due to the presence of a release notes `pdf` in the `dmg`:
```
uninstall delete: "#{staged_path}/#{token}"
```

Tamerz:
@miccal I gave it a shot but still seems to have a problem. Perhaps I'm doing it wrong?

miccal:
@Tamerz more likely my suggestion was not correct, I thought it may help.

ran-dall:
@miccal I would suggest not doing things to the cask to try to fool the CI. It's something we don't want to get into the habit of doing at all. It's very possible that the CI can fail and the update just merged manually.

ran-dall:
@Tamerz Can you undo that last commit you did and I'll try to figure this out for you later? I'm having one of those days, so it would be later or tomorrow morning latest.

Tamerz:
@ran-dall Thank you, I removed that line.

ran-dall:
@vitorgalvao So the problem with this cask is that I can't really do anything with the file `Axe-Edit-III-Release-Notes.pdf` because it's locked. I tried invoking `chflags` with `nouchg` in `preflight` but I don't think it's working. 

Is that allowed? Also, may there be a similar way of fixing this that has escaped me? Maybe I'm not targeting the file? Is there a way to target files that are put into `/private`.

ran-dall:
@miccal @Tamerz If you guys have any ideas, you're more than welcome to share.

miccal:
@ran-dall how about an `uninstall_preflight` on `"#{staged_path}/Axe-Edit-III-Release-Notes.pdf"`?

ran-dall:
@miccal FMU both install and uninstall actions have a `*flight` for pre- and post- actions. `uninstall_preflight` would be referring to pre-uninstall; however the problem is happen in the directory that `brew cask` uses to mount the files in prior to moving the files into their `Caskroom` home. Thus again FMU `#{staged_path}` refers to the `Caskroom` home of the cask. I don't think there's a way to target the directory in `/var/private/`. Please correct me if I'm wrong...

miccal:
@ran-dall yes you are right -- at the moment I am not sure what to suggest, but I will keep thinking about it.

On a side note, your message initially made me laugh -- FMU in Australia is often taken to mean "f*** me up", not "from my understanding", he he :)

vitorgalvao:
@ran-dall You assessment looks accurate.

We need to unlock the PDF but we don’t know where it is unless we dive into core functionality. A few solutions:

1. Change the core so when it mounts a DMG every file is checked and unlocked.
2. Mount the DMG ourselves inside the cask, and do the process manually.
3. Ask upstream to not lock the PDF.
4. Remove the cask.

In turn, each option is:

1. The most robust solution. But it’s adding a solution to a problem so rare it’s negligible.
2. Hideous. It might be possible, but we we shouldn’t even attempt it.
3. Unknown. It depends on upstream so success is unknown until tried.
4. More than acceptable. Frankly this cask does not deserve all the talk it generated around its problem:

```
axe-edit-iii
30 days: 2 (#4569)
90 days: 3 (#5381)
365 days: 5 (#5975)
Age: 489 days (added 2019, February 06)
```

My recommended course of action: Someone (@Tamerz?) contacts upstream and asks them to provide the PDF unlocked. If they don’t do it, we remove the cask. So we’d be doing **3** with **4** in case of failure.

ran-dall:
@miccal @Tamerz I'm in agreement to take @vitorgalvao recommendation.

Thanks for looking that over @vitorgalvao 🤝

Tamerz:
Seems like it is fixed now.

ran-dall:
@Tamerz Thanks for sticking with this. Good work. 👍

PR #83684, opened 2020-06-03
Revert "Update balenaetcher from 1.5.95 to 1.5.96"

suschizu:
Reverts Homebrew/homebrew-cask#83670

the download still contains version 1.5.95


miccal:
https://github.com/Homebrew/homebrew-cask/pull/83738

PR #83962, opened 2020-06-08
Update adding_a_cask.md

iMonZ:
Some fixes for better usability


vitorgalvao:
Thank you, but we don’t intend to start changing everything to add support for fish as well.

iMonZ:
Hey this is the only problem that I have with fish!
I can remove this one if you want but there is also another change

vitorgalvao:
Does the `--download` being after causes problems with fish?

PR #84020, opened 2020-06-09
Cask suggestion: StoryMill (by Mariner Software)

jakewilliami:
There is no doubt that the Mariner Software team makes intuitive and efficient-to-use software (though I think it is personally overpriced, but it certainly works for some).  I don't own any of their other software (future endeavour to put on `homebrew-cask` by those who do?) but I think that it'd be nice to have this as an installable cask.  

Keep in mind that **this is available on the app store**.  What do y'all think?  Is there any point if you can get it from the app store?  It is *paid software* but *downloadable for free*.

---

❌ `brew cask install {{cask_file}}` worked successfully.
    Everything worked except:
    ```
    Error: It seems the App source '/usr/local/Caskroom/storymill/4.0.5/StoryMill' is not there.
    ```

❌ `brew cask uninstall {{cask_file}}` worked successfully.
    Due to the above error:
    ```
    Error: Cask 'storymill' is not installed.
    ```

### DIFF
--- a/Casks/storymill.rb
+++ b/Casks/storymill.rb
@@ -1,0 +1,20 @@
+cask 'storymill' do
+  version '4.0.5'
+  sha256 :no_check
+
+  # s3.amazonaws.com/ was verified as official when first introduced to the cask
+  url 'https://s3.amazonaws.com/MarinerDownloads/StoryMill/StoryMill_Mac_4_0_5.dmg'
+  name 'StoryMill'
+  homepage 'https://marinersoftware.com/'
+
+  app 'StoryMill'
+
+  zap trash: [
+               '~/Library/Caches/com.marinersoftware.StoryMill',
+               '~/Library/Preferences/com.marinersoftware.StoryMill.plist',
+               '~/Library/Application Support/.StoryMill.plist',
+             ],
+      rmdir: [
+               '~/Library/Application Support/StoryMill/',
+             ]
+end
--- a/Casks/storymill.rb
+++ b/Casks/storymill.rb
@@ -1,20 +1,19 @@
 cask 'storymill' do
   version '4.0.5'
-  sha256 :no_check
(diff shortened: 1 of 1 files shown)


miccal:
[Their website says that StoryMill is discontinued?](https://marinersoftware.com/discontinuedproducts/)

jakewilliami:
@miccal Oh my, I didn't know this.  Weird, because I checked to see that it was 64-bit before submitting the PR, and indeed it was 64-bit (see attached).  

Screen Shot 2020-06-09 at 6 59 36 PM

I could consider adding to [homebrew-cask-versions](https://github.com/Homebrew/homebrew-cask-versions) but it doesn't seem worth it if it is discontinued and it is on the app store.

Just before closing this, I was wondering why some applications are on the App Store AND homebrew-cask, but some are only on the App Store?  Is there any reason for this other than perhaps someone hasn't added the cask yet?

kmutahar:
> Just before closing this, I was wondering why some applications are on the App Store AND homebrew-cask, but some are only on the App Store? Is there any reason for this other than perhaps someone hasn't added the cask yet?
@jakewilliami 

It all depends in the developer. Some developers give the user to download the app from their own site and from the Apple App Store. This can be see more in paid apps so that they don't have to pay Apple its cut if purchased from the developer directly.

Another thing is that most time the direct site can be updated faster then the App Store since they don't have to wait for Apple approval. Another thing is that there may be a small difference due to sandboxing requirements. At the end it all depends on the developer. 

Homebrew-cask has no say in that only if a non-App store exist and fits the requirements it can be add. 

(someone correct me if I am wrong in anything)

vitorgalvao:
We don’t generally care if an app is on the MAS or not. We need apps to be downloadable directly from an official source, and that it can be made into the full app (i.e. apps that are downloadable as trials and the only way to get the full version is from the MAS aren’t accepted).

Thank you for the submission, but we’ll refuse due to being discontinued. Also, this cask has a lot of errors. I’ll fix them directly so you can see where they are for next time.

PR #84133, opened 2020-06-10
Remove support for Mavericks

claui:
With https://github.com/Homebrew/brew/pull/7716 and https://github.com/Homebrew/brew/pull/7698 now merged, Homebrew no longer runs on Mavericks, and support for the `:mavericks` symbol has been removed.

This causes `brew cask install` and `brew cask uninstall` to fail with the following error message:

```
Error: Cask 'deeper' definition is invalid: invalid 'depends_on macos' value: :mavericks
```

This PR fixes the issue by removing all references to `:mavericks`.

Please mind that **CI will fail** because this is a single PR for all 39 casks (albeit multiple commits).

___

- [ ] `brew cask audit --download {{cask_file}}` is error-free. **Audit fails for 3 out of 39 casks: for `adobe-dng-converter`, the server keeps killing the connection; `geekbench` and `time-out` suffer from the [TLS issue](https://github.com/Homebrew/brew/issues/7667) but are fine otherwise. All other casks pass the audit.**

### DIFF
--- a/Casks/adobe-dng-converter.rb
+++ b/Casks/adobe-dng-converter.rb
@@ -1,11 +1,6 @@
 cask 'adobe-dng-converter' do
-  if MacOS.version <= :mavericks
-    version '9.6.1'
-    sha256 '087eac5026667e4e6e3c156fd13243c9ea00f6c0238cbbb94d3099ae8772603f'
-  else
-    version '12.2'
-    sha256 'cc2962dd051c29bb2c4737a831477be5b0f35531f0fbc1ea2c10c2449201f0f7'
-  end
+  version '12.2'
+  sha256 'cc2962dd051c29bb2c4737a831477be5b0f35531f0fbc1ea2c10c2449201f0f7'
 
   url "https://download.adobe.com/pub/adobe/dng/mac/DNGConverter_#{version.dots_to_underscores}.dmg"
   appcast 'https://supportdownloads.adobe.com/product.jsp?product=106&platform=Macintosh'
--- a/Casks/amethyst.rb
+++ b/Casks/amethyst.rb
@@ -1,9 +1,5 @@
 cask 'amethyst' do
-  if MacOS.version <= :mavericks
-    version '0.9.10'
-    sha256 '82adf42ce6031ab59a3072e607788e73f594ad5f21c7118aabc6c5dafe3d0b47'
-    url "https://ianyh.com/amethyst/versions/Amethyst-#{version}.zip"
-  elsif MacOS.version <= :el_capitan
+  if MacOS.version <= :el_capitan
     version '0.10.1'
     sha256 '9fd1ac2cfb8159b2945a4482046ee6d365353df617f4edbabc4e8cadc448c1e7'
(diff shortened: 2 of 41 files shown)


vitorgalvao:
Merging fast before there are merge conflicts.

Though this will be problematic now, because every installed cask with `mavericks` will fail, giving rise to bug reports that necessitated https://github.com/Homebrew/homebrew-cask/issues/58046.

We really need to fix that problem.

claui:
Thanks @vitorgalvao!

vitorgalvao:
Thank *you*!

MikeMcQuaid:
> Though this will be problematic now, because every installed cask with `mavericks` will fail, giving rise to bug reports that necessitated #58046.

@vitorgalvao if you can provide some reproduction steps for this I can give it a go at digging into the cask code and fixing it once and for all.

danielbayley:
So should `depends_on macos: '>= :mavericks'` in existing Casks not now be replaced with `depends_on macos: '>= :yosemite'`, rather than removing the line altogether?

MikeMcQuaid:
@danielbayley I don't think so because `brew cask` can no longer be run on pre-yosemite.

danielbayley:
Ah ok, so really only `depends_on macos: '>= :el_capitan'` would be used with `>=` in practice…

claui:
@MikeMcQuaid @vitorgalvao Here’s a simple, self-contained snippet to reproduce the issue:

**(Make a backup before running the following Bash script. It will mess with your installation!)**

```bash
(
    set -ex -o pipefail;

    # Forcibly remove mpv if present, even in case cask is not loadable
    rm -rf 2>/dev/null \
        "$(brew --prefix)/Caskroom/mpv" \
        '/Applications/mpv.app' \
        | true
    # Restore pristine state of `version.rb`
    git -C "$(brew --repository)" checkout -- \
        'Library/Homebrew/os/mac/version.rb'

    # Install cask that depends on Yosemite
    brew cask install mpv

    # Simulate upstream dropping Yosemite support
    sed -i '' -e '/yosemite/d' \
        "$(brew --repository)/Library/Homebrew/os/mac/version.rb"

    # Try to do something with the installed cask
    brew cask info mpv
    echo 'Success'  # this line is never reached
)
```

vitorgalvao:
Building on @claui’s code, a self-contained version that won’t touch your local installation of either `mpv` or `brew`.

```bash
# Temporary directories
brew_tmp_dir="$(mktemp -d)"
mpv_tmp_dir="$(mktemp -d)"

# Clone upstream Homebrew/brew
git clone 'https://github.com/Homebrew/brew.git' "${brew_tmp_dir}"

# When calling brew, use the one from the temporary directory
# Closing the shell is all you need to do to get back to normal
alias brew="${brew_tmp_dir}/bin/brew"

# Install mpv to temporary directory
brew cask install --appdir="${mpv_tmp_dir}" mpv

# Simulate upstream dropping Yosemite support
sed -i '' -e '/yosemite/d' "${brew_tmp_dir}/Library/Homebrew/os/mac/version.rb"

# Try to do something with the installed cask
brew cask info mpv && echo 'Success'

```

PR #84210, opened 2020-06-12
Add Kui v8.9.1

LouisStAmour:
Added Kui but in testing the included kubectl-kui binary in /usr/local/bin, I found I needed to make an upstream PR to have it look for Kui in /Applications.
So I opened a PR at https://github.com/IBM/kui/pull/4878 ~and temporarily added a workaround to the caveats section~.

### DIFF
--- a/Casks/kui.rb
+++ b/Casks/kui.rb
@@ -1,0 +1,12 @@
+cask 'kui' do
+  version '8.9.1'
+  sha256 '41778159c8fe32d77ca84449d81ccb2674656a5bbe3a1d45eea187543e572b9e'
+
+  # github.com/IBM/kui was verified as official when first introduced to the cask
+  url "https://github.com/IBM/kui/releases/download/v#{version}/Kui-darwin-x64.tar.bz2"
+  appcast 'https://github.com/IBM/kui/releases.atom'
+  name 'Kui'
+  homepage 'https://kui.tools/'
+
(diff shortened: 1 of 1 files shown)


ran-dall:
@vitorgalvao Can you look over the `caveat`? It seems long to me.

vitorgalvao:
The caveat is definitely too long. All of it should go.

Also, we don’t add `TODO` comments: those never get done, they just stay as eternal comments.

LouisStAmour:
Updated as requested, while I would prefer to ship instructions on how to get the binary working from command line as a Kubectl plugin as per the installation instructions in the Kui repo, I can submit an update later with a binary rule when the upstream PR fix is merged.

I'll point out that if there is a length requirement on caveats, it is not documented in either `brew cask style` nor in the Github docs. I may have been using Homebrew for far too long, but I've seen far longer caveat texts historically, including those which linked to GitHub issues when workarounds were required. Perhaps rules differed historically, or perhaps Homebrew upstream has different rules than Homebrew-Casks. But I thought my Caveats section was relatively succinct given the 3 commands that were suggested. I recognize that the Installation instructions could be adapted to suggest the same, but dynamically telling users what homebrew path to use for a temporary workaround seemed more like a caveats task than a Markdown task.

vitorgalvao:
> or perhaps Homebrew upstream has different rules than Homebrew-Casks.

We do.

PR #84292, opened 2020-06-13
Update screen from 1.2.6 to 1.2.7

ran-dall:

### DIFF
--- a/Casks/screen.rb
+++ b/Casks/screen.rb
@@ -1,6 +1,6 @@
 cask 'screen' do
-  version '1.2.6'
-  sha256 '354dc29002bb146279d4b4843288e09f96a2708918f1ec66ae90a28651bb3612'
+  version '1.2.7'
+  sha256 'a80f6e5dd8c2e6dd03503bba0d222d7bbef96aab4a815c71ab8b7becd1fb66d9'
 
   url "https://download.screen.so/desktop-app/darwin/#{version}/Screen.dmg"
   appcast 'https://download.screen.so/'


jsherwani:
Hello @ran-dall 👋  I'm the founder of Screen, and I wanted to find out how homebrew interoperates with Squirrel's update mechanism (which we use to manage auto-updates on Mac).

Is homebrew responsible just for downloading the app, and then it functions just as though the user downloaded the app themselves from our website?

I couldn't find any documentation on this, other than: https://github.com/Homebrew/homebrew-cask/pull/72166/files, which seemed somewhat eerie — is it no longer considered generally safe for apps to update themselves? If so, how do apps that have update mechanisms inter-operate with homebrew?

ran-dall:
@jsherwani Thanks for the ping! Sorry for the delay in getting back to you. HBC itself has a [`appcast`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) system that deals with updates universally. 

> Is homebrew responsible just for downloading the app, and then it functions just as though the user downloaded the app themselves from our website?

That's the idea. But upon uninstall, we maybe cleanout files/folders better than just dragging the `.app` into the trash.

> I couldn't find any documentation on this, other than: #72166 (files), which seemed somewhat eerie — is it no longer considered generally safe for apps to update themselves? If so, how do apps that have update mechanisms inter-operate with homebrew?

I think it's still generally safe. I think the reason for removal was because people were thinking that adding `auto-update` to everything was okay when in many situations it isn't. If an app isn't set up properly to `auto-update` there could be some data-loss on the user's side, because we'll essentially uninstall and reinstall the latest version.

> If so, how do apps that have update mechanisms inter-operate with homebrew?

I don't think we have mechanisms that 'inter-operate', but rather it's more about the `brew cask upgrade` process.

I hope that answers all your questions. Feel free to reply, if have any more or I didn't properly address anything.

PR #84357, opened 2020-06-14
Update qsync-client from 3.4.0.0416 to 3.3.7.0514

suschizu:

### DIFF
--- a/Casks/qsync-client.rb
+++ b/Casks/qsync-client.rb
@@ -1,6 +1,6 @@
 cask 'qsync-client' do
-  version '3.4.0.0416'
-  sha256 '704638070a9e557682ca19c1aa62f6496fdc86ff3aa918b55d32ade859415aad'
+  version '3.3.7.0514'
+  sha256 '97d7813e1c2a171d07355969236719e7a70e5e475522ea840e87594ab8ea478b'
 
   url "https://download.qnap.com/Storage/Utility/QNAPQsyncClientMac-#{version}.dmg"
   appcast 'https://update.qnap.com/SoftwareRelease.xml'


John-Dennert:
@suschizu 
Why did you roll this back to an older version?

suschizu:
@John-Dennert - because this is the version from the website  (https://www.qnap.com/en-in/utilities/essentials#block_3)
I think it has been reverted (?)

John-Dennert:
@suschizu 
Interesting... I was helping a friend get reconnected a couple of weeks ago and it wouldn't connect (they were running an older 3.3.x version). When I installed it on my computer with Homebrew it installed 3.4.0.0416. When I went to QNAP website on their computer to download the update I noticed that the version I had was newer than what was available on the website. I thought that was kind of strange but I just copied the newer installer from my computer to their computer and it was finally able to connect.

I just tested 3.3.7.0514 and got login errors in the client but none on the server. When I reinstalled 3.4.0.0416 it worked fine.

John-Dennert:
@suschizu 
Just to be clear, I understand why you "updated" the version. Hopefully QNAP gets whatever the issues are straightened out.

In the meantime, I'm going to modify this cask locally and hang onto a copy of the 3.4.0.0416 installer in case QNAP actually removes it from their website and I need it again.

suschizu:
@John-Dennert thanks for the explanation - I don't use this app I just checked if its up-to-date ...  so its interesting that the "official" version is buggy and the non-official one works fine - 
but also the build in update check also doesn't know about version 3.4.0 ... 
not sure what is the best solution here …
image

PR #84304, opened 2020-06-13
Update santa from 0.9.33 to 1.13

peterlewis:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #84366, opened 2020-06-14
santa.rb: add caveat

vitorgalvao:
Continuation of https://github.com/Homebrew/homebrew-cask/pull/84304#issuecomment-643760204.


peterlewis:
That's fab, thank you for adding the caveat, @vitorgalvao! :)

PR #84439, opened 2020-06-16
Add Micronaut CLI 2.0.0.M3

alvarosanchez:

### DIFF
--- a/Casks/micronaut.rb
+++ b/Casks/micronaut.rb
@@ -5,6 +5,7 @@ cask 'micronaut' do
   # github.com/micronaut-projects/micronaut-starter was verified as official when first introduced to the cask
   url "https://github.com/micronaut-projects/micronaut-starter/releases/download/v#{version}/mn-darwin-amd64-v#{version}.zip"
   appcast 'https://github.com/micronaut-projects/micronaut-starter/releases.atom'
+  name 'Micronaut'
   homepage 'https://micronaut.io/'
 
   depends_on formula: 'openjdk'
--- a/Casks/micronaut.rb
+++ b/Casks/micronaut.rb
@@ -1,0 +1,14 @@
+cask 'micronaut' do
+  version '2.0.0.M3'
+  sha256 '15a748072060263c0eee2b2bb808c3eed75516066542fcd2d1bccb39ae3e64e5'
+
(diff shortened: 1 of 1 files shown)


alvarosanchez:
Note that [there is a `micronaut` formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/micronaut.rb) that is intended to be used for 1.x versions.

From 2.x onwards, we are also shipping pre-compiled native executables built with GraalVM, which is what this cask installs.

vitorgalvao:
This should continue to be a formula, upgrading the one that‘s there. Why isn’t it?

Having v1 of a software as a formula and v2 as a cask is confusing to a point I don’t think we want to pursue.

alvarosanchez:
I agree it should be a formula, and that was my initial approach.

However, the `audit` command failed saying:

```
https://github.com/micronaut-projects/micronaut-starter/releases/download/v/mn-darwin-amd64-v.zip looks like a binary package, not a source archive; homebrew/core is source-only.
```

That led me to create a cask instead.

vitorgalvao:
Well, yes, but if v1 can be built from source as a formula, why can’t v2?

alvarosanchez:
v1 isn't built from source either. The zip contains a pre-copmiled JAR file and a startup script that does `java -jar` under the hood. I believe this is the common approach for Java-based applications. 

Note that we (the Micronaut team) didn't create the original formula, but is our plan to keep it up-to-date automatically going forward.

vitorgalvao:
> Having v1 of a software as a formula and v2 as a cask is confusing to a point I don’t think we want to pursue.

You haven’t addressed that. Why have both?

alvarosanchez:
No reason why. I don't want to pursue having both either. Happy to contribute an upgrade of the existing formula, but like I said, the audit fails.

vitorgalvao:
> I don't want to pursue having both either.

Then we should only have one. You should first submit a PR to [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) to remove v1, and when that’s accepted (alternatively, that part of the team may be able to get it to compile from source as a formula) we can merge the cask.

alvarosanchez:
https://github.com/Homebrew/homebrew-core/pull/56440

PR #84443, opened 2020-06-16
Update activedock from 2.0 to 2.0,2001

suschizu:


suschizu:
perhaps add the bundle version here too because they are changing more often the bundle version than the short version string

PR #84455, opened 2020-06-16
wireshark: Add shim script

gechr:

---

Prompted by https://github.com/eldadru/ksniff/issues/63, it seems that the `wireshark` CLI script was removed somewhere along the line when the cask was upgraded from v2 to v3.

This adds a shim script in the same way it's done in other Casks to address the issue.

### DIFF
--- a/Casks/wireshark.rb
+++ b/Casks/wireshark.rb
@@ -14,6 +14,16 @@ cask 'wireshark' do
   app 'Wireshark.app'
   pkg 'Install ChmodBPF.pkg'
   pkg 'Add Wireshark to the system path.pkg'
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/wireshark.wrapper.sh"
+  binary shimscript, target: 'wireshark'
+
+  preflight do
+    IO.write shimscript, <<~EOS
+      #!/bin/sh
+      exec #{appdir}/Wireshark.app/Contents/MacOS/Wireshark "$@"
+    EOS
+  end
 
   uninstall_preflight do
     system_command '/usr/sbin/installer',


miccal:
@gechr would you know what exactly the `pkg 'Add Wireshark to the system path.pkg'` does? I assume the `shimscript` that you have added does not interfere?

gechr:
Hmm according to https://github.com/wireshark/wireshark/blob/77d94aea2/CMakeLists.txt#L2850-L2881, it seems that `Add Wireshark to the system path.pkg` adds https://github.com/wireshark/wireshark/blob/77d94aea2/packaging/macosx/path_helper/root/etc/paths.d/Wireshark to the system `PATH`:

```
$ cat /etc/paths.d/Wireshark
/Applications/Wireshark.app/Contents/MacOS

$ find /Applications/Wireshark.app/Contents/MacOS/ -maxdepth 1 -executable -type f
/Applications/Wireshark.app/Contents/MacOS/capinfos
/Applications/Wireshark.app/Contents/MacOS/captype
/Applications/Wireshark.app/Contents/MacOS/dftest
/Applications/Wireshark.app/Contents/MacOS/dumpcap
/Applications/Wireshark.app/Contents/MacOS/editcap
/Applications/Wireshark.app/Contents/MacOS/idl2wrs
/Applications/Wireshark.app/Contents/MacOS/mergecap
/Applications/Wireshark.app/Contents/MacOS/mmdbresolve
/Applications/Wireshark.app/Contents/MacOS/randpkt
/Applications/Wireshark.app/Contents/MacOS/rawshark
/Applications/Wireshark.app/Contents/MacOS/reordercap
/Applications/Wireshark.app/Contents/MacOS/sharkd
/Applications/Wireshark.app/Contents/MacOS/text2pcap
/Applications/Wireshark.app/Contents/MacOS/tshark
/Applications/Wireshark.app/Contents/MacOS/Wireshark
```

Not sure if something's wrong with my setup, but none of these are available to me when I `<tab>` complete. Also, the `wireshark` binary is not lowercase, as it was in v2.x, so I think this change is still valid (or an upstream patch).

miccal:
@gechr so the `pkg 'Add Wireshark to the system path.pkg'` seems to add `/Applications/Wireshark.app/Contents/MacOS` to the `PATH`, implying that calling `Wireshark` from the command line should already work, correct?

gechr:
@miccal Yep, `Wireshark` works after I sorted out `/usr/libexec/path_helper`. However, there is still the issue of case-(in)sensitivity.

It _happens_ to work for me regardless of case because my filesystem is case-insensitive, i.e. `WIRESHARK`, `Wireshark`, `wireshark` all work, but for other people who have a case-sensitive filesystem, it won't work, so I think this PR might still be valid for those folk.

Casks/wireshark.rb
```diff
@@ -14,6 +14,16 @@
...
   pkg 'Install ChmodBPF.pkg'
   pkg 'Add Wireshark to the system path.pkg'
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/wireshark.wrapper.sh"
+  binary shimscript, target: 'wireshark'
+
+  preflight do
+    IO.write shimscript, <<~EOS
+      #!/bin/sh
+      exec #{appdir}/Wireshark.app/Contents/MacOS/Wireshark "$@"
```
vitorgalvao, line 24: This probably could be a `binary` on its own. But to be considered at all, there needs to be an official source that recommends it, otherwise every cask would do it.

vitorgalvao (changes requested):

Casks/wireshark.rb
gechr: I don't think it can be a symlinked `binary` on its own. I tried that prior to raising this PR and got the following error:

```
$ ln -s /Applications/Wireshark.app/Contents/MacOS/Wireshark /usr/local/bin/wireshark

$ wireshark
qt.qpa.plugin: Could not find the Qt platform plugin "cocoa" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

fish: Job 1, 'wireshark' terminated by signal SIGABRT (Abort)
```

> _[...] there needs to be an official source that recommends it_

Not entirely sure what you mean by this – do you mean an official upstream Wireshark source that recommends creating a shim script? If so, I don't think there is one, I just borrowed the shim pattern from various other Casks:

- https://github.com/Homebrew/homebrew-cask/search?q=shim

vitorgalvao: > I just borrowed the shim pattern from various other Casks

Those casks have a reason for the shim script: upstream directly recommends doing that linking for using the application from the command line because they behave like CLI tools.

If there’s no such upstream recommendation, we don’t add those. The result of a cask installation should be as close to a manual installation as possible. Otherwise someone would add shim scripts to every cask (not an exaggeration, it’s been tried), complicating them and polluting users’ paths with a ton of garbage.

PR #84520, opened 2020-06-17
Delete tcl

miccal:
Downloads are now walled:

Screen Shot 2020-06-17 at 20 34 56


vitorgalvao:
Downloads only count as walled if `url` is from a different host than the `homepage` (i.e. if it would need the `was verified as official when first introduced to the cask` comment), because users need to endure it isn’t just a random URL someone made up.

When the host is the same it’s not walled, because if you trust the home source, you trust the download.

miccal:
Thanks @vitorgalvao -- `brew cask fetch tcl` did not work for several hours yesterday and I could not find an updated `url` because I could not get past the registration page, but it is working now. 

[Note however that it may not work for much longer:](https://downloads.activestate.com/ActiveTcl/releases/)

Screen Shot 2020-06-18 at 03 27 38

PR #84527, opened 2020-06-17
Add Okta CLI v 0.3.1

bdemers:
Adds new cask for binary based Okta CLI tool
This project is OSS, however it depends on a host of pre-build binaries that are outside the Brew ecosystem (GraalVM, pre built Java artifacts, etc), so going the Cask route seems like the correct path to take.

### DIFF
--- a/Casks/okta.rb
+++ b/Casks/okta.rb
@@ -1,0 +1,11 @@
+cask 'okta' do
+  version '0.3.1'
+  sha256 'f063964e4017c98e193f235453575cd14e9ee42a68482c02923e777bd4b7bfdc'
+
+  url "https://github.com/oktadeveloper/okta-cli/releases/download/okta-cli-tools-#{version}/okta-cli-macos-#{version}-x86_64.zip"
+  appcast 'https://github.com/oktadeveloper/okta-cli/releases.atom'
+  name 'okta'
+  homepage 'https://github.com/oktadeveloper/okta-cli'
+
+  binary 'okta'
+end


vitorgalvao:
Let’s see how https://github.com/Homebrew/homebrew-core/pull/56446 goes, first.

MikeMcQuaid:
@vitorgalvao My take: this should also be rejected from both places if it can be built from source but doing so is "too hard".

graemerocher:
@MikeMcQuaid I don't think anyone is saying building from source is "too hard" or impossible for either the Micronaut PR or this one. With the Micronaut PR we actually provided a formula to build from source and you guys pushed back saying we had to provide a GraalVM formula. Frankly it is you guys making it hard not us since you are trying to place developers independent of the GraalVM project in charge of coming up with a formula for GraalVM which could be defined in all manner of ways and is best done by someone from the GraalVM community. The "too hard" comment is quite frankly patronizing.

MikeMcQuaid:
> With the Micronaut PR we actually provided a formula to build from source and you guys pushed back saying we had to provide a GraalVM formula.

"Building from source" implies "building from source using dependencies in Homebrew or that are built from source where it is possible to do so".

> Frankly it is you guys making it hard not us since you are trying to place developers independent of the GraalVM project in charge of coming up with a formula for GraalVM which could be defined in all manner of ways and is best done by someone from the GraalVM community.

Micronaut decided to move to adopting a dependency that is not packaged by Homebrew and that they aren't willing to work with us to get it built from source. 

Look, we're taking the same attitude here that other open source system package managers will. Go suggest to Debian that they use upstream binaries rather than building from source and see how that goes there.

> The "too hard" comment is quite frankly patronizing.

It's not intended to be patronising. I'm not stating whether it is or is not too hard for any individual or project to so do. I am stating that when there's an impasse on such things we're going to fall back to the existing policies rather than creating one-off exceptions (that quickly become many-off exceptions).

graemerocher:
Well hopefully one day a GraalVM forumula exists and we can provide a better experience for Homebrew users, ultimately the majority of our users use https://sdkman.io/ to install Micronaut so this is not important enough for us.

We will either continue to publish the non-native universal Java version of Micronaut to Homebrew or just live with the Homebrew tap solution. Thanks for your time.

MikeMcQuaid:
> We will either continue to publish the non-native universal Java version of Micronaut to Homebrew

Yes, this is what we'd prefer, thanks.

bdemers:
Closing in favor of a tap: https://github.com/oktadeveloper/homebrew-tap

PR #84577, opened 2020-06-18
Remove Spoofax

Baltoli:
The maintainers have their own homebrew tap where this Cask can live.


vitorgalvao:
Usually that wouldn’t be a reason to remove it, but seeing as this was added one day ago by yourself and has no downloads, fine.

PR #85037, opened 2020-06-26
Add Unclutter 2.1.22d

NanoSector:
This cask has been submitted a few times before but rejected because it was unclear if the app could be purchased from within the app and not requiring a download: #46823 #37546

I have bought the app and it is possible to activate it with an activation link from the e-mail you receive afterwards. The app also mentions activation using a license file. It does not require a re-download, clicking the link brings you back to Unclutter which then activates and removes the trial restrictions. Therefore I don't think it differs from apps like Little Snitch or Alfred which are included in the repo.

### DIFF
--- a/Casks/unclutter.rb
+++ b/Casks/unclutter.rb
@@ -8,4 +8,12 @@ cask 'unclutter' do
   homepage 'https://unclutterapp.com/'
 
   app 'Unclutter.app'
+
+  zap trash: [
+               '~/Library/Application Support/Unclutter',
+               '~/Library/Caches/com.softwareambience.Unclutter',
+               '~/Library/Containers/com.softwareambience.Unclutter',
+               '~/Library/Group Containers/*.com.softwareanbience.Unclutter',
+               '~/Library/Preferences/com.softwareambience.Unclutter.plist',
+             ]
 end
--- a/Casks/unclutter.rb
+++ b/Casks/unclutter.rb
@@ -1,0 +1,10 @@
+cask 'unclutter' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://unclutterapp.com/files/Unclutter.zip'
+  name 'Unclutter'
+  homepage 'https://unclutterapp.com/'
+
+  app 'Unclutter.app'
+end
--- a/Casks/unclutter.rb
+++ b/Casks/unclutter.rb
@@ -1,5 +1,5 @@
 cask 'unclutter' do
-  version '2.1.22d'
+  version '2.1.2200'
   sha256 'f2a60f896e8c8c1f21a1beb3b9c5a833343a2932ab423ebc8b791eb8db1cca1b'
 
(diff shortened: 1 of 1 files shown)


Casks/unclutter.rb
```diff
@@ -0,0 +1,10 @@
+cask 'unclutter' do
+  version :latest
+  sha256 :no_check
```
ran-dall, line 3: Needs a `version` preferably.

Casks/unclutter.rb
```diff
@@ -0,0 +1,10 @@
...
+
+  url 'https://unclutterapp.com/files/Unclutter.zip'
+  name 'Unclutter'
+  homepage 'https://unclutterapp.com/'
+
+  app 'Unclutter.app'
+end
```
ran-dall, line 10: Missing a `zap` stanza.

ran-dall (changes requested):

Casks/unclutter.rb
Yoshi2889: The site only gives out an Unclutter.zip without version info. I could version and hash this but whenever they push out an update it’ll break this cask, I’m not sure that’s preferable. Maybe there’s a different endpoint it uses for updates, I’ll have to check.

Casks/unclutter.rb
```diff
@@ -0,0 +1,10 @@
+cask 'unclutter' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://unclutterapp.com/files/Unclutter.zip'
```
vitorgalvao, line 5: Please add an [`appcast`](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md): `https://unclutterapp.com/files/Unclutter.zip`

vitorgalvao (changes requested):

Casks/unclutter.rb
vitorgalvao: Usually you’d be right. But because we can have an `appcast`, it can be versioned.

vitorgalvao:
> but rejected because it was unclear if the app could be purchased from within the app

No, it was rejected because *it was clear it couldn’t* be purchased from outside the MAS. It’s irrelevant if it can be done from inside the app.

Right now there’s a clear direct payment option, so it can be accepted.

ran-dall:
@Yoshi2889 All you're missing now is a `zap` stanza.

Yoshi2889:
I’ll take a look at the zap stanza later tonight, thank you for the feedback and changes!

Yoshi2889:
I've added a zap stanza as requested; I wasn't sure if the `/var/folders` entries should also be included, I remember reading that macOS cleans those up automatically on reboot.

I mainly used AppCleaner to scan for these as it seems to usually be fairly accurate, this is what it came up with:

image

Please let me know if I missed anything.

PR #85311, opened 2020-07-02
fix version string

luckman212:
`1.8,2020.06` isn't a valid version string - and messes up MacUpdater's auto-update functionality

### DIFF
--- a/Casks/silentknight.rb
+++ b/Casks/silentknight.rb
@@ -1,5 +1,5 @@
 cask 'silentknight' do
-  version '1.8,2020.06'
+  version '1.8'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> `1.8,2020.06` isn't a valid version string

Yes, it is. Plenty of casks use it. Look at this cask itself; by removing the string after the comma you’re breaking it.

> and messes up MacUpdater's auto-update functionality

Worry not, that will be fixed soon, if it hasn’t been already.

core-code:
thanks, fixed on our end. holding down shift while selecting 're-scan' forces a config-file update which resolves it.

luckman212:
@core-code Nice! Hidden feature I did not know about 👍

PR #85314, opened 2020-07-02
libreoffice-language-pack 6.4.5

miccal:
```
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/developer/bin/update_multilangual_casks /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Casks/libreoffice-language-pack.rb 6.4.5
```

### DIFF
--- a/Casks/libreoffice-language-pack.rb
+++ b/Casks/libreoffice-language-pack.rb
@@ -1,558 +1,493 @@
 cask 'libreoffice-language-pack' do
-  version '6.4.4'
+  version '6.4.5'
 
   language 'af' do
-    sha256 '01a49b01c55578fddfd218439bed47888c0ddbbc6da78a5b888cb82730673241'
+    sha256 'e5c59d35e577cbb49ae9408d4de67ef99f7c27d946955896cde7b2eb8bfb4e4d'
     'af'
   end
 
   language 'am' do
-    sha256 'ff4928cc831ba964dc3844c203601d5ad8e420fc26684da7994348e1eb25bf10'
+    sha256 '037a2a7a356cb5eddedbe5bbd48b7b1fa2178bf8462e0c621347dd69ccc62ae8'
     'am'
   end
 
   language 'ar' do
-    sha256 '09098d6a64dbf4f6e824bd344918772f58b30939b9f9628ee49e437a8c5792c9'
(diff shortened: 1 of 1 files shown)


miccal:
```
audit for libreoffice-language-pack: failed
 - locale ast, brx, ca-valencia, dgo, dsb, gug, hsb, kab, kmr-Latn, kok, mai, mni, nso, sat, sid, sr-Latn, szl, vec are invalid
```
I only added four of these -- the other have been there all along ...

vitorgalvao:
Either we fix the cask to conform to the audit, or we change the audit. The solution is probably somewhere in the middle, as we should be able to make this closer to the audit but then might need to have it support three letter codes.

We could also remove every language that doesn’t conform and add them correctly one by one as they are needed. I doubt all of this are in use.

miccal:
@vitorgalvao this is the only Cask using three-character locale codes, so I have just removed them all -- some of them are not even listed as valid [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2) codes.

I think we should also remove the reference to the ISO 639-2 codes in [language.md](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/language.md), and change the opening sentence to:
```
The `language` stanza can match [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) language codes, ...
```

vitorgalvao:
You’re right, I forgot about that documentation, which is now wrong. Care to open a PR (or just an issue so we don’t forget) on it?

Also, from the `audit` the languages need to be of the form `sw-TZ`, `pt-PT`, now.

miccal:
> Care to open a PR

Sure.

> Also, from the `audit` the languages need to be of the form `sw-TZ`, `pt-PT`, now.

I don't follow, they are already ... ?

vitorgalvao:
> I don't follow, they are already ... ?

My mistake. What I meant was that `pt-PT` was valid and `pt` wouldn’t be. But I just looked at the code and both are indeed valid.

miccal:
The [LibreOffice mirrors](https://download.documentfoundation.org/mirmon/) are acting up a bit at the moment, hence the CI failures -- I had a bit of trouble with them when I updated my [personal tap version of LibreOffice](https://github.com/miccal/homebrew-miccal/blob/master/Casks/m-libreoffice.rb).

PR #85365, opened 2020-07-03
CONTRIBUTING: Update to use ssh for remote

dmlemos:
Using ssh as opposed to HTTPs is a better choice for working with git repositories.
Contributor has to have ssh keys configured, but that must be assumed at this point. The use of git is a requirement anyway :)


miccal:
I think it is better to leave this as-is, which basically mirror the instructions on [docs.brew.sh](https://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request#set-up-your-own-fork-of-the-homebrew-repository).

ran-dall:
@dmlemos My colleague @miccal brings up a very good point, but in addition to that, we can't be making assumptions that everyone has to do what a `Contributor` has to do; therefore, I'm closing this.

PR #85456, opened 2020-07-06
Revert seemingly unintentional changes to package installation for Miniconda and update version

albertmichaelj:
In a previous commit (24f7d49cceb9cfcd420720de877e75c293acc447), the git commit message just said that the version was being updated, but I think the author of that commit @b0ttle actually used an old version of the file. They ended up undoing all of the great changes that @xyu had made to the cask in order to ensure that the cask installs at the standard location. This commit reverts the changes and updates the version to 4.8.3. Again, this commit is not introducing novel changes. It is reverting what seems to be unintentional changes that reverted great work done by @xyu.

### DIFF
--- a/Casks/miniconda.rb
+++ b/Casks/miniconda.rb
@@ -1,6 +1,6 @@
 cask 'miniconda' do
   version 'py37_4.8.3'
-  sha256 'bb2e3cedd2e78a8bb6872ab3ab5b1266a90f8c7004a22d8dc2ea5effeb6a439a'
+  sha256 'ccc1bded923a790cd61cd17c83c3dcc374dc0415cfa7fb1f71e6a2438236543d'
 
   # repo.anaconda.com/miniconda/ was verified as official when first introduced to the cask
   url "https://repo.anaconda.com/miniconda/Miniconda3-#{version}-MacOSX-x86_64.sh"
--- a/Casks/miniconda.rb
+++ b/Casks/miniconda.rb
@@ -1,27 +1,32 @@
 cask 'miniconda' do
-  version 'py37_4.8.2'
-  sha256 'f3ede3a58d82fb5dcbca52d291a9edb5cd962d84d823a20693dd4bb27506cdd0'
+  version 'py37_4.8.3'
+  sha256 'bb2e3cedd2e78a8bb6872ab3ab5b1266a90f8c7004a22d8dc2ea5effeb6a439a'
 
   # repo.anaconda.com/miniconda/ was verified as official when first introduced to the cask
-  url "https://repo.anaconda.com/miniconda/Miniconda3-#{version}-MacOSX-x86_64.pkg"
+  url "https://repo.anaconda.com/miniconda/Miniconda3-#{version}-MacOSX-x86_64.sh"
   name 'Continuum Analytics Miniconda'
   homepage 'https://conda.io/miniconda.html'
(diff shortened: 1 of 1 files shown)


albertmichaelj:
I just realized that this was an intentional change. I misunderstood the commit history, though I strongly prefer the way that this cask install prior to commit https://github.com/Homebrew/homebrew-cask/commit/24f7d49cceb9cfcd420720de877e75c293acc447. Also, the current method of installation for the cask is inconsistent with the cask for Anaconda, which does the use the install script. For my personal use case, I strongly prefer the cleanness of the install script and installing to the caskroom instead of a system directory. This also seems to be more consistent with how Miniconda is manually installed (which is to a .miniconda folder in the home directory).

However, given that this reversion was intentional, feel free to close this pull request if the current cask does reflect the desired behavior.

miccal:
```
audit for miniconda: failed
- download not possible: Checksum for Cask 'miniconda' does not match.
Expected: bb2e3cedd2e78a8bb6872ab3ab5b1266a90f8c7004a22d8dc2ea5effeb6a439a
Actual: ccc1bded923a790cd61cd17c83c3dcc374dc0415cfa7fb1f71e6a2438236543d
```

vitorgalvao:
> though I strongly prefer the way that this cask install prior to commit
> (…)
> For my personal use case, I strongly prefer the cleanness of the install script and installing to the caskroom instead of a system directory.

Unfortunately, a random shell script is way harder for us to maintain and cleanup than a pkg. If they had an uninstaller script, it would be another matter.

> Also, the current method of installation for the cask is inconsistent with the cask for Anaconda, which does the use the install script.

We may also argue the reverse, that the Anaconda cask is wrong and should be changed to a pkg.

> However, given that this reversion was intentional, feel free to close this pull request if the current cask does reflect the desired behavior.

It does. Thank you anyway.

albertmichaelj:
Hi @vitorgalvao . Thanks for taking the time to respond. I would like to point out that despite the pkg being presumably easier to maintain and cleanup, uninstallation doesn't actually work for the current Cask recipe. If you do `brew cask install miniconda`, and then `brew cask zap miniconda` or `brew cask uninstall miniconda`, both of those leave behind the `/opt/miniconda3` folder. If you then try to reinstall the cask with `brew cask install miniconda` you will get an error.

The previous version of this Cask recipe (and the one in my pull request) uninstalled perfectly cleanly.

The script is the way that miniconda itself seems to expect people to install because they only give uninstallation instructions for the script method. They don't even bother to give uninstallation instructions for the pkg method of installation.

I'm fine to leave this closed if everyone thinks this is the appropriate behavior, but from my perspective the current Cask recipe is broken based on the uninstallation problems and the expectations from the miniconda website.

vitorgalvao:
@albertmichaelj I’m convinced, but can’t reopen because (it seems) you removed the repo. If you want to resubmit, ping me and I’ll accept it.

Thank you for taking the time to explain.

albertmichaelj:
@vitorgalvao Thanks for reconsidering! I really appreciate you taking another look at this. I was cleaning up my git forks, and I went ahead and deleted the old version. However, I had created my own brew tap with the old version of the formula for my own use, so it was easy to just use that to create a new pull request. The new request is #86120. Please let me know there if you have any comments or necessary changes before merging.

PR #85573, opened 2020-07-08
.github/workflows: run `brew cask style`.

MikeMcQuaid:
Do this for all casks instead of a single cask.

As-of https://github.com/Homebrew/brew/pull/7867 we will need to ensure that this never fails to avoid breaking Homebrew/brew CI (and vice versa)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,9 @@ jobs:
         env:
           HOMEBREW_COLOR: 1
           HOMEBREW_DEVELOPER: 1
+      - name: brew cask style
+        run: |
+          # don't care about `brew cask style` here.
+          brew untap adoptopenjdk/openjdk
+
+          brew cask style


MikeMcQuaid:
> Run failed for master (0dadf1e)

Got an email notification about this. That's super odd/confusing because it isn't the actual commit I made. Any reason this isn't part of the `CI` actions job?

reitermarkus:
> Any reason this isn't part of the `CI` actions job?

It's because annotations weren't possible to do without a token before. I think it's possible to do now without a privileged job.

vitorgalvao:
If we’re merging both styles (and commands), wouldn’t it make sense to make `brew style` work for both casks and formulae and get rid of `brew cask style` altogether?

MikeMcQuaid:
> If we’re merging both styles (and commands), wouldn’t it make sense to make `brew style` work for both casks and formulae and get rid of `brew cask style` altogether?

Yes, I'm game for this but not in this PR. CC @whoiswillma who has been working on this

PR #85684, opened 2020-07-10
Move to using an upstream installer script in yandex-cloud-cli

andreiborisov:
Move to using installer script provided by Yandex which allows for the installation of all components of the CLI with `brew` without the need to execute additional commands after that (which are poorly documented on their part). It is also should be less brittle in case of the upstream changes and make shell completions available for the user.

In the same time, make the cask unversioned so it won't conflict with `yc` self-update and component management mechanism (the same situation as in https://github.com/Homebrew/homebrew-cask/blob/master/Casks/google-cloud-sdk.rb)


andreiborisov:
Something weird is going on with Travis, I’ve tried to force-push to fix it, but no luck

ran-dall:
@andreiborisov Thanks for the contribution. It's all good; we only require one of the CI checks to pass, which it did.

PR #85770, opened 2020-07-11
Delete opencore-configurator.rb

core-code:
this does seem to have download protection now. please double-check


miccal:
@core-code this has just been updated in https://github.com/Homebrew/homebrew-cask/pull/85743

PR #85774, opened 2020-07-11
Update acslogo

core-code:

### DIFF
--- a/Casks/acslogo.rb
+++ b/Casks/acslogo.rb
@@ -2,7 +2,8 @@ cask 'acslogo' do
   version '1.6'
   sha256 '467e4ff74265308de2262dc46d88f15841618c6fab4b4b761f32191214123a1c'
 
-  url "https://www.alancsmith.co.uk/logo/ACSLogo#{version.no_dots}.dmg"
+  url "https://www.alancsmith.co.uk/logo/ACSLogo#{version.no_dots}.dmg",
+      user_agent: :fake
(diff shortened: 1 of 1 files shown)


ran-dall:
@core-code Do you know if there's there a way to add `user agent: fake` to the `appcast` so it can pass the CIs?

core-code:
i don't think thats possible but @vitorgalvao may know better

ran-dall:
@core-code I'll merge it for now. I like to have a clean-ish PR log when I do updates later.

vitorgalvao:
> Do you know if there's there a way to add `user agent: fake` to the `appcast` so it can pass the CIs?

I don’t understand the question. If you mean a way of making the check behave as if `user_agent: :fake` was there when it isn’t, then no. But I also don’t get why you’d want to do that.

core-code:
i think the point would be to be able to set the user-agent which is used to retrieve the appcast (instead of the download URL)

vitorgalvao:
Ah, thank you for the clarification. The answer is still no, *but* what we can do is change the appcast check itself in the core to always use `:fake` (or whatever).

core-code:
just for reference - note that the only feeds i've found that require special user-agents *and* have a cask are from the techsmith apps. i  also remember it from 'transmit' but that doesn't seem to work anymore anyway.

ran-dall:
> ...what we can do is change the appcast check itself in the core to always use the fake user agent in the curl request.

That's a great idea.

PR #85779, opened 2020-07-11
Update kext-updater from 3.5.5 to 3.5.6

core-code:


ran-dall:
@core-code I'm pretty sure we have coverage for `kext-updater` with the ML. Just an FYI.

core-code:
@ran-dall sounds good. i am running a full 'cask fetch' for every cask in our repositieres every few weeks to find casks that are broken. 95% of these will be casks that have been broken some time e.g. due to URL or checksum changes. the remaining 5% is just things that would be fixed by the daily updates anyway. unfortunately there is no way to easily determine which casks are in those 5% to skip sending PRs for those...

ran-dall:
@core-code Cool. I'll update these on my end too. Good work on this.

PR #85815, opened 2020-07-12
Update avidemux from 2.7.4 to 2.7.6

hannesbraun:


core-code:
i am not sure if others agree but since this is a very popular app and catalina is still 'new' i think we might want to have an if/else to still support mojave. can you add that?

core-code:
if not we need at least a proper depends_on stanza

hannesbraun:
Oh sure, I didn't think about that. It probably makes sense to still support Mojave. I'll add it.

core-code:
good job, thanks!

PR #85850, opened 2020-07-13
Add Fing Desktop

peterlewis:

### DIFF
--- a/Casks/fing-desktop.rb
+++ b/Casks/fing-desktop.rb
@@ -10,7 +10,7 @@ cask 'fing-desktop' do
   
   uninstall launchctl: 'com.fing.service'
   
-  zap trash [
+  zap trash: [
               '~/Library/Application Support/Fing',
               '~/Library/Preferences/com.fing.app.plist',
               '~/Library/Saved Application State/com.fing.app.savedState',
--- a/Casks/fing-desktop.rb
+++ b/Casks/fing-desktop.rb
@@ -7,12 +7,12 @@ cask 'fing-desktop' do
   homepage 'https://www.fing.com/products/fing-desktop'
 
   app 'Fing.app'
-  
+
   uninstall launchctl: 'com.fing.service'
-  
+
   zap trash: [
-              '~/Library/Application Support/Fing',
-              '~/Library/Preferences/com.fing.app.plist',
-              '~/Library/Saved Application State/com.fing.app.savedState',
+               '~/Library/Application Support/Fing',
+               '~/Library/Preferences/com.fing.app.plist',
+               '~/Library/Saved Application State/com.fing.app.savedState',
              ]
 end
--- a/Casks/fing-desktop.rb
+++ b/Casks/fing-desktop.rb
@@ -13,6 +13,6 @@ cask 'fing-desktop' do
   zap trash: [
               '~/Library/Application Support/Fing',
               '~/Library/Preferences/com.fing.app.plist',
-              '~/Library/Saved Application State/com.fing.app.savedState'
+              '~/Library/Saved Application State/com.fing.app.savedState',
              ]
 end
--- a/Casks/fing-desktop.rb
+++ b/Casks/fing-desktop.rb
@@ -1,0 +1,10 @@
+cask 'fing-desktop' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://get.fing.com/fing-desktop-releases/mac/Fing.dmg'
+  name 'Fing Desktop'
+  homepage 'https://www.fing.com/products/fing-desktop'
+
+  app 'Fing.app'
+end
--- a/Casks/fing-desktop.rb
+++ b/Casks/fing-desktop.rb
@@ -7,4 +7,12 @@ cask 'fing-desktop' do
   homepage 'https://www.fing.com/products/fing-desktop'
 
   app 'Fing.app'
+  
+  uninstall launchctl: 'com.fing.service'
+  
+  zap trash [
+               '~/Library/Application Support/Fing',
+               '~/Library/Preferences/com.fing.app.plist'
+               '~/Library/Saved Application State/com.fing.app.savedState',
+            ]
 end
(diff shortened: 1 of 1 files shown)


peterlewis:
Cask fails on `brew cask audit --new-cask fing-desktop` as it includes the word 'desktop'. However, this is necessary as Fing offers two products:

- Fing CLI (currently named 'fing' within Homebrew)
- Fing Desktop

FYI, if agreeable to maintainers, I am planning on submitting additional PR to rename 'fing' to 'fing-cli'.

Casks/fing-desktop.rb
```diff
@@ -0,0 +1,18 @@
+cask 'fing-desktop' do
+  version :latest
+  sha256 :no_check
```
miccal, line 3: Please change this to:
```
version '2.0.0'
sha256 'fceb7cf920214f447333e80bf7b141f67661a423eda03041bc651e3323763810'
```

Casks/fing-desktop.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  url 'https://get.fing.com/fing-desktop-releases/mac/Fing.dmg'
```
miccal, line 5: Please add the following `appcast`:
```
appcast 'https://get.fing.com/fing-desktop-releases/mac/latest-mac.yml'
```

peterlewis: thanks, done!

Casks/fing-desktop.rb
peterlewis: Sure thing! Done! Good to know, thanks!

vitorgalvao:
> Cask fails on `brew cask audit --new-cask fing-desktop` as it includes the word 'desktop'. However, this is necessary

It isn’t. Name it `fing`.

vitorgalvao (changes requested):
> Cask fails on `brew cask audit --new-cask fing-desktop` as it includes the word 'desktop'. However, this is necessary

It isn’t. Name it `fing`.

peterlewis:
> > Cask fails on `brew cask audit --new-cask fing-desktop` as it includes the word 'desktop'. However, this is necessary
> 
> It isn’t. Name it `fing`.

OK, I'm happy to adjust the name to 'fing'. This follows the [renaming of the existing 'fing' cask to 'fing-cli'](https://github.com/Homebrew/homebrew-cask/pull/85851).

Can you advise whether this will have any impact on existing installations of 'fing' cask (now called fing-cli)? I.e. will they find themselves with an entirely different application installed, due to it using the old cask's name? Sorry, not entirely aux-fait with the underlying mechanics, yousee...

Thanks!

Peter

vitorgalvao:
It may impact current installations on `upgrade`, as we still lack a proper way to rename casks. Unsure on how much, would have to check. But it’s not like we have a better way at the moment.

vitorgalvao:
@peterlewis By the way, thank you for the continued contributions and for being a nice user in general.

peterlewis:
Hmm... Whilst ppl who have the Fing Desktop Embedded CLI installed may also want Fing Desktop, for it to become installed like that doesn't sound ideal. Shall we pause on this for now?

peterlewis:
> @peterlewis By the way, thank you for the continued contributions and for being a nice user in general.

Kind of you to say, thank you. It's a pleasure! I'm enjoying contributing to Open Source after too many years of simply consuming! :)

peterlewis:
Ok. So went through the following troubleshooting steps:

1. `brew cask install pl-fing`
2. Renamed `pl-fing` to `pl-fing-cli`
3. Added 'Fing Desktop' as `pl-fing`
4. `brew update && brew cask upgrade`

The outcome of the above was that the existing Fing CLI installation (that was renamed from `fing` to `fing-cli`) became orphaned and Fing Desktop (which was introduced to the repo as `fing`) was installed, following `brew cask upgrade`.

Not sure what the right thing to do here is, tbh... :-/

ran-dall:
@peterlewis Can you "retrace" your steps and undo it? I don't think there a simple command I could give you that would fix this. "Retracing" is probably your best bet.

vitorgalvao:
The only thing that’s done is the renaming on the CLI version, and that’s fine. Let’s give it a while so people have time to move over, then merge this one, or add this one with `-desktop` and remove the suffix later.

peterlewis:
> The only thing that’s done is the renaming on the CLI version, and that’s fine. Let’s give it a while so people have time to move over, then merge this one, or add this one with `-desktop` and remove the suffix later.

Sounds good to me. Thanks for your understanding/flexibility on this!

vitorgalvao:
Merging around the 20<sup>th</sup>.

peterlewis:
@vitorgalvao Apologies. I had misread the terminal output (which mentioned nothing about uninstalling the older cask) but I've just re-tested it and can confirm that my testing shows that the older cask is _uninstalled_ and not orphaned.

I've updated my previous testing notes to reflect this.

peterlewis:
Just to further add that whilst I personally feel it would be preferred to add this cask as `fing-desktop` for now, so as to avoid unintentional uninstallation of 'Fing CLI' and installation of 'Fing Desktop', I understand why the maintainers may prefer to add it as `fing`, so as to avoid introducing inconsistencies to Homebrew Cask.

If it is decided to add to Homebrew Cask as `fing`, would it be worth considering adding a caveat for a period of time, something along the lines of the following?

```
This cask for 'Fing Desktop' uses the same cask name (`fing`) that was used for 'Fing CLI' (renamed to 'fing-cli'). If you installed the 'Fing CLI' cask prior to it being renamed, then you may find that Fing CLI is uninstalled and replaced with 'Fing Desktop'. You can reinstall 'Fing CLI' via the `fing-cli` cask.
```

I appreciate the above caveat is somewhat verbose and would likely require editing prior to its inclusion, but just wanted to provide the suggestion.

vitorgalvao:
Caveats like that end up being forgotten and never dealt with. Our current solution is bad but we don’t have anything better yet (a proper way to rename). Sticking to the original plan.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask/pull/85850#issuecomment-662433374

reitermarkus:
@vitorgalvao, I think this should still be called just `fing`.

vitorgalvao:
> @vitorgalvao, I think this should still be called just `fing`.

You’re right, that’s what I meant to do, I thought we had already made the change. Will fix directly.

PR #85913, opened 2020-07-14
Update airdisplay from 3.3.3 to 3.4.2

suschizu:

### DIFF
--- a/Casks/airdisplay.rb
+++ b/Casks/airdisplay.rb
@@ -7,7 +7,7 @@ cask 'airdisplay' do
   name 'Air Display'
   homepage 'https://avatron.com/applications/air-display/'
 
-  pkg "ad#{version.no_dots}.pkg"
+  pkg 'Air Display Installer.pkg'
 
   uninstall pkgutil: [
                        'com.avatron.pkg.AirDisplay',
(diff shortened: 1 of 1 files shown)


suschizu:
xcode10.1 fail: 
```
Error: Failure while executing; `/usr/bin/sudo -E -- /usr/bin/env LOGNAME=travis USER=travis USERNAME=travis /usr/sbin/installer -pkg /usr/local/Caskroom/airdisplay/3.4.2/Air\ Display\ Installer.pkg -target / -verboseR` exited with 1. Here's the output:
installer: Cannot install on volume / because it is disabled.
installer: macOS High Sierra versions 10.13.4–10.13.6 are not compatible with virtual video displays. Please contact support@avatron.com for more information.
```

miccal:
From the `homepage`:

Screen Shot 2020-07-15 at 07 21 33

suschizu:
@miccal - oke thanks - this makes sense  … but is it possible to ignore this exception?

miccal:
The simplest way would be to add:
```
depends_on macos: '>= :mojave'
```
but that might be a bit too restrictive?

suschizu:
> The simplest way would be to add:
> 
> ```
> depends_on macos: '>= :mojave'
> ```
> 
> but that might be a bit too restrictive?

no I think that's the best way to solve this issue :)

PR #86024, opened 2020-07-15
Update fly from 6.1.0 to 6.3.0

therealmarv:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #86027, opened 2020-07-15
Fix CI

SeekingMeaning:
Not sure if this works

Related: #86024


ran-dall:
@SeekingMeaning Thanks for the PR, however this is really doing is working around the problem; as opposed to addressing it. I've seen this happen before, and it normally happens because something is not running correctly inside the CI. I would rather take some time tomorrow and try to find the problem before merging a workaround.

PR #86070, opened 2020-07-16
Fix GitHub Actions’s `Please tell me who you are` Fail

ran-dall:
Fixes the GitHub Action's Failure


ran-dall:
@SeekingMeaning Turns out your way, is the only way I find to address this...for now. That said, it's better to use the default GitHub Action user, which I think should have already be set, but isn't for some reason.

PR #86177, opened 2020-07-18
Update lando from 3.0.8 to 3.0.9

philltran:

### DIFF
--- a/Casks/lando.rb
+++ b/Casks/lando.rb
@@ -1,6 +1,6 @@
 cask 'lando' do
-  version '3.0.8'
-  sha256 'b2b966f15dc87372e0c54d2fb883f4ad1e193e4f9d76f93cc970a982ea3fd20d'
+  version '3.0.9'
+  sha256 '4a4d4c90e326f706554b52be58af1b997857a4f519f1b90108df1492abfc8f10'
 
   # github.com/lando/lando/ was verified as official when first introduced to the cask
   url "https://github.com/lando/lando/releases/download/v#{version}/lando-v#{version}.dmg"


core-code:
@philtran thanks for taking care of updates for 'lando' but both version 3.0.8 and 3.0.9 are marked as 'pre-release' in GitHub and therefore not suitable for inclusion in Homebrew Cask. we only want the last stable official version. here is more info on the 'lando' release cycle: https://docs.lando.dev/config/releases.html#toggling-channels   . if you personally need the 'edge' releases of lando please subit an 'lando-edge' cask over at 'homebrew-cask-versions' where unstable releases are allowed. i an going to submit a PR to revert this to 3.0.7 now

philltran:
Hi

Sorry about that. Thanks for clarifying. I will make sure to submit pre-release versions to homebrew-cask-versions. 

I appreciate your help. 

Have a wonderful day.

core-code:
thanks!

philltran:
No thank you. I want to help and not be a pain. 

Have a wonderful weekend.

core-code:
> No thank you. I want to help and not be a pain.

no pain at all, thanks for your contributions

> Have a wonderful weekend.

thanks you too

vitorgalvao:
> I will make sure to submit pre-release versions to homebrew-cask-versions.

Before you do, run a `brew cask info` on the stable cask and check the analytics numbers. If they’re low, they don’t justify the maintenance effort of a new one in homebrew/cask-versions and would be rejected.

We don’t have a definitive number for what constitutes “low”, as it varies with the cask’s complexity (the more complex it is, the higher its maintenance cost, thus the higher its analytics numbers need to be).

As a rough guideline, if download numbers come up to less than one or two downloads per day on the stable cask, it’s likely not worth submitting an alternate version.

To anyone stumbling here in the future, that’s not official policy. Please don’t link here later as an argument against a refusal.

PR #86186, opened 2020-07-18
Fix incorrect auto_update method call in ableton*

lucsky:
Casks definitions were using `auto_update` instead of `auto_updates`.


lucsky:
Commented on the initial commit (9a217a528548c6f4c504345f14bc6046dbbf84aa) which introduced those errors and let the original author fix those.

PR #86216, opened 2020-07-18
Change container image for `BrewTestBot`

ran-dall:
This fixes the `BrewTestBot` failure because the GitHub Action runs as the `root` user by default; which isn't permitted by Homebrew any longer.

```shell
...
 Step 4/9 : RUN brew update-reset
 ---> Running in 86e8e5610ff9
Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.
The command '/bin/sh -c brew update-reset' returned a non-zero code: 1

##[error]Docker build failed with exit code 1
```

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   automerge:
     if: startsWith( github.repository, 'Homebrew/' )
-    runs-on: ubuntu-latest
+    runs-on: homebrew/ubuntu16.04:master
     steps:
       - uses: actions/checkout@master
       - uses: ./.github/actions/automerge


issyl0 (commented):
Homebrew was fixed to allow the new Actions runners in https://github.com/Homebrew/brew/pull/8009, but it hasn't made it into a stable release yet. I don't know how Cask CI works, but switching to the main branch of `Homebrew/brew` could save you some pain?

ran-dall:
@issyl0 THANK YOU ISSY! This is an outstanding idea. However how could I do that? From what it looks like, we're pulling the [`homebrew/brew` container](https://hub.docker.com/r/homebrew/brew). Wouldn't that have to moved over to the main branch?

Bo98:
The Ubuntu 16.04 image has a `master` tag: `homebrew/ubuntu16.04:master`

`latest` refers to latest stable tag.

issyl0:
What I assumed was happening was that Cask CI was using a tagged release of Homebrew instead of the bleeding edge `master`. Now I've looked into it further, turns out that Action runs through Docker. Behaviour on my local machine is that [the Dockerfile for that automerge Action](https://github.com/Homebrew/homebrew-cask/blob/master/.github/actions/automerge/Dockerfile) pulls `homebrew/brew:latest` which... oh, which doesn't have the changes from the PR I linked above. That said, since it's via Docker, it shouldn't complain about being root because Docker was allowed prior to that PR. Hmmm. 🤔 

So nothing works how I thought it did and computers are hard. 😄

ran-dall:
@issyl0 I tried it locally and it seems to build for me.

@Bo98 So should I change it over? But isn't running the CI under a user a better option though?

reitermarkus:
> The Ubuntu 16.04 image has a `master` tag: `homebrew/ubuntu16.04:master`

Let's use the `homebrew/brew:master` image then.

Bo98:
> But isn't running the CI under a user a better option though?

I'm perhaps not the best person to ask, but I believe the container intentionally uses root by default because it's more standard for Docker containers to do that. linuxbrew-core CI runs fine under this.

> Should there also be a `homebrew/brew:master` image?

There isn't one because no one has needed it to date - and it's cheaper on CI to only offer what is necessary. linuxbrew-core runs on `homebrew/ubuntu16.04:master`.

`homebrew/ubuntu18.04` and `homebrew/ubuntu20.04` are available (without `master`) and `homebrew/brew` is the same as 20.04 at the moment.

ran-dall:
Anyone else notice that `BrewTestBot` started working again ~12 hours ago? Would that mean a change is no longer required and we can close this up?

issyl0:
Yep, [release 2.4.8 of Homebrew](https://github.com/Homebrew/brew/releases/tag/2.4.8) shipped 13 hours ago, with the fix for this problem.

reitermarkus:
We should still use the `master` image for CI.

ran-dall:
@reitermarkus Alright, I'll take care of it then.

ran-dall:
@reitermarkus Can you take a look 👀 at it? and merge if everything looks good?

Issue #86221, opened 2020-07-19
Do not overwrite files in /usr/local/bin without warning when installing MacVim.app


williamsmj:
### Description of feature/enhancement

During `brew cask install macvim`, symlinks to /Applications/MacVim.app/Contents/bin/mvim are installed in /usr/local/bin:

```
/Applications/MacVim.app/Contents/bin/mvim -> gview (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> gvim (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> gvimdiff (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> mview (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> mvimdiff (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> view (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> vim (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> vimdiff (Binary)
/Applications/MacVim.app/Contents/bin/mvim -> mvim (Binary)
```

1. There should be a warning about this if it will overwrite existing files (which it will if the user has previously installed vim)
2. Creating these symlinks is a reasonable default, but it should be optional.

### Justification

1. There must be a warning. These symlinks overwrite existing files. Depending on how the files that get overwritten were created, this may result in data loss. (In that sense, this is a bug report, not a feature request.)
2. Creating these links should be optional. There are lots of situations in which a user might not want these symlinks to exist (e.g. they have already installed vim and don't want `vim` on the command line to run the build of vim that happens to be bundled in MacVim.app).

vitorgalvao:
Making it optional isn’t a good solution, because that requires yet another option to support at installation time for a situation that almost never happens.

https://github.com/Homebrew/homebrew-cask/pull/86222

williamsmj:
Thanks for the quick response. I think the solution in #86222 is a shame.

There is no reason from the user's point of view to forbid the coexistence of the vim brew and the macvim cask, other than the cask's insistence on overwriting the brew's files, which is something MacVim.app does not need to do in order to function. Wanting both brew vim and macvim cask is a pretty reasonable request. While macvim comes with a handy GUI, it uses an older version of vim itself and is generally kept less up to date.

[...]

PR #86222, opened 2020-07-19
Update macvim conflicts_with

miccal:
Updated `conflicts_with`:

- Conflicts with formula [`ex-vi`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ex-vi.rb) since `ex-vi` also installs a `bin/view` binary.
- Conflicts with formula [`macvim`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/macvim.rb), obviously.
- Conflicts with formula [`vim`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/vim.rb) since `vim` also installs `bin/vi*` binaries.

Adresses https://github.com/Homebrew/homebrew-cask/issues/86221

### DIFF
--- a/Casks/macvim.rb
+++ b/Casks/macvim.rb
@@ -8,7 +8,11 @@ cask 'macvim' do
   homepage 'https://github.com/macvim-dev/macvim'
 
   auto_updates true
-  conflicts_with formula: 'macvim'
+  conflicts_with formula: [
+                            'ex-vi',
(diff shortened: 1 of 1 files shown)


vitorgalvao:
This is the similar to [the solution used in the formula](https://github.com/Homebrew/homebrew-cask/pull/86222), so thumbs up for consistency.

PR #85691, opened 2020-07-10
Add Apparency.app v1.0

silver886:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Additionally, **if adding a new cask**:

- [x] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] `brew cask audit --new-cask {{cask_file}}` worked successfully.
- [x] `brew cask install {{cask_file}}` worked successfully.
- [x] `brew cask uninstall {{cask_file}}` worked successfully.
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).

Issue #62223, opened 2019-04-20
New strategy for `.metadata`


vitorgalvao:
When we install a cask, we keep a copy of it at that moment in a `.metadata` directory. This method causes problems when we remove a feature from the core, as the old casks throw errors when read back. An example of widespread breakage due to this method is https://github.com/Homebrew/homebrew-cask/issues/58046.

https://github.com/Homebrew/brew/pull/4730#issuecomment-415322677 suggests a few alternative to the current method. The first one is the most viable: when reading a cask in `.metadata`, if it gives an error, ignore it and read the current cask.

reitermarkus:
> The only thing we need from the cask copy is the uninstall logic.

Actually, we need everything. In case an upgrade fails, we should still be able to reinstall the old version.

vitorgalvao:
> Actually, we need everything. In case an upgrade fails, we should still be able to reinstall the old version.

Edited the top post to only reference the first option.

vitorgalvao:
I’ve been thinking a bit more about this. The top post include another suggestion:

> The only thing we need from the cask copy is the uninstall logic. When installing, instead of keeping a copy of the cask, create an uninstall bash script. We already run system commands via ruby to perform the uninstallation anyway, so might as well save those commands.

That was removed due to the point that [we need everything, in the case of an upgrade fail](https://github.com/Homebrew/homebrew-cask/issues/62223#issuecomment-485206817). But how often does that happen? We could just as well do both. We always uninstall from the shell script, and only reference the `.metadata` cask if an upgrade fails. At which point, if that fails we’re back where we started, but in the way we’ve already got rid of most errors by offloading uninstall to script.

MikeMcQuaid:
> [Homebrew/brew#4730 (comment)](https://github.com/Homebrew/brew/pull/4730#issuecomment-415322677) suggests a few alternative to the current method. The first one is the most viable: when reading a cask in `.metadata`, if it gives an error, ignore it and read the current cask.

I think we should do this. We're getting more and more reports of this and the status-quo is pretty untenable and undesirable. Any objections?

vitorgalvao:
> I think we should do this. We're getting more and more reports of this and the status-quo is pretty untenable and undesirable. Any objections?

None whatsoever. We should also think about what to do if the failing cask no longer exists. My suggestion would be to inform the user they’ll need to uninstall it manually, `cat` the cask, and remove its directory from the Caskroom (so it no longer shows as installed).

[...]

Issue #40557, opened 2017-11-10
Large zap updates


miccal:
I have noticed (under MacOS 10.13 at least) that there has been a slight change the filenames stored in the folders
```
~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/
```
and
```
~/Library/Caches/com.apple.helpd/SDMHelpData/Other/English/HelpSDMIndexFile/
```
For example, on my system:
```
-bash-4.4.12- /Users/miccal/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments (9) [> ls -l1
com.apple.ichat.sfl2
com.apple.preview.sfl2
com.apple.textedit.sfl2
com.github.atom.sfl2
org.libreoffice.script.sfl2
org.m0k.transmission.sfl2
texshop.sfl2
```
and:
```
-bash-4.4.12- /Users/miccal/Library/Caches/com.apple.helpd/SDMHelpData/Other/English/HelpSDMIndexFile (5) [> ls -l1
TeX Live Utility Help*1.30
TeXShop Help*3.88
Transmission Help*2.92
```

Notice that the files in 
```
~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/
```
now have extension `.sfl2` instead of the usual `.sfl`, and the files in 
```
~/Library/Caches/com.apple.helpd/SDMHelpData/Other/English/HelpSDMIndexFile/
```
now have `*{version}` added to the end.

Now, looking at the `zap` stanza for `texshop`/`mactex` as an example, which originally were:
```
zap delete: [
              '~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/texshop.sfl',
              '~/Library/Caches/com.apple.helpd/SDMHelpData/Other/English/HelpSDMIndexFile/TeXShop Help',
              [snip]
```
I recently changed them to
```
zap delete: [
              '~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/texshop.sfl*',
              '~/Library/Caches/com.apple.helpd/SDMHelpData/Other/English/HelpSDMIndexFile/TeXShop Help*',
              [snip]
```

[...]

PR #86432, opened 2020-07-23
Make `zap` a requirement

ran-dall:
Updated `zap` and related documentation to make `zap` a requirement.

ping @vitorgalvao @core-code @miccal @nrlquaker 

Comments and changes welcome...


miccal:
Below are my thoughts on this:

- I don't think `appcast` should be moved from _Required_ to _Optional_, as there are many casks for which there is no suitable `appcast`; and of course there are casks that use `version :latest` and `sha256 :no_check` and thus by definition do not have an `appcast`.

- For `zap` the _simplest_ way to find any user files and the shared resources associated with a particular cask in my opinion is to run the command `sudo find / -iname "*{search item}*"` and an inspection of the output (well, at least that is how I do it). I do think it is fine to also mention the "CreateZap Project" and an app such as "AppCleaner" as other examples.

- As for making the `zap` _Required_ for new casks, I do not agree with this. In my years of using `homebrew-cask` I have found that the `zap` stanza is greatly influenced by the macOS version and the version of the cask, and can change quite a lot -- see [this issue](https://github.com/Homebrew/homebrew-cask/issues/40557), for example. I have also found that one has to use a particular cask for a while (involving opening the app, creating new files, opening existing files, saving new and existing files, exporting, etc.) before one would get a _complete_ list of user files and shared resources associated with a particular cask -- the many contributors to `homebrew-cask` (and the other cask repo's) already do a great job of adding and /or updating the `zap` stanza based on their personal experience of using the app, and this should continue to be encouraged.

miccal:
Here is an example of my thoughts behind `zap` [this PR](https://github.com/Homebrew/homebrew-cask/pull/85691) added a new cask for `Apparency.app`, a wonderful and interesting app that I have been using since it was added. Running `sudo find` just now revealed [another file](https://github.com/Homebrew/homebrew-cask/pull/86443) that needed to be added to `zap`.

core-code:
i agree that appcasts are absolutely essential to keep casks updated. that said, sometimes there are just none...

regarding the zap stanza, i think we've already talked about the idea that if homebrew knew the bundle-identifier of the installed app, all the 'standard' locations could be zapped automatically. i think this would be great. but unless someone is willing to actually implement this...

ran-dall:
> i think we've already talked about the idea that if homebrew knew the bundle-identifier of the installed app, all the 'standard' locations could be zapped automatically

I like this idea.

doc/cask_language_reference/stanzas/zap.md
```diff
@@ -22,3 +22,7 @@ $ brew cask zap td-toolbelt             # also removes org.ruby-lang.installer
 The form of `zap` stanza follows the [`uninstall` stanza](uninstall.md). All of the same directives are available. The `trash:` key is preferred over `delete:`.
 
 Example: [dropbox.rb](https://github.com/Homebrew/homebrew-cask/blob/31cd96cc0e00dab1bff74d622e32d816bafd1f6f/Casks/dropbox.rb#L17-L35)
+
+## zap Creation
+
+There are several ways to create a `zap` stanza. The simplest method is to use @nrlquaker's [CreateZap Project](https://github.com/nrlquaker/homebrew-createzap), which can automatically generate the stanza. In the few instances, it fails to pick anything up; manual creation may be required. Manual creation can be facilitated with some of the developer tools that are already available in Homebrew Cask or an uninstaller tool (such as [AppCleaner](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/appcleaner.rb)).
```
vitorgalvao, line 28: The first sentence can be cut and the paragraph will retain the meaning.

doc/cask_language_reference/all_stanzas.md
```diff
@@ -9,8 +9,10 @@ Each of the following stanzas is required for every Cask.
 | `version`          | no                            | application version; give value of `:latest`  if versioned downloads are not offered (see also [Version Stanza Details](stanzas/version.md))
 | `sha256`           | no                            | SHA-256 checksum of the file downloaded from `url`, calculated by the command `shasum -a 256 <file>`. Can be suppressed by using the special value `:no_check` (see also [Checksum Stanza Details](stanzas/sha256.md))
 | `url`              | no                            | URL to the `.dmg`/`.zip`/`.tgz`/`.tbz2` file that contains the application.<br />A [comment](stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) should be added if the hostnames in the `url` and `homepage` stanzas differ. Block syntax should be used for URLs that change on every visit.<br />See [URL Stanza Details](stanzas/url.md) for more information.
+| `appcast`          | no                            | a URL providing an appcast feed to find updates for this Cask (see also [Appcast Stanza Details](stanzas/appcast.md))
```
vitorgalvao, line 12: This part needs to be reverted. A required stanza isn’t one that we want, but one without which a cask will not function and will fail `audit`.

That doesn’t apply to neither `appcast` nor `zap`, and it never can because there are situations when indeed we don’t have those.

vitorgalvao (changes requested):

vitorgalvao:
> I was under the impression that `appcast`s are now required, unless it's a special case

Heavily encouraged, but not required.

> I know it's not a 'hard' rule, but it's a harder rule than `optional`

As hinted in the review, it’s optional in the sense it’s not mandatory for the cask to be functional, not in the sense of “do it only if you want to”.

> and I think people should make an effort to find the `appcast` or ask for help in finding it.

I do too, but I don’t think changing its position on that list will accomplish that. Adding the appcast in the PR itself, before being merged, is an effective step we take.

> I still believe in practice we should have a way of getting rid of what we install in a end-users computer.

I agree with the sentence, but I don’t agree that’s the reality. We’re not the ones installing the preference files, the apps are.

ran-dall:
> I agree with the sentence, but I don’t agree that’s the reality. We’re not the ones installing the preference files, the apps are.

I see what you're saying, but the apps are there because they were installed with HBC. So I'd argue there's a level of 'cause' to the 'effect' that separates us from being directly responsible; but I still feel that we're the force that pushed the events forwarded. 

Furthermore, if a user where to remove a `cask`, then removing the remaining 'support' files become exponentially harder because you can't use something like 'AppCleaner' to remove them. You'd have to use something like 'CleanMyMac X' or 'Onyx' and hope they pick up the orphans.

doc/cask_language_reference/all_stanzas.md
```diff
@@ -11,6 +11,7 @@ Each of the following stanzas is required for every Cask.
 | `url`              | no                            | URL to the `.dmg`/`.zip`/`.tgz`/`.tbz2` file that contains the application.<br />A [comment](stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment) should be added if the hostnames in the `url` and `homepage` stanzas differ. Block syntax should be used for URLs that change on every visit.<br />See [URL Stanza Details](stanzas/url.md) for more information.
 | `name`             | yes                           | a string providing the full and proper name defined by the vendor (see also [Name Stanza Details](stanzas/name.md))
 | `homepage`         | no                            | application homepage; used for the `brew cask home` command
+| `zap`              | yes                           | additional procedures for a more complete uninstall, including user files and shared resources (see also [Zap Stanza Details](stanzas/zap.md))
```
vitorgalvao, line 14: I meant both. Even if we start requesting `zap`s on submissions, “required” should mean “the simplest possible cask”.

vitorgalvao:
> Furthermore, if a user where to remove a `cask`, then removing the remaining 'support' files become exponentially harder because you can't use something like 'AppCleaner' to remove them. You'd have to use something like 'CleanMyMac X' or 'Onyx' and hope they pick up the orphans.

That’s a good point. Though we still fail there—you can’t `zap` an uninstalled cask, but we should be able to.

ran-dall:
I like @core-code's idea and it's something I'll be looking into the upcoming days. Still I think making `zap` required now it a step in the right direction.

That said, should I look into doing some modifications so that we can run `zap` even if the cask isn't installed?

vitorgalvao:
> should I look into doing some modifications so that we can run `zap` even if the cask isn't installed?

That would be great, thank you. Note that `zap` implicitly calls `uninstall` so the goal here would be to, if `zap` is called but the cask isn’t installed, skip `uninstall` and proceed to `zap` only.

`zap` can do everything `uninstall` can. I f that becomes a problem, if would be fine to make it so that a `zap` on an uninstalled cask only does the path operations (`rmdir`, `delete`, `trash`).

core-code:
> That would be great, thank you. Note that zap implicitly calls uninstall so the goal here would be to, if zap is called but the cask isn’t installed, skip uninstall and proceed to zap only.

why? wouldn't the ability to 'uninstall' *also* be benificial even if the cask isn't installed? i only ever install manually and the removal of launchctl jobs and other things would also be great without actually having to install it...

vitorgalvao:
> wouldn't the ability to 'uninstall' _also_ be benificial even if the cask isn't installed?

Our uninstallation isn’t robust enough for that. As is it, I wouldn’t even say we’re good at uninstall what we install (https://github.com/Homebrew/homebrew-cask/issues/62223). That’s why I mentioned:

> If that becomes a problem, if would be fine to make it so that a `zap` on an uninstalled cask only does the path operations (`rmdir`, `delete`, `trash`).

doc/cask_language_reference/stanzas/zap.md
```diff
@@ -22,3 +22,7 @@ $ brew cask zap td-toolbelt             # also removes org.ruby-lang.installer
 The form of `zap` stanza follows the [`uninstall` stanza](uninstall.md). All of the same directives are available. The `trash:` key is preferred over `delete:`.
 
 Example: [dropbox.rb](https://github.com/Homebrew/homebrew-cask/blob/31cd96cc0e00dab1bff74d622e32d816bafd1f6f/Casks/dropbox.rb#L17-L35)
+
+## zap Creation
+
+The simplest method is to use @nrlquaker's [CreateZap Project](https://github.com/nrlquaker/homebrew-createzap), which can automatically generate the stanza. In the few instances, it fails to pick anything up; manual creation may be required. Manual creation can be facilitated with some of the developer tools that are already available in Homebrew Cask, using `sudo find / -iname "*{search item}*"`, or an uninstaller tool (such as [AppCleaner](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/appcleaner.rb)).
```
claui, line 28:
```suggestion
The simplest method is to use @nrlquaker's [CreateZap Project](https://github.com/nrlquaker/homebrew-createzap), which can automatically generate the stanza. In a few instances, it fails to pick anything up so manual creation may be required.

Manual creation can be facilitated with:
- some of the developer tools that are already available in Homebrew Cask,
- `sudo find / -iname "*{search item}*"`,
- an uninstaller tool (such as [AppCleaner](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/appcleaner.rb))
- or visually inspecting the usual suspects, i.e. `/Library/{'Application Support',LaunchAgents,LaunchDaemons,Frameworks,Logs,Preferences,PrivilegedHelperTools}` and `~/Library/{'Application Support',Caches,Containers,LaunchAgents,Logs,Preferences,'Saved Application State'}`.
```

claui:
> you can’t `zap` an uninstalled cask, but we should be able to.

@vitorgalvao I think we can do it with `zap --force`. I use it regularly while working on `zap` stanza entries.

vitorgalvao:
These instructions don’t suggest `zap` is mandatory, but encourage it and explain how to make it. I don’t expect there to be controversy in merging this.

vitorgalvao (approved):

PR #86613, opened 2020-07-27
Remove Travis CI

MikeMcQuaid:
GitHub Actions is now used instead.


vitorgalvao (approved):
I agree with only having one CI and making it GitHub’s, but I’ll point out that lately it has been the more unreliable of the two, in case someone has experienced the same and wants to comment on it / knows why / has a solution.

core-code:
i don't have any hard numbers but i think i can confirm that the Github Actions are even more (!) unreliable than Travis.

however since currently CI fails if *any* of the checks fails, we are still better off even if we remove the more reliable one.

MikeMcQuaid:
> I agree with only having one CI and making it GitHub’s
> however since currently CI fails if _any_ of the checks fails, we are still better off even if we remove the more reliable one.

👍🏻 Merging given this for now.

reitermarkus:
Reverting this since this broke CI and `automerge` still relies on Travis.

PR #86619, opened 2020-07-27
Update saoimageds9.rb

core-code:
get rid of non-standard syntax. please double check this is what we want.

### DIFF
--- a/Casks/saoimageds9.rb
+++ b/Casks/saoimageds9.rb
@@ -1,11 +1,10 @@
 cask "saoimageds9" do
   version "8.1"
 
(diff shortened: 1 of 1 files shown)


Casks/saoimageds9.rb
```diff
@@ -1,11 +1,10 @@
 cask "saoimageds9" do
```
miccal, line 1: Might be wise to add a comment as is done in other casks:
```
# note: "9" is not a version number, but an intrinsic part of the product name
```

core-code: good idea!

Issue #86628, opened 2020-07-27
kafka-tool cask: application does not work when JDK11 is already installed


timothybasanov:
#### Description of issue

Kafka Tool cask installs the app successfully and wars that it needs JDK8 to run correctly.
I already have JDK11 installed.
When I try to run an app it fails with some of the rendering of data.

Installing the same application from the web site works just fine as application packages JRE8 already.

#### Command that failed
```sh
brew cask install kafka-tool
```

vitorgalvao:
> Kafka Tool cask installs the app successfully and wars that it needs JDK8 to run correctly.
I already have JDK11 installed.

It says it needs “8”, nor “8 or later”.

> Installing the same application from the web site works just fine as application packages JRE8 already.

If you can update the cask to use that app instead, that’d work.

> When I try to run an app it fails with some of the rendering of data.

But otherwise there’s nothing we can do. The java version warning is just that, a warning. It’s up to the user to fulfil the requirement.

timothybasanov:
I think that cask already downloads the right DMG. 
Not sure what is the difference that causes this issue.

miccal:
@timothybasanov can you please test if the [changes I made](https://github.com/Homebrew/homebrew-cask/pull/86632) work for you, thanks.

PR #86632, opened 2020-07-28
Update kafka-tool from latest to 2.0.7

miccal:

From the `homepage`:

> All versions of Kafka Tool come with a bundled JRE with the exception of the Linux version.

so the `depends_on_java "8"` should not be required.

Addresses https://github.com/Homebrew/homebrew-cask/issues/86628

### DIFF
--- a/Casks/kafka-tool.rb
+++ b/Casks/kafka-tool.rb
@@ -10,7 +10,6 @@ cask "kafka-tool" do
   installer script: {
     executable: "Kafka Tool #{version.major} Installer.app/Contents/MacOS/JavaApplicationStub",
     args:       ["-q"],
-    sudo:       true,
   }
 
   uninstall delete: "/Applications/Kafka Tool.app"
--- a/Casks/kafka-tool.rb
+++ b/Casks/kafka-tool.rb
@@ -8,12 +8,12 @@ cask "kafka-tool" do
   homepage "https://www.kafkatool.com/index.html"
 
   installer script: {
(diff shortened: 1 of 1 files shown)


miccal:
I believe this is timing-out due to the addition of `sudo: true` to the `installer script:`.

core-code:
can you change the appcast to the download page?

https://www.kafkatool.com/download.html

chanlog-sites are the worst possible kind of appcasts since they also contain the version numbers of the old releases. 

this effectively makes a check that sees whether the version number of the cask is contained in the appcast to detect updates impossible. and for HTML pages this is often the only possible check since they are often too noisy to observe for changes.

miccal:
> I believe this is timing-out due to the addition of `sudo: true` to the `installer script:`.

I have tried removing `sudo: true` to the `installer script:` and the CI still times out -- I do not understand why.

core-code:
are we seeing real failures or just travis problems here?

miccal:
@core-code CI issues, I think. 

I have successfully tested this on my Mojave machine with and without `sudo: true`, and it _seems_ to be fine.

reitermarkus:
I tried debugging this on CI but no idea what's going on. Also installed fine for me locally.

miccal:
Thank you @reitermarkus.

I suspect it may have something to do with the CI machines probably having Java pre-installed, which _sometimes_ interferes with anything that bundles its own JRE.

PR #86684, opened 2020-07-28
Update openzfs.rb

core-code:
get rid of non-standard syntax, similar to onyx/saoimageds9

also, while we are at at, the 'unusual case' thing is *below* the version numbers in all other cases. it seems to be one of the few things that the style check doesn't care about


vitorgalvao:
It makes sense to have this one further down, because it concerns the `pkg` stanza which is typically under `depends_on`.

core-code:
ok i've restored the order.

PR #86829, opened 2020-07-30
Update malus from 1.0.5 to 1.1.1

i0ntempest:

### DIFF
--- a/Casks/malus.rb
+++ b/Casks/malus.rb
@@ -1,9 +1,9 @@
 cask "malus" do
-  version "1.0.5"
-  sha256 "7771d5026201b8b47a648aa115f1d628ea0fcc27e13db8a7afd5f4c5233558c9"
+  version "1.1.1,1.0.1"
+  sha256 "2a9998cbe5c4b291e627a5c8c7b3e7a853e6a87e60128679675de802cb9c6e5b"
 
   # malus.s3cdn.net/ was verified as official when first introduced to the cask
-  url "https://malus.s3cdn.net/uploads/Malus-#{version.no_dots}-mac.dmg"
+  url "https://malus.s3cdn.net/uploads/Malus-#{version.after_comma.dots_to_underscores}.dmg"
   appcast "https://api.getmalus.com/api/checkDesktopUpdate?type=mac"
   name "Malus"
   homepage "https://getmalus.com/"


Casks/malus.rb
```diff
@@ -1,9 +1,8 @@
 cask "malus" do
-  version "1.0.5"
-  sha256 "7771d5026201b8b47a648aa115f1d628ea0fcc27e13db8a7afd5f4c5233558c9"
+  version "1.1.0"
+  sha256 "2a9998cbe5c4b291e627a5c8c7b3e7a853e6a87e60128679675de802cb9c6e5b"
 
-  # malus.s3cdn.net/ was verified as official when first introduced to the cask
-  url "https://malus.s3cdn.net/uploads/Malus-#{version.no_dots}-mac.dmg"
+  url "https://getmalus.com/r/mac"
```
miccal, line 5: This re-directs to:
```
-bash-5.0.18- /Users/miccal (29) [> curl -sLI https://getmalus.com/r/mac | grep -i Location
location: https://malus.s3cdn.net/uploads/Malus-1_0_1.dmg
```
so please change the `url` to
```
url "https://malus.s3cdn.net/uploads/Malus-#{version.dots_to_underscores}.dmg"
```
and re-add the comment above the `url`.

i0ntempest: Notice the version in the URL is different from the actual version, which is the reason I switched to an upper level URL. Maybe the vendor screwed up the file name.

i0ntempest: Plus for every version they change how they name file and write version so I need to change the URL for every update, which is VERY annoying.

miccal: Oh I see, missed that. Please disregard my suggestions.

miccal: Oh I see, missed that. Please disregard my suggestions.

core-code: maybe having two versions in the the version stanza separated by ',' helps - one real version number and one filename token

miccal: The version inside the `app` itself says 1.1.1!

Screen Shot 2020-07-31 at 17 17 15

i0ntempest: I know, yet their appcast and change log says its 1.1.0... I've no idea what they're thinking.

core-code: the only thing that helps in these cases is being persistent with polite feedback about these kinds of issues :->

core-code: how about

```
version "1.1.1,1.0.1"
url "https://malus.s3cdn.net/uploads/Malus-#{version.after_comma.dots_to_underscores}.dmg"

```

i0ntempest: Right thanks, I did just that. I'll also try contacting the vendor.

core-code:
thanks, looks fine to me except for the comment. while it contains valuable information i don't think it should be added directly to the caskfile and i don't think we've done it before. i think we can document information about apps directly here in the pull requests

core-code:
thanks there

PR #86932, opened 2020-08-01
Update netron to 4.4.3

lutzroeder:


miccal:
@lutzroeder you will need to fix your auto-PR mechanism with the updated `brew cask style` measures.

PR #86967, opened 2020-08-02
Update the-battle-for-wesnoth.rb

stonerl:
Use version specific appcast.
Use enhance download link.


core-code:
i am fine with the URL change but i am not happy about the appcast change. it works fine to exclude any 1.5 dev version but will prevent us from discovering the update to 1.6 once it is officially released

vitorgalvao:
I don’t see how that’s an “enhance download link”. It’s just longer; it doesn’t offer any more granularity of `version`, and changes the Sourceforge format we use everywhere else.

The appcast also becomes problematic, as pointed out by @core-code.

Thank you, but we’ll be refusing unless you can make a stronger argument for the change.

PR #87001, opened 2020-08-03
ci.yml: Run style before CI

vitorgalvao:
`style` should be run first because that always works, unlike CI which sometimes points errors that aren’t there (e.g. asking to remove a `launchctl` that’s already covered in the cask, or a checking an outdated appcast).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,12 @@ jobs:
           PR_URL: ${{ github.event.pull_request._links.html.href }}
           HOMEBREW_COLOR: 1
           HOMEBREW_DEVELOPER: 1
+      - name: brew cask style
+        run: |
+          # don't care about `brew cask style` here.
+          brew untap adoptopenjdk/openjdk
+
+          brew cask style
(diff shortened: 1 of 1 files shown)


reitermarkus:
> that always works

Not quite. If style fails, `brew cask ci` is not run at all.

I think adding 

```
if: always()
```

to the `brew cask style` step will do what we want: Run `brew cask style` unconditionally, even if `brew cask ci` fails.

vitorgalvao:
> > that always works
> 
> Not quite. If style fails, `brew cask ci` is not run at all.

It `style` fails we won’t merge it anyway, so no need to run `ci`. What I meant by “always works” is that it never gives a false negative (unlike `ci`), it’s reliable.

But if you want to make improvements to the PR, it’s fine to edit it directly.

reitermarkus:
`brew cask ci` also adds annotations for style errors which won't get added if `brew cask style` fails.

vitorgalvao:
I’m fine with whatever solution you prefer here. I’d just like for a failed `ci` to not prevent `brew cask style` from running.

PR #87010, opened 2020-08-03
Update routeconverter.rb from 2.29.116 to 2.28.116

suschizu:

### DIFF
--- a/Casks/routeconverter.rb
+++ b/Casks/routeconverter.rb
@@ -1,6 +1,6 @@
 cask "routeconverter" do
-  version "2.29.116"
-  sha256 "2f0df34831e8ee1a864070c5d11fe4b596ea32ed5a8acf1728220fc64547db51"
+  version "2.28.116"
+  sha256 "4fef088ac9a83f3f938b7ec0f967f504385c495715173f3e1ca7336c37661944"
 
   url "https://static.routeconverter.com/download/RouteConverterMac.app.zip"
   appcast "https://static.routeconverter.com/download/previous-releases/",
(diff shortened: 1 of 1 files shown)


Casks/routeconverter.rb
```diff
@@ -1,6 +1,6 @@
 cask "routeconverter" do
-  version "2.29.116"
-  sha256 "2f0df34831e8ee1a864070c5d11fe4b596ea32ed5a8acf1728220fc64547db51"
+  version "2.28.116"
```
miccal, line 2: The `minor` part is not used, so I would suggest:
```
version "2.28"
```
and then you can remove the line `must_contain: version.major_minor` from the `appcast`.

Casks/routeconverter.rb
```diff
@@ -1,6 +1,6 @@
...
+  sha256 "4fef088ac9a83f3f938b7ec0f967f504385c495715173f3e1ca7336c37661944"
 
   url "https://static.routeconverter.com/download/RouteConverterMac.app.zip"
```
miccal, line 5: This may make a better `url`:
```
url "https://static.routeconverter.com/download/previous-releases/#{version}/RouteConverterMacOpenSource.app.zip"
```

suschizu: thanks @miccal :)

core-code: hm. the minor part isn't used but it makes things easier on our side to keep things updated. also removing it will lead to SHA-only changes (refer to  #75425 ).

miccal: @core-code I did not realise that the `patch` (not `minor` as I mistakenly said above) changed -- thanks for the link. 

@suschizu best to put it back in, thank you.

core-code:
thanks @suschizu for your persistance. i think the only remaining problem is that their server is (too) slow today

PR #87166, opened 2020-08-06
Update expressvpn from 7.9.5.2 to 7.9.7.8

hannesbraun:

### DIFF
--- a/Casks/expressvpn.rb
+++ b/Casks/expressvpn.rb
@@ -1,6 +1,6 @@
 cask "expressvpn" do
-  version "7.9.5.2"
-  sha256 "dabb1d0ac109547dd8662a36c03e33058cfcb1052de2bf9064228788840f2a03"
+  version "7.9.7.8"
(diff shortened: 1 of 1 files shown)


miccal:
Lines 109-114 of the `uninstall.tool` script state that:
```
# Initiate the actual uninstall, which requires admin privileges.
echo
echo "The uninstallation process requires administrative privileges"
echo "because some of the installed files cannot be removed by a"
echo "normal user. You may now be prompted for a password..."
echo
```
so the `uninstall script:` may need `sudo: true`.

hannesbraun:
> so the `uninstall script:` may need `sudo: true`.

This doesn't work either. I also tried to install and uninstall this cask on my local machine with no problems.
Any other ideas why the CI is timing out?

miccal:
@hannesbraun I am not sure, sorry -- it could be nay number of things (terminal requiring full disk access, etc.).

>  I also tried to install and uninstall this cask on my local machine with no problems.

Excellent, that is good to know, thanks for testing.

PR #87168, opened 2020-08-06
Update musescore from 3.4.2 to 3.5.0

lpla:

### DIFF
--- a/Casks/musescore.rb
+++ b/Casks/musescore.rb
@@ -1,9 +1,15 @@
 cask "musescore" do
-  version "3.4.2"
-  sha256 "59f7ee60d77b63a67a89f998f19ae3434959b6ed2eeb20c9e90ff56f5659f824"
+  version "3.5.0"
+  shortversion = if version.patch.to_s == "0"
+    version.major_minor.to_s
+  else
+    version
+  end
+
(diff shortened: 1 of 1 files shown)


miccal:
I am guessing that the line:
```
exec '#{appdir}/MuseScore #{version.major}.app/Contents/MacOS/mscore' "$@"
```
needs to be changed to:
```
exec '#{appdir}/MuseScore #{version.major_minor}.app/Contents/MacOS/mscore' "$@"
```
also?

lpla:
You are right, I missed that one. Thank you!

core-code:
i am really unhappy with that shortversion stuff. can we use a solution here that is more in line to what we've used in other casks?

basically if the url changes we've just adjusted the URL stanza instead of trying to find a system that will work in any way the authors might change in the future. if we really want to try to do something that is 'future proof' i suggest a single-line solution like this

`  url "https://github.com/musescore/MuseScore/releases/download/v#{version.chomp(".0")}/MuseScore-#{version}.dmg"
`

lpla:
I am okay with that. I am not familiar with Ruby lang and code style in other brew formulas, so if this solution is better, I will use it.

I know that Musescore devs changed their numbering format in Github tags, release names and app title along time, but looks like the current format is the one they kept most time.

core-code:
thanks!

PR #87246, opened 2020-08-07
sage: Fix version choosing for macOS < Catalina

leogama:
There are currently two binary versions available for the SageMath app:
one for the latest macOS version (10.15 Catalina) and one for a much
older one (10.11 El Capitan). Just the later seems to work on macOS
versions less than 10.15.

Also fix 'depends_on macos'.

Note: I've installed the 10.15 version on my 10.13 system and got a 'malformed application' error when tried to launch it. The 10.11 version worked fine. [This person](https://ask.sagemath.org/question/52776/sagemath-on-macos-app-components-packages-sagetex/) apparently also had to install the version for 10.11 on her 10.14 system and it also worked as expected (besides the current Tk and SSL problems present in both binaries).

### DIFF
--- a/Casks/sage.rb
+++ b/Casks/sage.rb
@@ -1,6 +1,11 @@
 cask "sage" do
-  version "9.1,10.15.4"
-  sha256 "139fd510245d94e9027fa6fff280b41a238b2d6044f1923c49e66a031da166e4"
+  if MacOS.version <= :mojave
+    version "9.1,10.11.6"
+    sha256 "23c13690b8a72deca1628dd0e66a0f7b83557f98c13c3db1dc7eb15d80cf3a8d"
+  else
+    version "9.1,10.15.4"
+    sha256 "139fd510245d94e9027fa6fff280b41a238b2d6044f1923c49e66a031da166e4"
+  end
 
   # mirrors.mit.edu/sage/osx/intel/ was verified as official when first introduced to the cask
   url "https://mirrors.mit.edu/sage/osx/intel/sage-#{version.before_comma}-OSX_#{version.after_comma}-x86_64.app.dmg"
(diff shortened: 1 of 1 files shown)


Casks/sage.rb
```diff
@@ -1,14 +1,19 @@
 cask "sage" do
-  version "9.1,10.15.4"
-  sha256 "139fd510245d94e9027fa6fff280b41a238b2d6044f1923c49e66a031da166e4"
+  if MacOS.version < :catalina
```
reitermarkus, line 2: `<= :mojave`

Casks/sage.rb
```diff
@@ -21,4 +26,11 @@
     "~/Library/Logs/sage.log",
     "~/Library/Preferences/org.sagemath.Sage.plist",
   ]
+
+  if (MacOS.version > :el_capitan) && (MacOS.version < :catalina)
+    caveats <<~EOF
+      This #{token} app was compiled for OS X 10.11 El Capitan, but should also
```
reitermarkus, line 32: We don't need this caveat. This applies to virtually every app which was compiled for an older version.

Casks/sage.rb
leogama: OK. The `brew cask style --fix` didn't catch this.

Casks/sage.rb
leogama: Does this rule apply in general? When there are two or more binaries of an app, for the newest and some older macOS versions but not for intermediate ones, is it expected that the binary for the closest preceding version should work on your system?

For example, suppose there are binaries for macOS 10.10, 10.13 and 10.15. Then one should install the version for 10.10 on macOS 10.11 or 10.12, and install 10.13 on macOS 10.14?

reitermarkus: Yes, it should always install the latest possible one. Generally macOS versions are backwards compatible but binaries are not.

reitermarkus:
Thanks!

Casks/sage.rb
leogama: Makes sense. The system is backwards compatible so that the binary apps work in a "forwards compatible" way. Good to know! Thanks for accepting the PR.

PR #87495, opened 2020-08-12
Delete cutbox

miccal:
The `url` is dead and it looks like the developer has pulled the `dmg`'s -- from `https://github.com/cutbox/CutBox`:

> Download / install...

> CutBox is Free OpenSource software. It isn't available via Apple's AppStore. or possible to download and install a CutBox.app. This is thanks to Apple's poorly thought through (or deliberately toxic to Open Source apps?) Gatekeeper / Notarization policy.

> I recommend compiling from source (I will make this easier for users via Homebrew.) See the developer notes at the bottom of this README if you'd like to try this method and bootstrap your very own CutBox.app.


vitorgalvao:
Markdown tip: If you use a `>` on the empty lines, the grey left bar will be uninterrupted, making it clear it’s a single quote.

```
> this is

> your example
```

> this is

> your example

```
> but it becomes uninterrupted
>
> like this
```

> but it becomes uninterrupted
>
> like this

miccal:
Good to know, thanks for the tip.

PR #87497, opened 2020-08-12
Update bettertouchtool url

miccal:

### DIFF
--- a/Casks/bettertouchtool.rb
+++ b/Casks/bettertouchtool.rb
@@ -1,9 +1,9 @@
 cask "bettertouchtool" do
-  version "3.402"
+  version "3.402-1633"
   sha256 "578f2e62805ae21e880c60e13ec455ec6057745edaff14214eb5c6eac5cf7705"
 
   url "https://folivora.ai/releases/BetterTouchTool.zip"
-  appcast "https://updates.folivora.ai/bettertouchtool_release_notes.html"
+  appcast "https://www.corecode.io/macupdater/appcasts/bettertouchtool.txt"
   name "BetterTouchTool"
   homepage "https://folivora.ai/"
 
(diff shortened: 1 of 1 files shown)


core-code:
we've had *many* problems with this app here previously so i am not really happy to revert to the status quo here.

core-code:
i can change the cronjob that generates the appcast to run more than once daily if that is desired though

miccal:
I think we should try using the `url` from the `homepage` -- if it turns out this link points to non-stable releases, we can revert these changes.

miccal:
@core-code perhaps you could change your cronjob to download from `https://folivora.ai/releases/BetterTouchTool.zip` and monitor the `version` that way?

core-code:
> @core-code perhaps you could change your cronjob to download from

this is what is already done. here is the source: https://textbin.net/LsFiK2qiTu

> I think we should try using the url from the homepage -- if it turns out this link points to non-stable releases

well, any release linked from the homepage is by definition 'stable'.


i want to avoid a situation like we had before. we had 'https://folivora.ai/releases/' as appcast and people would submit *any* version listed there, which were often newer than the one in the homepage download *and* newer than in the default update check. i.e. these were beta releases listed only in their "alpha update check". 


switching to the unversioned download will have one possible downside. in case the download from the homepage lags the download from the update check we won't have the flexibility to just submit the newer version from the update check without changing back to the versioned download. i am not sure what we gain on the other hand. but i am fine with the change since the downside is pretty hyptothetical in this particular case.

however i don't think we should change the appcast here. their release notes are notoriously outdated to the point of uselessness. as mentioned i'll increase frequency of the cronjob instead so that necessary updates can go through sooner.

core-code:
> as mentioned i'll increase frequency of the cronjob instead so that necessary updates can go through sooner.

i've had a look and this already runs every two hours

miccal:
I have restored the original `appcast`.

core-code:
thanks

PR #87562, opened 2020-08-13
Upgrade FileMaker Pro Advanced to v19.0.1.116

johnchristopherjones:
For major version 19, there are some significant changes:
1. FileMaker, Inc is now Claris, hence the domain change.
2. The latest version is now packaged as an App instead of a pkg installer.

I'm not sure what to do about supporting older pkg versions under the same
token.  This commit just switches to the current version.

Also ^ not clear if this qualifies for "trial" tap since Claris introduced a 45-day trial and you need a license key past that point. No re-download is necessary, so I think this qualifies it for "Stable".


miccal:
So is this now just called `FileMaker Pro`, not `FileMaker Pro Advanced`?

vitorgalvao:
Please add a [description](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/desc.md).

johnchristopherjones:
@miccal Ah, good catch. "FileMaker Pro" seems to be the official name now.  I was confused because when they announced consolidating "Pro" with "Pro Advanced", they originally settled on the former.

@vitorgalvao Done

miccal:
> Description is too long. It should be less than 80 characters. The current length is 103.

How about instead:
```
desc "Cross-platform relational database and rapid application development platform"
```
which is 77 characters.

vitorgalvao (approved):

PR #87678, opened 2020-08-16
macvim: remove caveat

vitorgalvao:
Waiting on https://github.com/Homebrew/brew/pull/8370.

### DIFF
--- a/Casks/macvim.rb
+++ b/Casks/macvim.rb
@@ -29,8 +29,4 @@ cask "macvim" do
     "~/Library/Preferences/org.vim.MacVim.LSSharedFileList.plist",
     "~/Library/Preferences/org.vim.MacVim.plist",
   ]
-
-  caveats <<~EOS
-    To avoid overwriting vim binaries from other formulae or casks, install with the "--no-binaries" flag.
-  EOS
 end


miccal:
@vitorgalvao I have had a quick look in the `homebrew-cask` repository for any casks that may be affected by the changes I have proposed in https://github.com/Homebrew/brew/pull/8370, and I have summarised my findings below:

- [cmake](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/cmake.rb); installs an `.app` and a `binary`; `conflicts_with formula: "cmake"`.

- [emacs](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/emacs.rb); installs an `.app` and a `binary`; `conflicts_with formula: "emacs"`.

- [kotlin-native](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/kotlin-native.rb); installs a `binary` only; `conflicts_with formula: "kotlin"`.

- [macvim](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/macvim.rb); installs an `.app` and a `binary`; `conflicts_with formula: "macvim"`.

- [mkvtoolnix](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/mkvtoolnix.rb); installs an `.app` and a `binary`; `conflicts_with formula: "mkvtoolnix"`.

- [mpv](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/mpv.rb); installs an `.app` and a `binary`; currently does not have a `conflicts_with formula: "mpv"`.

- [rar](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/rar.rb); installs a `binary` only; `conflicts_with formula: "unrar"`.

- [unison](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/unison.rb); installs an `.app` and a `binary`; `conflicts_with formula: "unison"`.

- [wine-stable](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/wine-stable.rb); installs an `.app` and a `binary`; `conflicts_with formula: "wine"`.

Now, for the ones that install an `.app` (all but `kotlin-native` and `rar`) I think that there may be those who want both the formula version for the fully-fledged binary and the cask version for the `.app`, so I am thinking that the `conflicts_with formula:` should be removed from these and replaced with a `caveat` similar to the one you have proposed here for `macvim`.

I am also thinking of adding a new "caveats Mini-DSL" [here](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/dsl/caveats.rb), something like:
```
caveat :binary_conflict do
  <<~EOS
    If the installation fails due to a binary already existing
    in /usr/local/bin, retry with the "--no-binaries" flag:
      brew cask install #{@cask} --no-binaries
  EOS
end
```
and then we could just add:
```
caveats do
  binary_conflict
end
```
where necessary.

As always, I would value your thoughts and opinions on this before I go ahead with this.

Thanks!

vitorgalvao:
I’m undecided if removing the conflict is the best course of action. I’d like to see reasonings of people who want both the formula and the cask on why that’s a superior combination over just using the cask, as well as some proof that doesn’t cause issues.

miccal:
> I’m undecided if removing the conflict is the best course of action. I’d like to see reasonings of people who want both the formula and the cask on why that’s a superior combination over just using the cask, as well as some proof that doesn’t cause issues.

Fair enough, but I still think that having another `caveats` DSL would be useful in light of https://github.com/Homebrew/brew/pull/8370 being merged.

PR #87855, opened 2020-08-19
Update hand-mirror from 1.1.2 to 1.3.1

tedski:

### DIFF
--- a/Casks/hand-mirror.rb
+++ b/Casks/hand-mirror.rb
@@ -1,9 +1,8 @@
 cask "hand-mirror" do
-  version "1.1.2"
-  sha256 "da590bc7666646185e647c7ab53c28d007f37bd32360d3decf78729677ea798e"
+  version "1.3.1"
+  sha256 "0c88e844d5161f093e374ef0e89c0eeddf589eab2db7e5b2c15c810255496167"
 
   url "https://handmirror.app/Hand%20Mirror%20#{version}.dmg"
-  appcast "https://handmirror.app"
   name "Hand Mirror"
   homepage "https://handmirror.app/"
 


core-code:
according to the homepage this only exists in the Mac App Store now. where does the link to version 1.3.1 come from?

tedski:
> according to the homepage this only exists in the Mac App Store now. where does the link to version 1.3.1 come from?

I reached out to the developer and asked if he could start hosting the latest versions on his site so we can properly upgrade.  I'm still working with him to provide the version number on the site so we can use appcast, but he said he will upload new versions to the same site.

vitorgalvao:
> To be clear, the appcast stuff will be handled in a future PR.

I’d rather it be done now.

From experience, this is the kind of thing the developer will either drag on or forget to do. Without an appcast or a page to download there’s no way for us to know when the app is updated (checking the MAS is not a reasonable solution, especially if those aren’t updated in tandem).

The cask isn’t particularly popular now, so if we’re going to remove it, it should be sooner rather than later.

Let’s give it a month to avoid disruption. If the situation hasn’t changed by then, removing the cask will be sensible. That doesn’t mean it will be removed forever; if the problem gets correct after it’s removed, it can be resubmitted.

I’ve merged this for now since it’s an update, but [the removal PR is ready](https://github.com/Homebrew/homebrew-cask/pull/88059).

```
hand-mirror
30 days: 8 (#2698)
90 days: 31 (#2491)
365 days: 73 (#3544)
Age: 154 days (added 2020, March 23)
```

tedski:
Thanks, @vitorgalvao.  I feel that's a perfectly reasonable path forward.  I will contact the developer and set the appropriate sense of urgency.

vitorgalvao:
One day early, but I doubt it makes a difference. [Removing it](https://github.com/Homebrew/homebrew-cask/pull/89712).

tedski:
Agreed, @vitorgalvao.  I appreciate you following up with @rafaelconde about this.  It's unfortunate we weren't able to get the appcast stuff figured out.  I apologize for the inconvenience and toil here.

vitorgalvao:
> I appreciate you following up with @rafaelconde about this.

[Credit for that](https://github.com/Homebrew/homebrew-cask/pull/88059#issuecomment-695771157) goes to @core-code.

> I apologize for the inconvenience and toil here.

Nothing to apologise for, you were an exemplary contributor. Thank you for reaching out upstream and working with us on this.

Issue #24377, opened 2016-09-10
Allow GUI un/installs


vitorgalvao:
Ever since the inception of Homebrew-Cask, we (mostly<sup id="a1">[1](#f1)</sup>) haven’t allowed casks to call GUIs when installing and uninstalling, in the name of automation.

However, in some cases it actually prevents desirable automation: not auto-opening [`installer manual:`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/installer.md#installer-manual) targets and not using tool-provided uninstallers come to mind.

I postulate most users would actually prefer these GUI actions to be called automatically in most cases, the most notable exception being when reinstalling a system (since a bunch of casks are installed).

It also seems more in line with the goal of #13201 of making our automation as close to the manual installation as possible.

---

Proposal:

Change policy and start allowing GUIs to be called, *but only allow GUIs as the last resort*.

When calling a GUI is necessary for an action, allow the user to pick if a GUI will be called or the action will be interrupted. Make allowing GUIs the default.

Implement `--no-gui` as a flag to be given at runtime or in (`HOMEBREW_CASK_OPTS`). It means “If this action needs a GUI, don’t run it, and instead output a warning on how to proceed manually” (i.e. the same instructions we have now).

Implement `installing_calls_gui` and `uninstalling_calls_gui`. Similarly to `auto_updates` and `caskroom_only`, this should only exist when the value is `true`. This serves a single purpose: when we’re calling `install`/`uninstall` and the corresponding `*_calls_gui` stanza is `true`, check is `--no-gui` is set and act accordingly. In terms of [stanza order](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/readme.md#stanza-order), right above `auto_updates` seems fine.

When opening GUIs, do so with `open -W`. This way the command will be blocked until it finishes, preventing a barrage of GUIs left open when installing multiple apps and allowing for `&&` and other subsequent commands.

Rename `installer manual:` to `installer gui:`.

Add `uninstall gui:`. It runs *in addition to* the other flags, and before them (otherwise it could very well be removed by them).

---

Checklist:

+ [ ] Add `installing_calls_gui` and `uninstalling_calls_gui` stanzas.
+ [ ] Add `--no-gui` flag.
+ [ ] Rename (and update implementation of) `installer manual:`.
+ [ ] Implement `uninstall gui:`.
+ [ ] Update documentation.
+ [ ] Add tests.
[...]

PR #88039, opened 2020-08-23
Use Backblaze Uninstaller and switch to latest version as Backblaze downloads are unversioned

mudge:
Backblaze [do not offer versioned downloads of their installer](https://secure.backblaze.com/download.htm) so attempting to upgrade the Cask causes issues with mismatched checksums if the previous installer is still cached.
    
This means we lose the ability to detect updates but forcibly upgrading the Cask (e.g. with `brew cask upgrade backblaze`) should now work as expected.

Rather than having to manually delete the various applications, preference panes and `launchctl` jobs installed by the Backblaze Installer ourselves, use the included Backblaze Uninstaller script instead.
    
Note that the Uninstaller will not delete user-specific files so we retain the existing zap stanza.

### DIFF
--- a/Casks/backblaze.rb
+++ b/Casks/backblaze.rb
@@ -8,14 +8,7 @@ cask "backblaze" do
 
   installer manual: "Backblaze Installer.app"
 
-  uninstall launchctl: [
-    "com.backblaze.bzserv",
-    "com.backblaze.bzbmenu",
-  ],
-            delete:    [
-              "#{appdir}/Backblaze.app",
-              "/Library/PreferencePanes/BackblazeBackup.prefPane",
-            ]
+  uninstall script: "#{staged_path}/Backblaze Uninstaller.app/Contents/MacOS/Backblaze Uninstaller"
 
   zap trash: [
     "/Library/Backblaze.bzpkg",
--- a/Casks/backblaze.rb
+++ b/Casks/backblaze.rb
@@ -1,10 +1,8 @@
 cask "backblaze" do
-  version "7.0.2.464"
-  sha256 "4956a72674d1901a6ec5a98ec0089ebfde743c5bb63e7644c3103e113e26d6d8"
+  version :latest
+  sha256 :no_check
 
   url "https://secure.backblaze.com/mac/install_backblaze.dmg"
-  appcast "https://secure.backblaze.com/api/clientversion.xml",
-          must_contain: "mac_version=\"#{version}\""
   name "Backblaze"
   homepage "https://backblaze.com/"
 


Casks/backblaze.rb
```diff
@@ -1,23 +1,14 @@
...
 
-  uninstall launchctl: [
-    "com.backblaze.bzserv",
-    "com.backblaze.bzbmenu",
-  ],
-            delete:    [
-              "#{appdir}/Backblaze.app",
-              "/Library/PreferencePanes/BackblazeBackup.prefPane",
-            ]
+  uninstall script: "Backblaze Uninstaller.app/Contents/MacOS/Backblaze Uninstaller"
```
miccal, line 11: Please change this to:
```
uninstall script: "#{staged_path}/Backblaze Uninstaller.app/Contents/MacOS/Backblaze Uninstaller"
```

mudge: Updated to use `staged_path`.

Casks/backblaze.rb
```diff
@@ -1,23 +1,14 @@
...
 
-  uninstall launchctl: [
-    "com.backblaze.bzserv",
-    "com.backblaze.bzbmenu",
-  ],
-            delete:    [
-              "#{appdir}/Backblaze.app",
-              "/Library/PreferencePanes/BackblazeBackup.prefPane",
-            ]
+  uninstall script: "#{staged_path}/Backblaze Uninstaller.app/Contents/MacOS/Backblaze Uninstaller"
```
vitorgalvao, line 11: This is not a script, it will call a GUI which complicates automation and which we do not want [until we can do it properly](https://github.com/Homebrew/homebrew-cask/issues/24377).

Casks/backblaze.rb
```diff
@@ -1,23 +1,14 @@
 cask "backblaze" do
-  version "7.0.2.464"
-  sha256 "4956a72674d1901a6ec5a98ec0089ebfde743c5bb63e7644c3103e113e26d6d8"
+  version :latest
+  sha256 :no_check
 
   url "https://secure.backblaze.com/mac/install_backblaze.dmg"
-  appcast "https://secure.backblaze.com/api/clientversion.xml",
```
vitorgalvao, line 6: The whole point of the appcast is that we can detect when updates are available.

Casks/backblaze.rb
```diff
@@ -1,23 +1,14 @@
 cask "backblaze" do
-  version "7.0.2.464"
```
vitorgalvao, line 2: It’s by having a version that `brew upgrade` can work properly. You’re removing that ability.

vitorgalvao (changes requested):

vitorgalvao:
The cask is correct as is. Everything you mentioned is done intentionally.

mudge:
Thanks for reviewing, @vitorgalvao.

It's especially useful to know [support for GUI uninstalls are a separate issue](https://github.com/Homebrew/homebrew-cask/issues/24377).

Re the versioning of Backblaze: in retrospect, sacrificing the functioning `appcast` and `version` is overkill when my original issue is a much smaller one: that upgrading Backblaze in-place fails because the checksum fails due to all versions of the installer being cached as `install_backblaze.dmg`.

Frustratingly, this is hard to reproduce due to the fact it's not possible to download an older version of the installer and I notice that all downloads in `~/Library/Caches/Homebrew/downloads` seem to be prefixed with a checksum.

I'll re-test when a new version of the Backblaze installer is released to see if I can reproduce the problem more thoroughly then.

PR #88059, opened 2020-08-24
Delete hand-mirror.rb

vitorgalvao:
See https://github.com/Homebrew/homebrew-cask/pull/87855#issuecomment-679147633.


core-code:
@rafaelconde is the non-Mac App Store version 1.3.1 of Hand Mirror going to be listed on your website? else we'll have to remove it from Homebrew Cask as discussed here: https://github.com/Homebrew/homebrew-cask/pull/87855#issuecomment-679147633

vitorgalvao:
This now has conflicts, so removing in another PR: https://github.com/Homebrew/homebrew-cask/pull/89712. Only needs to be accepted.

PR #88094, opened 2020-08-24
Update backblaze url and installer manual

miccal:
Use the versioned `url` provided by the `appcast`. 

Although the `installer manual:` has a different name, it is identical to the original one provided by the `dmg` -- the only difference is that that `dmg` includes a `Backblaze Uninstaller.app`, which is actually buried inside both the `Backblaze Installer.app` and `bzdoinstall.app` anyway.

This should help with any checksum errors as reported in https://github.com/Homebrew/homebrew-cask/pull/88039.

Ping @mudge.

### DIFF
--- a/Casks/backblaze.rb
+++ b/Casks/backblaze.rb
@@ -1,14 +1,15 @@
 cask "backblaze" do
   version "7.0.2.464"
-  sha256 "4956a72674d1901a6ec5a98ec0089ebfde743c5bb63e7644c3103e113e26d6d8"
+  sha256 "1fed02eb7190c0ce052b6f1c0f5b33a7e89178cd1501a46cf99c2842109f9edf"
 
-  url "https://secure.backblaze.com/mac/install_backblaze.dmg"
+  url "https://secure.backblaze.com/api/install_backblaze?file=bzinstall-mac-#{version}.zip"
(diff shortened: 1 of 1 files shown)


mudge:
Great find, @miccal. Thanks for looking into this further.

As this means the URL will differ for every version of the Backblaze installer and the checksum of the download URL is used to prefix the cached filename, it should fix the issue I've been seeing.

Hopefully the fact these download URLs appear in the [appcast feed](https://secure.backblaze.com/api/clientversion.xml) itself mean they are stable enough for us to depend on, e.g. at the time of writing:

```xml
<update win32_version="7.0.2.463" win32_url="%DEST_HOST%/api/install_backblaze?file=bzinstall-win32-7.0.2.463.exe" win32_msi_version="7.0.2.463" win32_msi_url="%DEST_HOST%/api/install_backblazemsi?file=bzinstall-win32-msi-7.0.2.463.msi" mac_version="7.0.2.464" mac_url="%DEST_HOST%/api/install_backblaze?file=bzinstall-mac-7.0.2.464.zip" linux_version="1.0.0.16" linux_url="%DEST_HOST%/api/install_backblaze?file=bzinstall-linux-1.0.0.16"/>
```

PR #88250, opened 2020-08-27
Remove Podman

ashley-cui:
Podman now builds on a mac, and is now available through homebrew-core


reitermarkus:
Please also add it to `tap_migrations.json`.

PR #88172, opened 2020-08-26
Update kafka-tool from 2.0.7 to 2.0.8

ran-dall:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Created with [`cask-repair`](https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask).

PR #88301, opened 2020-08-27
Check for `skip-install` label.

reitermarkus:
Added a `skip-install` label to skip installing a cask on CI, in cases such as https://github.com/Homebrew/homebrew-cask/pull/88172 where CI cannot be fixed any other way.


vitorgalvao:
@reitermarkus If we’re adding magic labels that interact with CI, I think we should affix them in a clear manner (e.g. `ci-skip-install`) and possibly use the same (exclusive) color in all of them.

PR #88355, opened 2020-08-28
Update metasploit from 6.0.2,20200820102910 to latest

mojotx:

### DIFF
--- a/Casks/metasploit.rb
+++ b/Casks/metasploit.rb
@@ -8,7 +8,7 @@ cask "metasploit" do
 
   depends_on formula: "nmap"
 
-  pkg "metasploit-framework-latest.pkg"
+  pkg "metasploitframework-latest.pkg"
   binary "/opt/metasploit-framework/bin/msfbinscan"
   binary "/opt/metasploit-framework/bin/msfconsole"
   binary "/opt/metasploit-framework/bin/msfd"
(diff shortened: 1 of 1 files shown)


mojoTX:
This is a work-in-progress. The metasploit packages change so quickly that we need to either just use the "latest" URL, or else automate the process to update the cask based on the latest URL and SHA256.

mojoTX:
This seems to work, if we want to grab the latest version. Ideally we would validate the PGP signature of the package, by downloading from https://osx.metasploit.com/metasploitframework-latest.pkg.asc and using `gpg --verify metasploitframework-latest.pkg.asc metasploitframework-latest.pkg`. 

However, that would also create a dependency on gpg, unless there's a better way to do it.

reitermarkus:
> or else automate the process to update the cask based on the latest URL and SHA256.

That's exactly why we prefer versioned URLs with an `appcast`. There's no official tool for this yet but some contributors are already using this to generate automatic update PRs.

vitorgalvao:
Please revert this to how it was, updating only the `version` and `shasum`. The cask is built as intended.

ran-dall:
@mojoTX As FTR, our ML system does a pretty good job of picking up the `metasploit` updates within 24hrs in the past. If we didn't have coverage on this cask, then I would say that you're solution could be considered; but if we have coverage, then as @vitorgalvao said...

> The cask is built as intended.

mojoTX:
I will close the MR, but it seems to be a common problem where the cask falls behind the nightly releases.

PR #88388, opened 2020-08-29
Update app-cleaner from 7.0.3 to 7.1

ran-dall:

### DIFF
--- a/Casks/app-cleaner.rb
+++ b/Casks/app-cleaner.rb
@@ -1,6 +1,6 @@
 cask "app-cleaner" do
-  version "7.0.3"
-  sha256 "6b83428a2a0267df6eb8f9637ced57ac2e1832aea69568c8d82f49bd6186d4d4"
+  version "7.1"
+  sha256 "4809a00410ab09a88e70e66aed1e88a3f359c0c26a3004e124f467d7924d7fb2"
 
   url "https://nektony.com/download/app-cleaner-uninstaller/app-cleaner-uninstaller.dmg"
   appcast "https://nektony.com/pro-support/app-cleaner/update/update.xml"


core-code:
i can confirm this is available on the homepage and the correct version number

core-code:
would like to merge but something is broken:

Screenshot 2020-08-30 at 15 33 43

reitermarkus:
> would like to merge but something is broken

Nothing is broken, it's working as intended. You cannot merge if CI is failing.

We should either fix the appcast, wait for it to be updated or add a specific label which lets CI pass (like https://github.com/Homebrew/homebrew-cask/pull/88301).

core-code:
> We should either fix the appcast, wait for it to be updated or add a specific label which lets CI pass (like #88301).

if we add a label, *anything* will pass and be auto-merged and this is definitely not what we want.

given how often CI fails although everything is fine i am not a fan of this new behaviour.

reitermarkus:
> if we add a label, anything will pass and be auto-merged and this is definitely not what we want.

Well, we only apply the label once we have determined that it should be merged even with the failure. Also `automerge` will not merge any PR which has any labels now.

> given how often CI fails

It only fails so often because we keep merging failing PRs. If we never fix CI, it will obviously keep failing.

core-code:
> Well, we only apply the label once we have determined that it should be merged even with the failure.

i don't see a problem with manually merging a PR after determining that is correct.
however i see a big problem with applying a label that will let *anything* be automerged in the future. we had many issues there and those are much harder to catch than missing updates.

> It only fails so often because we keep merging failing PRs. If we never fix CI, it will obviously keep failing.

thats a great goal and of course i wish that CI would work 100%. however i don't think it is realistic to expect that just because manatainance work is made next-to-impossible with these changes that suddently some people that are willing and able to fix CI are going to magically appear out of thin air.

also, regarding the appcast check specifically - this is not a failure that should prevent maintainters from merging:

> We should either fix the appcast, 

as can be seen in this case, there is no appcast that has version 7.1. however it *is* version 7.1 so it *should* be merged. anything that prevents us from doing the right thing sounds broken to me.

> wait for it to be updated

currently /any/ user trying to do 'brew cask install app-cleaner' will just see a SHA error which is not a good thing.
why hold up a perfectly fine PR that fixes this problem just because they haven't updated their website? the reality is that developers are lazy and make mistakes and need to have the capability to deal with those situations gracefully without breaking the whole system (see below).

> or add a specific label which lets CI pass (like #88301).

thats the worst option, because it will allow *anything* to pass through and be automerged unchecked in the future. i am strongly opposed to this.


regarding appcasts, the basic issue is that we'd ideally want to have several appcasts for a single app. e.g. all of the release-notes site, the download-site and the sparkle-feed. the newest version might be contained in any of those. however this capability does not exist yet. and we can not shuffle the appcast for each and every update just because merging with an appcast-failure is now impossible. maintainance is already enough work as-is and this will make it a lot harder for no upside at all. 

i am sorry to disagree on all fronts here. since this mainly affects the maintainers doing all the work on merging PRs i think it is important to know what @vitorgalvao and @ran-dall think here.

reitermarkus:
> there is no appcast that has version 7.1. however it is version 7.1 so it should be merged

Then the appcast is wrong or broken and should be updated or removed. I don't see the usefulness in an appcast that is broken half of the time.

reitermarkus:
> it will allow anything to pass through and be automerged unchecked in the future

It will let a specific PR pass once it is labelled, which won't be automerged because it then has a label. Any future PR will not have a label to begin with and will fail as usual.

core-code:
> Then the appcast is wrong or broken and should be updated or removed. I don't see the usefulness in an appcast that is broken half of the time.

the reality isn't binary good versus broken here. appcasts are somtimes late, sometimes wrong, and might still be good 90% of the time.

and an appcast is valuable for both of its purposes (finding updates & preventing auto-merging bad updates) even if it is 'good' just one percent - anything is better than nothing here.

> It will let a specific PR pass once it is labelled, which won't be automerged because it then has a label. Any future PR will not have a label to begin with and will fail as usual.

thanks that sounds good

ran-dall:
I'm with @core-code on this one. Furthermore, this action inhibits some of the feedback required by the ML. This would add a decent amount of work to work around, and there's already a decent amount of work that needs to be to the ML for it to be up up-to-par with all the changes that have been happening recently. So I don't know how much longer the ML will be around to keep casks up-to-date if we continue in this direction.

reitermarkus:
> this action inhibits some of the feedback required by the ML

@ran-dall, can you elaborate on that? How exactly does ensuring CI is passing inhibit anything?

ran-dall:
@reitermarkus Sorry for the delay in my reply, I've been crazy busy at work lately; since folks is coming back from the COVID break. Anyways, the old CI system was referenced by the CI. Mainly, it used to understand when something was force-merged, as usually the update was correct (so it was marked correct in the analysis). If the update wasn't merged, it prefers something like `bad-update` or `build-fail` to be posted for it's analysis. 

Now that the updates are more CI and tag dependent, I'd have to rework the tags into the system. Currently `skip-install` doesn't mean anything to the ML, so it's something that I'd manually have to input until I find a better solution.

So inhibit might be the wrong choice of words. I don't think it really inhibits anything, more than it just requires an update on my side.

Also FTR it picked up the output of Travis a bit better than GitHub Actions.

reitermarkus:
> used to understand when something was force-merged, as usually the update was correct

Wouldn't it still be marked as a correct update if there was no error, i.e. a normal merge when CI is passing?

ran-dall:
I believe so

reitermarkus:
Then is there still an issue?

ran-dall:
Kinda, it's just time I need to invest into the ML to make it all work.

ran-dall:
I'll let you know if become a real problem.

vitorgalvao:
I thought about this for a while, and think a solution that would satisfy everyone is an action that merges *any* PR (i.e. not only the ones that change *just* `version` and `sha256` as long as):

* All checks passed.
* A maintainer (i.e. someone with merge access) approved it (through GitHub’s approve feature).
* _Optional_: More granular magic `ci` labels. `ci-verified-manually` covers everything, but something like `ci-skip-appcast` might also be a good one.

On that last point, think how better organised the repo would be. In the future, just by counting the number of tags we’d know which tests we skip more often—hence the ones we should prioritise improving. And we’d also be able to search for a cask and from the issue list see which test we skip more often and improve *the cask itself* on that point.

But specific `ci` tags are also useful to catch mistakes. If a maintainer thinks a cask is only failing due to one test, applying only that label will point out the oversight by not merging.

Either way, following the two bullets points alone would mean that applying a label is no more effort than direct merging, which is part of the issue here: currently it’s more work—apply label, *wait* for tests again, merge.

reitermarkus:
> GitHub’s approve feature

This is only really possible if we also require an approval, since we need to ensure the approval is dismissed after a new push. I think it's possible if we make `BrewTestBot` an admin which approves version bump PRs.

> More granular magic `ci` labels.

Definitely, however this will also need some more granular options for the `audit` command.

vitorgalvao:
> Definitely, however this will also need some more granular options for the `audit` command.

I’d say we start with whatever’s already possible through flags (like `--appcast`) and after we see how those work decide what (if anything) we need more.

reitermarkus:
> I’d say we start with whatever’s already possible through flags (like `--appcast`)

It's not actually possible to disable though, since `--online` implies `--appcast` currently.

PR #88410, opened 2020-08-30
Update redisinsight

suschizu:

### DIFF
--- a/Casks/redisinsight.rb
+++ b/Casks/redisinsight.rb
@@ -3,7 +3,8 @@ cask "redisinsight" do
   sha256 "337ebb483582742749a1a9312d35e06b4f82f3b8dd5b79523744073f4bc635e6"
 
   url "https://downloads.redisinsight.redislabs.com/#{version}/redisinsight-mac.dmg"
-  appcast "https://docs.redislabs.com/latest/ri/release-notes/"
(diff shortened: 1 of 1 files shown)


Casks/redisinsight.rb
```diff
@@ -3,7 +3,8 @@
   sha256 "337ebb483582742749a1a9312d35e06b4f82f3b8dd5b79523744073f4bc635e6"
 
   url "https://downloads.redisinsight.redislabs.com/#{version}/redisinsight-mac.dmg"
-  appcast "https://docs.redislabs.com/latest/ri/release-notes/"
+  appcast "https://docs.redislabs.com/latest/ri/release-notes/",
+          must_contain: version.major_minor
```
reitermarkus, line 7: Are appcasts which don't contain the full version actually useful?

vitorgalvao: > Are appcasts which don't contain the full version actually useful?

Yes. *At the very least* it allows detection of `1.7` and beyond. We shouldn’t ignore the ability to detect major and minor updates only because we can’t see patch bumps.

core-code: even more importantly the appcast will prevent auto-merging of beta versions of major and semi-major releases.

PR #88427, opened 2020-08-30
Update vmware-horizon-client from 5.4.3-16499473,CART21FQ1 to 2006-8.0.0-16531451,CART21FQ2

beargle:

### DIFF
--- a/Casks/vmware-horizon-client.rb
+++ b/Casks/vmware-horizon-client.rb
@@ -1,6 +1,6 @@
 cask "vmware-horizon-client" do
-  version "5.4.3-16499473,CART21FQ1"
-  sha256 "3fe45ce4f1a84049ed9d7e445fdc0397da7c8b420ae870e686fd66876b2003d6"
+  version "2006-8.0.0-16531451,CART21FQ2"
+  sha256 "ef63a658b2233a8490421e33ecfbafb2a05ef57f0d3810bb6912b92f117cf468"
 
   url "https://download3.vmware.com/software/view/viewclients/#{version.after_comma}/VMware-Horizon-Client-#{version.before_comma}.dmg"
   appcast "https://docs.vmware.com/en/VMware-Horizon-Client-for-Mac/",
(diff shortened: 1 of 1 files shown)


core-code:
hm, it does have a built-in update check but i was unable to determine the URL they use

beargle:
Closest thing to an update URL so far, but requires the `CART`-prefixed version/string/whatever to be manually supplied

https://my.vmware.com/channel/public/api/v1.0/dlg/details?locale=en_US&downloadGroup=CART21FQ2_MAC_800&productId=1027&rPId=48989

core-code:
thanks that one definitely looks better than nothing at all

> CART-prefixed version/string/whatever to be manually supplied

ok whats the actual significance here. will we habe to update it on the next major version? every minor version?


btw how did you obtain the URL? ;)

core-code:
thanks that one definitely looks better than nothing at all

> CART-prefixed version/string/whatever to be manually supplied

ok whats the actual significance here. will we habe to update it on the next major version? every minor version?


btw how did you obtain the URL? ;)

beargle:
> ok whats the actual significance here? will we habe to update it on the next major version? every minor version?

Looks like the URL will need to be changed for every major version for sure. Likely every minor version (which is consistent with previous releases), but not sure what we'll be getting in future - especially considering this is the first release using some calver/semver hybrid.

> btw how did you obtain the URL? ;)

Poking around in browser developer tools for XHR activity on the [download page](https://my.vmware.com/en/web/vmware/downloads/details?downloadGroup=CART21FQ2_MAC_800&productId=1027&rPId=48989), good bit of info buried in there.

image

Let me know if you find something more useful. Thanks!

core-code:
ok thanks for the explanation! lets try this for an appcast and lets see where it leads us.

core-code:
> Let me know if you find something more useful. Thanks!

of course. i did try to find out the URL contact by the update-check built into the app, but since it ignored the system proxy (bah!) and also ignored the http_proxy variable, 'sniffing' it wasn't possible. i gave up after a few minutes...

PR #88682, opened 2020-09-04
USAGE.md: various updates

miccal:
Commands that still need the `cask` prefix and do not accept the `--cask` flag:
- `fetch`
- `zap`
- `audit`
- `cat`
- `create`
- `edit`


miccal:
This one might need more re-working ...

vitorgalvao:
Like on the other PR, I don’t think we should be mentioning `--cask` that often. The point of `--cask` is not to replace `cask` (which was mandatory), but to offer more granularity *when needed* (otherwise it would’ve been an awful tradeoff, two extra characters every time).

It makes sense to mention the flag once, but on the explanation it should be ommited.

miccal:
Pushed some more changes.

vitorgalvao:
I still feel like there’s going to be confusion. For context, both in these repos and on the web it’s common to find people who don’t understand the difference between HB and HBC and which are confused as hell and never know when to use `brew cask`. I believe having our documentation now interchangeably mention *three* methods of dealing with casks (`brew`, `brew cask`, `brew --cask`) will only make it worse.

The end goal is to be able to get rid of `cask` as much as possible. `--cask` (which we should document as `--casks` because it’s typically used in situations of multiple results) is analogous to `--formulae` in that it’s *a command for more granularity, not a requirement*.

So we should mention all commands as `brew` only, and somewhere in the document (perhaps at the top) mention “these commands show results for both casks and formulae. To filter for casks alone, append the `--casks` flag”, thus making the documentation more focused and making it clear what `--casks` is and when to use it (most users will never need to).

miccal:
> mention all commands as brew only

I do not think we can do that yet since `brew install`, arguably the **most** important one, does not work for Casks at the moment:
```
|-> brew install google-chrome
Error: No available formula with the name "google-chrome" 
Found a cask named "google-chrome" instead. Try
  brew cask install google-chrome
```

> document as `--casks`

Happy to change that, but I used `--cask` since that is the flag that `brew` _always_ suggests:
```
|-> brew cask upgrade
Error: Calling brew cask upgrade is deprecated! Use brew upgrade --cask instead.
|-> brew cask outdated
Error: Calling brew cask outdated is deprecated! Use brew outdated --cask instead.
|-> brew cask list
Error: Calling brew cask list is deprecated! Use brew list --cask instead.
```

miccal:
Also, this is a _draft_ PR -- @vitorgalvao, @reitermarkus and @claui please feel free to push your own suggestions and changes.

vitorgalvao:
> > mention all commands as brew only
> 
> I do not think we can do that yet since `brew install`, arguably the **most** important one, does not work for Casks at the moment:

I meant “all commands in the scope of this PR change”. All the ones you touched can be `brew`-only.

ralphrmartin:
... "brew install, arguably the most important one, does not work for Casks at the moment"

This means the different things you want to do are done in inconsistent ways, which is not good. Surely this partial change should not have gone live, and a change should have waited until the whole thing is ready.

vitorgalvao:
> Surely this partial change should not have gone live, and a change should have waited until the whole thing is ready.

Things have to be tested and we need to see how they work in the wild. Commands have been losing their `cask` progressively and we can start to see where that works or not. `brew cask search` has been gone for a good while with no big complaints.

While it’s bad to bring live half-baked features, these do not depend on each other, they’re interface changes. Furthermore, a lot of the recent ones were made by a non-maintainer to whom we’re grateful for the contributions and the time spent on them, we’re not just going to say “yeah, thanks a lot for taking time out of your vacation to make this well done work that we want, but unless you take the major undertaking of doing everything, we don’t want it”.

You’re not being pragmatic. This change isn’t disruptive and doesn’t break functionality, so we’re evolving it as we can. If you want it done faster, PRs are welcome.

miccal:
Pushed some further changes after more testing of the commands.

miccal:
Updated to reflect the changes made in https://github.com/Homebrew/brew/pull/8808.

USAGE.md
```diff
@@ -103,27 +94,36 @@ $ brew install brew-cask-completion
 List all installed Casks:
 
 ```bash
-$ brew cask list
+$ brew list --casks
 atom          google-chrome     1password
 ```
 
+With `--full-name` the Tap name will also be included, _eg_: `foo/bar/boofar` for the Cask `boofar` installed from the Tap `foo/bar`.
```
vitorgalvao, line 101: `foobar` is confusing for a lot of people. Let’s use a real example.

core-code: you mean like `foobar2000`? :p

MikeMcQuaid:
> Commands that still need the `cask` prefix and do not accept the `--cask` flag:

Note: I'd hope that eventually this list can drop down to zero!

USAGE.md
vitorgalvao: > you mean like `foobar2000`? :p

Well, yes! That would be both fun and a good example (if not for us having removed it already) because the example would be clearer: `Homebrew/homebrew-cask/foobar2000`.

miccal: `--full-name` does not show the Tap name for `homebrew-cask-versions`, `homebrew-cask-drivers` or `homebrew-cask-fonts`; it only shows the Tap name for "outside" Taps.

For example:
```
|-> brew list --casks --full-name
apparency			# from main repo
bose-soundtouch			# from drivers
slack-beta			# from versions
miccal/miccal/m-mactex-no-gui	# from my own tap
```

vitorgalvao: > `--full-name` does not show the Tap name for `homebrew-cask`, `homebrew-cask-versions`, `homebrew-cask-drivers` or `homebrew-cask-fonts`; it only shows the Tap name for "outside" Taps.

That’s fine. Then we can use a real example from an external popular tap—like the adoptopenjdk—or make something up as long as it follows a realistic pattern as is clear (`kubrick/unfinished/napoleon.rb` or `apple/old-drivers/apology-mouse.rb`).

ralphrmartin:
The documentation probably needs to make clear what to do when there is both a cask and a bottle with the same name. In particular, cmake exists as both, and it needs to be clear how to refer to these separately. 
(It could, however, be simpler if the cask form of cmake were renamed cmake-app or similar, to avoid such confusion).

vitorgalvao:
> The documentation probably needs to make clear what to do when there is both a cask and a bottle with the same name.

For your reference now, the solution is to use the path with the tap name, much as you’d do if you had that conflict with a third-party tap.

> In particular, cmake exists as both, and it needs to be clear how to refer to these separately.

It’s just as important to think *why* both exist and if we could do with just one. Judging by the analytics, that’s not such a farfetched idea:

Formula: 115,113 (30 days), 349,395 (90 days), 1,488,711 (365 days)
Cask: 299 (30 days), 1,209 (90 days), 6,232 (365 days)

While the cask is far from unpopular in absolute terms, in comparison is multiple orders of magnitude behind.

> (It could, however, be simpler if the cask form of cmake were renamed cmake-app or similar, to avoid such confusion).

Unfortunately, that *creates* confusion, it doesn’t get rid of it. We’ve tried that exact approach and it was an awful decision we backtracked on.

MikeMcQuaid:
Not to say what we should do but: when we have duplicates it's often because one is a GUI and one isn't. Just to avoid rehashing past discussions: a solution that isn't on the table is shipping more GUIs in Homebrew/homebrew-core.

ralphrmartin:
Sorry if I was unclear. My request was not a plea for help to solve _this_ particular problem (with cmake), but a request to add to the _documentation_ how the user can express what he wants to do, for all who come across it and _similar_ problems in future.

USAGE.md
miccal: I have added a new one -- Google it if you don't get it 😆

miccal:
> > The documentation probably needs to make clear what to do when there is both a cask and a bottle with the same name.
> 
> For your reference now, the solution is to use the name with the tap path, much as you’d do if you had that conflict with a third-party tap.

Which is explained in the section `Other Ways to Specify a Cask`.

USAGE.md
```diff
@@ -1,22 +1,13 @@
...
-```bash
-$ brew --version
-0.9.5
-```
-
 ## Frequently Used Commands
 
-Homebrew Cask is implemented as a subcommand of Homebrew. All Homebrew Cask commands begin with `brew cask`. Homebrew Cask has its own set of command verbs many of which are similar to Homebrew’s. The most frequently-used
-commands are:
+Homebrew Cask is implemented as part of Homebrew. All Homebrew Cask commands begin with `brew`, which works for both Casks and Formulae. To filter for Casks alone, append the `--casks` flag. The most frequently-used commands are:
```
reitermarkus, line 5: Remove “To filter for Casks alone, append the `--casks` flag.” from here.

USAGE.md
```diff
@@ -47,48 +38,48 @@ $ brew cask install google-chrome
 Easy peasy:
 
 ```bash
-$ brew cask uninstall google-chrome
+$ brew uninstall google-chrome
 ```
 
 This will both uninstall the Cask and remove applications which were moved to `/Applications`.
 
 To uninstall all versions of a Cask, use `--force`:
```
reitermarkus, line 46: Let's remove this. It should not be possible anymore to have multiple versions installed.

USAGE.md
```diff
@@ -103,27 +94,36 @@ $ brew install brew-cask-completion
 List all installed Casks:
 
 ```bash
-$ brew cask list
+$ brew list --casks
 atom          google-chrome     1password
 ```
 
+With `--full-name` the Tap name will also be included, _eg_: `purple/monkey/dishwasher` for the Cask `dishwasher` installed from the Tap `purple/monkey`.
```
reitermarkus, line 101: `e.g.`

USAGE.md
```diff
@@ -103,27 +94,36 @@ $ brew install brew-cask-completion
...
 Show details about a specific Cask:
 
 ```bash
-$ brew cask info caffeine
-caffeine: 1.1.1
-http://lightheadsw.com/caffeine/
-Not installed
-From: https://github.com/Homebrew/homebrew-cask/blob/master/Casks/caffeine.rb
+$ brew info keepingyouawake
+Error: No available formula with the name "keepingyouawake"
```
reitermarkus, line 109: This should be fixed not to show an error if a cask exists.

USAGE.md
```diff
@@ -139,37 +139,38 @@ The primary Homebrew Cask Tap includes most of the Casks that a typical user wil
 | [Homebrew/cask-fonts](https://github.com/Homebrew/homebrew-cask-fonts)       | contains Casks that install fonts
 | [Homebrew/cask-drivers](https://github.com/Homebrew/homebrew-cask-drivers)   | contains Casks that install drivers for various devices
 
-You can tap any of the above with a `brew tap` command:
+You can Tap any of the above with a `brew tap` command:
```
reitermarkus, line 142: “tap” is a verb here.

reitermarkus (changes requested):

USAGE.md
miccal: Perhaps this could be a warning instead of an error? Personally, I do not think showing this kind of message is a bad thing.

MikeMcQuaid: Agreed with @reitermarkus; the error isn't "bad" but it is inconsistent with other commands at this point.

miccal: I will investigate this over the weekend.

reitermarkus (approved):

miccal:
Thank you @reitermarkus.

PR #88696, opened 2020-09-04
Update orange from 3.24.1 to 3.26.0

suschizu:


suschizu:
the download is since 3.25.0 bundled with python37 (or python36) - see: https://download.biolab.si/download/files/
no more download without python - not sure if we want this @vitorgalvao

core-code:
ready-to-use sounds good.

core-code (approved):

vitorgalvao:
We prefer it, even. Apps that bundle their own dependencies (usually python or java) are way less trouble for us (because they’re less trouble for users and we get less bug reports).

PR #88940, opened 2020-09-09
Link binary makemkvcon in makemkv cask

sammys:
Make the console binary, makemkvcon, accessible when cask is installed.


miccal:
Justification for adding the `binary` stanza can be found [here](https://www.makemkv.com/developers/).

reitermarkus (approved):

PR #88936, opened 2020-09-08
Update shotcut from 20.07.11 to 20.09.01

SeekingMeaning:
Created with `brew bump-cask-pr`.

PR #88947, opened 2020-09-09
Revert "Update shotcut from 20.07.11 to 20.09.01"

core-code:
Reverts Homebrew/homebrew-cask#88936

### DIFF
--- a/Casks/shotcut.rb
+++ b/Casks/shotcut.rb
@@ -1,6 +1,6 @@
 cask "shotcut" do
-  version "20.09.01"
-  sha256 "c3838f2c7c09ba749c5f66bbf03c8041795ed0b2c38a2f32654e8bf85ac7a806"
+  version "20.07.11"
+  sha256 "8b94e22bc54fcb5ee6fb5f1642ecb66ef92b128df6a5b488f2415c41c7a6175f"
 
   # github.com/mltframework/shotcut/ was verified as official when first introduced to the cask
   url "https://github.com/mltframework/shotcut/releases/download/v#{version}/shotcut-macos-signed-#{version.no_dots}.dmg"


core-code:
@vitorgalvao can you merge this? it seems i can't merge my own PRs anymore. *grrr*

vitorgalvao (approved):

vitorgalvao:
> it seems i can't merge my own PRs anymore. _grrr_

It’s because PRs now need to be approved and we can’t approve our own PRs. Thought the bot should have approved this one.

core-code:
thanks for the quick help

> It’s because PRs now need to be approved and we can’t approve our own PRs. 

makes sense for code changes but not for caskfiles. things are bad enough already with recent changes, and now this?

vitorgalvao:
Note that this way we no longer need to purposefully merge, we can just approve and the bot will merge it typically in less than a minute. So it does allow us to approve something before the tests finish and have it merged as soon as they’re done. Essentially we’re replacing merging with approving, though I agree it is annoying we can’t merge our own cask fixes.

I’m guessing this was implemented as part of [my suggested compromise](https://github.com/Homebrew/homebrew-cask/pull/88388#issuecomment-685828038), which I’m fine with scrapping if it’s causing problems. I used to [merge with a script](https://github.com/vitorgalvao/tiny-scripts/blob/master/cli-merge-button) and now I [approve with a script](https://github.com/vitorgalvao/tiny-scripts/blob/master/cli-approve-button), so I’m fine either way.

reitermarkus:
You can let the bot approve here: https://github.com/Homebrew/homebrew-cask/actions?query=workflow%3A%22Self-approve+a+Pull+Request%22

core-code:
thanks i've bookmarked the approval

reitermarkus:
> Though the bot should have approved this one.

Should it? It's a version decrease.

> things are bad enough already with recent changes, and now this?

What is so bad still about the recent changes? I think the process is now as streamlined as it gets, you don't even have to wait for CI and press the merge button manually afterwards, just approve and let @BrewTestBot handle the rest.

core-code:
> What is so bad still about the recent changes? 

the workflow that was suggested to me a week back, e.g. applying ci-verified-manually or ci-skip-appcast is definitely a large step back. previously i would just verify thats things are OK manually and then hit merge. applying a tag instead, waiting for half an hour constantly checking if the second CI did finish again, see if if will be automerged or submit a review and/or merge manually. this is already a burden for one PR, but if you want to do 5 of them you are busy for 2 hours. removing the ability to approve the own PRs doesn't help either.

approving using the bot definitely helps restore some of previous efficiency, so thanks there. and since @vitorgalvao does the bulk of the work there i'm fine with whatever he can work with.

what i find more disturbing is that the changes that impact all the work done on the caskfiles have not been discussed beforehand with the actual people involved. feel free to refer to the last few posts on https://github.com/Homebrew/homebrew-cask/issues/55268

reitermarkus:
> see if if will be automerged or submit a review and/or merge manually

It will be auto merged if it is approved and CI is passing, so just apply the label and approve immediately. (CI will now actually re-run automatically after applying the label since I added https://github.com/Homebrew/homebrew-cask/blob/master/.github/workflows/rerun-workflow.yml.)

If it doesn't pass after applying the label, that means it would previously have been merged even if there was another problem.

core-code:
> It will be auto merged if it is approved and CI is passing, so just apply the label and approve immediately. 

thanks, will do that

vitorgalvao:
> > Though the bot should have approved this one.
> 
> Should it? It's a version decrease.

Unwanted version increases are considerably more common than unwanted version decreases.

The most common occurrence of the former is users sending version bumps to non-stable versions, while the most common occurrence of the latter is reverting erroneous merges of the former.

In sum, version decreases are more likely to be correct—and have a bigger urgency—than version increases.

reitermarkus:
This logic exists because of a security concern, not a correctness concern.

I think however that auto-approving PRs with the following criteria would be no problem:

- author is a member
- version decreased to *exactly* the previous version according to `git log`

vitorgalvao:
@reitermarkus The self-approving Workflow isn’t available on other repos. Shouldn’t actions be auto-syncing?

Until we can self-approve in other places, someone please approve https://github.com/Homebrew/homebrew-cask-versions/pull/9610 so we can fix the CI errors.

Unrelated, the fonts repo has a [`Scheduled automerge.`](https://github.com/Homebrew/homebrew-cask-fonts/actions?query=workflow%3A%22Scheduled+automerge.%22) weirdly named with a period at the end.

reitermarkus:
> The self-approving Workflow isn’t available on other repos. Shouldn’t actions be auto-syncing?

Will need to be added to the glob pattern.

> `Scheduled automerge.` 

I renamed this to just `Automerge`, not sure why the old name is also still showing up.

PR #89077, opened 2020-09-11
brew bump-cask-pr above cask-repair

vitorgalvao:
Continuation of https://github.com/Homebrew/brew/pull/7986.

Leaving a link to `cask-repair` for people who are accustomed to it or might prefer it, while the new method gains traction.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,18 +11,13 @@
 
 ## Updating a Cask
 
-Notice an application that's out-of-date in Homebrew Cask? In most cases, it's very simple to update it. We have a [script](https://github.com/vitorgalvao/tiny-scripts/blob/master/cask-repair) that will ask for the new version number, and take care of updating the Cask file and submitting a pull request to us:
+Notice an application that's out-of-date in Homebrew Cask? In most cases, it's very simple to update it. We have a command that will accept a new version number and take care of updating the Cask file and submitting a pull request to us:
 
 ```bash
-# install and setup script - only needed once
-brew install vitorgalvao/tiny-scripts/cask-repair
-cask-repair --help
-
-# use to update <outdated_cask>
-cask-repair <outdated_cask>
+brew bump-cask-pr --version <new_version> <outdated_cask>
 ```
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -11,18 +11,13 @@
...
-brew install vitorgalvao/tiny-scripts/cask-repair
-cask-repair --help
-
-# use to update <outdated_cask>
-cask-repair <outdated_cask>
+brew bump-cask-pr --version <new_version> <outdated_cask>
 ```
 
-If there is a more complicated change, or there is a case where `cask-repair` fails (for example, where a Cask uses a [`url do` block](https://github.com/Homebrew/homebrew-cask-versions/blob/2bf0f13dd49d263ebec0ca56e58ad8458633f789/Casks/vlc-nightly.rb#L5L10) or the [`language` stanza](https://github.com/Homebrew/homebrew-cask/blob/306b8fbd9502036f1ca742f70c569d8677b62403/Casks/firefox.rb#L4L74)), you can also follow the steps in [Adding a Cask](doc/development/adding_a_cask.md) to do the same thing manually. Remember to update the `version` and `sha256` values.
+If you want more control over the PR or are looking for our old tool, install [`cask-repair`](https://github.com/vitorgalvao/tiny-scripts/blob/master/cask-repair) (`brew install vitorgalvao/tiny-scripts/cask-repair`). You can also follow the steps in [Adding a Cask](doc/development/adding_a_cask.md) for more complicated changes.
```
reitermarkus, line 20: > If you want more control over the PR

What's the difference in control to `bump-cask-pr`?

vitorgalvao: > What's the difference in control to `bump-cask-pr`?

I doesn’t require all the information when called. `cask-repair` started as (and has first-class support for) being called with minimal info and asking for what’s missing as it goes along.

Before any edits, it shows you the cask as it is.

It can open the appcast and homepage in the browser automatically for a user to inspect, allowing one to start fixing before they have all the information.

It allows directly editing the cask (via a text editor) on a loop until it passes tests, meaning one can iterate in real-time until it works, as opposed to having to start all over again when the command fails.

All of these make it a more powerful fit when you know a cask needs updating but you don’t know exactly to what (e.g. a version with a comma is not immediately obvious from the website alone, you need to see the cask to know what format it expects).

All that said, I’m expecting to eventually remove `cask-repair` completely. I still fix (or more likely accept PRs for fixes) bugs but I’m not adding features. I’m only leaving it here to make for an easier transition.

reitermarkus (approved):

PR #89147, opened 2020-09-12
Update gpg-suite-pinentry.rb

maboloshi:
Add a reminder for `~/.gnupg/gpg-agent.conf` settings.
Refer to the description of `caveats` in [pinentry-mac.rb](https://github.com/Homebrew/homebrew-core/blob/5e6b22d05db84553765213aa8bbb73dfa5ae4b24/Formula/pinentry-mac.rb#L35).

### DIFF
--- a/Casks/gpg-suite-pinentry.rb
+++ b/Casks/gpg-suite-pinentry.rb
@@ -5,6 +5,7 @@ cask "gpg-suite-pinentry" do
   url "https://releases.gpgtools.org/GPG_Suite-#{version}.dmg"
   appcast "https://gpgtools.org/releases/gka/appcast.xml"
   name "GPG Suite Pinentry"
+  desc "A GUI pinentry program from the GPG Suite for GPG"
   homepage "https://gpgtools.org/"
 
   auto_updates true
--- a/Casks/gpg-suite-pinentry.rb
+++ b/Casks/gpg-suite-pinentry.rb
@@ -5,7 +5,7 @@ cask "gpg-suite-pinentry" do
   url "https://releases.gpgtools.org/GPG_Suite-#{version}.dmg"
   appcast "https://gpgtools.org/releases/gka/appcast.xml"
   name "GPG Suite Pinentry"
-  desc "A GUI pinentry program from the GPG Suite for GPG"
+  desc "Pinentry GUI for GPG Suite"
   homepage "https://gpgtools.org/"
 
   auto_updates true
--- a/Casks/gpg-suite-pinentry.rb
+++ b/Casks/gpg-suite-pinentry.rb
@@ -106,5 +106,12 @@ cask "gpg-suite-pinentry" do
 
   caveats do
     files_in_usr_local
+
+    <<~EOS
+      You can now set this as your pinentry program like
+
+      ~/.gnupg/gpg-agent.conf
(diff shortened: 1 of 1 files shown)


Casks/gpg-suite-pinentry.rb
```diff
@@ -106,5 +106,12 @@
 
   caveats do
     files_in_usr_local
+
+    <<~EOS
+      You can now set this as your pinentry program like
+
+      ~/.gnupg/gpg-agent.conf
+          pinentry-program #{HOMEBREW_PREFIX}/MacGPG2/libexec/pinentry-mac.app/Contents/MacOS/pinentry-mac
```
reitermarkus, line 114: If it's installed by an installer, it will always be `/usr/local` irregardless of what `HOMEBREW_PREFIX` is.

vitorgalvao: This seems to fall under “you should know this about the app, it’s not installation related”. I don’t see a strong reason to keep this one.

Casks/gpg-suite-pinentry.rb
```diff
@@ -106,5 +106,12 @@
 
   caveats do
     files_in_usr_local
+
+    <<~EOS
+      You can now set this as your pinentry program like
```
vitorgalvao, line 111: We don’t use `like`. `by running` is more appropriate.

maboloshi: This is largely copied here from the description [pinentry-mac. rb#L37](https://github.com/Homebrew/homebrew-core/blob/5e6b22d05db84553765213aa8bbb73dfa5ae4b24/Formula/pinentry-mac.rb#L37)

Casks/gpg-suite-pinentry.rb
maboloshi: In fact, this cask is a substitute for [pinentry-mac](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pinentry-mac.rb), which achieves exactly the same functions and has better UI effect, and the [project](https://github.com/GPGTools/pinentry-mac) of `pinentry-mac` has not been updated since 2015.

Casks/gpg-suite-pinentry.rb
vitorgalvao: > This is largely copied here from the description pinentry-mac. rb#L37

That’s a formula, not a cask. Different messaging.

Casks/gpg-suite-pinentry.rb
```diff
@@ -106,5 +106,12 @@
...
+
+      ~/.gnupg/gpg-agent.conf
```
vitorgalvao, line 113: Why is this two lines. You mean the second line should be placed inside the file mentioned on the first? That’s not very clear.

Casks/gpg-suite-pinentry.rb
vitorgalvao: > In fact, this cask is a substitute for [pinentry-mac](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pinentry-mac.rb), which achieves exactly the same functions and has better UI effect, and the [project](https://github.com/GPGTools/pinentry-mac) of `pinentry-mac` has not been updated since 2015.

We had already noticed and commented on that internally. Mind making a PR to remove the formula as well? Please @mention me so I can post the previous discussion as a refresher to other maintainers.

vitorgalvao (approved):

PR #89251, opened 2020-09-14
Update minecraft-server from 1.15.2 to 1.16.3

Blacktiger:

### DIFF
--- a/Casks/minecraft-server.rb
+++ b/Casks/minecraft-server.rb
@@ -1,6 +1,6 @@
 cask "minecraft-server" do
-  version "1.15.2,bb2b6b1aefcd70dfd1892149ac3a215f6c636b07"
-  sha256 "80cf86dc2004ec6a2dc0183d1c75a9af3ba0669f7c332e4247afb1d76fb67e8a"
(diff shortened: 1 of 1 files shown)


Blacktiger:
It looks like the ci build couldn't reach https://minecraft.net/ for some reason.

reitermarkus:
Weird that only the homepage fails and the appcast doesn't.

Blacktiger:
Perhaps it's related to the homepage doing a 302 redirect to https://minecraft.net/en-us?

core-code:
> Weird that only the homepage fails and the appcast doesn't.

i think i've requested a re-run here, and before the re-run the appcast check did also have an error because it did not contain the version number ...instead it contained HTML about some 'security protection'

hypothesis: all URLs are queried in short succession, triggering some defence on the server. if the appcast check runs before the others, it might still be OK while later reachability checks only get either 'defence html' or just some bad status code

reitermarkus:
The homepage failure is because of this: https://github.com/Homebrew/brew/pull/8730

core-code:
wow. must be a Microsoft server.

reitermarkus:
You'd think it would work in a Microsoft CI environment. 😄

PR #89314, opened 2020-09-15
Update youdaonote from 3.6.2 to 3.6.3

ran-dall:

### DIFF
--- a/Casks/youdaonote.rb
+++ b/Casks/youdaonote.rb
@@ -1,5 +1,5 @@
 cask "youdaonote" do
-  version "3.6.2"
+  version "3.6.3"
   sha256 "6529627d2b61ec455e45580c1bd47aa7585f6c173f77bb486700f0e7fbe1cd1a"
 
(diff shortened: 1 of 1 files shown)


ran-dall:
This is a downgrade but it's showing `3.6.3` on the `appcast`

core-code:
if the SHA didn't change, we shouldn't change the version either. it definitely does not contain version 3.6.3 regardless of what the appcast says

core-code:
```
svs: 3.6.2
bv: 6371
```

ran-dall:
@core-code Great, can we do something about flagging this for the MLv1? Also `wordpresscom` has a similar problem.

core-code:
i think it  just needs to add this specific version  to the 'fp.cask.appcastcontains.blacklist' file so it doesn't show up anymore. 

@suschizu can you do that? similar for `wordpresscom`?

ran-dall:
Maybe said blacklist can reset every 30 days or something?

core-code:
the idea there is that by only putting specific versions on it, it kind of has a reset anyway. but i agree that basically most lists ideally need some reset or expiration

PR #89421, opened 2020-09-16
Add Nova 1.0

hugo:

### DIFF
--- a/Casks/nova.rb
+++ b/Casks/nova.rb
@@ -1,0 +1,19 @@
+cask "nova" do
+  version "1.0"
+  sha256 "c5414f5cf6888a8c53f91b53d5bdd1cd246e31aafad3c3bec381c10ef6a44a94"
+
+  # panic-c157.kxcdn.com was verified as official when first introduced to the cask
+  url "https://panic-c157.kxcdn.com/nova/Nova%20#{version}.zip"
+  name "Panic Nova"
+  desc "Native code editor"
+  homepage "https://nova.app/"
+
+  auto_updates true
+
+  app "Nova.app"
+  binary "#{appdir}/Nova.app/Contents/MacOS/Nova", target: "nova"
+
+  zap trash: [
+    "~/Library/Application Support/Nova",
+  ]
+end
--- a/Casks/nova.rb
+++ b/Casks/nova.rb
@@ -13,7 +13,5 @@ cask "nova" do
   app "Nova.app"
   binary "#{appdir}/Nova.app/Contents/MacOS/Nova", target: "nova"
 
-  zap trash: [
-    "~/Library/Application Support/Nova",
-  ]
+  zap trash: "~/Library/Application Support/Nova"
 end


Casks/nova.rb
```diff
@@ -0,0 +1,19 @@
+cask "nova" do
+  version "1.0"
+  sha256 "c5414f5cf6888a8c53f91b53d5bdd1cd246e31aafad3c3bec381c10ef6a44a94"
+
+  # panic-c157.kxcdn.com was verified as official when first introduced to the cask
+  url "https://panic-c157.kxcdn.com/nova/Nova%20#{version}.zip"
+  name "Nova"
+  desc "Native code editor, from Panic"
```
reitermarkus, line 8: Add the vendor to the name instead.

reitermarkus (changes requested):

Casks/nova.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage "https://nova.app/"
+
+  auto_updates true
+
+  app "Nova.app"
+
+  zap trash: [
+    "/usr/local/bin/nova",
```
reitermarkus, line 16: Add this as a `binary`.

hugo: I don’t know what this means. It would have been helpful if you pointed to documentation that described what you are asking for.

Casks/nova.rb
hugo: I don’t see any guidance or documentation about this. If you want the change, please make it yourself. Otherwise, you are welcome to close the PR.

reitermarkus: > I don’t see any guidance or documentation about this.

You're right, there wasn't much at the time you submitted this. We have since added some examples here: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/desc.md

Casks/nova.rb
reitermarkus: Take a look at https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/binary.md.

hugo: I'm not sure I understand from that document why this stanza would be appropriate here. This binary is installed optionally via the Nova UI. There isn't a guarantee, as far as I know, of where the source file will reside. Arguably it's also not guaranteed that it will be copied to `/usr/local/bin/nova` and so we might be deleting the wrong binary inadvertently... I'm not sure what the best approach is here honestly.

reitermarkus: Well, it has to be copied from within the `.app` bundle, so we might as well link it ourselves, which also avoids deleting a wrong binary if one is already present and the manual installation step using the GUI.

hugo: Would I be right to expect that the `audit` command would catch if the source binary moved location or was renamed?

Casks/nova.rb
hugo: Thanks for the link, I've made the recommende change.

vitorgalvao (approved):

reitermarkus (approved):

PR #89424, opened 2020-09-16
Remove `container` workaround for Parallels.

reitermarkus:
This was added for Parallels 14, seems like it is not needed anymore.


core-code:
the "Parallels Desktop.app" stored on the DMG has the invisibiity flag set. will HBC make it visible after copying it over?

reitermarkus:
@core-code, should be visible now.

PR #89651, opened 2020-09-22
Update emclient from 7.2.37965 to 8.0.2958.0

suschizu:

### DIFF
--- a/Casks/emclient.rb
+++ b/Casks/emclient.rb
@@ -1,8 +1,8 @@
 cask "emclient" do
-  version "7.2.37965"
-  sha256 "3d8e9f7714143adacfcc8256c45beebaae58256a813eb56f46cf3d44d772103f"
+  version "8.0.2958.0"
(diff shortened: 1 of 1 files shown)


Casks/emclient.rb
```diff
@@ -1,8 +1,8 @@
 cask "emclient" do
-  version "7.2.37965"
-  sha256 "3d8e9f7714143adacfcc8256c45beebaae58256a813eb56f46cf3d44d772103f"
+  version "8.0.2958.0"
+  sha256 "236ced99ef79c3d4973d4a9a69bba03a7d0e29586035823f20f8f70bf2426d83"
 
-  url "https://cdn-dist.emclient.com/dist/v#{version}_Mac/setup.pkg"
```
reitermarkus, line 5: This still seems to exist.

core-code: thanks, i've reverted to the previous url, good catch

i've also updated the appcast since the release-notes page is both noisy html and contains old versons too - which is the reason why this update was missing so long

reitermarkus (approved):

PR #89822, opened 2020-09-24
Update rekordbox appcast

core-code:
HTML release-notes pages should only be used as a last resort

### DIFF
--- a/Casks/rekordbox.rb
+++ b/Casks/rekordbox.rb
@@ -3,7 +3,7 @@ cask "rekordbox" do
   sha256 "3ad7e16b5c72e7f5a0b90fd7cdce51eb744dc0fc0679c949ed07b15f33595fa2"
 
   url "https://cdn.rekordbox.com/files/#{version.after_comma}/Install_rekordbox_#{version.before_comma.dots_to_underscores}.pkg_.zip"
-  appcast "'https://rekordbox.com/en/download/"
+  appcast "https://rekordbox.com/en/download/"
   name "rekordbox"
   desc "Free Dj app to prepare and manage your music files"
   homepage "https://rekordbox.com/en/"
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao i often get the impression that release-notes pages are used preferentially as an appcast. however, HTML release notes are absolutely the worst for the purpose of finding updates. most of them are too noisy to observe for changes. and all of them also contain old version numbers so it is impossible to check when the current version stops being contained in them. this combination makes them useless for finding updats. it would be great if our official documentation could be changed to mention that pages that *also* contain old version numbers (like release-notes pages) should only be used as a last-resort when pages that *only* contain the current version number are not available.

reitermarkus:
> impossible to check when the current version stops being contained in them

In my opinion, that's just a limitation of the `appcast` stanza, not the URL itself. I think with `livecheck` it will be possible to extract the latest (or first matching) version with a regex (or other method), then you can just compare the matched version with the current version. Having an actual version to compare is much better than just knowing if the current version is contained or not.

core-code:
yes, the 'contains' check is a last resort that has to be used when the appcast is too noisy to observe for changes (unfortunately 'too noisy' is true for most HTML pages). 

being a last resort doesn't mean it is not incredibly useful - it yields us a dozen cask-repairs per day ( e.g. http://macupdater.net/cask-repair/appcast_containsversion-2020-09-24.txt ) and has yielded thousands so far.

i am currently putting more noisy appcasts on a blacklist so if they also contain old versions (like this one does now) this just means no more PRs will come for affected casks. and if we had good guidelines on that point all of the manual work of switching appcasts over wouldn't even be necessary.

i agree that `livecheck` is great and will (hopefully) make most of our system obsolete. but 1.) it is not in place yet, i haven't seen hordes of PRs coming from livecheck yet. 2.) extracting the current version with a regexp is also easier and less error-prone when old version numbers aren't contained in the document. so avoiding changelogs and preferring download sites with versioned download links will be useful even when and if `livecheck` is in place

reitermarkus:
> easier and less error-prone when old version numbers aren't contained in the document

Good point.

reitermarkus (approved):

vitorgalvao:
> it would be great if our official documentation could be changed to mention that pages that _also_ contain old version numbers (like release-notes pages) should only be used as a last-resort when pages that _only_ contain the current version number are not available.

We do point that out:

> Webpages that only change on new version releases are preferred, as are sites that do not contain previous version strings (i.e. avoid changelog pages if the download page contains the current version number but not older ones).

I’m fine with changing that whole paragraph to make it clearer.

core-code:
ah ok thanks i remember now that this is already the case.

PR #89960, opened 2020-09-27
Update markdown-service-tools from 2.16 to 3.0.3

jfexyz:


joshuajabbour:
Ok, sorry for the multiple fixups... some issues found when installing the updated formula. But all should work now!

Casks/markdown-service-tools.rb
```diff
@@ -1,35 +1,42 @@
 cask "markdown-service-tools" do
-  version "2.16"
-  sha256 "fcc9c9b88b164547315cde305d4b4096243268b84d24c054e2731449c9e47633"
+  version "3.0.3"
+  sha256 "7bdfc033d4fb84818b7062eda5aae73c6b593477a7511d3d438909d0b7f52cad"
 
   url "https://brettterpstra.com/downloads/MarkdownServiceTools#{version}.zip"
   name "Markdown Service Tools"
+  desc "Collection of services designed to make writing Markdown text that much easier"
```
miccal, line 7:
```suggestion
  desc "Collection of services for Markdown-formatted text"
```

Casks/markdown-service-tools.rb
```diff
@@ -38,6 +45,13 @@
...
   service "MarkdownServiceTools#{version}/md - Wrap - Angle Brackets.workflow"
+  service "MarkdownServiceTools#{version}/md - Wrap - Backticks.workflow"
   service "MarkdownServiceTools#{version}/md - Wrap - Parenthesis.workflow"
   service "MarkdownServiceTools#{version}/md - Wrap - Square Brackets.workflow"
+
+  caveats <<~EOS
+    All Markdown conversion services require MultiMarkdown:
+
+      brew install multimarkdown
+  EOS
```
miccal, line 56: Is this `caveat` still relevant? I see no mention that this is required on the `homepage`.

reitermarkus: If it is relevant, it might be best to just depend on `multimarkdown`.

joshuajabbour: It's a caveat called out on the project's website. A bunch of the workflows will not work without multimarkdown.

I did not just depend on multimarkdown because a) I couldn't figure out how to depend on a formula in a cask b) there are multiple packages that provide a markdown converter (see the notes for multimarkdown itself). I didn't want to force one (even though MM is the default and probably best option)

Casks/markdown-service-tools.rb
joshuajabbour: Ok. I just copied the description from the project's own website. But whatever you want ¯\_(ツ)_/¯

Casks/markdown-service-tools.rb
joshuajabbour: And, multimarkdown requirement called out twice on this page: https://brettterpstra.com/2020/06/15/markdown-service-tools-v3-dot-0/

miccal: Thank you for your reply -- I have added a `depends_on formula: "multimarkdown"` instead of the `caveat`.

reitermarkus (approved):

PR #90037, opened 2020-09-29
New prowritingaid appcast based on web request

vitorgalvao:


core-code:
it is online now. i thought there was an option to "must_contain" to disable looking for the version number but didn't find it

vitorgalvao:
> i thought there was an option to "must_contain" to disable looking for the version number but didn't find it

`must_contain: ''` (empty string) or `must_contain: ' '` (space) should work.

But that’s a good point. It’s a weird case for `appcast` and I wonder if we could have something better.

core-code:
i do think we have a few caks with checksum appcasts which never contain the version:

```
armitage
cr
folding-at-home
igv
monal
paraview
xdm

```

core-code:
of course when merging that PR updates to beta versions will not be prevented any more.

shows again that appcasts serve two purposes, both discovering new versions as well as preventing invalid versions. 

in an ideal world we'd have two different stanzes, with both being arrays and accepting multiple URLs.

vitorgalvao:
> shows again that appcasts serve two purposes, both discovering new versions as well as preventing invalid versions.
>
> in an ideal world we'd have two different stanzes, with both being arrays and accepting multiple URLs.

Agreed we could rethink `appcast` for its current reality. When it was first introduced it had a mandatory parameter that was the checksum of the appcast’s page itself (so if it ever changed, you’d know there was an update).

An option would be to block version bump PRs from users to certain casks. We could warn them those already have automated systems in place, and that if there’s an error, the `appcast` itself should be fixed.

core-code:
> checksum of the appcast’s

seems like a good-in-theory-bad-in-practice thing

> An option would be to block version bump PRs from users to certain casks

the problem is, which ones are those? we are generally quite good with everything having a sparkle, electron/squirell or github feeds but even here i am not sure if its 100%.

in practice the system of appcasts preventing automerging bad PRs works reasonably well for the time being

vitorgalvao:
> seems like a good-in-theory-bad-in-practice thing

It was!

core-code (approved):

PR #90145, opened 2020-09-30
Update archi.rb

core-code:
i think we should give up on this app


core-code:
ping @vitorgalvao should we delete this cask?

vitorgalvao:
> ping @vitorgalvao should we delete this cask?

[Definitely.](https://github.com/Homebrew/homebrew-cask/pull/90364)

PR #90018, opened 2020-09-28
Update macx-dvd-ripper-pro from 6.5.3,20200630 to 6.5.3,20200927

ran-dall:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Created with [`cask-repair`](https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask).

PR #90157, opened 2020-10-01
prowritingaid: remove must_contain, add desc

miccal:

See https://github.com/Homebrew/homebrew-cask/pull/90018 for background.

### DIFF
--- a/Casks/prowritingaid.rb
+++ b/Casks/prowritingaid.rb
@@ -3,9 +3,9 @@ cask "prowritingaid" do
   sha256 "43f30a5a7135026f36849c6b21b1321a82684d189cf7494ca4cff6c3e53805ee"
 
   url "https://cdn.prowritingaid.com/prowritingaid-desktop/desktop_#{version.dots_to_underscores}/ProWritingAid.dmg"
-  appcast "https://macupdater.net/cgi-bin/check_urls/check_url_sizeandmodified.cgi?url=https://cdn.prowritingaid.com/prowritingaid-desktop/ProWritingAid.dmg",
-          must_contain: ""
+  appcast "https://macupdater.net/cgi-bin/check_urls/check_url_sizeandmodified.cgi?url=https://cdn.prowritingaid.com/prowritingaid-desktop/ProWritingAid.dmg"
   name "ProWritingAid"
+  desc "Grammar checker, style editor, and writing mentor"
   homepage "https://prowritingaid.com/"
 
   app "ProWritingAid.app"


Casks/prowritingaid.rb
```diff
@@ -4,8 +4,9 @@
 
   url "https://cdn.prowritingaid.com/prowritingaid-desktop/desktop_#{version.dots_to_underscores}/ProWritingAid.dmg"
   appcast "https://macupdater.net/cgi-bin/check_urls/check_url_sizeandmodified.cgi?url=https://cdn.prowritingaid.com/prowritingaid-desktop/ProWritingAid.dmg",
-          must_contain: ""
+          must_contain: "GMT"
```
reitermarkus, line 7: How is that helpful?

reitermarkus (changes requested):

Casks/prowritingaid.rb
miccal: [See the comment here](https://github.com/Homebrew/homebrew-cask/pull/90018#issuecomment-700867090) for more background.

It was noticed that having `""` or `" "` was not working.

reitermarkus: This is beginning to look very workaround-y. I'd rather this functionality be implemented in `brew cask audit`.

miccal: I do not have much knowledge of the ML update system, so I will leave it to @ran-dall and @core-code to explain.

However, I think that if _no other_ suitable `appcast` can be found, then it does no harm to have _something_, no?

reitermarkus: Sure it does no harm to have *something*, but it would be better to have something *official*.

core-code: feel free to find something better :) for some casks the only way to detect updates is to look at the file itself. this is also similar to having a checksum file in the appcast, which is already done for a handful.

core-code: also, looking at the downloaded file from the homepage kind-of is official, no? the appcast will only change if they modify the official download.

reitermarkus: > looking at the downloaded file from the homepage kind-of is official, no?

It's not the homepage itself I am talking about. I am talking about the various CGI scripts which are living on your server instead of being incorporated into `brew`. There's currently no official way to do anything useful with just the file size/modified time unless you keep track of these yourself with some custom tool or script. Then again, there's also no official tool to do anything useful with the appcast itself either … so I guess my point is moot until `livecheck` works for casks.

core-code: > I am talking about the various CGI scripts which are living on your server instead of being incorporated into brew.

i did offer to provide (and pay for) a server to host these and other stuff for HBC with full access for HBC member(s). i even offered to buy a suitable domain name. the response was that there is no interest. i am not sure it would have been possible to offer anything more than that.

> There's currently no official way to do anything useful with just the file size/modified time unless you keep track of these yourself with some custom tool or script. 

indeed. we obviously check each and every appcast every day, but publish only the changes

http://macupdater.net/cask-repair/

publishing the contents of each appcast every day could become expensive since it is around 60MB per day

> Then again, there's also no official tool to do anything useful with the appcast itself either

yes, everyone is free to check appcasts himself whether they refer to our server or any other server. no real distinction here.

reitermarkus: > the response was that there is no interest

I think there is no interest because this doesn't need to exist on a server at all. The same checks can be implemented without having a server running by just doing what the CGI scripts are doing locally.

core-code: sure, and i've advocated the very same thing in the past.

but who will implement it?

core-code: if there is ever work in that direction, i think the first step is disentangling the dual use of the 'appcast' stanza. it is used 1.) to detect new updates and 2.) to find out which updates are valid and can be automerged

those do sometimes overlap, but not always.

core-code:
we could just omit the 'must_contain' stanza.  since the version number is not contained in the appcast, we cannot automatically reject bad PRs. it may be better if a maintainer verifies upcoming PRs anyway.

core-code:
i don't get it. CI fails with appcast error despite the 'ci-skip-appcast' label

core-code (approved):

reitermarkus (approved):

PR #90163, opened 2020-10-01
Update sage.rb

hypersad:


m416kar98k:
image

vitorgalvao:
Homebrew itself doesn’t support El Capitan, so the conditional is too low.

While it’s great that the software runs under your machine, there’s no guarantee it will work for other users. I’ve both seen that happen and reproduced it—something as simple as doing a clean OS install VS an upgrade makes a difference. For OS compatibility, we follow whatever upstream recommends. If they change their recommendation, so will we.

Thank you anyway.

PR #90229, opened 2020-10-02
Update cisco-jabber to add appcast

core-code:


Casks/cisco-jabber.rb
```diff
@@ -3,6 +3,7 @@
   sha256 "41c70fc9dc30ad04ccfb1c06c2ee1416bb09fccfb0f38b7c37166b78cde195e7"
 
   url "https://binaries.webex.com/jabberclientmac/#{version}/Install_Cisco-Jabber-Mac.pkg"
+  apcast "https://www.webex.com/downloads/jabber.html"
```
reitermarkus, line 6: Typo here.

reitermarkus: This should be fixed to actually produce an error, at least when auditing.

core-code: i've noticed this before but i fear an 'unknown stanza' error might be difficult to implement because caskfiles can contain arbitrary ruby code

miccal (approved):

Issue #88467, opened 2020-08-31
Error: Cask 'aquiline-check' definition is invalid: invalid 'depends_on macos' value: ">= 10.11"


masonke:
#### General troubleshooting steps

- [X ] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [X] I have retried my command with `--force`.
- [X] I ran `brew update-reset && brew update` and retried my command.
- [X] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [X] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [X] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

#### Description of issue

{{
First noticed Aug 31st.Almost all commands return the error Error: Cask 'aquiline-check' definition is invalid: invalid 'depends_on macos' value: ">= 10.11". The exception is brew cask doctor, which does not return an error
}}

#### Command that failed

```
{{
brew update brew doctor
}}
```

#### Output of command with `--force --verbose --debug`

```
{{
$ brew upgrade --force --verbose --debug
Updating Homebrew...
/usr/local/Homebrew/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/pyenv.rb
/usr/local/Homebrew/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/lame.rb
/usr/local/Homebrew/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/pkg-config.rb
/usr/local/Homebrew/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/gtk+.rb
/usr/local/Homebrew/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/vim.rb
[...]

PR #90326, opened 2020-10-04
Fix doc about macOS version comparison

dlackty:
While implementing custom Cask, I found that the following statement doens't work and raises invalid cask errors:

```ruby
  depends_on macos: '>= 10.15'
```

However, this is declared valid in documentaiton.

After digging into [the source code of homebrew](https://github.com/Homebrew/brew/blob/5e09bb256411d8ed9388fa2cd64dcbb8be847bdf/Library/Homebrew/requirements/macos_requirement.rb#L18), I found that it uses `MacOS::Version.from_symbol` to parse the version, and thus confirm that using version code is not valid anymore.

I also found #88467 reported this issue.

### DIFF
--- a/doc/cask_language_reference/stanzas/depends_on.md
+++ b/doc/cask_language_reference/stanzas/depends_on.md
@@ -51,11 +51,10 @@ depends_on macos: ['10.14', '10.15']
 
 ### Setting a Minimum macOS Release
 
-`depends_on macos:` can also accept a string starting with a comparison operator such as `>=`, followed by an macOS release in the form above. The following are both valid expressions meaning “at least macOS 10.15”:
+`depends_on macos:` can also accept a string starting with a comparison operator such as `>=`, followed by an macOS release in the form above. The following is a valid expression meaning “at least macOS Catalina (10.15)”:
 
 ```ruby
 depends_on macos: '>= :catalina'
-depends_on macos: '>= 10.15'
 ```
 
 A comparison expression cannot be combined with any other form of `depends_on macos:`.
--- a/doc/cask_language_reference/stanzas/depends_on.md
+++ b/doc/cask_language_reference/stanzas/depends_on.md
@@ -27,11 +27,11 @@ depends_on formula: 'unar'
 
 ### Requiring an Exact macOS Release
 
-The value for `depends_on macos:` may be a symbol, string, or an array, listing the exact compatible macOS releases.
+The value for `depends_on macos:` may be a symbol or an array, listing the exact compatible macOS releases.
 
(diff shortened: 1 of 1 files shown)


miccal:
These do not work either:
```
depends_on macos: '10.15'
depends_on macos: ['10.14', '10.15']
```
or
```
depends_on macos: "10.15"
depends_on macos: ["10.14", "10.15"]
```
May be nest to remove all mention of `string` and stick to just `symbol` for `depends_on macos:`.

reitermarkus:
This should work again with https://github.com/Homebrew/brew/pull/8859.

core-code:
i also think sticking to a single style is best. better not mention a style which is used in zero caskfiles anyway.

btw there is a single caskfile that sticks out using the ">" operator

`  depends_on macos: "> :yosemite"
`

in `whichspace`

vitorgalvao:
> i also think sticking to a single style is best. better not mention a style which is used in zero caskfiles anyway.

It was decided early in the feature’s implementation that we’d always use the name style in the main repos, but using numbers would be available for convenience of third-parties. But I mention that only for context; I have no quarrel with removing the number style.

> btw there is a single caskfile that sticks out using the ">" operator
> 
> ` depends_on macos: "> :yosemite"`
> 
> in `whichspace`

Nice catch. Fixing in https://github.com/Homebrew/homebrew-cask/pull/90463

doc/cask_language_reference/stanzas/depends_on.md
```diff
@@ -51,11 +51,10 @@ depends_on macos: ['10.14', '10.15']
 
 ### Setting a Minimum macOS Release
 
-`depends_on macos:` can also accept a string starting with a comparison operator such as `>=`, followed by an macOS release in the form above. The following are both valid expressions meaning “at least macOS 10.15”:
+`depends_on macos:` can also accept a string starting with a comparison operator such as `>=`, followed by an macOS release in the form above. The following is a valid expression meaning “at least macOS 10.15”:
```
reitermarkus, line 54:
```suggestion
`depends_on macos:` can also accept a string starting with a comparison operator such as `>=`, followed by an macOS release in the form above. The following is a valid expression meaning “at least macOS Catalina (10.15)”:
```

dlackty: Nice suggestion. Just updated.

miccal:
I have added a few more updates.

reitermarkus (approved):

vitorgalvao (approved):
Next maintainer to agree with the changes feel free to remove the label and merge.

PR #90349, opened 2020-10-04
Update karabiner-elements to version 13

core-code:


miccal:
This updated version `13.0.0` brings along with it some substantial changes, in particular for `uninstall`.

There is an extra `pkg` receipt `org.pqrs.Karabiner-DriverKit-VirtualHIDDevice` and another `uninstall script` at `/Library/Application Support/org.pqrs/Karabiner-DriverKit-VirtualHIDDevice/scripts/uninstall/remove_files.sh`. 

Perhaps it may be best to move the `<= :el_capitan` and `<= :mojave` versions to `homebrew-cask-versions` and just have the newest one in the main repo.

core-code:
is everyone on catalina already? given how awful catalina is, i am still on mojave.

would it hurt to just add the additional uninstalls unconditionally?

core-code:
> here is an extra pkg receipt org.pqrs.Karabiner-DriverKit-VirtualHIDDevice 

should we even remove that given this comment?

`  # kext: 'org.pqrs.driver.Karabiner.VirtualHIDDevice.v*': Should not be uninstalled by Cask
`

miccal:
I have tested the `install`, `uninstall` and `zap` on my 10.14 and 10.15 machines.

The 10.14 version is all good and there is no longer any `kext` installed, just a `kextd` which is successfully handled by `launchctl` via `org.pqrs.karabiner.karabiner_kextd` and the `uninstall script`.

For the 10.15 version, in preparation for macOS Bug Sur the developers install a driver extension (`dext`) in `/Library/DriverExtensions/org.pqrs.Karabiner-DriverKit-VirtualHIDDevice.dext`, which is a symlink to `/Library/SystemExtensions/{a random string}/org.pqrs.Karabiner-DriverKit-VirtualHIDDevice.dext`, which (I think) cannot be removed without disabling SIP (at this stage).

There is also a left-over ad-hoc `launchctl` job, namely `org.pqrs.Karabiner-DriverKit-VirtualHIDDevice-({a random string})`, which can be removed by a script located at:
```
/Library/Application Support/org.pqrs/Karabiner-DriverKit-VirtualHIDDevice/scripts/uninstall/deactivate_driver.sh
```
&nbsp;
The contents of this `deactivate_driver.sh` script are as follows:
```
#!/bin/bash

#
# Note: This script must be run with console user account.
#

PATH=/bin:/sbin:/usr/bin:/usr/sbin
export PATH

osascript -e 'display dialog "Removing Karabiner VirtualHIDDevice driver.\nThe administrator password will be required to complete." buttons {"OK"}'
/Applications/.Karabiner-VirtualHIDDevice-Manager.app/Contents/MacOS/Karabiner-VirtualHIDDevice-Manager deactivate
```
&nbsp;
Notice it calls a GUI, and my attempts to run `/Applications/.Karabiner-VirtualHIDDevice-Manager.app/Contents/MacOS/Karabiner-VirtualHIDDevice-Manager deactivate` manually were not successful.

core-code:
thanks for testing. seems difficult after all. let me know if we should stick to your initial plan and remove support for older versions to make this easier.

miccal:
@core-code it is done now and it passes CI, so might as well leave it as-is. The older versions are fine, it is just the newer one that is a bit quirky in preparation for the next macOS release.

core-code:
thanks!

miccal:
Note: I kept the two `uninstall`'s deliberately separated to make it easier for if/when the older versions are dropped/migrated to `homebrew-versions`.

vitorgalvao (approved):
Fantastic work. Despite being a complex installation, it’s thorough and well organised.

miccal:
Thank you @vitorgalvao.

PR #90351, opened 2020-10-04
Update ncar-ncl.rb

core-code:
== is not something that is normally used in caskfiles

### DIFF
--- a/Casks/ncar-ncl.rb
+++ b/Casks/ncar-ncl.rb
@@ -2,10 +2,10 @@ cask "ncar-ncl" do
   version "6.6.2"
 
   if MacOS.version <= :high_sierra
-    sha256 "b107934b17085c39053467aa3faa07f00b4e18a89271a0b15eec6768d6ab06fb"
+    sha256 "4e937a6de4303a4928f0f42390d991b12a37659726d15b9da7e8072db74e1867"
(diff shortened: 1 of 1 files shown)


Casks/ncar-ncl.rb
```diff
@@ -1,22 +1,23 @@
...
   end
 
   appcast "https://www.ncl.ucar.edu/current_release.shtml"
   name "NCAR Command Language"
   name "ncl"
+  desc "Interpreted language for scientific data analysis and visualization"
   homepage "https://www.ncl.ucar.edu/"
 
-  depends_on x11: true
+  depends_on cask: "xquartz"
```
vitorgalvao, line 18: @miccal Why the change? Was it failing before? These are not equivalent; `depends_on x11` is meant to work even if the user hasn’t installed through a cask.

miccal: > Was it failing before?

Yes:
```
Error: Cask 'ncar-ncl' requires XQuartz/X11, which can be installed using Homebrew Cask by running:
  brew cask install xquartz
or manually, by downloading the package from:
  https://www.xquartz.org/
Error: Process completed with exit code 1.
```
which come from [here](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/exceptions.rb#L107-L120).

I made the same change in https://github.com/Homebrew/homebrew-cask/pull/90345

I can revert these changes, but then a `ci-skip-install` label may be required to pass CI.

miccal: Thought about it, and I will revert this.

miccal (approved):

PR #90356, opened 2020-10-04
Update boom to 1.6.11

core-code:
ok we have a bit of a problem here. they don't use sparkle anymore, but a homegrown solution. which is difficult to piggyback on even with sophisticated CGIs. basically they do this:

`curl -H 'Host: www.globaldelight.net' -H 'Content-Type: application/octet-stream' -H 'Accept: */*' -H 'User-Agent: Boom%202/1.6.9003 CFNetwork/1200.1 Darwin/20.1.0' -H 'Accept-Language: en-gb' --data-binary "@data.txt" --compressed 'https://www.globaldelight.net/tower/registrationauth/boom/boomCheckUpdateOffer.php'`

where you need a 500 byte binary file. 

=> current output:
`{"ResponseStatus":"1","AppVersionAvailable":"1.6.11","DownloadURL":"https:\/\/d3jbf8nvvpx3fh.cloudfront.net\/Boom2\/Update\/v1.6.11\/Boom_2_Update.dmg","ReleaseNotes":"Boom 2 (v1.6.11) comes with: \n- Minor defect fixes."}%   
`
so the other option is to use and try to observe the static download file from the homepage. lets try that

### DIFF
--- a/Casks/boom.rb
+++ b/Casks/boom.rb
@@ -1,10 +1,10 @@
 cask "boom" do
-  version "1.6.9,1575451705"
-  sha256 "444b5513c92eb0975494509908786a31f087a0af0e58fa5f312a156318be22f8"
+  version "1.6.11"
+  sha256 "624f8f01299d657e677098acfe4df6291136da5cff9949218168f4aced9a3be3"
 
-  # devmate.com/com.globaldelight.Boom2/ was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


core-code:
data file fyi

[data.txt](https://github.com/Homebrew/homebrew-cask/files/5324168/data.txt)

core-code:
ping @suschizu

core-code:
not sure why CI ignores the label

suschizu (approved):

PR #90364, opened 2020-10-05
Delete archi.rb

vitorgalvao:
Ping @Homebrew/cask. This cask has been removed and readded before (merged by me because I didn’t recognise it that time) but it should be kept out. They keep changing the `after_comma` value without any other changes, as some kind of download protection we have to keep fighting and isn’t worth it.

I have a feeling the `- [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).` section of the template isn’t effective. I can understand why, as after clicking that link it isn’t obvious you should search for the cask being submitted. Having a separate document that lists refusals and their reasonings should work better, but open to other ideas.


core-code:
perhaps somehow keep a list of casks that have been deleted?

core-code (approved):

PR #90531, opened 2020-10-08
Expose the usage of MiniZinc executables from MiniZincIDE Cask

Dekker1:

- [X] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.

### Proposed Changes

This PR exposes the MiniZinc executables from the bundled package. This allows users to use `minizinc` directly from command-line.

Although the binary would currently conflict with the current `minizinc` formula in the Homebrew core packages., it should be mentioned that this binary is an incomplete version of MiniZinc. The official MiniZinc bundle links against the CoinBC and Gecode solvers and is compiled with the headers for the CPlex, Gurobi, SCIP, and XPress solvers to allow users to load these solvers at runtime. This means that users of the `minizinc` formula do not have access to various features of the MiniZinc compiler, both during compilation of the MiniZinc language and in the solving of the produced models. Therefore, I would suggest that once this cask can expose the executables from within this cask, we have a look at deprecating the `minizinc` formula and aliasing `minizinc` to this cask.

### DIFF
--- a/Casks/minizincide.rb
+++ b/Casks/minizincide.rb
@@ -6,7 +6,10 @@ cask "minizincide" do
   url "https://github.com/MiniZinc/MiniZincIDE/releases/download/#{version}/MiniZincIDE-#{version}-bundled.dmg"
   appcast "https://github.com/MiniZinc/MiniZincIDE/releases.atom"
   name "MiniZincIDE"
-  homepage "https://www.minizinc.org/ide/index.html"
+  desc "Open-source constraint modeling language and IDE"
+  homepage "https://www.minizinc.org/index.html"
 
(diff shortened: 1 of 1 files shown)


miccal:
Justification for `binary`: https://www.minizinc.org/doc-2.5.0/en/basic_steps.html#the-minizinc-command-line-tool

miccal:
> I assume this should have been the "minizinc" formula that it conflicts with. My apologies for not adding this. I'm a bit new to creating Cask definitions

Whoops, thanks!

miccal:
Thank you @Dekker1.

miccal (approved):

PR #90533, opened 2020-10-08
Add MGTA Pro Tracker v2.1.10

lucassaldanha:

### DIFF
--- a/Casks/mtgaprotracker.rb
+++ b/Casks/mtgaprotracker.rb
@@ -1,9 +1,9 @@
 cask "mtgaprotracker" do
-  version "v2.1.10"
+  version "2.1.10"
   sha256 "fb620da85fd36c76eaaa460d143b59d61f6967c00e1680c84911d34c08d7db99"
 
   # github.com/Razviar/mtgap was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


lucassaldanha:
@miccal thanks for the update. I completely forgot to parameterize the version 👍 

Maybe it would be worth adding an extra check on the audit command to prevent this kind of mistake? Something like: if there is a version in the Cask file, but no wildcard for the version is being used, generate an alert? Thoughts?

miccal:
> @miccal thanks for the update. I completely forgot to parameterize the version 👍

No problem.

> Maybe it would be worth adding an extra check on the audit command to prevent this kind of mistake? Something like: if there is a version in the Cask file, but no wildcard for the version is being used, generate an alert? Thoughts?

That would be handy, yes!

vitorgalvao:
> Something like: if there is a version in the Cask file, but no wildcard for the version is being used, generate an alert? Thoughts?

Would lead to false negatives. We allow versioned casks that don’t use `#{version}` anywhere as long as they have an `appcast`. This isn’t a common problem, so the extra (imperfect) check isn’t worth it.

vitorgalvao (approved):

PR #90540, opened 2020-10-08
Update evernote to v10

core-code:


core-code:
things would be easier if 'brew cask style' worked on Big Sur

core-code:
@miccal @vitorgalvao should we drop the old OS versions to simplify?

vitorgalvao:
> should we drop the old OS versions to simplify?

Fine with me. Anything below High Sierra isn’t officially supported by Homebrew in general anyway.

core-code:
well it works now so i don't really see a reason to remove something that could be useful to someone

vitorgalvao (approved):

Issue #90629, opened 2020-10-09
“NBTExplorer” needs to be updated.


Masononic:
The developer of this app needs to update it to work with this version of macOS.

miccal:
This is an issue for the developer of `nbtexplorer`, not `homebrew-cask` -- see https://github.com/jaquadro/NBTExplorer/issues/83.

This Cask may need to be removed.

PR #90643, opened 2020-10-10
nbtexplorer: remove

miccal:
See https://github.com/Homebrew/homebrew-cask/issues/90629.


core-code:
do we have a rule to remove casks that do not on the newest macOS? 

looking for "depends_on macos: "[^>]" reveals at least 18 casks that do not run on 10.15. maybe add a depends_on here too as it may still be useful to users on old macOS versions?

PR #91545, opened 2020-10-28
Delete cloudflare-warp

miccal:
`homepage` and `url` appear to be dead.


reitermarkus:
Both are up for me.

miccal:
Thanks again for confirming @reitermarkus.

PR #91578, opened 2020-10-28
Export Wireshark CLI tools from formula

luizberti:
Wireshark comes with several bundled CLI tools, which currently are not
exposed through the cask, and people have to manually alias or link them.

This makes the Wireshark cask work more in line with what other casks
that bundle CLI binaries (e.g. `julia` or `visual-studio-code`) do.

The following command was used to extract the available binaries from
`Wireguard.app` if needed for future reference (needs `fd` and `rg`):

```bash
fd --type=executable . /Applications/Wireshark.app/ | \
rg -F Wireshark.app/Contents/MacOS | rg -v 'Wireshark$' | \
rg -F /Applications --replace='binary "#{appdir}' | awk '{print $0 "\""}'
```

The outputs of it are in line with the documentation currently hosted at:
https://www.wireshark.org/docs/man-pages/

A great follow up addition to this would be adding the `man` pages (e.g.
`/Applications/Wireshark.app/Contents/Resources/share/man/man1/tshark.1`)
automatically to `man`'s path through the Cask formula, but I am not sure
how to accomplish this **(I'll gladly make another PR  for this if a maintainer
can point me to an example from another formula that does this)**.

### DIFF
--- a/Casks/wireshark.rb
+++ b/Casks/wireshark.rb
@@ -14,7 +14,26 @@ cask "wireshark" do
 
   app "Wireshark.app"
   pkg "Install ChmodBPF.pkg"
-  pkg "Add Wireshark to the system path.pkg"
+  pkg "Add Wireshark to the system path.pkg"  
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/capinfos"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/captype"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/dftest"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/dumpcap"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/editcap"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/idl2wrs"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/mergecap"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/mmdbresolve"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/randpkt"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/rawshark"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/reordercap"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/sharkd"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/text2pcap"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/tshark"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/extcap/androiddump"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/extcap/ciscodump"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/extcap/randpktdump"
+  binary "#{appdir}/Wireshark.app/Contents/MacOS/extcap/sshdump"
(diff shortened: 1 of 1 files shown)


luizberti:
**For the reviewers:** I fixed some style errors that were already present by running the `brew cask style --fix` command, except for the `Missing frozen string literal comment` one (cause I'm not exactly sure what that's about, since I'm not a rubyist)

The only style violation that was actually introduced by my PR was the `Block has too many lines` one, but I am not exactly sure what to do about it given that all those lines are necessary, so some guidance would be appreciated here.

miccal:
Thank you @luizberti, but I believe the CLI tools can be installed via the [`wireshark` formula](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/wireshark.rb). Note that the `caveats` for the formula state that:
```
==> Caveats
This formula only installs the command-line utilities by default.

Install Wireshark.app with Homebrew Cask:
  brew cask install wireshark

If your list of available capture interfaces is empty
(default macOS behavior), install ChmodBPF:
  brew cask install wireshark-chmodbpf
```
&nbsp;
@vitorgalvao I would suggest that we add a similar `caveat` message to the cask version to match the formula.

luizberti:
@miccal @vitorgalvao Fair enough, I removed the linkage statements in favor of a caveat message to avoid confusion, can you take a look at it again?

luizberti:
Actually, scratch what I just said. I just tried installing the `wireshark` formula, and it skips linking the binaries if the cask is installed, because it assumes the cask does the linking, and indeed none of the CLI tools (e.g. `tshark` and `randpkt`) are available after installing the formula. See the output of `brew install wireshark`:
```
...
==> Installing wireshark dependency: nghttp2
==> Pouring nghttp2-1.41.0_1.catalina.bottle.tar.gz
🍺  /usr/local/Cellar/nghttp2/1.41.0_1: 26 files, 2.7MB
==> Installing wireshark
==> Pouring wireshark-3.2.7.catalina.bottle.tar.gz
==> wireshark cask is installed, skipping link.                  <<========= HERES THE RELEVANT PART
==> Caveats
This formula only installs the command-line utilities by default.

Install Wireshark.app with Homebrew Cask:
  brew cask install wireshark

If your list of available capture interfaces is empty
(default macOS behavior), install ChmodBPF:
  brew cask install wireshark-chmodbpf
==> Summary
🍺  /usr/local/Cellar/wireshark/3.2.7: 1,168 files, 91.3MB
==> Caveats
==> wireshark
This formula only installs the command-line utilities by default.
...
```

I really do think linking from the cask is what makes the most sense, as it keeps the CLI tools and Wireshark.app versioned and updated together, and given that this seems to be the behavior the regular `wireshark` brew formula expects, I think I should revert my commit that removed the symlinking instructions and added the caveat.

Do the two of you agree? @miccal @vitorgalvao

vitorgalvao:
> I really do think linking from the cask is what makes the most sense, as it keeps the CLI tools and Wireshark.app versioned and updated together, and given that this seems to be the behavior the regular `wireshark` brew formula expects, I think I should revert my commit that removed the symlinking instructions and added the caveat.

Agreed. That’s consistent with other casks ([`mpv`](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/mpv.rb), [`mkvtoolnix`](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/mkvtoolnix.rb)).

luizberti:
@miccal @vitorgalvao the commit has been reverted back to linking all the binaries from the cask's app

anything else pending for a merge?

miccal:
Thank you @luizberti.

miccal (approved):

PR #91805, opened 2020-10-31
parallels: improve uninstall procedure

emilazy:
This allows the package to be cleanly uninstalled even while running, unloads the kexts, and removes a bunch of extra leftover files and other remnants I found.

Things I’m unsure about:
* I don’t know if all of these `uninstall` parameters are required or if some can be inferred automatically.
* I couldn’t get the files outside `$HOME` to be trashed, so they’re just `delete:`d currently.
* Should files in `/Library/StagedExtensions` be removed? Looks like it involves `kextcache(8)`; this is probably something that should be handled by `uninstall kext:` in general?
* The `pkill`/`sed` scripts could probably use review.


Casks/parallels.rb
```diff
@@ -30,29 +30,88 @@
...
+    "/Library/Logs/DiagnosticReports/prl_vm_app_*.diag",
+    "/Library/Logs/parallels.log",
+    "/Library/Preferences/Parallels",
+    "/Users/Shared/Parallels",
+    "/private/var/.Parallels_swap",
+    "/private/var/db/Parallels",
+    "/private/var/folders/qm/0fjbztdj28d5d9x2h411gxm00000gn/C/com.parallels.desktop.console",
+    "/private/var/folders/qm/0fjbztdj28d5d9x2h411gxm00000gn/C/com.parallels.vm",
+    "/private/var/folders/qm/0fjbztdj28d5d9x2h411gxm00000gn/C/com.parallels.winapp.*",
+    "/private/var/root/Library/Preferences/com.parallels.desktop.plist",
```
miccal, line 110: These files and folders in `/private` should not be removed by Cask.

Casks/parallels.rb
```diff
@@ -30,29 +30,88 @@
...
     "~/Library/Preferences/com.parallels.Parallels Desktop.plist",
     "~/Library/Preferences/com.parallels.Parallels.plist",
-    "~/Library/Preferences/com.parallels.PDInfo.plist",
+    "~/Library/Preferences/com.parallels.desktop.console.LSSharedFileList.plist",
+    "~/Library/Preferences/com.parallels.desktop.console.plist",
+    "~/Library/Saved Application State/com.parallels.desktop.console.savedState",
+  ], delete: [
+    "/Library/Logs/DiagnosticReports/prl_vm_app_*.diag",
+    "/Library/Logs/parallels.log",
+    "/Library/Preferences/Parallels",
```
miccal, line 103: These three `/Library` entries are not user-specified files, so they belong in `uninstall delete`.

Casks/parallels.rb
```diff
@@ -30,29 +30,88 @@
...
+    sudo:         true,
+    must_succeed: false,
+  }, quit: [
+    "com.parallels.vm",
+    "com.parallels.desktop.console",
+  ], kext: [
+    # com.parallels.kext.netbridge goes first, because it depends
+    # on com.parallels.kext.{hypervisor,vnic}
+    "com.parallels.kext.netbridge",
+
```
miccal, line 46: Remove this blank line.

Casks/parallels.rb
emilazy: Surely `uninstall` shouldn’t destructively erase logs and preferences? There seem to be a lot of casks with `/Library/{Logs,DiagnosticReports}` paths and the like in `zap` stanzas; e.g. (from a quick `rg '[^~]/Library/Logs'` and verifying the stanza) `vmware-fusion`, `firefox`, `keybase`, `little-snitch`.

Casks/parallels.rb
emilazy: As with https://github.com/Homebrew/homebrew-cask/pull/91805#discussion_r515525079, these are just leftover cache and preference files, so `zap`? Is the problem just that they're not owned by the user? The uninstall procedure for Parallels already requires `sudo`, and there seem to be a few casks removing `/var/folders` paths already.

Casks/parallels.rb
emilazy: Rebased to remove the blank line and also the `/Users/Shared/Parallels` zap (since it's the storage location for actual VMs, a la `~/Parallels`, which probably shouldn't get erased); I could still use some elaboration on the other issues, per my comments.

Casks/parallels.rb
```diff
@@ -36,77 +36,82 @@
...
+            ],
+            kext:         [
+              # com.parallels.kext.netbridge goes first, because it depends
+              # on com.parallels.kext.{hypervisor,vnic}
+              "com.parallels.kext.hypervisor",
+              "com.parallels.kext.netbridge",
+              "com.parallels.kext.usbconnect",
+              "com.parallels.kext.vnic",
+            ],
+            delete:       [
```
emilazy, line 52: The ordering here is sensitive; `com.parallels.kext.netbridge` has to be the first entry.

Casks/parallels.rb
```diff
@@ -30,29 +30,91 @@
...
+              "/usr/local/bin/prlsrvctl",
+              "/usr/local/etc/bash_completion.d/prlctl",
+              "/usr/local/etc/bash_completion.d/prlsrvctl",
+              "/usr/local/share/man/man8/prl_convert.8",
+              "/usr/local/share/man/man8/prl_disk_tool.8",
+              "/usr/local/share/man/man8/prlctl.8",
+              "/usr/local/share/man/man8/prlexec.8",
+              "/usr/local/share/man/man8/prlsrvctl.8",
+            ],
+            script:       {
```
vitorgalvao, line 69: `script` is for scripts, not commands. This should be in an `uninstall_postflight`.

Casks/parallels.rb
```diff
@@ -30,29 +30,91 @@
...
-    "/usr/local/bin/prl_disk_tool",
-    "/usr/local/bin/prl_perf_ctl",
-    "/usr/local/bin/prlcore2dmp",
-    "/usr/local/bin/prlctl",
-    "/usr/local/bin/prlexec",
-    "/usr/local/bin/prlsrvctl",
-  ]
-
-  zap trash: [
+  uninstall early_script: {
```
vitorgalvao, line 33: `script` is for scripts, not commands. This should be in an `uninstall_preflight`. Or better yet, if `quit` can’t do it maybe [`signal`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/uninstall.md#uninstall-key-signal) might.

vitorgalvao (changes requested):

Casks/parallels.rb
emilazy: Moved these to the `uninstall_{pre,post}flight` blocks; I'm not sure `signal` would suffice here since they're just random executables inside `/Applications/Parallels Desktop.app/Contents/MacOS` spawned outside of the normal app flow?

emilazy:
Not sure I understand [the CI failure](https://github.com/Homebrew/homebrew-cask/runs/1725252527?check_suite_focus=true); shouldn’t the main application bundle get removed automatically?

emilazy:
I see; it’s because the `pkill` is failing. I’m not sure why, though, as I used `system_command` rather than `system_command!`.

Casks/parallels.rb
vitorgalvao: > and there seem to be a few casks removing `/var/folders` paths already.

We might want to rethink that as well. These should be deleted by the system.

Casks/parallels.rb
```diff
@@ -14,47 +14,105 @@
...
+    system_command "/usr/bin/pkill",
+                   args: ["-f", "#{appdir}/Parallels Desktop.app"],
+                   sudo: true
+  end
+
+  uninstall_postflight do
+    # "Parallels records in the /etc/hosts file have #prl_hostonly in the string."
+    #   -- https://kb.parallels.com/en/123300
+    system_command! "/usr/bin/sed",
+                    args: ["-i", ".orig", "/#prl_hostonly/d", "/etc/hosts"],
```
vitorgalvao, line 42: Good thinking to delete these, but at the same time you’re creating another file. `.orig` isn’t descriptive when someone bumps into it later on.

More explicit solution: Make the extension `.before_cask_uninstall_parallels`; move the file somewhere else (like the Trash); warn the user this is going on.

PR #91837, opened 2020-10-31
Using a 64-bit build of ZipCleaner

ItalyPaleAle:
ZipCleaner has been mostly [abandoned](https://roger-jolly.nl/software/) from its maintainer, and while the code still works, it's never been re-compiled for 64-bit. Because of that, it stopped working on Catalina.

This updates the link to ZipCleaner's builds to use the ones by @AmigaAbattoir: https://github.com/AmigaAbattoir/ZipCleaner/releases

(Note: I have no relationship with either ZipCleaner nor the maintainer of the repo above)

### DIFF
--- a/Casks/zipcleaner.rb
+++ b/Casks/zipcleaner.rb
@@ -3,6 +3,7 @@ cask "zipcleaner" do
   sha256 "bfdb7d451bc2b0e8f65711e09690dc4527ed924f0d187e4c7eacd0d824e24c58"
 
   url "https://github.com/AmigaAbattoir/ZipCleaner/releases/download/V#{version}/ZipCleaner.app.zip"
+  appcast "https://github.com/AmigaAbattoir/ZipCleaner/releases.atom"
   name "ZipCleaner"
   homepage "https://github.com/AmigaAbattoir/ZipCleaner"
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you and welcome.

This is a common mistake for first-time contributors, but **we can’t accept any random fork as a substitute for an existing cask; that’d be a huge security hole.**

[There are only two situation in which a fork can overturn the main cask](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#forks-and-apps-with-conflicting-names), and this satisfies neither. It would also not be accepted as a fork cask anyway because the repository is unpopular.

ItalyPaleAle:
So what are the options? The app is not working anymore, should it just be removed from Homebrew?

vitorgalvao:
> So what are the options?

Leave it be for now and [make sure it doesn’t try to be installed under Catalina and later](https://github.com/Homebrew/homebrew-cask/pull/91879).

> should it just be removed from Homebrew?

And eventually remove it.

> The app is not working anymore

For you. Plenty of people aren’t on Catalina and never will be. The app is unpopular but it’s not problematic right now.

```
zipcleaner
30 days: 6 (#3095)
90 days: 13 (#3622)
365 days: 52 (#4088)
Age: 2550 days (added 2013, November 09)
```

Issue #91744, opened 2020-10-30
google-cloud-sdk broken


xaocon:
### Description of issue

Not sure if this is the right place for this but the google-cloud-sdk is broken under default situations. It broke when after installing python@3.9 and is I tried reinstalling to fix and now it can't be installed. I'm pretty sure this is because of python@3.9 being the default and linked python version and google-cloud-sdk not supporting it. I imagine it's fixable by having a dependency on 3.8 and setting up the install to use it. Not sure about the inner workings of brew and if you can have a formulae dep in a cask?

#### Command that failed

```
brew cask install google-cloud-sdk
```

#### Output of command with `--verbose --debug`

```
==> Cask::Installer#install
==> Printing caveats
==> Caveats
google-cloud-sdk is installed at /usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk. Add your profile:

  for bash users
    source "/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/path.bash.inc"
    source "/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/completion.bash.inc"

  for zsh users
    source "/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/path.zsh.inc"
    source "/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/completion.zsh.inc"

  for fish users
    set -g fish_user_paths "/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/path.fish.inc" $fish_user_paths

==> Cask::Installer#fetch
==> Downloading
==> Downloading https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz
/usr/bin/curl --disable --globoff --show-error --user-agent Homebrew/2.5.7\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 10.15.7\)\ curl/7.64.1 --header Accept-Language:\ en --retry 3 --location --silent --head --request GET https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz
Already downloaded: /Users/epitstick/Library/Caches/Homebrew/downloads/cda39f18c3069c1a4ac0cd5b1d91541af945cd331eba8443c58ec5c890343c0a--google-cloud-sdk.tar.gz
==> Checking quarantine support
/usr/bin/xattr
[...]

PR #91856, opened 2020-11-01
Updated google-cloud-sdk for python39

xaocon:
Issue #91744

google-cloud-sdk fails to install or work because it doesn't work with
python 3.9 yet and that is the default python3 in homebrew. Please
review this PR closely as I'm new to the inner workings of homebrew/casks 
 and I don't know Ruby or fish shell. These changes may also be a
bit "hacky" but were the best I could come up with under the
circumstances.

These changes require and force usage of python 3.8 for installation
and adds information to the `caveats` section about setting the required
`CLOUDSDK_PYTHON` shell variable.

I removed all the `binary` stanzas as the `caveats` section mentions
that you must source files for your shell that set $PATH to look for the
bins in their unlinked locations.

Additionally, I cleaned up the cask a bit and added a `desc` stanza.

### DIFF
--- a/Casks/google-cloud-sdk.rb
+++ b/Casks/google-cloud-sdk.rb
@@ -28,12 +28,12 @@ cask "google-cloud-sdk" do
     #{token} is installed at #{staged_path}/#{token}. Add your profile:
 
       for bash users
-        export "CLOUDSDK_PYTHON=#{HOMEBREW_PREFIX}/opt/python@3.8/libexec/bin/python"
+        export CLOUDSDK_PYTHON="#{HOMEBREW_PREFIX}/opt/python@3.8/libexec/bin/python"
         source "#{staged_path}/#{token}/path.bash.inc"
         source "#{staged_path}/#{token}/completion.bash.inc"
 
       for zsh users
-        export "CLOUDSDK_PYTHON=#{HOMEBREW_PREFIX}/opt/python@3.8/libexec/bin/python"
+        export CLOUDSDK_PYTHON="#{HOMEBREW_PREFIX}/opt/python@3.8/libexec/bin/python"
         source "#{staged_path}/#{token}/path.zsh.inc"
         source "#{staged_path}/#{token}/completion.zsh.inc"
 
--- a/Casks/google-cloud-sdk.rb
+++ b/Casks/google-cloud-sdk.rb
@@ -9,15 +9,16 @@ cask "google-cloud-sdk" do
 
   depends_on formula: "python@3.8"
 
-  installer script: {
-    executable: "/usr/bin/env",
-    args:       [
-      "CLOUDSDK_PYTHON=#{HOMEBREW_PREFIX}/opt/python@3.8/libexec/bin/python",
-      "#{caskroom_path}/latest/google-cloud-sdk/install.sh",
-      "--usage-reporting", "false", "--bash-completion", "false", "--path-update", "false",
-      "--rc-path", "false", "--quiet"
-    ],
-  }
+  stage_only true
+
+  postflight do
+    system_command "#{staged_path}/#{token}/install.sh",
+                   args: [
+                     "--usage-reporting", "false", "--bash-completion", "false", "--path-update", "false",
+                     "--rc-path", "false", "--quiet"
+                   ],
+                   env:  { "CLOUDSDK_PYTHON" => "#{HOMEBREW_PREFIX}/opt/python@3.8/libexec/bin/python" }
+  end
 
   # Not actually necessary, since it would be deleted anyway.
   # It is present to make clear an uninstall was not forgotten and that for this cask it is indeed this simple.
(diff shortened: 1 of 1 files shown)


Casks/google-cloud-sdk.rb
```diff
@@ -4,20 +4,20 @@
...
   homepage "https://cloud.google.com/sdk/"
 
+  depends_on formula: "python@3.8"
+
   installer script: {
-    executable: "google-cloud-sdk/install.sh",
+    executable: "/usr/bin/env",
     args:       [
+      "CLOUDSDK_PYTHON=#{HOMEBREW_PREFIX}/opt/python@3.8/libexec/bin/python",
+      "#{caskroom_path}/latest/google-cloud-sdk/install.sh",
```
miccal, line 16: Does this not work with just `google-cloud-sdk/install.sh`?

Casks/google-cloud-sdk.rb
```diff
@@ -4,20 +4,20 @@
...
       "--usage-reporting", "false", "--bash-completion", "false", "--path-update", "false",
       "--rc-path", "false", "--quiet"
     ],
   }
-  binary "google-cloud-sdk/bin/bq"
-  binary "google-cloud-sdk/bin/docker-credential-gcloud"
-  binary "google-cloud-sdk/bin/gcloud"
-  binary "google-cloud-sdk/bin/git-credential-gcloud.sh"
-  binary "google-cloud-sdk/bin/gsutil"
```
miccal, line 20: Are these all linked into `/usr/local` via the `install.sh`?

Casks/google-cloud-sdk.rb
xaocon: Not on my machine, which is "ready to brew". I'm up for any suggestions though.

Casks/google-cloud-sdk.rb
xaocon: They are. They aren't linked to /usr/local/bin but the files are in path if you follow the instructions of the installer or the caveats. Users that don't follow caveats won't be able to run them on a default system. If it's preferred to include them in /usr/local/bin anyways I can add them back. I just didn't see a good reason to link files to a location they weren't going to be called from anyways.

miccal: Ok, thanks, but running `gsutil` (for example) directly from the command line (once `CLOUDSDK_PYTHON` is set) works as expected, yes?

franklouwers:
Can confirm the Formula in this PR works.

Casks/google-cloud-sdk.rb
xaocon: No, you have to source the other files too. There is another line for each shell that is instructed to be added and it adds the path where the bin is. On mine for my shell it's `source "/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/path.zsh.inc"`. It's in the formula as it's stands today (not added in this PR). Running gcloud jor gsutil just works after following the instructions in Caveats in this PR and doesn't work at all without following the instructions in Caveats, even if you link the bins to /usr/local/bin.

xaocon: Another thing to consider is that this package has its own mini package manager that will download and install additional components. Those components will not be linked into /usr/local/bin because they aren't managed by homebrew. They will only be in the location added to `PATH` via the sourced file.

Casks/google-cloud-sdk.rb
```diff
@@ -4,20 +4,20 @@
...
   url "https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz"
   name "Google Cloud SDK"
+  desc "Set of tools to manage resources and applications hosted on Google Cloud"
   homepage "https://cloud.google.com/sdk/"
 
+  depends_on formula: "python@3.8"
+
   installer script: {
-    executable: "google-cloud-sdk/install.sh",
+    executable: "/usr/bin/env",
```
vitorgalvao, line 13: This is really weird. Can’t we make this work within `install_postflight` using `system_command`?

xaocon: This felt awkward to me too but it wasn't clear to me how to do it otherwise. I probably just don't know all the options as this is my first delve into homebrew inner workings. I'll see what I can come up with using that guidance.

Casks/google-cloud-sdk.rb
miccal: Thanks for the explanation @xaocon.

Casks/google-cloud-sdk.rb
xaocon: Yeah, that felt a lot more natural. Kind of surprised that system_command and installer executable have different options but that seems to be the easier to use.

vitorgalvao (approved):

dpacaud:
it seems this PR introduced a sha256 bug : 
```Error: Cask 'google-cloud-sdk' does not have a sha256 checksum defined and was not installed.
This means you have the --require-sha option set, perhaps in your HOMEBREW_CASK_OPTS.```

vitorgalvao:
@dpacaud That’s not a bug. The message tells you *exactly* what’s the issue; you have to fix on your side.

dpacaud:
oops sorry for this, will look into it.
thanks

Casks/google-cloud-sdk.rb
jmo-qap: This breaks automated installs that relied on the old approach

PR #91993, opened 2020-11-04
Update gotomeeting.rb from 10.12.0 to 10.13.1

iMonZ:
Updated to 10.13.1

### DIFF
--- a/Casks/gotomeeting.rb
+++ b/Casks/gotomeeting.rb
@@ -1,6 +1,6 @@
 cask "gotomeeting" do
-  version "10.12.0,18425"
-  sha256 "ac6ecdd0f45f50bbae3ff2d4b1e02ba404af28027f76079c8d0e67a3484439d9"
+  version "10.14.0,18962"
+  sha256 "ffcc6a3000eea1841bd79ba67f59036ed8392097f58a2bd438f37239c2c2e7df"
 
   # builds.cdn.getgo.com/builds/g2m/ was verified as official when first introduced to the cask
   url "https://builds.cdn.getgo.com/builds/g2m/#{version.after_comma}/GoToMeeting.dmg"
--- a/Casks/gotomeeting.rb
+++ b/Casks/gotomeeting.rb
@@ -1,6 +1,6 @@
 cask "gotomeeting" do
-  version "10.14.0,18962"
-  sha256 "ffcc6a3000eea1841bd79ba67f59036ed8392097f58a2bd438f37239c2c2e7df"
+  version "10.13.1,18705"
+  sha256 "e8f62a7715e6e28f3973b653d3ac8a771d6dc2183311234e8e47ace5e56f4b17"
 
   # builds.cdn.getgo.com/builds/g2m/ was verified as official when first introduced to the cask
   url "https://builds.cdn.getgo.com/builds/g2m/#{version.after_comma}/GoToMeeting.dmg"


iMonZ:
The change log is outdated
the last two updates are missing...
LogMein just doesn't like change logs

vitorgalvao:
If I go to the website and [the download page](https://support.goto.com/meeting), I still get on the latest URL (https://link.gotomeeting.com/latest-dmg) `10.13.1.18705`.

iMonZ:
And if you try to download this from the link inside the cask?
Because this is not the latest version!

vitorgalvao:
> And if you try to download this from the link inside the cask?

The link inside the cask is relative to the version, so that doesn’t help. I believe you that the version you’ve submitted is the latest, but is it the latest *stable* version? Because that’s what matters.

iMonZ:
Yes this is the latest stable version!
If you curious you can install this application and press on update!

vitorgalvao:
> If you curious

Not in the slightest. I dislike this company and their practices. Yet, I still need to verify it. There’s an amount of due diligence that has to be taken, and that you also benefit from.

> you can install this application and press on update!

I see no way to update on the installed application.

iMonZ:
> > If you curious
> 
> Not in the slightest. I dislike this company and their practices. Yet, I still need to verify it. There’s an amount of due diligence that has to be done, and that you also benefit from.
> 
> > you can install this application and press on update!
> 
> I see no way to update on the installed application.

That’s right this application is horrible from the technical view still I’m forced to use it :(

You can update this tool from the menu bar

vitorgalvao:
> You can update this tool from the menu bar

That’s where I’m looking. Can you be more specific or show a screenshot?

iMonZ:
Oh the update button is gone 
https://ibb.co/TYCwS2N

vitorgalvao:
> Then you get an error message that your version is too old to join the meeting with an update message.

Alright, that worked. *But* I also know from past research on other GoTo apps that they want people on the exact same versions of interacting software. GoToAssist will even upgrade or downgrade itself for version parity. So that isn’t conclusive evidence that this is the stable version. We need to rely on *something* from upstream. So we have two choices:

1. Change this cask to `version :latest`, `sha256 :no_check`, and `url "https://link.gotomeeting.com/latest-dmg"`.
2. Leave as is and upgrade to `10.13.1,18705`.

**2** seems more sensible to me. We rely on upstream and don’t lose `appcast` functionality.

Ping @Homebrew/cask.

iMonZ:
> > Then you get an error message that your version is too old to join the meeting with an update message.
> 
> Alright, that worked. _But_ I also know from past research on other GoTo apps that they want people on the exact same versions of interacting software. GoToAssist will even upgrade or downgrade itself for version parity. So that isn’t conclusive evidence that this is the stable version. We need to rely on _something_ from upstream. So we have two choices:
> 
> 1. Change this cask to `version :latest`, `sha256 :no_check`, and `url "https://link.gotomeeting.com/latest-dmg"`.
> 2. Leave as is and upgrade to `10.13.1,18705` (#92006).
> 
> **2** seems more sensible to me. We rely on upstream and don’t lose `appcast` functionality.
> 
> Ping @Homebrew/cask.

If there are only two choices I would recommend the first one, cause 10.14.0,18962 is definitely the latest stable version.

vitorgalvao:
> cause 10.14.0,18962 is definitely the latest stable version.

No, not _definitely_. Not according to upstream’s website. If you’re sure, we need proof. As stated, the software wanting to update itself when joining a specific meeting is not enough, due to the way they do things.

core-code:
personally i'd just use 'version :latest' in this case. we've even ignored the app from MacUpdater because tracking version numbers for this just doesn't make sense

iMonZ:
I will contact the LogMein support to ask them which version is the latest and if they can update the change log on there website.

iMonZ:
Okay I have contacted the support and they told me that the 10.14 is not out yet and this is an bug!
I will change my commit to 10.13.1

vitorgalvao:
> Okay I have contacted the support and they told me that the 10.14 is not out yet and this is an bug!

Thank you for checking! Closing in favour of https://github.com/Homebrew/homebrew-cask/pull/92006 which also has other fixes.

Still open to changing this to `version :latest` in the future if it’s justified.

PR #92006, opened 2020-11-04
gotomeeting.rb: update to 10.13.1,18705

vitorgalvao:
Refs https://github.com/Homebrew/homebrew-cask/pull/91993#issuecomment-721808117.

@core-code Since both work, which `appcast` do you prefer in this case?


core-code:
looks fine to me this way, thanks!

PR #91915, opened 2020-11-02
Update little-snitch from 4.6 to 5.0.1

core-code:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Created with [`cask-repair`](https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask).

PR #92106, opened 2020-11-06
Update bartender

tiiiecherle:


tiiiecherle:
Thanks a lot ;)

miccal (approved):

miccal:
Thank you @tiiiecherle.

core-code:
bartender 4 is a beta version and clearly labeled as such. even if it is the only version that runs on 'big sur' i think we should always have a discussion first if we make exceptions to our rules.

miccal:
[I assumed it followed this exception.](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version)

core-code:
i was under the assumption this talks about: "there is no stable version for this app at all". whether this applies in this case is debatable.

miccal:
From https://www.macbartender.com/Bartender4/:

> Bartender 4 is our biggest update yet, specifically created for macOS Big Sur. 
Bartender 4 is currently in Public Beta as more exciting new features are added.
It is free to use whilst in Beta. It will then be a paid upgrade.

and therefore satisfies our guidelines:

and adding this version has not fundamentally changed how this Cask would be installed for macOS version Catalina or lower (unlike the changes required for [`little-snitch`](https://github.com/Homebrew/homebrew-cask/pull/91915) which _would_ fundamentally change the install procedure).

So please explain exactly what "exceptions to our rules" you think have been made here that are "debatable"?

core-code:
> and therefore satisfies our guidelines:

it would only do so if the word 'only' was omitted in the sentence from the documentation which you've quoted. 

> has not fundamentally changed how this Cask would be installed for macOS version Catalina or lower 

which would be why it is ok to make an exception here. still i think its preferable to discuss when and if to make exceptions.

> little-snitch 

the difference is that little-snitch 5 is not in beta, it it official. but i agree it makes sense to delay merging that one.

> So please explain exactly what "exceptions to our rules" you think have been made here that are "debatable"?

we never carry beta software, except when there is *only* a beta version available. this is the sentence you've quoted above. but this doesn't apply here because there is a stable version of bartender, called bartender 3. 

we can make an exception here because of  the fact that on Big Sur there is no stable version. 

i don't think this is meant by the existing rules however, hence i think we are making an exception here.

PR #91944, opened 2020-11-03
Update aliwangwang from 20180413-1510-8.00.44 to 20200723-8.01.03

ElfSundae:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Created with [`cask-repair`](https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask).

PR #92111, opened 2020-11-06
Delete aliwangwang.rb

vitorgalvao:
`url` looks to be very outdated (over two years, see https://github.com/Homebrew/homebrew-cask/pull/91944) and there seems to be no way to get the current version without placing a lot of trust in URLs neither we nor users can verify without an unreasonable amount of effort (https://github.com/Homebrew/homebrew-cask/pull/91944#issuecomment-723195002).

Suggesting removal.


core-code:
we have a lot of appcasts that can only be determined with complicated checks (find-appcasts simplifies them but still) or have been determined by network sniffing. also we often have URLs that are taken from these appcasts, instead of the homepage.

why is it a problem in this case but OK in other cases?

vitorgalvao:
> we have a lot of appcasts that can only be determined with complicated checks (find-appcasts simplifies them but still) or have been determined by network sniffing. also we often have URLs that are taken from these appcasts, instead of the homepage.

In those cases, there’s *something* we can use as reasonable assumption they are legitimate. Examples: the `appcast` itself is from the same host as `homepage`, so even if `url` isn’t a user can verify it. Or the `appcast` isn’t from the same host as `homepage` but `url` is, so it’s presumed safe.

In those cases, even if network sniffing was required *the first time*, any user can audit the cask later and verify the URLs as legitimate without having to install the software.

> why is it a problem in this case but OK in other cases?

In this case, no URL (`appcast` or `url`) can be reasonably traced back to its source without considerable effort, beyond the reach of most users. To verify this, a user would have to install the software and perform the network sniffing themselves. To boot, the software is in a language we don’t read, making it hard for even maintainers to vouch for.

In sum, this cask (not in its current form, but in how it would need to be updated) takes a lot of work and responsibility to vouch for—more than I find reasonable—especially considering its numbers (high enough to cause damage, low enough to go undetected).

```
aliwangwang
30 days: 62 (#909)
90 days: 195 (#867)
365 days: 1,120 (#1057)
Age: 2499 days (added 2014, January 03)
```

core-code:
i'm not sure we don't have other cases where neither the url nor the appcast can be 'reasonably traced back' by an average user.

> takes a lot of work and responsibility to vouch for—more than I find reasonable

ok then lets remove it

miccal (approved):

levinit:
I am a chinese user, aliwangwang  has not been updated for two years, Many Chinese Internet companies do not pay much attention to the development of desktop applications, these companies always like to make mobile apps and force us to use them！！！

JounQin:
Can we add this cask back? It seems it is maintained again.

See also https://github.com/homebrew-x/homebrew-x/blob/master/Casks/aliwangwang.rb

vitorgalvao:
The main issue wasn’t the software’s maintenance status, it was how hard it is to audit for regular users. That’s explained in the link above. If that has changed, we can reconsider the cask’s inclusion.

JounQin:
@vitorgalvao There are two domains used `alicdn.com` as resource link and `tmall.com` as appcast and homepage.

vitorgalvao:
Which is precisely the issue discussed above. Unless they’ve changed something to make it clear the download URL belongs to that homepage, nothing changed.

JounQin:
@vitorgalvao It does. I just copied the download link from the homepage.

vitorgalvao:
In that case, feel free to submit a PR for review.

PR #92405, opened 2020-11-11
Use BetterTouchTool vendors appcast, not a third-party one

tomtastic:
Problem : 
www.corecode.io/macupdater is a paid app, and frequently carries incorrect appcast information in it's feed (https://www.corecode.io/macupdater/appcasts), we shouldn't rely on this, and it was incorrect at the time of committing this change.

Solution :
The vendors appcast should be used : https://updates.folivora.ai/appcast_manual.xml

Reasoning :
BetterTouchTool developer discussion here : https://community.folivora.ai/t/appcast-wrong-current-version/19605/3


vitorgalvao:
> Problem :
> www.corecode.io/macupdater is a paid app

That’s neither a problem nor related to the situation. Quite the contrary, HBC is in way better shape and cask updates happen faster because of MacUpdater. You need only search for contributions by @core-code and @suschizu (who are HBC maintainers!) to see what I mean.

> and frequently carries incorrect appcast information in it's feed (https://www.corecode.io/macupdater/appcasts), we shouldn't rely on this, and it was incorrect at the time of committing this change.

That specific appcast is unique in its solution. See https://github.com/Homebrew/homebrew-cask/pull/75004 as to why and how it was implemented. While imperfect—and we knew it would be because it’s dependent on when and how often [its creation script is run](https://github.com/Homebrew/homebrew-cask/pull/75004#issuecomment-570617349)—it was a vast improvement over what we had before and a 100% HBC official solution which just so happened to be hosted on that server.

> Solution :
> The vendors appcast should be used : https://updates.folivora.ai/appcast_manual.xml

That is the correct solution, yes. The only reason we hadn’t done it before is that we didn’t know about that appcast and were working with the best we could under the circumstances.

Thank you for the update.

@core-code Good news, that script and appcast can be retired!

vitorgalvao (approved):

core-code:
>  frequently carries incorrect appcast information in it's feed 

not really true. the information is correct but the script to generate the appcast runs only every 2 hours. so there is a window of up  to two hours when the information might be 'old'


>  The only reason we hadn’t done it before is that we didn’t know about that appcast and were working with the best we could under the circumstances.

we've used the official appcasts for BetterTouchTool before but decided against using it because it is very very very frequently out-of-sync with the version offered for download for the homepage. creating an appcast that specifically tracks the version on their hompage was our solution to this problem.

vitorgalvao:
Because we’ve never (that I recall) had this specific appcast and it’s [recommended by the developer](https://community.folivora.ai/t/appcast-wrong-current-version/19605/2), I’m cautiously optimistic. We should be able to know soon if this one finally fixes our problems.

tomtastic:
> > frequently carries incorrect appcast information in it's feed
> 
> not really true. the information is correct but the script to generate the appcast runs only every 2 hours. so there is a window of up to two hours when the information might be 'old'

I stand corrected, thank you for the explanation. 👍

vitorgalvao:
@tomtastic The new appcast isn’t good either. Since you’ve already initiated a conversation with the developer, do you mind enquiring as to why that’s the case? Specifically: **why is there no appcast that is kept in sync with the download offered on the homepage**? Seems like a weird oversight; this should be automated on the developer’s side.

core-code:
i don't think we need to ask this question. developers almost never have appcasts that are in sync with the homepage download. we don't have either. it is a standard procedure to use the homepage download as kind of 'beta test light' and only bring over existing users once no new user which has downloaded from the homepage complains.

tomtastic:
> @tomtastic The new appcast isn’t good either. Since you’ve already initiated a conversation with the developer, do you mind enquiring as to why that’s the case? Specifically: **why is there no appcast that is kept in sync with the download offered on the homepage**? Seems like a weird oversight; this should be automated on the developer’s side.

Developer response :

> Sorry I really don't want to maintain more appcasts :slight_smile:
> The manual one is the one you should use, everything else is irrelevant.

PR #92765, opened 2020-11-15
specify a specific version for spotify

zhao-li:
This change specifies a specific version for the Spotify cask so that brew does not constantly reinstall Spotify when the user runs `brew upgrade --greedy`

Spotify's download page lists it's version `10.5` with an URL pointing to `0.8.5` so I wasn't sure how to handle that scenario.
https://www.spotify.com/us/download/other/

Please let me know if there's anything else additional I should do on my end.

Thank you for an amazing product 🥇 

Rubocop found this error:
```
Casks/spotify.rb:1:1: C: Missing frozen string literal comment.
# typed: false
^
```
But none of the other Cask files seem to have this:
`# frozen_string_literal: true`
or this:
`# frozen_string_literal: false`

### DIFF
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -1,9 +1,9 @@
 cask "spotify" do
-  version :latest
+  version "10.5"
   sha256 :no_check
 
   # scdn.co/ was verified as official when first introduced to the cask
-  url "https://download.scdn.co/Spotify.dmg"
+  url "https://download.scdn.co/Spotify-0.8.5.dmg"
   name "Spotify"
   desc "Music streaming service"
   homepage "https://www.spotify.com/"
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -1,3 +1,4 @@
+# typed: false
 cask "spotify" do
   version "10.5"
   sha256 :no_check


reitermarkus:
> Rubocop found this error:

How/where did you run `brew cask style`?

zhao-li:
> > Rubocop found this error:
> 
> How/where did you run `brew cask style`?

I ran this command:
`brew cask style --fix Casks/spotify.rb`
at the root of the project folder.

miccal:
Thank you @zhao-li, bu the downloadn link `https://download.scdn.co/Spotify-0.8.5.dmg` provides version `0.8.5.1333.g822e0de8` which appears damaged, while the original download link provides version `1.1.45.621.gdddebadc-22` which works fine.

If you do not want `brew` to constantly reinstall Spotify, simply do not use the `--greedy` flag.

zhao-li:
@miccal thank you for reviewing the change.

The `--greedy` flag is desired because the other apps upgrade the way I want.  Spotify is the only app that has the behavior that I want to avoid.

Would it be acceptable to change `version` from:
`:latest`
to
`1.1.45.621.gdddebadc-22`
and keep the original download link of?:
`https://download.scdn.co/Spotify.dmg`

Thank you again for your time and help 👍

vitorgalvao:
> Would it be acceptable to (…)

No. Because `url` is unversioned and there’s no `appcast`, we wouldn’t know when to update the cask anyway, meaning you’d get the worse of all worlds, a stuck cask that wouldn’t update and would lie to you about the installed version.

If you want this cask to be versioned, you’ll need to find an [`appcast`](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) or convince Spotify to provide versioned downloads.

In short, we’ve done all we can on this side; improvements require either upstream collaboration or information we don’t have.

zhao-li:
That is fair enough.  Thank you all for your time and enjoy the rest of your weekend. :)

PR #92907, opened 2020-11-17
Update catalina-cache-cleaner.rb

core-code:
this seems to be in the process of being released. the webpage is already linked but not online yet. download can be found on the downloads page


core-code:
not sure why a depends on catalina was in there  - the app has never required more than 10.10

miccal (approved):

PR #92997, opened 2020-11-18
Add Avast Secure Browser

krtek:
Recently released for Mac at https://www.avast.com/secure-browser#mac

EDIT: Fixed link.

### DIFF
--- a/Casks/avast-secure-browser.rb
+++ b/Casks/avast-secure-browser.rb
@@ -10,6 +10,6 @@ cask "avast-secure-browser" do
 
   pkg "AvastSecureBrowserSetup.pkg"
 
-  uninstall delete:  "~/Library/LaunchAgents/com.avast.browser.*",
+  uninstall delete:  "/Library/LaunchAgents/com.avast.browser.*",
             pkgutil: "com.avast.browser"
 end
--- a/Casks/avast-secure-browser.rb
+++ b/Casks/avast-secure-browser.rb
@@ -11,9 +11,10 @@ cask "avast-secure-browser" do
   pkg "AvastSecureBrowserSetup.pkg"
 
   uninstall_preflight do
-    avast_launch = Open3.capture2("launchctl", "list").first.split("\n").map { |a| a.sub(/^.*\t/, "") }.select { |a| a.start_with?("com.avast.browser.") }
+    avast_launch = Open3.capture2("launchctl", "list").first.split("\n")
+                        .map { |a| a.sub(/^.*\t/, "") }.select { |a| a.start_with?("com.avast.browser.") }
   end
 
-  uninstall launchctl: "#{avast_launch}",
+  uninstall launchctl: avast_launch,
             pkgutil:   "com.avast.browser"
 end
--- a/Casks/avast-secure-browser.rb
+++ b/Casks/avast-secure-browser.rb
@@ -10,11 +10,19 @@ cask "avast-secure-browser" do
 
   pkg "AvastSecureBrowserSetup.pkg"
 
-  uninstall_preflight do
-    avast_launch = Open3.capture2("launchctl", "list").first.split("\n")
-                        .map { |a| a.sub(/^.*\t/, "") }.select { |a| a.start_with?("com.avast.browser.") }
-  end
+  uninstall quit:    "com.avast.browser",
+            pkgutil: "com.avast.browser"
 
-  uninstall launchctl: avast_launch,
-            pkgutil:   "com.avast.browser"
+  zap trash: [
+    "~/Library/Application Support/AVAST Software/Browser",
+    "~/Library/Caches/AVAST Software/Browser",
+    "~/Library/Caches/com.avast.browser",
+    "~/Library/Preferences/com.avast.AvastSecureBrowser.plist",
+    "~/Library/Preferences/com.avast.browser.plist",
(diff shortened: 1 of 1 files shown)


krtek:
Thank you, @miccal! 🙏

miccal:
Ah, I forgot that [`launchctl:` does not support wildcards](https://github.com/Homebrew/homebrew-cask/issues/79661), and since the `com.avast.browser.*` required here contains a random string, I am not sure what we can do here.

vitorgalvao:
> any thoughts or suggestions?

1. Add wildcard or regular expression support to `launchctl`.
2. Feasible now: use `delete` to get rid of the plist itself. If it’s no longer there, it will not be loaded on next boot anyway (I’ve tested in the past but don’t recall if it’ll keep working in the mean time).

Casks/avast-secure-browser.rb
```diff
@@ -0,0 +1,19 @@
...
+  # cdn-download.avastbrowser.com/ was verified as official when first introduced to the cask
+  url "https://cdn-download.avastbrowser.com/AvastSecureBrowserSetup.pkg"
+  name "Avast Secure Browser"
+  desc "Web browser focusing on privacy"
+  homepage "https://www.avast.com/secure-browser#mac"
+
+  pkg "AvastSecureBrowserSetup.pkg"
+
+  uninstall_preflight do
+    avast_launch = system_command "/bin/launchctl", args: ["list | grep -i com.avast.browser.*"]
```
vitorgalvao, line 14: Nice idea!

Should probably add a comment to explain why it’s needed. Also, as it currently is it, if they ever add more then one launch agent this will be a wrong result. Finally, a pipe as part of an argument is weird. Should probably use proper ruby for the command.

miccal: I might need some help with putting this in the proper Ruby format.

vitorgalvao:
```suggestion
    avast_launch = Open3.capture2('launchctl', 'list').first.split("\n").map { |a| a.sub(/^.*\t/, '') }.select { |a| a.start_with?('com.avast.browser') }
```

vitorgalvao: May also work with

```suggestion
    avast_launch = system_command("/bin/launchctl", args: ["list"]).split("\n").map { |a| a.sub(/^.*\t/, '') }.select { |a| a.start_with?('com.avast.browser') }
```

Casks/avast-secure-browser.rb
```diff
@@ -0,0 +1,20 @@
...
+  # cdn-download.avastbrowser.com/ was verified as official when first introduced to the cask
+  url "https://cdn-download.avastbrowser.com/AvastSecureBrowserSetup.pkg"
+  name "Avast Secure Browser"
+  desc "Web browser focusing on privacy"
+  homepage "https://www.avast.com/secure-browser#mac"
+
+  pkg "AvastSecureBrowserSetup.pkg"
+
+  uninstall_preflight do
+    stdout, *  = system_command "/bin/launchctl", args: ["list | grep -i com.avast.browser.*"]
```
vitorgalvao, line 14: As it is, it will match too much. `list` doesn’t just output the name of the agent, but other information as well. And it still doesn’t solve the issue of possible multiple agents in the future. Use one of the suggestions (the first one should work, the second one maybe).

If you any part of it isn’t clear, do let me know.

Casks/avast-secure-browser.rb
```diff
@@ -0,0 +1,19 @@
...
+  # cdn-download.avastbrowser.com/ was verified as official when first introduced to the cask
+  url "https://cdn-download.avastbrowser.com/AvastSecureBrowserSetup.pkg"
+  name "Avast Secure Browser"
+  desc "Web browser focusing on privacy"
+  homepage "https://www.avast.com/secure-browser#mac"
+
+  pkg "AvastSecureBrowserSetup.pkg"
+
+  uninstall_preflight do
+    avast_launch = Open3.capture2("launchctl", "list").first.split("\n").map { |a| a.sub(/^.*\t/, '') }.select { |a| a.start_with?("com.avast.browser.") }
```
vitorgalvao: Probably going to fail with line too long. We might need to break it up.

Casks/avast-secure-browser.rb
```diff
@@ -0,0 +1,20 @@
...
+  # cdn-download.avastbrowser.com/ was verified as official when first introduced to the cask
+  url "https://cdn-download.avastbrowser.com/AvastSecureBrowserSetup.pkg"
+  name "Avast Secure Browser"
+  desc "Web browser focusing on privacy"
+  homepage "https://www.avast.com/secure-browser#mac"
+
+  pkg "AvastSecureBrowserSetup.pkg"
+
+  uninstall_preflight do
+    avast_launch = Open3.capture2("launchctl", "list").first.split("\n")
```
vitorgalvao, line 14: This might still not work as is, because this variable may not be available outside the block

miccal:
@vitorgalvao would you beliueve that none of this was actually necessary -- the `.pkg` installer runs the `.app` once it succeeds, which creates an ad-hoc `launchctl` job while the `.app` is running.

So all that was required was to `quit` the `.app`.

krtek:
Oops. I didn't know it makes a difference. Thanks for all the work you put into this!

vitorgalvao (approved):

PR #92998, opened 2020-11-18
Use non-cask commands where possible.

reitermarkus:
Only `brew cask zap` and `brew cask audit` to go now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Clean up CI machine
         run: |
-          if ! brew cask ls visual-studio &>/dev/null; then
+          if ! brew list --cask visual-studio &>/dev/null; then
             if ! rm -r '/Applications/Visual Studio.app'; then
               echo '::warning::Removing Visual Studio is no longer necessary.'
             fi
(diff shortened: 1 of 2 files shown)


miccal:
@reitermarkus I have found that the `--cask` flag in `brew uninstall --cask {cask_name}` to be a little flaky for those (rare) cases where there is a Formula and a Cask that share the same name -- it might be easier to use `brew uninstall homebrew/cask/{cask_name}` for a Cask, and just keep using `brew uninstall` for Formula.

Just a thought ... :)

reitermarkus:
>  a little flaky

What exactly does that mean? Specifying the `--cask` flag should definitely not be flaky.

reitermarkus:
Ah, `brew uninstall` doesn't even support the `--cask` flag, that's why …

miccal:
> What exactly does that mean? Specifying the `--cask` flag should definitely not be flaky.

Oh sorry, i meant with `brew install`, not `brew uninstall`.

I recently did a clean install of Big Sur and while installing the Casks `gitup` and `transmission` (which each share names with a Formula) I noticed that the `--cask` flag did not work for one of them for some reason.

I was installing many things via a single command, and hence could not isolate and further investigate why it did not work for one of them.

vitorgalvao:
> Only `brew cask zap` and `brew cask audit` to go now.

`brew cask cat` and `brew cask _stanza` still only work with `cask`.

reitermarkus:
I meant this in terms of the CI workflow.

reitermarkus:
Also `fetch` is still missing.

miccal:
And `create` and `edit`.

PR #93121, opened 2020-11-19
Add Auto Update to Zoom Cask

ghost:
It auto updates.

- [X] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.

### DIFF
--- a/Casks/zoomus.rb
+++ b/Casks/zoomus.rb
@@ -8,6 +8,8 @@ cask "zoomus" do
   name "Zoom.us"
   desc "Video communication and virtual meeting platform"
   homepage "https://www.zoom.us/"
+  
+  auto_updates true
 
   conflicts_with cask: "zoom-for-it-admins"
 


miccal:
https://github.com/Homebrew/homebrew-cask/pull/93083

rhanneken:
I think it's confusing for the user to have competing upgrade notifications. Today I was prompted by the Zoom app to upgrade, and I did so. Then I did `brew update` and `brew outdated`, and now I'm being told that Zoom is out of date, even though it's not. I'm going to have to do `brew upgrade` just to eliminate noise from the `brew outdated` output.

Going forward, I guess I can remember to always do the upgrade via Homebrew, but if other cask maintainers do the same thing, it might be hard to remember which apps need to be upgraded by which method.

When Zoom was marked as auto-updating, couldn't someone have upgraded it with Homebrew by doing `brew upgrade --force` or maybe `brew reinstall`?

vitorgalvao:
> When Zoom was marked as auto-updating, couldn't someone have upgraded it with Homebrew by doing `brew upgrade --force` or maybe `brew reinstall`?

`--force` might not even be necessary, but being explicit on the cask would need to be.

> I think it's confusing for the user to have competing upgrade notifications.

I see your point, though Zoom is not alone. Calibre also warns of updates but can’t download them. While Zoom does download the update, having to go through and installation wizard ([if I understood correctly](https://github.com/Homebrew/homebrew-cask/pull/93083)) begins to go over *the threshold of inconvenience*.

> it might be hard to remember which apps need to be upgraded by which method.

In sum, the point here is that Zoom is *worse than the norm* (i.e. this isn’t common), so an app that doesn’t have `auto_updates` is one that either can’t update itself or that doing so is a noticeable inconvenience.

Like every other policy, I’m willing to be convinced otherwise. I have no sympathy for user-hostile companies like Zoom and Adobe so I’m not as familiar with all ways they inconvenience users or how bad their upgrade policy was.

This cases seems like a sensible rule, but if you disagree, @adunning should have a change to counter-point too.

PR #93247, opened 2020-11-21
google-chrome.rb: remove cross-platform from desc

vitorgalvao:
It’s irrelevant what platforms it runs on; we only support one.

### DIFF
--- a/Casks/google-chrome.rb
+++ b/Casks/google-chrome.rb
@@ -5,7 +5,7 @@ cask "google-chrome" do
   url "https://dl.google.com/chrome/mac/stable/GGRO/googlechrome.dmg"
   appcast "https://omahaproxy.appspot.com/history?os=mac;channel=stable"
   name "Google Chrome"
-  desc "Cross-platform web browser"
+  desc "Web browser"
   homepage "https://www.google.com/chrome/"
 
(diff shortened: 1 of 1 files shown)


reitermarkus (approved):

core-code:
i agree in principle but am not sure in this case. the main "selling point" of Chrome versus Safari is that you can use it identically on windows/linux too.

vitorgalvao:
We still hold the position that users should be reasonably familiar with what they are installing, so `desc` serves as a kind of disambiguator.

Practical issues I have with `cross-platform`:

* It kind of puts an onus on us to have it on every cross-platform tool. And to have a standard way in which we add it. It’s extra work, and the kind that eventually leads to small bikeshedding disagreements.
* As as descriptive term, it isn’t too far from `open-source`, which is another term we don’t want to pursue in `desc` let it become another nightmarish `license`. Point being, I’d rather nip in the bud these terms, which are midway between useful terms and marketing fluff (e.g. `modern`).

> the main "selling point" of Chrome versus Safari is that you can use it identically on windows/linux too.

But is that useful to an HBC user when `search --desc`ing? I’d argue it’s not.

core-code:
agreed

PR #93302, opened 2020-11-21
Update icollections from 6.6,66012 to 6.7,67008

ran-dall:

### DIFF
--- a/Casks/icollections.rb
+++ b/Casks/icollections.rb
@@ -1,5 +1,5 @@
 cask "icollections" do
-  version "6.6,66012"
+  version "6.7,67008"
   sha256 "f386cc0326970da0e4ab488169abbca5709f49e47b6571d9037cb35c7053293b"
 
   url "https://naarakstudio.com/download/iCollections.dmg"


core-code:
this is another bogus update. can you disable any PR that only updates the version number but not the SHA? they seem to be always wrong

core-code:
@vitorgalvao is there any way to make travis not auto-merge these?

vitorgalvao:
> is there any way to make travis not auto-merge these?

Haven’t looked into it, but it should be on [the review action](https://github.com/Homebrew/actions/blob/master/review-cask-pr/review.rb).

reitermarkus:
> they seem to be always wrong

All of them, or just the ones where the URL is unversioned?

core-code:
> All of them, or just the ones where the URL is unversioned?

i've had a look at the last few cases where the problem happened, it was in 

```
macx-dvd-ripper-pro
naver-whale
nvidia-geforce-now
```

the first two are static downloads just like `icollections` but `nvidia-geforce-now` is not - and with this cask the problem even happened twice in the last 30 days.

so i'd prefer if 'version number only' updates would never get automerged - regardless of the URL being versioned or not.

i guess this should only apply to casks that actually have a SHA though.

PR #93326, opened 2020-11-22
Switch google-chrome to Universal App build

unitof:
I've tested & confirmed this DMG installs & runs on both Apple Silicon Big Sur & Intel Catalina. Should probably confirm Yosemite can run Universal Apps. If not, update `depends_on macos:` to `">= catalina"`.

Also wasn't sure if `appcast` should be updated to `https://omahaproxy.appspot.com/history?os=mac_arm64;channel=stable` to show only Apple Silicon-compatible build history. Left as-is for now.

- [x] `brew cask audit --download google-chrome` is error-free.
- [x] `brew cask style --fix google-chrome` reports no offenses.

### DIFF
--- a/Casks/google-chrome.rb
+++ b/Casks/google-chrome.rb
@@ -1,8 +1,8 @@
 cask "google-chrome" do
   version "87.0.4280.67"
-  sha256 "d768f823e56b0dde4e68fd785db0c535ceb9fb1b4d5a5e9bf673e5b4a59e8ea4"
+  sha256 "c3e771740a0e9398d188b9cf9c7806f61a7d5775b056abec8f608a3561d4cf35"
 
-  url "https://dl.google.com/chrome/mac/stable/GGRO/googlechrome.dmg"
+  url "https://dl.google.com/chrome/mac/universal/stable/GGRO/googlechrome.dmg"
   appcast "https://omahaproxy.appspot.com/history?os=mac;channel=stable"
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):

i0ntempest:
@vitorgalvao Is there a way homebrew can check CPU arch before installation (and how to code it in this cask)? The universal app is 2 times the size of x86 only one and I don't really want to waste my disk space if not needed.

core-code:
> don't really want to waste my disk space if not needed.

thousands of Mac apps will switch to 'FAT' binaries in the coming weeks and 99% of those aren't going to continue shipping a Intel-only version like Chrome does. 

what i want to say is, if disk space is that important, a more general solution will be needed - not a special case for Google Chrome.

vitorgalvao:
Not sure how our `arch` bindings are these days. But I don’t think we’d want that burden right now while HB itself isn’t fully ARM-ready yet. For the moment we should do what gives the most results with least effort, and that’s the universal binary.

> thousands of Mac apps will switch to 'FAT' binaries in the coming weeks and 99% of those aren't going to continue shipping a Intel-only version like Chrome does.

I think so too. Which reinforces the idea that separating by arch inside the cask right now wouldn’t be a good way to spend effort.

i0ntempest:
Okay fair enough. I'll wait a bit and see how HB changes for ARM.

PR #93336, opened 2020-11-22
google-chrome: support both universal and x86_64 version

i0ntempest:

### DIFF
--- a/Casks/google-chrome.rb
+++ b/Casks/google-chrome.rb
@@ -1,8 +1,14 @@
 cask "google-chrome" do
-  version "87.0.4280.67"
-  sha256 "c3e771740a0e9398d188b9cf9c7806f61a7d5775b056abec8f608a3561d4cf35"
+  if RUBY_PLATFORM.include? "x86_64"
+    sha256 "d768f823e56b0dde4e68fd785db0c535ceb9fb1b4d5a5e9bf673e5b4a59e8ea4"
+    url "https://dl.google.com/chrome/mac/stable/GGRO/googlechrome.dmg"
+  else
+    sha256 "c3e771740a0e9398d188b9cf9c7806f61a7d5775b056abec8f608a3561d4cf35"
+    url "https://dl.google.com/chrome/mac/universal/stable/GGRO/googlechrome.dmg"
+  end
 
-  url "https://dl.google.com/chrome/mac/universal/stable/GGRO/googlechrome.dmg"
+  version "87.0.4280.67"
+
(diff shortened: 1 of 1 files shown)


i0ntempest:
The universal version is 2x bigger (200MB bigger) than x86_64 version and I think most people don't want to waste disk space on codes they can't run.

reitermarkus:
I think we should prefer easier to maintain casks over disk space. Also, there is no way to test this on CI since we don't have CI for ARM.

core-code:
> I think we should prefer easier to maintain casks over disk space.

agreed.

how about solving this by adding a new cask 'google-chrome-intel' to the versions repo for those people who require minimal disk-space consumption?

vitorgalvao:
Closing since we’re in agreement.

> how about solving this by adding a new cask 'google-chrome-intel' to the versions repo for those people who require minimal disk-space consumption?

I think we should recommend personal taps for that. I doubt the number of people with this specific need will be common enough.

core-code:
note if you do add a 'google-chrome-intel' to the versions repo i can keep it up-to-date with the new system i have to update casks with static urls.

otherwise, we have a version here: http://macupdater.net/casks_manual/google-chrome-intel.rb

i0ntempest:
> note if you do add a 'google-chrome-intel' to the versions repo i can keep it up-to-date with the new system i have to update casks with static urls.
> 
> otherwise, we have a version here: http://macupdater.net/casks_manual/google-chrome-intel.rb

Is this identical to the version of the google-chrome cask file before the universal binary switch? If so I'll just add this to the versions repo.

vitorgalvao:
> If so I'll just add this to the versions repo.

I’d rather not. I’ll gladly let this go if other maintainers disagree, but the versions repo—despite stricter rules—is still a mess. Are we prepared to start getting a bunch of low usage `-intel` casks that (I predict) will start to be abandoned as upstream and users cease to care (which I bet won’t take long)?

Will we then start having `-arm` or `-universal` versions as well when upstream starts changing what they provide?

So far we’ve had **one** user mentioning this as desirable. And it’s not even that big of a deal, but a minor inconvenience. Unless we have a horde of users asking for this, I think it’s premature to set the precedent.

miccal:
Until Homebrew officially supports Apple Silicon, we should not accept Casks that install things that are specifically for Apple Silicon only.

See the discussion [here](https://github.com/Homebrew/brew/issues/9177), which indicates that there may be some quite fundamental changes when Apple Silicon is officially supported.

PR #93438, opened 2020-11-24
rejected casks: obscure: remove reference to self-submitted

vitorgalvao:
That part of the reasoning is the least important and it’s what trips people over more often.

### DIFF
--- a/doc/faq/rejected_casks.md
+++ b/doc/faq/rejected_casks.md
@@ -15,7 +15,7 @@ Common reasons to reject a Cask entirely:
 + The download URL for the app is both behind a login/registration form and from a host that differs from the homepage, meaning users can’t easily verify its authenticity. [alehouse/homebrew-unofficial](https://github.com/alehouse/homebrew-unofficial) is a sister repo where you may wish to submit your cask.
 + The Cask is for an app that is unmaintained (no releases in the last year, or [explicitly discontinued](https://github.com/Homebrew/homebrew-cask/pull/22699)).
 + The Cask is for an app that is too obscure. Examples:
-  + A self-submitted app from a code repository that is not notable enough (under 30 forks, 30 watchers, 75 stars).
+  + An app from a code repository that is not notable enough (under 30 forks, 30 watchers, 75 stars).
   + [Electronic Identification (eID) software](https://github.com/Homebrew/homebrew-cask/issues/59021).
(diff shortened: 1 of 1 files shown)


doc/faq/rejected_casks.md
```diff
@@ -15,7 +15,7 @@ Common reasons to reject a Cask entirely:
 + The download URL for the app is both behind a login/registration form and from a host that differs from the homepage, meaning users can’t easily verify its authenticity. [alehouse/homebrew-unofficial](https://github.com/alehouse/homebrew-unofficial) is a sister repo where you may wish to submit your cask.
 + The Cask is for an app that is unmaintained (no releases in the last year, or [explicitly discontinued](https://github.com/Homebrew/homebrew-cask/pull/22699)).
 + The Cask is for an app that is too obscure. Examples:
-  + A self-submitted app from a code repository that is not notable enough (under 30 forks, 30 watchers, 75 stars).
+  + An app from a code repository that is not notable enough (under 30 forks, 30 watchers, 75 stars).
```
miccal, line 18: Perhaps?
```suggestion
  + An app from a GitHub code repository that is not notable enough (under 30 forks, 30 watchers, 75 stars).
```

vitorgalvao: I think it also works for Gitlab. And being overly specific can both lead to complaints and be outdated sooner, so I’d rather have messaging we don’t need to touch.

miccal (approved):

strafe:
@vitorgalvao unsure but it seems like the `brew cask audit --new-cask <cask>` / `brew audit --new-cask <cask>` has never been mentioned in the documentation regarding [adding new casks](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md) which is the only audit check that checks for this(?).

Obviously there's the bigger issue of a lot of the example commands across all of the docs using the deprecated `brew cask` command.

vitorgalvao:
> it seems like the `brew cask audit --new-cask <cask>` / `brew audit --new-cask <cask>` has never been mentioned in the documentation

It might be only on the `man` page.

> Obviously there's the bigger issue of a lot of the example commands across all of the docs using the deprecated `brew cask` command.

`brew cask` isn’t deprecated in all cases yet. But yes, the document you linked to is still missing fixes.

Further PRs are welcome. @miccal has been doing a lot of work there, so feel free to ping either (or both of use) if you submit one.

PR #93444, opened 2020-11-24
zoomus.rb: add comment on auto_updates

vitorgalvao:
The alternative to this comment is to readd `auto_updates`.

From https://github.com/Homebrew/homebrew-cask/pull/93083:

> To say that Zoom auto-updates is a bit of a stretch: it has a menu option that checks for updates, downloads a new installer, and runs it, requiring extensive user intervention. It would be far quicker to have Homebrew update it.

Some discussion going on at https://github.com/Homebrew/homebrew-cask/pull/93121.


miccal:
I personally use Zoom.us and I agree that the internal update mechanism is _terrible_ and I rely on `homebrew-cask` for updates.

miccal (approved):

Issue #93482, opened 2020-11-24
Tracking Issue: “bom contains no files, nothing copied”


reitermarkus:
Let's try to figure out this error. @Homebrew/cask, please append PRs where this happens to the list below:

- https://github.com/Homebrew/homebrew-cask/pull/85673 (balenaetcher)
- https://github.com/Homebrew/homebrew-cask/pull/89709 (browserosaurus)
- https://github.com/Homebrew/homebrew-cask/pull/90070 (sizzy)
- https://github.com/Homebrew/homebrew-cask/pull/92091 (vrew)
- https://github.com/Homebrew/homebrew-cask/pull/93470 (goodsync)
- https://github.com/Homebrew/homebrew-cask/pull/93534 (trainerroad)

core-code:
https://github.com/Homebrew/homebrew-cask/pull/93534

core-code:
fixed? awesome!

reitermarkus:
fixed?

Worked around, rather. Hopefully … 😉

PR #93484, opened 2020-11-24
Try reproducing “bom contains no files, nothing copied” error.

reitermarkus:
Try reproducing by triggering CI for all casks in https://github.com/Homebrew/homebrew-cask/issues/93482.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,13 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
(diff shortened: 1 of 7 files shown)


reitermarkus:
Observation:

`find` does not find anything, on retry, `find` finds necessary files

Assumption:

DMG needs some time to be mounted, `find` fails because path doesn't exist yet on first try.

reitermarkus:
`find` seems to fail with `Interrupted system call`

PR #93628, opened 2020-11-25
Fix checksum in jump 8.6.19

SimplyDanny:


core-code:
this PR is broken - the version is not 8.6.19 but 8.6.15

the mismatch comes because we use the static download from the homepage but the version information from the appcast

@vitorgalvao should we switch over to the versioned download from the appcast?

vitorgalvao:
> should we switch over to the versioned download from the appcast?

[Definitely.](https://github.com/Homebrew/homebrew-cask/pull/93856)

PR #93739, opened 2020-11-27
Update naver-whale.rb

core-code:
fix bad pr https://github.com/Homebrew/homebrew-cask/pull/93713

### DIFF
--- a/Casks/naver-whale.rb
+++ b/Casks/naver-whale.rb
@@ -1,5 +1,5 @@
 cask "naver-whale" do
-  version "2.8.109.22"
+  version "2.8.108.15"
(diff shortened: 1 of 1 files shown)


ran-dall (approved):

reitermarkus:
@core-code, please don't apply `ci-syntax-only` in this case. A version change is not only a syntax change.

The description of the label even states: “use only for bulk changes”

I think `ci-skip-install` should probably also not be applied in this case since we still want version reversals to be fully tested.

core-code:
@reitermarkus ok thanks will try to do that. still unhappy about the time it takes to revert. magic didn't work either. https://github.com/Homebrew/homebrew-cask/actions/runs/387282862/workflow

PR #93774, opened 2020-11-27
Update naver-whale.rb

core-code:
this should prevent problems similar to https://github.com/Homebrew/homebrew-cask/pull/93739


ran-dall:
@core-code I can't pick this up with ML.

core-code:
no problem since this is a static download-url it will get updated anyway

reitermarkus (approved):

PR #93787, opened 2020-11-27
README.md: update gif and installation for lack of `cask`

vitorgalvao:


strafe:
Would you mind documenting how the GIF was made, just out curiosity. I had attempted to create a similar one but couldn't seem to get such a high fps/low file size.

vitorgalvao:
> Would you mind documenting how the GIF was made, just out curiosity.

[Screenflow](https://www.telestream.net/screenflow/), really old version (5), on an (also really old) retina Macbook Pro. Exported at 100% size ([which looks way larger when played back!](https://i.imgur.com/FNNM6WL.gif)) in the `Lossless - ProRes with Alpha` preset. Conversion to gif is via [a personal script](https://github.com/vitorgalvao/tiny-scripts/blob/master/gfv) (`brew install vitorgalvao/tiny-scripts/gfv`, use `gfv -h` for help) which uses painstakingly researched techniques to achieve the best result. Then it’s included here at full size, but shrunk in the `img` tag.

strafe:
Ah... that's a bit better, still not able to get under 1MB though. Maybe that's due to my source video? ([vid.zip](https://github.com/Homebrew/homebrew-cask/files/5609736/vid.zip))

reitermarkus (approved):

PR #93802, opened 2020-11-28
Rename zoomus to zoom

vitorgalvao:
The `zoom` cask is [a 38 star open-source project](https://github.com/Logicalshift/zoom). How many erroneous downloads do you think it gets?

```
zoom
30 days: 2,288 (#65)
90 days: 5,708 (#70)
365 days: 21,843 (#316)
Age: 2242 days (added 2014, October 09)

zoomus
30 days: 6,918 (#18)
90 days: 17,365 (#19)
365 days: 56,838 (#60)
Age: 2466 days (added 2014, February 27)
```

And we’re not consistent, considering we have `zoomus-outlook-plugin.rb` but `zoom-for-it-admins.rb`. Without a facility for cask renames (we should perhaps announce this on the HB twitter), this will cause some disruption, but I’d argue there’s already at least a thousand wrong installs a month from people who assume the `zoom` cask is the video conferencing app people are still [inexplicably using](https://gist.github.com/dacruz21/dd2480f195f5b48a9ab7af8b41c21404).


core-code:
i agree that the renaming makes sense

claui (approved):
This is going to improve UX by a lot, given the enormous download numbers we have have for that 38-star app.
Can be merged as-is if you ask me.

KizzyCode:
Sorry what? (insert cursing admin here 😅).
This change just broke a few dozen installations 😕

Moreover, IMHO there are a few things that are not ok here: 
 1. You really must not silently change the application a formula points to. Even if `zoom` hasn't been used much, this is really bad practice, because if such a user runs `brew upgrade` they end up with an installation of zoom.us while their original `zoom` installation has been deleted.
    No matter how rarely `zoom` was used – stuff like this is an absolute no-go!
 2. Everyone that uses some kind of deployment like `brew bundle` or ansible etc. with `zoomus` has a broken deployment now. Even worse, this change was made in less than 24h on a weekend – without any announcement on twitter or what so ever.

My two cents:
Unless there's a really important reason (e.g. malicious or abandoned/unmaintained formulas), you should not rename popular formulas until there's a way to do this without some disruption.
And you should never silently exchange the software a user has installed – even if the intentions are good, that's pretty shady and could be another candidate for "Good apps behaving bad"…

Best,
Kizzy

claui:
> And you should never silently exchange the software a user has installed

You have a valid point there. Thank you for the feedback.

The fact that the PR was merged so quickly was entirely my mistake, and I apologize for that. I wasn’t aware that BrewTestBot now auto-merges casks on approval, too. It used to not do that. I’ll absolutely consider it when reviewing future cask PRs.

With that out of the way: last time I checked, we did have a process to migrate users of renamed formulae but not for installed casks if I remember correctly.

Even if we tweeted and printed warnings all over the place, only very few users would get the message so I believe the best solution would be to auto-migrate users just like with formulae. I’m sure we’d consider a PR for such a feature. Until then, please understand that renaming casks is sometimes necessary and that we have limited means to mitigate the effects of doing so.

vitorgalvao:
> The fact that the PR was merged so quickly was entirely my mistake, and I apologize for that. I wasn’t aware that BrewTestBot now auto-merges casks on approval, too.

I got the impression that might’ve been the case, because you don’t usually comment and merge immediately, and I wouldn’t have expected you to have done so after saying “Can be merged as-is if you ask me”.

I made a mistake there as well, as I should’ve labeled it with `awaiting maintainer feedback`. For reference, both `awaiting` labels and the `do not merge` label will stop the bot from merging even on an approved PR.

> With that out of the way: last time I checked, we did have a process to migrate users of renamed formulae but not for installed casks if I remember correctly.

[Yep, that’s correct](https://docs.brew.sh/Rename-A-Formula). We’re missing a `cask_renames.json`.

> Even if we tweeted and printed warnings all over the place, only very few users would get the message

Agreed. Even when we post something everywhere, we get complaints. That said, in this case I think we should’ve done it (there’s still time).

> I believe the best solution would be to auto-migrate users just like with formulae. I’m sure we’d consider a PR for such a feature.

We would! I mention it every time someone complains of a rename (often it’s someone who never before interacted with the project) but never got any takers (usually there‘s not even a reply).

core-code:
> I made a mistake there as well, as I should’ve labeled it with awaiting maintainer feedback. For reference, both awaiting labels and the do not merge label will stop the bot from merging even on an approved PR.

me too. i remembered that i should not approve in the webinterface as it might get automerged. i thought about adding the do-not-merge label but failed to do so - sorry.

MikeFoden:
Interestingly enough, I saw the zoomus cask was deleted and was quite confused - I was a bit confused that Zoom had done something nefarious that would cause the cask to be deleted.

Coming here I am much happier to see that it was a rename, as even on brew.sh you can't see why a cask has been deleted - just that it no longer exists.

KizzyCode:
Ah, if this was by mistake please excuse my grumpy comment 😕 

>  Until then, please understand that renaming casks is sometimes necessary and that we have limited means to mitigate the effects of doing so.

While I understand this position from a maintainer perspective, IMHO it is dangerous for casks. For formulas it would be somehow okayish, because the entire build process is sandboxed.
However, this is not the case with casks - with the result that you might silently "force-execute" unvalidated 3rd party code from an untrusted/unwanted vendor on the users system.

If you insist in renaming formulas just for convenience, an alternative could have been to rename `zoom` to `logicalshift-zoom`. This would probably have eliminated the mix-ups without causing these security implications.

> I believe the best solution would be to auto-migrate users just like with formulae. I’m sure we’d consider a PR for such a feature.

Yeah, I know as a non-contributor it's easy whine 😕 
My python and ruby skills are… not that well developed; however I'll probably take a look at it during the holidays 😊

mislav:
As short-term mitigation, could a `zoomus` alias be added for the new `zoom` cask?

vitorgalvao:
> Ah, if this was by mistake please excuse my grumpy comment
> (…)
> I know as a non-contributor it's easy whine

Thank you for that.

> While I understand this position from a maintainer perspective
> (…)
> If you insist in renaming formulas just for convenience

It’s not for *our* convenience; it’s to stop literally thousands of people from installing the wrong thing every month.

> with the result that you might silently "force-execute" unvalidated 3rd party code from an untrusted/unwanted vendor on the users system.

Casks are quarantined by default, so that’s unlikely.

> As short-term mitigation, could a `zoomus` alias be added for the new `zoom` cask?

No, because casks don’t support aliases.

vitorgalvao:
> I was a bit confused that Zoom had done something nefarious

I mean, [it’s done several](https://gist.github.com/dacruz21/dd2480f195f5b48a9ab7af8b41c21404). Considering [even Apple treated it as malware at one point](https://www.theverge.com/2019/7/10/20689644/apple-zoom-web-server-automatic-removal-silent-update-webcam-vulnerability), I’d say we’d have been plenty justified in removing it.

KizzyCode:
> Casks are quarantined by default, so that’s unlikely.

No it's not – Homebrew itself says:
> Package installers may write to any location; options such as --appdir are ignored.

If an installer does not require root permissions, it can execute arbitrary code as current user without any interaction. Even if an installer requires root permissions, preinstallation scripts are executed before the user is asked for his password.

To put it differently: this change is the reason some unsuspecting oldie gamers now have zoom.us' preinstallation scripts on their systems and probably even a full zoom.us installation on their Macs.
Which is even less understandable given the fact that apparently you don't like/trust zoom very much.

If you wouldn't want to be polite, you could even say it like this:
> A package manager that silently swaps installed applications with zoom.us just for convenience – and even worse executes zoom.us code during this exchange – should maybe not use the same thread to criticise zoom.us for shady behavior.

Best,
Kizzy

vitorgalvao:
> This means: If an installer does not require root permissions

Which PKG installers do.

> Even if an installer requires root permissions, preinstallation scripts are executed before the user is asked for his password.

I don’t think that’s the case when using `installer` (which we do).

But even if all of your points were the case, you’re discussing an irrelevant hypothetical. We’ve already agreed the `zoom` cask shouldn’t have been replaced like this. There’s nothing to be gained by continuing to hammer that point.

> given the fact that apparently you don't like/trust zoom very much.

I don’t like or trust them *at all*. Just like I don’t like or trust Adobe, but I don’t let my personal philosophy dictate the course of HBC; I try to make decisions that are best for its users and its long-term longevity.

> and even worse executes zoom.us code during this exchange 

No, it does not (without consent), because the user will have to give their password.

> If you wouldn't want to be polite, you could even say it like this

Again, *way hyperbolic*. You’re comparing Zoom’s **years of continued hidden user-hostile behaviour which was admitted to be on purpose and continues to this day with no signs of stopping despite it being so serious the richest company in the world and one of the most powerful governments intervened** with a single PR on an open-source project, made openly for everyone to see and comment on, with a clear and immediately seen effect, with a change that frankly isn’t that problematic but was already agreed it wasn’t thought through all the way and an (also admitted) premature merge.

We can’t have a productive discussion if you keep being unreasonable with the scale of your arguments. You’re arguing as if shooing a fly is equivalent to hunting white rhinos.

It’s getting difficult to believe you’re arguing in good faith. Quite frankly, it makes it hard to want to continue the discussion when you’re making me waste a chunk of an already freezing and rainy day debunking your wrong assertions.

I’d really prefer not to have to lock this discussion. I want others to have a chance to speak as well. You’ve made your point; it’s been answered. Some things were agreed upon, others were not, let’s not repeat ourselves.

> Best,
> Kizzy

You don’t need to sign your posts. We’re not exchanging letters, I can see your username right there.

jacktose:
For an example of a confusing user experience, here's my adventure this morning (abridged):
```
$ brew update
zoomus ✔
$ brew upgrade zoomus
$ brew upgrade zoomus
$ brew outdated
$ brew outdated --greedy
$ brew upgrade
$ brew list
[google "brew zoomus", find this]
$ brew install zoom
$ brew list
$ brew uninstall zoomus
$ brew install zoom
$ brew reinstall zoom
```
Some kind of migration would be really, really nice.

vitorgalvao:
> Some kind of migration would be really, really nice.

**Yes, we know! PRs are welcome!** I’m out of ways of saying that. Continuing to say “it would be nice if this feature existed” won’t make it happen. There’s no magic here, we’re all *volunteer* flawed humans, same as you.

This project is for everyone, *by* everyone. It’s tiring when people who never contributed with either code or to discussions come out only to complain. That’s one of the reasons open-source developers burn out (don’t worry, not the case here).

You’re not the only ones who want this feature, we as maintainers do too! And just like you, we have lives and priorities. This might not have been handled the best way (again, we’ve already established that and learned from it) but it’s done. Frankly, there might be a silver lining here. If this affects so many people that much, maybe someone will finally come forward to implement the feature.

I can’t tell you how simple or complicated it will be, though I can say the core has been going through a lot of changes and unification so it should be easier now than in the past.

Further replies who do not advance the conversation will be hidden, so as not to muddle possible actionable steps.

jacktose:
Let me rephrase: I think this PR should be backed out ASAP and held until there is a migration path or plan, or at least messaging for the change.

core-code:
> Let me rephrase: I think 

you a entitled to you opinion but i haven't heard any arguments to support that this would be necessary. the 'logicalshift-zoom' cask is quite unpopular and almost everyone who installed it, did so by mistake. i got these usage numbers from our system:

zoom: 6006 users
logicalshift-zoom: 325 users

if the install numbers from HBC are equal this means 95% of the installs were by accident 

> and held until there is a migration path ...

will you send PRs for this?

jacktose:
I'm all in favor of `zoom` meaning the popular video app, eventually. My argument is that the change, as it happened, creates confusing and potentially dangerous behavior for users. For evidence of confusing, [see my first comment](https://github.com/Homebrew/homebrew-cask/pull/93802#issuecomment-735990217). For evidence of potentially dangerous, see KizzyCode's comments.

I'd love to make PRs someday, but for now I don't know Ruby, or the structure of Homebrew, or have any time. I'm well aware that there's no resources to create a migration script right now, which is why I'm arguing for the simple remedy of pulling this patch.

KizzyCode:
> You’re comparing Zoom’s
I'm sorry, because I didn't want to compare you to zoom.us – English is not my mother tongue. I just wanted to point out the irony that you're installing zoom.us on systems while criticising zoom.us.
It really was not my intention to discredit homebrew or make you upset – I'm a pretty happy homebrew user; thats why I'm here 😊

-----

> But even if all of your points were the case, you’re discussing an irrelevant hypothetical. We’ve already agreed the zoom cask shouldn’t have been replaced like this. There’s nothing to be gained by continuing to hammer that point.

Then that's a misunderstanding on my side – I read claui's comment
> Until then, please understand that renaming casks is sometimes necessary and that we have limited means to mitigate the effects of doing so.

like: "Ok, this was a bit hasty by accident, but we might nonetheless do the same again (maybe a tad slower)".

vitorgalvao:
Again, you’re hammering on both irrelevant *and disproven* points without advancing the discussion. Instead of repeating myself, I’ve hidden those comments for the reasons mentioned above.

What saddens me most about this issue is that we only got replies from people who never contributed and [only one real solution](https://github.com/Homebrew/homebrew-cask/pull/93802#issuecomment-735888572) (which unfortunately isn’t supported). It would have been useful useful to have had this discussion with contributors who are minimally familiar with our processes instead of wasting the better part of a day.

At least it was mostly civil, and for that I thank you. More often than not, people who never contributed and complain come in guns blazing and insulting us directly. Glad we hadn’t had to deal with that, at least.

A mitigating PR [was already underway](https://github.com/Homebrew/homebrew-cask/pull/93930). **Testers needed and welcome**.

peterlewis:
Probably not the right place to raise this as a potential solution, but with regards to the cask rename issue, it feels that having a GUID stanza present for each cask may help with a potential future solution.

Obviously having the cask.rb named as GUID.rb wouldn't be very nice, and in the case of cask-mismatch ideally there would be some sort of index that could be used to re-attribute to the appropriate .rb, as opposed to parsing 1000s of casks, but just a thought.

As to the implementation, I'm not sure I would have the skills currently to make this dream a reality, but just wanted to get thoughts on this idea...

vitorgalvao:
> it feels that having a GUID stanza present for each cask may help with a potential future solution.

That’s not necessary; there’s a perfectly acceptable and working solution in `formula_renames.json` which needs only be ported to casks. We shouldn’t make something new when a solution already exists and can potentially be shared.

joshmanders:
I just found this PR after doing `brew update` and seeing `zoomus` under Deleted Casks, and I wanted to go against the grain of the responses here and formally thank @vitorgalvao and the rest of the contributors of Homebrew. Ya'll have made my life exponentially easier and as a maintainer of open source myself, I feel for you.

<3 you all.

claui:
@joshmanders Thank you so much for your kind words.
Comments like yours mean the world to us, and help us keep going even when things are a mess.
I can’t thank you enough!

PR #93838, opened 2020-11-29
Update frescobaldi to new build (same version 3.1.2)

dliessi:

### DIFF
--- a/Casks/frescobaldi.rb
+++ b/Casks/frescobaldi.rb
@@ -1,9 +1,9 @@
 cask "frescobaldi" do
-  version "3.1.2"
-  sha256 "d70c9004c6745c46b04bc28cda9bc1e6e664d4208090dd92e29f65cf10135d7f"
+  version "3.1.2-10.11"
+  sha256 "ffa585b4e8c6e553ed32a2aae1d9d821f49f444bbb7d57e236ce6ee9f71a8915"
 
   # github.com/frescobaldi/frescobaldi/ was verified as official when first introduced to the cask
-  url "https://github.com/frescobaldi/frescobaldi/releases/download/v#{version}/Frescobaldi-#{version}-x86_64.dmg"
+  url "https://github.com/frescobaldi/frescobaldi/releases/download/v#{version.sub(/-.+/, "")}/Frescobaldi-#{version.sub(/-.+/, "")}-x86_64-#{version.sub(/.+-/, "")}.dmg"
   appcast "https://github.com/frescobaldi/frescobaldi/releases.atom"
   name "Frescobaldi"
   desc "LilyPond Editor"
--- a/Casks/frescobaldi.rb
+++ b/Casks/frescobaldi.rb
@@ -1,9 +1,9 @@
 cask "frescobaldi" do
-  version "3.1.2-10.11"
+  version "3.1.2,10.11"
   sha256 "ffa585b4e8c6e553ed32a2aae1d9d821f49f444bbb7d57e236ce6ee9f71a8915"
 
   # github.com/frescobaldi/frescobaldi/ was verified as official when first introduced to the cask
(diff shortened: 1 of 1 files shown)


Casks/frescobaldi.rb
```diff
@@ -1,9 +1,9 @@
 cask "frescobaldi" do
-  version "3.1.2"
-  sha256 "d70c9004c6745c46b04bc28cda9bc1e6e664d4208090dd92e29f65cf10135d7f"
+  version "3.1.2-10.11"
+  sha256 "ffa585b4e8c6e553ed32a2aae1d9d821f49f444bbb7d57e236ce6ee9f71a8915"
 
   # github.com/frescobaldi/frescobaldi/ was verified as official when first introduced to the cask
-  url "https://github.com/frescobaldi/frescobaldi/releases/download/v#{version}/Frescobaldi-#{version}-x86_64.dmg"
+  url "https://github.com/frescobaldi/frescobaldi/releases/download/v#{version.sub(/-.+/, "")}/Frescobaldi-#{version.sub(/-.+/, "")}-x86_64-#{version.sub(/.+-/, "")}.dmg"
```
vitorgalvao, line 6: The correct way to handle this is to make the `version` `3.1.2,10.11` and the using `version.before_comma` and `version.after_comma`.

But this is for adding support to an older macOS version that isn’t officially supported by HB, so rejecting. Thank you anyway.

dliessi: > The correct way to handle this is to make the `version` `3.1.2,10.11` and the using `version.before_comma` and `version.after_comma`.

From the [documentation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md) I had the impression that the part after the comma was reserved to a commit name, but apparently that's not the case: I added a commit to follow your suggestion.

> But this is for adding support to an older macOS version that isn’t officially supported by HB, so rejecting.

That is not correct.
I could have explained the following when I opened the PR, but I didn't expect the change to be controversial.

With the first build we had several reports that the application was not working correctly, or not working at all, in many systems with several different versions of macOS including recent ones, although the application worked correctly on other systems with the same OS versions.
When building the application again (with the same version of the source code) I decided to use a clean installation of the oldest macOS I could, trying to solve the problem for as many users as possible.
The new DMG apparently solved the problem on all versions of macOS.
The latest bug reports ([1](https://github.com/frescobaldi/frescobaldi/issues/1339), [2](https://github.com/frescobaldi/frescobaldi/issues/1336)) were related to Big Sur (including at least one Homebrew user), where the old build does not work at all and the new one works correctly (although with reduced functionality as noted in the release page).

I will remove the old build in the next days: I can wait for this PR to be merged, but I will proceed anyway if it gets rejected, which would unfortunately break this cask.
Admittedly, the naming of the DMG may be confusing: I considered first removing the old DMG and renaming the new one and then updating only the checksum in this cask, but that would only solve the problem for new installations via Homebrew.
In fact, if I understand correctly, only updating the checksum would not trigger an upgrade in existing installations, and I think that upgrading the file without changing its name, even if the version is upgraded, could confuse Homebrew, given that it would have already downloaded a file with the same name.
If this is not the case, I'm happy to proceed differently and any hint in the right direction is welcome.

-----

On a side note, given that the prebuilt application bundle lacks some functionalities due to bugs in how PyQt is packaged (as noted in the release page), I plan to move Homebrew's Frescobaldi from cask to formula soon.

vitorgalvao: > From the documentation I had the impression that the part after the comma was reserved to a commit name, but apparently that's not the case

It’s whatever’s convenient for the interpolation.

Thank you for the explanation regarding the version.

vitorgalvao (approved):

dliessi:
@vitorgalvao Thank you for taking care of this.

PR #93842, opened 2020-11-29
Update pdf-over from 4.2.2 to version :latest

suschizu:

### DIFF
--- a/Casks/pdf-over.rb
+++ b/Casks/pdf-over.rb
@@ -1,6 +1,6 @@
 cask "pdf-over" do
   version :latest
-  sha256 :no_checksum
+  sha256 :no_check
 
   url "https://webstart.buergerkarte.at/pdf-over/PDF-Over.dmg"
   name "PDF-Over"
--- a/Casks/pdf-over.rb
+++ b/Casks/pdf-over.rb
@@ -1,6 +1,6 @@
 cask "pdf-over" do
-  version "4.2.3"
-  sha256 "f8aee47f5cfe03ce3c5a1fda3f961a10aacba802650c855d8dc9bc33af16566d"
+  version :latest
+  sha256 :no_checksum
 
   url "https://webstart.buergerkarte.at/pdf-over/PDF-Over.dmg"
   name "PDF-Over"
--- a/Casks/pdf-over.rb
+++ b/Casks/pdf-over.rb
@@ -1,8 +1,8 @@
 cask "pdf-over" do
-  version "4.2.2"
-  sha256 "ea7bd079bc1eea5819401d0a81e7766c13baa35214dadcd0bb6824b79c36e479"
+  version "4.2.3"
+  sha256 "f8aee47f5cfe03ce3c5a1fda3f961a10aacba802650c855d8dc9bc33af16566d"
 
-  url "https://webstart.buergerkarte.at/pdf-over/PDF-Over-#{version}.dmg"
+  url "https://webstart.buergerkarte.at/pdf-over/PDF-Over.dmg"
   name "PDF-Over"
   desc "Digitally sign PDFs with the Austrian Buergerkarte"
   homepage "https://webstart.buergerkarte.at/pdf-over/"


core-code:
where does the URL come from? i don't see it on the homepage

Casks/pdf-over.rb
```diff
@@ -1,8 +1,8 @@
 cask "pdf-over" do
-  version "4.2.2"
```
vitorgalvao, line 2: With no version in `url` and no `appcast`, this should be `version :latest` and `sha256 :no_checksum`.

But this is an unpopular app and cask which might get few updates, so keeping the version and checksum might not be problematic. So leave as is or make the change; whichever is more convenient to you. If you opt to leave the version and it becomes an issue later on, we can revisit it then.

suschizu:
> where does the URL come from? i don't see it on the homepage

its the download from this site: https://webstart.buergerkarte.at/pdf-over/

core-code:
> With no version in url and no appcast, this should be version :latest and sha256 :no_checksum.

i do now have a system which runs every 4 hours which is able to update casks with static downloads based on their contents. no appcast needed.

core-code:
> > where does the URL come from? i don't see it on the homepage
> 
> its the download from this site: https://webstart.buergerkarte.at/pdf-over/

sorry i was blind, thought the version has been added not removed

vitorgalvao:
> i do now have a system which runs every 4 hours which is able to update casks with static downloads based on their contents. no appcast needed.

That’s great! So it’d make sense to keep the version and checksum in this case, correct?

Does the system run on every cask? Would it make sense to switch to versioned more of the `version :latest` cases?

core-code:
> That’s great! So it’d make sense to keep the version and checksum in this case, correct?

well, you can switch it over to 'version :latest' too if you prefer. but if the version number stays in there, my system should keep it up-to-date reliably.

> Does the system run on every cask? 

no it only runs on casks with a completely static (no #{} in the URL stanza) or effectively static ( just #{version.major}) URL. also i've put a few on a blacklist too, so its not 'every'. i've attached the current list.

> Would it make sense to switch to versioned more of the version :latest cases?

we can do that if we see a benefit there. for what i personally (actually professionally) want, "version :latest" cask are absolutely fine, except when the vendor also publishes beta versions which are not marked in any way. in this case a version number would be better and the new system should be able to ensure those are kept current easily.  

i do see issues with 'checksum mismatch' though in that 4-hour window. on the other hand, casks with appcast normally take longer to become updated too.

just for reference, the system does only re-download casks if the modification date or filesize of the download has changed.

[Uploading list.txt…]()

core-code:
also note that the system is currently seriously hampered by the bugs in bump cask pr.

if 'cask-repair' could take a 'sha' flag to avoid re-downloading, i could use that one ^^

vitorgalvao:
> no it only runs on casks with a completely static (no #{} in the URL stanza) or effectively static ( just #{version.major}) URL.

That’s what I meant to ask, if it ran on every applicable (unversioned) cask by default or some configuration was needed.

> we can do that if we see a benefit there.

There’s a major benefit for users, in that `upgrade` will work for the cask without `--greedy`. For us, knowing when and how often the software was updated (via cask version bumps) can also come in handy. The more casks we can have without `:latest`, the better.

> i do see issues with 'checksum mismatch' though in that 4-hour window. on the other hand, casks with appcast normally take longer to become updated too.

I think it will be fine for the overwhelmingly majority of casks. If it isn’t fine for a certain cask, we can revert that one.

> just for reference, the system does only re-download casks if the modification date or filesize of the download has changed.

Good to know. I was going to ask about the bandwidth usage.

> if 'cask-repair' could take a 'sha' flag to avoid re-downloading, i could use that one ^^

Added. It’s `-s`/`--cask-sha`. It’s untested, so do let me know if it’s not working.

core-code:
> That’s what I meant to ask, if it ran on every applicable (unversioned) cask by default or some configuration was needed.

not really by default but we get a message for each cask that is not enabled for this and not on the blacklist either, so new additions should be picked up.

> There’s a major benefit for users, in that upgrade will work for the cask without --greedy
 
given the troubles HBC has with updating casks which have been highlighted by the recent 'zoom' rename i wonder if updating apps isn't something that isn't better handled by external updaters anyway :) 

not really being serious here of course

> I think it will be fine for the overwhelmingly majority of casks. If it isn’t fine for a certain cask, we can revert that one.

sure we can add more and see how it works out. i'll let you know when adding some to the blacklist because i cannot support it

> Added. It’s -s/--cask-sha. It’s untested, so do let me know if it’s not working.

awesome, will be sure to try it, thanks!

miccal (approved):

PR #93889, opened 2020-11-30
Add XMind 2020 v10.2.1

Ninzero:

- [*] `brew audit --cask {{cask_file}}` is error-free.
- [*] `brew style --fix {{cask_file}}` reports no offenses.
- [*] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [*] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

- [*] `brew audit --new-cask {{cask_file}}` worked successfully.
- [*] `brew install --cask {{cask_file}}` worked successfully.
- [*] `brew uninstall --cask {{cask_file}}` worked successfully.
- [*] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [*] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).


vitorgalvao:
This requires explanation. What’s the difference between this and `xmind`?

Ninzero:
On the [Xmind website](https://www.xmind.net/), Xmind 8 and Xmind 2020 are two different products. It seems like they add many new features to Xmind 2020 and supports cross platforms. Besides, if you try to buy Xmind, Xmind 8 and Xmind 2020 have different prices.

vitorgalvao:
> On the [Xmind website](https://www.xmind.net/), Xmind 8 and Xmind 2020 are two different products.

That doesn’t explain the difference. It looks to me like v8 is on its way out and that v2020 might be the replacement. This isn’t a good name for this cask; I don’t expect it to be called “2020” indefinitely.

The best course of action would likely to make this the main `xmind` and move the other one to `cask-versions`. That’s why we need more information. As it stands, we don’t have a good enough understanding of the situation to merge this (and again, not with this name).

Ninzero:
> It looks to me like v8 is on its way out and that v2020 might be the replacement.
I looked through their website and found this [post](https://www.xmind.net/blog/en/xmind%3A-zen-is-now-xmind%3A-2020/). It seems like instead of XMind 8, v2020 replaced another cask 'xmind-zen'. So we should remove the Zen Cask. But I am still not sure how to name them.

Casks/xmind-2020.rb
```diff
@@ -1,12 +1,15 @@
...
   version "10.2.1-202007271856"
   sha256 "a41f8647ef655e19e4a8066746475be95fd13b501c9fa09cc81abadb1da29580"
 
   url "https://www.xmind.net/xmind/downloads/XMind-2020-for-macOS-#{version}.dmg"
   appcast "https://www.xmind.net/download/"
-  name "XMind"
-  desc "Brainstorming and mind mapping app"
-  homepage "https://www.xmind.net/zen/"
+  name "Xmind 2020"
+  desc "Full-featured mindmap software"
```
vitorgalvao, line 8: “Full-featured” is marketing fluff and needs to be removed.

Casks/xmind-2020.rb
```diff
@@ -1,12 +1,15 @@
-cask "xmind-zen" do
+cask "xmind-2020" do
```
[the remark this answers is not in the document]
Ninzero: I agree. But their website has already removed Xmind-Zen, so this name might cause confusion. Maybe we should leave it as xmind-zen for now and make a note that cask xmind-zen is actually Xmind 2020 in description part. Then we can wait and see how will the name change?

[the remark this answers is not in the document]
vitorgalvao: If you’re still interested, this sounds like the best option for now. Ensuring that the `caveat` has a link to the post you linked before.

Casks/xmind-zen.rb
```diff
@@ -4,11 +4,20 @@
...
+  name "Xmind 2020"
+  desc "Mindmap and brainstorming app"
+  homepage "https://www.xmind.net/xmind2020/"
+
+  auto_updates true
+  conflicts_with cask: "xmind"
 
   app "XMind.app"
 
   zap trash: "~/Library/Application Support/XMind ZEN"
```
vitorgalvao, line 16: Won’t this need changing?

Ninzero: No need.

vitorgalvao (approved):

PR #93892, opened 2020-11-30
pdf-over.rb: make versioned

vitorgalvao:
Per conversation in https://github.com/Homebrew/homebrew-cask/pull/93842.

### DIFF
--- a/Casks/pdf-over.rb
+++ b/Casks/pdf-over.rb
@@ -1,6 +1,6 @@
 cask "pdf-over" do
   version "4.2.3"
-  sha256 "f8aee47f5cfe03ce3c5a1fda3f961a10aacba802650c855d8dc9bc33af16566d"
+  sha256 :no_check # required as upstream package is updated in-place
 
   url "https://webstart.buergerkarte.at/pdf-over/PDF-Over.dmg"
   name "PDF-Over"
--- a/Casks/pdf-over.rb
+++ b/Casks/pdf-over.rb
@@ -1,6 +1,6 @@
 cask "pdf-over" do
   version "4.2.3"
-  sha256 :no_check # required as upstream package is updated in-place
+  sha256 :no_check
 
   url "https://webstart.buergerkarte.at/pdf-over/PDF-Over.dmg"
   name "PDF-Over"
--- a/Casks/pdf-over.rb
+++ b/Casks/pdf-over.rb
@@ -1,6 +1,6 @@
 cask "pdf-over" do
-  version :latest
-  sha256 :no_check
+  version '4.2.3'
+  sha256 'f8aee47f5cfe03ce3c5a1fda3f961a10aacba802650c855d8dc9bc33af16566d'
 
   url "https://webstart.buergerkarte.at/pdf-over/PDF-Over.dmg"
   name "PDF-Over"
--- a/Casks/pdf-over.rb
+++ b/Casks/pdf-over.rb
@@ -1,6 +1,6 @@
 cask "pdf-over" do
-  version '4.2.3'
-  sha256 'f8aee47f5cfe03ce3c5a1fda3f961a10aacba802650c855d8dc9bc33af16566d'
+  version "4.2.3"
+  sha256 "f8aee47f5cfe03ce3c5a1fda3f961a10aacba802650c855d8dc9bc33af16566d"
 
   url "https://webstart.buergerkarte.at/pdf-over/PDF-Over.dmg"
   name "PDF-Over"


Casks/pdf-over.rb
```diff
@@ -1,6 +1,6 @@
 cask "pdf-over" do
-  version '4.2.3'
```
vitorgalvao: Force of habit! Thank you, @miccal.

miccal:
What is the reasoning for this? The `url` is un-versioned and there is no `appcast`?

vitorgalvao:
> What is the reasoning for this? The `url` is un-versioned and there is no `appcast`?

See #93842. With @core-code’s system, we don’t need them. We should start switching more `version :latest` casks over, slowly.

reitermarkus:
Making casks with unversioned URLs versioned doesn't make sense to me at all. This will make the checksum fail every time there is an update, which is precisely why we have `:no_check` for unversioned casks.

vitorgalvao:
> This will make the checksum fail every time there is an update

During a 4 hour window, max. And that will only happen for a new user installing the software for the first time.

The problem with those breakages was always that we couldn’t detect them until said new user tried to install the cask and it failed. Now we can, so it’s no longer an issue. And in return users get a checksummed cask (`--require-sha` users, rejoice) and `upgrade` will work on those as well (previously `--greedy` users rejoice).

Most casks aren’t updated that often that the 4 hour window will be problematic. For those that are, we can always scale back. But the benefits of this change *far* outweigh the theoretical drawbacks (which I’m willing to bet won’t ever be felt in the vast majority of cases where it’s applied, barring upstream server oddities).

reitermarkus:
> During a 4 hour window, max.

I'd like this to **never** happen at all. Unversioned URLs should never fails due to a checksum error.

> And that will only happen for a new user installing the software for the first time.

How is that? The same problem is the case when running `brew upgrade` and the checksum fails.

vitorgalvao:
> I'd like this to **never** happen at all. Unversioned URLs should never fails due to a checksum error.

This is no different from unversioned `url`s on casks with an `appcast`. It’s the same situation with the same people ensuring those don’t get outdated for long. There is no issue here; we’re already been doing this for long and it has been working without a hitch, we’re only expanding its reach.

> How is that? The same problem is the case when running `brew upgrade` and the checksum fails.

If `version` hasn’t changed, `brew upgrade` won’t try to do anything, thus no user with the cask already installed will notice.

vitorgalvao:
The **only** reason casks with unversioned URLs and and no `appcast` couldn’t be versioned is that there was no way to periodically check for their versions (because no `appcast`). Now that reason no longer applies.

reitermarkus:
I think the best solution would be to make these casks versioned with `:no_check`. That way they get updated and never fail the checksum.

core-code:
> make these casks versioned with :no_check

shouldn't we do this for all casks with static URLs then? many have unreliable appcasts and even if they have good appcasts they are usually only updated once per day.

core-code:
while appcasts are essential to what we do, they are not bulletproof for preventing checksum failures on casks with static urls

even a very quick check reveals that we have checksum failures on `wifispoof elan anydo qiyimedia airy`, despite appcasts

of course the fact that a problem already exists doesn't mean we can just make changes that could intensify the problem

vitorgalvao:
> of course the fact that a problem already exists doesn't mean we can just make changes that could intensify the problem

True, but these aren’t problems until someone tries to install the cask for the first time and it fails, which is increasingly rare. Some casks are dead for months before anyone even notices. Now that I think about, the new system will even help with those, because it should detect dead URLs.

@reitermarkus Agreed with @core-code [above](https://github.com/Homebrew/homebrew-cask/pull/93892#issuecomment-737840646) The situation isn’t meaningfully different.

If the people who deal with the day to day (and get annoyed when it doesn’t work) are willing to experiment with giving users a better experience, we should do it. If it turns out it doesn’t work (which is unlikely, because it has been working for a long time already), we’ll revert it soon enough because we’ll get frustrated soon enough.

I get what your position is, but you haven’t justified it. We can’t reach a consensus if you don’t share your reasoning for being against the change in a way that addresses the points in favour of it.

core-code:
> Some casks are dead for months before anyone even notices. 

or people just don't report problems. i remember when i started with HBC there were hundreds of SHA mismatches. not hundreds of bug reports though. 99 in 100 people turn away when they see a problem. i did too.

i do think that the system of checking casks every 4 hours means *less* problems than similar casks with static URL and an appcast. still, leaving them un-versioned would mean no problem at all so there is a downside.

i question the idea of verifying the checksum in principle. why is it important? i don't think it provides meaningful security (checking the code signature would!) and just creates problems. i think it is a leftover from normal brew where mirrors are common and no other line of defense exists.

vitorgalvao:
> or people just don't report problems.

Fair enough.

> still, leaving them un-versioned would mean no problem at all so there is a downside.

But there’s also a huge downside to leaving them unversioned, in that users aren’t notified of updates or have to download the software every time they `upgrade`, by using `--greedy`. That’s frustrating and wastes a lot of time and bandwidth.

> i question the idea of verifying the checksum in principle. why is it important? i don't think it provides meaningful security

I wholeheartedly agree and have made that same argument more than once. All it’s good for is checking the integrity of the download, and in return it causes several issues.

> (checking the code signature would!)

And in the meantime we quarantine downloads anyway.

I’m in no way opposed to getting rid of checksum verification. Though I think this might be contentious and as such should get a broader discussion with the whole of Homebrew.

And with that I guess I’m advocating for the `:no_check` compromise.

core-code:
> And with that I guess I’m advocating for the :no_check compromise.

what i wonder where is not having a checksum at all in the caskfile would make my update system less efficient

the ideal solution here would be to make 'brew cask install' not fail if the checksum mismatches for casks with static download URL (either no #{} interpolation in the URL at all or just #{version.major} which makes it 'effectively static')

vitorgalvao:
> what i wonder where is not having a checksum at all in the caskfile would make my update system less efficient

How so? How do you use the checksum?

> the ideal solution here would be to make 'brew cask install' not fail if the checksum mismatches for casks with static download URL (either no #{} interpolation in the URL at all or just #{version.major} which makes it 'effectively static')

From a user’s point of view, having a stanza that’s interpreted differently depending on the contents of another might get very confusing. And it would break `--require-sha`.

core-code:
> How so? How do you use the checksum?

i think you are right - shouldn't pose a problem apart from a few minutes necessary to cope with the "no_check"

reitermarkus:
> shouldn't we do this for all casks with static URLs then?

Yes, we should.

core-code:
i think we are in agreement :-)

Casks/pdf-over.rb
```diff
@@ -1,6 +1,6 @@
 cask "pdf-over" do
-  version :latest
-  sha256 :no_check
+  version "4.2.3"
+  sha256 :no_check # required as upstream package is updated in-place
```
reitermarkus, line 3: I think this comment really only makes sense for versioned URLs, since with unversioned URLs this is implied.

core-code: does #{version.major} count as versioned or unversioned?

reitermarkus: I'd say if not all parts of the version are used in the URL, it counts as unversioned.

vitorgalvao: > I think this comment really only makes sense for versioned URLs, since with unversioned URLs this is implied.

It may stop contributors from trying to submit a checksum for these.

But I think there’s a style cop that would complain without the comment there, right?

reitermarkus: > It may stop contributors from trying to submit a checksum for these.

There should be an audit for this.

> I think there’s a style cop

I don't think there is a RuboCop for `:no_check`.

core-code:
one thing i worry about with all the checksums gone now is protection from bad PRs. previously we basically knew if someone was modifying the version but not the checksum then the PR was either a mistake or correcting a previous mistake. we could have prevented auto-merging those PRs as i've advocated a few times before. with the checksums gone we don't even have that option anymore.

should we restrict automerging of PRs for casks with static download URLs and without appcast to known-good-PR-sources? i am all for it :)

vitorgalvao:
> should we restrict automerging of PRs for casks with static download URLs and without appcast to known-good-PR-sources? i am all for it

That’s a good point. With an auto-merging bot and no `:no_checksum`, it means users can change cask versions willy nilly with just a PR.

core-code:
also basically all open PRs are stuck with travis false-positives. can we restart them all?

vitorgalvao:
> can we restart them all?

Yes. For reference, you can do it by ticking their checkbox [on the PR list](https://github.com/Homebrew/homebrew-cask/pulls) and then applying the `ci-requeue` label.

core-code:
thanks i tried that and it didn't seem to work. i'll try again :)

core-code:
> > one thing i worry about with all the checksums gone now is protection from bad PRs. 

> users can change cask versions willy nilly with just a PR.

it already happened https://github.com/Homebrew/homebrew-cask/pull/95290/files

miccal:
Superseded by https://github.com/Homebrew/homebrew-cask/pull/94705.

Issue #79028, opened 2020-03-21
Add `verified` parameter to `url`


vitorgalvao:
We have a [magic comment required when `url` and `homepage` hostnames differ](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment).

In addition to it being imperfect in select cases where “the smallest possible portion of the URL that uniquely identifies the app or vendor” includes a URL part that is interpolated, we shouldn’t hide functionality in comments.

The option is important for user security, and a great visual cue for maintainers reviewing a cask—if it’s there it’s an extra line; if it’s not it’s an `audit` warning. So we should not get rid of the rule, but instead make it a proper option.

Instead of (from [activitywatch](https://github.com/Homebrew/homebrew-cask/blob/b3ed4165e8cacc2a0ce63b32ac2b9eecf22a45b4/Casks/activitywatch.rb)):

```
# github.com/ActivityWatch/activitywatch was verified as official when first introduced to the cask
url "https://github.com/ActivityWatch/activitywatch/releases/download/v#{version}/activitywatch-v#{version}-macos-x86_64.zip"
```

do

```
url "https://github.com/ActivityWatch/activitywatch/releases/download/v#{version}/activitywatch-v#{version}-macos-x86_64.zip",
    verified: 'https://github.com/ActivityWatch/activitywatch'
```

This could be a simple string comparison with the start of `url`.

Ping @Homebrew/cask.

claui:
According to the documentation, the primary goal is:

> so a user auditing the cask knows the URL was verified by the Homebrew Cask team as the one provided by the vendor, even though it may look unofficial or suspicious.

I like the idea of a user casually inspecting a cask without having to resort to the stanza documentation.

I’d love to collect a few alternatives so we can decide on a solution that preserves as much of the spirit of that goal as possible.

claui:
### Option 2

An attempt to make this unambiguous:

```rb
url "https://github.com/ActivityWatch/activitywatch/releases/download/v#{version}/activitywatch-v#{version}-macos-x86_64.zip"

confirmed url_starts_with: 'https://github.com/ActivityWatch/activitywatch',
          because: '79028#issuecomment-602017687'
```

By allowing multiple copies, the `confirmed` stanza would be open for future uses:

```rb
[...]

PR #94171, opened 2020-12-03
spyder.rb: remove verified parameter

vitorgalvao:

### DIFF
--- a/Casks/spyder.rb
+++ b/Casks/spyder.rb
@@ -3,7 +3,7 @@ cask "spyder" do
   sha256 "4c92ca2e93498a93d14c49e8c46b61220d3cd99c003ffaf176e6feaa3d70887a"
 
   url "https://github.com/spyder-ide/spyder/releases/download/v#{version}/Spyder.dmg",
-      verified: "https://github.com/spyder-ide/spyder/"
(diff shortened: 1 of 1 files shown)


miccal:
SO we do not add the `http[s]://` part here? In your [original issue](https://github.com/Homebrew/homebrew-cask/issues/79028#issue-585411432) it was included?

vitorgalvao:
> In the example given in your [original issue](https://github.com/Homebrew/homebrew-cask/issues/79028#issue-585411432) it was included.

But on [the documentation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-verified) I wrote “excluding the protocol”.

Both work and I have no strong preference, so if someone does, we should go with that. Though we should decide on it before I keep adding these. Or I can keep adding with the protocol, as that’ll be easier to remove if we decide to go the other way.

Do you have a preference?

miccal:
> Do you have a preference?

I think we should go with how it was from the comment above the `url` and not include the protocol.

I _may_ have already merged some things with the protocol included, though ...

miccal (approved):

PR #94179, opened 2020-12-03
Update icollections from 6.7.2,67201 to 6.7.2,67210

core-code:

### DIFF
--- a/Casks/icollections.rb
+++ b/Casks/icollections.rb
@@ -1,6 +1,6 @@
 cask "icollections" do
-  version "6.7.2,67201"
-  sha256 "0817e68a6599fbf65344eafdd5693233aa9f645f7e0b0cf4a58dcc454b7c1695"
+  version "6.7.2,67210"
+  sha256 "e173229182d69ad4bbdbb3f134af3d0d48f131528520a4e8217a9291fe85364a"
 
   url "https://naarakstudio.com/download/iCollections.dmg"
   appcast "https://naarakstudio.com/download/icollections/update.ver",


core-code:
@reitermarkus would there be a way to make `muUpdateStaticHomebrewCask PRs` exempt from the appcast check?

reitermarkus:
I have not seen the source code for `muUpdateStaticHomebrewCask PRs` or any detailed explanation of how it works. Ideally it should be an official Homebrew command before we make it exempt from any checks by default.

I guess if the appcast isn't reliable the simplest solution would be to remove it.

core-code:
i can certainly explain how it works.

i also could extract the source code to be self-contained and publish it under an open-source license. however that is some significant work and i don't want to do it in-vain. it is ObjC so i guess it won't be added to HBC. and it seems the VirusTotalCheck code that i've extracted hasn't been in use by anyone else either. so unless someone is really committed to run the code on his Mac to help the project i don't feel like spending significant time on this. i can't even remember when the last day with less than 10 hours of work was. so if just slapping `ci-skip-appcast `labels is the cheaper option time-wise i'll take it.

core-code:
> I guess if the appcast isn't reliable the simplest solution would be to remove it.

no appcast is ever totally reliably and not having any will allow *anything* to be merged. i feel like this is the worst option.

core-code:
> i can certainly explain how it works.

it is super simple really. casks with static download URLs don't have any requirements on the the version stanza to be in a specific format to find the actual URL. we already have the actual URL. so it is just a matter of updating the version number to match what is actually contained in the download. this is exactly what happens.

every 4 hours around 300 casks with static download URL are checked for changes (both filesize and file date as reported by the server). if any of those changed, it is downloaded and extracted. the checksum of the download and the version number are compared with the stanzas in the cask. if any of those changed, a PR is sent.

the only trick is maintaining the same version number format that has been used previously. i.e. BV,SVS or just SVS or just BV, etc. also if the version number is equal but not identical (e.g. 3.0.0 and 3.0) no update is sent either so there is no noise.

reitermarkus:
> so unless someone is really committed to run the code on his Mac to help the project i don't feel like spending significant time on this

I think the goal should be that *no one* has to run this on their Mac. The goal should be to make @BrewTestBot run it with a scheduled action or similar. And obviously it is much easier to reach that goal if such a system is out in the open and can be worked on and improved upon by multiple people.

> what is actually contained in the download

So this only works for `.app` downloads by reasing the `.plist`, I assume?

> every 4 hours around 300 casks

Is this a fixed set of 300 specific casks or a random set every 4 hours?

core-code:
> I think the goal should be that no one has to run this on their Mac. The goal should be to make @BrewTestBot run it with a scheduled action or similar. And obviously it is much easier to reach that goal if such a system is out in the open and can be worked on and improved upon by multiple people.

i am fine with that goal. if someone wants to work on that i'm happy to provide my code as a base.

> So this only works for .app downloads 

thanks to @vitorgalvao 's pkg-extract script it also works on most pkg downloads. unless it is a super old/weird mpkg file. 'triple compression' also inst supported.

> by reading the .plist, I assume?

true it reads the Info.plist

> Is this a fixed set of 300 specific casks or a random set every 4 hours?

it is a fixed set of all casks with static download urls which haven't been blacklisted. 18 are blacklisted. a bit more than 300 remain.

miccal (approved):

reitermarkus:
> 18 are blacklisted.

Which ones and why?

core-code:
```
### currently has no version number in the app itself
videostream
next
=lovense-remote
=logitech-firmwareupdatetool
### broken version number in the app
zerotier-one
superbeam
bricklink-partdesigner
nodeclipse
appstudio
### doen't directly contain the app
amazon-workspaces
=jabra-direct
=canon-imageformula-driver
### filesize AND file date reporting on the server broken
macx-video
rightfont
sizzy
piskel
latest
rescuetime
eaccess
### triple compression
ilok-license-manager
### unsupported MPKG and super old
iceberg
### unsupported "lossy" mapping
=caldigit-docking-utility
qiyimedia
### has no checksum and only partial version number in cask => therefore no mapping
snagit

```

if the comments are unclear, feel free to ask for more info

PR #94528, opened 2020-12-07
Update blackhole and add support for 2/16 channels

leafac:

- [ ] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md). **I don’t know what to do here because BlackHole now supports 2 versions: 16-channel (similar to previous versions) and 2-channel (new). I broke the `blackhole` cask into `blackhole-16ch` and `blackhole-2ch`. Maybe we should just stick with `blackhole` and continue to install the 16-channel version. Maybe we should switch to the 2-channel version, which is probably more useful to most people. Maybe we should have one of the two as a `cask-versions`.**


miccal:
I think what you have proposed is is for the best, thanks @leafac.

miccal (approved):

leafac:
Thank you very much, @miccal

loikein:
Wait, people, I'm just a random user here, but would homebrew please give more helpful information when renaming casks instead of actually deleting them? It's been confusing two in a row now (Zoom seveal days ago).

leafac:
@loikein I think the non-cask part of homebrew does something along those lines. Also, it seems that the non-cask and cask parts of homebrew are becoming closer and closer. Maybe something like what you want will exist for casks soon.

loikein:
@leafac Thank you, fingers crossed…

PR #94919, opened 2020-12-08
Update deepl from 1.13.1,115943 to 1.12.0,1120000

core-code:

### DIFF
--- a/Casks/deepl.rb
+++ b/Casks/deepl.rb
@@ -1,6 +1,6 @@
 cask "deepl" do
   version "1.12.0,1120000"
-  sha256 :no_check
+  sha256 :no_check 
 
   url "https://appdownload.deepl.com/macos/DeepL.dmg"
   name "DeepL"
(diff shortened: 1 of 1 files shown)


core-code:
@vitorgalvao @reitermarkus it seems 'cask-repair' automatically re-adds checksums to casks with :no_check. what shall we do?

vitorgalvao:
> it seems 'cask-repair' automatically re-adds checksums to casks with :no_check. what shall we do?

Not when `# required as upstream package is updated in-place` is used, which these don’t. I’m not entirely convinced leaving that comment out is the right choice, because its absence might incentivise contributors to (again) start submitting versioned cask with versioned URLs and `:no_check`.

I’ve pushed a change to `cask-repair` to not update the checksum if it already was `:no_check` and the cask is versioned. This won’t be optimal in very specific use cases, like an unversioned cask becoming versioned with a new URL, but handling that as well is introducing even more complexity to a script that should be deprecated and phased out—it’s orders of magnitude more complex than it was when I introduced it and the use cases it’s serving are different from the original goals.

`bump-cask-pr` needs to be fixed to properly handle the needs of casks as they *are*, not treat them like formula.

core-code:
thanks for the quick fix, 'cask-repair' has the best customer support!

PR #95183, opened 2020-12-10
Revert "Update tresorit from latest to 3.5.1967.1350,1967"

Cyberbeni:
Reverts Homebrew/homebrew-cask#94816

### DIFF
--- a/Casks/tresorit.rb
+++ b/Casks/tresorit.rb
@@ -1,5 +1,5 @@
 cask "tresorit" do
-  version "3.5.1967.1350,1967"
+  version :latest
   sha256 :no_check
 
   # installerstorage.blob.core.windows.net/ was verified as official when first introduced to the cask


Cyberbeni:
As I was trying to explain in the original PR, people have been doing staged rollout for decades and specifying the version number in the formula makes more people update than intended, so please don't do that.

reitermarkus:
If you add `auto_updates true`, this will behave the same way as `:latest`. People will have to use `--greedy` to upgrade it.

Cyberbeni:
No, people just need to have the app run it to update it because it automatically updates. I can't explain staged rollout any clearer, please look it up on google or wikipedia if you still don't understand it.

reitermarkus:
I know what a staged roll-out is. What I am telling you is this: Whether this cask uses `:latest` or a version does not make a difference to users, given it specifies `auto_updates true`. And we prefer versions over `:latest`.

Cyberbeni:
You explained in your previous comment that it does make a difference.

reitermarkus:
I didn't. To be explicit, these are the two equal options in terms of behaviour:

- If it uses `:latest`, people will have to specify `--greedy` to force an upgrade.
- If it uses a version and `auto_updates true`, people will have to specify `--greedy` to force an upgrade.

Cyberbeni:
What happens when the file behind the link changes and you download a different version than what is specified in the formula?

reitermarkus:
The same as with `:latest`, except that it will only upgrade again if the version in the cask changes instead of every time you run with `--greedy`.

vitorgalvao:
> I can't explain staged rollout any clearer, please look it up on google or wikipedia if you still don't understand it.

Please understand you’re the one who’s wrong here. You seem to be getting frustrated, and it’s without reason. We already understood your use case, it’s you who hasn’t understood how Homebrew Cask works. *This will work as you expect it to*.

Because of `auto_updates true`, the cask will behave on `upgrade` exactly like it did when it had `:latest`. And when `upgrade --greedy` is called, it won’t force a reinstall unless the version in the cask has changed. So scratch what I said before; **by being versioned, the cask will work better than what you expected**, because even users using `--greedy` won’t be updated immediately.

I’m not sure we can be any clearer than that.

PR #95208, opened 2020-12-11
version.md: mention :latest

vitorgalvao:
About the one place where it should be mentioned.

### DIFF
--- a/doc/cask_language_reference/stanzas/version.md
+++ b/doc/cask_language_reference/stanzas/version.md
@@ -37,7 +37,7 @@ url "https://example.com/#{version.sub(%r{build\d+}, '')}/file-version-#{version
 The special value `:latest` is used on casks which:
 
 1. Have a `url` without a version in it.
-2. Having a correct value to `vesion` is too difficult or impractical, even with our automated systems.
+2. Having a correct value to `version` is too difficult or impractical, even with our automated systems.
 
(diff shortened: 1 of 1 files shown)


miccal (approved):

doc/cask_language_reference/stanzas/version.md
```diff
@@ -32,6 +32,15 @@ version "1.2.3build4"
...
 ```
 
+## version :latest
+
+The special value `:latest` is used on casks which:
+
+1. `url` doesn’t contain a version.
+2. Having a correct value to `version` is too difficult or impractical, even with our automated systems.
+
+Example: [spotify.rb](https://github.com/Homebrew/homebrew-cask/blob/f56e8ba057687690e26a6502623aa9476ff4ac0e/Casks/spotify.rb#L2)
```
reitermarkus, line 42: I don't really get why Spotify is an exception.

reitermarkus: Has this been discussed somewhere?

vitorgalvao: > I don't really get why Spotify is an exception.

I don’t expect this one to be definitive, I just had to put *something* for now.

> Has this been discussed somewhere?

Though I’ve picked it because I thought I saw it mentioned recently as possibly problematic. But I’m not finding the issue, so I must be misremembering the cask.

vitorgalvao: [Found it](https://github.com/Homebrew/brew/pull/9423#issuecomment-742749871).

PR #95218, opened 2020-12-11
update upic from v0.21.0 to v0.21.1

gee1k:

### DIFF
--- a/Casks/upic.rb
+++ b/Casks/upic.rb
@@ -1,6 +1,6 @@
 cask "upic" do
-  version "0.21.0"
-  sha256 "6957497e865b7ecba634ddbba77001f0881fbe7402c1381a470e0f201bdb677c"
+  version "0.21.1"
+  sha256 "1152e2f3995cc33d16d764348618a70a9fb067f2b17f548a813646809aa1154c"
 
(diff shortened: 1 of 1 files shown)


core-code:
>  binary "#{appdir}/uPic.app/Contents/MacOS/uPic", target: "upic"

why was this removed?

gee1k:
Tested must be executed through the full path to run properly, no solution has been found yet, so first remove the

miccal:
> Tested must be executed through the full path to run properly, no solution has been found yet, so first remove the

Please elaborate -- what is the error you are experiencing?

reitermarkus:
> must be executed through the full path to run properly

This is a common problem. Add a shim script instead.

gee1k:
> > must be executed through the full path to run properly
> 
> This is a common problem. Add a shim script instead.

What is the shim script you are talking about?

could u show me an example?

gee1k:
Is it possible to add a script that calls the app via an absolute path and then link to that script?

gee1k:
Like this?

image

gee1k:
Thanks for the reminder, I've solved the problem using shims script in the new PR #95315

PR #95475, opened 2020-12-14
Remove depends on x11

MikeMcQuaid:
Remove `depends_on :x11` from all Casks.

This is needed for https://github.com/Homebrew/brew/pull/9403

- `brl-cad-mged`: this needs XQuartz specifically (tried with formulae but couldn't get it to work)
- `dia`: this prompts for XQuartz to be installed on startup (tried with formulae but couldn't get it to work)
- `gtkwave`: this doesn't need XQuartz at all (I don't have it installed and worked without)
- `near-ncl`: this needs XQuartz specifically (tried with formulae but couldn't get it to work)
- `x2goclient`: this doesn't need XQuartz at all (I don't have it installed and worked without)

### DIFF
--- a/Casks/brl-cad-mged.rb
+++ b/Casks/brl-cad-mged.rb
@@ -8,7 +8,7 @@ cask "brl-cad-mged" do
   name "BRL-CAD"
   homepage "https://brlcad.org/"
 
-  depends_on x11: true
+  depends_on cask: "xquartz"
 
   app "BRL-CAD : MGED #{version}.app"
 end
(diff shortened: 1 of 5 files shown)


vitorgalvao (approved):

vitorgalvao:
Removing from documentation in https://github.com/Homebrew/homebrew-cask/pull/95483.

MikeMcQuaid:
Thanks for merge and the follow-up PR @vitorgalvao!

PR #95514, opened 2020-12-15
Update balenaetcher.rb

Mactastic1-5:
Removed auto-update since it never automatically updates. Bug in repo.

### DIFF
--- a/Casks/balenaetcher.rb
+++ b/Casks/balenaetcher.rb
@@ -9,8 +9,6 @@ cask "balenaetcher" do
   desc "Tool to flash OS images to SD cards & USB drives"
   homepage "https://balena.io/etcher"
 
-  auto_updates true
-
   app "balenaEtcher.app"
 
   uninstall quit: [


vitorgalvao:
> Removed auto-update since it never automatically updates. Bug in repo.

Is there an issue you can point us to? I can imagine someone else readding that stanza, so we should have something to show as the reason to not have it.

Stellarspace:
> > Removed auto-update since it never automatically updates. Bug in repo.
> 
> Is there an issue you can point us to? I can imagine someone else readding that stanza, so we should have something to show as the reason to not have it.

https://github.com/balena-io/etcher/issues/3100

I’ve tested the app as well. There’s no way to manually trigger an update except toggling the auto-update setting, which appears to do nothing. It’s been this way for a while now.

vitorgalvao:
Thank you.

vitorgalvao (approved):

PR #95563, opened 2020-12-15
Update transcribe

core-code:

### DIFF
--- a/Casks/transcribe.rb
+++ b/Casks/transcribe.rb
@@ -4,8 +4,8 @@ cask "transcribe" do
     sha256 "f01781100cd3b9987c8f8892145a2eaa358df07b92e10e26f30b6a877f5b352c"
     url "https://www.seventhstring.com/xscribe/downmac_old/transcribe#{version.no_dots}.dmg"
   else
-    version "8.80.0"
-    sha256 :no_check
+    version "8.80.2"
+    sha256 "f01781100cd3b9987c8f8892145a2eaa358df07b92e10e26f30b6a877f5b352c"
     url "https://www.seventhstring.com/xscribe/transcribe.dmg"
   end
 


reitermarkus (approved):

core-code:
sorry - no idea why this happened but i guess it is a cask-repair bug 

ping @vitorgalvao

vitorgalvao:
Does the latest version of `bump-cask-pr` handle it correctly?

core-code:
seems so https://github.com/Homebrew/homebrew-cask/pull/95616/files

vitorgalvao:
I think I’ve found the issue. Did you run the command on Big Sur or under (guessing under)?

@reitermarkus Setting `MACOS_VERSION` in the shell no longer seems to work to force Homebrew to use a specific version. I grepped the code and found references to `HOMEBREW_MACOS_VERSION` and `HOMEBREW_MACOS_VERSION_NUMERIC`, but neither seems to do it.

core-code:
the Mac running muUpdateStaticHomebrewCask is on 10.15

vitorgalvao:
> the Mac running muUpdateStaticHomebrewCask is on 10.15

[This line](https://github.com/vitorgalvao/tiny-scripts/blob/d36ac0e809717c1e15335c9d6e27a1660d9b7a23/cask-repair#L4) (used to) force Hoembrew to act like the user was on the latest version of macOS, so commands consider that branch of the MacOS.version conditional.

In this case, `brew cask _stanza transcribe` should have returned `:no_check`, but because that variable [is no longer working](https://github.com/Homebrew/homebrew-cask/pull/95563#issuecomment-745573675), it’s using the macOS version of the machine, grabbing *that* checksum, which exists and therefore is not ignored. Additionally, the last `sha256` is always the one modified (because it assumes it to be the latest version).

Put it all together, and we get that specific bug. It will be fixed if the variable is.

core-code:
cask-repair has never worked 'correctly' for me on anything but the latest macOS version (for casks that have relevant if/else) and i've assumed this was a known limitation.

vitorgalvao:
> cask-repair has never worked 'correctly' for me on anything but the latest macOS version (for casks that have relevant if/else) and i've assumed this was a known limitation.

It was, until that variable was introduced to the script. Even then it would only update the latest version, but it would work correctly independently of your installed OS.

No idea when that variable stopped working.

PR #95706, opened 2020-12-16
Remove SoftU2F

fletchto99:
[SoftU2F](https://github.com/github/SoftU2F) is no longer being maintained by GitHub. Furthermore, the functionality it provided is now natively supported by browsers using TouchID. Since the project is no longer being supported it probably makes sense to also remove it from homebrew.


miccal:
Thank you @fletchto99.

miccal (approved):

PR #95810, opened 2020-12-18
Revert "Update pixelpeeper from 1.0 to 1.0.1"

core-code:
Reverts Homebrew/homebrew-cask#95694


core-code:
@suschizu can you have a look where the faulty pr did come from?

PR #95897, opened 2020-12-18
Update macpilot

core-code:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).
- [x] `brew audit --cask {{cask_file}}` is error-free.
- [x] `brew style --fix {{cask_file}}` reports no offenses.

Automatically created and submitted by muUpdateStaticHomebrewCask

PR #96051, opened 2020-12-19
Update macpilot.rb

core-code:
as discussed in https://github.com/Homebrew/homebrew-cask/pull/95897

### DIFF
--- a/Casks/macpilot.rb
+++ b/Casks/macpilot.rb
@@ -11,7 +11,7 @@ cask "macpilot" do
     url "https://www.koingosw.com/products/macpilot/download/old/macpilot_#{version.no_dots}_intel_for_1015.dmg"
   else
     version "12.0.5"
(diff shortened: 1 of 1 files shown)


samford (approved):
Unfortunately I can't merge this myself, as I haven't requested commit access to homebrew-cask up to this point (though I'll probably do that soon, now that we're getting to a point where we'll be adding `livecheck` blocks to casks).

I think this temporary fix is the way to go until there's a better solution for this situation, so please do merge this when you get a chance. Thanks for the quick turnaround, @core-code!

reitermarkus (approved):

reitermarkus:
This should have failed CI in https://github.com/Homebrew/homebrew-cask/pull/95897, not sure why it didn't.

reitermarkus:
Oh right, CI is using Cataline, so it doesn't recognise the `else` branch.

PR #96152, opened 2020-12-21
airtool: add livecheck

reitermarkus:

### DIFF
--- a/Casks/airtool.rb
+++ b/Casks/airtool.rb
@@ -11,7 +11,7 @@ cask "airtool" do
     strategy :sparkle
   end
 
-  pkg "Airtool_#{version}.pkg"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
It’s fine that `livecheck` replaces `appcast`; it makes sense to convert a read-only stanza to something actively useful inside casks.

But as these are being merged, just want to make sure @core-code @suschizu @ran-dall are aware, because the systems in place may suffer breakages as `appcasts` are replaced en masse.

core-code:
ugh. how about talking about making these changes before just merging them?

core-code:
@reitermarkus doing breaking changes like these without giving the slightest heads-up before (and on 21. dec no less)  means you have zero interest in the continued cooperation with me, @suschizu and @ran-dall . fine by me and good luck.

PR #96421, opened 2020-12-21
macdown: add livecheck

reitermarkus:


miccal:
Version `0.7.3` has an issue as reported in https://github.com/MacDownApp/macdown/issues/1173, and until this is resolved we cannot upgrade to this version.

PR #96504, opened 2020-12-21
Migrate gfortran to homebrew/core

### LINKED ISSUE #96135: gfortran cask duplicates the gcc formula in homebrew-core
I discovered today that there is a `gfortran` cask, which is distributing the gfortran installers that I build and make available. It's been happening for 3 years and I never knew 😄 

 I don't think that it fits the homebrew-cask rules, though:

- despite their name, these installers are a full GCC install, not just gfortran
- gfortran is fully part of GCC and cannot be separated, anyway
- this is purely command-line software
- that is already available as the `gcc` formula in homebrew-core

I could understand if the two distributions were different, but as I'm basically maintaining both, I don't think it should be kept that way. I suggest the cask be removed and users redirected to the gcc formula.

carlocab:
Closes https://github.com/Homebrew/homebrew-cask/issues/96135.

Companion PR at https://github.com/Homebrew/homebrew-core/pull/67387.


vitorgalvao (approved):

imrehorvath:
Hello! This change breaks compatibility with R CRAN source packages with fortran linkage, since they rely on the location /usr/local/gfortran. The gfortran Cask with the .pkg in it installed gfortran to that location. Simply aliasing gfortran to gcc does not provide the same setup as the Cask did.

imrehorvath:
> Hello! This change breaks compatibility with R CRAN source packages with fortran linkage, since they rely on the location /usr/local/gfortran. The gfortran Cask with the .pkg in it installed gfortran to that location. Simply aliasing gfortran to gcc does not provide the same setup as the Cask did.

Problem solved. Solution is in: https://github.com/Homebrew/homebrew-cask/issues/96135

PR #96686, opened 2020-12-22
vienna: add livecheck

reitermarkus:


barijaona:
I think that the change in version format is not pertinent in this case.
What is the motivation behind this ?

PR #97130, opened 2020-12-31
Update vienna from 3.6.2,7371 to 3.7.0

barijaona:

- [X] `brew audit --cask {{cask_file}}` is error-free.
- [X] `brew style --fix {{cask_file}}` reports no offenses.

### DIFF
--- a/Casks/vienna.rb
+++ b/Casks/vienna.rb
@@ -1,5 +1,5 @@
 cask "vienna" do
-  version "3.7.0"
+  version "3.7.0,7429"
   sha256 "a896d241622aac516ca26064d972d20be1324dc32c1f6c22a24703e8a6561ea0"
 
   url "https://dl.bintray.com/viennarss/vienna-rss/Vienna#{version.before_comma}.tar.gz",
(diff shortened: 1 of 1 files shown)


Casks/vienna.rb
```diff
@@ -1,6 +1,6 @@
 cask "vienna" do
-  version "3.6.2,7371"
-  sha256 "c43487411c8f39d29e82f0cc618a04d00a7147afa593dfe4cd45de2be5280bc6"
+  version "3.7.0"
```
Amorymeltzer, line 2: The short `sparkle:version` from the actual feed is required:

```suggestion
  version "3.7.0,7429"
```

barijaona: I am not convinced by the change introduced by PR #96686
Vienna versioning scheme never referred to the sparkle:version data. Cf. the [releases page](https://github.com/ViennaRSS/vienna-rss/releases) and previous PR #94328 or #94480.

If for some reason, it is deemed necessary to add data to the official numbering, I think it should be the commit short SHA (in this case `:5698f89d:`) not the commit sequence (7371), as this later might create confusions when dealing with different branches (Vienna has `master`, `stable` and `old stable` branches)

Amorymeltzer: Perhaps @reitermarkus can explain a bit about it since it's new to me, but for now, *not* making the change means that either the CI fails or nobody updates vienna.

miccal (approved):

PR #97235, opened 2021-01-02
steam: auto_updates

vitorgalvao:
@reitermarkus The version for this one is dead wrong. It should be excluded from automated checks because I don’t think they care about having the proper version in the app. The DMG also doesn’t necessarily contain the latest version of the app; this one is very dependent on updating itself.


vitorgalvao:
Ping @reitermarkus

reitermarkus (approved):

PR #97261, opened 2021-01-03
Fix pkg and add sha256 for mbed-studio

GHF:
Download URL redirects to a file with a versioned file name

### DIFF
--- a/Casks/mbed-studio.rb
+++ b/Casks/mbed-studio.rb
@@ -1,12 +1,12 @@
 cask "mbed-studio" do
   version "1.3.1.3"
-  sha256 :no_check
+  sha256 "cc4561b6ce367ed9fa330dcc3a88469a662c90c9be43b6e6dbecd430cf0bfea2"
 
   url "https://studio.mbed.com/installers/latest/mac/MbedStudio.pkg"
   name "Mbed Studio"
   homepage "https://os.mbed.com/studio/"
 
-  pkg "MbedStudio.pkg"
+  pkg "MbedStudio-#{version.major_minor_patch}.pkg"
 
   uninstall pkgutil: "com.arm.mbed.studio"
 


vitorgalvao (changes requested):
This is a case which should have a `sha256`, because the version number matters, even if not on `url`.

reitermarkus:
Doesn't `https://studio.mbed.com/installers/latest/mac/MbedStudio.pkg` redirect to a versioned URL?

vitorgalvao:
> Doesn't `https://studio.mbed.com/installers/latest/mac/MbedStudio.pkg` redirect to a versioned URL?

Nope. It’s `content-disposition: attachment; filename=MbedStudio-1.3.1.pkg`, not a redirect.

GHF:
> This is a case which should have a `sha256`, because the version number matters, even if not on `url`.

Good catch! I didn't even realize it was `:no_check`.

> > Doesn't `https://studio.mbed.com/installers/latest/mac/MbedStudio.pkg` redirect to a versioned URL?
> 
> Nope. It’s `content-disposition: attachment; filename=MbedStudio-1.3.1.pkg`, not a redirect.

Today I learned. 😁

vitorgalvao (approved):

GHF:
Hmmm...

```
audit for mbed-studio: failed
11
 - Use `sha256 :no_check` when URL is unversioned.
```

Any way to suppress that audit diagnostic?

vitorgalvao:
I had a fling that might be an issue. Ignoring that check for this PR, but we should make the `audit` take these into account, @reitermarkus.

vitorgalvao (approved):

MikeMcQuaid:
@vitorgalvao @reitermarkus Need to figure out a way to make these checks not get ignored as they break CI for all PRs on Homebrew/brew. Perhaps this should be a `required` CI check?

reitermarkus:
It is a required check, but applying the `ci-syntax-only` of course skipped. This should not have been marked with that label as it clearly isn't a syntax-only change.

MikeMcQuaid:
Thanks @reitermarkus. Reverted in https://github.com/Homebrew/homebrew-cask/pull/97518

reitermarkus:
It should just use `sha256 :no_check`, as the audit suggests.

vitorgalvao:
> It should just use `sha256 :no_check`, as the audit suggests.

I disagree. It’s the `audit` that is inadequate—as pointed out above—because it’s not fulfilling its purpose.

The version number is very much required for this cask to be installed correctly and not break. If this case doesn’t require a checksum, then neither do any of the other casks.

Blindly following the `audit` here is ignoring the problem; it’s changing the rules to fit the system, when it should always be the other way around.

reitermarkus:
> it’s changing the rules to fit the system

The rule is: Unversioned URLs should use `sha256 :no_check`. Why should it use a checksum in this case?

reitermarkus:
> The version number is very much required for this cask to be installed correctly and not break.

The cask will fail to install either way: If we add a checksum and the download changes, or if the version is wrong. So why not remove one of the possible ways this can fail?

reitermarkus:
We could potentially do something like this so it doesn't depend on the version and never fails:

```ruby
preflight do
  FileUtils.mv Dir["#{staged_path}/*.pkg"].first, "#{staged_path}/MbedStudio.pkg"
end
```

vitorgalvao:
> The rule is: Unversioned URLs should use `sha256 :no_check`.

Correct. But I was arguing the rule should be “Require `sha256 :no_check` when there’s no version interpolation *anywhere* on the case”.

> The cask will fail to install either way: If we add a checksum and the download changes, or if the version is wrong.

Correct, but it’s better to fail at the download stage than at the installation stage, because the former is easier to diagnose and a more common error we even reference on the issue template. Meaning that given the choice, that’s a better error to have because it will lead to quicker fixes.

> We could potentially do something like this so it doesn't depend on the version and never fails

[Agreed. That’s better.](https://github.com/Homebrew/homebrew-cask/pull/97539) Don’t we have access to `pathname` (without `require`), though? That reads way nicer.

PR #97381, opened 2021-01-07
duplicati: add livecheck

reitermarkus:


freayd:
@reitermarkus, may I know why you changed from a beta to a canary release? I'm not sure everyone wants to run a potentially less stable version.

reitermarkus:
This was part of a batch of changes and both beta and canary versions are marked as pre-releases so this was missed. The beta is over a year old at this point so I'm not sure it's worth switching back to it.

reitermarkus:
Okay, given that the homepage links to the beta, we can revert to the beta and use the homepage for the `livecheck` instead of GitHub releases. Feel free to open a PR for that.

PR #97499, opened 2021-01-08
Fix livecheck.md documentation issues

JeffreyBytes:
PR https://github.com/Homebrew/homebrew-cask/pull/97486 added livecheck documentation, but the `livecheck.md` was located in the wrong spot.
I've moved it so the links pointing to it now work.

@reitermarkus


reitermarkus (approved):
Thanks!

PR #97505, opened 2021-01-08
Update smultron from 12.2.5 to 12.2.6

JeffreyBytes:
This is a draft at the moment because I want some input.

Smultron has releases going back to version 7 to support 10.10+. So I added these.

I'm unable to test these older releases because apparently `MACOS_VERSION` environment variable no longer works (or maybe this is just on Big Sur).

Nonetheless, I'd like input on the style. Starting at version 10, the app name changed to just `Smultron.app` instead of the previously used `Smultron #{version.major}.app`. I didn't want to triple code in `app "Smultron.app"`, so I put it outside of the if-elsif statement, and only included it in version 7 and 8 (yosemite and el capitan versions).
Is this the correct way to do it or should each if-elsif statement include the `app` stanza?

### DIFF
--- a/Casks/smultron.rb
+++ b/Casks/smultron.rb
@@ -1,6 +1,6 @@
 cask "smultron" do
   version "12.2.5,12250"
-  sha256 "4fd4a60746bf49bab5aeff124ff36e5506e71dd3ab768d72cdcd817bb0147c6e"
+  sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


reitermarkus:
Generally, if you don't need these versions yourself, don't add them, since it only increases the maintenance burden.

miccal:
@reitermarkus since this only uses `#{version.major}` in the `url` shoudl this use `:no_check` for the `sha256` also?

vitorgalvao (approved):

PR #97510, opened 2021-01-08
Update encryptr from 2.0.0 to 2.1.0

JeffreyBytes:
`version` update; fixed `url`; added `desc`

### DIFF
--- a/Casks/encryptr.rb
+++ b/Casks/encryptr.rb
@@ -1,8 +1,8 @@
 cask "encryptr" do
   version "2.1.0"
-  sha256 "7b0613067aa2cd078327564273d7b907df09417dc20bb7d26d83974d487cf163"
+  sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


spaz926:
Looks like my download url is incorrect. I can't figure out how to get the direct download url.
I was able to download it by navigating here: http://www.spideroak.com/release/Encryptr/osx

spaz926:
I didn't realize we could just use that address instead of a direct link to the file.
Thanks.

miccal (approved):

spaz926:
@miccal while I have you.. It seems they will be discontinuing the product on March 4th. How does homebrew deal with apps that will be/are discontinued?
Will it get moved into a legacy tap?

miccal:
We add a `discontinued` caveat as explained [here](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md#caveats-mini-dsl).

PR #97538, opened 2021-01-08
Miniforge support ARM

### LINKED ISSUE #95776: How to create a cask with different binaries for intel / M1
I would like to create a cask that uses different binaries (i.e. URLs) for Intel and M1 processors. Unfortunately the app is not available (yet) as a universal binary. What is the recommend way to do this?  Or better: is it possible at all?

domoritz:
fixes https://github.com/Homebrew/homebrew-cask/issues/95776

### DIFF
--- a/Casks/miniforge.rb
+++ b/Casks/miniforge.rb
@@ -4,9 +4,19 @@ cask "miniforge" do
   if Hardware::CPU.intel?
     sha256 "a33aca4dd6686b1dba790b224c9e686d7e08c86e6074379194f3bde478d883ed"
     url "https://github.com/conda-forge/miniforge/releases/download/#{version}/Miniforge3-#{version}-MacOSX-x86_64.sh"
+
+    installer script: {
+      executable: "Miniforge3-#{version}-MacOSX-x86_64.sh",
+      args:       ["-b", "-p", "#{caskroom_path}/base"],
+    }
   else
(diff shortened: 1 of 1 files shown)


vitorgalvao (changes requested):
Because the script’s name changes, `installer script` will also need to be inside the conditional.

domoritz:
Good catch. Updated.

domoritz:
Would it make sense to instead of putting the whole property into the conditional to use a ternary and change the strings (similar to the version)?

vitorgalvao:
This way is more consistent with other casks. It’s not that much code, so the small amount of repetition isn’t problematic and will make fixes easier if they ever diverge.

domoritz:
I updated the formula with what I meant. Let me know what you think.

vitorgalvao:
Clever, but the way you had it is better because it’s simpler for others to understand. Those are lines that should seldom need to be revisited (due to the interpolation), so clarity is more important than brevity.

domoritz:
I like the current version and think it's quite readable (note, I am not a ruby developer and I was able to write this by just guessing). If you disagree, I am happy to revert 073cf7db3991cedb446f460f381b1c1e32129a83.

domoritz:
Reverted to the explicit version.

vitorgalvao (approved):

PR #97539, opened 2021-01-08
mbed-studio.rb: avoid versioned pkg

vitorgalvao:

---

Refs https://github.com/Homebrew/homebrew-cask/pull/97261#issuecomment-756799901.

### DIFF
--- a/Casks/mbed-studio.rb
+++ b/Casks/mbed-studio.rb
@@ -14,7 +14,7 @@ cask "mbed-studio" do
   pkg "MbedStudio.pkg"
 
   preflight do
-    Pathname(staged_path).glob("MbedStudio-*.pkg").first.rename("#{staged_path}/MbedStudio.pkg")
+    staged_path.glob("MbedStudio-*.pkg").first.rename("#{staged_path}/MbedStudio.pkg")
   end
 
   uninstall pkgutil: "com.arm.mbed.studio"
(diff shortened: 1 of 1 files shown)


Casks/mbed-studio.rb
```diff
@@ -11,7 +11,11 @@
     strategy :header_match
   end
 
-  pkg "MbedStudio-#{version}.pkg"
+  pkg "MbedStudio.pkg"
+
+  preflight do
+    Pathname.glob("#{staged_path}/MbedStudio-*.pkg").first.rename("#{staged_path}/MbedStudio.pkg")
```
vitorgalvao, line 17: `glob` should be avoided with variables, because if the value inside contains a globabble character, it will be interpreted instead of taken literally.

reitermarkus: How else are you supposed to find the `.pkg`? `Dir[]` is `Dir.glob`, so it's not doing anything differently than before.

vitorgalvao: Already did the change. With `Pathname.new(staged_path).glob("MbedStudio-*.pkg")`, even in the off chance the variable would have a globabble character, it will be interpreted literally and not as part of the glob.

reitermarkus: TIL that `Pathname#glob` exists.

reitermarkus: In that case we can just call `staged_path.glob`.

vitorgalvao: Even better!

reitermarkus (approved):

PR #97584, opened 2021-01-09
macintoshjs: ARM support

JeffreyBytes:
Added ARM support.

### DIFF
--- a/Casks/macintoshjs.rb
+++ b/Casks/macintoshjs.rb
@@ -1,14 +1,14 @@
 cask "macintoshjs" do
   version "1.1.0"
 
-  if Hardware::CPU.arm?
-    sha256 "0af95bbcc075f939d6c3e1fdcdab0d8da2760e5546aecfbc82767e326640740c"
-
-    url "https://github.com/felixrieseberg/macintosh.js/releases/download/v#{version}/macintosh.js-darwin-arm64-#{version}.zip"
-  else
+  if Hardware::CPU.intel?
     sha256 "4ca41517f15c594e718da622074a2160754cb8f1293cb1fad908f6d0ae585384"
 
     url "https://github.com/felixrieseberg/macintosh.js/releases/download/v#{version}/macintosh.js-darwin-x64-#{version}.zip"
+  else
+    sha256 "0af95bbcc075f939d6c3e1fdcdab0d8da2760e5546aecfbc82767e326640740c"
+
+    url "https://github.com/felixrieseberg/macintosh.js/releases/download/v#{version}/macintosh.js-darwin-arm64-#{version}.zip"
   end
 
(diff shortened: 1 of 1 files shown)


Casks/macintoshjs.rb
```diff
@@ -1,14 +1,14 @@
 cask "macintoshjs" do
   version "1.1.0"
 
-  if Hardware::CPU.intel?
-    sha256 "4ca41517f15c594e718da622074a2160754cb8f1293cb1fad908f6d0ae585384"
-
-    url "https://github.com/felixrieseberg/macintosh.js/releases/download/v#{version}/macintosh.js-darwin-x64-#{version}.zip"
-  else
+  if Hardware::CPU.arm?
```
vitorgalvao, line 4: @reitermarkus There are already other (few) casks with `if Hardware::CPU.intel?`. We should pick which to use for consistency. My though process for `intel?` instead of `arm?` is the same as for `MacOS.version`, where the latest released comes last on the `else`. Because all macs will eventually transition, I think it makes sense to have ARM later.

Casks/macintoshjs.rb
```diff
@@ -1,8 +1,16 @@
 cask "macintoshjs" do
   version "1.1.0"
-  sha256 "4ca41517f15c594e718da622074a2160754cb8f1293cb1fad908f6d0ae585384"
 
-  url "https://github.com/felixrieseberg/macintosh.js/releases/download/v#{version}/macintosh.js-darwin-x64-#{version}.zip"
+  if Hardware::CPU.intel?
```
vitorgalvao, line 4: Reverted to `.intel?` because most inclusions are following that pattern. Might as well be consistent now and change everything later if we decide that.

vitorgalvao (approved):

i0ntempest:
@vitorgalvao Just asking, can this `if Hardware::CPU.intel?` thing be used only with software that provides arm64 only and x86_64 only builds, or it can also be used with softwares like chrome that provides universal and x86_64 only builds?

reitermarkus (approved):

vitorgalvao:
@i0ntempest Depends if you’re asking if it can technically be done (it can) or if we allow it as a policy (we haven’t discussed it).

If the latter, I’m not sure that’s a good idea because every conditional (including in `MacOS.version`) is another place for versions to be out-of-sync. It’d be weird to have, in the same cask, the universal build be up-to-date but arch-specific versions be behind (or vice-versa).

Though I’m not against what you’re proposing. Maybe we should concentrate first in the cases where we don’t have a choice, and after those settle a bit and we see there’s no major issues regarding version parity (which I believe will be the case), start optimising the others.

If you have done some research on that regard and have, for example, a list of apps which provide both universal and arch-specific builds, I’d be interested in seeing it for context.

PR #97854, opened 2021-01-10
seafile-client: add livecheck

reitermarkus:


miccal (approved):

guillaumealgis:
Hi, I just realised that Seafile on my machine has been updated to a beta version (8.0.0). This originated in this PR, and I'm not sure it was intentional?

In addition, looks like Seafile is using Sparkle for its updates, so we should be able to use that for the livecheck, and it should be more reliable?

Apologies if I'm misunderstanding something, I never really dug in brew's internals :/

PR #97969, opened 2021-01-12
Delete arrayfire

ankane:
Hi, thanks for all the hard work maintaining Homebrew!

ArrayFire is now in homebrew-core, so this removes it here.

Ref: https://github.com/Homebrew/homebrew-core/blob/master/Formula/arrayfire.rb


miccal:
Thank you @ankane.

miccal (approved):

ankane:
Thanks @miccal!

PR #98100, opened 2021-01-13
Delete tinkerwell.rb

vitorgalvao:
Not only is this a [walled build](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) (which [we do not accept](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#unofficial-vendorless-and-walled-builds)), I’m not finding a way to download the app without buying it (even after registering an account).

I’m not sure we’re not inadvertently bypassing their purchasing mechanism, which we shouldn’t do.

```
tinkerwell
30 days: 52 (#1061)
90 days: 155 (#1085)
365 days: 365 (#1794)
Age: 277 days (added 2020, April 11)
```

I’ve [contacted them on Twitter](https://twitter.com/vhgalvao/status/1349390826590384129), as well as [the person who originally submitted it](https://github.com/Homebrew/homebrew-cask/pull/81062#issuecomment-759558344).


sschlein:
Currently, you need to purchase the app to download it but the app itself has a license key, and so it is fine to have it here as you are not bypassing our mechanism.

The new download link is https://download.tinkerwell.app instead of the digitaloceanspaces.com link.

vitorgalvao:
Thank you for the fast response! Fixing in https://github.com/Homebrew/homebrew-cask/pull/98103. It looks more complex that it is, because I took the chance to add extra changes; you would have only had to update the `url` line to fix this.

PR #98103, opened 2021-01-13
tinkerwell: add livecheck

vitorgalvao:

### DIFF
--- a/Casks/tinkerwell.rb
+++ b/Casks/tinkerwell.rb
@@ -9,7 +9,7 @@ cask "tinkerwell" do
   livecheck do
     url "https://download.tinkerwell.app/tinkerwell/latest-mac.yml"
     strategy :page_match
-    regex(/^version: (.+)/i)
+    regex(/version:\s*(\d+(?:\.\d+)*)/i)
   end
 
   auto_updates true
--- a/Casks/tinkerwell.rb
+++ b/Casks/tinkerwell.rb
@@ -2,11 +2,15 @@ cask "tinkerwell" do
   version "2.11.0"
   sha256 "322070bd4973a7e8669af74b508710de527b6f7c6c7d8316553c22904b377e10"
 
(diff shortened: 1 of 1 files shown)


Casks/tinkerwell.rb
```diff
@@ -9,7 +9,7 @@
   livecheck do
     url "https://download.tinkerwell.app/tinkerwell/latest-mac.yml"
     strategy :page_match
-    regex(/^version: (.+)/i)
+    regex(/version:\s*(\d+(?:\.\d+)*)/i)
```
vitorgalvao, line 12: @reitermarkus Why do we need such a complicated regex? What does it catch that `[\d\.]+` doesn’t?

reitermarkus: Basically to have some sort of consistency and future-proofing.

If the version changed to e.g. `1.2.3-beta.1`, we wouldn't want to match it (at least not completely, to avoid auto-merging). Of course, this specific appcast doesn't seem like it will ever contain a beta, but I think it's good to consistently use the `(\d+(?:\.\d+)*)` pattern where possible.

vitorgalvao: > If the version changed to e.g. `1.2.3-beta.1`, we wouldn't want to match it

It’s my understanding these matches only care for the first capture group. That being the case, `(\d+(?:\.\d+)*)` and `([\d\.]+)` capture the exact same thing in `1.2.3-beta.1`. You can verify that on https://regex101.com/

I understand the desire to future proof, but I’m still not seeing any case where `(\d+(?:\.\d+)*)` does a correct match and `([\d\.]+)` doesn’t.

reitermarkus: Given `1.2.3.ga`, it will match`1.2.3.` instead of `1.2.3`, for example.

miccal (approved):

PR #98206, opened 2021-01-16
Mozilla VPN 2.0.1 (new cask)

msanders:


miccal:
This seems to have some tight regional restrictions:

Screen Shot 2021-01-17 at 05 53 44

It also appears to not accept any `curl` requests.

Thank you @msanders, but I am afraid we will not be able to have this in Cask for now.

msanders:
The issue with curl requests seems to be due to the region restriction. It works correctly when issued by an IP from one of the locations described.

miccal:
> The issue with curl requests seems to be due to the region restriction. It works correctly when issued by an IP from one of the locations described.

True, but that would mean that this would not work for many users who would then submit bug reports, as they do not realise there are region restrictions on this. If/when Mozilla remove these restrictions, we could then consider adding this to `homebrew-cask`.

PR #98451, opened 2021-01-20
suspicious-package: add livecheck

miccal:
Output before:
```
|-> brew livecheck suspicious-package --debug
suspicious-package : unversioned
```
Output after:
```
|-> brew livecheck suspicious-package --debug

Cask:             suspicious-package
Livecheckable?:   Yes

URL:              https://www.mothersruin.com/software/SuspiciousPackage/data/SuspiciousPackageVersionInfo.plist
Strategy:         PageMatch

Matched Versions:
3.6,736
suspicious-package : 3.6,736 ==> 3.6,736
```

### DIFF
--- a/Casks/suspicious-package.rb
+++ b/Casks/suspicious-package.rb
@@ -10,8 +10,8 @@ cask "suspicious-package" do
   livecheck do
     url "https://www.mothersruin.com/software/SuspiciousPackage/data/SuspiciousPackageVersionInfo.plist"
     strategy :page_match do |page|
-      v = page.match(/CFBundleShortVersionString.*?(\d+(?:\.\d+)*)/i)
-      id = page.match(/CFBundleVersion.*?(\d+)/i)
+      v = page.match(/CFBundleShortVersionString.*?\n.*?(\d+(?:\.\d+)*)/i)
+      id = page.match(/CFBundleVersion.*?\n.*?(\d+(?:\.\d+)*)/i)
       "#{v[1]},#{id[1]}"
     end
   end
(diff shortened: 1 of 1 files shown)


miccal:
@reitermarkus I tried it that way first, and it does not work because of the new lines and tabs in the `plist`.

Casks/suspicious-package.rb
```diff
@@ -10,8 +10,8 @@
   livecheck do
     url "https://www.mothersruin.com/software/SuspiciousPackage/data/SuspiciousPackageVersionInfo.plist"
     strategy :page_match do |page|
-      v = page.match(/<string>(\d+(?:\.\d+)*)/i)
-      id = page.match(/<string>(\d{2,})/i)
+      v = page.match(/CFBundleShortVersionString.*?(\d+(?:\.\d+)*)/i)
+      id = page.match(/CFBundleVersion.*?(\d+)/i)
```
ykursadkaya, line 14: By default "." does not match "\n" character, so you can enable dot match newline by using "/m" modifier or you can manually match "\n".

```suggestion
      v = page.match(/CFBundleShortVersionString.*?(\d+(?:\.\d+)*)/im)
      id = page.match(/CFBundleVersion.*?(\d+)/im)
```

```suggestion
      v = page.match(/CFBundleShortVersionString.*?\n.*?(\d+(?:\.\d+)*)/i)
      id = page.match(/CFBundleVersion.*?\n.*?(\d+)/i)
```

Both versions do work as below:

```
ykursadkaya@YKKs-MacBook-Air: ~% brew livecheck --cask suspicious-package
suspicious-package : 3.6,736 ==> 3.6,736
```

vitorgalvao: Not the first case of this. It might be worth adding a specific strategy for plists, as they’re *the* way to define app versions.

miccal: Thank you @ykursadkaya -- I was sure I tried the `\n` switch!

vitorgalvao (approved):

PR #98465, opened 2021-01-21
Update uninstall stanza for folding-at-home

unitof:

### DIFF
--- a/Casks/folding-at-home.rb
+++ b/Casks/folding-at-home.rb
@@ -15,9 +15,6 @@ cask "folding-at-home" do
             quit:      [
               "org.foldingathome.fahviewer",
               "org.foldingathome.fahcontrol",
-            ]
-
-  zap trash: [
-    "/Library/Application Support/FAHClient",
-  ]
+            ],
+            rmdir:     "/Library/Application Support/FAHClient"
 end
--- a/Casks/folding-at-home.rb
+++ b/Casks/folding-at-home.rb
@@ -16,4 +16,8 @@ cask "folding-at-home" do
               "org.foldingathome.fahviewer",
               "org.foldingathome.fahcontrol",
             ]
+
+  zap trash: [
+    "/Library/Application Support/FAHClient",
+  ]
 end


Casks/folding-at-home.rb
```diff
@@ -16,4 +16,8 @@
               "org.foldingathome.fahviewer",
               "org.foldingathome.fahcontrol",
             ]
+
+  zap trash: [
+    "/Library/Application Support/FAHClient",
```
miccal, line 21: Just confirming that these files are in `/Library`, and not  `~/Library`?

unitof: Yep, confirmed.
image

miccal: Thanks @unitof -- that means that they are not user-specific files, and belong in `uninstall` not `zap`.

unitof: Gotcha, makes sense!

miccal:
Thank you @unitof.

miccal (approved):

unitof:
Hm, it seems to only remove the `FAHClient` directory (and any subdirectories) if empty, which is not desired.

Does `rmdir` by default only delete empty directories?

```
==> Uninstalling packages:
==> Removing directories if empty:
/Library/Application Support/FAHClient
/Library/Application Support/FAHClient/logs
/Library/Application Support/FAHClient/work
/Library/Application Support/FAHClient/work/00
/Library/Application Support/FAHClient/work/00/01
/Library/Application Support/FAHClient/cores
/Library/Application Support/FAHClient/cores/cores.foldingathome.org
/Library/Application Support/FAHClient/cores/cores.foldingathome.org/osx
/Library/Application Support/FAHClient/cores/cores.foldingathome.org/osx/64bit-sse2
/Library/Application Support/FAHClient/cores/cores.foldingathome.org/osx/64bit-sse2/a8-0.0.9
/Library/Application Support/FAHClient/cores/cores.foldingathome.org/osx/64bit-sse2/a8-0.0.9/Core_a8.fah
/Library/Application Support/FAHClient/cores/cores.foldingathome.org/osx/64bit-sse2/a7-0.0.19
/Library/Application Support/FAHClient/cores/cores.foldingathome.org/osx/64bit-sse2/a7-0.0.19/Core_a7.fah
```

miccal:
> Does rmdir by default only delete empty directories?

Yes it does -- do you thinks these should be removed and re-created when this is upgraded and/or reinstalled.

unitof:
Probably not. But it's definitely an artifact left by `folding-at-home` installation which is not removed by `brew uninstall --cask folding-at-home` currently. Perhaps it should move back to the `zap` stanza?

miccal:
Ping @vitorgalvao for his thoughts.

vitorgalvao:
> Does `rmdir` by default only delete empty directories?

Not “by default”, but at all. [That’s how `rmdir` is supposed to work](https://en.wikipedia.org/wiki/Rmdir).

Do these directories (being in `/Library` as opposed to `~/Library`) ever store user data or settings? I.e., do its contents change from usage? If not, they should be in `uninstall`, otherwise `zap`. But it’s not a big issue if they’re in `zap`.

unitof:
I'm not sure exactly what's stored in the `/Library/.../FAHClient` directory. I'll dig around after running it next time, but my guess is it's not user data.

Issue #98458, opened 2021-01-20
fly cask has unsigned binary; needs quarantine xattr deleted


todd-a-jacobs:
#### General troubleshooting steps

- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [x] I have retried my command with `--force`.
- [x] I ran `brew update-reset && brew update` and retried my command.
- [x] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

#### Description of issue

On Big Sur and testing on multiple machines, running the golang binary installed by the `fly` cask is not permitted without taking additional steps because it's unsigned. While the user can force the binary to run via the GUI, this can also be done using the *xattr* command as outlined below:

``` bash
xattr -dr com.apple.quarantine "$(brew --prefix)/bin/fly"
```

This (or something functionally equivalent) should be added to the cask's recipe.

#### Command that failed

```
brew install --cask fly

# these will trigger an unidentified developer warning; see image below
fly
fly --version
```

<img width="372" alt="Screen Shot 2021-01-20 at 6 01 39 PM" src="https://user-images.githubusercontent.com/385462/105252486-81020480-5b4b-11eb-9961-81ec5139c3fd.png">

#### Output of command with `--verbose --debug`

```
$ brew install --cask fly --verbose --debug
==> Cask::Installer#install
==> Printing caveats
==> Cask::Installer#fetch
==> Downloading https://github.com/concourse/concourse/releases/download/v6.7.2/fly-6.7.2-darwin-amd64.tgz
/usr/bin/curl --disable --globoff --show-error --user-agent Homebrew/2.7.5-10-g8869208\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 11.1\)\ curl/7.64.1 --header Accept-Language:\ en --retry 3 --location --silent --head --request GET https://github.com/concourse/concourse/releases/download/v6.7.2/fly-6.7.2-darwin-amd64.tgz
[...]

PR #98525, opened 2021-01-22
fly: remove removal of quarantine

vitorgalvao:
We (HBC) are the ones who add the quarantine, so we should never be the ones removing it. That breaks expectations. Removing quarantine ourselves in one cask means users who rely on it being active have to check every cask installation to see if it’s there.

[We explain how to remove quarantine when needed](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/app_cant_be_opened.md) and have `--no-quarantine`, both of those need to be explicitly called by users thus indicating explicit consent. That’s the way it should be. Either that or we [remove quarantining altogether](https://github.com/Homebrew/homebrew-cask/issues/70798), but we should never say we do something and then do the reverse (bugs excluded).

Refs https://github.com/Homebrew/homebrew-cask/issues/98458.

cc @miccal @todd-a-jacobs.

### DIFF
--- a/Casks/fly.rb
+++ b/Casks/fly.rb
@@ -9,9 +9,4 @@ cask "fly" do
   homepage "https://github.com/concourse/concourse"
 
   binary "fly"
-
-  preflight do
-    system_command "xattr",
-                   args: ["-dr", "com.apple.quarantine", "#{staged_path}/fly"]
-  end
 end


todd-a-jacobs:
Requiring this to b e done through the GUI is very much an anti-pattern. While I respect the decision to avoid automatically untainting it, the xattr instructions should be provided as caveats during install, rather than forcing the user to use the GUI.

My problem isn't with the security considerations here, it's with enforcing a multi-click, GUI approach to solving a known issue with a specific binary when a command-line alternative _with no root privileges required_ can be presented to the user during cask installation.

vitorgalvao:
> Requiring this to be done through the GUI is very much an anti-pattern.

Nobody is requiring you to go through the GUI. I’ve pointed out above how to it at install time: `brew install --no-quarantine fly`.

todd-a-jacobs:
> > Requiring this to be done through the GUI is very much an anti-pattern.
> 
> Nobody is requiring you to go through the GUI. I’ve pointed out above how to it at install time: `brew install --no-quarantine fly`.

The point is that it's non-intuitive for *users* who may not realize that they need to install it with different arguments. That's why I'm suggesting putting the solution into the caveats displayed to the user, rather than being something they need to know *before* attempting the installation.

vitorgalvao:
> The point is that it's non-intuitive for _users_ who may not realize that they need to install it with different arguments.

Which is the case for literally every cask which isn’t signed. There’s nothing special about this one. You’re arguing for it because you use it.

miccal:
Apologies for all the trouble I caused with this @vitorgalvao.

vitorgalvao:
> Apologies for all the trouble I caused with this @vitorgalvao.

Not at all! It’s two minor casks; no big deal in the slightest. By now I’ve said it multiple times and it continues to hold: in my book, you’re a fantastic maintainer.

PR #98527, opened 2021-01-22
kieler.rb: remove removal of quarantine

vitorgalvao:
Same as https://github.com/Homebrew/homebrew-cask/pull/98525.

cc @miccal @xaaronc.

### DIFF
--- a/Casks/kieler.rb
+++ b/Casks/kieler.rb
@@ -14,4 +14,13 @@ cask "kieler" do
     "~/Library/Preferences/de.cau.cs.kieler.product.plist",
     "~/Library/Saved Application State/de.cau.cs.kieler.product.savedState",
   ]
+
+  caveats <<~EOS
+    According to the vendor (https://rtsys.informatik.uni-kiel.de/confluence/display/KIELER/Downloads#Downloads-StandaloneApplications),
+    the quarantine attribute breaks the app and needs to be removed. To do so, install this cask with:
+
+      brew install --cask --no-quarantine #{token}
+
+    WARNING: quarantining is a macOS security feature and you disable it at your own risk!
+  EOS
 end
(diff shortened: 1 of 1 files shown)


xaaronc:
The case is slightly different here. The app is unusable when the quarantine attribute is set, presumably because something is up with the signature block (I don’t know for sure, I’m just a user of this app). In particular, the instructions documented at:
  https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/app_cant_be_opened.md
do not work.

If the policy is to leave the app in the quarantine no matter what, we should at least drop a post-install message with instructions to the user, since there’s no way to fix this in the UI.

vitorgalvao:
> The case is slightly different here.

Nevertheless, blindly disabling a security feature a user will expect to be there isn’t an acceptable solution; especially on software that isn’t widely recognisable.

> The app is unusable when the quarantine attribute is set

Then it can be installed with `--no-quarantine`, [as per the updated instructions](https://github.com/Homebrew/homebrew-cask/pull/98528).

But that’s also a bug that needs to be solved upstream. Can you find official documentation that points out that the GUI method of right-click and open doesn’t work and the quarantine needs to be explicitly removed (or not added)? If you do, we can add a `caveats` to this.

miccal:
https://github.com/Homebrew/homebrew-cask/pull/98525#issuecomment-765705733

xaaronc:
What’s official documentation got to do with it? Documented or not, the method doesn’t work, and it’s not an "unidentified developer" warning in any case.

vitorgalvao:
> What’s official documentation got to do with it?

From the vendor of the app. Official documentation we can link to so whoever comes across the cask later has context.

xaaronc:
Apologies, I misinterpreted your request to mean Apple documentation.
Looks like you found it in any case, caveats looks good to me. Thanks!

PR #98725, opened 2021-01-26
VS Code 1.54 for ARM64

ItalyPaleAle:
VS Code 1.54 is coming out soon and will include support for Apple Silicon.  There will be 3 bundles published:

- the old URL (package "darwin") will remain for Intel only
- "darwin-arm64" will be for AMR64 only (currently can be tried for insiders here: https://update.code.visualstudio.com/1.54.0-insider/darwin/insider )
- "darwin-universal" will be a universal build (currently can be tried for insiders here: https://update.code.visualstudio.com/1.54.0-insider/darwin-universal/insider )

While the Universal build will be the default on the website when users download that, it is significantly larger than the arch-specific builds because of Electron (they'll also use much more disk space after installation). With a tool like Homebrew that is capable of automatically detecting the architecture, our recommendation would be to pick the architecture-specific build here when possible.

---


ItalyPaleAle:
Update: because of issues with Node.js and Big Sur 11.2 on ARM, VS Code will not ship a stable for ARM64 in 1.53 and will be happening on the next release instead. I'll update this PR in early March.

brainsik (commented):
AFAICT, the `livecheck` block also needs to get updated to use the arch specific URL.

ItalyPaleAle:
@brainsik is that a requirement of Homebrew?

From the point of view of VS Code, we won't update an arch-specific build without updating the other ones too. Or, at least without updating the universal one. So, maybe I could switch the livecheck to use the universal build?

ItalyPaleAle:
Hey @miccal please don't merge this yet. Even though the binaries might be available, we're not making VS Code for ARM64 available in the Stable channel until the next release because of an [issue](https://github.com/microsoft/vscode/issues/115646) with Big Sur 11.2 and Node.js.

Would appreciate if you could wait till the 1.54 release so we are best equipped to support Apple Silicon.

CC: @deepak1556

vitorgalvao:
@ItalyPaleAle Please ping either me or @miccal when ready.

ItalyPaleAle:
@vitorgalvao @miccal the release (now 1.54) should be coming out in the next few hours. I have updated this PR with what we believe will be the hashes of the new build. I'll ping you once the release is officially out, but for now marking this ready for review :)

ItalyPaleAle:
Build is out, so triggered the CI again. SHould work this time.

CC: @miccal @vitorgalvao

thecesrom:
Hello, @ItalyPaleAle.

Since there are three types of installers:

- Universal (`appcast`: https://update.code.visualstudio.com/api/update/darwin-universal/stable/VERSION)
- Intel Chip (`appcast`: https://update.code.visualstudio.com/api/update/darwin/stable/VERSION)
- Apple Silicon (`appcast`: https://update.code.visualstudio.com/api/update/darwin-arm64/stable/VERSION)

Would it be advisable to use the Universal installer for M1 and Intel or continue checking `Hardware::CPU.intel`?

I have installed the Universal kind on my Intel Mac, and it works just like the 64bit, but I don't have any M1 for testing.

Thanks!

ItalyPaleAle:
(Note that 1.54.0 got pulled shortly after the release for a regression - 1.54.1 should be released very shortly - I'll update this PR after)

@thecesrom The Universal installer is 2x the size because of the way Universal Electron apps are packaged today. This will continue to be the case for the foreseeable future. My recommendation would be that, for a tool like brew that has the ability to automatically detect the arch, it should install arch-specific builds, to save bandwidth and disk space. But I'm happy to cancel this PR and switch to Universal for everyone if that's the maintainers' preference.

thecesrom:
> The Universal installer is 2x the size

I did notice that.

VSCode-darwin.zip = 99,570 KB
VSCode-darwin-arm64.zip = 99,058 KB
VSCode-darwin-universal.zip = 166,947 KB

I have **manually** bumped `visual-studio-code-insiders` since they started offering Intel and M1 installers (no Universal yet), and in my experience `brew bum-cask-pr` does not update both `sha256`. See discussion here https://github.com/Homebrew/discussions/discussions/561.

Taking that into account I consider that going with the Universal installer will make it easier to maintain, since users will still be able to bump this Cask via `brew bump-cask-pr`, something that is not possible when having an installer per Chip (as far as I know).

ItalyPaleAle:
Shouldn't this be something that needs to be fixed in bump-cask-pr though? I don't think this is an issue with VS Code only.

In any case, happy to close this PR and just open one to update visual-studio-code to use the universal build rather than the arch-specific one if you prefer.

Also, sorry again this release has been so complicated :(

ItalyPaleAle:
Shouldn't this be something that needs to be fixed in bump-cask-pr though? I don't think this is an issue with VS Code only. Many Universal apps are 2x the size and I wanted to be mindful of people around the world who don't have fast internet (or who have ridiculous data caps like my _lovely_ new ISP)

In any case, happy to close this PR and just open one to update visual-studio-code to use the universal build rather than the arch-specific one if you prefer.

Also, sorry again this release has been so complicated :(

thecesrom:
> Shouldn't this be something that needs to be fixed in bump-cask-pr though?

I would think that the Homebrew team should consider that, yes. It would make it easier to `bump` Casks with one simple command.

> I wanted to be mindful of people around the world who don't have fast internet (or who have ridiculous data caps like my lovely new ISP)

Definitely a fair point.

Stable versions are not updated as often as "nightlies' or "insiders", so maybe sticking with both `darwin` and `darwin-arm64` might be better. So manually bumping wouldn't be as painful as downloading ~2x ZIP installers.

ItalyPaleAle:
Ok finally 1.54.1 is out, and CI is green!

miccal:
Thank you @ItalyPaleAle.

miccal (approved):

PR #99388, opened 2021-02-05
Update mathpix-snipping-tool

singularitti:


singularitti:
Why change back to no-hash while there is an explicit link to download (like https://mathpix.com/dmg/SnippingTool-v3.0.0.201.dmg)?

miccal:
> Why change back to no-hash while there is an explicit link to download (like https://mathpix.com/dmg/SnippingTool-v3.0.0.201.dmg)? @miccal

We have experienced issued with the versioned `url` for this Cask in the past, and this way it also matches the `livecheck` results.

PR #99389, opened 2021-02-05
Set safe permissions when ngrok is installed

gedigi:

`ngrok` is installed as a world-writeable file, which is insecure. On a shared system, a malicious unprivileged user could replace `ngrok` with malware, which could be then executed by other users.

### DIFF
--- a/Casks/ngrok.rb
+++ b/Casks/ngrok.rb
@@ -11,4 +11,9 @@ cask "ngrok" do
   binary "ngrok"
 
   zap trash: "~/.ngrok2"
+
+
+  postflight do
+    FileUtils.chmod 0755, Dir.glob("#{staged_path}/ngrok")
+  end
 end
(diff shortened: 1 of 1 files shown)


Casks/ngrok.rb
```diff
@@ -10,5 +10,9 @@
 
   binary "ngrok"
 
+  postflight do
+    FileUtils.chmod 0755, Dir.glob("#{staged_path}/ngrok")
```
vitorgalvao, line 14: You’re looking for [`set_permissions`](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/flight.md#flight-mini-dsl) instead.

vitorgalvao:
Thank you, but this looks like an upstream bug. It should be fixed there instead, where more people will benefit and we won’t need a workaround. Have you contacted them?

gedigi:
by upstream you mean ngrok? I installed/downloaded the file manually and it sets the correct permissions (0755). See https://asciinema.org/a/bTYfjjcitUdhalWlRk075BQ8U

Is it possible that Homebrew ignores umask?

vitorgalvao (approved):

PR #99419, opened 2021-02-06
Rename google-drive-file-stream to google-drive

vitorgalvao:
Can anyone confirm with this changes pulled that `brew install google-drive-file-stream` or `brew upgrade google-drive-file-stream` results in an installation of `google-drive` and no `google-drive-file-stream`?

### DIFF
--- a/Casks/google-drive-file-stream.rb
+++ b/Casks/google-drive-file-stream.rb
@@ -10,9 +10,9 @@ cask "google-drive-file-stream" do
   stage_only true
 
   postflight do
-    system_commad "brew",
+    system_command HOMEBREW_BREW_FILE,
       args: ["install", "google-drive"]
-    system_commad "brew",
+    system_command HOMEBREW_BREW_FILE,
       args: ["uninstall", "google-drive-file-stream"]
   end
 
--- a/Casks/google-drive-file-stream.rb
+++ b/Casks/google-drive-file-stream.rb
@@ -7,6 +7,7 @@ cask "google-drive-file-stream" do
   homepage "https://www.google.com/drive/"
 
   depends_on macos: ">= :el_capitan"
+
   stage_only true
 
   postflight do
--- a/Casks/google-drive-file-stream.rb
+++ b/Casks/google-drive-file-stream.rb
@@ -1,5 +1,5 @@
 cask "google-drive-file-stream" do
-  version "44.0.14.1" # So it triggers an update for users of the cask
+  version "44.0.14.1" # So it triggers an upgrade for users of the cask
   sha256 :no_check
 
   url "https://dl.google.com/drive-file-stream/GoogleDriveFileStream.dmg"
--- a/Casks/google-drive-file-stream.rb
+++ b/Casks/google-drive-file-stream.rb
@@ -7,26 +7,21 @@ cask "google-drive-file-stream" do
   desc "Client for the Google Drive storage service"
   homepage "https://www.google.com/drive/"
 
-  depends_on macos: ">= :el_capitan"
+  depends_on cask: "google-drive"
 
   stage_only true
-
-  postflight do
-    system_command HOMEBREW_BREW_FILE,
-                   args: ["install", "google-drive"]
-    system_command HOMEBREW_BREW_FILE,
-                   args: ["uninstall", "google-drive-file-stream"]
-    FileUtils.rm_r(staged_path.to_s)
-  end
 
   caveats <<~EOS
     RENAME WARNING
 
-    Google Drive File Stream has been renamed to Google Drive.
-
-    `#{cask}` has been renamed `google-drive`. We will uninstall the former and install the latter. but you should update your scripts. `#{cask}` will be deleted on August 1, 2021.
+    `#{cask}` will be renamed `google-drive`.
+    In the meantime, `#{cask}` will install `google-drive` for you as a dependency, but you should update your scripts.
 
     We’re aware this solution is subpar. If you’d like to help us improve it,
     we accept PRs and need the equivalent of formula_renames.json for casks: https://docs.brew.sh/Rename-A-Formula
+
+    To migrate now, do:
+      brew uninstall #{cask}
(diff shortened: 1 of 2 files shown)


Casks/google-drive-file-stream.rb
```diff
@@ -1,40 +1,29 @@
...
   url "https://dl.google.com/drive-file-stream/GoogleDriveFileStream.dmg"
   name "Google Drive File Stream"
   homepage "https://www.google.com/drive/"
 
   depends_on macos: ">= :el_capitan"
+  stage_only true
 
-  pkg "GoogleDriveFileStream.pkg"
+  postflight do
+    system_commad "brew",
```
bevanjkay, line 13: @vitorgalvao There is a typo here it should be `system_command`

I also only had success changing the lines to:
`system_command(HOMEBREW_BREW_FILE, args:["install", "google-drive"], print_stdout: true)
    system_command(HOMEBREW_BREW_FILE, args:["uninstall", "google-drive-file-stream"], print_stdout: true)`

and even then the installer requires a password input to install which I'm not getting prompted for.

Casks/google-drive.rb
```diff
@@ -0,0 +1,41 @@
+cask "google-drive-file-stream" do
```
bevanjkay, line 1: @vitorgalvao 
The install is failing because this should be "google-drive"

Casks/google-drive.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  url "https://dl.google.com/drive-file-stream/GoogleDriveFileStream.dmg"
+  name "Google Drive"
+  homepage "https://www.google.com/drive/"
+
+  depends_on macos: ">= :el_capitan"
+
+  pkg "GoogleDrive.pkg"
+
+  uninstall login_item: "Google Drive File Stream",
```
bevanjkay, line 13: @vitorgalvao This can be changed to "Google Drive" as the login_item has changed

vitorgalvao:
Thank you @bevanjkay. Round 2.

miccal:
Fixed the `audit` and `style` errors, but I am not sure about the `uninstall` errors.

bevanjkay:
@miccal @vitorgalvao 
On both my machines, the upgrade command runs without errors, but after running `brew ls` includes both google-drive and google-drive-file-stream.

And running `brew uninstall google-drive-file-stream` returns `Error: Cask 'google-drive-file-stream' is not installed.`

miccal:
@vitorgalvao maybe add a `rm -rf "$(brew --prefix)/Caskroom/google-drive-file-stream`?

vitorgalvao:
Round 3.

Casks/google-drive-file-stream.rb
```diff
@@ -16,6 +16,7 @@
                    args: ["install", "google-drive"]
     system_command HOMEBREW_BREW_FILE,
                    args: ["uninstall", "google-drive-file-stream"]
+    FileUtils.rm_r(staged_path.to_s)
```
bevanjkay, line 19: @vitorgalvao I am now getting an error as it appears this line is trying to remove the versioned staging folder for the Cask. Changing to `FileUtils.rm_r(FIle.expand_path("..", staged_path.to_s))` targets the folder and removes it.

This results in the `brew ls` showing that only google-drive is installed, and not google-drive-file-stream.

**Caveat**: This is resulting in a new error:
`Error: google-drive-file-stream: Could not find PKG source file 'GoogleDrive.pkg'.`

This is because the `brew upgrade google-drive-file-stream` command uses a cached version of the ruby file. So it fails to upgrade and at the end tries to reinstall the original google-drive-file-stream package (which we deleted with the remove command above).

vitorgalvao: It’s getting unwieldy. Perhaps we should consider another path. Any suggestions? As a last resort we have what we did with `zoomus` → `zoom`.

bevanjkay: @vitorgalvao That may well be a better interim solution until there is an equivalent of the formula_renames strategy.

vitorgalvao:
Ugly, but it should work.

Casks/google-drive-file-stream.rb
```diff
@@ -1,40 +1,27 @@
...
   url "https://dl.google.com/drive-file-stream/GoogleDriveFileStream.dmg"
   name "Google Drive File Stream"
+  desc "Client for the Google Drive storage service"
   homepage "https://www.google.com/drive/"
 
-  depends_on macos: ">= :el_capitan"
+  depends_on cask: "google-drive"
 
-  pkg "GoogleDriveFileStream.pkg"
+  stage_only true
```
bevanjkay, line 13: @vitorgalvao The only recommendation I would make, depending on how it sits with you would be to retain 

`uninstall login_item: "Google Drive File Stream"`

As the login item has changed name with the application.

Otherwise, the login item may remain in system preferences broken. In saying this, I believe even Google's auto update of the app may be causing the broken login item to remain also.

vitorgalvao: Good point.

Casks/google-drive-file-stream.rb
```diff
@@ -1,40 +1,29 @@
 cask "google-drive-file-stream" do
-  version "44.0.14"
+  version "44.0.14.1" # So it triggers an upgrade for users of the cask
   sha256 :no_check
 
   url "https://dl.google.com/drive-file-stream/GoogleDriveFileStream.dmg"
```
bevanjkay, line 5: There is a new download URL also as follows
`https://dl.google.com/drive-file-stream/GoogleDrive.dmg`
See https://support.google.com/a/answer/7491144?hl=en#zippy=%2Cmac

vitorgalvao:
Everyone fine with merging as is?

miccal (approved):

vitorgalvao:
Thank you for all the help, @bevanjkay!

PR #99447, opened 2021-02-06
Update url.md

marble-sh:
- [x] `brew audit` is error-free.
- [x] `brew style` reports no offenses.

Sister pull request to https://github.com/Homebrew/brew/pull/10548

### DIFF
--- a/doc/cask_language_reference/stanzas/url.md
+++ b/doc/cask_language_reference/stanzas/url.md
@@ -116,4 +116,11 @@ Example for using the block syntax: [vlc-nightly.rb](https://github.com/Homebrew
 
 In rare cases, you might need to set URL parameters like `cookies` or `referer` while also using the block syntax.
 
-This is possible by returning a two-element array as a block result. The first element of the array must be the download URL; the second element must be a `Hash` containing the parameters.
+This is possible by passing the options as part of the initial block syntax.
+
+
+```rb
+url referer: "https://example.com", using: MyDownloadStrategy do
+  ...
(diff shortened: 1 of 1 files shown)


doc/cask_language_reference/stanzas/url.md
```diff
@@ -116,4 +116,11 @@ Example for using the block syntax: [vlc-nightly.rb](https://github.com/Homebrew
 
 In rare cases, you might need to set URL parameters like `cookies` or `referer` while also using the block syntax.
 
-This is possible by returning a two-element array as a block result. The first element of the array must be the download URL; the second element must be a `Hash` containing the parameters.
+This is possible by passing the options as part of the initial block syntax.
+
+
+```rb
+url referer: "https://example.com", using: MyDownloadStrategy do
+  ...
```
vitorgalvao, line 124: We’d rather have a real example. We don’t do the ellipses on examples anywhere else. Also, `MyDownloadStrategy` will be confusing. It’s best to have one specific real example people can adapt than an example that is too generic.

marblenix: Would it be too much to use my own repository as an example? As far as I know I'm the only one who uses both `url do` blocks and download options.

vitorgalvao: > Would it be too much to use my own repository as an example?

Go ahead!

marblenix:
closing due to alternate fix in original pull request

PR #99908, opened 2021-02-15
Update AlDente from 1.2 to 2.0

lpla:
Adds support for Apple Silicon (M1 MacBooks)

### DIFF
--- a/Casks/aldente.rb
+++ b/Casks/aldente.rb
@@ -1,7 +1,8 @@
 cask "aldente" do
-  version "v2.0-alpha"
+  version "2.0"
   sha256 "03859d6ca653b59bb32b5ddff1806ec07c24523f871d941fadf940f27ac8139e"
-  url "https://github.com/davidwernhart/AlDente/releases/download/#{version}/AlDente_2.0Alpha.app.zip"
+
+  url "https://github.com/davidwernhart/AlDente/releases/download/v#{version}-alpha/AlDente_#{version}Alpha.app.zip"
   appcast "https://github.com/davidwernhart/AlDente/releases.atom"
   name "AlDente"
   desc "Menu bar tool to limit maximum charging percentage"
--- a/Casks/aldente.rb
+++ b/Casks/aldente.rb
@@ -1,8 +1,7 @@
 cask "aldente" do
-  version "1.2"
-  sha256 "a588dc29faca894b7321e23420ca17d6a944b9b3b46412435f519b96e4ebee7b"
-
-  url "https://github.com/davidwernhart/AlDente/releases/download/#{version}/AlDente.app.zip"
+  version "v2.0-alpha"
+  sha256 "03859d6ca653b59bb32b5ddff1806ec07c24523f871d941fadf940f27ac8139e"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> * [x]  The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

You sure? It’s tagged as stable on the releases but it has “alpha” right in the name.

lpla:
Devs didn't advertise it as alpha in the release page or the README.m. They simply added the 'alpha' string to the file. But in any case, looks like there are issues with the main improvement in this update, M1 Macbook support: https://github.com/davidwernhart/AlDente/issues/109

So maybe we should really treat this version as an alpha and dismiss this PR.

vitorgalvao:
> But in any case, looks like there are issues with the main improvement in this update, M1 Macbook support: davidwernhart/AlDente#109
>
> So maybe we should really treat this version as an alpha and dismiss this PR.

Given that, let’s close. Thank you for checking.

On another note, why use an external app instead of relying on macOS’ own “Optimised battery charging”, on by default?

lpla:
There is a whole topic about this, regarding battery aging through calendar if kept at static charge or cycles by use, but mainly, macOS AI-driven feature keeps battery on 90% until it "predicts" your next battery use, which, while better than keeping it at 100%, is not good enough for batteries in case that you keep your computer on charger, let's say, all working days. Check this [use case](https://github.com/davidwernhart/AlDente/issues/52#issuecomment-779292359):

"[...] during the week my work laptop is on charger exclusively. During the weekend I often drive to our cottage where I use the laptop mostly without charger. If on Sunday evening the laptop sits at 90% then discharging to 30% to avoid at least 5 days of calendar aging at such a high charge level is definitely worth it.

What you don't want to do is operate the laptop between 20% and 35% and each night discharge down to 20% to get a bit less calendar aging. At these levels cycle aging is much more expensive than calendar aging. What does make sense is to leave the laptop at 20% if it happens to be there in the evening and only charge to 35% in the morning. Of course you have to do that manually - at least currently 🙂..."

There are several studies regarding Li-ion batteries aging (most of them listed in [Battery University](https://batteryuniversity.com/learn/article/how_to_charge_when_to_charge_table)) which point towards keeping batteries between 75–25% under nice temperatures for the longest life without compromising half battery usage, as
- high temperatures in full batteries affect as "calendar aging" (that's why manufacturers, including Apple, recommend keeping your batteries at 50% for long term storage)
- and full cycles (from 100% to 0%) affect as "cycle aging".

This topic is very interesting for electric cars and satellites, as they use the same technology. Check more technical reports about prolonging battery life on Battery University, specially text around Figure 6: https://batteryuniversity.com/learn/article/how_to_prolong_lithium_based_batteries

davidwernhart:
Hey guys! I released a new version 2 days ago, which fixed a few bugs and is not in PR state anymore. Is it possible to incorporate this release into the brew formula?
David

davidwernhart:
https://github.com/Homebrew/homebrew-cask/pull/101047
I created a new pull request with the updated cask.

PR #99944, opened 2021-02-16
Changed logisim-evolution to use .dmg

kqct:

I've updated this cask to use the Mac-specific `.app/.dmg`, instead of using the Java universal `.jar`. Because of quirks in how the developer packages the disk image, I have to run `xattr -cr` on the application in order for it to launch properly. In addition, I've removed the Java 9 dependency, since the app bundle ships with a compatible version of Java. Finally, I wasn't sure if the application auto-updates; the description states that it does, but I'm not sure if the functionality is built-in or if it just redirects to the GitHub project, so I left it out for now. Thanks for reviewing, and I hope there aren't any issues!

### DIFF
--- a/Casks/logisim-evolution.rb
+++ b/Casks/logisim-evolution.rb
@@ -1,20 +1,21 @@
 cask "logisim-evolution" do
   version "3.4.1"
-  sha256 "1b72dd3397290b7be95f886b933bd37a2d9182b8d33247fa5f059d0867e8111c"
+  sha256 "253290b94ae1dba6d5528ff2ea478c2750cee4a40c4710f7da6006dc3f0b4b52"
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> Because of quirks in how the developer packages the disk image, I have to run `xattr -cr` on the application in order for it to launch properly.

Specifically what? This would remove quarantine, which we do not want.

meow10811:
That's a good question - I can look into it and try to figure out why.

meow10811:
it looks like there's a codesigning issue - I might have to submit an issue on their end to fix it. Unless a better solution exists than removing quarantine, this might not be worth merging for now.

vitorgalvao:
> I might have to submit an issue on their end to fix it.

Please do. We can revisit this at any time. Thank you.

PR #100248, opened 2021-02-22
Improved Camo Studio cask

aydenp:

---

Disclosure: I work on Camo Studio at Reincubate.

This PR improves the `camo-studio` cask by adding full install/uninstall support (which will properly setup the A/V plugins), bumping Camo Studio to today's 1.2.4 release, adding a description, and replacing the `appcast` stanza with `livecheck`.

On uninstall: there is a temporary workaround in this PR which is setting `must_succeed` on the uninstall script to `false`, as we sometimes exit with code 9 after a successful uninstall. This should be resolved in a future Camo Studio release and I will update the cask when that happens, but we figure it is better than the current revision's incomplete uninstall.

### DIFF
--- a/Casks/camo-studio.rb
+++ b/Casks/camo-studio.rb
@@ -1,6 +1,6 @@
 cask "camo-studio" do
-  version "1.2.3,139"
-  sha256 "dc5983b84f2a0b5500fa44e59acf744228cf03d3e10497b05ba09944d6acfa08"
+  version "1.2.4,144"
+  sha256 "37aea984109f4c8bf6f007a8bd8cb55de2e745773684c8cca3d736b421023ce0"
 
   url "https://reincubate.com/res/labs/camo/Camo%20Studio%20#{version.before_comma}%20%28#{version.after_comma}%29%20%5BRelease%5D.app.zip"
   appcast "https://uds.reincubate.com/release-notes/camo/?format=sparkle"
--- a/Casks/camo-studio.rb
+++ b/Casks/camo-studio.rb
@@ -3,10 +3,14 @@ cask "camo-studio" do
   sha256 "37aea984109f4c8bf6f007a8bd8cb55de2e745773684c8cca3d736b421023ce0"
 
   url "https://reincubate.com/res/labs/camo/Camo%20Studio%20#{version.before_comma}%20%28#{version.after_comma}%29%20%5BRelease%5D.app.zip"
-  appcast "https://uds.reincubate.com/release-notes/camo/?format=sparkle"
   name "Camo Studio"
   desc "Use your phone as a pro webcam to look amazing on video calls"
   homepage "https://reincubate.com/camo/"
+
+  livecheck do
+    url "https://uds.reincubate.com/release-notes/camo/?format=sparkle"
+    strategy :sparkle
+  end
 
   auto_updates true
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you! I do have some questions, which I’ll make in the file.

Casks/camo-studio.rb
```diff
@@ -1,15 +1,32 @@
...
 
+  livecheck do
+    url "https://uds.reincubate.com/release-notes/camo/?format=sparkle"
+    strategy :sparkle
+  end
+
   auto_updates true
 
   app "Camo Studio.app"
+  installer script: {
```
vitorgalvao, line 18: `installer` *and* `app`? Shouldn’t it be one *or* the other?

Casks/camo-studio.rb
```diff
@@ -1,15 +1,32 @@
...
+    must_succeed: false,
+  }
 
-  uninstall delete: "/Library/CoreMediaIO/Plug-Ins/DAL/ReincubateCamoDAL.plugin"
+  uninstall quit:   "com.reincubate.macos.cam",
+            script: {
+              executable:   "/Applications/Camo Studio.app/Contents/MacOS/Camo Studio",
+              args:         ["-uninstall"],
+              sudo:         true,
+              must_succeed: false, # necessary for now as sometimes will exit(9) even on success
```
vitorgalvao, line 30: Please add a link back to this PR so users can see the reason.

Casks/camo-studio.rb
```diff
@@ -1,15 +1,32 @@
 cask "camo-studio" do
-  version "1.2.3,139"
-  sha256 "dc5983b84f2a0b5500fa44e59acf744228cf03d3e10497b05ba09944d6acfa08"
+  version "1.2.4,144"
+  sha256 "37aea984109f4c8bf6f007a8bd8cb55de2e745773684c8cca3d736b421023ce0"
 
   url "https://reincubate.com/res/labs/camo/Camo%20Studio%20#{version.before_comma}%20%28#{version.after_comma}%29%20%5BRelease%5D.app.zip"
-  appcast "https://uds.reincubate.com/release-notes/camo/?format=sparkle"
   name "Camo Studio"
+  desc "Use your phone as a pro webcam to look amazing on video calls"
```
vitorgalvao, line 7:
```suggestion
  desc "Use your phone as a webcam"
```

The extra text is what we describe as “marketing fluff” (no offence!). Descriptions should be brief in telling *objectively* what the app does or is for.

Casks/camo-studio.rb
aydenp: In this case, both are necessary. The app is normally self-installing upon launch. Via the CLI (initially intended for sysadmins), the app only installs its A/V plugins, but won't move itself, to allow placement anywhere.

Casks/camo-studio.rb
aydenp: That makes sense. Was referencing our public copy. 

The last fragment may indeed be extraneous fluff, but I would consider 'pro' as useful; Camo Studio offers many controls that go beyond what the proposed would describe. Perhaps there's a better, more precise description for this field without the fluff.

vitorgalvao: What makes it a “pro” webcam? What’s the most important feature of those?

aydenp: I've just updated the description with a short list of some of our most important features, to avoid any potential subjective adjectives.

vitorgalvao (approved):

PR #100317, opened 2021-02-24
fix source tree formula for macOS bigsur

adelwin:

### DIFF
--- a/Casks/sourcetree.rb
+++ b/Casks/sourcetree.rb
@@ -11,6 +11,12 @@ cask "sourcetree" do
 
     url "https://product-downloads.atlassian.com/software/sourcetree/ga/Sourcetree_#{version.before_comma}_#{version.after_comma}.zip",
         verified: "atlassian.com/software/sourcetree/"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
That’s what the `else` section is doing.

adelwin:
Sorry, i didn't realize there's questions here.
I noticed the main trunk is already updated with another commit.
I'd like to weigh in on that if/else branching tho.
We shouldn't just make an assumption that anything not "sierra" and not "high_sierra" is definitely must be "big_sur"
I would actually argue that we should fail the installation if it falls under "else" block, because then it's an unknown os version.

vitorgalvao:
> We shouldn't just make an assumption that anything not "sierra" and not "high_sierra" is definitely must be "big_sur"

That’s not what the conditional says. It’s `<=`, not `=`.

> I would actually argue that we should fail the installation if it falls under "else" block, because then it's an unknown os version.

No, the version will be very well known as “anything above the previous one mentioned in the conditional”.

We’ve been using this method in tons of casks for several years. It works well.

PR #100494, opened 2021-03-01
Delete pins.rb

mindfulmonk:
No longer available. Project moved to https://get-pins.app/ and is only available from app store which means `mas`


miccal:
Analytics:
```
==> Analytics
install: 2 (30 days), 6 (90 days), 8 (365 days)
```
Thank you @mindfulmonk.

miccal (approved):

PR #100509, opened 2021-03-02
Add MailTrackerBlocker v0.3.16

apparition47:

### DIFF
--- a/Casks/mailtrackerblocker.rb
+++ b/Casks/mailtrackerblocker.rb
@@ -16,12 +16,4 @@ cask "mailtrackerblocker" do
   uninstall pkgutil: "com.onefatgiraffe.mailtrackerblocker",
             delete:  "/Library/Mail/Bundles/MailTrackerBlocker.mailbundle",
             quit:    ["com.apple.mail"]
-
-  def caveats
-    <<~EOS
-      To enable:
-        Open Mail, goto Preferences > General > Manage Plug-ins... >
-        check "MailTrackerBlocker.mailbundle" > Apply and Restart Mail.
-    EOS
-  end
 end
--- a/Casks/mailtrackerblocker.rb
+++ b/Casks/mailtrackerblocker.rb
@@ -1,6 +1,6 @@
 cask "mailtrackerblocker" do
-  version "0.3.15"
-  sha256 "f2ba1b0a35bb404ca306b84604ee4998b855a33e456d74414d37607ee2177253"
+  version "0.3.16"
+  sha256 "feb3f217a6bf08423c7bcd2b4e97113384e1314123e37f57e2981a4b4d3cda46"
 
   url "https://github.com/apparition47/MailTrackerBlocker/releases/download/#{version}/MailTrackerBlocker.pkg",
       verified: "github.com/apparition47/MailTrackerBlocker/"
@@ -19,15 +19,9 @@ cask "mailtrackerblocker" do
 
   def caveats
     <<~EOS
-      ℹ️  To enable and use (you'll need to do this again after updating):
-        1. Open Mail, goto Preferences > General > Manage Plug-ins... >
-           check "MailTrackerBlocker.mailbundle" > Apply and Restart Mail.
-        2. Tap on the ⓧ  button to find out what was blocked.
-
-      ⚠️  Note:
-        Disabling "load remote content in messages"
-        with MailTrackerBlocker enabled is redundant;
-        re-enable this option for the best experience.
+      To enable:
+        Open Mail, goto Preferences > General > Manage Plug-ins... >
+        check "MailTrackerBlocker.mailbundle" > Apply and Restart Mail.
     EOS
   end
 end
--- a/Casks/mailtrackerblocker.rb
+++ b/Casks/mailtrackerblocker.rb
@@ -1,0 +1,33 @@
+cask "mailtrackerblocker" do
+  version "0.3.15"
+  sha256 "f2ba1b0a35bb404ca306b84604ee4998b855a33e456d74414d37607ee2177253"
+
+  url "https://github.com/apparition47/MailTrackerBlocker/releases/download/#{version}/MailTrackerBlocker.pkg",
+      verified: "github.com/apparition47/MailTrackerBlocker/"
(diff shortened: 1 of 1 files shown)


miccal:
Thank you for your contribution to `homebrew-cask` @apparition47.

Could you please try to add a [`zap` stanza](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/zap.md), thanks.

apparition47:
Does the `zap` stanza run after the `uninstall` stanza? All my (only 1) of my files should already be covered in `uninstall`.

Also tried the suggested `createzap`:

```bash
$ brew createzap mailtrackerblocker
No mailtrackerblocker settings found.
```

Casks/mailtrackerblocker.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  depends_on macos: ">= :el_capitan"
+
+  pkg "MailTrackerBlocker.pkg"
+
+  uninstall pkgutil: "com.onefatgiraffe.mailtrackerblocker",
+            delete:  "/Library/Mail/Bundles/MailTrackerBlocker.mailbundle",
+            quit:    ["com.apple.mail"]
+
+  def caveats
```
vitorgalvao, line 20: From the [`caveats` documentation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md):

> To avoid flooding users with too many messages (thus desensitising them to the important ones), `caveats` should be used sparingly and exclusively for installation-related matters.

From the [Homebrew Cask is not a discoverability service FAQ document](https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Casks.md#homebrew-cask-is-not-a-discoverability-service)

> users are expected to have reasonable knowledge about the apps they’re installing through us before doing so.

Also, we don’t use emoji in cask caveats. Please remove this.

Casks/mailtrackerblocker.rb
```diff
@@ -0,0 +1,33 @@
...
+  desc "Email tracker, read receipt and spy pixel blocker plugin for Apple Mail"
+  homepage "https://apparition47.github.io/MailTrackerBlocker/"
+
+  depends_on macos: ">= :el_capitan"
+
+  pkg "MailTrackerBlocker.pkg"
+
+  uninstall pkgutil: "com.onefatgiraffe.mailtrackerblocker",
+            delete:  "/Library/Mail/Bundles/MailTrackerBlocker.mailbundle",
+            quit:    ["com.apple.mail"]
```
vitorgalvao, line 18: Is this *needed* for the uninstall to work? Not too comfortable closing another app without warning.

vitorgalvao (changes requested):

Casks/mailtrackerblocker.rb
apparition47: Sure, It won't prevent the uninstallation but it will create user confusion if someone decides **to uninstall with my plugin enabled while Mail is open**. The plugin will still be activated but with all the localized strings missing (see screenshot). As the user, I'd expect all remnants of this plugin to be gone upon uninstall and that doesn't happen until Mail is closed to allow the plugin to be finally released from memory.

Casks/mailtrackerblocker.rb
apparition47: Removed the caveat completely since it isn't "installation-related".

Casks/mailtrackerblocker.rb
vitorgalvao: Agreed it may cause some confusion, but that’s better than the alternative of closing an another application from under the user’s feet, which may have consequences we aren’t envisioning.

We could check at uninstall time if Mail.app is running, and if yes, print a message (untested code):

```ruby
uninstall_postflight do
  if system_command("ps", args: ["x"]).stdout.match?("Mail.app/Contents/MacOS/Mail")
    opoo "Restart Mail.app to finish uninstalling #{token}"
  end
end
```

Casks/mailtrackerblocker.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  uninstall_postflight do
+    if system_command("ps", args: ["x"]).stdout.match?("Mail.app/Contents/MacOS/Mail")
+      opoo "Restart Mail.app to finish uninstalling #{token}"
+    end
+  end
+
+  uninstall pkgutil: "com.onefatgiraffe.mailtrackerblocker",
+            delete:  "/Library/Mail/Bundles/MailTrackerBlocker.mailbundle",
+            quit:    "com.apple.mail"
```
vitorgalvao, line 23: The goal of the `uninstall_postflight` is so we can remove this.

Casks/mailtrackerblocker.rb
```diff
@@ -0,0 +1,24 @@
...
+      verified: "github.com/apparition47/MailTrackerBlocker/"
+  name "MailTrackerBlocker"
+  desc "Email tracker, read receipt and spy pixel blocker plugin for Apple Mail"
+  homepage "https://apparition47.github.io/MailTrackerBlocker/"
+
+  depends_on macos: ">= :el_capitan"
+
+  pkg "MailTrackerBlocker.pkg"
+
+  uninstall_postflight do
```
vitorgalvao, line 15: Is this working as is? Meaning, if Mail.app is is running and we `brew uninstall mailtrackerblocker`, it shows the message?

apparition47: Yes, seems to work

```bash
brew remove mailtrackerblocker
==> Uninstalling Cask mailtrackerblocker
==> Uninstalling packages:
com.onefatgiraffe.mailtrackerblocker
==> Removing files:
/Library/Mail/Bundles/MailTrackerBlocker.mailbundle
Warning: Restart Mail.app to finish uninstalling mailtrackerblocker
==> Purging files for version 0.3.16 of Cask mailtrackerblocker
```

vitorgalvao: Great! I asked because untested code that works on the first try is always suspicious.

Thank you for the submission, the fast changes, and for working on something that makes it harder to be tracked.

vitorgalvao (approved):

PR #100679, opened 2021-03-04
Update yandex-disk from 3.2.1,30 to 3.2.6

AlekSi:
"30" in the file name was removed some time ago.

### DIFF
--- a/Casks/yandex-disk.rb
+++ b/Casks/yandex-disk.rb
@@ -1,13 +1,11 @@
 cask "yandex-disk" do
-  version "3.2.1,30"
-  sha256 "cf69a3ff787ad970715c95aceaeb0e10de4664a8c97e60c314f3735eb44338ec"
+  version :latest
+  sha256 :no_check
 
-  url "https://disk.yandex.ru/download/YandexDisk#{version.after_comma}.dmg/?instant=1"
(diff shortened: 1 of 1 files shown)


vitorgalvao (approved):

Casks/yandex-disk.rb
```diff
@@ -1,8 +1,8 @@
 cask "yandex-disk" do
-  version "3.2.1,30"
-  sha256 "cf69a3ff787ad970715c95aceaeb0e10de4664a8c97e60c314f3735eb44338ec"
+  version "3.2.6"
```
AlekSi, line 2: Why specify a version if there is no version in the URL and no stable checksum?

AlekSi: @vitorgalvao ^

vitorgalvao: Because then it still works with `upgrade` (and we have systems to check for those versions). Though in this case it has `auto_updates`, so that’s moot, you’re right. I’ll revert that part.

AlekSi: Yeah, I was following https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/version.md#version-latest. Thanks for explanation.

PR #100776, opened 2021-03-06
Update Logisim-evolution to use app .dmg file

Mick235711:

The `logisim-evolution` cask currently use the universal built `.jar` file provided by the maintainer. However, starting from version 3.4.1, the maintainer had provided prebuilt binary for Windows, Linux and macOS. Therefore we can now directly use the `.dmg` file, since it directly produce a `.app` file that can be directly run and placed on the dock.

However, the 3.4.1 `.dmg` file will report `Application “Logisim-evolution.app” is damaged and can’t be opened. You should move it to the Trash.`. This is not due to the dmg file damaged, but due to the fact that it is improperly codesigned. The maintainer decided in [#505](https://github.com/reds-heig/logisim-evolution/issues/505) that in the build process, the codesign will be removed, and resulting in an app file that can be open under `Anywhere` security checks (since it is now unsigned). This fix is already [merged](https://github.com/reds-heig/logisim-evolution/pull/512), and is already included in the pre-release 3.4.2. However, a new stable release containing this fix is not available by now, and therefore I used a `postflight` block to manually remove the codesigns (suggested in [#499](https://github.com/reds-heig/logisim-evolution/issues/499)). That block should be able to be safely deleted when a new stable version come out.

### DIFF
--- a/Casks/logisim-evolution.rb
+++ b/Casks/logisim-evolution.rb
@@ -1,16 +1,20 @@
 cask "logisim-evolution" do
   version "3.4.1"
-  sha256 "1b72dd3397290b7be95f886b933bd37a2d9182b8d33247fa5f059d0867e8111c"
+  sha256 "253290b94ae1dba6d5528ff2ea478c2750cee4a40c4710f7da6006dc3f0b4b52"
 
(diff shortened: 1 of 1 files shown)


Mick235711:
Sorry for duplicating

Issue #88183, opened 2020-08-26
google-cloud-sdk versioned and maybe formular


Mazorius:
### Description of feature/enhancement

Google has a versioned archive so that the cask does not need latest anymore — I think.
e.g.: https://storage.googleapis.com/cloud-sdk-release/google-cloud-sdk-307.0.0-darwin-x86_64.tar.gz

As mentioned in #62715 maybe a switch to formular is possible as it is a CLI?

### Justification

A update is much easier to see … and I find cask a bit strange for a CLI — hmmm.
Also it can be listed by `brew cask outdated` or `brew outdated` if no more a cask.

### Example use case

Checking output of outdated via script is much easier.

miccal:
Although the `google-cloud-sdk` is a CLI tool, as per the [Homebrew documentation](https://docs.brew.sh/Acceptable-Formulae#we-dont-like-binary-formulae) a formula must either build from source or produce cross-platform binaries. Binary-only tools belong in `homebrew-cask`.

Also, there is a note in the [cask file itself](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/google-cloud-sdk.rb#L2) explaining why it is un-versioned.

Mazorius:
Ok – so without cask is not possible. Thank you for the clarification!

I read the comment but the comment blame is the following.

```
fcb36135f6e (Ondřej Führer 2018-04-11 23:52:56 +0200
```

I think maybe this is a outdated comment?

If I know how to test this, I would do this in a PR … but I not really have an idea about that.

vitorgalvao:
You’re welcome to try to make this into a formula, but that’s the domain of https://github.com/Homebrew/homebrew-core; this repo is only concerned with casks.

If you can make it into a formula and have it merged in that repo, we’ll remove it from here.

PR #45818, opened 2018-04-11
Update Google Cloud SDK: add comment about versioning

ondrejfuhrer:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.

PR #101021, opened 2021-03-11
Version google-cloud-sdk

sparhomenko:

This is an attempt to version `google-cloud-sdk`. This has been tried before (#37317, #45818, #88183), but wasn't accepted for reasons described by @commitay in https://github.com/Homebrew/homebrew-cask/pull/37317#issuecomment-319856337, namely that each upgrade would:
1. require users to update the `$(brew --prefix)/Caskroom/google-cloud-sdk/<version>/google-cloud-sdk/` path in their profile;
2. remove manually installed components (as currently documented near `version`).

This PR proposes to address these concerns by:
1. using a static `#{HOMEBREW_PREFIX}/share/#{token}` path as in other SDKs ([Adobe Air](https://github.com/Homebrew/homebrew-cask/blob/3f07f816a5d8919e14c43fced157684cbf2b0ff7/Casks/adobe-air-sdk.rb), Android [SDK](https://github.com/Homebrew/homebrew-cask/blob/3b53a9694ffdb6e3af14dd3c56a3be8d37b0d721/Casks/android-sdk.rb) and [NDK](https://github.com/Homebrew/homebrew-cask/blob/81ff6deb5c7a752e01a07ff88194d98f10cec683/Casks/android-ndk.rb), [Crystax NDK](https://github.com/Homebrew/homebrew-cask/blob/60f20d3f711eeb0ef023cad662eecbea02b96465/Casks/crystax-ndk.rb));
2. adding a caveat describing how to use a new `HOMEBREW_GOOGLE_CLOUD_SDK_ADDITIONAL_COMPONENTS` environment variable to keep the additional components.

This also adds `auto_updates true` since Google Cloud SDK automatically checks and informs when a new version is available (unless `component_manager.disable_update_check` [config property](https://cloud.google.com/sdk/gcloud/reference/config) is set) and users can run `gcloud components update` to upgrade.

Manually (`gcloud components install`) installed additional components would still be lost after an upgrade unless also added to `HOMEBREW_GOOGLE_CLOUD_SDK_ADDITIONAL_COMPONENTS`, which is admittedly not ideal. However:
- users that don't use `brew upgrade` will not experience that;
- users of `brew upgrade` without `--greedy` will also not experience that as `auto_updates true` casks are not included;
- users of `brew upgrade --greedy` were losing the additional components anyway since `--greedy` includes unversioned casks, and now at least have a way to prevent that, and will only be upgrading the casks when the version actually changed.

### DIFF
--- a/Casks/google-cloud-sdk.rb
+++ b/Casks/google-cloud-sdk.rb
@@ -1,22 +1,29 @@
 cask "google-cloud-sdk" do
-  version :latest # Must remain unversioned, else all installed gcloud components would be lost on upgrade
-  sha256 :no_check
+  version "331.0.0"
+  sha256 "32c7b961f4c3d18f772a8900c9bc1300e7f81c0c05d71f93b770a222f00490ba"
 
-  url "https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz"
+  url "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/#{token}-#{version}-darwin-x86_64.tar.gz"
   name "Google Cloud SDK"
   desc "Set of tools to manage resources and applications hosted on Google Cloud"
   homepage "https://cloud.google.com/sdk/"
 
+  auto_updates true
   depends_on formula: "python"
 
-  stage_only true
+  artifact token, target: "#{HOMEBREW_PREFIX}/share/#{token}"
 
   postflight do
     system_command "#{staged_path}/#{token}/install.sh",
                    args: [
                      "--usage-reporting", "false", "--bash-completion", "false", "--path-update", "false",
                      "--rc-path", "false", "--quiet"
-                   ],
+                   ] + (
+                     if ENV.key?("HOMEBREW_GOOGLE_CLOUD_SDK_ADDITIONAL_COMPONENTS")
+                       ["--additional-components"] + ENV["HOMEBREW_GOOGLE_CLOUD_SDK_ADDITIONAL_COMPONENTS"].split(",")
+                     else
+                       []
+                     end
+                   ),
                    env:  { "CLOUDSDK_PYTHON" => Formula["python"].opt_bin/"python3" }
   end
 
@@ -25,17 +32,20 @@ cask "google-cloud-sdk" do
   uninstall delete: "#{staged_path}/#{token}"
 
   caveats <<~EOS
-    #{token} is installed at #{staged_path}/#{token}. Add your profile:
+    #{token} is installed at #{HOMEBREW_PREFIX}/share/#{token}. Add your profile:
 
       for bash users
-        source "#{staged_path}/#{token}/path.bash.inc"
-        source "#{staged_path}/#{token}/completion.bash.inc"
+        source "#{HOMEBREW_PREFIX}/share/#{token}/path.bash.inc"
+        source "#{HOMEBREW_PREFIX}/share/#{token}/completion.bash.inc"
 
       for zsh users
-        source "#{staged_path}/#{token}/path.zsh.inc"
-        source "#{staged_path}/#{token}/completion.zsh.inc"
+        source "#{HOMEBREW_PREFIX}/share/#{token}/path.zsh.inc"
+        source "#{HOMEBREW_PREFIX}/share/#{token}/completion.zsh.inc"
 
       for fish users
-        source "#{staged_path}/#{token}/path.fish.inc"
+        source "#{HOMEBREW_PREFIX}/share/#{token}/path.fish.inc"
+
+    To keep additional components after an upgrade or reinstallation add to your profile:
+      export HOMEBREW_GOOGLE_CLOUD_SDK_ADDITIONAL_COMPONENTS='COMPONENT_ID,COMPONENT_ID'
   EOS
 end
--- a/Casks/google-cloud-sdk.rb
+++ b/Casks/google-cloud-sdk.rb
@@ -2,7 +2,7 @@ cask "google-cloud-sdk" do
   version "332.0.0"
   sha256 "1d158a84dfff8f3aa3cacdbbd3fb9cfbea3178728367b20c986760edc1665026"
 
-  url "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/#{token}-#{version}-darwin-x86_64.tar.gz"
+  url "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-#{version}-darwin-x86_64.tar.gz"
   name "Google Cloud SDK"
   desc "Set of tools to manage resources and applications hosted on Google Cloud"
   homepage "https://cloud.google.com/sdk/"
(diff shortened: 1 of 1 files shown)


Casks/google-cloud-sdk.rb
```diff
@@ -25,17 +32,20 @@
...
 
   caveats <<~EOS
-    #{token} is installed at #{staged_path}/#{token}. Add your profile:
+    #{token} is installed at #{HOMEBREW_PREFIX}/share/#{token}. Add your profile:
 
       for bash users
-        source "#{staged_path}/#{token}/path.bash.inc"
-        source "#{staged_path}/#{token}/completion.bash.inc"
+        source "#{HOMEBREW_PREFIX}/share/#{token}/path.bash.inc"
+        source "#{HOMEBREW_PREFIX}/share/#{token}/completion.bash.inc"
```
sparhomenko, line 39: This theoretically can be replaced by `artifact #{staged_path}/#{token}/completion.bash.inc, target: "#{HOMEBREW_PREFIX}/etc/bash_completion.d/google-cloud-sdk` or a symlink in postflight, making user experience a bit better (they don't need to manually add this line as they probably already added Homebrew completions). It won't work for ZSH completions however as `completion.zsh.inc` cannot be added to `share/zsh/site-functions` (it should be `source`d, not added to `FPATH`). Not sure if it makes sense to change it or not.

Casks/google-cloud-sdk.rb
```diff
@@ -25,17 +32,20 @@
   uninstall delete: "#{staged_path}/#{token}"
 
   caveats <<~EOS
-    #{token} is installed at #{staged_path}/#{token}. Add your profile:
+    #{token} is installed at #{HOMEBREW_PREFIX}/share/#{token}. Add your profile:
```
sparhomenko, line 35: This is theoretically redundant as the executed installer scripts will print the following:
> ==> Source [/usr/local/share/google-cloud-sdk/completion.zsh.inc] in your profile to enable shell command completion for gcloud.
> ==> Source [/usr/local/share/google-cloud-sdk/path.zsh.inc] in your profile to add the Google Cloud SDK command line tools to your $PATH.

This is even better than the caveat as it'll automatically detect the shell and show the files to be sourced. Not sure however if it makes sense to rely on that or it's better to keep the caveat.

Casks/google-cloud-sdk.rb
```diff
@@ -1,41 +1,57 @@
...
+  artifact token, target: "#{HOMEBREW_PREFIX}/share/#{token}"
 
   postflight do
     system_command "#{staged_path}/#{token}/install.sh",
                    args: [
                      "--usage-reporting", "false", "--bash-completion", "false", "--path-update", "false",
                      "--rc-path", "false", "--quiet"
-                   ],
+                   ] + (
+                     if ENV.key?("HOMEBREW_GOOGLE_CLOUD_SDK_ADDITIONAL_COMPONENTS")
```
vitorgalvao, line 21: This is suboptimal. We shouldn’t be introducing homebrew environment variables for use in single casks.

Casks/google-cloud-sdk.rb
```diff
@@ -1,41 +1,57 @@
...
+  url "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/#{token}-#{version}-darwin-x86_64.tar.gz"
   name "Google Cloud SDK"
   desc "Set of tools to manage resources and applications hosted on Google Cloud"
   homepage "https://cloud.google.com/sdk/"
 
+  auto_updates true
   depends_on formula: "python"
 
-  stage_only true
+  artifact token, target: "#{HOMEBREW_PREFIX}/share/#{token}"
```
vitorgalvao, line 13: `token` can be used where it will always be correct (i.e. messaging), it should never be used for code.

Casks/google-cloud-sdk.rb
```diff
@@ -1,41 +1,57 @@
...
+  version "331.0.0"
+  sha256 "32c7b961f4c3d18f772a8900c9bc1300e7f81c0c05d71f93b770a222f00490ba"
 
-  url "https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz"
+  url "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/#{token}-#{version}-darwin-x86_64.tar.gz"
   name "Google Cloud SDK"
   desc "Set of tools to manage resources and applications hosted on Google Cloud"
   homepage "https://cloud.google.com/sdk/"
 
+  auto_updates true
```
vitorgalvao, line 10: This doesn’t auto update. Users have to run a specific command for it to happen, which isn’t the same thing.

vitorgalvao (commented):
Thank you for your work on this. Clearly you gave it some thought.

However, I’m unconvinced this will be an improvement to most users, which is why I’d like for users of this cask to weigh in.

Casks/google-cloud-sdk.rb
sparhomenko: `#{token}` has already been used in this cask before my changes for more than messaging, eg. in `system_command` and `uninstall`. If I understand you correctly that should not be done so I've removed all usages except `caveats`.

Casks/google-cloud-sdk.rb
sparhomenko: With default configuration it checks for updates not only when you run the specific command but also when you use it normally (eg. run other commands), giving this message:
>Updates are available for some Cloud SDK components.  To install them, please run:
>$ gcloud components update

It indeed does not install updates automatically. From what 'check automatically but do not install automatically' is typical for `auto_updates true` casks, eg. `iterm`, `alfred` and `bartender`.

georgettica:
Is there a way I can assist you in making this work?
It really annoys me that each upgrade the version changes

Chris-Hopkins:
Google now supply an arm64 tar of google-cloud-sdk, I've played around with the current cask locally to check and install that version on my m1. The only issue I ran into was the versioning requirement for including the arm64 download link.

cho-m:
@Chris-Hopkins Based on some quick checks, I think the current unversioned installation is aware of hardware.

For example, if I install current Cask, I see arm64 binaries:
```console
❯ cd /opt/homebrew/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin

❯ file * | grep Mach
anthoscli:                Mach-O 64-bit executable arm64
```

Extracting the versioned URLs locally, this is similar to the ARM-specific variant.

So, I think "versioned vs. unversioned (`latest`)" should have no impact on installing native binaries.

Chris-Hopkins:
@cho-m You're correct! I only ran a simple comparison of the directories for x86_64 and arm64 downloads. Doh!

tgbauer:
Would like some version control on google-cloud-sdk.
Would also be nice to have options for including the "alpha" and "beta" components.  Maybe have additional casks like 'google-cloud-sdk-beta' that installs base as well as beta, 'google-cloud-sdk-alpha' that installs base and beta and alpha?

Currently using '--greedy' option (which reinstalls every time) and adding gcloud commands to reinstall beta and alpha

vitorgalvao:
@tgbauer This cask is already complex enough as it is and it’s not even merged. It’s not yet the time to take about alpha and beta versions, and we’re unlikely to accept those.

tgbauer:
@vitorgalvao ,  totally understand and for sure would want version control first/more.
 Specifically called out the alpha and beta components that are wiped out because they are used so often (more than other alpha or beta)

georgettica:
what is the state of this?
I don't have enough experience in this to drive it forward but if I would have guidance I can try!

(also, can building this be done in a containerized env? last time it required cleaning up)

vitorgalvao:
> what is the state of this?

There hasn’t been any solutions regarding [my comment](https://github.com/Homebrew/homebrew-cask/pull/101021#discussion_r595442491) and this is now conflicting with other updates to the cask.

Best to close and anyone can try to improve on a new PR.

PR #101208, opened 2021-03-15
Update to 2.0.23

tomschlenkhoff:


jconley:
fyi, UTM only runs on big sur... errors on startup with message:

jconley:
fyi, UTM only runs on big sur... errors on startup with message:
```You have macOS 10.15.7. The application requires macOS 11.0 or later.```

might consider adding ```depends_on macos: ">= :big_sur"``` on the next version bump.

PR #102417, opened 2021-03-22
mellowplayer: add livecheck

reitermarkus:


vitorgalvao:
Seems these are all adding `discontinued`. The commit message doesn’t make sense, then.

reitermarkus:
Technically I am running a script for adding implicit livechecks (i.e. the `appcast` can be removed), and some of these are also failing since they're discontinued.

vitorgalvao (approved):

PR #102520, opened 2021-03-24
Delete adobe-dng-converter.rb

vitorgalvao:
Adobe made new versions only downloadable with an account, and the URL has copy protection.


lkrms:
Hi @vitorgalvao, just wondering where you saw this change from Adobe?

This [Adobe DNG Converter page](https://helpx.adobe.com/photoshop/using/adobe-dng-converter.html) doesn't report any restrictions and I can reach the 13.2 [dmg file](https://download.adobe.com/pub/adobe/dng/mac/DNGConverter_13_2.dmg) via `curl` without any issues.

vitorgalvao:
> just wondering where you saw this change from Adobe?

Not where, *when*. It’s working now but wasn't at the time. You may resubmit it updated if you’d like. Leave the High Sierra version out.

PR #102523, opened 2021-03-24
Delete adobe-air.rb

vitorgalvao:
https://get.adobe.com/air/ redirects to https://blog.adobe.com/en/publish/2019/05/30/the-future-of-adobe-air.html#gs.x16gec


reitermarkus:
I think we have a unpack strategy specifically for AIR applications. We should remove that and them too.

vitorgalvao:
Agreed, but tests on the core repo never pass for me (even without modifications) so I’d be removing it blindly.

PR #102562, opened 2021-03-24
lbry: Update to v0.50.1, add shimscripts for lbrynet and lbry-first, add desc

jivanpal:

- [x] *(Note from PR creator: There is no mention of stable versions in the linked document)* The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Additionally, **if adding a new cask**: N/A

### DIFF
--- a/Casks/lbry.rb
+++ b/Casks/lbry.rb
@@ -1,11 +1,29 @@
 cask "lbry" do
-  version "0.49.4"
-  sha256 "dba1adcc44c8548459e51ff690d67973215502fcf92623ee15cf2837f7f65989"
+  version "0.50.1"
+  sha256 "483a14e53a1638b6aa586444724e1d33d7ec60be2fc72f2a47dab3f6d66840f2"
 
   url "https://github.com/lbryio/lbry-desktop/releases/download/v#{version}/LBRY_#{version}.dmg",
       verified: "github.com/lbryio/lbry-desktop/"
-  name "LBRY"
+  name "LBRY Desktop"
+  desc "Official client for LBRY, a decentralised file-sharing and payment network"
   homepage "https://lbry.com/"
 
   app "LBRY.app"
+  # shim scripts (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shim_lbrynet = "#{staged_path}/lbrynet.wrapper.sh"
+  shim_lbryfirst = "#{staged_path}/libry-first.wrapper.sh"
+  binary shim_lbrynet, target: "lbrynet"
+  binary shim_lbryfirst, target: "lbry-first"
+
+  preflight do
+    IO.write shim_lbrynet, <<~EOS
+      #!/bin/sh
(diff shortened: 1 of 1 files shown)


Casks/lbry.rb
```diff
@@ -1,11 +1,40 @@
...
+  sha256 "483a14e53a1638b6aa586444724e1d33d7ec60be2fc72f2a47dab3f6d66840f2"
 
   url "https://github.com/lbryio/lbry-desktop/releases/download/v#{version}/LBRY_#{version}.dmg",
       verified: "github.com/lbryio/lbry-desktop/"
-  name "LBRY"
+  name "LBRY Desktop"
+  desc "Official client for LBRY, a decentralised file-sharing and payment network"
   homepage "https://lbry.com/"
 
+  livecheck do
```
vitorgalvao, line 11: Remove this and try `brew livecheck lbry` to see if it works. Because this uses a GitHub `url`, it should.

Casks/lbry.rb
```diff
@@ -1,11 +1,40 @@
...
+    url :url
+    strategy :github_latest
+  end
+
   app "LBRY.app"
+  # shim scripts (https://github.com/Homebrew/homebrew-cask/issues/18809)
```
vitorgalvao, line 17: Any official documentation where they recommend making these?

vitorgalvao (changes requested):

Casks/lbry.rb
jivanpal: Not required, [official docs](https://lbry.com/faq/how-to-cli) instruct user to `cd` to the location of `lbrynet` if they wish to use it, but this is tedious. I've opted to create the shimscripts for convenience, in the same vein as Homebrew's `vlc` package. I cannot find any official documentation for `lbry-first`, but it can also be used just fine without `cd`'ing to its directory. Both apps interface directly with the LBRY RPC server when it is running (when the app is running or when `lbrynet start` is run).

Casks/lbry.rb
jivanpal: Commited by @miccal, not me, so I don't know what the purpose of this is. Removing it and then running `brew livecheck lbry` outputs `lbry : 0.50.1 ==> 7110823363bdf8f3e1e` — shall I commit the removal?

vitorgalvao: Ah, didn’t check who the committer was. No, please don’t remove—if @miccal added it, it’s needed (as shown by your result; you should’ve gotten the same result with and without for it to not be needed).

Casks/lbry.rb
vitorgalvao: Good enough. I agree that `cd`ing would be a bad experience for this.

vitorgalvao (approved):

Issue #66273, opened 2019-07-15
Updating MultiMC removes user data


haykam821:
**Expected behavior:** User data is preserved between updates to the `multimc` cask.  
**Actual behavior:** When upgrading the `multimc` cask, all user data is removed. 

This is caused by MultiMC storing user data in `/Applications/MultiMC.app/Contents/MacOS` by default (which is irresponsible). This could be fixed by temporarily copying the user data in that folder to a different location when overwriting the app, and placing it back in the updated app.

It may also be helpful to use the `caveats` stanza to warn the user to configure the folders differently, like I have now done:

![Updated folders configuration](https://user-images.githubusercontent.com/24855774/61245854-75927980-a71b-11e9-925b-bdb44e939f31.png)

vitorgalvao:
> This is caused by MultiMC storing user data in `/Applications/MultiMC.app/Contents/MacOS` by default (which is irresponsible)

Agreed. But that’s an upstream issue, not a Homebrew Cask one.

> It may also be helpful to use the `caveats` stanza to warn the user to configure the folders differently

From the [`caveats` documentation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md):

> To avoid flooding users with too many messages (thus desensitising them to the important ones), `caveats` should be used sparingly and exclusively for installation-related matters.

From the [Homebrew Cask is not a discoverability service FAQ document](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md)

> users are expected to have reasonable knowledge about the apps they’re installing through us before doing so.

But I thank you for the bug report. It was well made, understandable, and actionable.

PR #102590, opened 2021-03-25
Delete multimc.rb

kb-1000:
Remove MultiMC.
To quote the MultiMC author:
> that thing is dangerous and should not be used

(about this package)
MultiMC is not meant for being used like this, and does not support it. For example, under no circumstances will updating it using this keep the settings and instances.
MultiMC has its own updater.

References #66273.


kb-1000:
That might have been fixed, actually, however I do not actuallly know if that's what auto_updates does.

miccal:
Thanks @kb-1000, but I believe this has been fixed by adding `auto_updates true`.

PR #102953, opened 2021-04-01
butt: remove livecheck

miccal:


Bo98:
Hmm, there seems to be a bit of disparity here:

```
$ brew cask --audit butt
audit for butt: failed
 - Download is hosted on SourceForge, please add a livecheck. See https://github.com/Homebrew/homebrew-cask/blob/HEAD/doc/cask_language_reference/stanzas/livecheck.md
```

The CI here adds an `--appcast` flag which silences that:

```
$ brew cask --audit --appcast butt
audit for butt: passed
```

Same with `gpxsee`.

PR #103136, opened 2021-04-05
Update datadog-agent from 7.22.0-1 to 7.26.0-1 and add livecheck to cask

cpmsmith:

## Caveat
While testing this, I realized that the agent stores its config files in its own little virtual environment in `/opt/datadog-agent/` and symlinks them into `~/.datadog-agent/`, rather than the reverse. Since `/opt/datadog-agent/` also contains the entire installation, it's in the `uninstall` stanza, meaning config files get deleted when upgrading/uninstalling. This isn't new, but it'll presumably be triggered every time the cask is updated. Does Homebrew have any tooling for dealing with something like this?

### DIFF
--- a/Casks/datadog-agent.rb
+++ b/Casks/datadog-agent.rb
@@ -10,7 +10,7 @@ cask "datadog-agent" do
 
   livecheck do
     url "https://s3.amazonaws.com/dd-agent/"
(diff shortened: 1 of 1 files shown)


miccal:
Thank you for your contribution to `homebrew-cask` @cpmsmith.

miccal (approved):

PR #103216, opened 2021-04-06
google-chrome: add livecheck

miccal:


reitermarkus (approved):

miccal:
@reitermarkus but this way `brew bump-unversioned-casks` does not work:
```
|-> brew bump-unversioned-casks --dry-run google-chrome
==> Unversioned Casks: 1 (0 cached)
==> Checking google-chrome
==> Downloading https://dl.google.com/chrome/mac/universal/stable/GGRO/googlechrome.dmg
==> Would bump google-chrome from 89.0.4389.114 to 89.0.4389.114,4389.114
```

reitermarkus:
@miccal, my plan is that once we have https://github.com/Homebrew/brew/pull/11028 we can get rid of `brew bump-unversioned-casks` and consistently use `brew livecheck` to detect updates for all casks. In this case, there is already an actual appcast, so this should be preferred.

PR #103378, opened 2021-04-08
Revert "Add Emacs man pages"

miccal:
Reverts Homebrew/homebrew-cask#103373


Vynce:
Thanks! Verified that cask install works now 😊.

PR #22699, opened 2016-07-10
Add cask for Magic Number Machine.

yacoob:
#### Adding a new cask
- [x] Checked there aren’t open [pull requests](https://github.com/caskroom/homebrew-cask/pulls) for the same cask.
- [x] Checked there aren’t closed [issues](https://github.com/caskroom/homebrew-cask/issues) where that cask was already refused.
- [x] When naming the cask, followed the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] Commit message includes cask’s name.
- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [x] `brew cask install {{cask_file}}` worked successfully.
- [x] `brew cask uninstall {{cask_file}}` worked successfully.

PR #103412, opened 2021-04-09
Remove discontinued battery-guardian cask

thatrobotdev:
Battery Guardian is no longer being updated and is labeled a legacy product on its website (https://www.dssw.co.uk/batteryguardian/).

To my knowledge (and after searching around on their website and on google), there is no updated version of Battery Guardian, and development hasn't started again after it was discontinued. However, I haven't used this software personally, so I could be wrong.

Because Battery Guardian is discontinued, I think it can be deleted? It's not clear to me if the guidelines for "[Rejected Casks](https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Casks.md#rejected-casks)" apply only to when you are submitting a new cask, or that casks that don't fulfill these that are already in the repo should be deleted, but if it's the latter than it doesn't follow this reason:

> The Cask is for an unmaintained app (no releases in the last year, or [explicitly discontinued](https://github.com/Homebrew/homebrew-cask/pull/22699)).


miccal:
```
==> Analytics
install: 3 (30 days), 18 (90 days), 83 (365 days)
```

miccal (approved):

miccal:
Thank you @thatrobotdev.

PR #103769, opened 2021-04-13
Windscribe 2

jivanpal:

Additionally, **if adding a new cask**: N/A, but...

- [ ] `brew audit --new-cask {{cask_file}}` worked successfully. **(didn't check this)**

***

Windscribe v2 uses an installer app (as opposed to v1, which just used drag-and-drop). If there is any way to automate installation rather than requiring the user to open the file from Caskroom due to `installer manual`, that would be good to know. However, I have not been able to discover any command-line arguments taken by the installer program that may make this possible.

Should v1 be added to [homebrew-cask-versions](https://github.com/Homebrew/homebrew-cask-versions)? It is still available on Windscribe.com as ["Legacy version"](https://assets.totallyacdn.com/desktop/mac/Windscribe-legacy.dmg).

Also, `brew style --fix` seems to have messed with the indentation of my `uninstall launchctl` block (see second commit). Should this "fix" be reverted?

### DIFF
--- a/Casks/windscribe.rb
+++ b/Casks/windscribe.rb
@@ -10,9 +10,9 @@ cask "windscribe" do
   installer manual: "WindscribeInstaller.app"
 
   uninstall launchctl: [
-              "com.windscribe.helper.macos",
-              "com.windscribe.launcher.macos",
-            ],
+    "com.windscribe.helper.macos",
+    "com.windscribe.launcher.macos",
+  ],
             quit:      [
               "com.windscribe.gui.macos",
               "com.windscribe.helper.macos",
--- a/Casks/windscribe.rb
+++ b/Casks/windscribe.rb
@@ -1,11 +1,17 @@
 cask "windscribe" do
-  version :latest
+  version "2.02.10"
   sha256 :no_check
 
   url "https://assets.windscribe.com/desktop/mac/Windscribe.dmg"
   name "Windscribe"
   desc "VPN client for secure internet access and private browsing"
   homepage "https://windscribe.com/"
+
+  livecheck do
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> If there is any way to automate installation rather than requiring the user to open the file from Caskroom due to `installer manual`, that would be good to know.

Check inside the installer app. Often they are thin wrappers for a `.pkg`, likely in the `Resources` or `MacOS` directories. They may also have a zip or similar with an app inside. The more corporate an app is, the likelier they’ll have a bullshit installer.

> Should v1 be added to [homebrew-cask-versions](https://github.com/Homebrew/homebrew-cask-versions)?

No. If anyone needs it, they can submit it then.

jivanpal:
> Check inside the installer app. Often they are thin wrappers for a `.pkg`

Alas, not the case here, the `.app` appears to just be a wrapper for an executable which is a one-click GUI installer, and it apparently takes no command line arguments. Does Homebrew support/condone launching the installer `.app` so that the user doesn't have to open it from Caskroom themselves (e.g. by using `open` within a `do` block)?

vitorgalvao:
> Does Homebrew support/condone launching the installer `.app` so that the user doesn't have to open it from Caskroom themselves (e.g. by using `open` within a `do` block)?

Not at present. If we did, we wouldn’t have `installer manual:`. See https://github.com/Homebrew/homebrew-cask/issues/24377 for future plans, but those are unlikely to happen any time soon unless we get a contribution.

vitorgalvao:

vitorgalvao (approved):

PR #103774, opened 2021-04-13
dbvisualizer: add livecheck

reitermarkus:


cho-m:
Recursively same `location`.

One option would be to use variant of appcast `https://www.dbvis.com/download/#{version.major_minor}`
and then try to regex match on the dropdown menu for all versions, e.g.
```rb
regex(/Latest\s*Version:\s*(\d+(?:\.\d+)*)/i)
```

This works even when accessing older versions of page, e.g. https://www.dbvis.com/download/9.5
It will break if `dbvisualizer` removes the version page, e.g. https://www.dbvis.com/download/9.4 (ends up redirecting to /12.0 and hits recursive issue again).

miccal (approved):

PR #103997, opened 2021-04-14
mysqlworkbench: add livecheck

reitermarkus:


miccal (approved):

bpoland:
@miccal it looks like this partially reverted a previous commit to use 8.0.22 on catalina -- 8.0.23 is broken on catalina so I don't think this is a good change.  Can we go back to using 8.0.22 on catalina?

bpoland:
(see https://github.com/Homebrew/homebrew-cask/pull/99603)

mynameisbogdan:
8.0.22 isn't compatible with mojave either.

Screenshot 2021-04-17 at 14 50 14

PR #104067, opened 2021-04-15
Add ARM support for parallels

cbrockschmidt:
Parallels has two build numbers for each version (ARM/Intel), so I had to declare a different version for each of them. An alternative would be to use `version:latest`, but I think keeping the version numbers would be better.


miccal:
Thank you for your contribution to `homebrew-cask` @cbrockschmidt.

miccal (approved):

PR #104199, opened 2021-04-18
Remove all automerge code.

MikeMcQuaid:
This is not secure enough.


bbenzikry:
@MikeMcQuaid, @Ry0taK  I understand this has to do with #104191 
I looked at the disclosure on hackerone and it seems like you opted against opening this up to the world. 
Is it possible to change your mind? I think others will benefit from seeing what was wrong here

MikeMcQuaid:
@bbenzikry We will be publishing a security disclosure blog post for this. We just need to write it up and coordinate with @Ry0taK.

bbenzikry:
@MikeMcQuaid Great, thank you 👍

MikeMcQuaid:
We have posted about this in https://brew.sh/2021/04/21/security-incident-disclosure/

PR #104716, opened 2021-04-28
cuteclips: remove

cho-m:
Looks like `cuteclips` renamed to `tes`.

The current Cask download URL appears to be a dead link.

As download count is low and there appears to be no-one who raised an issue or tried to fix this, I think this Cask should be considered for removal.


miccal:
```
==> Analytics
install: 0 (30 days), 0 (90 days), 4 (365 days)
b
```

miccal (approved):

PR #104732, opened 2021-04-28
Update camtasia from 2020.0.18 to 2021.0.0 and add livecheck

cho-m:

---

### Versioned URL.
e.g. Version **2021.0.0** is shown as **21.0.0** in https://www.techsmith.com/download/oldversions
The following appear to be same DMG:
- https://download.techsmith.com/camtasiamac/releases/2100/Camtasia.dmg
- https://download.techsmith.com/camtasiamac/releases/Camtasia.dmg

### Compatibility
https://support.techsmith.com/hc/en-us/articles/219910027-Mac-OS-X-Support-Matrix
- High Sierra: 3.1.1 - 2020.x.x.
  - 20.0.18 released: 2021 Apr 07
- Mojave - Big Sur all work with latest
  -  21.0.0 released: 2021 Apr 27

### DIFF
--- a/Casks/camtasia.rb
+++ b/Casks/camtasia.rb
@@ -1,14 +1,20 @@
 cask "camtasia" do
-  version "2020.0.18,114362"
-  sha256 :no_check
+  version "2021.0.0"
+  sha256 "1afed5fb6bef606eae88d116247bb8ba5d678b33716c342ea8a548d2a3f05240"
 
-  url "https://download.techsmith.com/camtasiamac/releases/Camtasia.dmg"
-  appcast "https://support.techsmith.com/hc/en-us/articles/115006624748-Camtasia-Mac-Version-History"
+  url "https://download.techsmith.com/camtasiamac/releases/#{version.major[-2..]}#{version.minor_patch.no_dots}/Camtasia.dmg"
   name "Camtasia"
+  desc "Screen recorder and video editor"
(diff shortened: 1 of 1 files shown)


SMillerDev (dismissed):

Casks/camtasia.rb
```diff
@@ -1,10 +1,27 @@
 cask "camtasia" do
-  version "2020.0.18,114362"
-  sha256 :no_check
+  if MacOS.version <= :high_sierra
```
vitorgalvao, line 2: Unless you personally need support under High Sierra, please remove this (i.e. don’t add it).

These conditionals, especially when introducing more complex blocks like `livecheck` incur a maintenance cost. In our experience, when they’re added unprompted, for the most part they just sit abandoned and forgotten until they break. So unless you need it (and thus have an investment in maintaining it), we’d rather not have it (if someone in the future needs it, we can consider it then). Plus, High Sierra is below the recommended version for Homebrew itself.

vitorgalvao (changes requested):

Casks/camtasia.rb
cho-m: Sure. Updated to only latest version and bumped requirement to Mojave.

vitorgalvao: Thank you!

vitorgalvao (approved):

PR #104764, opened 2021-04-28
lingo: remove

cho-m:
I think hockeyapp.net was retired and apps were all migrated to Microsoft App Center.
So, the download URL for `lingo` appears to be using an old dead link.

The latest version of `lingo` appears to be **12.4 (3190)** but this Cask wasn't maintained and no-one raised issue / opened PR on this.
Given very low downloads, I think it is worth considering removal.

If someone wants this app, they can create new Cask with functioning URLs.


miccal:
```
|-> brew info lingo
==> Analytics
install: 0 (30 days), 0 (90 days), 13 (365 days)
```

miccal (approved):

PR #104799, opened 2021-04-29
Add description to SpiderOakOne

enorms:

- [X] `brew audit --cask spideroakone` is error-free.
- [X] `brew style --fix spideroakone` reports no offenses.


Casks/spideroakone.rb
```diff
@@ -9,7 +9,9 @@
   else
     url "https://spideroak.com/release/spideroak/osx"
   end
+
   name "SpiderOakOne"
+  desc "Cloud backup and storage"
   homepage "https://spideroak.com/"
```
cho-m, line 16:
```suggestion

  livecheck do
    url :url
    strategy :header_match
  end

```

cho-m: This will only detect version = `7.5.0`, so please update `version` number too.
i.e. Line 2 `version "7.5.0.10333"` => `version "7.5.0"`

cho-m:
@lifekaizen can you modify the `version` to be
```rb
  version "7.5.0"
```
I can't suggest the change directly since it is outside the range that GitHub allows and I don't have write permissions to directly modify.

I don't see any reference to extra numbers (probably only available inside App), so I think it should be fine with the main 3 version parts.

lifekaizen:
> @lifekaizen can you modify the `version` to be
> 
> ```ruby
>   version "7.5.0"
> ```
> 
> I can't suggest the change directly since it is outside the range that GitHub allows and I don't have write permissions to directly modify.
> 
> I don't see any reference to extra numbers (probably only available inside App), so I think it should be fine with the main 3 version parts.

I am not sure where the version string came from and was just researching. When I started modifying it was set to ':latest' which seems appropriate since the 'URL doesn't contain a version'. Would you be ok with using ':latest'? Or prefer '7.5.0' (which I agree is the current Mac version)

cho-m:
Since it is currently possible to `livecheck` a version number, it would be better to keep a specific `version` but leave `sha256 :no_check` (in case URL has newer version).

I believe this is the general preference in Homebrew Cask and is probably the reason the specific version was added by bot.

It also works better with Homebrew's version tracking, upgrade logic, etc.

Casks/spideroakone.rb
lifekaizen: Added

SMillerDev (approved):

PR #104806, opened 2021-04-29
Description for Pure data package

roolo:
Without this change it is really hard to tell if this package is really for Pure data


miccal (approved):

roolo:
@miccal Hi, it was my goal to clarify, the package is about Pure data. In 6c10da4 you're removing any mention of the Pure data name. Could you please rephrase the description so it mentions the name?

PR #104889, opened 2021-05-01
Update qt3dstudio from 2.6.0 to 2.8.0 and add livecheck

cho-m:

### DIFF
--- a/Casks/qt3dstudio.rb
+++ b/Casks/qt3dstudio.rb
@@ -1,12 +1,20 @@
 cask "qt3dstudio" do
-  version "2.6.0"
-  sha256 "840333076975954923dbc1bfc5833521f228a4e1408d846f42acd57d7e32f65b"
+  version "2.8.0"
+  sha256 "bdaa9ede1c9dcf2137d14ad312bafaab08c6b37e7ad9286160240e38daf56857"
 
   url "https://download.qt.io/official_releases/qt3dstudio/#{version.major_minor}/qt-3dstudio-opensource-mac-x64-#{version}.dmg"
-  appcast "https://download.qt.io/official_releases/qt3dstudio/",
-          must_contain: version.major_minor
   name "Qt 3D Studio"
+  desc "Compositing tool"
   homepage "https://www.qt.io/developers/"
+
+  livecheck do
+    url "https://download.qt.io/official_releases/qt3dstudio/"
+    strategy :page_match do |page|
+      version_major_minor = page[%r{href="(\d+(?:\.\d+)*)/"}i, 1]
+      version_page = Net::HTTP.get(URI.parse("#{url}#{version_major_minor}/"))
+      version_page[/qt-3dstudio-opensource-mac-x64-(\d+(?:\.\d+)*)\.dmg/i, 1]
+    end
+  end
 
   depends_on macos: ">= :sierra"
 


Casks/qt3dstudio.rb
```diff
@@ -1,13 +1,21 @@
...
 
+  livecheck do
+    url "https://download.qt.io/official_releases/qt3dstudio/"
+    strategy :page_match do |page|
+      version_major_minor = page[%r{href="(\d+(?:\.\d+)*)/"}i, 1]
+      version_page = Net::HTTP.get(URI.parse("#{url}#{version_major_minor}/"))
+      version_page[/qt-3dstudio-opensource-mac-x64-(\d+(?:\.\d+)*)\.dmg/i, 1]
+    end
+  end
+
```
SMillerDev, line 18: @nandahkrishna or @samford can you give your opinion?

SMillerDev (approved):

Casks/qt3dstudio.rb
```diff
@@ -1,13 +1,21 @@
...
   homepage "https://www.qt.io/developers/"
 
+  livecheck do
+    url "https://download.qt.io/official_releases/qt3dstudio/"
+    strategy :page_match do |page|
+      version_major_minor = page[%r{href="(\d+(?:\.\d+)*)/"}i, 1]
+      version_page = Net::HTTP.get(URI.parse("#{url}#{version_major_minor}/"))
+      version_page[/qt-3dstudio-opensource-mac-x64-(\d+(?:\.\d+)*)\.dmg/i, 1]
+    end
+  end
```
samford, line 17:
```suggestion
  # File names contain the full major/minor/patch version but they're kept in
  # major/minor version directories, so it's unfortunately necessary to fetch
  # two pages to obtain the latest version.
  livecheck do
    url "https://download.qt.io/official_releases/qt3dstudio/?C=M;O=D"
    strategy :page_match do |page|
      version_major_minor = page[%r{href=["']?v?(\d+(?:\.\d+)+)/?["' >]}i, 1]
      version_page = Homebrew::Livecheck::Strategy.page_content(url.sub("/?", "/#{version_major_minor}/?"))
      return "" unless version_page[:content].present?

      version_page[:content].scan(/href=.*?qt-3dstudio-opensource-mac-x64[._-]v?(\d+(?:\.\d+)+)\.dmg/i).map { |match| match.first }
    end
  end
```

The `strategy` block in this PR assumes that the default sorting will always have the newest version directory at the top but it's better to ensure this by adding URL parameters to sort by modified date in a descending fashion. The alternative is to use `page.scan`, convert the matched text to `Version`s, and use the newest version. The latter is more reliable but will be verbose, so it's probably fine to take the easy path for now.

This suggestion modifies the first regex to use the standard regex for matching version directories where `href` URLs only consist of the version and an optional forward slash at the end.

Regarding the second HTTP request, we should use either a `Livecheck::Strategy` method for fetching content (e.g., `Strategy#page_content`) or simply call a strategy's `find_versions` method (e.g., `PageMatch#find_versions`, which calls `Strategy#page_content`). This ensures that we're using the established methods for fetching content in livecheck and won't run into any weirdness by using something different. This also means that the `strategy` block will benefit from any future internal changes in `Strategy#page_content` without needing to be manually modified to maintain parity. [I'll have to look into making the strategy methods available in `strategy` blocks so we don't need to use the full `Homebrew::Livecheck::Strategy` name.]

In light of the added query string parameters, this updates the URL-generating code to add the major/minor version before the parameters.

I've also updated the second regex to follow the standard pattern for matching a file link on an HTML page. This also updates the matching behavior on the second page to use `scan` and `map`, so we can simply return all the matches and allow livecheck to determine which version is newest. Ideally we want to avoid/minimize duplicating livecheck's behavior in `strategy` blocks.

There may be more room for improvement here but this is more in line with how I would approach this.

samford (changes requested):

PR #104894, opened 2021-05-01
izip: add livecheck

cho-m:

### DIFF
--- a/Casks/izip.rb
+++ b/Casks/izip.rb
@@ -3,10 +3,17 @@ cask "izip" do
   sha256 :no_check
 
   url "https://www.izip.com/izip.dmg"
-  appcast "https://www.izip.com/download"
   name "iZip"
   desc "App to manage ZIP, ZIPX, RAR, TAR, 7ZIP and other compressed files"
   homepage "https://www.izip.com/"
+
+  livecheck do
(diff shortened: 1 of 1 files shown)


SMillerDev (dismissed):

Casks/izip.rb
```diff
@@ -3,10 +3,17 @@
...
 
   url "https://www.izip.com/izip.dmg"
-  appcast "https://www.izip.com/download"
   name "iZip"
   desc "App to manage ZIP, ZIPX, RAR, TAR, 7ZIP and other compressed files"
   homepage "https://www.izip.com/"
 
+  livecheck do
+    url "https://www.izip.com/download"
+    strategy :page_match
```
samford, line 12:
```suggestion
```

From the `brew livecheck` documentation:

> **Only use `strategy` when it's necessary**. For example, if livecheck is already using `Git` for a URL, it's not necessary to use `strategy :git`. However, if `Git` applies to a URL but we need to use `PageMatch`, it's necessary to specify `strategy :page_match`.

This URL uses `PageMatch` by default, so `strategy :page_match` isn't necessary and doesn't make a difference at the moment.

samford (changes requested):

miccal (approved):

PR #104912, opened 2021-05-02
sublime-merge: add livecheck

cho-m:

### DIFF
--- a/Casks/sublime-merge.rb
+++ b/Casks/sublime-merge.rb
@@ -4,10 +4,15 @@ cask "sublime-merge" do
 
   url "https://download.sublimetext.com/sublime_merge_build_#{version}_mac.zip",
       verified: "download.sublimetext.com/"
-  appcast "https://www.sublimemerge.com/updates/stable_update_check"
   name "Sublime Merge"
   desc "Git client"
   homepage "https://www.sublimemerge.com/"
+
+  livecheck do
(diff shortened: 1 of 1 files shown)


SMillerDev (dismissed):

Casks/sublime-merge.rb
```diff
@@ -4,11 +4,16 @@
...
   url "https://download.sublimetext.com/sublime_merge_build_#{version}_mac.zip",
       verified: "download.sublimetext.com/"
-  appcast "https://www.sublimemerge.com/updates/stable_update_check"
   name "Sublime Merge"
   desc "Git client"
   homepage "https://www.sublimemerge.com/"
 
+  livecheck do
+    url "https://www.sublimemerge.com/updates/stable_update_check"
+    strategy :page_match
```
samford, line 13:
```suggestion
```

From the `brew livecheck` documentation:

> **Only use `strategy` when it's necessary**. For example, if livecheck is already using `Git` for a URL, it's not necessary to use `strategy :git`. However, if `Git` applies to a URL but we need to use `PageMatch`, it's necessary to specify `strategy :page_match`.

This URL uses `PageMatch` by default, so `strategy :page_match` isn't necessary and doesn't make a difference at the moment.

samford (changes requested):

miccal (approved):

PR #104923, opened 2021-05-02
Update qt-creator from 4.14.1 to 4.15.0 and add livecheck

cho-m:

---

For `livecheck`:
- Can't use `page.match` since https://download.qt.io/official_releases/qtcreator/ shows empty upcoming release number 4.15, so need to at least go through first 2 (4.15 and 4.14) to find latest version
- Using `lazy` with selecting first non-nil to try to avoid doing 9 extra HTTP GETs. The versions are ordered with newest first.

### DIFF
--- a/Casks/qt-creator.rb
+++ b/Casks/qt-creator.rb
@@ -1,12 +1,21 @@
 cask "qt-creator" do
-  version "4.14.1"
-  sha256 "7fb68dff6ed1fd3b02bd82857e9394afa240fb89d2bdfe8a672335319d9906e6"
+  version "4.14.2"
+  sha256 "1a2b1c5a85f253c78085914094e6163e6b6630bdf0b3ad1b6ad6631466c57410"
 
   url "https://download.qt.io/official_releases/qtcreator/#{version.major_minor}/#{version}/qt-creator-opensource-mac-x86_64-#{version}.dmg"
-  appcast "https://download.qt.io/official_releases/qtcreator/",
-          must_contain: version.major_minor
   name "Qt Creator"
+  desc "IDE for application development"
   homepage "https://www.qt.io/developers/"
+
+  livecheck do
+    url "https://download.qt.io/official_releases/qtcreator/"
+    strategy :page_match do |page|
+      page.scan(%r{href="(\d+(?:\.\d+)*)/"}i).lazy.map do |match|
+        version_page = Net::HTTP.get(URI.parse("#{url}#{match[0]}/"))
+        version_page[%r{href="(\d+(?:\.\d+)*)/"}i, 1]
+      end.reject(&:nil?).first
+    end
+  end
 
   depends_on macos: ">= :sierra"
 
--- a/Casks/qt-creator.rb
+++ b/Casks/qt-creator.rb
@@ -1,6 +1,6 @@
 cask "qt-creator" do
-  version "4.14.2"
-  sha256 "1a2b1c5a85f253c78085914094e6163e6b6630bdf0b3ad1b6ad6631466c57410"
+  version "4.15.0"
+  sha256 "94d5e2a85d68beae77d3026b562fd30fff9a928b0b7889f58d0a31121e16c7c5"
 
   url "https://download.qt.io/official_releases/qtcreator/#{version.major_minor}/#{version}/qt-creator-opensource-mac-x86_64-#{version}.dmg"
   name "Qt Creator"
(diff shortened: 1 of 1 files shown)


Casks/qt-creator.rb
```diff
@@ -1,13 +1,22 @@
...
 
+  livecheck do
+    url "https://download.qt.io/official_releases/qtcreator/"
+    strategy :page_match do |page|
+      page.scan(%r{href="(\d+(?:\.\d+)*)/"}i).lazy.map do |match|
+        version_page = Net::HTTP.get(URI.parse("#{url}#{match[0]}/"))
+        version_page[%r{href="(\d+(?:\.\d+)*)/"}i, 1]
+      end.reject(&:nil?).first
+    end
+  end
```
SMillerDev, line 18: @nandahkrishna and @samford?

nandahkrishna: Can possibly use `url "https://download.qt.io/official_releases/qtcreator/?C=N;O=D"` to ensure that we have a descending order for versions.

nandahkrishna: Similarly, we could use `Net::HTTP.get(URI.parse("#{url}#{match[0]}/?C=N;O=D"))` for the subdirectories.

Casks/qt-creator.rb
```diff
@@ -1,13 +1,22 @@
...
+  desc "IDE for application development"
   homepage "https://www.qt.io/developers/"
 
+  livecheck do
+    url "https://download.qt.io/official_releases/qtcreator/"
+    strategy :page_match do |page|
+      page.scan(%r{href="(\d+(?:\.\d+)*)/"}i).lazy.map do |match|
+        version_page = Net::HTTP.get(URI.parse("#{url}#{match[0]}/"))
+        version_page[%r{href="(\d+(?:\.\d+)*)/"}i, 1]
+      end.reject(&:nil?).first
```
nandahkrishna, line 16: I should've also mentioned that there would be additional changes based on my previous comments, so here's a suggestion:
```suggestion
    url "https://download.qt.io/official_releases/qtcreator/?C=N;O=D"
    strategy :page_match do |page|
      page.scan(%r{href="(\d+(?:\.\d+)*)/"}i).lazy.map do |match|
        version_page = Net::HTTP.get(URI.parse("#{url.split("?")[0]}#{match[0]}/C=N;O=D"))
        version_page[%r{href="(\d+(?:\.\d+)*)/"}i, 1]
      end.reject(&:nil?).first
```
Not sure if there's a better way to remove the GET parameters than using `split("?")`.

Casks/qt-creator.rb
samford: The suggested `livecheck` block and feedback in #104889 should apply here as well, since the situation is the same (files in major/minor version directories). In general: use the query string parameters for sorting instead of relying on the default, use `Strategy#page_content` instead of `Net::HTTP#get`, align the regexes with current standards, etc.

I would add that the approach in #104889 is preferable to the `scan.lazy.map` approach here. In my testing, this made three HTTP requests when we should only be making two (one for the `qtcreator` directory listing page and one for the newest major/minor version directory page).

samford, line 18:
```suggestion
  # It's necessary to check within a major/minor version directory
  # (fetching an additional page) to obtain the full version.
  livecheck do
    url "https://download.qt.io/official_releases/qtcreator/?C=M;O=D"
    strategy :page_match do |page|
      # These version directories can sometimes be empty, so this will 
      page.scan(%r{href=["']?v?(\d+(?:\.\d+)+)/?["' >]}i).lazy.map do |match|
        version_page = Homebrew::Livecheck::Strategy.page_content(url.sub("/?", "/#{match[0]}/?"))
        next if version_page[:content].blank?

        versions = version_page[:content].scan(%r{href=["']?v?(\d+(?:\.\d+)+)/?["' >]}i).map(&:first)
        next if versions.blank?

        versions
      end.reject(&:nil?).first
    end
  end
```

Here's a suggestion that synthesizes the approaches in #104889 and this PR. I won't repeat the lengthy explanation from #104889 but some of the same feedback applies here.

samford (changes requested):

NikolausDemmel:
4.15 is out now :-)

Casks/qt-creator.rb
cho-m: Thanks @nandahkrishna and @samford.
I have added the recommendations and will look into update the some other Casks when I have the chance.

miccal (approved):

PR #105192, opened 2021-05-06
feat(framer): add new desktop app url

dalisoft:
See [here](https://www.framer.com/support/using-framer/windows-mac/) for more info.

Also would be nice remove deprecated **Framer X** as they do not support it

**NOTE**: I do not know where is URL of **livecheck** of new Framer Desktop app, if could knew, i would be add

### DIFF
--- a/Casks/framer.rb
+++ b/Casks/framer.rb
@@ -1,14 +1,10 @@
 cask "framer" do
   version "2021.16.5"
+  sha256 :no_check
   
   if Hardware::CPU.intel?
-    sha256 "008279eca248e6c7cad163223b1e7df2de330e62b918216f15aba469cd09d6da"
-
     url "https://updates.framer.com/electron/darwin/x64/Framer.zip"
-    end
   else
-    sha256 "296efac5623ce68a5832b9be20a6483ee2c7d9692778f403c69d489c75042f53"
-
     url "https://updates.framer.com/electron/darwin/arm64/Framer.zip"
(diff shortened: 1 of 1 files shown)


cho-m:
Fixed some syntax issues.

Also, sadly can't verify `sha256` when using unversioned URL.
It would be nice if there was another versioned URL, but I didn't see any from quick glance.

CI is probably going to fail on `livecheck`
Will need to see what is best way to update that later.

dalisoft:
@cho-m I verified `sha256` sums manually by downloading two different files. Please revert these. About versioning, may be we update manually and Framer team later will add some `livecheck` URL?

cho-m:
@dalisoft this is a Homebrew/cask requirement for unversioned URL.
It is part of CI checks done for unversioned URLs.

---

Broken `livecheck` will also fail CI.

dalisoft:
@cho-m Sorry then, i didn't read docs enough yet, so my bad. Thank you for fixing

cho-m:
@dalisoft no problem.

It is probably required for end-user as Framer file will be updated in future versions, which will change SHA and cause install to fail. `sha256 :no_check` avoids unnecessary issues raised by users who aren't familiar with behavior.

---

I will try running CI with `extract_plist` livecheck to see how it behaves.
It would be better to find a webpage/RSS-feed with version information though.

miccal (approved):

PR #105367, opened 2021-05-08
Use File instead of IO for file operations

Rylan12:

See https://github.com/Homebrew/brew/pull/11140

Replace `IO.write` and `IO.read` with `File.write` and `File.read`, respectively, in `j`, `minecraft-server`, and `openzfs`.

Note that I was unable to test this for `openzfs` because it depends on certain macOS versions that I don't have access to (i.e. doesn't work on Big Sur).

### DIFF
--- a/Casks/j.rb
+++ b/Casks/j.rb
@@ -4,11 +4,17 @@ cask "j" do
 
   url "https://www.jsoftware.com/download/j#{version}/install/j#{version}_mac64.zip"
   name "J"
+  desc "Programming language for mathematical, statistical and logical analysis of data"
   homepage "https://www.jsoftware.com/"
 
   apps = %w[jbrk jcon jhs jqt]
   apps.each do |a|
     app "j#{version}/#{a}.app"
+  end
+
+  livecheck do
+    url "https://code.jsoftware.com/wiki/System/Installation"
+    regex(%r{href=.*?/ReleaseNotes/J(\d+(?:\.\d+)*)/?["' >]}i)
   end
 
   installer script: "j#{version}/macos-fix.command"
@@ -30,13 +36,13 @@ cask "j" do
     # Use `readlink` to get full path of symlinked commands.
     commands.each do |c|
       command = "#{staged_path}/j#{version}/bin/#{c}.command"
-      IO.write command, IO.read(command).gsub("$0", '$(/usr/bin/readlink "$0" || /bin/echo "$0")')
+      File.write command, File.read(command).gsub("$0", '$(/usr/bin/readlink "$0" || /bin/echo "$0")')
     end
 
     # Fix relative paths inside App bundles.
     apps.each do |a|
       apprun = "#{appdir}/#{a}.app/Contents/MacOS/apprun"
-      IO.write apprun, IO.read(apprun).gsub(%r{`dirname "\$0"`.*?/bin}, "#{staged_path}/j#{version}/bin")
+      File.write apprun, File.read(apprun).gsub(%r{`dirname "\$0"`.*?/bin}, "#{staged_path}/j#{version}/bin")
     end
   end
 
--- a/Casks/minecraft-server.rb
+++ b/Casks/minecraft-server.rb
@@ -4,9 +4,17 @@ cask "minecraft-server" do
 
   url "https://launcher.mojang.com/v#{version.major}/objects/#{version.after_comma}/server.jar",
       verified: "launcher.mojang.com/"
-  appcast "https://minecraft.net/en-us/download/server/"
   name "Minecraft Server"
-  homepage "https://minecraft.net/"
+  desc "Run a Minecraft multiplayer server"
+  homepage "https://www.minecraft.net/en-us/"
+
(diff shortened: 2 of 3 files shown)


Rylan12:
CC @samford because I'm not sure about some of these `livecheck` blocks (especially `j`)

vitorgalvao (dismissed):

Casks/minecraft-server.rb
```diff
@@ -4,10 +4,18 @@
...
-  appcast "https://minecraft.net/en-us/download/server/"
   name "Minecraft Server"
+  desc "Run a Minecraft multiplayer server"
   homepage "https://minecraft.net/"
 
+  livecheck do
+    url "https://minecraft.net/en-us/download/server/"
+    strategy :page_match do |page|
+      match = page.match(%r{href=.*?/objects/(\h+)/server\.jar[^>]*>minecraft[_-]server[._-]v?(\d+(?:\.\d+)*).jar}i)
+      "#{match[2]},#{match[1]}"
```
samford, line 15:
```suggestion
    url "https://minecraft.net/en-us/download/server/"
    strategy :page_match do |page|
      page.scan(%r{href=.*?/objects/(\h+)/server\.jar[^>]*>minecraft[_-]server[._-]v?(\d+(?:\.\d+)*)\.jar}i)
          .map do |match|
        "#{match[1]},#{match[0]}"
```

The `minecraft.net` URL redirects to `www.minecraft.net`, so we can update the URL to avoid the redirection.

In general, I would say it's better to always use `page.scan(regex).map do |match|` instead of `match = page.match(regex)`, as `#scan` will match all occurrences of a regex but `#match` only matches the first occurrence. `PageMatch` uses `#scan` internally and we simply leave it up to livecheck to go through the matched versions and determine which is newest (using `Version` comparison).

Basically, `#scan` will continue to work properly if the page is updated to list more than one version whereas `#match` could potentially return an incorrect result. It's a small change but it can help reduce the maintenance burden over time.

Casks/openzfs.rb
```diff
@@ -4,8 +4,17 @@
...
   url "https://openzfsonosx.org/w/images/#{version.after_comma[0]}/#{version.after_comma}/OpenZFS_on_OS_X_#{version.before_comma}.dmg"
   name "OpenZFS on OS X"
+  desc "Open source port of OpenZFS"
   homepage "https://openzfsonosx.org/"
 
+  livecheck do
+    url "https://openzfsonosx.org/wiki/Downloads"
+    strategy :page_match do |page|
+      match = page.match(%r{href=['"]?/w/images/\d/(\h+)/openzfs[_-]on[_-]os[_-]x[._-](\d+(?:\.\d+)*)\.dmg}i)
+      "#{match[2]},#{match[1]}"
```
samford, line 14:
```suggestion
      page.scan(%r{href=['"]?/w/images/\d/(\h+)/openzfs[_-]on[_-]os[_-]x[._-](\d+(?:\.\d+)*)\.dmg}i).map do |match|
        "#{match[1]},#{match[0]}"
      end
```

As above, this suggestion uses `#scan` instead of `#match`.

samford (changes requested):

Casks/j.rb
```diff
@@ -4,13 +4,19 @@
...
   homepage "https://www.jsoftware.com/"
 
   apps = %w[jbrk jcon jhs jqt]
   apps.each do |a|
     app "j#{version}/#{a}.app"
   end
 
+  livecheck do
+    url "https://github.com/jsoftware/jsource.git"
+    regex(/^jv?(\d+(?:\.\d+)*)-release$/i)
```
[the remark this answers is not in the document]
Rylan12: Yeah, this is much better. I missed the fact that the `ReleaseNotes` pages were different from the `Installation` pages so I figured that wouldn't work. Glad it does though, this is much better 👍

Casks/minecraft-server.rb
samford: I usually check the `homepage`, `stable`/`url`, etc. when I find a redirection somewhere in a formula/cask and update any URLs that are redirecting. In this case, the `homepage` also redirects to `https://www.minecraft.net/`, so we may want to update that while we're at it.

Rylan12: I'm happy to do whatever you prefer but the reason I split it up was because `brew style` complained that the line was too long.

Bo98 (dismissed):

Bo98:
Thanks!

Casks/minecraft-server.rb
samford: It may be easy to overlook but I addressed the line length issue in the suggestion above by having `.map` on a separate line (indented to align with `.scan`). This is what the current `PageMatch` `strategy` block example in the `brew livecheck` documentation (following the homebrew/cask style) does to work around longer regexes that go past the line length limit.

Rylan12: Ah, gotcha 👍

SMillerDev (approved):

PR #105594, opened 2021-05-12
dotnet: various updates

miccal:
Not sure how best to add a `livecheck` here -- the page:

https://dotnet.microsoft.com/download/dotnet

lists the versions, and opening the "recommended" one takes you to:

https://dotnet.microsoft.com/download/dotnet/5.0

and then clicking on either the "SDK" or "Runtime" pages take you to:

https://dotnet.microsoft.com/download/dotnet/thank-you/sdk-5.0.203-macos-x64-installer

or:

https://dotnet.microsoft.com/download/dotnet/thank-you/runtime-5.0.6-macos-x64-installer

which have the links with all the required `version` parts, but how to extract them?

There is also this page with the direct links:

https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0.6/5.0.6.md

### DIFF
--- a/Casks/dotnet.rb
+++ b/Casks/dotnet.rb
@@ -1,5 +1,5 @@
 cask "dotnet" do
-  version "5.0.6,4e2e2c76-626f-4927-8753-55d47ab79e06:4417509d68777cab1ed5f927e86db82"
+  version "5.0.6,4e2e2c76-626f-4927-8753-55d47ab79e06:24417509d68777cab1ed5f927e86db82"
   sha256 "208eae389ea28d76b6b081f944cecca9ec7f0d59bba6711ade976a8c2bf18994"
 
   url "https://download.visualstudio.microsoft.com/download/pr/#{version.after_comma.before_colon}/#{version.after_colon}/dotnet-runtime-#{version.before_comma}-osx-x64.pkg"
--- a/Casks/dotnet.rb
+++ b/Casks/dotnet.rb
@@ -1,11 +1,16 @@
 cask "dotnet" do
-  version "5.0.5,0e2acfdc-5dad-48a8-935d-6095a0ae2217:3ffa41859dbb1ca40a7e1ff704651178"
-  sha256 "f4f3b02899e48b7cb0f1ed8b99f38db630f2e957d47b96de1ee76444b9201dae"
+  version "5.0.6,4e2e2c76-626f-4927-8753-55d47ab79e06:4417509d68777cab1ed5f927e86db82"
+  sha256 "208eae389ea28d76b6b081f944cecca9ec7f0d59bba6711ade976a8c2bf18994"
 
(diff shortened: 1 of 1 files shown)


miccal:
Ping @samford for his expertise.

miccal:
FYI regarding the addition of the `depends_on macos: ">= :high_sierra"`:

https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0-supported-os.md#macos

Casks/dotnet.rb
```diff
@@ -1,18 +1,22 @@
...
 
   url "https://download.visualstudio.microsoft.com/download/pr/#{version.after_comma.before_colon}/#{version.after_colon}/dotnet-runtime-#{version.before_comma}-osx-x64.pkg"
-  appcast "https://dotnet.microsoft.com/download/dotnet-core"
-  name ".Net Core Runtime"
+  name ".Net Runtime"
+  desc "Developer platform"
   homepage "https://www.microsoft.com/net/core#macos"
 
+  livecheck do
+    skip "No version information available"
```
samford, line 11: From my perspective, we have at least two options:

* Check the [`releases-index.json`](https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json) file (also found in the `dotnet/core` GitHub repository), parse the JSON, identify the latest version where `support-phase` is `current`, fetch the related `releases.json` file for the `channel-version` (e.g., `5.0` for `5.0.6`), use a `strategy` block to match the URL for the `dotnet-runtime-osx-x64.pkg` and assemble the parts into the cask `version` format.
* Use the `releases.json` file for the cask's major/minor version (corresponding with the `channel-version`) as the `url` in the `livecheck` block and do the matching/assembling mentioned above.

The first option requires two HTTP requests but would continue to work when the "current" `channel-version` is updated in the future (e.g., switching to `6.0` in November 2021). The second option only makes one HTTP request but the check wouldn't inform us of when there's a new "current" version, so someone would have to notice this outside of livecheck and bump the cask version.

`dotnet` is pretty notable and major releases only happen once a year (November), so I'm not concerned about the second option being pinned to the cask's current major/minor version; the check will automatically work again when the cask is bumped to a version with a new major/minor. With the first option, we would only benefit from the additional HTTP request once a year but we would be paying for it all year, which doesn't seem worthwhile.

The second option is pretty straightforward and looks like this:

```suggestion
    url "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/#{version.major_minor}/releases.json"
    strategy :page_match do |page|
      page.scan(%r{/download/pr/([^/]+)/([^/]+)/dotnet-runtime-v?(\d+(?:\.\d+)+)-osx-x64\.pkg}i).map do |match|
        "#{match[2]},#{match[0]}:#{match[1]}"
      end
    end
```

samford (changes requested):

Casks/dotnet.rb
miccal: Thank you @samford.

PR #105630, opened 2021-05-12
prey: avoid recommending environment filtering.

### LINKED ISSUE #10786: Remove HOMEBREW_NO_ENV_FILTERING uses
Places we use this within Homebrew:

* [x] [`remove-disabled-formulae` action](https://github.com/Homebrew/actions/blob/aa48b95887a989761d7a234ef1a9a3e6d3803bda/remove-disabled-formulae/action.yml#L18) - can be removed without issue.
* [x] [`brew-script` action](https://github.com/Homebrew/actions/blob/dfc55f5712c0b8bfd57344763a58d6381750c5eb/brew-script/action.yml#L27) - I see the usefulness of this, though the regular Actions `{{ ... }}` system covers most use cases CC @dawidd6 
* [ ] [`prey` Cask](https://github.com/Homebrew/homebrew-cask/blob/e3a962226f40b2b3736b9b2e6fcbc6c18e7db290/Casks/prey.rb#L28) - sounds like they need some way to pass arbitrary user-supplied data to the installation of a Cask. 
  - My (@MikeMcQuaid) suggestion is this uses something like `HOMEBREW_PREY_TOKEN` which is then converted in the Cask to be the correct environment variable that Prey expects.

_Originally posted by @Bo98 in https://github.com/Homebrew/brew/issues/10778#issuecomment-790599632_

MikeMcQuaid:
Instead, use a `preflight` to set the desired API key environment variable from a value that Homebrew doesn't filter.

I have no idea whether this works as I want it to, some @homebrew/cask help/testing would be appreciated 🙇🏻 

Fixes https://github.com/Homebrew/brew/issues/10786

### DIFF
--- a/Casks/prey.rb
+++ b/Casks/prey.rb
@@ -14,6 +14,10 @@ cask "prey" do
 
   pkg "prey-mac-#{version}-x64.pkg"
 
+  preflight do
+    ENV["API_KEY"] = ENV["HOMEBREW_PREY_SETUP_API_KEY"]
+  end
+
   uninstall pkgutil:   "com.prey.agent",
             launchctl: "com.prey.agent"
 
@@ -25,6 +29,6 @@ cask "prey" do
 
     The API key may be set as an environment variable as follows:
 
-      HOMEBREW_NO_ENV_FILTERING=1 API_KEY="foobar123" brew install --cask prey
+      HOMEBREW_PREY_SETUP_API_KEY="foobar123" brew install --cask prey
   EOS
 end


vitorgalvao:
`export API_KEY=<whatever>` before the installation should also work, and it would avoid the roundabout to the `preflight`.

```
export API_KEY="foobar123"
brew install --cask prey
```

Pinging the last people who updated the cask to see if they can test it (because we can’t without a key): @adithyasunil26 @spaz926 @peterlewis.

peterlewis:
> Pinging the last people who updated the cask to see if they can test it (because we can’t without a key): @adithyasunil26 @spaz926 @peterlewis.

Howdy @vitorgalvao! Would love to help but unfortunately I don't have a key either.

peterlewis:
> Pinging the last people who updated the cask to see if they can test it (because we can’t without a key): @adithyasunil26 @spaz926 @peterlewis.

Howdy @vitorgalvao! Would love to help but unfortunately I don't have a key either.

MikeMcQuaid:
> `export API_KEY=<whatever>` before the installation should also work, and it would avoid the roundabout to the `preflight`.

Hmm, I'm pretty sure it won't given that anything run by `brew` has all these environment variables stripped out?

MikeMcQuaid:
Confirmed locally that these credentials are being passed through 🎉

Issue #98231, opened 2021-01-17
Document livecheck strategy :git


vitorgalvao:
It’s being added to casks ([`kotlin-native`](https://github.com/Homebrew/homebrew-cask/blob/8037d1db8d2464d9d4c94dda7c0c09f6411578d6/Casks/kotlin-native.rb#L13); [`adopopenjdk`](https://github.com/Homebrew/homebrew-cask/pull/98130/files)) but [has no documentation yet](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/livecheck.md).

We should also mention that when `url` is from GitHub releases, `livecheck` isn’t generally needed.

PR #105633, opened 2021-05-12
Add Menu Bar Splitter.app v2.0

jaklan:

### DIFF
--- a/Casks/menu-bar-splitter.rb
+++ b/Casks/menu-bar-splitter.rb
@@ -7,11 +7,6 @@ cask "menu-bar-splitter" do
   desc "Utility for keeping menu bar organized"
   homepage "https://github.com/jwhamilton99/menu-bar-splitter"
 
-  livecheck do
-    url "https://github.com/jwhamilton99/menu-bar-splitter"
-    strategy :git
-  end
-
   depends_on macos: ">= :catalina"
 
   app "Menu Bar Splitter.app"
--- a/Casks/menu-bar-splitter.rb
+++ b/Casks/menu-bar-splitter.rb
@@ -1,0 +1,25 @@
+cask "menu-bar-splitter" do
+  version "2.0"
+  sha256 "4f1db97404fe48eaef14b8fd05c938a6ca9e6c6cd546730e25e85b709ed8fa8a"
+
+  url "https://github.com/jwhamilton99/menu-bar-splitter/releases/download/#{version}/menu-bar-splitter-v#{version}.zip"
+  name "Menu Bar Splitter"
+  desc "Utility for keeping menu bar organized"
+  homepage "https://github.com/jwhamilton99/menu-bar-splitter"
+
+  livecheck do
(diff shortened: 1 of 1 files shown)


Casks/menu-bar-splitter.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  url "https://github.com/jwhamilton99/menu-bar-splitter/releases/download/#{version}/menu-bar-splitter-v#{version}.zip"
+  name "Menu Bar Splitter"
+  desc "Utility for keeping menu bar organized"
+  homepage "https://github.com/jwhamilton99/menu-bar-splitter"
+
+  livecheck do
+    url "https://github.com/jwhamilton99/menu-bar-splitter"
+    strategy :git
+  end
```
cho-m, line 13: This looks like the default behavior for GitHub download URLs.

Can you try removing the `livecheck` block and checking if `brew livecheck --cask menu-bar-splitter` picks up the latest version?

jaklan: @cho-m - confirm, it seems to work without the livecheck. Is that actually documented anywhere? I based this solution on some other casks:
https://github.com/Homebrew/homebrew-cask/search?q=%22strategy+git%22

jaklan: Okay, I guess it's described here:

> 1. For formulae: Collect the head, stable, and homepage URLs, in that order.
For casks: Collect the url and homepage URLs, in that order.
> 2. Determine if any strategies apply to the first URL. If not, try the next URL.
> 3. If a strategy can be applied, use it to check for new versions.
> 4. Return the newest version (or an error if versions could not be found at any available URLs).

The second point is quite mysterious to me (what particular strategies are automatically checked?), but anyway - I'll remove it then 😉

cho-m: I don't think we have documented the proper approach for using `:git` strategy: #98231

Excluding the `livecheck` when there are no "problematic" tags is in-line with the main `homebrew-core` repo.

Usually, adding a `livecheck` section for GitHub URL is useful for:
- Bad tags, which require a dedicated `regex(...)` section to avoid
- Pre-releases that are versioned similar to actual releases, which require `:github_latest` strategy.

GitHub URLs have default `livecheck` using `:git` strategy: https://rubydoc.brew.sh/Homebrew/Livecheck/Strategy/Git.html

Sourceforge URLs have default `livecheck` using `:sourceforge` strategy: https://rubydoc.brew.sh/Homebrew/Livecheck/Strategy/Sourceforge.html

I think Bitbucket and Apache also have default `livecheck`.

Most other URLs will require dedicated `livecheck`.
The default strategy applied to these is usually `:page_match`, but will require a `regex(...)` section to work.

jaklan: Big thanks for the clarification, I will definitely keep this in mind when creating any PRs in the future 🙂 And definitely it would be really valuable to see such a detailed explanation in docs!

vitorgalvao (approved):

PR #105636, opened 2021-05-12
Sync triage workflow between cask repos

Rylan12:
This adds the [`triage` workflow](https://github.com/Homebrew/homebrew-cask/blob/master/.github/workflows/triage.yml) to the list of workflows to by synced among the cask repos.

Also, while I'm at it, I noticed that the [`review` workflow](https://github.com/Homebrew/homebrew-cask/blob/master/.github/workflows/review.yml) (which doesn't seem to run very often at all, and maybe used to be disabled) can be merged into the `triage` workflow.


Rylan12:
The `documentation` label was applied as expected in https://github.com/Homebrew/homebrew-cask/pull/105637

SMillerDev (approved):

Issue #71503, opened 2019-10-26
Rename Cask


jhabdas:
As it stands `brew install mongodb` offers up the `mongodb` cask even though it is not official. Please rename `mongodb` cask to `mongodb-app` to avoid developer confusion. Author is OK with this. Ref linked issue.

vitorgalvao:
Thank you for the report.

> Please rename `mongodb` cask to `mongodb-app`

We don’t do the `-app` suffix; it’s confusing (we know because we’ve tried it) and wouldn’t make it clear it’s an unofficial app.

> Author is OK with this.

Authors don’t have direct say in names of casks in official repos. They can influence them through renaming their own apps, but otherwise they follow a strict [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).

This has been partially addressed in https://github.com/Homebrew/homebrew-core/pull/43770#issuecomment-527255004, but that didn’t address the software not being official. We [have a protocol for this situation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md#potentially-misleading-name). Fixing in https://github.com/Homebrew/homebrew-cask/pull/71506.

vitorgalvao:
Small note: please include the cask name in the title, next time.

jhabdas:
Thanks. For contunity https://discourse.brew.sh/t/mongodb-sspl-license/4058/14?u=jos

PR #105689, opened 2021-05-13
Change the name of the unofficial rabbitmq cask

jaklan:

`rabbitmq` is normally installed from homebrew-core via `brew install rabbitmq`. `brew install --cask rabbitmq` install some [unofficial `rabbitmq` app](https://github.com/jpadilla/rabbitmqapp) and can lead to confusion. The change is also driven by the [token name guidelines](https://docs.brew.sh/Cask-Cookbook#potentially-misleading-name).

It's a bit similar issue to [the one with `mongodb`](https://github.com/Homebrew/homebrew-cask/issues/71503), but `rabbitmq` is actually located in homebrew-core, not in a tap - however it can be misleading anyway, so that's why I suggest the change.

### DIFF
--- a/Casks/jpadilla-rabbitmq.rb
+++ b/Casks/jpadilla-rabbitmq.rb
@@ -5,6 +5,7 @@ cask "jpadilla-rabbitmq" do
   url "https://github.com/jpadilla/rabbitmqapp/releases/download/#{version}/RabbitMQ.zip",
       verified: "github.com/jpadilla/rabbitmqapp/"
   name "RabbitMQ"
+  desc "App wrapper for RabbitMQ"
   homepage "https://jpadilla.github.io/rabbitmqapp/"
 
   app "RabbitMQ.app"
--- a/Casks/jpadilla-rabbitmq.rb
+++ b/Casks/jpadilla-rabbitmq.rb
@@ -1,4 +1,4 @@
-cask "rabbitmq" do
+cask "jpadilla-rabbitmq" do
   version "3.6.1-build.1"
   sha256 "1838afcece704ab1d23645d5d44953b809474f1f67ec4b18f3f98d440e5b5aad"
 
--- a/Casks/jpadilla-rabbitmq.rb
+++ b/Casks/jpadilla-rabbitmq.rb
@@ -4,7 +4,7 @@ cask "jpadilla-rabbitmq" do
 
   url "https://github.com/jpadilla/rabbitmqapp/releases/download/#{version}/RabbitMQ.zip",
       verified: "github.com/jpadilla/rabbitmqapp/"
-  name "RabbitMQ"
+  name "RabbitMQ.app"
   desc "App wrapper for RabbitMQ"
   homepage "https://jpadilla.github.io/rabbitmqapp/"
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Needs a `desc`, but other than that seems fine. If it’s unofficial, you’re correct it needs the prefix.

Should probably also submit a PR to the repo once this is merged, to change their instructions.

jaklan:
@vitorgalvao done, when merged I'll create a PR in the original repo

jaklan:
@vitorgalvao I've also changed the `name` to `RabbitMQ.app` - it's recommended to delete the `.app` suffix when selecting a proper token name, but as I understand - in the `name` field we should keep the official vendor name (and it's also less confusing than pure `RabbitMQ`).

vitorgalvao:
> I've also changed the `name` to `RabbitMQ.app`

Please revert that. That we don’t want.

jaklan:
> > I've also changed the `name` to `RabbitMQ.app`
> 
> Please revert that. That we don’t want.

What's the reasoning behind that? As mentioned in [docs](https://docs.brew.sh/Cask-Cookbook#stanza-name):
>Stanza: name
name accepts a UTF-8 string defining the name of the software, including capitalization and punctuation. It is used to help with searchability and disambiguation.
>
>Unlike the token, which is simplified and reduced to a limited set of characters, **the name stanza can include the proper capitalization, spacing and punctuation to match the official name of the software**. For disambiguation purposes, it is recommended to spell out the name of the application, and including the vendor name if necessary. A good example is pycharm-ce, whose name is spelled out as Jetbrains PyCharm Community Edition, even though it is likely never referenced as such anywhere.

and as well here:
> name | the full and proper name defined by the vendor, and any useful alternate names (see Name Stanza Details)

The official name of the app is `RabbitMQ.app`, not `RabbitMQ`. It also works as a distinction between the official `RabbitMQ` and the wrapper.

vitorgalvao:
> The official name of the app is `RabbitMQ.app`, not `RabbitMQ`.

Technically yes. In practice `.app` is like having `Desktop` or `for Mac`. It’s redundant and does nothing to help with search (it would do the opposite!), which is the point of `name`. It also looks like an error because we don’t do that anywhere else.

> It also works as a distinction between the official `RabbitMQ` and the wrapper.

No, it doesn’t. HBC is for installing are macOS apps, that distinction is meaningless.

vitorgalvao (approved):

jaklan:
@vitorgalvao done.

PS 
>No, it doesn’t. HBC is for installing are macOS apps, that distinction is meaningless.

From HBC perspective sure, but I was referring to the general convention of naming such apps with `XYZ.app` suffix to indicate it's a _bundled version of tool XYZ_. But it's more like philosophical discussion, that's not the place for that ;) And ofc there's no sense to break current approach.

vitorgalvao:
> I was referring to the general convention of naming such apps with `XYZ.app` suffix to indicate it's a _bundled version of tool XYZ_.

Sure, but it’s no different than the `Desktop`/`for Mac` we eschew. In can be helpful in some contexts but not in ours—to us it’s noise.

PR #105988, opened 2021-05-18
workflows: enable commit signing for BrewTestBot

Rylan12:
This PR adds the `setup-commit-signing` action to enable commit signing for BrewTestBot, wherever required. See Homebrew/brew#11403.

CC @nandahkrishna

### DIFF
--- a/.github/workflows/bump-unversioned-casks.yml
+++ b/.github/workflows/bump-unversioned-casks.yml
@@ -44,6 +44,12 @@ jobs:
         with:
           username: BrewTestBot
 
+      - name: Set up commit signing
+        id: set-up-commit-signing
+        uses: Homebrew/actions/setup-commit-signing@master
+        with:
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):

SMillerDev (approved):

Rylan12:
@nandahkrishna is there an easy way for us to test this out or should we just merge and assume it will work (I'll try to keep an eye out for failed workflows or unsigned commits in PRs)

nandahkrishna:
Do we have a way to manually run any of these workflows? We could try creating a test PR (perhaps with `bump-unversioned-casks`) and see if the commits are signed.

Rylan12:
So the `bump-unversioned-casks` workflow can be run manually but I didn't realize that it's been manually disabled (looks like for around a month).

I think I'd feel comfortable merging without testing because this is something that works in lots of other places

PR #106031, opened 2021-05-19
Add gnosis-safe 3.1.3

jcvernaleo:


Casks/gnosis-safe.rb
```diff
@@ -0,0 +1,23 @@
+cask "gnosis-safe" do
+  version "3.1.3"
+  sha256 "5f0f94623d4a9bac4db13690fc97993e1ee94bc56354ecc5f67997879aee658e"
+
+  url "https://github.com/gnosis/safe-react/releases/download/v#{version}/Safe-Multisig-#{version}.dmg",
```
vitorgalvao, line 5: I don’t see anything on the homepage linking here.

Casks/gnosis-safe.rb
```diff
@@ -0,0 +1,23 @@
+cask "gnosis-safe" do
+  version "3.1.3"
```
vitorgalvao, line 2: This is not the latest version on Github.

Casks/gnosis-safe.rb
```diff
@@ -0,0 +1,23 @@
...
+  sha256 "5f0f94623d4a9bac4db13690fc97993e1ee94bc56354ecc5f67997879aee658e"
+
+  url "https://github.com/gnosis/safe-react/releases/download/v#{version}/Safe-Multisig-#{version}.dmg",
+      verified: "github.com/gnosis/safe-react/"
+  name "Gnosis Safe"
+  desc "Gnosis Safe Multisig Desktop Application"
+  homepage "https://gnosis-safe.io/"
+
+  livecheck do
```
vitorgalvao, line 11: Likely unnecessary in this case.

Casks/gnosis-safe.rb
```diff
@@ -0,0 +1,23 @@
+cask "gnosis-safe" do
+  version "3.1.3"
+  sha256 "5f0f94623d4a9bac4db13690fc97993e1ee94bc56354ecc5f67997879aee658e"
+
+  url "https://github.com/gnosis/safe-react/releases/download/v#{version}/Safe-Multisig-#{version}.dmg",
+      verified: "github.com/gnosis/safe-react/"
+  name "Gnosis Safe"
+  desc "Gnosis Safe Multisig Desktop Application"
```
vitorgalvao, line 8: From the [`desc` documentation](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/desc.md):

> `desc` is not for app slogans! Vendors’ descriptions tend to be filled with generic adjectives such as “modern” and “lightweight”. Those are meaningless marketing fluff (do you ever see apps proudly describing themselves as outdated and bulky?) which must the deleted. It’s fine to use the information on the software’s website as a starting point, but it will require editing in almost all cases.

“Desktop Application” is redundant; that’s what casks are.

Casks/gnosis-safe.rb
```diff
@@ -0,0 +1,23 @@
+cask "gnosis-safe" do
```
vitorgalvao, line 1: Doesn’t follow the [token reference](https://github.com/Homebrew/brew/blob/master/docs/Cask-Cookbook.md#token-reference).

Casks/gnosis-safe.rb
```diff
@@ -0,0 +1,23 @@
...
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
+
+  auto_updates false
+
+  app "Safe Multisig.app"
+
+  zap trash: [
```
vitorgalvao, line 20: If it’s just one path, it should be a string instead of an array.

vitorgalvao (changes requested):

Casks/gnosis-safe.rb
jcvernaleo: I updated to 3.2.1 which is the last full release (the other tags in the release directory lack builds so they aren't useful).

Casks/gnosis-safe.rb
jcvernaleo: Removed.

Casks/gnosis-safe.rb
jcvernaleo: Updated.  Hopefully this is better.

Casks/gnosis-safe.rb
jcvernaleo: Changed to match the name of the actual app and hopefully follow the standard.

Casks/gnosis-safe.rb
jcvernaleo: Updated.

Casks/safe-multisig.rb
```diff
@@ -0,0 +1,16 @@
...
+  version "3.2.1"
+  sha256 "f100a2be2d9e3dea01f7a612418e5296dca2dd26488dd6cae1dd3b7407900345"
+
+  url "https://github.com/gnosis/safe-react/releases/download/v#{version}/Safe-Multisig-#{version}.dmg",
+      verified: "github.com/gnosis/safe-react/"
+  name "Gnosis Safe Multisig"
+  desc "Ethereum multisig wallet"
+  homepage "https://gnosis-safe.io/"
+
+  auto_updates false
```
vitorgalvao, line 11: Remove this. We only use it when it’s `true`, never when `false`.

vitorgalvao:
Good progress. But if using GitHub for the `livecheck` isn’t good enough (due to tags not having downloads), `livecheck` will need to be done base on the page.

Casks/safe-multisig.rb
jcvernaleo: removed.

jcvernaleo:
Thanks for all the help!

I'm trying to work out a good livecheck scheme for this.  The fact that so many of the tags lack downloadable files complicated it a bunch of course.  Trying to find some canonical place to get the 'best' version.

vitorgalvao:
This will work:

```ruby
  livecheck do
    url "https://api.github.com/repos/gnosis/safe-react/releases"
    strategy :page_match
    regex(/(\d+(?:\.\d+)*).dmg/)
  end
```

Check the URL to see what it outputs. Then we grab on to whatever `.dmg` they have.

Pinging @samford, our resident `livecheck` expert in case there’s something better. Sam, quick recap:

* Some GitHub tags for this app do not have downloads, so they’re no good.
* Checking network connections on the home page does yield a file which has the version information, but it’s https://gnosis-safe.io/component---src-pages-index-tsx-ba83e56c466f979407fd.js and that string at the end might indicate that it’s not a fixed, long-term bet.

Casks/safe-multisig.rb
```diff
@@ -0,0 +1,20 @@
...
+  url "https://github.com/gnosis/safe-react/releases/download/v#{version}/Safe-Multisig-#{version}.dmg",
+      verified: "github.com/gnosis/safe-react/"
+  name "Gnosis Safe Multisig"
+  desc "Ethereum multisig wallet"
+  homepage "https://gnosis-safe.io/"
+
+  livecheck do
+    url "https://api.github.com/repos/gnosis/safe-react/releases"
+    strategy :page_match
+    regex(/(\d+(?:\.\d+)*).dmg/)
```
samford, line 14:
```suggestion
  # It's necessary to check the GitHub releases page and match versions
  # from dmg files, as not all releases provide a dmg file. Normally we avoid
  # checking the releases page and use the Git or, if necessary, GithubLatest
  # strategy but this is an exception.
  livecheck do
    url "https://github.com/gnosis/safe-react/releases"
    regex(/href=.*?Safe-Multisig[._-]v?(\d+(?:\.\d+)+)\.dmg/i)
    strategy :page_match
```

Some feedback on the previous `livecheck` block:

* `strategy` should come after other DSL method calls in a `livecheck` block (i.e., after `regex` in this instance). [The idea behind this is that we sometimes reference variables that are defined in the `livecheck` block within a `strategy` block and the values need to be set before the `strategy` block. We're not doing that here but we still standardize on having `strategy` come after other DSL method calls because it allows us to easily modify an existing `strategy` call to use a block without having to move anything around.]
* I generally try to anchor the start/end of the regex to ensure that we're only matching what we want to be matching. The standard pattern for matching the version from the filename in an `href` attribute on an HTML page does this and that's what I'm using here (though the end isn't as strict as it could be). Alternatively, if we went with the previous api.github.com check, I would have modified the regex to the following: `/["']["']*?Safe-Multisig[._-]v?(\d+(?:\.\d+)+)\.dmg["']/i` (anchoring the start/end with the quotes).
* We use case-insensitive regexes (i.e., the trailing `i` flag in `/.../i`) unless we specifically need it to be case sensitive to match. Case-sensitivity is almost always unnecessary for livecheck and avoiding it prevents checks from breaking due to simple changes in case upstream.
* It's important to remember to escape periods (`\.`) when we intend to match a period, as `.` matches anything.

---

One thing to note is that we don't use `api.github.com` in any other `livecheck` blocks in homebrew/core or homebrew/cask because it's rate limited and the limit is pretty low (60 requests an hour, I believe?). It's higher if we use authentication but we don't do this in livecheck, so we can hit the limit in certain circumstances (or if more checks using api.github.com were added over time). github.com may also do some rate limiting but it's not advertised in response headers and it seems to be much more forgiving in practice.

As a rule, we avoid checking the `releases` page for a GitHub repository if the `Git` strategy is sufficient, as the `releases` page is paginated (10 releases per page) but `Git` provides all the tags in the repository. Sometimes `Git` isn't sufficient and it's necessary to use `GithubLatest`. Between these two strategies, we almost never check a `releases` page itself.

However, for a small number of casks (zero formulae), it's necessary to check the `releases` page and match versions from an attached file (there isn't an alternative to this). This cask is one of those exceptions, so I've added an explanatory comment before the `livecheck` block in the suggestion to make it clear that this behavior shouldn't be copied under normal circumstances.

samford (changes requested):

jcvernaleo:
Thanks.  All those changes you did make sense.

I don't love having to do such a special case here, but that seems to just be the way upstream works.

vitorgalvao (approved):

PR #106089, opened 2021-05-21
Sublime Text 4 released

unitof:

### DIFF
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -5,7 +5,7 @@ cask "sublime-text" do
   url "https://download.sublimetext.com/sublime_text_build_#{version.no_dots}_mac.zip"
   name "Sublime Text"
   desc "Text editor for code, markup and prose"
-  homepage "https://www.sublimetext.com/#{version.major}"
+  homepage "https://www.sublimetext.com"
 
   livecheck do
     url "https://www.sublimetext.com/updates/#{version.major}/stable/appcast_osx.xml"
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -9,11 +9,12 @@ cask "sublime-text" do
 
   livecheck do
     url "https://www.sublimetext.com/updates/#{version.major}/stable_update_check"
-    strategy :sparkle do |item|
-      match = item.version.match(/(\d)(\d+)/)
-      "#{match[1]}.#{match[2]}"
-    end
+    regex(/"latest_version":\s*(\d+)/i)
+    strategy :page_match do |page, regex|
+      match = page.match(regex)[1]
+      "#{match[0]}.#{match[1..]}"
   end
+end
 
   auto_updates true
   conflicts_with cask: "sublime-text-dev"
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -27,6 +27,7 @@ end
   zap trash: [
     "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.sublimetext.#{version.major}.sfl*",
     "~/Library/Application Support/Sublime Text",
+    "~/Library/Application Support/Sublime Text (Safe Mode)",
     "~/Library/Caches/com.sublimetext.#{version.major}",
     "~/Library/Preferences/com.sublimetext.#{version.major}.plist",
     "~/Library/Saved Application State/com.sublimetext.#{version.major}.savedState",
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -13,8 +13,8 @@ cask "sublime-text" do
     strategy :page_match do |page, regex|
       match = page.match(regex)[1]
       "#{match[0]}.#{match[1..]}"
+    end
   end
-end
 
   auto_updates true
   conflicts_with cask: "sublime-text-dev"
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -1,8 +1,8 @@
 cask "sublime-text" do
-  version "3.211"
-  sha256 "531c84e24983927c59dc0c5611f605776f917d1c516af80c69c09ea232d24e01"
+  version "4.107"
+  sha256 "5da344779f92c7eb3ac279a6b3cc8810e86480b3e71fba79d8e09e94a3631602"
 
-  url "https://download.sublimetext.com/Sublime%20Text%20Build%20#{version.no_dots}.dmg"
+  url "https://download.sublimetext.com/sublime_text_build_#{version.no_dots}_mac.zip"
   name "Sublime Text"
   desc "Text editor for code, markup and prose"
   homepage "https://www.sublimetext.com/#{version.major}"
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -5,7 +5,7 @@ cask "sublime-text" do
   url "https://download.sublimetext.com/sublime_text_build_#{version.no_dots}_mac.zip"
   name "Sublime Text"
   desc "Text editor for code, markup and prose"
-  homepage "https://www.sublimetext.com"
+  homepage "https://www.sublimetext.com/"
 
   livecheck do
     url "https://www.sublimetext.com/updates/#{version.major}/stable/appcast_osx.xml"
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -8,7 +8,7 @@ cask "sublime-text" do
   homepage "https://www.sublimetext.com/"
 
   livecheck do
-    url "https://www.sublimetext.com/updates/#{version.major}/stable/appcast_osx.xml"
+    url "https://www.sublimetext.com/updates/#{version.major}/stable_update_check"
     strategy :sparkle do |item|
       match = item.version.match(/(\d)(\d+)/)
       "#{match[1]}.#{match[2]}"
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -24,12 +24,19 @@ cask "sublime-text" do
 
   uninstall quit: "com.sublimetext.#{version.major}"
 
+  # Sublime Text 4 uses `Sublime Text 3` and `com.sublimetext.3` dirs if they exist
+  # Otherwise, it creates `Sublime Text` and `com.sublimetext.4`
+  # More info: https://www.sublimetext.com/docs/side_by_side.html
   zap trash: [
     "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.sublimetext.#{version.major}.sfl*",
     "~/Library/Application Support/Sublime Text",
     "~/Library/Application Support/Sublime Text (Safe Mode)",
+    "~/Library/Application Support/Sublime Text 3",
     "~/Library/Caches/com.sublimetext.#{version.major}",
+    "~/Library/Caches/com.sublimetext.3",
     "~/Library/Preferences/com.sublimetext.#{version.major}.plist",
+    "~/Library/Preferences/com.sublimetext.3.plist",
     "~/Library/Saved Application State/com.sublimetext.#{version.major}.savedState",
+    "~/Library/Saved Application State/com.sublimetext.3.savedState",
   ]
 end
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -8,8 +8,8 @@ cask "sublime-text" do
   homepage "https://www.sublimetext.com/"
 
   livecheck do
-    url "https://www.sublimetext.com/updates/#{version.major}/stable_update_check"
-    regex(/"latest_version":\s*(\d+)/i)
+    url "https://www.sublimetext.com/download"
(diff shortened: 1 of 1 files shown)


unitof:
By default, ST4 uses and writes data to `Sublime Text 3` and `com.sublimetext.3` directories if they exist for a clean drop-in upgrade from ST3.

If they do not, _then_ it creates `Sublime Text 4` and `com.sublimetext.4` directories.

What's the best way to handle this in the zap stanza, considering most users will be upgrading from Sublime Text 3? Should it explicitly find and trash all versions, or only the current major version?

unitof:
I don't understand the error in the latest run: https://github.com/Homebrew/homebrew-cask/pull/106089/checks?check_run_id=2636063271#step:11:13.

Says it gets a 404 when downloading https://download.sublimetext.com/sublime_text_build_4107_mac.zip. However that link just worked for me.

Any way to force a re-check?

unitof:
Weird, 404ing on https://download.sublimetext.com/sublime_text_build_4107_mac.zip again, even though I can download it myself.

fanvinga:
> Weird, 404ing on https://download.sublimetext.com/sublime_text_build_4107_mac.zip again, even though I can download it myself.

The download link of ST4 just worked for me either, but the details of the CI shows the download step finished, maybe not the CDN problem ? And I noticed that the livecheck url is broken for major version 4, you will get 404 error by accessing `https://www.sublimetext.com/updates/4/stable/appcast_osx.xml` , will it cause the problem ?

cho-m:
> And I noticed that the livecheck url is broken for major version 4, you will get 404 error by accessing `https://www.sublimetext.com/updates/4/stable/appcast_osx.xml` , will it cause the problem ?

Yes. The CI error message isn't the greatest, but it refers to `livecheck` URL not working.

The audit check is run with `--online` flag, which compares the version with value returned by `livecheck`

fanvinga:
> > And I noticed that the livecheck url is broken for major version 4, you will get 404 error by accessing `https://www.sublimetext.com/updates/4/stable/appcast_osx.xml` , will it cause the problem ?
> 
> Yes. The CI error message isn't the greatest, but it refers to `livecheck` URL not working.
> 
> The audit check is run with `--online` flag, which compares the version with value returned by `livecheck`

I think the release note and the version info have been migrated to `https://www.sublimetext.com/download`,maybe we need to update the livecheck block and use regex to match latest version string ?

fanvinga:
@cho-m @unitof I've used Charles to figure out how sublime text check for updates, exactly they use json to announce the latest version.

image

## Here is the raw request export from Charles

* For Sublime Text 3
```
curl -sL 'https://www.sublimetext.com/updates/3/stable/updatecheck'
```
```
{
    "latest_version": 3211,
    "update_url": "https://www.sublimetext.com/3",

    "manifest_host": "www.sublimetext.com",
    "update_host": "download.sublimetext.com",

    "manifest_path_osx": "/_pak/sublime_text_osx_3211.manifest.xz",
    "update_path_osx": "/sublime_text_osx_3211.pak.xz",

    "manifest_path_windows_x64": "/_pak/sublime_text_windows_x64_3211.manifest.xz",
    "update_path_windows_x64": "/sublime_text_windows_x64_3211.pak.xz",

    "manifest_path_windows_x32": "/_pak/sublime_text_windows_x32_3211.manifest.xz",
    "update_path_windows_x32": "/sublime_text_windows_x32_3211.pak.xz"
}
```

* For Sublime Text 4
```
curl -sL 'https://www.sublimetext.com/updates/4/stable_update_check'
```
```
{
    "latest_version": 4107,
    "update_url": "https://www.sublimetext.com/download",

    "manifest_host": "download.sublimetext.com",
    "update_host": "download.sublimetext.com",

    "manifest_path_osx": "/_pak/sublime_text_osx_4107.manifest.xz",
    "update_path_osx": "/sublime_text_osx_4107.pak.xz",

    "manifest_path_windows_x64": "/_pak/sublime_text_windows_x64_4107.manifest.xz",
    "update_path_windows_x64": "/sublime_text_windows_x64_4107.pak.xz",

    "manifest_path_windows_x32": "/_pak/sublime_text_windows_x32_4107.manifest.xz",
    "update_path_windows_x32": "/sublime_text_windows_x32_4107.pak.xz"
}
```

Casks/sublime-text.rb
```diff
@@ -1,11 +1,11 @@
...
+  sha256 "5da344779f92c7eb3ac279a6b3cc8810e86480b3e71fba79d8e09e94a3631602"
 
-  url "https://download.sublimetext.com/Sublime%20Text%20Build%20#{version.no_dots}.dmg"
+  url "https://download.sublimetext.com/sublime_text_build_#{version.no_dots}_mac.zip"
   name "Sublime Text"
   desc "Text editor for code, markup and prose"
-  homepage "https://www.sublimetext.com/#{version.major}"
+  homepage "https://www.sublimetext.com/"
 
   livecheck do
```
jrappen, line 10: This should be:

```rb
  appcast "https://www.sublimetext.com/updates/#{version.major}/dev_update_check",
          configuration: version.no_dots
```

instead of:

```rb
  livecheck do
      url "https://www.sublimetext.com/updates/#{version.major}/stable/appcast_osx.xml"
```

Casks/sublime-text.rb
```diff
@@ -25,7 +25,7 @@
 
   zap trash: [
     "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.sublimetext.#{version.major}.sfl*",
-    "~/Library/Application Support/Sublime Text #{version.major}",
+    "~/Library/Application Support/Sublime Text",
```
jrappen, line 28: This should be:

```rb
    "~/Library/Application Support/Sublime Text",
    "~/Library/Application Support/Sublime Text (Safe Mode)",
```

as you can run Sublime Text in safe mode now:

<https://www.sublimetext.com/docs/safe_mode.html>

Casks/sublime-text.rb
fanvinga: Currently people often use the bump-version-pr command to issue a pull request for a version update, I would like to ask what is the difference between appcast and livecheck?

jrappen: Not sure, it is used for the cask of the dev builds:

<https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/sublime-text-dev.rb>

which I had proposed to update in Homebrew/homebrew-cask-versions#8245 a while back.

jrappen: Using sparkle her however isn't needed, though.

fanvinga: Thanks for the explanation, I also agree that sparkle is not necessary here.In that case, maybe we need to just remove the livecheck and only use the json style appcast

unitof: I thought appcast stanzas had been deprecated and replaced by livecheck?

SMillerDev: They are: 
> Note: The livecheck stanza should be preferred in most cases, as it allows casks to be updated automatically.

https://docs.brew.sh/Cask-Cookbook#stanza-appcast

Casks/sublime-text.rb
```diff
@@ -8,7 +8,7 @@
   homepage "https://www.sublimetext.com/"
 
   livecheck do
-    url "https://www.sublimetext.com/updates/#{version.major}/stable/appcast_osx.xml"
+    url "https://www.sublimetext.com/updates/#{version.major}/stable_update_check"
     strategy :sparkle do |item|
```
fanvinga, line 12: According to https://docs.brew.sh/Brew-Livecheck , we can use regex to match version.So I think we can change livecheck block like this.

```
livecheck do
  url "https://www.sublimetext.com/updates/#{version.major}/stable_update_check"
  regex(/\"latest_version\":\s(\d+)/i)
end
```

Casks/sublime-text.rb
```diff
@@ -1,14 +1,14 @@
...
+  sha256 "5da344779f92c7eb3ac279a6b3cc8810e86480b3e71fba79d8e09e94a3631602"
 
-  url "https://download.sublimetext.com/Sublime%20Text%20Build%20#{version.no_dots}.dmg"
+  url "https://download.sublimetext.com/sublime_text_build_#{version.no_dots}_mac.zip"
   name "Sublime Text"
   desc "Text editor for code, markup and prose"
-  homepage "https://www.sublimetext.com/#{version.major}"
+  homepage "https://www.sublimetext.com/"
 
   livecheck do
```
nandahkrishna, line 10:
```
livecheck do
  url "https://www.sublimetext.com/updates/#{version.major}/stable_update_check"
  regex(/"latest_version":\s*(\d+)/i)
    match = page.match(regex)[1]
    "#{match[0]}.#{match[1..]}"
  end
end
```
We should probably use this suggested `livecheck` block. The `regex` suggested here is the one used in `sublime-merge`. We use the `strategy` block to modify the version format from `4107` to `4.107` which is used by the Cask.

unitof: https://github.com/Homebrew/homebrew-cask/commit/06365f50346112170bdfc582efd1066eeac4783b

Casks/sublime-text.rb
unitof: https://github.com/Homebrew/homebrew-cask/commit/81b0362f848c365f2ee0e70a3c65e65574958372

Casks/sublime-text.rb
nandahkrishna: @unitof I updated my suggestion for the `livecheck` block just a short while back, sorry. The updated suggestion should be able to prevent `livecheck` from breaking on major version upgrades.

unitof: Ahh, nice call. On it!

unitof:
> By default, ST4 uses and writes data to `Sublime Text 3` and `com.sublimetext.3` directories if they exist for a clean drop-in upgrade from ST3.
> 
> If they do not, _then_ it creates `Sublime Text` and `com.sublimetext.4` directories.
> 
> Documentation: https://www.sublimetext.com/docs/side_by_side.html
> 
> What's the best way to handle this in the zap stanza, considering most users will be upgrading from Sublime Text 3? Should it explicitly find and trash all versions, or only the current major version?

@nandahkrishna @jrappen @fanvinga Would love your thoughts on the above. I have a zap ready (not pushed) that explicitly zaps both 3 and 4 files.

Since I upgraded from 3, the current zap would not actually delete most ST4 files, since it uses 3’s Application Support if it exists.

fanvinga:
> > By default, ST4 uses and writes data to `Sublime Text 3` and `com.sublimetext.3` directories if they exist for a clean drop-in upgrade from ST3.
> > If they do not, _then_ it creates `Sublime Text` and `com.sublimetext.4` directories.
> > Documentation: https://www.sublimetext.com/docs/side_by_side.html
> > What's the best way to handle this in the zap stanza, considering most users will be upgrading from Sublime Text 3? Should it explicitly find and trash all versions, or only the current major version?
> 
> @nandahkrishna @jrappen @fanvinga Would love your thoughts on the above. I have a zap ready (not pushed) that explicitly zaps both 3 and 4 files.
> 
> Since I upgraded from 3, the current zap would not actually delete most ST4 files, since it uses 3’s Application Support if it exists.

Personally, I would consider the `zap` operation to be the same as removing all traces of the software installation to clean up the environment and prepare it for the next installation.

> I have a zap ready (not pushed) that explicitly zaps both 3 and 4 files.

So this kind of behavior looks good for me, and if a user who do want to keep their configurations, I think `zap` is not a suitable option.

SMillerDev:
> Since I upgraded from 3, the current zap would not actually delete most ST4 files, since it uses 3’s Application Support if it exists.

I believe it should only uninstall files that the app created. It won't be removing `~/Library/Caches` just because it shares that with other apps.

unitof:
> I believe it should only uninstall files that the app created. It won't be removing `~/Library/Caches` just because it shares that with other apps.

That's kind of the tricky part though. If installed fresh, then the Zap as written in this PR will remove everything.

However if you install ST4 on a machine that has ST3 (or upgrade, as most existing Homebrew Cask users likely will), then it inherits and writes its own settings to `Sublime Text 3` and `com.sublimetext.3` folders/files, without renaming them.

unitof:
Passing! Victory!

I will go ahead and push one more commit adding `Sublime Text 3` files/folders to the Zap stanza, as discussed above.

Curious: I've noticed the `test` CI task takes snapshots of various system files and services before/after running the `zap`. Where can I view/compare those? All I can find is the giant mostly unhelpful JSON object at https://github.com/Homebrew/homebrew-cask/pull/106089/checks?check_run_id=2641744858#step:21:7.

jrappen:
> @nandahkrishna @jrappen @fanvinga Would love your thoughts on the above. I have a zap ready (not pushed) that explicitly zaps both 3 and 4 files.

Looks fine to me. You can ask the Sublime developers any remaining questions on their Discord at:

[SublimeHQ Discord](https://discord.gg/D43Pecu)

unitof:
Ready to merge upon review then!

themoah:
any updates on merge ? any help needed ?

SMillerDev (approved):

cho-m (approved):
Thanks @unitof and everyone who helped out

PR #106533, opened 2021-05-29
netbeans-php: remove as same app as `netbeans` cask

cho-m:
~~Created with `brew bump-cask-pr`.~~

**EDIT:** Consider for deletion as seems to be same app as #106532
Netbeans may have collapsed their apps into one.


cho-m:
Hmm... I think this is now installing same thing as `netbeans` Cask.
Let me modify for deletion instead.

miccal (approved):

PR #106570, opened 2021-05-30
jameica: add livecheck and desc

tiiiecherle:

### DIFF
--- a/Casks/jameica.rb
+++ b/Casks/jameica.rb
@@ -5,6 +5,12 @@ cask "jameica" do
   url "https://www.willuhn.de/products/jameica/releases/current/jameica/jameica-macos64-#{version}.zip"
   name "Jameica"
   homepage "https://www.willuhn.de/products/jameica/"
+  
+  livecheck do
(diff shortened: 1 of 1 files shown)


cho-m:
Please fix the style issues reported by CI on this PR and the other opened PRs.

These should have been shown as errors when you ran the `brew style ...` command

Casks/jameica.rb
```diff
@@ -1,13 +1,16 @@
+# typed: false
+# frozen_string_literal: true
+
```
cho-m, line 3: These should be deleted.

There is a slight issue with `brew style` where if you run `brew style` on filename (`jameica.rb`), it may recommend general Ruby style fixes.

Running the command on the Cask name (e.g. `brew style jameica`) uses Homebrew's guidelines.

On Homebrew side, we probably need to fix `brew style` command to deal with filename situation better.

tiiiecherle:
all done ;)

cho-m (approved):
Thanks, @tiiiecherle

PR #83359, opened 2020-05-28
Update miniconda from 4.7.12 to py37_4.8.2

b0ttle:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #106582, opened 2021-05-30
miniconda: add desc and livecheck

cho-m:

---

For livecheck, trying to grab version info from latest shell script
https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
```sh
#!/bin/sh
#
# NAME:  Miniconda3
# VER:   py38_4.9.2
# PLAT:  osx-64
# LINES: 577
# MD5:   6e21e31f666558d485dfdab74c1bc1c8
```

Seems the easiest way, though it will require curl'ing a ~55M file.

A more complicated option would be to regex matching the MD5 checksum for latest shell script inside https://repo.anaconda.com/miniconda/ and then using that to find the versioned shell script that has same MD5 checksum.

### DIFF
--- a/Casks/miniconda.rb
+++ b/Casks/miniconda.rb
@@ -5,7 +5,13 @@ cask "miniconda" do
   url "https://repo.anaconda.com/miniconda/Miniconda3-#{version}-MacOSX-x86_64.sh",
       verified: "repo.anaconda.com/miniconda/"
   name "Continuum Analytics Miniconda"
+  desc "Minimal installer for conda"
   homepage "https://conda.io/miniconda.html"
+
+  livecheck do
+    url "https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh"
+    regex(/^#\s*VER:\s*((?:py\d*[._-])?v?(\d+(?:\.\d+)+))/i)
+  end
 
   auto_updates true
   conflicts_with cask: "miniforge"
--- a/Casks/miniconda.rb
+++ b/Casks/miniconda.rb
@@ -8,9 +8,12 @@ cask "miniconda" do
   desc "Minimal installer for conda"
   homepage "https://conda.io/miniconda.html"
 
+  # This regex restricts matching to a specific Python version. This will need
+  # to be updated when the prefix changes in the latest version at the top of:
+  # https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
   livecheck do
-    url "https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh"
-    regex(/^#\s*VER:\s*((?:py\d*[._-])?v?(\d+(?:\.\d+)+))/i)
+    url "https://repo.anaconda.com/miniconda/"
+    regex(/>\s*Miniconda3-(py38[._-]\d+(?:\.\d+)+)-MacOSX-x86_64\.sh\s*</i)
   end
 
   auto_updates true


vitorgalvao:
Another option would be to run `curl` directly with `--range` before trying to match:

```console
$ curl --range 0-100 'https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh'
#!/bin/sh
#
# NAME:  Miniconda3
# VER:   py38_4.9.2
# PLAT:  osx-64
# LINES: 577
# MD5:   6e21e31f666
```

We don’t even need the first full 100 bytes, but it’s a good idea to have a buffer.

cho-m:
> Another option would be to run `curl` directly with `--range` before trying to match:

@vitorgalvao Do we have a way to do that for `livecheck` block?

Also ping @samford and @nandahkrishna for possible input on livecheck.

Casks/miniconda.rb
```diff
@@ -5,8 +5,14 @@
   url "https://repo.anaconda.com/miniconda/Miniconda3-#{version}-MacOSX-x86_64.sh",
       verified: "repo.anaconda.com/miniconda/"
   name "Continuum Analytics Miniconda"
+  desc "Minimal installer for conda"
   homepage "https://conda.io/miniconda.html"
 
+  livecheck do
+    url "https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh"
+    regex(/^#\s*VER:\s*((?:py\d*[._-])?v?(\d+(?:\.\d+)+))/i)
```
samford, line 13:
```suggestion
  # This regex restricts matching to the specific Python version used in the
  # cask but upstream also supports other versions.
  livecheck do
    url "https://conda.io/en/latest/miniconda_hashes.html"
    regex(/>\s*Miniconda3-(py38[._-]\d+(?:\.\d+)+)-MacOSX-x86_64\.sh\s*</i)
```

samford (changes requested):
> > Another option would be to run `curl` directly with `--range` before trying to match:
> 
> @vitorgalvao Do we have a way to do that for `livecheck` block?

Not yet but this is a specific use case that I already plan to support in the future (see Homebrew/homebrew-core#77685).

My PR to migrate `brew livecheck` to `curl` (from `open-uri`) needs to be merged first  but this should happen in the very near future. I'll also have to create a PR for my local work on expanding the `livecheck` DSL to allow for options to be specified in a `livecheck` block (for strategies, `Curl`, etc.).

That lays the foundation for allowing this in a general sense (i.e., passing options from a `livecheck` block to a strategy to `Strategy#page_content`) but I would need to create a follow-up PR afterward to actually implement support for handling this in `#page_content` (i.e., passing `curl` arguments to `Utils::Curl#curl_args`). That should be easy but I have to put some pieces in place first.

> Also ping @samford and @nandahkrishna for possible input on livecheck.

I've added a suggestion that checks the [latest `miniconda_hashes.html` page](https://conda.io/en/latest/miniconda_hashes.html), as the [main `miniconda` docs page](https://docs.conda.io/en/latest/miniconda.html) only lists versioned filenames (e.g., `Miniconda3-py38_4.9.2-MacOSX-x86_64.sh`) for the `py39` versions and the others are simply given as "latest" (e.g., `Miniconda3-latest-MacOSX-x86_64.sh`). The `regex` in the `livecheck` block currently limits matching to versions that start with `py38` (the version used in the cask) but if we want to match the version for the latest Python version available (i.e., `py39_4.9.2`), we would simply replace `py38` with `pyd+`.

Is there any particular reason why the cask is using `py38` versions instead of `py39`?

Casks/miniconda.rb
cho-m: I believe the issue is that Anaconda determines which Python version they consider for the "latest".

When the version was first introduced #83359, there was `py37` & `py38`, but `py37` was used for the latest as it matched what Anaconda used.

Afterward, in #86755, the version was switched `py38`.

This is why the logic I used is based on the `Miniconda3-latest-MacOSX-x86_64.sh` file, which is a copy of a specific version file.

Inside the latest shell file, there is a line:
```sh
# VER:   py38_4.9.2
```
Which will allow us to detect when `py38` becomes `py39`.

I wasn't able to find another webpage that would allow us to do this outside of querying https://repo.anaconda.com/miniconda/ and matching MD5 hashes.

samford: At the moment, our only choice is to use a fixed Python version in the regex and periodically check to update it. This is a short-term issue as we'll be able to pass a `--range` option to `curl` in the not-too-distant future (after my PR to migrate livecheck to `curl` is finished/merged and I create a PR for my work to allow for configuration options in `livecheck` blocks). I've made a note in my to-do list to revisit this cask and update the `livecheck` block once it's technically possible.

That said, I updated the suggested change above to make the preceding comment a bit clearer and to check the `https://repo.anaconda.com/miniconda/` directory listing page, as it's a few KB lighter than the hashes page. I think this is probably the best we can do for now.

cho-m: That's fine with me. We can reconsider after changes.

If we also have support for some arguments in livecheck url like the download url, it will help livecheck some other Casks.

There are some Casks I've seen that are difficult to livecheck since they require a POST request, or the default user-agent ends up loading a site with Windows-specific version.

samford: My PR to switch livecheck from using `open-uri` to `Utils::Curl` internally was merged today, so that part is done now. I'm planning to create a PR for my branch that adds support for configuration options (in the `livecheck` block DSL and within livecheck) but I have an open Homebrew/brew PR that I'm working to get merged before I create other substantial PRs.

The options PR is intended to handle a variety of use cases that require passing arguments to `curl` (e.g., configuring the user agent, creating POST requests, etc.). It should happen sooner rather than later but I just have to get through my other PR first. I've been keeping track of related formulae/casks that have been brought to my attention but I imagine you know others, so I'll try to remember to mention you in a comment (or message you on Slack) when the feature is finished and merged.

samford (approved):
I don't have approve/merge privileges in Homebrew/cask but this looks good to me at this point 👍

PR #106421, opened 2021-05-28
Ensure zoom postinstall is non-interactive

toobuntu:
This is because `open "$APP_PATH"&` is called from the postinstall
script of the package and we don't want any user intervention there.

**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask {{cask_file}}` is error-free.
- [x] `brew style --fix {{cask_file}}` reports no offenses.

Additionally, **if adding a new cask**:

- [ ] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [ ] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [ ] `brew audit --new-cask {{cask_file}}` worked successfully.
- [ ] `brew install --cask {{cask_file}}` worked successfully.
- [ ] `brew uninstall --cask {{cask_file}}` worked successfully.

PR #106752, opened 2021-06-03
zoom-for-it-admins: non-interactive postinstall

miccal:
See https://github.com/Homebrew/homebrew-cask/pull/106421

### DIFF
--- a/Casks/zoom-for-it-admins.rb
+++ b/Casks/zoom-for-it-admins.rb
@@ -19,6 +19,13 @@ cask "zoom-for-it-admins" do
 
   pkg "ZoomInstallerIT.pkg"
 
+  postflight do
+    # Description: Ensure console variant of postinstall is non-interactive.
+    # This is because `open "$APP_PATH"&` is called from the postinstall
+    # script of the package and we don't want any user intervention there.
+    system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"], must_succeed: false
+  end
+
   uninstall signal:  ["KILL", "us.zoom.xos"],
             pkgutil: "us.zoom.pkg.videmeeting",
             delete:  [


toobuntu:
This package is a little different from the standard Zoom.pkg. I had submitted PR #106559 (closed) and ran into the same CI failure. 

The Zoom.pkg postinstall script opens zoom.us.app by invoking `open "$APP_PATH"&`.
The ZoomInstallerIT.pkg postinstall script can open zoom.us.app in two different ways. It seems from reading the code that `open /Applications/zoom.us.app` is called for new installs, and 
```
osascript <<EOD
tell application "zoom.us" activate
end tell
EOD
```
is called for upgrades.

I don't know why the pkill succeeds for Zoom.pkg but fails for ZoomInstallerIT.pkg. One idea to satisfy the CI is to make it conditional:
```
  postflight do
    # Description: Ensure console variant of postinstall is non-interactive.
    # This is because zoom.us is opened by the postinstall script of the
    # package and we don't want any user intervention there.
    if system_command("ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
      system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"], must_succeed: false
    end
  end
```

Both the straight pkill approach and the conditional approach work in my testing of the zoom package. Unfortunately, I do not have zoom-for-it-admins installed, which means I haven't been able to test and investigate further. For this reason, I closed my zoom-for-it-admins PR after the CI failure.

A final note: The idea for the initial PR was inspired by Debian's approach to Dropbox.
[https://salsa.debian.org/debian/nautilus-dropbox/-/blob/master/debian/patches/non-interactive-update.patch](https://salsa.debian.org/debian/nautilus-dropbox/-/blob/master/debian/patches/non-interactive-update.patch)

PR #22995, opened 2016-07-22
Added Cask for p4 version 2016.1.1411799

mbogh:
#### Adding a new cask
- [x] Checked there aren’t open [pull requests](https://github.com/caskroom/homebrew-cask/pulls) for the same cask.
- [x] Checked there aren’t closed [issues](https://github.com/caskroom/homebrew-cask/issues) where that cask was already refused.
- [x] When naming the cask, followed the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] Commit message includes cask’s name.
- [x] `brew cask audit --download p4` is error-free.
- [x] `brew cask style --fix p4` left no offenses.
- [x] `brew cask install p4` worked successfully.
- [x] `brew cask uninstall p4` worked successfully.

I am not able to verify 32 bit.

PR #106786, opened 2021-06-03
Add p4

prantlf:
- [x] `brew audit --cask homebrew/cask/p4` is error-free.
- [x] `brew style --fix homebrew/cask/p4` reports no offenses.

Additionally, because adding a new cask:

- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues). See below, this is not clear.
- [x] `brew audit --new-cask homebrew/cask/p4` worked successfully.
- [x] `brew install --cask homebrew/cask/p4` worked successfully.
- [x] `brew uninstall --cask homebrew/cask/p4` worked successfully.

Added version 2021.1.2126753.

After having the `p4v` cask available for the visual client, the `p4` cask will allow installing the command-line client too, without all the server tools from the `perforce` cask.

It appears that the `p4` cask was added by #22995, but later removed and replaced with `perforce` by #29180. However, there is no comment there about why. While the server is needed only on one machine, all developers need just the client. It is not necessary installing the server package everywhere. The `p4v` cask does not include the server either.

### DIFF
--- a/Casks/p4.rb
+++ b/Casks/p4.rb
@@ -1,0 +1,28 @@
+cask "p4" do
+  version "2021.1,2126753"
+  sha256 "85c0d2f92b59564ea36a56a8f07b87aecf0b2806637e8b7d69d6edd0d79d6c40"
+
+  url "https://cdist2.perforce.com/perforce/r#{version.major[-2..]}.#{version.minor}/bin.macosx1015x86_64/p4"
+  name "Perforce Helix Command-Line Client (P4)"
+  desc "Use it to gain instant access to operations and complete control over the system"
+  homepage "https://www.perforce.com/products/helix-core-apps/command-line-client"
+
+  livecheck do
+    url "https://www.perforce.com/perforce/doc.current/user/relnotes.txt"
+    strategy :page_match do |page|
+      page.scan(%r{\((\d+(?:\.\d+)+)/(\d+)\)}i).map do |match|
+        "#{match[0]},#{match[1]}"
+      end
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


Casks/p4.rb
```diff
@@ -0,0 +1,31 @@
...
+  container type: :naked
+
+  binary "p4"
+
+  preflight do
+    # brew 3.1.9 saves the downloaded binary under the name of the directory
+    if File.file?("#{staged_path}/bin.macosx1015x86_64")
+      File.rename "#{staged_path}/bin.macosx1015x86_64", "#{staged_path}/p4"
+    end
+  end
```
miccal, line 30: Is this necessary? Does simply having
```
binary "bin.macosx1015x86_64", target: "p4"
```
not work?

prantlf: Yes, it works. Actually, I used it initially, but introduced the renaming in c253afc1df3fbf4347697959f871f382dcd767d2 to have a nicer cackroom :-)

    /usr/local/bin/p4 -> /usr/local/Caskroom/p4/2021.1,2126753/p4*

But it cost me some renaming code in the preflight block, which would not be necessary otherwise.

I do not mind reverting that change. Do you want me to do it?

The result will be:

    /usr/local/bin/p4 -> /usr/local/Caskroom/p4/2021.1,2126753/bin.macosx1015x86_64*

miccal (approved):

Casks/p4.rb
prantlf: Thanks a lot for merging this! :-)

PR #106838, opened 2021-06-06
Update mailtrackerblocker from 0.3.29 to 0.4.1

apparition47:
New version requires a stanza update due to the built-in update mechanism and the generation of some files in `Application Support`.


cho-m (approved):
Thanks, @apparition47

PR #106950, opened 2021-06-09
Add cask for Final Fantasy XIV

stevestreza:
Adds the critically acclaimed MMORPG FINAL FANTASY XIV that has a free trial up to level 60 including the Heavensward expansion.

Reviewers: Please be advised if testing this cask, the downloaded DMG is approximately 2.5 GB.

- [ ] It's unclear if I'm using the version tag correctly here, as I don't think this version string is _actually_ the version of the installer, but the audit tool wouldn't accept with an unversioned URL that would also allow SHA256 verification, and I erred on the side of favoring security, but I could also disable this with `:no_check` and go with a more semantic version.

### DIFF
--- a/Casks/final-fantasy-xiv-online.rb
+++ b/Casks/final-fantasy-xiv-online.rb
@@ -1,4 +1,4 @@
-cask "final-fantasy-xiv" do
+cask "final-fantasy-xiv-online" do
   version "pnvdkzgk77dj10"
   sha256 "a187414e0a0a101b995f08706d1bd341992258e72d535409f1136daaed413f17"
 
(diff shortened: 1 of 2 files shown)


miccal:
```
exception while auditing final-fantasy-xiv: Invalid argument @ io_fread - /private/tmp/20210609-2949-1spku3h
```
I believe this has something to do with large file sizes, but I am not sure.

stevestreza:
> I believe this has something to do with large file sizes, but I am not sure.

Yeah, looking around I see a few reports of this error with similarly large files. The downloaded DMG is over INT32_MAX bytes. This might be a bug in brew itself, or one of its dependencies. I can't reproduce it here, unfortunately, but I'm on macOS 11, not 10.15 like this test runner is.

stevestreza:
After looking at https://github.com/Homebrew/homebrew-cask/issues/107214 which suggests the build check issue is intermittent, I think the check that failed is now passing.

As a safety, I reran every command from the checklist above, including deleting the cached DMG and redownloading it. Every step worked as expected.

miccal:
Thank you @stevestreza.

PR #106996, opened 2021-06-10
Fix livecheck for eclipse-jee

nicerloop:

### DIFF
--- a/Casks/eclipse-jee.rb
+++ b/Casks/eclipse-jee.rb
@@ -4,6 +4,7 @@ cask "eclipse-jee" do
 
   url "https://www.eclipse.org/downloads/download.php?file=/technology/epp/downloads/release/#{version.after_comma.before_colon}/#{version.after_colon}/eclipse-jee-#{version.after_comma.before_colon}-#{version.after_colon}-macosx-cocoa-x86_64.dmg&r=1"
   name "Eclipse IDE for Java EE Developers"
+  desc "Eclipse IDE for Java EE developers"
   homepage "https://eclipse.org/"
 
   livecheck do
--- a/Casks/eclipse-jee.rb
+++ b/Casks/eclipse-jee.rb
@@ -11,8 +11,7 @@ cask "eclipse-jee" do
     url "https://www.eclipse.org/downloads/packages/"
     strategy :page_match do |page|
       page.scan(/Eclipse IDE (\d+-\d+) R Packages/i).map do |release|
-        require "net/http"
-        version_page = Net::HTTP.get(URI.parse("https://projects.eclipse.org/releases/#{release[0]}"))
+        version_page = Homebrew::Livecheck::Strategy.page_content("https://projects.eclipse.org/releases/#{release[0]}")[:content]
         version = version_page.scan(%r{href="/projects/eclipse/releases/(\d+(?:\.\d+)*)"}i)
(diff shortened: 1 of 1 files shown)


Casks/eclipse-jee.rb
```diff
@@ -4,12 +4,14 @@
 
   url "https://www.eclipse.org/downloads/download.php?file=/technology/epp/downloads/release/#{version.after_comma.before_colon}/#{version.after_colon}/eclipse-jee-#{version.after_comma.before_colon}-#{version.after_colon}-macosx-cocoa-x86_64.dmg&r=1"
   name "Eclipse IDE for Java EE Developers"
+  desc "Eclipse IDE for Java EE developers"
   homepage "https://eclipse.org/"
 
   livecheck do
```
SMillerDev, line 10: @samford or @nandahkrishna is `net/http` the best way here?

nandahkrishna: There's a `Homebrew::Livecheck::Strategy.page_content` method which I believe should be used instead of `net/http` in `livecheck` blocks, such as in `qt-creator`.

nicerloop: Hi, I am no Ruby nor Homebrew expert, just hacked a way to grab details through a second URL build from the first one; if there is as better way to do it, I'm all for it.

nicerloop: I will have a look at the linked PR and try to use the proposed way for all eclipse-based products.

samford: We can avoid making a second HTTP request in the strategy block if we simply parse the Eclipse version from the news URL that's present on the downloads page. One minute and I'll provide a suggestion.

nicerloop: The full version is not present on the page, only the major.minor.

nicerloop: But the patch part of the version was used when eclipse had a major.minor for 1 year, now the minor is incremented every 3 month instead of the patch. The patch part could be fixed to '0' to simplify things.

samford: Yes, the version on the downloads page uses a `4.19` format and the releases page uses `4.19.0`. I've chosen to pad the identified version string with trailing `.0` parts when there are fewer than three numeric parts (i.e., `4.19` becomes `4.19.0`, `5` becomes `5.0.0`).

Alternatively, the cask could switch to using the version format without the trailing `.0` and we could avoid this added logic. From the standpoint of `Version` comparison, the trailing `.0` doesn't matter (i.e., `4.19` == `4.19.0`) but it's nice to keep the livecheck and formula/cask version formats aligned.

That said, I ended up doing a complete reimplementation of the `strategy` block and it's pretty significant. Unfortunately, I don't have commit access in homebrew/cask, so I can't push a commit to this branch. Would someone with commit access mind pulling in the commit from my [`eclipse-jee-update-livecheck`](https://github.com/samford/homebrew-cask/tree/eclipse-jee-update-livecheck) branch into this PR?

nicerloop: Hi, is the use of Homebrew::Livecheck::Strategy.page_content instead of net/http sufficient for now? If so, I will also update the other eclipse-based products.

miccal:
Great work @nicerloop, and thanks to @SMillerDev, @nandahkrishna and @samford for their invaluable assistance.

Issue #106948, opened 2021-06-09
Zoom no longer installs or upgrades since last commit on postinstall


nsteinmetz:
### 

- [X] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [X] I have retried my command with `--force`.
- [X] I ran `brew update-reset && brew update` and retried my command.
- [X] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [X] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [X] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

### Description of issue

Post install script of zoom make upgrade or install from scratch failng

It's related to commit https://github.com/Homebrew/homebrew-cask/commit/cd5a6bdc4dc10c62e3f72f5f8ae905dbce5d35e8

### Command that failed

brew install --cask zoom (same for upgrade)

### Output of command with `--verbose --debug`

<details><summary>Click to expand</summary><pre><code>

❯ brew install --cask zoom --verbose --debug
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 2 taps (homebrew/core and homebrew/cask).
==> Updated Formulae
Updated 5 formulae.
==> Updated Casks
Updated 3 casks.

==> Cask::Installer#install
==> Printing caveats
==> Cask::Installer#fetch
==> Downloading https://cdn.zoom.us/prod/5.6.7.1020/Zoom.pkg
/usr/bin/curl --disable --globoff --show-error --user-agent Homebrew/3.1.11\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 11.4\)\ curl/7.64.1 --header Accept-Language:\ en --retry 3 --location --silent --head --request GET https://cdn.zoom.us/prod/5.6.7.1020/Zoom.pkg
Already downloaded: /Users/nsteinmetz/Library/Caches/Homebrew/downloads/aab7602a9c83a72a0af56acdccce2d814fad72c680cea2790f7bee8da345fdab--Zoom.pkg
==> Checking quarantine support
/usr/bin/xattr
/usr/bin/swift /usr/local/Homebrew/Library/Homebrew/cask/utils/quarantine.swift
==> Quarantine is available.
[...]

PR #107009, opened 2021-06-10
zoom: add error handling to postflight

toobuntu:
A more complete fix for #106948

### DIFF
--- a/Casks/zoom.rb
+++ b/Casks/zoom.rb
@@ -31,8 +31,17 @@ cask "zoom" do
     # Description: Ensure console variant of postinstall is non-interactive.
     # This is because `open "$APP_PATH"&` is called from the postinstall
     # script of the package and we don't want any user intervention there.
+    ohai "The Zoom package postinstall script launches the Zoom app"
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention"
     if system_command("ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
-      system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"], must_succeed: false
+      begin
+        retries ||= 0
+        system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"]
+      rescue RuntimeError
+        sleep 0.5
+        retry if (retries += 1) < 3
+        opoo "Unable to forcibly close zoom.us.app"
+      end
     end
   end
 
--- a/Casks/zoom.rb
+++ b/Casks/zoom.rb
@@ -32,14 +32,10 @@ cask "zoom" do
     # This is because `open "$APP_PATH"&` is called from the postinstall
     # script of the package and we don't want any user intervention there.
     retries ||= 3
-    return(raise(RuntimeError)) unless system_command("/bin/ps",
-                                                      args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
+    ohai "The Zoom package postinstall script launches the Zoom app" unless retries < 3
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" unless retries < 3
+    return unless system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
 
-    nfo ||= 0
-    ohai "The Zoom package postinstall script launches the Zoom app" if nfo.zero?
-    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" if nfo.zero?
-    nfo += 1
-    system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
     rescue RuntimeError
       sleep 1
       retry unless (retries -= 1).zero?
--- a/Casks/zoom.rb
+++ b/Casks/zoom.rb
@@ -35,8 +35,10 @@ cask "zoom" do
     return(raise(RuntimeError)) unless system_command("/bin/ps",
                                                       args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
 
-    ohai "The Zoom package postinstall script launches the Zoom app" if retries.equal? 3
-    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" if retries.equal? 3
+    nfo ||= 0
+    ohai "The Zoom package postinstall script launches the Zoom app" if nfo.zero?
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" if nfo.zero?
+    nfo += 1
     system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
     rescue RuntimeError
       sleep 1
--- a/Casks/zoom.rb
+++ b/Casks/zoom.rb
@@ -31,18 +31,17 @@ cask "zoom" do
     # Description: Ensure console variant of postinstall is non-interactive.
     # This is because `open "$APP_PATH"&` is called from the postinstall
     # script of the package and we don't want any user intervention there.
-    ohai "The Zoom package postinstall script launches the Zoom app"
-    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention"
-    if system_command("/bin/ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
-      begin
-        retries ||= 3
-        system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"]
-      rescue RuntimeError
-        sleep 1
-        retry unless (retries -= 1).zero?
-        opoo "Unable to forcibly close zoom.us.app"
-      end
-    end
+    retries ||= 3
+    return(raise(RuntimeError)) unless system_command("/bin/ps",
+                                                      args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
+
+    ohai "The Zoom package postinstall script launches the Zoom app" if retries.equal? 3
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" if retries.equal? 3
+    system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
+    rescue RuntimeError
+      sleep 1
+      retry unless (retries -= 1).zero?
+      opoo "Unable to forcibly close zoom.us.app"
   end
 
   uninstall signal:  ["KILL", "us.zoom.xos"],
--- a/Casks/zoom.rb
+++ b/Casks/zoom.rb
@@ -33,13 +33,13 @@ cask "zoom" do
     # script of the package and we don't want any user intervention there.
     ohai "The Zoom package postinstall script launches the Zoom app"
     ohai "Attempting to close zoom.us.app to avoid unwanted user intervention"
-    if system_command("ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
+    if system_command("/bin/ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
       begin
-        retries ||= 0
+        retries ||= 3
         system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"]
       rescue RuntimeError
-        sleep 0.5
-        retry if (retries += 1) < 3
+        sleep 1
+        retry unless (retries -= 1).zero?
         opoo "Unable to forcibly close zoom.us.app"
       end
     end


toobuntu:
As stated in https://github.com/Homebrew/homebrew-cask/pull/81081#discussion_r412631195, 

> First, there is no way to ignore command execution error inside a code block except wrapping it inside try-rescue. Simply put `must_succeed: false` is ignored here. Instead, you should use something that is guaranteed to succeed:

Being that we are calling `system_command` inside a code block, we should handle errors the ruby way, as `must_succeed: false` is ignored. This PR does so by attempting `pkill` and retrying up to 3x upon error. If the `pkill` still cannot succeed, then we bail out with a warning message.

This was tested by introducing a forced error and repeatedly installing, reinstalling, and uninstalling zoom, as follows:
```
    postflight do
      # Description: Ensure console variant of postinstall is non-interactive.
      # This is because `open "$APP_PATH"&` is called from the postinstall
      # script of the package and we don't want any user intervention there.
      ohai "The Zoom package postinstall script launches the Zoom app"
      ohai "Attempting to close zoom.us.app to avoid unwanted user intervention"
      if system_command("ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
        begin
          retries ||= 0
          system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"]
          # force a runtime error, should exit 1
          system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"]
        rescue RuntimeError
          sleep 1
          retry if (retries += 1) < 3
          opoo "Unable to forcibly close zoom.us.app"
        end
      end
    end
```

```
brew reinstall --no-quarantine --debug zoom
brew reinstall --no-quarantine zoom
```

miccal:
Thank you for all your efforts with this @toobuntu.

vitorgalvao:
Alright, now I feel this is crossing a threshold. It was one thing when doing this only took a bit of code, but now it’s turning into an ever complex and finicky multiline and multicommand workaround.

Seems to me the fix for this is no longer HBC’s domain, but rather “stop using Zoom” or “pressure them to change their shady user-hostile ways”.

Reverting this cask to what it was seems the most sensible solution now.

toobuntu:
That point of view is certainly understandable. Qué será, será.

Is there not some middle ground? The cask in its current state appears to be working without issue.

Both people who reported having the problem in Issue https://github.com/Homebrew/homebrew-cask/issues/106948 reported back that it was fixed after PR https://github.com/Homebrew/homebrew-cask/pull/106959 was merged. So, even if this PR is ultimately rejected, maybe we can leave the cask in its current state?

-----
As a result of the pandemic, Zoom is used by a lot of people. I do think there is effectively zero chance that Zoom will change its code to suit command line upgrades by non-IT admins. The expectation is that users will upgrade from the running app, so reopening the app afterward does make some sense to them I guess. For a console upgrade, however, the automatic opening of the Zoom app after install or upgrade is a nuisance. 

-----
If the postflight convenience gets completely removed (reverting the cask to what it was), then there might still be a solution for those wishing to avoid this nuisance. If zoom-for-it-admins is installed without a corresponding plist file, it will still use default settings. It has two code paths for automatically opening Zoom in its postinstall script. One path tests whether Zoom is running and should not be encountered in homebrew given that the KILL signal is sent to Zoom in the uninstall stanza. The other I haven't encountered yet in practice, seemingly because it is poorly written[1].

That said, there are issues with that package, too. Per https://github.com/autopkg/homebysix-recipes/pull/177#issue-203494371
>Even if it's an enterprise distribution one, we still have to fix it, the postinstall script is a poorly written mess…

Some highlights, from https://github.com/autopkg/homebysix-recipes/issues/99#issuecomment-296760439:
>The postinstall removes !!!! any preexisting /Library/Preferences/us.zoom.config.plist -- so if you are managing configuration preferences for Zoom by placing/managing that file, the installer kills it.

A possible `caveat` that could be added to that cask would be to inform the user to execute `cp -f <config_file_location> /Library/Preferences/` (as done in the postinstall script) after installing or upgrading zoom-for-it-admins if they are using a custom configuration file.

>The postinstall launches the Zoom application.


-----
[1] The problem seems to be in `cut`:
```
installer_pid=$(ps -ef -o pid -o command | grep "/System/Library/CoreServices/Installer.app"\
        | while read each_p
        do
        echo "$each_p"
        done \
        | grep "MacOS" | cut -d " " -f 1)

        echo "installer_pid:$installer_pid">> "$LOG_PATH"
        if [[ $installer_pid > 0 ]]; then
            ps -ef -o pid -o command | grep "installer" | grep "target"
            if [[ $? != 0 ]]; then
            echo "Reopen zoom for manual update">> "$LOG_PATH"
            open /Applications/zoom.us.app
            fi
        fi
```

Casks/zoom.rb
```diff
@@ -31,8 +31,17 @@
     # Description: Ensure console variant of postinstall is non-interactive.
     # This is because `open "$APP_PATH"&` is called from the postinstall
     # script of the package and we don't want any user intervention there.
-    if system_command("ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
-      system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"], must_succeed: false
+    ohai "The Zoom package postinstall script launches the Zoom app"
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention"
+    if system_command("/bin/ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
```
vitorgalvao, line 36: Instead of one more conditional indent, exit early. `return if system_command("/bin/ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")` might work.

Casks/zoom.rb
```diff
@@ -31,8 +31,17 @@
...
+      begin
+        retries ||= 3
+        system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"]
```
cho-m, line 39: As I mentioned in https://github.com/Homebrew/homebrew-cask/issues/107182#issuecomment-860095834, `zoom` uses a package installer that will ignore `appdir` input.

So, this should be fixed to `args: ["-f", "/Applications/zoom.us.app"]` otherwise it will always fail if someone uses sets `--appdir` (either in local command or global environment/homebrew-bundle).

toobuntu: Of course, you are correct. As I understand it, Zoom installs to `/Applications` if the user has administrator privileges and to `~/Applications` if not. Since `pkill` without `-x` will, by default, match any substring, it probably makes sense to match only `/zoom.us.app`.
```suggestion
        system_command "/usr/bin/pkill", args: ["-f", "/zoom.us.app"]
```

Casks/zoom.rb
toobuntu: Done!

Thinking about it some more, given that in Issue #106948 Zoom had not yet fully launched, it seems useful to rescue `ps` and only attempt `pkill` if Zoom is actually running.

Casks/zoom.rb
```diff
@@ -41,7 +41,7 @@
     rescue RuntimeError
       sleep 1
       retry unless (retries -= 1).zero?
-      opoo "Unable to forcibly close zoom.us.app"
+      opoo "Unable to forcibly close zoom.us.app" if retries.zero?
```
vitorgalvao, line 44: The `if` is redundant. Due to the preceding line, this one will never be reached unless retries is `zero`.

toobuntu: Okay, reverted

Casks/zoom.rb
```diff
@@ -31,9 +31,19 @@
...
-    end
+    retries ||= 3
+    return(raise(RuntimeError)) unless system_command("/bin/ps",
+                                                      args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
+
+    nfo ||= 0
+    ohai "The Zoom package postinstall script launches the Zoom app" if nfo.zero?
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" if nfo.zero?
+    nfo += 1
+    system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
```
toobuntu, line 42: In the CI, `ps` and `pkill` can execute out of top-down order. I suppose it's not so important, but seems odd and I'm not sure how to correct it.

Current CI run:
```
42 installer: The install was successful.
43 /bin/ps x
44 ==> Installing artifact of class Cask::Artifact::PostflightBlock
45 ==> The Zoom package postinstall script launches the Zoom app
46 /usr/bin/pkill -f /Applications/zoom.us.app
47 ==> Attempting to close zoom.us.app to avoid unwanted user intervention
48 /usr/local/Homebrew/Library/Homebrew/shims/scm/git --version
49 🍺  zoom was successfully installed!
```

Previous CI run:
```
42 installer: The install was successful.
43 ==> Installing artifact of class Cask::Artifact::PostflightBlock
44 /bin/ps x
45 ==> The Zoom package postinstall script launches the Zoom app
46 /usr/bin/pkill -f /Applications/zoom.us.app
47 ==> Attempting to close zoom.us.app to avoid unwanted user intervention
48 /usr/local/Homebrew/Library/Homebrew/shims/scm/git --version
49 🍺  zoom was successfully installed!
```

I tried this, but command flow in the CI was unchanged:
```
ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" if (nfo += 1).equal? 1
system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"] if (nfo += 1) > 1
```

cho-m: The `system_command`s are run in-order, but the stderr/stdout printing may not be guaranteed to output in order. I think `ohai` uses stdout while the command printing is using stderr.

Though, on side note, `ps` may not even be needed as the exception handling should allow the error raised by `system_command` to be dealt with.

toobuntu: Thanks. Indeed, done!

vitorgalvao (approved):

Casks/zoom.rb
toobuntu: >The system_commands are run in-order, but the stderr/stdout printing may not be guaranteed to output in order.

How about making it a one-liner?
```
 $ git diff -U0 --cached toobuntu/zoom-postflight-error-handling
 diff --git a/Casks/zoom.rb b/Casks/zoom.rb
 index bae149f3a1..534bbb6763 100644
 --- a/Casks/zoom.rb
 +++ b/Casks/zoom.rb
@@ -35,2 +35 @@ cask "zoom" do
 -    ohai "The Zoom package postinstall script launches the Zoom app" unless retries < 3
 -    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" unless retries < 3
 +    ohai "Zoom's postinstall script opens zoom.us; attempting to forcibly close it" unless retries < 3
 @@ -42 +41 @@ cask "zoom" do
 -      opoo "Unable to forcibly close zoom.us.app"
 +      opoo "Unable to forcibly close zoom.us"
```

cho-m: A shorter message is usually better.
It may reduce chances of out-of-order stdout/stderr, but still won't guarantee it.
Also moving `sleep` from exception to after `ohai` could also help a bit (though would add 1s in fastest install scenario).

Casks/zoom.rb
```diff
@@ -31,9 +31,15 @@
...
-    end
+    retries ||= 3
+    ohai "The Zoom package postinstall script launches the Zoom app" unless retries < 3
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" unless retries < 3
+    return unless system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
+
+    rescue RuntimeError
+      sleep 1
+      retry unless (retries -= 1).zero?
+      opoo "Unable to forcibly close zoom.us.app"
```
cho-m, line 42: In Ruby style, I think the `rescue` should be aligned with start of block (`postflight do`) and `end`.

Not sure if cask style check behaves differently.

Casks/zoom.rb
```diff
@@ -31,9 +31,15 @@
     # Description: Ensure console variant of postinstall is non-interactive.
     # This is because `open "$APP_PATH"&` is called from the postinstall
     # script of the package and we don't want any user intervention there.
-    if system_command("ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
-      system_command "/usr/bin/pkill", args: ["-f", "#{appdir}/zoom.us.app"], must_succeed: false
-    end
+    retries ||= 3
+    ohai "The Zoom package postinstall script launches the Zoom app" unless retries < 3
+    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" unless retries < 3
+    return unless system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
```
cho-m, line 37: I think `return unless system_command` can just be simplified to `system_command`.

`system_command` will either succeed or raise an exception.
If it succeeds, then it should just continue execution to end of postflight and exit block normally.

Casks/zoom.rb
toobuntu: Fixed. It was indented because originally it aligned with `begin`, since removed, and `brew style` didn't complain.

Casks/zoom.rb
toobuntu: Right. There was an earlier suggestion to `return` to avoid an additional indent, which is no longer there in the current iteration of this PR.

This should do it
```
diff --git a/Casks/zoom.rb b/Casks/zoom.rb
index bae149f3a1..e0c5847bb5 100644
--- a/Casks/zoom.rb
+++ b/Casks/zoom.rb
@@ -32,14 +32,14 @@ cask "zoom" do
     # This is because `open "$APP_PATH"&` is called from the postinstall
     # script of the package and we don't want any user intervention there.
     retries ||= 3
-    ohai "The Zoom package postinstall script launches the Zoom app" unless retries < 3
-    ohai "Attempting to close zoom.us.app to avoid unwanted user intervention" unless retries < 3
-    return unless system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
+    system_command("/bin/ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
+    ohai "Attempting to close zoom.us; vendor opens it after install" if $CHILD_STATUS.success?
+    system_command "/usr/bin/pkill", args: ["-f", "/Applications/zoom.us.app"]
 
-    rescue RuntimeError
-      sleep 1
-      retry unless (retries -= 1).zero?
-      opoo "Unable to forcibly close zoom.us.app"
+  rescue RuntimeError
+    sleep 1
+    retry unless (retries -= 1).zero?
+    opoo "Unable to close zoom.us"
   end
 
   uninstall signal:  ["KILL", "us.zoom.xos"],
```

I like the idea of only `pkill`-ing if we know zoom.us is actually open, but I agree that simply attempting `pkill` and handling an error also works. If that is maintainers' preference, the line could be  `ohai "Attempting to close zoom.us; vendor opens it after install" unless retries < 3`. I don't feel strongly about it printing before `pkill`. I also don't feel strongly about any particular wording or punctuation, so if maintainers prefer something different, please let me know.

Another alternative is removing `ohai` and `opoo` entirely and operating silently. None of this is terribly important in the grander scheme of things.

Please let me know how to proceed, and if I should squash commits and force push. I haven't done so because there is already one approval, so it seemed better to discuss things first.

Thank you for your attention and comments!

toobuntu: Actually the `if $CHILD_STATUS.success?` isn't necessary. But to really avoid printing the `ohai` line more than once in the event of an error I guess we'd need to toggle a variable and check that. Probably not important enough, so reordering back to
```
ohai "Attempting to close zoom.us; vendor opens it after install" unless retries < 3
system_command("/bin/ps", args: ["x"]).stdout.match?("zoom.us.app/Contents/MacOS/zoom.us")
```
I guess makes more sense if the decision is to keep the `ps`.

vitorgalvao (approved):
If everyone is happy with the current state, merge.

PR #107043, opened 2021-06-10
Update dosbox-x from 0.83.13 to 0.83.14

tomyun:

---

- version bump
- zap added

### DIFF
--- a/Casks/dosbox-x.rb
+++ b/Casks/dosbox-x.rb
@@ -1,7 +1,7 @@
 cask "dosbox-x" do
   if Hardware::CPU.intel?
-    version "0.83.13,20210430230655"
-    sha256 "0854427a1fec933d886a65285d0fe5dc54e5cb842d2a614efd40176ffdd931d5"
+    version "0.83.14,20210531184050"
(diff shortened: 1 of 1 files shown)


Casks/dosbox-x.rb
```diff
@@ -34,4 +34,8 @@
   homepage "https://dosbox-x.com/"
 
   app "dosbox-x/dosbox-x.app"
+
+  zap trash: [
+    "~/Library/Preferences/com.dosbox-x.plist",
+  ]
```
cho-m, line 40: For single trash path, the style preference is a single line:
```suggestion
  zap trash: "~/Library/Preferences/com.dosbox-x.plist"
```

tomyun: updated

cho-m (dismissed):
Thanks, @tomyun

Casks/dosbox-x.rb
tomyun: actually just found another file for zap. list updated again.

cho-m (approved):

PR #107204, opened 2021-06-12
openzfs 2.0.1

yurikoles:
Based on [openzfs-dev](https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/openzfs-dev.rb)

### DIFF
--- a/Casks/openzfs.rb
+++ b/Casks/openzfs.rb
@@ -1,61 +1,69 @@
 cask "openzfs" do
-  version "1.9.4,6f"
-  sha256 "4dc027cec9cb2fcbf572c1dfe5b1a31f0e14d3f55f1b4f8787903f8332db838b"
+  if MacOS.version <= :yosemite
+    version "2.0.1,316"
+    sha256 "b7fc93ae63de5dc348d82b3a3e1b7acc0ea84c43f6f9f17a6c4e64e9fbbff849"
+    pkg "OpenZFSonOsX-#{version.before_comma}-YOSEMITE-10.10.pkg"
+  elsif MacOS.version <= :el_capitan
+    version "2.0.1,315"
+    sha256 "39ca712ce1b1e660659790ec32c4415460f66d2ce5d82cef0a5067a3193e2938"
+    pkg "OpenZFSonOsX-#{version.before_comma}-EL.CAPITAN-10.11.pkg"
+  elsif MacOS.version <= :sierra
+    version "2.0.1,314"
+    sha256 "d1096996a10bc7b8d06a7f037f93b22c45c4c79662ea7bf0a7bb3486e30e95cc"
+    pkg "OpenZFSonOsX-#{version.before_comma}-Sierra-10.12.pkg"
+  elsif MacOS.version <= :high_sierra
+    version "2.0.1,313"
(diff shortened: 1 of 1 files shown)


yurikoles:
Is there a way to `livechek` only `version.before_comma`?

Extra kexts failure is a known issue (See PRs [#10730](https://github.com/Homebrew/homebrew-cask-versions/pull/10730), [#10856](https://github.com/Homebrew/homebrew-cask-versions/pull/10856)), they aren't bundled in this cask.

yurikoles:
Ping?

cho-m:
I think `livecheck` needs to return full version.
May need to just skip

Let's check with @samford @nandahkrishna if they see any way to deal with this.

nandahkrishna:
> I think `livecheck` needs to return full version.

I concur with this, but I'll defer to Sam on whether we should return the version after the comma for this cask alone or skip `livecheck` here altogether. And as @cho-m rightly pointed out, I don't think our `bump` logic works with multiple OS-specific branches.

samford (changes requested):
The `version` in the cask differs based on the environment where `brew livecheck` is run. Even if we create a `livecheck` block that can give the full version, it would only be useful for updating one conditional branch out of eight. `brew livecheck` currently only returns one version and this isn't likely to change anytime soon.

However, it's easy to modify the `livecheck` block DSL to allow us to specify the "current" version to use for comparison (something like `current_version version.before_comma`). This would allow us to have a `livecheck` block for `openzfs` that can check the download page and return `2.0.1` as newest. This couldn't be used for automatic version bumps but it would at least let people know there's a new version.

I implemented this to test the idea ([code here](https://github.com/samford/brew/commit/7bf876828171cdd151b9e3b3afd0ee19a1dfc223)), so I can create a Homebrew/brew PR if we think this is a good idea. DSL modifications have to be merged and tagged in a `brew` version before they can be used in formulae/casks, so we should just update this `livecheck` block to use `skip` in the interim time.

---

GitHub won't allow me to create a suggestion for the `livecheck` block (since there are deleted and unmodified lines) but this is what I would use:

```ruby
  # This `livecheck` block should simply return the version before the comma
  # but we can't override the "current" version that livecheck uses for
  # comparison yet (so the cask version will always appear to be newer than
  # livecheck's version). If/when this is supported in `livecheck` blocks, we
  # can uncomment the check below and update it to specify
  # `version.before_comma` as the "current" version.
  livecheck do
    # url "https://openzfsonosx.org/wiki/Downloads"
    # regex(/href=["']?#v?(\d+(?:\.\d+)+(?:[._-]r\d+)?)["' >]/i)
    skip "Version comparison will always fail for now"
  end
```

The commented out code will give the latest version from the downloads page (e.g., `2.0.1`), so it could be used if/when something like `current_version` is available in the `livecheck` DSL. In my view, it's preferable to continue checking the downloads page instead of the changelog page.

You can pull my [commit](https://github.com/samford/homebrew-cask/commit/f971f09e6bfcf65e7e5ecb1ed3068b0b58b36dae) for this from my [openzfs-2.0.1-update-livecheck](https://github.com/samford/homebrew-cask/tree/openzfs-2.0.1-update-livecheck) branch (based on this PR branch) but I don't have commit access in homebrew/cask, so I can't simply push it here.

PR #107275, opened 2021-06-14
Remove livecheck from discontinued casks

samford:

---

This PR removes `livecheck` blocks for discontinued casks. `brew livecheck` will automatically skip a cask if it's discontinued and a `livecheck` block isn't present in the cask (i.e., having a `livecheck` block in a discontinued cask overrides the automatic skip).

Some of the casks in this PR use `skip "Discontinued"`, which unnecessarily duplicates the automatic skip for discontinued casks. Without these `livecheck` blocks, these casks will continue to be correctly skipped (e.g., `eclipse-javascript : discontinued`).

The other casks in this PR contain working `livecheck` blocks. If a discontinued cask won't be updated to a new version in the future, the `livecheck` block should be removed when the cask is discontinued. This is similar to how we remove `livecheck` blocks from deprecated/disabled formulae in homebrew/core (as livecheck skips them in the same manner when a `livecheck` block isn't present).

### DIFF
--- a/Casks/bootchamp.rb
+++ b/Casks/bootchamp.rb
@@ -5,12 +5,7 @@ cask "bootchamp" do
   url "https://kainjow.com/downloads/BootChamp.zip"
   name "BootChamp"
   desc "Menu bar app to boot into your Boot Camp Windows partition"
-  homepage "https://www.kainjow.com/"
-
-  livecheck do
-    url "https://kainjow.com/updates/bootchamp.xml"
-    strategy :sparkle
-  end
+  homepage "https://kainjow.com/"
 
   depends_on macos: "<= :yosemite"
 
--- a/Casks/candybar.rb
+++ b/Casks/candybar.rb
@@ -7,12 +7,6 @@ cask "candybar" do
   desc "Tool to modify file icons"
   homepage "https://panic.com/blog/candybar-mountain-lion-and-beyond/"
 
-  livecheck do
-    url :homepage
-    strategy :page_match
-    regex(%r{href=.*?/CandyBar%20(\d+(?:\.\d+)*)\.zip}i)
-  end
-
   app "CandyBar.app"
 
   caveats do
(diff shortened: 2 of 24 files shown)


cho-m:
`brew audit --cask` probably needs to be fixed to avoid running `livecheck` when discontinued.

samford:
I'll create a Homebrew/brew PR to account for this condition in `Cask::Audit#check_livecheck_version` but this PR will still fail CI after it's merged because the [homepage for `bootchamp`](https://www.kainjow.com/) gives an SSL error. This can't be fixed because the HTTP URL redirects to the HTTPS URL.

We have a couple options on how to proceed:

* Add the `ci-syntax-only` label to this PR. I audited these casks locally and the only failures are because of the `brew audit` issue and the `bootchamp` SSL issue.
* After the fix for the `brew audit` issue is merged: remove the `bootchamp` commit from this PR and handle it separately in a `ci-syntax-only` PR (i.e., allowing CI to run normally for the other casks in this PR).

cho-m:
> because the [homepage for `bootchamp`](https://www.kainjow.com/) gives an SSL error.

Locally, if I remove  `www`, I don't have any issues
```console
❯ /usr/bin/curl https://kainjow.com/ -I
HTTP/2 200
...

❯ /opt/homebrew/opt/curl/bin/curl https://kainjow.com/ -I
HTTP/2 200
...
```

samford:
My PR to fix the audit issue can be seen at Homebrew/brew#11535. Basically, I modified `Cask::Audit#check_livecheck_version` to respect cask skip conditions by using `Livecheck::SkipConditions#skip_information`. This ensures the audit uses the same skip conditions as the `brew livecheck` command.

> Locally, if I remove `www`, I don't have any issues

Good catch. I should have noticed that the `homepage` uses the `www` subdomain but the `url` doesn't. I've modified this locally to update the `homepage` for `bootchamp` but I'll wait to push this branch until my Homebrew/brew audit PR is merged (so CI doesn't run unnecessarily).

samford:
The Homebrew/brew PR to fix the aforementioned cask audit (Homebrew/brew#11535) has been merged. I rebased this and updated the `bootchamp` homepage and everything is passing now, so this is ready for review.

cho-m (approved):

PR #107328, opened 2021-06-15
Update Spatial Cask installer.

gavinelder:
The following makes a small number of changes to the existing `Spatial` Cli installer. 

- Update formulae to work.
- Add a `Desc` Stanza.
- Update the docs URL to the direct link vs the legacy documentation URL which has a redirect URL.

 

SMillerDev (approved):
Thanks @gavinelder !

PR #107355, opened 2021-06-16
copyclip: fix url, add livecheck, add desc

cho-m:

### DIFF
--- a/Casks/copyclip.rb
+++ b/Casks/copyclip.rb
@@ -1,12 +1,18 @@
 cask "copyclip" do
-  version "2.9.98.3"
+  version "2.9.98.7"
   sha256 :no_check
 
-  url "https://rink.hockeyapp.net/api/2/apps/ffb436060eb379c0cb23097402e92379?format=zip",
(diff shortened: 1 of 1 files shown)


Casks/copyclip.rb
```diff
@@ -1,13 +1,19 @@
...
-  appcast "https://rink.hockeyapp.net/api/2/apps/ffb436060eb379c0cb23097402e92379"
+  url "https://fiplab.com/app-download/CopyClip_#{version.major}.zip"
   name "CopyClip"
+  desc "Clipboard manager"
   homepage "https://fiplab.com/apps/copyclip-for-mac"
 
+  livecheck do
+    url :url
+    strategy :extract_plist do |versions|
+      versions.values.map(&:short_version).compact.first
```
cho-m, line 13: Default livecheck version from `:extract_plist` is **2.9.98.7,29.98.7**

Want to just use the `short_version`.

Could also try accessing value via key like:
`versions["com.fiplab.copyclip#{version.major}"].short_version`

PR #107461, opened 2021-06-18
crossover: don't install on monterey, update zap

yurikoles:
According to [official forum](https://www.codeweavers.com/support/forums/general/?t=27;msg=240166#c3)

### DIFF
--- a/Casks/crossover.rb
+++ b/Casks/crossover.rb
@@ -12,7 +12,7 @@ cask "crossover" do
     strategy :sparkle
   end
 
-  depends_on macos: "<= :monterey"
+  depends_on macos: "<= :big_sur"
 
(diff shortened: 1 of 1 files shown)


Casks/crossover.rb
```diff
@@ -12,17 +12,21 @@
     strategy :sparkle
   end
 
+  depends_on macos: "<= :monterey"
```
cho-m, line 15: I think this would still install on Monterey due to equality.
Probably should be `<= :big_sur`

yurikoles: Thanks, @cho-m, you are right.

cho-m (approved):
Thanks, @yurikoles

PR #107536, opened 2021-06-21
Update mediahuman-youtube-downloader

kvorobyov:

### DIFF
--- a/Casks/mediahuman-youtube-downloader.rb
+++ b/Casks/mediahuman-youtube-downloader.rb
@@ -2,10 +2,10 @@ cask "mediahuman-youtube-downloader" do
   version "3.9.9.57"
   sha256 :no_check
 
-  url "https://www.mediahuman.com/download/YouTubeDownloader.dmg"
+  url "https://www.mediahuman.com/files/YouTubeDownloader.dmg"
   name "MediaHuman YouTube Downloader"
   desc "YouTube videos downloader"
-  homepage "https://www.mediahuman.com/youtube-downloader/"
+  homepage "https://www.mediahuman.com/youtube-video-downloader/"
 
   livecheck do
     url :homepage


Casks/mediahuman-youtube-downloader.rb
```diff
@@ -2,10 +2,10 @@
   version "3.9.9.57"
   sha256 :no_check
 
-  url "https://www.mediahuman.com/download/YouTubeDownloader.dmg"
+  url "https://www.mediahuman.com/files/YouTubeDownloader.dmg"
```
bevanjkay, line 5: @kvorobyov Is there a reason for this change? The MediaHuman website seems to point to the prior URL (/download/)

kvorobyov: The first link is just a placeholder and after several redirects it goes to the second link. Not sure it can affect anything, though.

bevanjkay: @kvorobyov I'm not seeing a redirect on the download URL, however, you are correct that the homepage was being redirected. 

Where did you find the updated URL? I'm not sure we can consider it reliable if it is not published anywhere.

kvorobyov: They have PAD-xml file from the Shareware era, here is the link: http://www.mediahuman.com/pad/YouTubeDownloader.xml and it contains all the links.
And here is the difference:
```
$ curl -I https://www.mediahuman.com/download/YouTubeDownloader.dmg
HTTP/2 200 
server: nginx
date: Mon, 21 Jun 2021 05:26:52 GMT
content-type: application/x-apple-diskimage
content-length: 87870810
expires: Mon, 21 Jun 2021 06:26:52
content-transfer-encoding: binary
cache-control: public
content-disposition: attachment; filename=YouTubeDownloader.dmg
vary: User-Agent
last-modified: Tue, 24 May 2016 14:35:31 GMT

$ curl -I https://www.mediahuman.com/files/YouTubeDownloader.dmg
HTTP/2 200 
server: nginx
date: Mon, 21 Jun 2021 05:27:12 GMT
content-length: 87870810
x-accel-version: 0.01
last-modified: Tue, 15 Jun 2021 13:31:16 GMT
accept-ranges: bytes
cache-control: max-age=86400
expires: Tue, 22 Jun 2021 05:27:12 GMT
```

bevanjkay: @cho-m Any thoughts on this one?

cho-m: No strong opinion as both download checksum-equivalent DMG, are from same domain, and take about same time to download.

The original URL uses Content-Disposition attachment, which isn't the same as a HTTP 301 or 302 redirect. There should be no client-visible redirects with Content-Disposition. What happens on backend would depend on how Nginx server is configured.

kvorobyov: youtube-to-mp3 application from this vendor uses a similar link from the beginning https://github.com/Homebrew/homebrew-cask/blob/master/Casks/youtube-to-mp3.rb

bevanjkay: Thanks @kvorobyov for providing some more information.
I have now approved your changes.

bevanjkay (approved):

PR #107580, opened 2021-06-22
Add Bit Fiddle 1.4.2

nbadal:

### DIFF
--- a/Casks/bit-fiddle.rb
+++ b/Casks/bit-fiddle.rb
@@ -1,0 +1,11 @@
+cask "bit-fiddle" do
+  version "1.4.2"
+  sha256 "38e4b3bda550d61d22d0ca48efce7fc9fadbcda7685a243861742b792eee7ffb"
+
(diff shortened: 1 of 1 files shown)


SMillerDev (changes requested):

nbadal:
Thanks for the assist @cho-m ! I grabbed the scan+match statement directly from the docs, so if that's the preferred way we should probably update those as well

cho-m:
@nbadal it probably doesn't matter which style we exactly use for livecheck block, though, on cask-side, the majority opinion seems to be a more sentence-like style. So, just matching that for now.

Switching `gsub` with `tr` here makes sense given single character translation and is also the common choice in other casks.

I will note that the documentation for `livecheck` on Cask-side (https://docs.brew.sh/Cask-Cookbook#stanza-livecheck) is due for an update as the examples don't actually match some general recommendations (i.e. don't include `strategy` unless necessary, usually prefer `scan` over `match`, etc).

PR #107628, opened 2021-06-23
Add clarinet 0.12.0

CharlieC3:

### DIFF
--- a/Casks/clarinet.rb
+++ b/Casks/clarinet.rb
@@ -4,7 +4,7 @@ cask "clarinet" do
 
   url "https://github.com/hirosystems/clarinet/releases/download/v#{version}/macos-x64.zip"
   name "clarinet"
-  desc "Clarity CLI designed to facilitate smart contract understanding, development, testing and deployment."
+  desc "Clarity CLI for smart contract development, testing and deployment"
   homepage "https://github.com/hirosystems/clarinet/blob/master/README.md"
 
   binary "clarinet"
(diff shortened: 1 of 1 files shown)


cho-m:
Given that this is a binary with open-source license, have you considered submitting it as a Formula in https://github.com/Homebrew/homebrew-core/ ?

CharlieC3:
> Given that this is a binary with open-source license, have you considered submitting it as a Formula in https://github.com/Homebrew/homebrew-core/ ?

Hi @cho-m, indeed I have. Building this project from source takes quite a while for a single binary -- upwards of 5+ minutes of downloading, compiling, and fans spinning.

After discussing with the lead dev behind `clarinet`, I thought it would be best to instead first submit it as a Cask since it takes seconds to install and provides a better overall user experience. I know this goes against what's suggested in the docs, but given the intense install thought it might be eligible for an exception.

cho-m:
When a Formula is added to homebrew-core, we generate pre-built "bottles" for the most common/recent macOS (right now from Mojave to Big Sur). This is what is installed when user runs `brew install ...` as long as they are on a supported OS. It should be pretty quick to install for users.

A lot of our Formulae would normally take a long time to build (some take a couple hours), which is part of reason why our CI nodes create the pre-builts.

CharlieC3:
Interesting, I did not know that. I'm reading up on this now, thanks for the info!

Assuming I don't find any issues getting it to build as a Formula, I think you're right about submitting it there. I'll close this PR for now and reopen it if I hit any blockers trying that method.

PR #107642, opened 2021-06-24
sage: various updates

miccal:
The [downloads page](https://www.sagemath.org/download.html) from the [homepage](https://www.sagemath.org/index.html) seems to point to this [Github releases page](https://github.com/3-manifolds/Sage_macOS/releases) as the recommended way to install Sage.

### DIFF
--- a/Casks/sage.rb
+++ b/Casks/sage.rb
@@ -18,8 +18,8 @@ cask "sage" do
 
   depends_on macos: ">= :high_sierra"
 
-  app "SageMath-#{version.before_comma}.app"
-  binary "#{appdir}/SageMath-#{version.before_comma}.app/Contents/Resources/sage/sage"
+  app "SageMath-#{version.before_comma.dots_to_hyphens}.app"
+  binary "#{appdir}/SageMath-#{version.before_comma.dots_to_hyphens}.app/Contents/Resources/sage/sage"
 
   uninstall quit: "org.sagemath.Sage"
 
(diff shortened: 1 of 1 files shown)


SMillerDev (dismissed):

bevanjkay (dismissed):

miccal:
Note from the `README.rst` on the GitHub page:

> There is a backwards incompatibility with the former Sage app distributed by sagemath.org which is slated for removal from Sage 9.3. That app included a bash script at the top level of the bundle (e.g. /Applications/SageMath.app/sage). But placing a script in that location prevents signing the bundle, since it violates Apple's specifications for the structure of an application bundle. Instead, this app provides a separate installer package which creates an executable bash script named /usr/local/bin/sage that can be used to run sage from a shell or script.

May need to investigate this further and remove the `binary`.

miccal:
Done.

cho-m (dismissed):

miccal:
Now thoroughly tested too (on Big Sur).

SMillerDev (approved):

deatondg (approved):
Thanks for doing this. I wanted to make similar changes in my other PR, but I didn't have the time.

I wouldn't say I've done "rigorous" testing, but installing worked, uninstalling worked, and installing again worked on my Big Sur machine. After the first install, I ran `sage` and got an error about not being able to find the app, but I think this is a 9.3 change and not a problem with the cask. As soon as I opened the app through Finder, I got the system dialogue about opening an app from a third party, and then `sage` worked fine. I did not have this issue on the second install either. Both installs were able to compile and run the project I'm working on.

 Thanks for your work!

miccal:
Thanks @deatondg.

Issue #107289, opened 2021-06-14
[Cleanup] Cask updates for dead URLs, try changing `appcast` to `livecheck`, etc. (June 2021)


cho-m:
Just tracking some work items for various Cask updates. May add extra items as needed.
Free for anyone to work on, but will assign myself to look into them when I have time.

Not including SHA checksum updates as tasks since they are ephemeral issues.

### Dead download URLs (local scan through, so could be false positives).
Unpopular casks should probably be deleted. Popular casks (or new casks) should be updated to functioning URL.

- [x] [`adware-removal-tool`](https://formulae.brew.sh/cask/adware-removal-tool) - #107351
- [x] [`aleo-studio`](https://formulae.brew.sh/cask/aleo-studio) - #107350
- [x] [`appdelete`](https://formulae.brew.sh/cask/appdelete) - #107445
- [x] [`beoplay-software-update`](https://formulae.brew.sh/cask/beoplay-software-update) - #107352
- [x] [`chirp`](https://formulae.brew.sh/cask/chirp) - #107337
- [x] [`copyclip`](https://formulae.brew.sh/cask/copyclip) - #107355
- [x] [`flomo`](https://formulae.brew.sh/cask/flomo) - #107334
- [x] [`flrig`](https://formulae.brew.sh/cask/flrig) - #107411
- [x] [`foreman`](https://formulae.brew.sh/cask/foreman) - #107339
- [x] [`hashbackup`](https://formulae.brew.sh/cask/hashbackup) - #107348
- [x] [`hbuilderx`](https://formulae.brew.sh/cask/hbuilderx) - #108234 
- [x] [`ioquake3`](https://formulae.brew.sh/cask/ioquake3) - #108172 
- [x] [`jamkazam`](https://formulae.brew.sh/cask/jamkazam) - #107335
- [x] [`jump`](https://formulae.brew.sh/cask/jump) - #108163
- [x] [`mapillary-uploader`](https://formulae.brew.sh/cask/mapillary-uploader) - #107297
- [x] [`miditrail`](https://formulae.brew.sh/cask/miditrail) - #108162 
- [x] [`monitorearth`](https://formulae.brew.sh/cask/monitorearth) - #107674
- [x] [`mycloud`](https://formulae.brew.sh/cask/mycloud) - #108165 
- [x] [`scriptql`](https://formulae.brew.sh/cask/scriptql) - #107324
[...]

PR #107674, opened 2021-06-24
Delete monitorearth.rb

vitorgalvao:
Download has been dead for a while.

```
install: 0 (30 days), 0 (90 days), 26 (365 days)
```


carlocab (approved):

cho-m:
One of casks with dead download link listed in #107289.

PR #108001, opened 2021-07-02
Update flipper from 0.96.0 to 0.96.1

chungweileong94:


cho-m (approved):
Thanks for contributing to `homebrew-cask`, @chungweileong94

PR #108034, opened 2021-07-03
google-chrome: alternative download for Intel chips

vladimyr:

---

This cask was altered to use universal download https://github.com/Homebrew/homebrew-cask/pull/93326 but in the meantime, `Hardware::CPU` check got implemented. This is done specifically to reduce download and install size as @i0ntempest mentioned https://github.com/Homebrew/homebrew-cask/pull/93326#issuecomment-731753577

### DIFF
--- a/Casks/google-chrome.rb
+++ b/Casks/google-chrome.rb
@@ -2,7 +2,12 @@ cask "google-chrome" do
   version "91.0.4472.114"
   sha256 :no_check
 
-  url "https://dl.google.com/chrome/mac/universal/stable/GGRO/googlechrome.dmg"
+  if Hardware::CPU.intel?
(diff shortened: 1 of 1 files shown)


i0ntempest:
Doubt they’ll accept this, since they said a while ago universal binary is preferred for everything unless it doesn’t exist for a particular app. Might have changed though.

vladimyr:
> Doubt they’ll accept this, since they said a while ago universal binary is preferred for everything unless it doesn’t exist for a particular app. Might have changed though.

Well here is what (rip)grep says about it:
```
$ rg -F 'Hardware::CPU.intel?' --files-with-matches Casks | wc -l
     106
```

So there is certainly an established practice.

cho-m (approved):

PR #108136, opened 2021-07-05
Update Clocker cask

n0shake:


n0shake:
`brew audit` fails with:

> audit for clocker: failed
 - Download uses GitHub releases, please add an appcast. See https://github.com/Homebrew/homebrew-cask/blob/HEAD/doc/cask_language_reference/stanzas/appcast.md

However, the appcast link is 404ing 😅

bevanjkay (approved):
Thanks @n0shake 

It looks like the app creator has also changed Github tag style (with and without v) and the download URL previously. Hopefully it will remain consistent in future.

PR #108166, opened 2021-07-06
sip: update livecheck, versions

bevanjkay:

### DIFF
--- a/Casks/sip.rb
+++ b/Casks/sip.rb
@@ -2,23 +2,26 @@ cask "sip" do
   if MacOS.version <= :sierra
     version "1.2"
     sha256 "93569421eef761ccdd2784d73e5f201d29e5e22ad6814a7a169f459f460bf4ff"
+  elsif MacOS.version <= :high_sierra
+    version "2.4.1"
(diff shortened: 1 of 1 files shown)


Casks/sip.rb
```diff
@@ -2,23 +2,26 @@
...
   livecheck do
-    url "https://sipapp.io/updates/v#{version.major}/sip.xml"
-    strategy :sparkle
+    url "https://sipapp.io/updates/"
+    regex(/href=.*?sip[._-]v?(\d+(?:\.\d+)+)\.dmg/i)
   end
 
   auto_updates true
-  depends_on macos: ">= :sierra"
+  depends_on macos: ">= :mojave"
```
cho-m, line 24: If we want to keep `MacOS.version <= ...` logic, then probably need to allow `sierra`/`high_sierra`.

Alternatively, can consider simplifying Cask and dropping support for older versions and only track `mojave` onward.

bevanjkay: @cho-m I have kept the `MacOS.version` logic for now, as the cask already had it included, and upstream made it very easy to identify which versions supported the different versions.

SMillerDev (approved):

PR #108252, opened 2021-07-08
workflows/ci: enable concurrency limiting.

MikeMcQuaid:
Let's cancel in-progress builds on the same PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ env:
   HOMEBREW_NO_AUTO_UPDATE: 1
   HOMEBREW_GITHUB_API_TOKEN: ${{ github.token }}
 
+concurrency:
+  group: "${{ github.ref }}"
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   generate-matrix:
     outputs:
--- a/.github/workflows/bump-unversioned-casks.yml
+++ b/.github/workflows/bump-unversioned-casks.yml
@@ -5,7 +5,7 @@ on:
     paths:
       - .github/workflows/bump-unversioned-casks.yml
   schedule:
-    - cron: '*/30 * * * *'
+    - cron: "*/30 * * * *"
   workflow_dispatch:
     inputs:
       clear_cache:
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -7,6 +7,10 @@ env:
   HOMEBREW_NO_AUTO_UPDATE: 1
   HOMEBREW_GITHUB_API_TOKEN: ${{ github.token }}
 
+concurrency:
+  group: "${{ github.ref }}"
```
Rylan12, line 11: Do we need quotes?

.github/workflows/ci.yml
```diff
@@ -7,6 +7,10 @@ env:
...
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
Rylan12, line 12: When is `github.event_name` not `pull_request`?

Rylan12 (commented):
Currently, do we not cancel previous runs for CI?

.github/workflows/ci.yml
MikeMcQuaid: For `master` branch builds.

MikeMcQuaid:
> Currently, do we not cancel previous runs for CI?

I do not believe so but hard to tell. Looks like there's 3+ different ways of doing so in the Homebrew org... CC @dawidd6 in case you have any ideas here too.

.github/workflows/ci.yml
Rylan12: Looks like the CI workflow only runs on `pull_request` atm

https://github.com/Homebrew/homebrew-cask/blob/b72bb2072a8465ca820933d124502564d02bd823/.github/workflows/ci.yml#L3

Rylan12: But if you anticipate wanting it to run on master too, I'm okay with including the conditional even though it will always be true

.github/workflows/ci.yml
dawidd6: Yes, cause there could be some weird characters and the value for this key must be a string.

Rylan12 (approved):

.github/workflows/ci.yml
dawidd6: In this case, `true` is enough.

dawidd6:
Let's use `concurrency` in the whole org for consistency.

dawidd6 (approved):

PR #108293, opened 2021-07-09
Using DNS host for Cloudflare WARP

fedotov-as:

### DIFF
--- a/Casks/cloudflare-warp.rb
+++ b/Casks/cloudflare-warp.rb
@@ -1,5 +1,5 @@
 cask "cloudflare-warp" do
-  version "1.5.294.0,20210701.2"
+  version "1.5.207.0"
   sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


bevanjkay:
@miccal There may not be much benefit in enabling livecheck on this cask. But extract_plist would work, albeit with a ~35mb fetch.

miccal:
@bevanjkay I do not believe it will work in this case as the `.app` is buried in a `.pkg`.

bevanjkay:
@miccal 

```livecheck do
    url :url
    strategy :extract_plist
end
```

returns
> cloudflare-warp : 1.5.294.0,20210701.2 ==> 1.5.207.0,20210616.5

miccal:
Oh really? I dod not think it would work! Please feel free to open a new PR to add this, thanks :)

PR #108374, opened 2021-07-12
update cloudflare-warp version to newest

naveenkrdy:
version 1.5.294.0,20210701.2


bevanjkay:
@naveenkrdy Version `1.5.294.0,20210701.2` is only available as an in-app update at this point, the download URL still points to the previous version.

PR #108422, opened 2021-07-13
microsoft-office: update livecheck for individual apps

vladimyr:

---

Microsoft AutoUpdate uses following feed format:
https://officecdn-microsoft-com.akamaized.net/pr/{{channel-id}}/MacAutoupdate/{{app-id}}-chk.xml
(e.g. https://officecdn-microsoft-com.akamaized.net/pr/C1297A47-86C4-4C1F-97FA-950631F94777/MacAutoupdate/0409XCEL2019-chk.xml)

App IDs were extracted by intercepting MAU network traffic and there is also a separate opensource update checker listing exact same IDs used for verification: https://github.com/pbowden-msft/OfficeCDNCheck/blob/9c33167778d276859b299a0ddec63ded46d12ef4/OfficeCDNCheck#L17-L46

Additional data about particular update is available through following URL format:
https://officecdn-microsoft-com.akamaized.net/pr/{{channel-id}}/MacAutoupdate/{{app-id}}.xml
(e.g. https://officecdn-microsoft-com.akamaized.net/pr/C1297A47-86C4-4C1F-97FA-950631F94777/MacAutoupdate/0409XCEL2019.xml)

I'm fully aware that existing livechecks work but using a stable update feed is better IMO than checking redirects on shortlinks that could change in the future.

### DIFF
--- a/Casks/microsoft-auto-update.rb
+++ b/Casks/microsoft-auto-update.rb
@@ -9,8 +9,8 @@ cask "microsoft-auto-update" do
   homepage "https://docs.microsoft.com/officeupdates/release-history-microsoft-autoupdate"
 
   livecheck do
-    url "https://go.microsoft.com/fwlink/?linkid=830196"
-    strategy :header_match
+    url "https://officecdn-microsoft-com.akamaized.net/pr/C1297A47-86C4-4C1F-97FA-950631F94777/MacAutoupdate/0409MSau04-chk.xml"
+    regex(%r{Version</key>\s*<string>(\d+(?:\.\d+)*)</}i)
   end
 
   auto_updates true
--- a/Casks/microsoft-excel.rb
+++ b/Casks/microsoft-excel.rb
@@ -9,8 +9,8 @@ cask "microsoft-excel" do
   homepage "https://products.office.com/en-US/excel"
 
   livecheck do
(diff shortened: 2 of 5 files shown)


miccal:
> I'm fully aware that existing livechecks work but using a stable update feed is better IMO than checking redirects on shortlinks that could change in the future.

I disagree -- the current `livecheck` blocks are simple in that they use the `url`'s provided from the `homepage`'s, which the ones you have created are much more prone to changing in the future.

I will let other maintainers chime in, though.

bevanjkay:
@miccal I agree that in this instance, using the links from the homepage for livecheck will ensure that we are serving the same version that would be received if installed manually. There is always a chance that the auto-update feed could be subject to staged releases, where I would think the download links would be more likely to be stable.

vitorgalvao:
Agreed with @miccal and @bevanjkay, for the stated reasons. I’ll only add to:

> There is always a chance that the auto-update feed could be subject to staged releases

That this is not rare at all. So while it may be tempting to use the feed and it can be fun to figure out the update check mechanism an app uses (and I suspect @vladimyr might enjoy that process, since they’re always very well researched!), an unversioned URL which redirects to the versioned URL is the most resilient choice.

Thank you anyway, as always.

PR #108537, opened 2021-07-16
Update mysql-shell from 8.0.22 to 8.0.25

philip:


philip:
It appears the macOS 10.5 to macOS 11 change is causing issues. The MySQL team only generates binaries for macOS 11 as of MySQL Shell 8.0.24. 

The official download page[1] includes this text "Packages for Big Sur (11) are compatible with Catalina (10.15)" so I assume it should work fine; but please advise is that's not the case.

[1] https://dev.mysql.com/downloads/shell/

bevanjkay:
@philip CI was failing due to there being no livecheck, we will now see if it passes install/uninstall

bevanjkay (approved):
Thank you @philip!

PR #108546, opened 2021-07-17
Remove version suffix from Pd.app and add `zap`

FnControlOption:


bevanjkay:
@FnControlOption Can you please explain the reason for using `target` on the app stanza?

https://docs.brew.sh/Cask-Cookbook#target-should-only-be-used-in-select-cases

FnControlOption:
Ah, I missed that. I'll update the PR accordingly.

bevanjkay (approved):
Thank you @FnControlOption!

PR #108548, opened 2021-07-17
Update menu-bar-splitter from 2.0 to 2.0.1, remove checksum, and update description

jwhamilton99:

Additionally, **if adding a new cask** (N/A):

### DIFF
--- a/Casks/menu-bar-splitter.rb
+++ b/Casks/menu-bar-splitter.rb
@@ -1,10 +1,10 @@
 cask "menu-bar-splitter" do
   version "2.0.1"
-  sha256 "4f1db97404fe48eaef14b8fd05c938a6ca9e6c6cd546730e25e85b709ed8fa8a"
+  sha256 :no_check
 
-  url "https://github.com/jwhamilton99/menu-bar-splitter/releases/download/#{version}/menu-bar-splitter-v#{version}.zip"
+  url "https://github.com/jwhamilton99/menu-bar-splitter/releases/download/#{version}/menu-bar-splitter.zip"
   name "Menu Bar Splitter"
-  desc "A macOS utility for keeping your menu bar organized."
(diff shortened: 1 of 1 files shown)


bevanjkay:
@jwhamilton99 What is the reason for removing the checksum?

jwhamilton99:
@bevanjkay I wasn't the one who originally made the cask, but I am the developer of the app. I think whoever submitted the cask used the wrong value for the token, and this is the only way I could get it to work without just straight up recreating the cask.

jwhamilton99:
In addition, the download file isn't versioned anymore. In the past it was `menu-bar-splitter-v2.0.zip`, but now it's just `menu-bar-splitter.zip`.

Casks/menu-bar-splitter.rb
```diff
@@ -1,10 +1,10 @@
 cask "menu-bar-splitter" do
-  version "2.0"
-  sha256 "4f1db97404fe48eaef14b8fd05c938a6ca9e6c6cd546730e25e85b709ed8fa8a"
+  version "2.0.1"
+  sha256 :no_check
```
bevanjkay, line 3:
```suggestion
  sha256 "3a48c97cb3594933c5b2f97b0e31373a45847b8aefec7aeceaf7f0ea350b789e"
```

@jwhamilton99 The file itself does not need to be versioned to allow using a checksum, only the URL. Because the URL does change for each version this will prevent there being shasum mismatches when the cask is updated.

bevanjkay (changes requested):

Casks/menu-bar-splitter.rb
jwhamilton99: Ok, thank you. I'm still new to making and updating casks.

bevanjkay (approved):
Thanks @jwhamilton99

PR #108822, opened 2021-07-19
aldente: add depends_on

miccal:
https://github.com/Homebrew/homebrew-cask/pull/108621#issuecomment-882597191


ngocphamm:
Thanks @miccal for the quick update. I just did `brew update` and Aldente still show up as a possible upgrade 🤔 

```
brew outdated
aldente (1.2) != 1.04```

PR #108835, opened 2021-07-19
sqlight: update livecheck

vladimyr:

---

Pass regex as a strategy block param.

### DIFF
--- a/Casks/sqlight.rb
+++ b/Casks/sqlight.rb
@@ -13,10 +13,11 @@ cask "sqlight" do
   # needs to be extracted from the download page.
   livecheck do
     url "https://www.aurvan.com/sqlight/"
-    strategy :page_match do |page|
+    regex(/sqlight-v?(\d+(?:[._]\d+)+)-app\.dmg/i)
+    strategy :page_match do |page, regex|
       js_file = page[%r{src=["']?(/com-aurvan-satva-ui-reactjs-release/static/js/main\.\w+\.chunk.js)["' >]}i, 1]
       version_page = Homebrew::Livecheck::Strategy.page_content("https://www.aurvan.com/#{js_file}")
-      version_page[:content].scan(/sqlight-v?(\d+(?:[._]\d+)+)-app\.dmg/i).flatten.map { |v| v.tr("_", ".") }
+      version_page[:content].scan(regex).map { |match| match&.first&.gsub("_", ".") }
     end
   end
 
--- a/Casks/sqlight.rb
+++ b/Casks/sqlight.rb
@@ -15,7 +15,9 @@ cask "sqlight" do
     url "https://www.aurvan.com/sqlight/"
     regex(/sqlight[._-]v?(\d+(?:[._]\d+)+)-app\.dmg/i)
     strategy :page_match do |page, regex|
-      js_file = page[%r{src=["']?(/com-aurvan-satva-ui-reactjs-release/static/js/main\.\w+\.chunk.js)["' >]}i, 1]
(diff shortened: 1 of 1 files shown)


bevanjkay:
@vladimyr I don't believe this change adds readability, especially given that two regexes are used in the livecheck.

vladimyr:
@bevanjkay It enables the reader to quickly figure out the download filename pattern without the need for reading/parsing the strategy block. I'd say that is an improvement compared to the current state of things.

miccal:
Ping @samford for a review.

Casks/sqlight.rb
```diff
@@ -13,10 +13,11 @@
   # needs to be extracted from the download page.
   livecheck do
     url "https://www.aurvan.com/sqlight/"
-    strategy :page_match do |page|
+    regex(/sqlight-v?(\d+(?:[._]\d+)+)-app\.dmg/i)
```
samford, line 16:
```suggestion
    regex(/sqlight[._-]v?(\d+(?:[._]\d+)+)-app\.dmg/i)
```

Casks/sqlight.rb
```diff
@@ -13,10 +13,11 @@
...
+    strategy :page_match do |page, regex|
       js_file = page[%r{src=["']?(/com-aurvan-satva-ui-reactjs-release/static/js/main\.\w+\.chunk.js)["' >]}i, 1]
```
samford, line 18:
```suggestion
      js_file_path = page[%r{src=["']?(/com-aurvan-satva-ui-reactjs-release/static/js/main\.\w+\.chunk\.js)["' >]}i, 1]
      next [] if js_file_path.blank?

```

Casks/sqlight.rb
```diff
@@ -13,10 +13,11 @@
...
       version_page = Homebrew::Livecheck::Strategy.page_content("https://www.aurvan.com/#{js_file}")
-      version_page[:content].scan(/sqlight-v?(\d+(?:[._]\d+)+)-app\.dmg/i).flatten.map { |v| v.tr("_", ".") }
+      version_page[:content].scan(regex).map { |match| match&.first&.gsub("_", ".") }
```
samford, line 20:
```ruby
      js_file_data = Homebrew::Livecheck::Strategy.page_content("https://www.aurvan.com/#{js_file_path}")
      next [] if js_file_data[:content].blank?

      js_file_data[:content].scan(regex).map { |match| match&.first&.gsub("_", ".") }
```

samford (changes requested):
There's a fair amount of history behind this particular question (i.e., using `#regex` and passing it into a `strategy` block vs. defining regexes in a `strategy` block) and I have strong opinions about it but I won't have the time to properly explain it until later today. I plan to open a brew PR to revisit this discussion in earnest and to revert the examples in the livecheck documentation back to its original state (passing `regex` into `strategy` blocks) but that won't happen until after I've worked through some other things on my livecheck to-do list (the last discussion about this was lengthy and time-consuming).

In the interim time, suffice to say that I'm strongly in favor of using `#regex` and passing the value into a `strategy` block in a majority of cases. The main exception is a `strategy` block that uses more than one regex to assemble the version but one regex can't be treated as the primary one over the others (this has only applied to homebrew/cask so far). `sqlight` isn't one of those exceptional cases, as there's only one regex that's used to identify the version and the one in the `livecheck` block is simply to identify the JS file that gets us to the version information.

With that in mind, I've added some suggestions to fix a syntactical issue and to bring this in line with typical patterns for this type of check (with respect to what we do in homebrew/core).

PR #108913, opened 2021-07-20
fix: SHA256 mismatch for mactracker

Anthonyive:

- [x] The submission is for **a stable version**

I'm sorry I'm not familiar with the debugging process of brew. I only changed the `shasum256` to the correct one.


miccal:
Thank you @Anthonyive.

PR #108961, opened 2021-07-20
Change gcc-arm-embedded to use pkg instead of bz2 and fix livecheck

3rror:

Additionally, **if adding a new cask** (NA here):


3rror:
When installed from the bz2 archive, binaries are not signed and notarized. Because of this, macOS shows a warning saying that the developer cannot be verified.

This PR makes the cask install the official pkg, which is signed and notarized. Contrary to the bz2, the pkg creates the directory /Applications/ARM.

Also, there is a fix for the livecheck.

miccal:
Thank you @3rror.

PR #109048, opened 2021-07-23
Add ARM version of Chromium

yooouuri:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making all changes to a cask, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [ ] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [ ] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [ ] `brew audit --new-cask <cask>` worked successfully.
- [ ] `brew install --cask <cask>` worked successfully.
- [ ] `brew uninstall --cask <cask>` worked successfully.

PR #109061, opened 2021-07-23
Revert "Add ARM version of Chromium"

miccal:
Reverts Homebrew/homebrew-cask#109048

### DIFF
--- a/Casks/chromium.rb
+++ b/Casks/chromium.rb
@@ -1,18 +1,9 @@
 cask "chromium" do
   version "904164"
+  sha256 "6f84b76f0ea3d69e8b516b4a8be92cd78d67906f974e6bdc783749c03f24c110"
 
-  if Hardware::CPU.intel?
-    sha256 "6f84b76f0ea3d69e8b516b4a8be92cd78d67906f974e6bdc783749c03f24c110"
-
-    url "https://commondatastorage.googleapis.com/chromium-browser-snapshots/Mac/#{version}/chrome-mac.zip",
-        verified: "commondatastorage.googleapis.com/chromium-browser-snapshots/Mac/"
-  else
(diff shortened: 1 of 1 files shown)


miccal:
Going to revert this -- the `version` numbers for the `intel` and `arm` versions are not necessarily the same, and the `livecheck` for the arm version would need to use `https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Mac_Arm%2FLAST_CHANGE?alt=media`.

Perhaps this one should be a separate Cask, like `chormium-arm` or something like that?

miccal:
Ping @vitorgalvao and @bevanjkay for thoughts and/or ideas.

vitorgalvao:
> Having a separate `livecheck` block for each version however would be a bad idea I think, since this Cask is already more-or-less constantly out-of-date.

The corollary might be that if the cask already lags behind, separate `livecheck` blocks might not make it worse. If a single user in each architecture commits to checking for it, it would be in great shape.

My initial reaction is that having two casks won’t make the situation better. But I’m very much open to be convinced otherwise.

The ideal solution would be to have `livecheck` and `bump-cask-pr` act on both in one swoop.

bevanjkay:
> My initial reaction is that having two casks won’t make the situation better. But I’m very much open to be convinced otherwise.

@vitorgalvao definitely agree with this statement. A key consideration for me is portability. I understand that our primary concern probably isn't `brew bundle` but my preference would be to keeping installs consistent as much as practically possible. 

> The ideal solution would be to have `livecheck` and `bump-cask-pr` act on both in one swoop.

I think there is value in discussing this further, and perhaps implementing a solution for `livecheck` and `bump-cask-pr` to check all stanzas and update accordingly. As an aside to this, is it worth pursuing a variant or flag for `bump-cask-pr` to raise a PR with the version returned from livecheck automatically? 

I'm happy to look at opening an issue if we want to take this further.

PR #109104, opened 2021-07-25
Update adobe-connect from 11,2021.4.36 to 11,2021.6.27

FnControlOption:

https://helpx.adobe.com/adobe-connect/connect-releasenotes.html
https://helpx.adobe.com/adobe-connect/release-note/adobe-connect-2021-6-application-release-notes.html
https://helpx.adobe.com/adobe-connect/connect-downloads-updates.html

### DIFF
--- a/Casks/adobe-connect.rb
+++ b/Casks/adobe-connect.rb
@@ -1,14 +1,19 @@
 cask "adobe-connect" do
-  version "11,2021.4.36"
-  sha256 "fe56b2e5e000f6f0ca80ecbae8b59798cb0bb9c944f39a556711465a9f4561d6"
+  version "11,2021.6.27"
+  sha256 "4037e55aadfe96964eb200c57913dbc51e3c6ecec64bfe9fbf212febcc73fedf"
 
-  url "https://download.adobe.com/pub/connect/updaters/meeting/#{version.before_comma}/AdobeConnect_#{version.after_comma.dots_to_underscores}.dmg"
+  url "https://download.adobe.com/pub/connect/updaters/meeting/#{version.before_comma.dots_to_underscores}/AdobeConnect_#{version.after_comma.dots_to_underscores}.dmg"
   name "Adobe Connect"
   desc "Virtual meeting client"
   homepage "https://www.adobe.com/products/adobeconnect.html"
 
   livecheck do
-    skip "No version information available"
+    url "https://www.adobe.com/go/ConnectMac11Plus"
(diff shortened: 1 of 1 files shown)


miccal:
Ping @samford for a review.

Casks/adobe-connect.rb
```diff
@@ -1,14 +1,19 @@
...
   name "Adobe Connect"
   desc "Virtual meeting client"
   homepage "https://www.adobe.com/products/adobeconnect.html"
 
   livecheck do
-    skip "No version information available"
+    url "https://www.adobe.com/go/ConnectMac11Plus"
+    regex(%r{/(\d+(?:_\d+)*)/AdobeConnect[._-](\d+(?:_\d+)*)\.dmg}i)
+    strategy :header_match do |headers, regex|
+      match = headers["location"].match(regex)
```
samford, line 14:
```suggestion
    regex(%r{/(\d+(?:[._]\d+)*)/AdobeConnect[._-](\d+(?:[._]\d+)*)\.dmg}i)
    strategy :header_match do |headers, regex|
      match = headers["location"]&.match(regex)
      next [] if match.blank?

```

samford (changes requested):
A couple tweaks:

* For a version that uses underscores, we generally use `[._]` instead of `_` in the `(?:[._]\d+)` part of a version regex, unless we specifically need to only match `_`. Sometimes version formats change from using underscores (or hyphens) to dots, so we use a character class that matches a dot or the current delimiter (`_`). This allows the regex to adapt to this change instead of breaking and needing to be manually updated.
* This returns an empty array if a `location` header isn't present in the response or the regex doesn't match the location URL. This is a little more graceful than allowing `#match` or `#tr` to error on a `nil` value.

FnControlOption:
The `:header_match` strategy doesn't seem to like arrays:

> Error: adobe-connect: Version value must be a string; got a Array ([])

samford:
> The `:header_match` strategy doesn't seem to like arrays:
> 
> > Error: adobe-connect: Version value must be a string; got a Array ([])

Ah yeah, I forgot that `HeaderMatch` only accepts a string from a `strategy` block and can handle a `nil` value. Most other strategies use `PageMatch` internally, which only accepts an array of strings or a string from a `strategy` block and doesn't handle a `nil` value.

I'll create a `brew` PR in a bit to update the other strategies to handle a `nil` value, so we can simply use `next if ...` in `strategy blocks without needing to remember what's an appropriate default value for a given strategy.

miccal:

PR #109111, opened 2021-07-25
adobe-lens-profile-creator: remove

FnControlOption:


bevanjkay:
It doesn't look like Adobe have updated this software since 2012, it isn't compatible with Big Sur, not sure about earlier versions. I would prefer to see this flagged for removal.

SMillerDev (approved):

PR #109120, opened 2021-07-25
Update appgate-sdp-client from 5.3.3 to 5.4.2

FnControlOption:

### DIFF
--- a/Casks/appgate-sdp-client.rb
+++ b/Casks/appgate-sdp-client.rb
@@ -1,15 +1,31 @@
 cask "appgate-sdp-client" do
-  version "5.3.3"
-  sha256 "935c87fcec29c6c7ab28ced0b3da8bb98db7f6b51303c3d651c53b14fc17fcbd"
+  if MacOS.version <= :high_sierra
+    version "5.3.3"
+    sha256 "935c87fcec29c6c7ab28ced0b3da8bb98db7f6b51303c3d651c53b14fc17fcbd"
+  else
+    version "5.4.2"
+    sha256 "d0880947d92035cb2783f995f69404085096326b24d8d58ba7f83a0411ee8efe"
+
+    livecheck do
+      url :homepage
+      strategy :page_match do |page|
+        page_regex = %r{href=["']?/(support/software-defined-perimeter-support/sdp-v?\d+(?:-\d+)*)["' >]}i
+        page.scan(page_regex).flat_map do |match|
+          version_page = Homebrew::Livecheck::Strategy.page_content("https://www.appgate.com/#{match[0]}")
+          version_regex = %r{href=.*?/Appgate[._-]SDP[._-](\d+(?:\.\d+)*)[._-]Installer\.dmg}i
+          version_page[:content].scan(version_regex).map(&:first)
+        end
+      end
+    end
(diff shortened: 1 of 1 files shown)


miccal:
Ping @samford for a review.

bevanjkay:
@FnControlOption It looks like the download link requires either a cookie or a header on the request to succeed, I couldn't get it to fetch locally either.

Casks/appgate-sdp-client.rb
```diff
@@ -1,15 +1,31 @@
...
+    livecheck do
+      url :homepage
+      strategy :page_match do |page|
+        page_regex = %r{href=["']?/(support/software-defined-perimeter-support/sdp-v?\d+(?:-\d+)*)["' >]}i
+        page.scan(page_regex).flat_map do |match|
+          version_page = Homebrew::Livecheck::Strategy.page_content("https://www.appgate.com/#{match[0]}")
+          version_regex = %r{href=.*?/Appgate[._-]SDP[._-](\d+(?:\.\d+)*)[._-]Installer\.dmg}i
+          version_page[:content].scan(version_regex).map(&:first)
+        end
+      end
```
samford, line 18:
```suggestion
      regex(%r{href=.*?/Appgate-SDP[._-](\d+(?:\.\d+)+)[._-]Installer\.dmg}i)
      strategy :page_match do |page, regex|
        support_versions =
          page.scan(%r{href=["']?([^"' >]*?/software-defined-perimeter-support/sdp[._-]v?(\d+(?:[.-]\d+)+))["' >]}i)
              .sort_by { |match| Version.new(match[1]) }
        next [] if support_versions.blank?

        # Assume the last-sorted version is newest
        version_page_path, = support_versions.last

        # Check the page for the newest major/minor version, which links to the
        # latest disk image file (containing the full version in the file name)
        version_page = Homebrew::Livecheck::Strategy.page_content(
          URI.join("https://www.appgate.com/", version_page_path),
        )
        next [] if version_page[:content].blank?

        version_page[:content].scan(regex).map(&:first)
      end
```

samford (changes requested):
Some of the [commentary for `aqua-data-studio`](https://github.com/Homebrew/homebrew-cask/pull/109121#pullrequestreview-714922181) applies here as well, so I'll focus on the differences:

* The `livecheck` block in this PR checks each of the major/minor version pages (since the `#page_content` call is in a `#flap_map` block) and this isn't necessary to find the latest version. I've reworked this to only check the page for the newest major/minor version. [If memory serves, there's only one exceptional formula/cask where we have to check version pages in a loop and we make sure to break out of the loop as soon as we've found version information.]
* I view the disk image regex as the primary regex (i.e., the one that identifies the full version) in this context, so I've moved it into a `#regex` call.
* Ideally we would use `URI#join` on the real homepage URL but `url` is a symbol in the `livecheck` block. This is a known issue and I think we'll have to deal with it in the future by making `formula_or_cask` available in a `livecheck` block, so we can optionally resolve a URL symbol like `url_string = url.is_a?(Symbol) ? Homebrew::Livecheck::livecheck_url_to_string(url, formula_or_cask) : url`.

miccal:

PR #109121, opened 2021-07-25
Update aqua-data-studio from 21.0.3 to 22.0.0

FnControlOption:

### DIFF
--- a/Casks/aqua-data-studio.rb
+++ b/Casks/aqua-data-studio.rb
@@ -12,7 +12,7 @@ cask "aqua-data-studio" do
     regex(%r{href=["']?([^"' >]*?/changelog/page/Version[._-]?(\d+(?:\.\d+)+)/[^"' >]*?)["' >]}i)
     strategy :page_match do |page, regex|
       changelog_matches = page.scan(regex)
-      next if changelog_matches.blank?
+      next [] if changelog_matches.blank?
 
       changelog_matches.uniq!(&:second)
       changelog_matches.sort_by! { |match| Version.new(match.second) }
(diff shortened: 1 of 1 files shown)


bevanjkay:
Ping @samford for review

samford (changes requested):
I've added a suggestion that follows the same basic idea but makes fewer assumptions. The suggested `livecheck` block is more in line with how we've approached checks in homebrew/core that require fetching more than one page.

A few things to note:

* I removed the `path` variable, as I would prefer to avoid establishing a precedent of defining variables outside of the provided `livecheck` block DSL methods. We can avoid duplicating the URL by using `URI#join` to combine the base `url` with the path to the changelog page for a given version. This also stands a better chance of continuing to work if the URLs on the changelog page are modified in the future.
* I've tweaked the regex to capture the entire `href` attribute value (to feed to `URI#join`) and I've made the preceding path a little looser (i.e., we usually include the minimum that's necessary to accurately match what we want).
* This gracefully fails when we're unable to match major/minor versions from URLs on the changelog page.
* This sorts versions from the changelog page, parsing the major/minor version strings as `Version`s and comparing them. This allows us to identify the newest major/minor version without assuming that the first match on the page will always be the newest version.
* This doesn't assume that the version from the changelog page URLs will always be only a major/minor version. I modified the regex to be able to match longer versions (whereas the existing regex would fail if this was changed) and this only appends `.0` to the `changelog_version` if it has fewer than three numeric parts separated by dots.

samford (changes requested):

miccal:

PR #109122, opened 2021-07-25
Update avidcodecsle from 2.3.7 to 2.7.6

FnControlOption:


bevanjkay:
@FnControlOption Looks like it may be a similar issue to #109120

bevanjkay:
@FnControlOption I couldn't find a location to livecheck this one either. 

We may need to just include a livecheck skip block - https://docs.brew.sh/Brew-Livecheck#skip

PR #109132, opened 2021-07-26
robo-3t: update livecheck

bevanjkay:
Updates livecheck to only check for stable releases. Cask was previously changed to allow beta-releases, however the in-app update is now only returning the latest stable release.

---

### DIFF
--- a/Casks/robo-3t.rb
+++ b/Casks/robo-3t.rb
@@ -9,8 +9,8 @@ cask "robo-3t" do
   homepage "https://robomongo.org/"
 
   livecheck do
-    url "https://github.com/Studio3T/robomongo/releases/latest"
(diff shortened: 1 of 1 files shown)


Casks/robo-3t.rb
```diff
@@ -8,9 +8,8 @@
   desc "MongoDB management tool"
   homepage "https://robomongo.org/"
 
-  # We need to check all releases since the current latest release is a beta version.
   livecheck do
-    url "https://github.com/Studio3T/robomongo/releases"
+    url "https://github.com/Studio3T/robomongo/releases/latest"
```
SMillerDev, line 12: wouldn't this make it `:github_latest`?

PR #109148, opened 2021-07-26
Add etlegacy v2.77.1

sebdanielsson:

Second cask so please correct any errors you may encounter.
- wget dependency is because the version of curl that macOS is shipping with doesn't allow the --output-dir option.
- Running wget 3 times because the command `wget https://mirror.etlegacy.com/etmain/pak{0..2}.pk3 -P etlegacy/etmain` don't work in Ruby that I know of. (Not a programmer tho)

### DIFF
--- a/Casks/etlegacy.rb
+++ b/Casks/etlegacy.rb
@@ -1,0 +1,23 @@
+cask "etlegacy" do
+  version "2.77.1"
+  sha256 :no_check
+
+  url "https://www.etlegacy.com/download/file/258"
+  name "ET: Legacy"
+  desc "Open source project based on the code of Wolfenstein: Enemy Territory"
+  homepage "https://www.etlegacy.com/"
+
+  depends_on macos: ">= :big_sur"
+  depends_on formula: "wget"
+
+  suite "ETLegacy"
+  installer script: {
+    executable: "wget",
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Neat trick! But we won’t accept it as is. No other cask uses this method and adding something different for what will be a niche cask isn’t a good idea.

If you can make it work with our regular methods, we’ll reopen and reconsider. Otherwise, you may wish to host on your own tap.

Tips: don’t download anything manually (i.e. outside of `url`); don’t install dependencies which are only useful for the installation process.

Thank you anyway.

SebDanielsson:
I'll try to figure something out :)

Any tips? Can one use multiple "url"?

Casks/etlegacy.rb
```diff
@@ -0,0 +1,28 @@
+cask "etlegacy" do
+  version "2.77.1"
+  sha256 :no_check
+
+  url "https://www.etlegacy.com/download/file/258"
+  name "ET: Legacy"
+  desc "Open source project based on the code of Wolfenstein: Enemy Territory"
```
vitorgalvao, line 7: `Game based on the code of Wolfenstein: Enemy Territory`

vitorgalvao:
> Any tips?

Not without investigating the software further.

> Would it be okay if I dropped the wget dep and tried to use a curl command to get the files?

No. That falls into my tip above.

> would it be okay if the etlegacy projekt offered a .sh script that I could call on to download the files?

Yes, as long as you get it with `url` and it does everything. That’s [`installer script:`](https://docs.brew.sh/Cask-Cookbook#installer-script).

PR #109193, opened 2021-07-27
calcservice: add livecheck

FnControlOption:

### DIFF
--- a/Casks/calcservice.rb
+++ b/Casks/calcservice.rb
@@ -9,7 +9,7 @@ cask "calcservice" do
 
   livecheck do
     url "https://www.devontechnologies.com/support/download"
-    regex(%r{<td>CalcService</td><td>(\d+(?:\.\d+)*)</td>}i)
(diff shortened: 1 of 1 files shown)


Casks/calcservice.rb
```diff
@@ -7,5 +7,12 @@
   name "CalcService"
   homepage "https://www.devontechnologies.com/apps/freeware"
 
+  livecheck do
+    url "https://www.devontechnologies.com/support/download"
+    regex(%r{<td>CalcService</td><td>(\d+(?:\.\d+)*)</td>}i)
```
bevanjkay, line 12: It may be better to use the URL used for the download for regex.

```suggestion
    regex(/href=.*?v?(\d+(?:\.\d+)+)\/CalcService\.zip/i)
```

miccal: It appears that the download `url` is not provided directly from `https://www.devontechnologies.com/support/download`.

PR #109195, opened 2021-07-27
caramba-switcher: remove

FnControlOption:
https://cdn.caramba-switcher.com/files/CarambaSwitcherBeta.pkg returns a 404; Caramba Switcher can only be installed from the App Store now.


bevanjkay (approved):
Thank you @FnControlOption!

PR #109223, opened 2021-07-28
Update monitorcontrol URL

jerr0328:
Repository is now under the "MonitorControl" org, so URL and homepage should be updated (old ones work, they just redirect to there). The previous behavior is confusing if you follow the instructions from MonitorControl's README but see a different URL used for the download.


bevanjkay (approved):
Thank you @jerr0328!

PR #109326, opened 2021-07-31
adobe-acrobat-pro: update livecheck

bevanjkay:


bevanjkay:
@miccal The previous URL for livecheck was using the release notes page, and was returning a `windows only` update at present. We could update the regex to filter them out, or the URL in in this PR is the endpoint checked by the in-app updater.

PR #109118, opened 2021-07-25
Update cloudflare-warp from 1.5.207.0,20210616.5 to 1.5.463.0,20210722.12

OliJWalker:
Created with `brew bump-cask-pr`.

PR #109339, opened 2021-07-31
Use Sparkle strategy in livecheck of cloudflare warp

n-thumann:

This PR changes the livecheck strategy of `cloudflare-warp` from `extract_plist` to `sparkle`, because that's what Cloudflare WARP uses behind the scenes. It also updates the version to the most recent one.

Not using `sparkle` has the disadvantage that the latest version is not detected properly: According to `extract_plist` `1.5.207.0,20210616.5` is the most recent one (because the binary says so), but immediately after install, a update window pops up, because `1.5.463.0,20210722.12` is actually the most recent one. Also, the check is much faster with `sparkle`.
FYI: I've extracted the livecheck url from the WARP binary itself, so it's the exact same source as the integrated updater uses :)

### DIFF
--- a/Casks/cloudflare-warp.rb
+++ b/Casks/cloudflare-warp.rb
@@ -1,5 +1,5 @@
 cask "cloudflare-warp" do
-  version "1.5.207.0,20210616.5"
+  version "1.5.463.0,20210722.12"
   sha256 :no_check
 
   url "https://cloudflarewarp.com/Cloudflare_WARP.zip"
--- a/Casks/cloudflare-warp.rb
+++ b/Casks/cloudflare-warp.rb
@@ -8,8 +8,8 @@ cask "cloudflare-warp" do
   homepage "https://cloudflarewarp.com/"
 
   livecheck do
-    url :url
-    strategy :extract_plist
+    url "https://api.appcenter.ms/v0.1/public/sparkle/apps/87d9992a-351c-44a9-849b-3f9a89d63d18"
+    strategy :sparkle
   end
 
   auto_updates true


miccal:
Thank you @n-thumann.

bevanjkay:
@n-thumann @miccal 

This PR may cause unintended results due to the reasons stated here - https://github.com/Homebrew/homebrew-cask/pull/109118#issuecomment-887184248

The download URL was previously serving a different version to the app centre update, and we are unable to use app centre for download URLs at this time (see link above). 

Depending on what happens with future updates we may need to revert to `extract_plist` and annotate the livecheck to explain why.

miccal:
Oh crap, sorry @bevanjkay -- please feel free to revert!

bevanjkay:
> Not using `sparkle` has the disadvantage that the latest version is not detected properly: According to `extract_plist` `1.5.207.0,20210616.5` is the most recent one (because the binary says so), but immediately after install, a update window pops up, because `1.5.463.0,20210722.12` is actually the most recent one. Also, the check is much faster with `sparkle`.
> FYI: I've extracted the livecheck url from the WARP binary itself, so it's the exact same source as the integrated updater uses :)

All of these are valid points, and thank you for looking into this. Livecheck should line up with the version that is served by `brew install` or `brew upgrade` commands, so at this point we need to stick with the strategy that returns the matching version from the download. It is worth noting the this cask has `auto_updates true`, and will therefore only be updated by homebrew when the `--greedy` flag is passed. This issue is problematic, and has caused manual version bumps to be closed in the past, however using `extract_plist` is the best course of action unless a versioned URL is made available upstream that doesn't utilise appcentre.

PR #109343, opened 2021-07-31
TextMate: link `mate_wait` binary

gaelicWizard:
If the `mate` CLI tool is invoked with name `mate_wait`, then it automatically acts as though `-w` was passed on the command line. This is helpful when something is invoking it via `$EDITOR` but doesn't interpret properly when `EDITOR="mate -w"`. Reference: https://macromates.com/manual/en/using_textmate_from_terminal#the_general_editor_variable

### DIFF
--- a/Casks/textmate.rb
+++ b/Casks/textmate.rb
@@ -18,6 +18,7 @@ cask "textmate" do
 
   app "TextMate.app"
   binary "#{appdir}/TextMate.app/Contents/Resources/mate"
+  binary "#{appdir}/TextMate.app/Contents/Resources/mate", target: "mate_wait"
 
   uninstall quit: "com.macromates.TextMate"
 


gaelicWizard:
I ran the `brew audit` and `brew style`. I didn't know how to do a version-number-bump. Formulas have `revision`, but Casks don't seem to have that. Thanks! 😃

vitorgalvao:
Welcome.

They do mention in the linked documentation that this trick is specifically for crontab, which—as they also mention—is obsoleted by launchd. Adding an extra binary which may confuse tab completion for such a narrow use case doesn’t seem worth it to have in the cask.

Thank you anyway.

gaelicWizard:
I run in to this *way* more than crontab (which I haven't used in a decade, and launchd has the same 'limitation'), but it's definitely somewhat niche (I imagine most people don't use $EDITOR), and I hadn't thought of it interfering with completion...which I just verified it definitely does. So, I'm ok with that. Thanks for looking at this! 😃

PR #109458, opened 2021-08-04
Update mattermost to support apple silicon build.

jbub:
Since v4.7.0 mattermost desktop is also shipping with native apple silicon (m1) build. Those builds are marked as beta though, i am not sure if this will get accepted because of that. But anyway i have been running those builds since RC releases with no issues.


miccal:
Thank you @jbub.

PR #109692, opened 2021-08-11
Update parallels from 16.5.1-49187 to 17.0.0-51461

g4brielvs:
Bump version to Parallels 17.


miccal:
This is now a universal binary, so no need for separate architectures.

PR #109765, opened 2021-08-12
Delete macclean.rb

vitorgalvao:
[This is not something we want to support!](https://www.virustotal.com/gui/file/ba88055972a8c01bf31aa9748b34ebc8ea54b90775ae29103c2455dbb320fcc6/detection)

We should consider removing the other casks from this vendor (search for `imobie`). The other’s weren’t flagged by VirusTotal, but they’re not very popular and can switch at any time.

These types of apps are prime choices for malicious vendors.


miccal:
All gone :)

vitorgalvao:
Thank you!

miccal:
> Thank you!

You are most welcome!

PR #110035, opened 2021-08-17
dupeguru 4.1.1

miccal:


mariopaumann:
@miccal why did you use the Qt Version of dupeguru ? i think the cocoa variant is better. The Qt Variant shows 0.0.0 as version number (which always shows up in MacUpdater), it is much bigger (150 vs 33 MB) and it looks worse.

miccal:
Because of what I read here:

https://dupeguru.voltaicideas.net/2021/08/14/macOS_releases_going_forward.html

PR #110109, opened 2021-08-17
Update Browserosaurus homepage

will-stone:


bevanjkay (changes requested):
Thanks @will-stone - I believe the verified parameters should be able to be removed from the download URLs also

will-stone:
Thanks 😊

PR #110134, opened 2021-08-18
Switch to supported download URL for Backblaze

mudge:
With the release of Backblaze 8.0.1.534, the versioned download URL for Backblaze (via https://secure.backblaze.com/api/install_backblaze) no longer works, instead redirecting users to https://www.backblaze.com/download_error.html

Having contacted their support team, they confirmed the following:

> We do not support [using https://secure.backblaze.com/api/install_backblaze for] installing the
> Backblaze program.
>
> It may have worked previously, but we are not able to support this currently.
>
> We don't have a way to specify a version for download, unfortunately.
>
> That said, you're always welcome to use this Update page to get the newest version of the Backblaze installer: https://secure.backblaze.com/update.htm
>
> This page will contain the following download links, which you can use directly if you prefer:
>
> * https://secure.backblaze.com/win32/install_backblaze.exe
> * https://secure.backblaze.com/mac/install_backblaze.dmg

Therefore switch to using the unversioned download URL and suppress the checksum as per the example from https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#examples as the checksum will change on the same URL when a new version is released.

We keep the livecheck stanza as the official download page uses that same method to load the current application version number.

### DIFF
--- a/Casks/backblaze.rb
+++ b/Casks/backblaze.rb
@@ -8,8 +8,9 @@ cask "backblaze" do
   homepage "https://backblaze.com/"
 
   livecheck do
-    url :url
-    strategy :extract_plist
+    url "https://secure.backblaze.com/api/clientversion.xml"
+    strategy :page_match
+    regex(/mac_version=.*?(\d+(?:\.\d+)*)/i)
   end
 
   auto_updates true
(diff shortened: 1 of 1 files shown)


miccal:
Thank you @mudge.

PR #110218, opened 2021-08-20
ghdl :latest

miccal:


miccal:
Not sure what to do with this one ... ?

bevanjkay:
@miccal Considering the latest version does not have built files, perhaps we just give this `version :latest` and point to the nightly?

miccal:
I have added this to the [`github_prerelease_allowlist.json`](https://github.com/Homebrew/homebrew-cask/blob/master/audit_exceptions/github_prerelease_allowlist.json) and now is passes the CI.

bevanjkay:
@miccal If we use nightly it will have to be `version :latest` because it doesn't look like it's the same as `1.0.0`

PR #110255, opened 2021-08-20
manuskript: add livecheck

miccal:


bevanjkay:
@miccal we should omit `strategy :page_March` if there is no block, as it's implied in those instances.

miccal:
@bevanjkay if the `url` is a `github` page the strategy will default to the `github` strategy, so I think the `:page_match` is necessary here.

bevanjkay:
@miccal sorry missed that, carry on! 🙄

PR #110333, opened 2021-08-21
wolfram-player: remove cask

bevanjkay:
`wolfram-player` now bundles as an app the downloads the wolfram player app


miccal:
```
==> Analytics
install: 0 (30 days), 9 (90 days), 9 (365 days)
```

PR #110485, opened 2021-08-22
smcfancontrol: remove

yeSpud:
```
==> Analytics
install: 455 (30 days), 1,642 (90 days), 5,956 (365 days)
```

The website for this cask seems to no longer exist


yeSpud:
Website listed was http://www.eidac.de/ for those who are curious

bevanjkay:
Given that the cask is reasonably popular (in the top 250 casks), despite note being updated for some time.
We should be able to use the Github repo for the download.
https://github.com/hholtmann/smcFanControl
```
30 days: 455 (#248)
90 days: 1,642 (#220)
365 days: 5,956 (#241)
```

yeSpud:
Ah

So would it be better just to update the urls in the file (as the site it points to seems to have expired and thus can only be downloaded via the github page)?

yeSpud:
Going to close this and fix the URLs under a new PR

Thanks!

PR #110486, opened 2021-08-22
smcfancontrol: update url

yeSpud:


yeSpud:
The url (and site) for smcfancontrol doesn't appear to exist anymore, so fallback to their Github

bevanjkay:
Thanks @yeSpud

PR #110569, opened 2021-08-24
Add Infra.app 0.46.0

lingrino:

### DIFF
--- a/Casks/infra.rb
+++ b/Casks/infra.rb
@@ -1,0 +1,11 @@
+cask "infra" do
+  version "0.46.0"
+  sha256 "d33af6e480e0fc04f7ad97b18a662fced3f0e7bc56cfc91fc13c3e1d85ad10ff"
+
+  url "https://download.infra.app/darwin/Infra-darwin-universal-#{version}.zip"
+  name "infra"
+  desc "Kubernetes desktop client"
+  homepage "https://infra.app/"
+
+  app "Infra.app"
+end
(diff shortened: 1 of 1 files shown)


lingrino:
The failure is that livecheck cannot find the latest version. In that case should I prefer to download from `-latest`?

bevanjkay:
@lingrino Please try to add a [zap stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap). The linked documentation provides some recommended methods to create it.

I will have a look at the livecheck for you now

lingrino:
The latest commit has a zap stanza that I made with https://github.com/nrlquaker/homebrew-createzap

bevanjkay:
I'm not 100% satisfied with the livecheck solution here, @miccal do you have any better ideas? Would prefer not to use `extract_plist` here because the download is ~330mb

miccal:
> I'm not 100% satisfied with the livecheck solution here (because it is passing a fake version number into the updater API - works for now?), @miccal do you have any better ideas? Would prefer not to use `extract_plist` here because the download is ~330mb

I have not been able to find a suitable alternative besides `:extract_plist`, I am afraid :(

bevanjkay:
We could ask upstream to see if there's another option, otherwise, I think the solution I provided should work.. it is just more susceptible to breaking than other methods.

miccal:
> We could ask upstream to see if there's another option, otherwise, I think the solution I provided should work.. it is just more susceptible to breaking than other methods.

Let's just use your idea, and see how it goes :)

bevanjkay:
Thanks @lingrino 👍🏻

lingrino:
Thank you all very much!

PR #110743, opened 2021-08-29
Update edrawmax language and app name

v00id852:

The old cask reports "It seems the App source is not there." error because the app name in dmg file is not "EdrawMax.app", it should be "亿图图示.app" in Chinese and "Wondershare EdrawMax.app" in English.

Another bug is that the cask didn't correctly detect the region when the language is zh_CN, so I added "zh", it works.


miccal:
Thank you @zianglei.

PR #110863, opened 2021-08-31
Add Keyman 14.0.280 (new cask)

myzhang1029:
Add a new cask [Keyman](https://keyman.com).

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) ~~or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version)~~.
- [x] `brew audit --cask keyman` is error-free.
- [x] `brew style --fix keyman` reports no offenses.

- [x] `brew audit --new-cask keyman` worked successfully.
- [x] `brew install --cask keyman` worked successfully.
- [x] `brew uninstall --cask keyman` worked successfully.

### DIFF
--- a/Casks/keyman.rb
+++ b/Casks/keyman.rb
@@ -1,0 +1,41 @@
+cask "keyman" do
+  version "14.0.280"
+  sha256 "8465f512b23900f8b6527cf4a2f21c550d526bac7135607b3abfd1609608b91a"
+
+  url "https://downloads.keyman.com/mac/stable/#{version}/keyman-#{version}.dmg"
+  name "Keyman"
+  desc "Reconfigures keyboard to type in another language"
+  homepage "https://keyman.com/"
+
+  livecheck do
+    url "https://github.com/keymanapp/keyman.git"
+    regex(/^(?:mac-)?release-(?:stable-)?(\d+(?:\.\d+)+)$/i)
+    # Parse macOS downloads listing, in case the tagging format changes again
+    # url "https://downloads.keyman.com/mac/stable"
+    # strategy :page_match
+    # regex(%r{href="(\d+(?:\.\d+)*)/"}i)
+  end
+
+  depends_on macos: ">= :yosemite"
+
+  input_method "Install Keyman.app/Contents/MacOS/Keyman.app"
+
+  uninstall quit: "keyman.inputmethod.Keyman"
+
+  zap trash: [
+    "~/Documents/Keyman-Keyboards",
+    "~/Library/Caches/keyman.inputmethod.Keyman",
+    "~/Library/Preferences/keyman.inputmethod.Keyman.plist",
+  ]
+
+  caveats <<~EOS
+    You must manually install the keyboard layout under:
+
+       Preferences → Keyboard → Input Sources
+
+    Before uninstalling Keyman, it is also suggested to manually remove the input method.
+
+    See https://help.keyman.com/products/mac/current-version/start/install-keyman.
+    and https://help.keyman.com/products/mac/current-version/troubleshooting/uninstall-keyman.
+  EOS
+end
--- a/Casks/keyman.rb
+++ b/Casks/keyman.rb
@@ -8,12 +8,8 @@ cask "keyman" do
   homepage "https://keyman.com/"
 
   livecheck do
-    url "https://github.com/keymanapp/keyman.git"
-    regex(/^(?:mac-)?release-(?:stable-)?(\d+(?:\.\d+)+)$/i)
-    # Parse macOS downloads listing, in case the tagging format changes again
-    # url "https://downloads.keyman.com/mac/stable"
-    # strategy :page_match
-    # regex(%r{href="(\d+(?:\.\d+)*)/"}i)
+    url "https://downloads.keyman.com/mac/stable"
+    regex(%r{href=["']?v?(\d+(?:\.\d+)+)/?["' >]}i)
   end
 
   depends_on macos: ">= :yosemite"
--- a/Casks/keyman.rb
+++ b/Casks/keyman.rb
@@ -8,8 +8,8 @@ cask "keyman" do
   homepage "https://keyman.com/"
 
   livecheck do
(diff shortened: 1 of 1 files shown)


myzhang1029:
I'm marking this as a draft until the folks at @keymanapp agrees to add this cask (keymanapp/keyman#5663). Feel free to tell me that this is not required.

Casks/keyman.rb
```diff
@@ -0,0 +1,41 @@
...
+  version "14.0.280"
+  sha256 "8465f512b23900f8b6527cf4a2f21c550d526bac7135607b3abfd1609608b91a"
+
+  url "https://downloads.keyman.com/mac/stable/#{version}/keyman-#{version}.dmg"
+  name "Keyman"
+  desc "Reconfigures keyboard to type in another language"
+  homepage "https://keyman.com/"
+
+  livecheck do
+    url "https://github.com/keymanapp/keyman.git"
```
darcywong00, line 11: Here's an API to query for the latest versions of Keyman for macOS
https://downloads.keyman.com/api/version/mac

bevanjkay:
@myzhang1029 I have updated the livecheck block to use the same location as the URL.

Casks/keyman.rb
```diff
@@ -0,0 +1,37 @@
...
+  version "14.0.280"
+  sha256 "8465f512b23900f8b6527cf4a2f21c550d526bac7135607b3abfd1609608b91a"
+
+  url "https://downloads.keyman.com/mac/stable/#{version}/keyman-#{version}.dmg"
+  name "Keyman"
+  desc "Reconfigures keyboard to type in another language"
+  homepage "https://keyman.com/"
+
+  livecheck do
+    url "https://downloads.keyman.com/mac/stable"
```
darcywong00, line 11:
```suggestion
    url "https://downloads.keyman.com/api/version/mac"
    strategy: page_match
```
and update the regex for the response

```
{
    "mac": {
        "alpha": "15.0.107",
        "beta": "14.0.268",
        "stable": "14.0.280"
    }
}
```

myzhang1029: Yes, the regex for that could be
```ruby
regex(%r{"stable": "(\d+(?:\.\d+)+)",?}i)
```
But I'm not sure if this is a good idea to @bevanjkay.

bevanjkay: @darcywong00 Is there an instance where the stable archive with the downloads folder would be ahead of the current release? Otherwise, as per [livecheck URL guidelines](https://docs.brew.sh/Brew-Livecheck#url-guidelines) we should check the same location as the stable archive.

myzhang1029: According to the [API doc](https://github.com/keymanapp/keyman/wiki/downloads.keyman.com-version-API#api-use-notice), this endpoint may "change without notice". This could be a reason not to use it though.

darcywong00: > Otherwise, as per livecheck URL guidelines we should check the same location as the stable archive.

Give all the above, go ahead and use your original url. My apologies.

myzhang1029: Well, then I was too eager...

vitorgalvao (approved):

Casks/keyman.rb
```diff
@@ -0,0 +1,37 @@
...
+    You must manually install the keyboard layout under:
+
+       Preferences → Keyboard → Input Sources
+
+    Before uninstalling Keyman, it is also suggested to manually remove the input method.
+
+    See https://help.keyman.com/products/mac/current-version/start/install-keyman.
+    and https://help.keyman.com/products/mac/current-version/troubleshooting/uninstall-keyman.
+  EOS
+end
```
mcdurdin, line 37: The normal installer for Keyman runs a [script](https://github.com/keymanapp/keyman/blob/ad99c6c687a992ba5d621e462f4c853e2a58dacf/mac/setup/source/install.applescript) in `Install Keyman.app` to both install the input method and configure security (e.g. Privacy/quarantine attributes). As this cask does not appear to run the installer, it would be worthwhile adding a note about security configuration required to run Keyman (per those links).

[Noting that a included helper program textinputsource in the installer is currently Intel-only, which is an open issue for us to resolve.]

myzhang1029: @mcdurdin Thanks for reminding. So is the script included as

`/Install Keyman.app/Contents/Resources/Scripts/main.scpt`(AppleScript compiled) ?

myzhang1029: Or, even better, if there is a way to run `Install Keyman.app` without user interaction, I guess that will be a good option. :-)

mcdurdin: At present there is no way to run the script without user interaction -- it does some steps automatically but prompts the user for others around security. As far as I know, changing security options on macos is always a user task and cannot be automated?

myzhang1029: Well, I guess we can integrate `main.scpt` into the Cask as long as the script requires user confirmation. I'm tracking the progress in #110993. Cheers!

mcdurdin: Awesome 😁 

> Noting that a included helper program textinputsource in the installer is currently Intel-only, which is an open issue for us to resolve.

Tracked in keymanapp/keyman#5573.

This also closes keymanapp/keyman#3881! Thank you! 😁

myzhang1029: @mcdurdin
Homebrew's maintainers suggested that we aren't supposed to disable macOS's security feature by touching the `com.apple.quarantine` attribute in https://github.com/Homebrew/homebrew-cask/pull/110993#issuecomment-913232293.

Currently, I think the installation script only removes the quarantine attribute and adds the input method, which I guess the users can handle. Should there be more complicated actions, feel free to drop me a message and I will try to work them out.

mcdurdin: I think it's good as it is. We want to rewrite the installer for macOS completely to make it a lot simpler for end users, and avoid using a script for it. Once we have done that, we can revisit this cask. Thanks!

PR #110993, opened 2021-09-05
Run Keyman post-install operations

myzhang1029:
Keyman's maintainers have kindly proposed to run the installer actions.

- [x] `brew audit --cask keyman` is error-free.
- [x] `brew style --fix keyman` reports no offenses.

### DIFF
--- a/Casks/keyman.rb
+++ b/Casks/keyman.rb
@@ -16,6 +16,17 @@ cask "keyman" do
 
   input_method "Install Keyman.app/Contents/MacOS/Keyman.app"
 
+  # Post-install operations originally in Resources/Scripts/main.scpt
+  postflight do
+    system_command "/usr/bin/xattr",
+                   args: [
+                     "-d",
+                     "-r",
+                     "com.apple.quarantine",
+                     "#{ENV["HOME"]}/Library/Input Methods/Keyman.app",
+                   ]
+  end
+
   uninstall quit: "keyman.inputmethod.Keyman"
 
   zap trash: [


myzhang1029:
macOS always refuses to run the `textinputsource` script, not sure if it is related to my M1 arch.
I'll update the "caveats" stanza when I get around that.

vitorgalvao:
We don’t delete the quarantine, *we add it* and go to some lengths to do so. A download via `curl` (which Homebrew uses) does not add the quarantine.

To install something without quarantining, use the `--quarantine` flag.

mcdurdin:
This is all from somewhat fuzzy memory, sorry...  

`Install Keyman.app` also has this quarantine bit applied, so the user would receive two quarantine prompts in the install -- one for the installer app, and one for `Keyman.app`, which just seems excessive.

Another reason we decided to remove the quarantine bit from `Keyman.app` was that in some circumstances macOS was not presenting UI to 'allow' Keyman to run (something to do with it being an input method, I think). This meant that Keyman would simply appear not to work, and the user would have to figure out how to go to ~/Library/Input Methods and manually remove the quarantine bit from Keyman.app.

I'd need to investigate further and do some testing to verify that my recollection is 100% correct here...

(We'd prefer not to do remove quarantine bits, but this seemed better than a broken install from the end user point-of-view. The other security prompts are not bypassed by the installer -- they are presented to the user as is appropriate, even if the current implementation is a little rough.)

vitorgalvao:
Like I said, installing with `--no-quarantine` will not add the quarantining. You’re free to suggest that to your users. But removing the quarantining we add doesn’t make sense.

I hope you understand that I don’t know you nor your app. My concern is the users of HBC, and they have an expectation that quarantine will be added unless *they* decide not to.

mcdurdin:
Yes, understood. I'm happy if we document the manual steps that users need to take in order to remove the quarantine bit given that macOS doesn't always prompt to do this for input methods, in my experience.

PR #111086, opened 2021-09-07
Add `allow_untrusted: true` to wine-stable

mathiasvr:
There was no way for me to install wine without this fix because the package wasn't signed.

### DIFF
--- a/Casks/wine-stable.rb
+++ b/Casks/wine-stable.rb
@@ -19,7 +19,7 @@ cask "wine-stable" do
   ]
   depends_on cask: "xquartz"
 
-  pkg "winehq-stable-#{version}.pkg",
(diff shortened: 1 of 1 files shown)


miccal:
See https://docs.brew.sh/Cask-Cookbook#stanza-pkg and https://github.com/Homebrew/homebrew-cask/issues/110763

Sorry @mathiasvr but `allow_untrusted: true` is not permitted for `pkg`'s in official Homebrew Cask taps.

mathiasvr:
@miccal Thanks, is there anyway to override this in cli when installing? I solved it by downloading the cask and editing manually but would have been great if there was an easier way.

miccal:
Perhaps consider hosting this [in your own Tap](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap) until upstream fix the signing issue.

PR #102776, opened 2021-03-28
add reMarkable.app

bernadinm:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [X] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).
- [X] `brew audit --cask {{cask_file}}` is error-free.
- [X] `brew style --fix {{cask_file}}` reports no offenses.

Additionally, **if adding a new cask**:

- [X] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [X] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [X] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).
- [X] `brew audit --new-cask {{cask_file}}` worked successfully.
- [X] `brew install --cask {{cask_file}}` worked successfully.
- [X] `brew uninstall --cask {{cask_file}}` worked successfully.

PR #45534, opened 2018-04-04
Delete remarkable.rb

vitorgalvao:
[It’s a walled build again](https://github.com/caskroom/homebrew-cask/pull/45531).

Pinging @sandsmark due to https://github.com/caskroom/homebrew-cask/pull/39286.

PR #111097, opened 2021-09-08
Add Remarkable Desktop app 2.8.2

gianpaj:

**new cask**:

### DIFF
--- a/Casks/remarkable.rb
+++ b/Casks/remarkable.rb
@@ -15,6 +15,14 @@ cask "remarkable" do
       page.scan(regex).map { |match| "#{match[1]},#{match[0]}" }
     end
   end
+  
+  zap trash: [
+    "~/Library/Application Support/remarkable",
+    "~/Library/Caches/com.remarkable.desktop",
+    "~/Library/Caches/remarkable",
+    "~/Library/Preferences/com.remarkable.desktop.plist",
+    "~/Library/Saved Application State/com.remarkable.desktop.savedState",
+  ]
 
   auto_updates true
   depends_on macos: ">= :high_sierra"
--- a/Casks/remarkable.rb
+++ b/Casks/remarkable.rb
@@ -15,7 +15,12 @@ cask "remarkable" do
       page.scan(regex).map { |match| "#{match[1]},#{match[0]}" }
     end
   end
-  
+
+  auto_updates true
+  depends_on macos: ">= :high_sierra"
+
+  app "reMarkable.app"
+
   zap trash: [
     "~/Library/Application Support/remarkable",
     "~/Library/Caches/com.remarkable.desktop",
@@ -23,9 +28,4 @@ cask "remarkable" do
     "~/Library/Preferences/com.remarkable.desktop.plist",
     "~/Library/Saved Application State/com.remarkable.desktop.savedState",
   ]
-
-  auto_updates true
-  depends_on macos: ">= :high_sierra"
-
-  app "reMarkable.app"
 end
--- a/Casks/remarkable.rb
+++ b/Casks/remarkable.rb
@@ -2,7 +2,8 @@ cask "remarkable" do
   version "2.8.2.170,11141120"
   sha256 "7669f1e8dd1b13fd5305af8a215b17b33b7ff6cc218a218c9992b7dd83223465"
 
-  url "https://eu-central-1.linodeobjects.com/remarkable-#{version.major}/sparkle/reMarkableMacOs/Prod/#{version.after_comma}/reMarkable-#{version.before_comma}.dmg"
+  url "https://eu-central-1.linodeobjects.com/remarkable-#{version.major}/sparkle/reMarkableMacOs/Prod/#{version.after_comma}/reMarkable-#{version.before_comma}.dmg",
+      verified: "eu-central-1.linodeobjects.com/"
   name "Remarkable"
   desc "Desktop application"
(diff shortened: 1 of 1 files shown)


Casks/remarkable.rb
```diff
@@ -0,0 +1,13 @@
+cask "remarkable" do
+  version "2.8.2.170"
+  sha256 "7669f1e8dd1b13fd5305af8a215b17b33b7ff6cc218a218c9992b7dd83223465"
+
+  url "https://downloads.remarkable.com/desktop/production/mac/reMarkable-#{version}.dmg"
```
miccal, line 5: How does one find this `url` from the `homepage`?

bevanjkay:
There is an appcast URL available at `https://get-updates.cloud.remarkable.engineering/sparkle/reMarkableMacOs/Prod/appcast.xml` - However the download URL provided by sparkle is different and it is hard to ascertain which parts would be dynamic until there is a new update available.

@gianpaj The cask should also include `auto_updates true`

It also seems that generally the app download is walled, however the download URL is active without any authorisation.

See: https://support.remarkable.com/hc/en-us/articles/360002665378-Desktop-app

miccal:
Also see previous PR's: https://github.com/Homebrew/homebrew-cask/pull/102776 https://github.com/Homebrew/homebrew-cask/pull/45534 https://github.com/Homebrew/homebrew-cask/pull/45531 https://github.com/Homebrew/homebrew-cask/pull/39286

vitorgalvao:
> It also seems that generally the app download is walled

Note that to be considered a walled download, it needs to be behind a login wall *and* be hosted on a different domain than the homepage. The second part doesn’t happen here, thus it’s not walled.

miccal:
Thank you for your contribution to `homebrew-cask` @gianpaj.

Could you please try to add a [`zap` stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap), thanks.

gianpaj:
@miccal how do I find which files should be zap'ed? Is there a way `brew` can automatically generate/find these files for me to add them to the cask definition?
And if so is there an example? I don't seem to understand the doc you provided. Feel free to update my PR if you're able to

miccal:
> @miccal how do I find which files should be zap'ed? Is there a way `brew` can automatically generate/find these files for me to add them to the cask definition?
> And if so is there an example? I don't seem to understand the doc you provided. Feel free to update my PR if you're able to

The documentation I linked explains very clearly what to do, and includes an example for the Dropbox Cask.

gianpaj:
I see now. That documentation text is hard to read with tired eyes... 

I've added the trash files now @miccal 

Thanks

bevanjkay:
Thank you @gianpaj 
In future, please run `brew style --fix <cask>` before committing.

gianpaj:
Got it. Thanks, guys!

BTW a little Github action checking the output of `brew style --fix <cask>` would be nice to have 🙃

adavidzh:
I was reading this thread to understand how to go about getting version `2.10` in. IIUC:
- Regarding [this comment](https://github.com/Homebrew/homebrew-cask/pull/111097#issuecomment-916302748) on the question of being walled, be as it may that the link is not readily available, it is a fact that [version `2.10`](https://downloads.remarkable.com/desktop/production/mac/reMarkable-2.10.0.172.dmg) can be downloaded by anyone (with `wget`) and is hosted under `remarkable.com`.
- Regarding [this comment](https://github.com/Homebrew/homebrew-cask/pull/111097#issuecomment-915690182) on `auto_updates true`: AFAICT, none of these application versions are self-updatable. At least `2.5` wasn't and `2.10` isn't either.
- I checked https://get-updates.cloud.remarkable.engineering/sparkle/reMarkableMacOs/Prod/appcast.xml and it proposes both `2.8.2.170` and `2.10.0.172 ` (how to choose?) so I was wondering if that is a better path somehow; I've never worked with `appcast` in formulae.

With these in mind, what do you all think would be the best way to transmogrify (and update) this cask?

bevanjkay:
> I checked https://get-updates.cloud.remarkable.engineering/sparkle/reMarkableMacOs/Prod/appcast.xml and it proposes both 2.8.2.170 and 2.10.0.172 (how to choose?) so I was wondering if that is a better path somehow; I've never worked with appcast in formulae.

@adavidzh If you look closely at the URL provided in the appcast for version `2.10.0.172` it is in the `dev` directory. This would indicate that it isn't the current stable release. However the app does seem to be returning this as the latest version with the in-app updater.

> Regarding this comment on auto_updates true: AFAICT, none of these application versions are self-updatable. At least 2.5 wasn't and 2.10 isn't either.

Version `2.8.2.170` definitely does allow checking for updates inside the app.

adavidzh:
Thanks @bevanjkay:

> @adavidzh If you look closely at the URL provided in the appcast for version `2.10.0.172` it is in the `dev` directory. This would indicate that it isn't the current stable release. However the app does seem to be returning this as the latest version with the in-app updater.

That is quite odd in the sense that I was just proposed to download that from `https://downloads.remarkable.com/desktop/production/mac/reMarkable-2.10.0.172.dmg` where there is no `dev` hint. I wonder if the company proposes different versions depending on what they know is installed in the device...

> Version `2.8.2.170` definitely does allow checking for updates inside the app.

Perhaps I am not looking in the right place; where do you see that in the app? In `2.10.0.172` I see nothing on the menus and the `About reMarkable` window also doesn't show something like that.

PR #111118, opened 2021-09-09
touchosc: add zap, update livecheck, bump version

bevanjkay:


Casks/touchosc.rb
```diff
@@ -1,16 +1,25 @@
 cask "touchosc" do
-  version "1.0.5.108"
-  sha256 "b1b76240bf302635f8dd4d02200edb166e536ef08a96acfc870af38401efc5fe"
+  version "1.0.5,109"
+  sha256 "ec7e563b5c2a1a267386f3f9941aea63953d0823e1f8c356e875059ad6893822"
 
-  url "https://hexler.net/pub/touchosc/touchosc-#{version}-macos.dmg"
+  url "https://hexler.net/pub/touchosc/touchosc-#{version.tr(",", ".")}-macos.dmg"
```
bevanjkay, line 5: @miccal Do we have a position on the best way to handle this?

1) Use .tr to change the `,` to a `.`
2) Use `#{version.before_comma}.#{version.after_comma}`
3) Use a block in the livecheck to output the version with a `.` instead of a `,`

miccal: Personally I would leave the `livecheck` as-is and just have `#{version.before_comma}.#{version.after_comma}`.

PR #111171, opened 2021-09-10
Upgrade kui from 10.4.6 to 10.4.13

starpit:
Note: This PR also **adds an arm64 download option**, and a zap stanza.

---

### DIFF
--- a/Casks/kui.rb
+++ b/Casks/kui.rb
@@ -24,5 +24,5 @@ cask "kui" do
 
   binary "#{appdir}/Kui.app/Contents/Resources/kubectl-kui"
 
-  zap trash: "~/Library/Application\ Support/Kui"
+  zap trash: "~/Library/Application Support/Kui"
 end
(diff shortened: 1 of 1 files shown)


Casks/kui.rb
```diff
@@ -1,18 +1,29 @@
...
   name "Kui"
   desc "Hybrid command-line/UI development experience for cloud-native development"
   homepage "https://kui.tools/"
 
   livecheck do
     url :url
-    strategy :github_latest
+    strategy :git
+    regex(/^v(\d+(?:\.\d+)*)$/)
   end
```
bevanjkay, line 24: @starpit I think the `:github_latest` strategy is probably still required as some tags do not include binaries. Also, for reference, when the strategy is `:git` and a custom regex is not required (and the download URL is from Github), the livecheck block should be removed as it is used by default.

bevanjkay (changes requested):

starpit:
Awesome thanks for the great feedback!

Casks/kui.rb
starpit: Awesome thanks for the great feedback! I have pushed some updates.

bevanjkay (approved):

bevanjkay:
Thanks @starpit

PR #111172, opened 2021-09-10
Add nimblenote.app v3.1.1

davidwinter:

### DIFF
--- a/Casks/nimblenote.rb
+++ b/Casks/nimblenote.rb
@@ -1,0 +1,19 @@
+cask "nimblenote" do
+  version "3.1.1"
+  sha256 "7dbdaa2d79baa13493e28c2ab0ba38671a6e2874dfa07c09278c8ebc278f775a"
+
+  url "https://github.com/nimblenote/nimblenote/releases/latest/download/nimblenote-#{version}.dmg", verified: "https://github.com/nimblenote/nimblenote"
+  name "nimblenote"
+  desc "Keyboard-driven note taking"
+  homepage "https://nimblenote.app/"
+
+  livecheck do
+    url "https://github.com/nimblenote/nimblenote/releases/latest/download/release.json"
+    strategy :page_match
+    regex(/"version":"(\d+\.\d+\.\d+)"/i)
+  end
+
+  auto_updates true
+
+  app "nimblenote.app"
+end
--- a/Casks/nimblenote.rb
+++ b/Casks/nimblenote.rb
@@ -13,6 +13,7 @@ cask "nimblenote" do
   app "nimblenote.app"
 
   zap trash: [
+    "~/Library/Application Support/nimblenote/blob_storage",
     "~/Library/Application Support/nimblenote/Cache",
     "~/Library/Application Support/nimblenote/Code Cache",
     "~/Library/Application Support/nimblenote/Cookies",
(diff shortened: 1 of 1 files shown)


davidwinter:
Appears as though this has failed as it is assuming that the GitHub repo is where the source is located, whereas it is only used for downloads and website source. Is there a way to workaround this?

Casks/nimblenote.rb
```diff
@@ -0,0 +1,19 @@
...
+  url "https://github.com/nimblenote/nimblenote/releases/latest/download/nimblenote-#{version}.dmg", verified: "https://github.com/nimblenote/nimblenote"
+  name "nimblenote"
+  desc "Keyboard-driven note taking"
+  homepage "https://nimblenote.app/"
+
+  livecheck do
+    url "https://github.com/nimblenote/nimblenote/releases/latest/download/release.json"
+    strategy :page_match
+    regex(/"version":"(\d+\.\d+\.\d+)"/i)
+  end
```
bevanjkay, line 14: @davidwinter The livecheck stanza is not required when the download URL is github, as it uses the `:git` strategy by default.

bevanjkay (changes requested):

bevanjkay:
> From what I can tell, if the app wasn't using GitHub for release downloads, then this check wouldn't even be happening and the submission would have passed the checks?

Generally, the notability threshold can be disregarded when it is only the binaries that are hosted on Github.

Additionally, can you please try to add a [zap stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap). The linked documentation provides some recommended methods to create it.

Casks/nimblenote.rb
davidwinter: Does that integrate with GitHub releases, as the main application source code does not actually live within that git repository?

davidwinter:
@bevanjkay thanks for the clarification - I've also added the `zap` stanza. I've purposefully not removed the entire `Application Support/nimblenote` directory, as some user-created files exist underneath it that we wouldn't want to delete.

bevanjkay:
> @bevanjkay thanks for the clarification - I've also added the `zap` stanza. I've purposefully not removed the entire `Application Support/nimblenote` directory, as some user-created files exist underneath it that we wouldn't want to delete.

What kind of user created files? The zap stanza only runs if it is invoked with the --zap switch on `brew uninstall`

davidwinter:
> > @bevanjkay thanks for the clarification - I've also added the `zap` stanza. I've purposefully not removed the entire `Application Support/nimblenote` directory, as some user-created files exist underneath it that we wouldn't want to delete.
> 
> What kind of user created files? The zap stanza only runs if it is invoked with the `--zap` switch on `brew uninstall`

User notes created by the application are stored within that directory by default, so not something to trash.

davidwinter:
@bevanjkay I _think_ I've done all the extra review points and checks I can now with this. It's still failing on the GitHub checks, but I'm assuming that is something that can be handled when it comes to merging?

miccal:
Thank you @davidwinter.

davidwinter:
Thanks @miccal - I had a few questions I just wanted to check up on which I'm then happy to open another PR if we decide appropriate:

- the `rmdir` that you added to the `zap` stanza; will that only remove the directory if it is empty? Otherwise, I didn't add this because any notes that a user creates with the application are stored within this directory, and I fear that adding this directory to be removed, rather than specific files and directories underneath it, will remove the user-created notes as well
- the name of the app is `nimblenote` rather than `Nimblenote`

PR #111250, opened 2021-09-13
Update multimc from 0.6.12 to 0.6.13

haykam821:
Created with `brew bump-cask-pr`.

PR #111253, opened 2021-09-13
multimc: update url, add depends_on

bevanjkay:
Continuation from #111250 111250

### DIFF
--- a/Casks/multimc.rb
+++ b/Casks/multimc.rb
@@ -2,7 +2,7 @@ cask "multimc" do
   version "0.6.13"
   sha256 :no_check
 
-  url "https://files.multimc.org/downloads/mmc-stable-osx64.tar.gz"
+  url "https://files.multimc.org/downloads/mmc-stable-osx64-5.5.12.tar.gz"
   name "Multi MC"
   desc "Minecraft launcher"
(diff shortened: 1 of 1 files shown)


bevanjkay:
Ping @haykam821

haykam821 (approved):

haykam821:
Thank you! This should fix issues with dark mode and using Microsoft accounts.

Casks/multimc.rb
```diff
@@ -2,7 +2,7 @@
   version "0.6.13"
   sha256 :no_check
 
-  url "https://files.multimc.org/downloads/mmc-stable-osx64.tar.gz"
+  url "https://files.multimc.org/downloads/mmc-stable-osx64-5.5.12.tar.gz"
```
haykam821, line 5:
```suggestion
  url "https://files.multimc.org/downloads/mmc-stable-osx64-5.15.2.tar.gz"
```

haykam821 (changes requested):

bevanjkay:
@haykam821 I will wait for other maintainer feedback before merging as this version is not the version provided by the `downloads` page on the upstream website. We will need to make the livecheck match with the version that is being downloaded also.

PR #111567, opened 2021-09-21
Add FirmaEC.app v2.10.0

sbibauw:
FirmaEC.app is the official digital signature (_firma electrónica_) recognized by Ecuadorian authorities.

The URL is sadly unversioned, but I added a `livecheck` for checking new versions on the official website.

### DIFF
--- a/Casks/firmaec.rb
+++ b/Casks/firmaec.rb
@@ -5,7 +5,7 @@ cask "firmaec" do
   url "https://www.firmadigital.gob.ec/firmaec/FirmaEC.zip"
   name "FirmaEC"
   desc "Firma electrónica en Ecuador"
-  homepage "https://www.firmadigital.gob.ec"
+  homepage "https://www.firmadigital.gob.ec/"
 
   livecheck do
     url "https://www.firmadigital.gob.ec/registro-de-cambios-de-firmaecchangelog/"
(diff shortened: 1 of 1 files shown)


miccal:
@sbibauw there are two `.app`'s inside the `.zip` file; one called `FirmaEC.app` and another called `FirmaECTransversal.app` -- can you please explain the difference between the two?

sbibauw:
> @sbibauw there are two `.app`'s inside the `.zip` file; one called `FirmaEC.app` and another called `FirmaECTransversal.app` -- can you please explain the difference between the two?

Good question and good call (+ thanks for the corrections in the code).
It's very poorly documented in the docs (PDF Manual available on the website). I've only used the main one (`FirmaEC.app`). It seems like `FirmaECTransversal.app` is used for some kind of integration with web services/web applications (typically from the Ecuadorian government) that might require it via some custom URLs like `firmaec:///` ([PDF Manual](https://www.firmadigital.gob.ec/wp-content/uploads/2021/08/Manual-Usuario-FirmaEC-v2.8.0.pdf), pp. 26-27).

I guess it should be included too...

sbibauw:
Not sure what to do here. The test seems to have failed due to a `curl timeout` error, which I assume shouldn't happen again... Is it possible to restart the tests? Or is there something else to do?

miccal:
Thank you @sbibauw.

PR #111732, opened 2021-09-25
Update bitwarden to use the universal binary on Apple silicon

danielkean:

### DIFF
--- a/Casks/bitwarden.rb
+++ b/Casks/bitwarden.rb
@@ -15,7 +15,7 @@ cask "bitwarden" do
 
   auto_updates true
 
-  app "Bitwarden.app"
+  pkg "Bitwarden-#{version}-universal.pkg"
 
   uninstall quit: [
     "com.bitwarden.desktop",
--- a/Casks/bitwarden.rb
+++ b/Casks/bitwarden.rb
@@ -8,6 +8,11 @@ cask "bitwarden" do
   desc "Desktop password and login vault"
   homepage "https://bitwarden.com/"
 
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
+
   auto_updates true
 
   pkg "Bitwarden-#{version}-universal.pkg"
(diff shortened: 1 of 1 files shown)


SMillerDev:
If it's universal, why not use it on Intel?

danielkean:
Yeah I wasn't sure whether it would be better to keep the smaller .zip file on the intel architecture whilst it exists, or to switch both to use the universal. Do you think it would be better to change it for both?

carlocab (dismissed):

SMillerDev:
I think if it's a universal app it'll be easier to rely on that.

danielkean:
@SMillerDev Hmm strange, when changed to use the universal binary for both architectures, it fails the tests. Any ideas?

neutric:
@danielkean "Error: It seems the App source '/usr/local/Caskroom/bitwarden/1.28.3/Bitwarden.app' is not there." You have to replace the app stanza in L18 with a [pkg stanza](https://docs.brew.sh/Cask-Cookbook#stanza-pkg) as you are no longer just unpacking a dmg.

danielkean:
@neutric Thanks for the help, I changed it to a pkg stanza as well as the uninstall to pkgutil, so the tests are now passing. It's my first time contributing to homebrew-cask so I would have been surprised if it went smoothly the first time hahaha.

miccal:
Thank you @danielkean.

danielkean:
@miccal Hi there, once installed, there seems to be an issue when running the application as can be seen in the first screenshot. Do you think this is something "pkg allow_untrusted:" would fix? I'm not sure why the tests didn't catch this problem.

I also wanted to make sure that the install location seems correct to you (It does from what I can see as it appears in the Applications folder), as I've never seen it say "Installing at base path /" , which can be seen in the second screenshot. Unless that's something specific to pkg files?

image


image

danielkean:
@miccal I've just submitted the issue to the Bitwarden/Desktop repo which can be found here:
https://github.com/bitwarden/desktop/issues/1090

Issue #39741, opened 2017-10-17
Rename casks with `Desktop` in name


vitorgalvao:
Refs https://github.com/caskroom/homebrew-cask/pull/39708.

commitay:
List of `*desktop*` Casks, not sure if all of them need to be renamed.

`caskroom`
- [ ] [acquia-dev-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/acquia-dev-desktop.rb)
- [ ] [authy-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/authy-desktop.rb)
- [ ] [bimedesktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/bimedesktop.rb)
- [ ] [chrome-remote-desktop-host](https://github.com/caskroom/homebrew-cask/blob/master/Casks/chrome-remote-desktop-host.rb)
- [ ] [colwiz-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/colwiz-desktop.rb)
- [ ] [desktop-log](https://github.com/caskroom/homebrew-cask/blob/master/Casks/desktop-log.rb)
- [ ] [desktoputility](https://github.com/caskroom/homebrew-cask/blob/master/Casks/desktoputility.rb)
- [ ] [freeswitch-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/freeswitch-desktop.rb)
- [ ] [github-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/github-desktop.rb)
- [ ] [google-play-music-desktop-player](https://github.com/caskroom/homebrew-cask/blob/master/Casks/google-play-music-desktop-player.rb)
- [ ] [isyncr-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/isyncr-desktop.rb)
- [ ] [jump-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/jump-desktop.rb)
- [ ] [lytro-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/lytro-desktop.rb)
- [ ] [macupdate-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/macupdate-desktop.rb)
- [ ] [mediafire-desktop](https://github.com/caskroom/homebrew-cask/blob/master/Casks/mediafire-desktop.rb)
[...]

PR #111861, opened 2021-09-29
Rename rancher to rancher-desktop

bashofmann:
The application is called "Rancher Desktop" not "Rancher", see https://rancherdesktop.io/.

"Rancher" is a different product, multi-cluster Kubernetes Management Solution and not a desktop app (https://rancher.com/products/rancher/).

The cask was just added a few days ago. Are any other steps necessary apart from renaming the file and the token?

### DIFF
--- a/Casks/rancher-desktop.rb
+++ b/Casks/rancher-desktop.rb
@@ -1,4 +1,4 @@
-cask "rancher" do
+cask "rancher-desktop" do
   version "0.5.0"
   sha256 "0ebdfb3e070e1c525a08730ccea761a040d366fa800c735870ff328080f8fa55"
 
--- a/audit_exceptions/github_prerelease_allowlist.json
+++ b/audit_exceptions/github_prerelease_allowlist.json
@@ -26,7 +26,7 @@
   "ql-ansilove": "all",
   "profilecreator": "all",
   "qdesktop": "all",
-  "rancher": "all",
+  "rancher-desktop": "all",
   "seaglass": "all",
   "servpane": "all",
   "shadowsocksx-ng-r": "all",


commitay:
> Named the cask according to the token reference.

> https://docs.brew.sh/Cask-Cookbook#token-reference
> Remove from the end: strings such as “Desktop”, “for Desktop”.

bashofmann:
I think that naming scheme does not really fit in this case as "rancher" and "rancher-desktop" both exist but are different products. Naming the cask "rancher" when it installs "rancher-desktop" will lead to all kinds of confusion.

bevanjkay:
Ping @vitorgalvao

jandubois:
> I think that naming scheme does not really fit in this case as "rancher" and "rancher-desktop" both exist but are different products. Naming the cask "rancher" when it installs "rancher-desktop" will lead to all kinds of confusion.

I think part of the confusion is that "Rancher" is both a company name (now part of SUSE), and the name of a product/project.

The separate project "Rancher Desktop" is not the "Rancher" project running on the desktop, but a tool for container runtimes on the desktop, made by Rancher, the company.

"Rancher", the project is not related to "Rancher Desktop", except by being produced by the same company.

Calling "Rancher Desktop" just "Rancher" would be like calling "Microsoft Office" just "Microsoft".

commitay:
> https://docs.brew.sh/Cask-Cookbook#token-reference
> By enforcing strict naming conventions we aim to:
>
> * Prevent duplicate submissions
> * Minimize renaming events
> * Unambiguously boil down the name of the software into a unique identifier
>
> Details of software names and brands will inevitably be lost in the conversion to a minimal token.

Also see https://github.com/Homebrew/homebrew-cask/issues/39741 were existing casks e.g. `github-desktop`, `parallels-desktop`, etc had the `-desktop` suffix removed.

bashofmann:
I understand the reasoning, but I still think this does not apply there. "rancher" as a product exists and is completely different to "rancher-desktop". You could even create a formula called "rancher" that installs the Rancher Management Server locally. Then you would have a formula and a cask that are both called "rancher" and install completely different products. And the only thing in common would be the vendor that created and maintains them.

vitorgalvao:
> Then you would have a formula and a cask that are both called "rancher"

Exactly like `docker` and a few others which cause no issue.

> I understand the reasoning, but I still think this does not apply there.

And we understand *your* reasoning and are explaining *it* doesn’t apply. @commitay is a former maintainer.

Until we have a way to rename casks properly (PR welcome), doing so is cumbersome and bad for everyone.

This in particular wouldn’t have passed if the cask were being submitted now; even worse if it’s renaming with no benefit or mitigation to the renaming.

AkihiroSuda:
> Exactly like docker and a few others which cause no issue.

The situation seems different.
Docker Desktop contains "Docker", but Rancher Desktop does not seem to contain "Rancher" (https://github.com/rancher/rancher) as of Rancher Desktop v0.5.0.

Also, I think Rancher/SUSE's opinion on the package name should be respected, as "Rancher" is a registered trademark in several countries. (IANAL, and I do not work for Rancher/SUSE.)

vitorgalvao:
> The situation seems different.

It’s not.

> Docker Desktop contains "Docker", but Rancher Desktop does not seem to contain "Rancher" (https://github.com/rancher/rancher) as of Rancher Desktop v0.5.0.

I have no idea of what you’re referring to.  That’s not the repo the casks uses and what matters is the name of the app.

> Also, I think Rancher/SUSE's opinion on the package name should be respected, as "Rancher" is a registered trademark in several countries. (IANAL, and I do not work for Rancher/SUSE.)

From my non-lawyer knowledge, that’s not how trademarks work. If they dislike the casks’s name, they can rename the app or ask us to not carry it. But we’ll not change casks to arbitrary names picked by vendors—the rules exist for Homebrew’s community and so we can keep consistent naming without duplicates.

I’m certainly open to changing our rules, but this does not warrant an exception. This case isn’t special.

PR #111971, opened 2021-10-02
Update cloudapp from 6.5.2,2245 to 6.6.0,2271

ghost:

### DIFF
--- a/Casks/cloudapp.rb
+++ b/Casks/cloudapp.rb
@@ -8,7 +8,7 @@ cask "cloudapp" do
   homepage "https://www.getcloudapp.com/"
 
   livecheck do
-    url "https://d2plwz9jdz9z5d.cloudfront.net/mac/latest/appcast.xml"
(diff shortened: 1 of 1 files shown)


zsoltion:
@miccal It seems like the appcast is behind (the website downloads an even older version). Can we still just merge it in?

bevanjkay:
> @miccal It seems like the appcast is behind (the website downloads an even older version). Can we still just merge it in?

The livecheck exists to ensure that we are providing the correct version from upstream. The appcast is the only place I can find a reference upstream to a specific version, so we will need to use that unless you can provide an alternative location to use for livecheck.

gdams:
According to https://downloads.getcloudapp.com/mac/latest/CloudApp.zip I get `6.5.3` (yet another version)?

zsoltion:
I'm not sure, that's what's running on my computer, but in that case, let's wait until it shows up again on the update server.

PR #112066, opened 2021-10-04
Update imagej to 1.53

takameron:

### DIFF
--- a/Casks/imagej.rb
+++ b/Casks/imagej.rb
@@ -17,5 +17,13 @@ cask "imagej" do
   desc "Image Processing and Analysis in Java"
   homepage "https://imagej.nih.gov/ij/index.html"
 
+  livecheck do
+    url "https://imagej.nih.gov/ij/download.html"
+    strategy :page_match do |page|
+      match = page.match(%r{href=.*?/ij(\d+)(\d+)(\d+)[._-]osx[._-]arm[._-]java\d+\.zip}i)
+      "#{match[1]}.#{match[2]}#{match[3]}"
+    end
+  end
+
   app "ImageJ.app"
 end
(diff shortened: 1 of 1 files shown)


Casks/imagej.rb
```diff
@@ -1,15 +1,29 @@
...
+
+  if Hardware::CPU.intel?
+    sha256 "4093259645569f5dc04ef5576a5f54ab878c0923f48c03b866f5d56f2edc078f"
+
+    url "https://wsr.imagej.net/distros/osx/ij#{version.no_dots}-osx-java8.zip",
+        verified: "wsr.imagej.net/distros/osx/"
+  else
+    sha256 "e464760174af0bcff7849db6e12090a4cbb1b0a1897a71dc1a999855d6e22e98"
+
+    url "https://wsr.imagej.net/distros/osx/ij#{version.no_dots}-osx-arm-java13.zip",
```
gdams, line 12: Is there a reason that Zulu 13 was chosen to be bundled here? Java 13 is unsupported upstream anymore?

takameron: [The download page](https://imagej.nih.gov/ij/download.html) only listed Java 13 bundles. However, Java 8 bundles are also [distributed](http://wsr.imagej.net/distros/osx/). Should I choose the one that bundles Java 8?

bevanjkay: @takameron We should distribute the version provided by upstream, which in this case look like the preference is the java13 variant as seen on [the download page](https://imagej.nih.gov/ij/download.html).

I would like to have a quick look at the livecheck block before merging.

bevanjkay:
Thanks @takameron

PR #112079, opened 2021-10-04
Update maciasl.rb

naveenkrdy:

### DIFF
--- a/Casks/maciasl.rb
+++ b/Casks/maciasl.rb
@@ -7,11 +7,6 @@ cask "maciasl" do
   desc "ACPI Machine Language (AML) compiler and IDE"
   homepage "https://github.com/acidanthera/MaciASL"
 
-  livecheck do
-    url :url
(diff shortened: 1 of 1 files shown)


bevanjkay:
@naveenkrdy The `:github_latest` livecheck strategy is only required in instances that the standard `:git` strategy isn't suitable. When the source url is from GitHub we can omit the livecheck stanza as it is inferred.

https://docs.brew.sh/Brew-Livecheck#general-guidelines

naveenkrdy:
> @naveenkrdy The `:github_latest` livecheck strategy is only required in instances that the standard `:git` strategy isn't suitable. When the source url is from GitHub we can omit the livecheck stanza as it is inferred.
> 
> [docs.brew.sh/Brew-Livecheck#general-guidelines](https://docs.brew.sh/Brew-Livecheck#general-guidelines)

thanks for the heads up.

PR #112101, opened 2021-10-05
Add IBM Semeru Cask

mbarbero:
I've verified that:

- [x] `brew audit --cask semeru-jdk-open` is error-free.
- [x] `brew style --fix semeru-jdk-open` reports no offenses.

- [x] `brew audit --new-cask semeru-jdk-open` worked successfully.
- [x] `brew install --cask semeru-jdk-open` worked successfully.
- [x] `brew uninstall --cask semeru-jdk-open` worked successfully.

### DIFF
--- a/Casks/semeru-jdk-open.rb
+++ b/Casks/semeru-jdk-open.rb
@@ -16,7 +16,7 @@ cask "semeru-jdk-open" do
     end
   end
 
-  pkg "ibm-semeru-open-jdk_x64_mac_#{version.before_comma}_#{version.after_comma}.pkg"
+  pkg "ibm-semeru-open-jdk_x64_mac_#{version.before_comma.tr("+", "_")}_#{version.after_comma}.pkg"
 
   uninstall pkgutil: "net.ibm-semeru-open.#{version.major}.jdk"
 end
--- a/Casks/semeru-jdk-open.rb
+++ b/Casks/semeru-jdk-open.rb
@@ -2,35 +2,22 @@ cask "semeru-jdk-open" do
   version "16.0.2+7,openj9-0.27.0"
   sha256 "151dfbf918682663e28187ed3b59b06e8c5c8ad39f57dada42612aa8290e8977"
 
-  version_before_plus = version.before_comma.sub(/\+\d+/, "")
-  version_after_plus = version.before_comma.sub(/^[^+]+\+/, "")
-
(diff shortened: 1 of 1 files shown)


bevanjkay:
Ping @samford - Can you please have a look at the livecheck here?

mstoodle:
Note that the JDK binary file names use underscore rather than plus (even though the directory name uses plus rather than underscore). So `16.0.2+7` is used in the directory name (because URLs can handle + as %2B) but `16.0.2_7` is used in the file name (because not all file systems can handle + in the file name). I believe that's the reason for the failed CI test.

samford (changes requested):

Casks/semeru-jdk-open.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  url "https://github.com/ibmruntimes/semeru#{version.major}-binaries/releases/download/jdk-#{version.before_comma}_#{version.after_comma}/ibm-semeru-open-jdk_x64_mac_#{version.before_comma.tr("+", "_")}_#{version.after_comma}.pkg",
+      verified: "github.com/ibmruntimes/semeru#{version.major}-binaries/"
+  name "IBM Semeru Runtime (JDK) Open Edition"
+  desc "Production-ready JDK with the OpenJDK class libraries and the Eclipse OpenJ9 JVM"
+  homepage "https://developer.ibm.com/languages/java/semeru-runtimes"
+
+  livecheck do
+    url "https://developer.ibm.com/languages/java/semeru-runtimes/downloads"
+    regex(%r{["'][^"']*ibm-semeru-open-jdk_x64_mac[._-](\d+(?:[._]\d+)+)[._-](.+)\.pkg["']}i)
```
samford, line 13:
```suggestion
    regex(/["'][^"']*ibm-semeru-open-jdk_x64_mac[._-](\d+(?:[._]\d+)+)[._-](.+)\.pkg["']/i)
```

I had originally included a `/` in the regex but forgot to switch this back to the `/.../` regex syntax after removing it.

bevanjkay:
Thanks @samford for your help here.

@gdams This looks good to go now for me. We can dismiss the notability audit, because the repo is only used to distribute the binaries.

mbarbero:
Thank you @samford, @bevanjkay, and @gdams for your help with this one! I've learned a lot!

> And BTW: SUPER thanks for putting this pull request together! We would be very happy to have the Semeru runtime available as a cask (I'm keen to install it myself :) ).

@mstoodle, you're welcome. I was missing it for myself as well ;)

PR #112116, opened 2021-10-05
Update leapp from 0.7.2 to 0.7.3

deiga:
Created with brew bump-cask-pr.

Change URL determining logic to match multi-url reality


gdams:
This change seems kinda awkward to me (I understand the problem it solves). Is it worth us asking if the leapp developers will just create the versioned URL path when they release a new version?

deiga:
Filed upstream issue https://github.com/Noovolari/leapp/issues/179

PR #112117, opened 2021-10-05
paper: add desc+livecheck+depends_on

neutric:

### DIFF
--- a/Casks/paper.rb
+++ b/Casks/paper.rb
@@ -9,7 +9,7 @@ cask "paper" do
 
   livecheck do
     url :homepage
-    regex(/["'][^"']*pap.er[._-]v?(\d+(?:\.\d+)+)\.dmg["']/i)
+    regex(/["'][^"']*pap\.er[._-]v?(\d+(?:\.\d+)+)\.dmg["']/i)
   end
 
   depends_on macos: ">= :el_capitan"
--- a/Casks/paper.rb
+++ b/Casks/paper.rb
@@ -4,7 +4,16 @@ cask "paper" do
 
   url "http://paperimg.meiyuan.in/pap.er_v#{version}.dmg"
   name "Paper"
+  desc "Live wallpaper app"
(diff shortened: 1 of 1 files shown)


Casks/paper.rb
```diff
@@ -4,8 +4,16 @@
 
   url "http://paperimg.meiyuan.in/pap.er_v#{version}.dmg"
   name "Paper"
+  desc "Live wallpaper app"
   homepage "https://paper.meiyuan.in/"
 
+  livecheck do
+    url :homepage
+    regex(/["'][^"']*pap\.er[._-]v?(\d+(?:\.\d+)+)\.dmg["']/i)
```
bevanjkay, line 12: @neutric I modified the regex slightly to be better anchored to the start and end of the area we are looking to capture. It also isn't necessary to use `strategy :page_match` as it is the default.

Casks/paper.rb
```diff
@@ -4,8 +4,16 @@
...
+  end
+
+  depends_on macos: ">= :el_capitan"
```
bevanjkay, line 15: @neutric According to the `info.plist` file in `pa.per.app` it should support back to Yosemite. Where did you find information about the minimum version?

```
	<key>LSMinimumSystemVersion</key>
	<string>10.10</string>
```

neutric: The homepage says "macOS X 10.11 or higher version" at the very bottom.

bevanjkay (approved):

Casks/paper.rb
neutric: @bevanjkay Thanks for the modification! Regarding ```strategy :page_match```: the documentation on the [livecheck](https://docs.brew.sh/Cask-Cookbook#stanza-livecheck) stanza might need some rewording then. It currently says 

> a livecheck should specify which strategy should be used to extract the version

whereas the documentation on [brew livecheck](https://docs.brew.sh/Brew-Livecheck) says
> Only use strategy when it’s necessary.

PR #112140, opened 2021-10-06
leapp 0.7.3

miccal:


bevanjkay:
@miccal There is still an upstream issue here that the previous version is not available using the same URL syntax

curl https://asset.noovolari.com/latest/Leapp-0.7.2.dmg
```
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>PFBDMVGMCAFR039N</RequestId><HostId>u7d4QCk8hqxZ9+WKAXmm8WyHQXHBMGT4A6siF26wQGsQ/73o1LF2uOXt5r9acsOf/burOIhMupY=</HostId></Error>%  
```

So the cask whenever a new release is published, until the version bump occurs.

deiga:
@miccal Should the documentation be changed to not reference `url do` blocks?

PR #112217, opened 2021-10-06
appium 1.22.0

miccal:
Token reference?

### DIFF
--- a/Casks/appium.rb
+++ b/Casks/appium.rb
@@ -1,10 +1,10 @@
 cask "appium" do
-  version "1.21.0"
-  sha256 "bdcb55be5c8193466f1bf6665e8f821711423e87abd7c836618a8fdded57eeea"
+  version "1.22.0"
+  sha256 "9d185b0bc905db84a199c71c7a5415dc5fd1d39c52487c4419eac21b9554ef9c"
 
-  url "https://github.com/appium/appium-desktop/releases/download/v#{version}/Appium-mac-#{version}.dmg",
+  url "https://github.com/appium/appium-desktop/releases/download/v#{version}/Appium-Server-GUI-mac-#{version}.dmg",
       verified: "github.com/appium/appium-desktop/"
-  name "Appium Desktop"
+  name "AAppium Server Desktop GUI"
   desc "Graphical frontend to Appium automation server"
   homepage "https://appium.io/"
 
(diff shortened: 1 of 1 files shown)


Casks/appium.rb
```diff
@@ -1,10 +1,10 @@
...
-  version "1.21.0"
-  sha256 "bdcb55be5c8193466f1bf6665e8f821711423e87abd7c836618a8fdded57eeea"
+  version "1.22.0"
+  sha256 "9d185b0bc905db84a199c71c7a5415dc5fd1d39c52487c4419eac21b9554ef9c"
 
-  url "https://github.com/appium/appium-desktop/releases/download/v#{version}/Appium-mac-#{version}.dmg",
+  url "https://github.com/appium/appium-desktop/releases/download/v#{version}/Appium-Server-GUI-mac-#{version}.dmg",
       verified: "github.com/appium/appium-desktop/"
-  name "Appium Desktop"
+  name "AAppium Server Desktop GUI"
```
gdams, line 7:
```suggestion
  name "Appium Server Desktop GUI"
```

gdams (approved):

miccal:
> Trying to understand what the difference is between this cask and https://github.com/Homebrew/homebrew-core/blob/master/Formula/appium.rb? Is this cask Appium Desktop? In which case it should likely be renamed to `appium-desktop` to avoid confusion

We drop `desktop` as per the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).

Following the token reference strictly, this should be re-named to `appium-server-gui`.

gdams:
>Following the token reference strictly, this should be re-named to appium-server-gui.

My interpretation of the token reference is that we should also remove `-gui` to it should just be `appium-server`?

vitorgalvao:
> My interpretation of the token reference is that we should also remove `-gui`.

The token reference is mostly comprehensive and `gui` isn’t singled out.

We remove “Desktop” and “Mac” because those are overused, redundant (*every* cask is for macOS and Desktop), and impair searchability.

`appium-server-gui` is the strictly correct name.

miccal:
@vitorgalvao I was a little hesitant on a rename:
```
==> Analytics
install: 151 (30 days), 409 (90 days), 1,268 (365 days)
```

PR #112276, opened 2021-10-07
XnViewMP: Add workaround for unidentified developer issue

cielavenir:


bevanjkay:
We don’t delete the quarantine, we add it and go to some lengths to do so.
Especially if there are security prompts, we shouldn’t be disabling a security feature.

To install something without quarantining, always at your own risk, use the --no-quarantine flag.

cielavenir:
Thank you `brew reinstall xnviewmp --cask --no-quarantine` worked. I added caveats instead because I believe some people should have the same problem.

miccal:
Thank you @cielavenir, but his is already mentioned in our documentation, for example [here](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/app_cant_be_opened.md).

PR #112285, opened 2021-10-08
Add Immersed (15.7)

stek29:
Immersed is desktop counterpart for the VR "office"/productivity app which allows to bring virtual screens into VR.

### DIFF
--- a/Casks/immersed.rb
+++ b/Casks/immersed.rb
@@ -2,7 +2,8 @@ cask "immersed" do
   version "15.7,129"
   sha256 :no_check
 
-  url "https://immersedvr.com/dl/Immersed.dmg"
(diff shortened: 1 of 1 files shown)


stek29:
Plugin installation is optional and can be skipped – but I haven't tested uninstall running without plugins installed, and I'm not sure if that should be handled differently.

coreaudiod kickstart comes from install.sh/uninstall.sh built into the app – unfortunately, those don't remove the plugins, so shelling out to them doesn't make sense.
coreaudiod kickstart is also in blackhole casks, but it's not used in casks for rogueamoeba apps, which also have audio plugins
so I'm not quite sure if coreaudiod kickstart should be present – I'd love to hear your opinion on this.

miccal:
Thank you @stek29.

PR #81315, opened 2020-04-24
Update R in r.rb from 3.6.3 to 4.0.0

vincentarelbundock:
After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/Homebrew/homebrew-cask/pulls
[already refused]: https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues
[the correct repo]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask

PR #112293, opened 2021-10-08
Fix and update versions for R, inlcuding ARM support

Amorymeltzer:
The current version is dependent on macOS >= 10.13, so this has been slightly wrong for about 18 months since #81315.  Furthermore, there is now a separate ARM build for the latest.

----

I'm not 100% on whether this is the ideal structure; is it okay to mix `MacOS.version` and `Hardware::CPU.intel?` so explicitly, or would it be preferable to nest the latter within an `else`?


SMillerDev (approved):
Thanks @Amorymeltzer !

aosavi:
Fixed it, thanks!

PR #112911, opened 2021-10-19
Add GitDock.app v0.1.11

kgoossens:

### DIFF
--- a/Casks/gitdock.rb
+++ b/Casks/gitdock.rb
@@ -8,8 +8,7 @@ cask "gitdock" do
   homepage "https://gitlab.com/mvanremmerden/gitdock"
 
   livecheck do
-    url "https://gitlab.com/api/v4/projects/28462485/releases"
-    regex(%r{gitdock/(\d+(?:\.\d+)+)/GitDock\.dmg}i)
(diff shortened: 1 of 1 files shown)


miccal:
Not sure what the issue is here:
```
Error: exception while auditing gitdock: undefined method `<' for nil:NilClass
```

miccal:
Did not work -- the problem must be in the `url`.

While we investigate this further @kgoossens, could you please try to add a [`zap` stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap), thanks.

miccal:
> @miccal It is because the `url` uses the `api/v4` syntax, so the audit for the repo is trying to navigate to `https://gitlab.com/api/v4/projects/api/v4` which returns a 404.

Yes I get that, but I am not 100 % sure that is the problem because both `cura-lulzbot` and `librewolf` _do not_ do that, but  `brew audit --online --new-cask` fails for them with the same error.

kgoossens:
I've added the `zap`-stanza but the brew audit failed again in the CI pipeline.
I'm checking with @MvRemmerden to see if we can create a permalink to the release artefacts which might be a workaround for the failure 🤞

PR #112973, opened 2021-10-20
Update oracle-jdk from 17 to 17.0.1, set sha256 to `:no_check`

jambt:

- [ ] `brew audit --cask <cask>` is error-free. (only adapted the checksums from [the downloads page](https://www.oracle.com/java/technologies/downloads/#jdk17-mac))
- [ ] `brew style --fix <cask>` reports no offenses. (only adapted the checksums from [the downloads page](https://www.oracle.com/java/technologies/downloads/#jdk17-mac))

### DIFF
--- a/Casks/oracle-jdk.rb
+++ b/Casks/oracle-jdk.rb
@@ -2,14 +2,9 @@ cask "oracle-jdk" do
   arch = Hardware::CPU.intel? ? "x64" : "aarch64"
 
   version "17.0.1"
+  sha256 :no_check   # required as upstream package is updated in place
 
-  url "https://download.oracle.com/java/#{version.major}/latest/jdk-#{version}_macos-#{arch}_bin.dmg"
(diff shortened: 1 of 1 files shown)


bevanjkay:
> Hi @bevanjkay and thanks for your help. The `nocheck`-option makes a lot of sense. However, should we also set the `version` back to `17` then, since it might upgrade in the future?

We should still set the version to be the correct version, as it can trigger `brew upgrade` to run for this cask.
I have updated the URL to use `#{version.major}` so only the `17` is used to generate the download link.

zhangyoufu:
> The versioned downloads are only available after the applicable version is no longer the stable release.
This is not true. All release versions can be downloaded from archive URL with precise version number.

I would expect homebrew to use archive URL instead of latest URL.

bevanjkay:
@zhangyoufu You are welcome to submit a PR with the changes

PR #113263, opened 2021-10-27
monitorcontrol: undo previous commit

neutric:
I must have not been awake yet when I read the changelog and mistook "only on Big Sur and above" regarding certain features to mean "only for Big Sur and above" regarding the current release. Sorry for that!

### DIFF
--- a/Casks/monitorcontrol.rb
+++ b/Casks/monitorcontrol.rb
@@ -1,22 +1,12 @@
 cask "monitorcontrol" do
-  if MacOS.version <= :catalina
-    version "3.1.1"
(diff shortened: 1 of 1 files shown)


quinncomendant:
@neutric, why did you revert commit 36601ff? It was useful to automatically get the legacy version on older systems. I'm still on Mojave, and can't run v4. 😢

v4.0.0 requires Big Sur and above.

neutric:
@quinncomendant Thanks for pointing that out! I wasn't as thorough as I should have been and didn't notice their Compatibility section. I was probably dizzy from working through the lengthy list of changes that came with v4.0.0. Fix is already merged.

PR #113481, opened 2021-10-30
livecheck update: graceful exit when regex fails page.match

bevanjkay:
This update causes livecheck to fail gracefully when `page.match` fails inside of the `do` block.
The documentation needs to be updated to match this pattern also.


miccal:
@bevanjkay any chance you can do the same thing for `versions` and `drivers`?

bevanjkay:
@miccal Yes I will do

PR #113597, opened 2021-11-01
Slack: pin pre-Mojave at v4.19.0

augmentedfourth:

_not applicable_

### DIFF
--- a/Casks/slack.rb
+++ b/Casks/slack.rb
@@ -1,12 +1,18 @@
 cask "slack" do
   arch = Hardware::CPU.intel? ? "x64" : "arm64"
 
-  version "4.21.1"
+  if MacOS.version >= :mojave
+    version "4.21.1"
 
-  if Hardware::CPU.intel?
-    sha256 "8f64fdefc5354fbb75a56f5797b6a84a27a2c6f6f05fd13a4a6efc550fa510f6"
+    if Hardware::CPU.intel?
+      sha256 "8f64fdefc5354fbb75a56f5797b6a84a27a2c6f6f05fd13a4a6efc550fa510f6"
+    else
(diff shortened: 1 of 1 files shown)


augmentedfourth:
I don't know what happened here, but the CI failure seems to be an aberration that is out of my control. Can we re-run this, or merge/approve anyway?

miccal:
According to https://slack.com/intl/en-au/help/articles/115002037526-System-requirements-for-using-Slack the Slack app requires macOS 10.14 or above, and according to https://slack.com/intl/en-au/help/articles/1500001836081-Slack’s-deprecation-schedule#mac-1 support for macOS versions 10.13 and below was deprecated on the 1st of September 2021.

I think it would be better, then, to add a `depends_on macos: ">= :mojave"` then continue to provide old versions that are deprecated upstream.

augmentedfourth:
I still have an old 2010 Mac that can't run an OS past 10.13, and Slack 4.19.0 still installs successfully on it, which is why I wrote the PR this way. I'd rather be able to still manage the installation via Homebrew. I know the tool is deprecated for High Sierra, but the warning message requests that users create PRs to resolve their own issues, which is what I've done here.

miccal:
> I still have an old 2010 Mac that can't run an OS past 10.13, and Slack 4.19.0 still installs successfully on it, which is why I wrote the PR this way. I'd rather be able to still manage the installation via Homebrew. I know the tool is deprecated for High Sierra, but the warning message requests that users create PRs to resolve their own issues, which is what I've done here.

Thanks @augmentedfourth, but since support for this version of macOS has been deprecated upstream we cannot accept this change.

Please consider hosting this version [in your own Tap](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap) so that your installation can still be managed by Homebrew.

PR #113602, opened 2021-11-01
Update gpg-suite from 2021.1,105 to 2021.2

lebdron:

### DIFF
--- a/Casks/gpg-suite-no-mail.rb
+++ b/Casks/gpg-suite-no-mail.rb
@@ -1,15 +1,14 @@
 cask "gpg-suite-no-mail" do
-  version "2021.1,105"
-  sha256 "383bd6ab4791ee51e0f67955ad1ab70bb3a2a1e5c71f6a7f42f53b92684106e0"
+  version "2021.2"
+  sha256 "828df766dec3aa1ad4fbd29c8dd6a830d756694fae1a9f38ae78b6cfcf403f52"
 
-  url "https://releases.gpgtools.org/GPG_Suite-#{version.before_comma}_#{version.after_comma}.dmg"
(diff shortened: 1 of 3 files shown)


samford:
The `appcast.xml` file used in the `livecheck` block doesn't include information for 2021.2 (yet?). Seeing as the appcast doesn't give us the correct latest version, it may be better to simply parse the version from the download link on the homepage:

```ruby
  livecheck do
    url :homepage
    regex(/href=.*?GPG_Suite[._-]v?(\d+(?:\.\d+)+)\.dmg/i)
  end
```

The `gpg-suite-pinentry` and `gpg-suite-no-mail` casks should be updated as well but I'll leave it up to our homebrew/cask maintainers whether we should do this altogether or in separate PRs. However that happens, the `livecheck` blocks in the variant casks can be updated as follows (since they use the same `url` as `gpg-suite` and `gpg-suite` is the canonical cask):

```ruby
  livecheck do
    cask "gpg-suite"
  end
```

This helps to ease our maintenance burden, as we only have to keep the `gpg-suite` `livecheck` block up to date and the others will automatically inherit any changes.

lebdron:
@samford Thank you for the reply! Should we also handle an optional `after_comma` case, or that should be done only if it appears in the link again?

miccal:
Let's update `gpg-suite-no-mail` and `gpg-suite-pinentry` in a separate PR.

NSBrianWard:
For whatever reason, updating via brew didn't actually increment my gpg-suite version, despite seeing that I got the update in brew. I was still stuck on 2021.1 according to the gpg app itself (from the preferences pane). I had to uninstall/reinstall the cask to get the correct version. Unsure what happened there. It seems to be working now though!

PR #113619, opened 2021-11-02
x2goclient: add desc

neutric:

### DIFF
--- a/Casks/x2goclient.rb
+++ b/Casks/x2goclient.rb
@@ -13,8 +13,8 @@ cask "x2goclient" do
   homepage "https://wiki.x2go.org/doku.php"
 
   livecheck do
-    url :url
-    strategy :extract_plist
+    url "https://wiki.x2go.org/doku.php/doc:release-notes-mswin"
+    regex(/x2goclient[._-]v?(\d+(?:\.\d+)+)/i)
   end
 
   depends_on macos: ">= :el_capitan"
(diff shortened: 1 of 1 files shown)


neutric:
Is there a way to tell livecheck to extract parts of the version string taken from the cask description? The current version `4.1.2.2` can be scraped off of the homepage easily, but the date that is part of the download URL and thus part of `version` (4.1.2.2,20200213 ) is a problem.

neutric:
`:extract_plist` is also failing locally with `Error: x2goclient: Unable to get versions`

miccal:
@neutric happy with this?

neutric:
@miccal Thanks for your help! I prefer versioned URLs, or rather the checksumming that comes with them, but this will have to do I suppose.

miccal:
> @miccal Thanks for your help! I prefer versioned URLs, or rather the checksumming that comes with them, but this will have to do I suppose.

I do too, but the problem was coming up with a `livecheck` that would work for the versioned `url`'s and I noticed from https://code.x2go.org/releases/ that the folder for `binary-macosx/` was not updated in a regular manner, so I think it is better to stick with the `X2GoClient_latest_macosx_10_*.dmg` links.

neutric:
Totally understand! I tried using `binary-macosx/x2goclient/releases/`, which at least for the current version was updated at the same time the file for `latest` was, but then got stuck missing the date that is part of the versioned URL.

PR #113621, opened 2021-11-02
gpg-suite-no-mail 2021.2

miccal:
https://github.com/Homebrew/homebrew-cask/pull/113602


miccal:
Ping @samford.

samford:
This `livecheck` block works locally but the audit failure occurs because `Cask::Audit#check_livecheck_version` wasn't updated to handle a `livecheck` block that uses `#cask` to reference another cask when I introduced that feature. At the time, I wasn't aware that the cask audit uses `Livecheck#latest_version` internally and would need to be kept up to date like `#livecheck_result` in `brew bump`. This will be fixed by Homebrew/brew#12368.

PR #113801, opened 2021-11-04
electrum: fix livecheck

n-thumann:

---
The livecheck of `electrum` is broken, because there're only tags, but not releases any more (see [Wayback Machine](https://web.archive.org/web/20201111162205/https://github.com/spesmilo/electrum/releases)).


bevanjkay:
Thanks @n-thumann - I have updated the livecheck to use the homepage download panel instead, because the binaries are not hosted with GitHub anymore.

PR #113824, opened 2021-11-04
Add uninstall script for Disk Drill.app

ZepDonald:
Disk Drill comes with it's own uninstall script. Running this script instead of removing anything via Cask stanzas is much easier, more future proof and more reliable.

### DIFF
--- a/Casks/disk-drill.rb
+++ b/Casks/disk-drill.rb
@@ -14,7 +14,10 @@ cask "disk-drill" do
 
   app "Disk Drill.app"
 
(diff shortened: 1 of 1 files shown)


bevanjkay:
@ZepDonald The zap could still be useful if the app was removed outside of Homebrew as you can still clear any remaining support files by passing `--force` to the `brew uninstall --zap <token>`

ZepDonald:
Good to know! :)
I will add some more files in the next days, thanks to the official installer.

PR #113913, opened 2021-11-06
Updated zap for Disk Drill.app

ZepDonald:
Transfered official uninstaller script for zap command. If Disk Drill
was uninstalled outside of Homebrew, a forced zap will still remove
everything that was installed by Disk Drill.

This pull request is a follow up to #113824.

### DIFF
--- a/Casks/disk-drill.rb
+++ b/Casks/disk-drill.rb
@@ -1,5 +1,5 @@
 cask "disk-drill" do
-  version "4.5.369"
+  version "4.6.370"
   sha256 :no_check
 
   url "https://dl.cleverfiles.com/diskdrill.dmg"
--- a/Casks/disk-drill.rb
+++ b/Casks/disk-drill.rb
@@ -19,9 +19,39 @@ cask "disk-drill" do
     sudo:       true,
   }
 
-  zap trash: [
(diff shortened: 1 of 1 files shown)


miccal:
Why is all this necessary in `zap`?

The `zap` stanza is really only for removing user files and shared resources.

ZepDonald:
As far as I can see, these are all shared resources, that are getting created while using the app. I'm not sure, if all of these files are always getting created, so it might be overkill?! But all of this is part of the official uninstaller script, therefore it seems reasonable to have such a huge 'zap' section. Especially for situations, where the app got uninstalled outside of Homebrew, a forced zap will remove everything (which will be impossible with just the 'uninstall' command, because the uninstall script ist part of the app bundle).

But feel free to remove parts you think are unnecessary. :)

miccal:
> all of this is part of the official uninstaller script, therefore it seems reasonable to have such a huge 'zap' section. 

If it is part of the `uninstall` script, why re-do it in the `zap`?

> Especially for situations, where the app got uninstalled outside of Homebrew, a forced zap will remove everything (which will be impossible with just the 'uninstall' command, because the uninstall script ist part of the app bundle).

We cannot really control the uninstall of things not managed by Cask.

PR #113955, opened 2021-11-07
Update deltawalker from 2.5.6 to 2.6.0

neutric:

> If installing over an existing copy, run the 'run-me-first' shell script from the DMG package, or the following in Terminal:
> rm -fr ~/Library/Containers/com.deltopia.DeltaWalker/Data/Library/Application\ Support

@miccal The shell script is a bit more granular than `rm -rf`. How would I invoke it? And is there a risk to be considered when automatically executing shell scripts provided by upstream which might make `rm -rf` preferable?

### DIFF
--- a/Casks/deltawalker.rb
+++ b/Casks/deltawalker.rb
@@ -15,6 +15,11 @@ cask "deltawalker" do
 
   app "DeltaWalker.app"
 
+  installer script: {
+    executable: "#{staged_path}/run-me-first",
+    sudo:       false,
+  }
+
   zap trash: [
     "~/Library/Caches/com.deltopia.DeltaWalker",
     "~/Library/Containers/com.deltopia.DeltaWalker",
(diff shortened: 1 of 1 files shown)


miccal:
> > If installing over an existing copy, run the 'run-me-first' shell script from the DMG package, or the following in Terminal:
> > rm -fr ~/Library/Containers/com.deltopia.DeltaWalker/Data/Library/Application\ Support
> 
> @miccal The shell script is a bit more granular than `rm -rf`. How would I invoke it? And is there a risk to be considered when automatically executing shell scripts provided by upstream which might make `rm -rf` preferable?

That should do it :)

neutric:
@miccal The script was supposed to run before the updated app is run. The uninstall script doesn't run during the removal of the old version during an update, does it?

miccal:
No, but it will for the next revision bump.

PR #113967, opened 2021-11-07
bbedit: Added binary installation for helpers

Kreeblah:

### DIFF
--- a/Casks/bbedit.rb
+++ b/Casks/bbedit.rb
@@ -24,6 +24,10 @@ cask "bbedit" do
   depends_on macos: ">= :mojave"
 
   app "BBEdit.app"
+  binary "#{appdir}/BBEdit.app/Contents/Helpers/bbedit_tool", target: "bbedit"
+  binary "#{appdir}/BBEdit.app/Contents/Helpers/bbdiff"
+  binary "#{appdir}/BBEdit.app/Contents/Helpers/bbfind"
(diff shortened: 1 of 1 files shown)


miccal:
Aren't these installed by running the command "Install Command Line Tools" from within the app itself?

I am concerned that adding these this way may mess with the recommended way to install them.

Kreeblah:
They are, yes, but that menu option puts them in a subdirectory in `/usr/local`, regardless of which architecture (ARM vs. Intel) it's running on.  Having them linked from the cask here does result in BBEdit reporting them as correctly installed, and they do work properly.  However, I did forget to add the man pages when I submitted this originally, which are the other things that get installed, so I updated this with another commit to do that.

This should, at this point, be identical in functionality to using the menu item to install them, except the man pages and binaries will be linked to where they're located within the app bundle instead of copied to locations within `/usr/local`.

miccal:
Thanks @Kreeblah, and one more question -- does running "Install Command Line Tools" from within the app change the name of `bbedit_tool` to just `bbedit` like you have done here?

Kreeblah:
Sure, no problem.

Regarding the renaming, the application's "Install Command Line Tools" option does also rename the binary from `bbedit_tool` to `bbedit`.  It also looks for `bbedit` (not `bbedit_tool`) when determining what the installed helper version is, and the man page refers to it as just `bbedit`.

PR #114303, opened 2021-11-14
update vimr and fix livecheck

jiaming-shi:

### DIFF
--- a/Casks/vimr.rb
+++ b/Casks/vimr.rb
@@ -10,7 +10,7 @@ cask "vimr" do
 
   livecheck do
     url :url
-    regex(%r{href=.*?/VimR-v?(\d+(?:\.\d+)*-\d+\.\d+)\.t}i)
+    regex(/^v?(\d+(?:[.-]\d+)+)$/i)
   end
 
   auto_updates true
--- a/Casks/vimr.rb
+++ b/Casks/vimr.rb
@@ -10,7 +10,6 @@ cask "vimr" do
 
   livecheck do
     url :url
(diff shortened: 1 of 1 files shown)


Casks/vimr.rb
```diff
@@ -11,7 +11,7 @@
   livecheck do
     url :url
     strategy :github_latest
```
samford, line 13:
```suggestion
```

Casks/vimr.rb
```diff
@@ -11,7 +11,7 @@
...
-    regex(%r{href=.*?/VimR-v?(\d+(?:\.\d+)*-\d+)\.t}i)
+    regex(%r{href=.*?/VimR-v?(\d+(?:\.\d+)*-\d+\.\d+)\.t}i)
```
samford, line 14:
```suggestion
    regex(/^v?(\d+(?:[.-]\d+)+)$/i)
```

samford (changes requested):
We only use the `GithubLatest` strategy when it's both correct and necessary. While it provides the correct version here, we can accurately obtain the same newest version from the Git tags using a modified version of the standard regex for tags like `1.2.3`/`v1.2.3`.

In this scenario, we can simply modify the delimiter in the repeating part from `\.` to `[.-]` and it will elegantly match existing formats (e.g., `v0.34.0-355`, `v0.35.0-20211105.213803`) while also accommodating potential variants that may appear in the future. This is a more flexible approach than requiring a specific format for the suffix (or any suffix at all), as that approach is more prone to break (as seen here).

PR #114311, opened 2021-11-14
Update fs-uae and create fs-uae-launcher

Kreeblah:

This submission is a bit unusual.  The FS-UAE project split up what used to be included in a single DMG into two tarballs, per https://fs-uae.net/download.  The old URL that used to reflect stable updates stopped being updated at some point as well.

So, this submission splits the old cask into two that depend on each other.  That doesn't feel ideal, but given that the applications need to be installed together, it was the only thing that came to mind.  Additionally, the FS-UAE project is also providing ARM builds now, so I added logic to select those when applicable.

### DIFF
--- a/Casks/fs-uae-launcher.rb
+++ b/Casks/fs-uae-launcher.rb
@@ -1,0 +1,26 @@
+cask "fs-uae-launcher" do
+  arch = Hardware::CPU.intel? ? "x86-64" : "ARM64"
+
+  version "3.1.42"
+
+  if Hardware::CPU.intel?
+    sha256 "c84db7c5bf04b183f86fcce6828d6e85968ddc3c350aef0a0abbfe68f506756e"
+  else
+    sha256 "a06656df30572e76623b7cad29b93eece0c7938029889574c36b65315a508ad1"
+  end
+
+  url "https://fs-uae.net/files/FS-UAE-Launcher/Stable/#{version}/FS-UAE-Launcher_#{version}_macOS_#{arch}.tar.xz"
+  name "FS-UAE Launcher"
+  desc "Amiga emulator launcher"
+  homepage "https://fs-uae.net/"
+
+  livecheck do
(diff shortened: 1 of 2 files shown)


Kreeblah:
It looks like the CI piece failed because of the cyclic dependency.  During my tests locally, I commented out the dependency since I was installing casks from a tree checked out from my cloned repo, and the existing cask installed an app bundle that conflicted with one of these.

So, I missed that check.  Are there any suggestions for how to better approach this?  I don't think Homebrew has a notion of metacasks or some such, which is what my first impulse would have been.

miccal:
I think if we just add `depends_on cask: "fs-uae-launcher"` to the original `fs-uae` Cask then this should be fine.

bevanjkay:
@miccal CI is failing on this because of `launcher` in the token. This may be a special case, as the cask needs to be split in two. 

@Kreeblah Is the launcher required to be able to use the emulator? If so then they probably should depend on each other, if the launcher isn't required then the current is okay.

Kreeblah:
@bevanjkay - It . . . depends.  The launcher is what's typically used to generate config files for the emulator (the emulator itself doesn't have the ability to do that), which are required to use it.  It's possible (if tedious and error-prone) to write them by hand, though, so I see valid arguments either way on whether the launcher is considered required for the emulator to function.

miccal:
> @miccal CI is failing on this because of `launcher` in the token. This may be a special case, as the cask needs to be split in two.

Yes it is a special case -- same thing happens in the `eve-launcher` Cask:
```
|-> brew audit --cask --appcast --online --new-cask eve-launcher
==> Downloading https://binaries.eveonline.com/EveLauncher-1952584.dmg
audit for eve-launcher: failed
 - cask token mentions launcher
```

miccal:
I am going to merge this and see how we go.

Thanks @Kreeblah.

PR #114328, opened 2021-11-14
spotify: minimize version string

FnControlOption:

### DIFF
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -18,9 +18,10 @@ cask "spotify" do
   livecheck do
     url :url
     strategy :extract_plist do |items|
-      items["com.spotify.client"].version.match(/^(\d+(?:\.\d+)+)[._-]g(\h+)[._-](\d+)$/)&.then do |match|
-        "#{match[1]},#{match[2]},#{match[3]}"
-      end
+      match = items["com.spotify.client"].version.match(/^(\d+(?:\.\d+)+)[._-]g(\h+)[._-](\d+)$/i)
+      next if match.blank?
+
+      "#{match[1]},#{match[2]},#{match[3]}"
     end
   end
 
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -2,9 +2,9 @@ cask "spotify" do
   arch = Hardware::CPU.intel? ? "" : "ARM64"
 
   if Hardware::CPU.intel?
-    version "1.1.72.439.gc253025e,1.1.72.439.gc253025e-34"
+    version "1.1.72.439,c253025e,34"
   else
-    version "1.1.69.612.gb7409abc,1.1.69.612.gb7409abc-13"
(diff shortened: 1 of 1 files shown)


bevanjkay:
Ping @samford

Casks/spotify.rb
```diff
@@ -17,7 +17,11 @@
 
   livecheck do
     url :url
-    strategy :extract_plist
+    strategy :extract_plist do |items|
+      items["com.spotify.client"].version.match(/^(\d+(?:\.\d+)+)[._-]g(\h+)[._-](\d+)$/)&.then do |match|
+        "#{match[1]},#{match[2]},#{match[3]}"
+      end
```
samford, line 23:
```suggestion
      match = items["com.spotify.client"].version.match(/^(\d+(?:\.\d+)+)[._-]g(\h+)[._-](\d+)$/i)
      next if match.blank?

      "#{match[1]},#{match[2]},#{match[3]}"
```

The `&.then do |match|` approach is one fewer line than the existing style (in the suggestion above) but it may not work as well for `strategy` blocks where we have more than one conditional early return (the [`bash` formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bash.rb) is the unofficial poster child for lengthy `strategy` blocks).

If we needed to extend a `strategy` block using `#then` in the future, we would seemingly have to either chain more `#then`s (which may be syntactically unappealing and/or harder to read) or reformat the `strategy` block to the early return format (which would be an unnecessary hurdle). We would also end up in a situation where some `strategy` blocks use `#then` and others use an early return and it may confuse maintainers/contributors about which should be used in a specific context. In general, I've been trying to improve consistency of `strategy` blocks (still quite a bit of work to do on that front) and this feels like it may be a step backward.

Additionally, we can only protect against a `nil` value when using the safe navigation operator, whereas we can enforce an arbitrary condition (or combination of conditions) using `next if`. Most of the existing early return conditions are like `next if match.blank?` but we do have some that use a combination of conditions (e.g., `next if download_group.blank? || product_id.blank? || pid.blank?`) or return a value (e.g., `next version if item.short_version == "0.7.3"`).

Though `#then` can be somewhat tidy in certain circumstances (e.g., the ~176 cask `strategy` blocks like this one), I'm of the opinion that we should stick to the existing style for the sake of consistency and the reasons mentioned above.

samford (commented):
General cask question: Should this `version` use a colon to separate the third part (e.g., `1.1.72.439,c253025e:34`) or is this only done when it's technically necessary to be able to separate the version parts (e.g., interpolating the individual `version` parts into the `url`)?

Casks/spotify.rb
bevanjkay: While I do understand that the `version` for this particular cask was lengthy, and it can make commit messages quite long, I see the addition of a `livecheck` block as being an unnecessary point of potential failure, as opposed to using the standard `extract_plist` strategy.

PR #114374, opened 2021-11-15
Combine `sbarex-qlmarkdown` with `qlmarkdown` for compatibility

elsiehupp:

***

I'm not sure if this is the correct approach, but the problem I'm trying to solve is that the existing [`qlmarkdown`](https://formulae.brew.sh/cask/qlmarkdown) is abandoned and does not work on Catalina or later ([and this is a known issue](https://github.com/toland/qlmarkdown/issues)), while [`sbarex-qlmarkdown`](https://formulae.brew.sh/cask/sbarex-qlmarkdown) works exclusively with Catalina or later but has limited discoverability.

I haven't consulted with either @toland or @sbarex about this, and the README at [sbarex/QLMarkdown](https://github.com/sbarex/QLMarkdown) would need to be updated with the new cask name.

I did `install` and `uninstall` on both macOS Big Sur and macOS High Sierra, but I'm not sure if I ran them correctly. `brew install --cask path/to/qlmarkdown.rb` on Big Sur reported no errors correctly and correctly installed the application, while on High Sierra it reported no errors but seemingly did not actually install anything. In both cases `brew uninstall --cask path/to/qlmarkdown.rb` worked correctly.


miccal:
Thank you @elsiehupp but this is not the way to handle this.

I will change the name of the old `qlmarkdown` Cask to `toland-qlmarkdown` and mark it as `discontinued`, and then rename the old `sbarex-qlmarkdown` Cask to just `qlmarkdown`.

elsiehupp:
@miccal yes, I wasn’t sure if this was the correct approach, so thanks for fixing the underlying problem regardless!

FWIW you probably do want something like `depends_on macos: "< :catalina"` on `toland-qlmarkdown` and something like `depends_on macos: ">= :catalina"` on `sbarex-qlmarkdown` (now just `qlmarkdown`). (It’s unclear whether or not `toland-qlmarkdown` runs on Catalina.)

miccal:
> @miccal yes, I wasn’t sure if this was the correct approach, so thanks for fixing the underlying problem regardless!

No problem.

> FWIW you probably do want something like `depends_on macos: "< :catalina"` on `toland-qlmarkdown`

This seems to work on Catalina and higher if installed with the `--no-quarantine` flag, so unless those requirements are specified upstream we would not add it.

> and something like `depends_on macos: ">= :catalina"` on `sbarex-qlmarkdown` (now just `qlmarkdown`). (It’s unclear whether or not `toland-qlmarkdown` runs on Catalina.)

This one already has a `depends_on macos: ">= :catalina"`.

PR #114455, opened 2021-11-16
Update media-center to 28.00.79

nrlquaker:

### DIFF
--- a/Casks/media-center.rb
+++ b/Casks/media-center.rb
@@ -10,7 +10,7 @@ cask "media-center" do
   livecheck do
     url "https://www.jriver.com/download.html"
     strategy :page_match do |page|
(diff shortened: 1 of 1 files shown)


nrlquaker:
closes #114454

nrlquaker:
`livecheck` fails. Seems like its because of the version I used is not listed on the download page. Only on forum. That's why its failing. Any ideas how to fix it?

bevanjkay:
@nrlquaker According to the downloads page, version `28.00.79` is the latest universal binary. We can bump to this for now?

nrlquaker:
@bevanjkay I think, yes.
I remembered that forum releases are not super stable. They do some user testing and post stable one to the download page.

PR #114623, opened 2021-11-19
Re-add TeamSpeak Client 3.5.6.

reitermarkus:

---

3.5.6 does not work on Big Sure or Monterey, but there is a 3.5.7-beta.1 which does. The `livecheck` only covers the stable version, so it is currently failing.

### DIFF
--- a/Casks/teamspeak-client.rb
+++ b/Casks/teamspeak-client.rb
@@ -1,0 +1,30 @@
+cask "teamspeak-client" do
+  desc "Voice communication client"
+
+  if MacOS.version <= :catalina
+    version "3.5.6"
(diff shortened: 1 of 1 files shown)


miccal:
@reitermarkus This was removed because it states on their [download page:](https://teamspeak.com/en/downloads/):

> Mirroring or linking directly to our files is prohibited without our express, written consent.

reitermarkus:
Ah, missed that. I have contacted them, let's see.

reitermarkus:
Okay, I got a response:

> You are welcome to link to our download page http://www.teamspeak.com/downloads but please don't mirror the actual files. In other words, please link to our servers.

So it should be fine as long as we only link to their official servers, which we do in any case.

PR #114668, opened 2021-11-20
Update alt-tab

lwouis:

### DIFF
--- a/Casks/alt-tab.rb
+++ b/Casks/alt-tab.rb
@@ -3,8 +3,8 @@ cask "alt-tab" do
   sha256 "f86ecc32fe99e58a6dede0364ea3490caee878eac6788975bb4976944ab1e2f0"
 
   url "https://github.com/lwouis/alt-tab-macos/releases/download/v#{version}/AltTab-#{version}.zip"
-  name "alt-tab"
-  desc "Utility to set up alt-tab to switch between windows"
+  name "AltTab"
+  desc "Windows alt-tab on macOS"
   homepage "https://github.com/lwouis/alt-tab-macos"
 
   livecheck do
@@ -24,5 +24,6 @@ cask "alt-tab" do
     "~/Library/Caches/com.lwouis.alt-tab-macos",
     "~/Library/Cookies/com.lwouis.alt-tab-macos.binarycookies",
     "~/Library/Preferences/com.lwouis.alt-tab-macos.plist",
+    "~/Library/LaunchAgents/com.lwouis.alt-tab-macos.plist",
   ]
 end
--- a/Casks/alttab.rb
+++ b/Casks/alttab.rb
@@ -1,10 +1,10 @@
-cask "alt-tab" do
+cask "alttab" do
   version "6.28.0"
   sha256 "f86ecc32fe99e58a6dede0364ea3490caee878eac6788975bb4976944ab1e2f0"
 
   url "https://github.com/lwouis/alt-tab-macos/releases/download/v#{version}/AltTab-#{version}.zip"
   name "AltTab"
-  desc "Windows alt-tab on macOS"
+  desc "Enable Windows-like alt-tab"
   homepage "https://github.com/lwouis/alt-tab-macos"
 
   livecheck do
@@ -23,7 +23,7 @@ cask "alt-tab" do
     "~/Library/Application Support/com.lwouis.alt-tab-macos",
     "~/Library/Caches/com.lwouis.alt-tab-macos",
     "~/Library/Cookies/com.lwouis.alt-tab-macos.binarycookies",
-    "~/Library/Preferences/com.lwouis.alt-tab-macos.plist",
     "~/Library/LaunchAgents/com.lwouis.alt-tab-macos.plist",
+    "~/Library/Preferences/com.lwouis.alt-tab-macos.plist",
   ]
 end


lwouis:
I see that the CI build failed with `Description shouldn't contain the platform.`. I think this is an improper observation. I changed the description to `Windows alt-tab on macOS` which is the [official slogan on our website](https://alt-tab-macos.netlify.app/):

image

AltTab brings the famous "alt-tab" feature from Windows, to macOS. I think the description is really helpful to understand what the app does, at a glance. Mentioning "Windows" and "macOs" in the description is an easy way to describe what it does, and does not go against the _spirit_ of the CI rule I believe.

neutric:
I am not sure that the majority of macOS users associates the same rich meaning with Windows alt-tab as you do. Also, Homebrew in my mind is not primarily a software discovery mechanism which further limits the role `desc` plays. On top, Homebrew Casks' only target platform is macOS, so mentioning it explicitly is unnecessary. How about something like "Enable Windows-like app switching"? Please also sort the zap stanza alphabetically.

miccal:
Re-named to `alttab` as per the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).

lwouis:
@neutric @miccal could you please revert the cask name change? It's going to break people scripts and i would have to update all the places where we mention the name on our website, docs, etc.

miccal:
@lwouis no, the token reference I linked above should be followed.

lwouis:
@miccal i'm finding the value of not breaking existing infra and people's workflow may be greater than the value of consistent naming.

i understand the position to have consistent naming. However if an inconsistent name passed the process, i think we could imagine an exception here. The cask has had this name for around 2 years now.

Is there any way you could avoid this rename and maintain compatibility. There is high value is not changing this name, and i believe little value in changing this name

miccal:
Ping @vitorgalvao.

ryenus:
@miccal I think your improvement to the description is ok, appreciated that.

But the [change](https://github.com/Homebrew/homebrew-cask/pull/114668/commits/6783510bd0) from **`alt-tab`** to `alttab` is a bit **overstepping**.

* First, `alt-tab` is not only the name for the cask, but also the key combination of <kbd>**Alt**</kbd> and <kbd>**Tab**</kbd>. Plus that this app brings a nice feature from other platforms to mac, so it feels more right for the target audience to refer to the key combination as `alt-tab`, **with the hyphen**, rather than `alttab` which seems really odd. Even though on mac the shortcut keys are typically showed without hyphens, but that's when the symbols <kbd>⌘⇧⌥⌃</kbd> are used, and they're mainly for the shortcuts themselves, not for places where a name is needed.

* From the same token reference, in the section [Converting the Simplified Name To a Token](https://docs.brew.sh/Cask-Cookbook#converting-the-simplified-name-to-a-token), it also stated that:
   > * Hyphens stay hyphens.

* Looking at existing casks, currently there're **1180** casks having `-` in their name, more than **1/4** of all the 3927 casks:

   ```sh
   $ echo "$(ls -1 Casks/*-* | wc -l) / $(ls -1 Casks/* | wc -l)"
   1180 / 3927
   ```

Last but not least, for such kind of naming change, why not check with the @lwouis first? He's not only the author of the PR, but also the author of the software itself. the community members are also a key part of the Homebrew ecosystem, they also need to be respected.

Thank you

neutric:
> But the [change](https://github.com/Homebrew/homebrew-cask/pull/114668/commits/6783510bd0) from **`alt-tab`** to `alttab` is a bit **overstepping**.

> so it feels more right for the target audience to refer to the key combination as alt-tab, with the hyphen, rather than alttab which seems really odd.

This cask was created in https://github.com/Homebrew/homebrew-cask/pull/70844/ with the app bundle name being `alt-tab-macos.app`, which, according to the above-mentioned token reference lead to the cask name being `alt-tab`. The app bundle was then renamed later on, and is `AltTab.app` right now. 

Since renaming the app bundle does not seem to be sacred to upstream, might I suggest a re-release with `Alt-Tab.app` as the bundle name? This would lead to the previous cask name of `alt-tab` and everybody could be 🥳 .

lwouis:
The app bundle was named `alt-tab-macos.app` in 2019, in the very early days of the app. It was named AltTab.app 2 days later in the cask, and has been stable for 2 years now.

Notice that both these commits were not for me. Users of the app did this work. Later on, I noticed people caring about homebrew, and fixed some issues with it, as well as to add a CI/CD step to keep the cask up-to-date automatically after each release.

This is for background. Now regarding the situation and how to move forward.

> Last but not least, for such kind of naming change, why not check with the @lwouis first? He's not only the author of the PR, but also the author of the software itself. the community members are also a key part of the Homebrew ecosystem, they also need to be respected.

Thank you a lot @ryenus for saying this. I must say I have felt quite disrespected so far in this PR. I can't chose the name and description of my own app I worked 2 years on my weekends for free to help other people? I didn't even complain about the description not being the one I want, and the one we use on the official website. The one that you guys selected is fine with me. I found it quite rude to not pick the one I wanted, and quickly merge things before I get a chance to react and give my opinion. However I'm open minded, and I think the description is also good, so no problem there.

Now when you change the name of the cask, it creates work on my side to find and update all the places on our website, documentation, discord, etc, to update it. It also break every user scripts/workflows that were relying on the name being `alt-tab`.

It's been stable for 2 years, everybody is used to it. I don't see value in breaking that situation. I also feel it would be nice to take it a bit slower and be careful before rushing changes, when I'm here the author, representing a project I worked 2 years on, and which I'm very in tune with as I've closed more than 800 tickets from users over the years. People coming from all kinds of backgrounds. This gave me an understanding of how people find the app, what they expect of it, etc.

In short, please revert the cask name, for everyone's benefit. Every other change in this PR is an improvement 👍

SMillerDev:
> I can't chose the name and description of my own app I worked 2 years on my weekends for free to help other people?

Not in someone else's project that they've worked for free on for well over 10 years on every day of the week.

I'm fine with you expressing your wishes for the naming of the cask and I even agree with them, but Homebrew has no obligation to you, or the project you represent and can pick their own rules. Regardless of how much time you spend on your app.

carlocab:
@lwouis I'm sorry that you feel disresepected, but @miccal was just following documented guidelines. The relevant guideline is even formalised in code which you can find in this repository at `developer/bin/generate_cask_token`:

```
❯ developer/bin/generate_cask_token AltTab.app
Proposed token:               alttab
Proposed file name:           alttab.rb
Cask Header Line:             cask 'alttab' do
```

That said, given that we have poor support for renaming casks, I also agree that renaming existing casks just to satisfy the cask token reference is almost always a terrible idea (with the possible exception of casks that have only very recently been added). We should probably change our guidelines regarding cask tokens here.

In the meantime, I'm happy to approve a PR that reverts the renaming of the cask.

lwouis:
> That said, given that we have poor support for renaming casks, I also agree that renaming existing casks just to satisfy the cask token reference is almost always a terrible idea (with the possible exception of casks that have only very recently been added). We should probably change our guidelines regarding cask tokens here.
>
> In the meantime, I'm happy to approve a PR that reverts the renaming of the cask.

I agree. Thank you!

I'll share a PR in a minute

ryenus:
Hi, team, thank you very much for the quick turnaround and have the wrongdoing corrected! Much appreciated! 👍 

About the token reference guidelines, I actually think it's a good thing, because naming is hard, really hard. When people are adding new casks and if they struggle with finding a proper name, having the token reference as a guideline, or even a tool to provide some naming recommendations is something that it's really helpful.

With regard to "obligation", I'd rather not think that way. There's a good old Chinese saying, "水能载舟, 亦能覆舟", which means, "Water can uphold boats, but also sink them". The relationship between the Homebrew core team and the community members is just like that. They simply cannot live without each other. As a user I really appreciate both Homebrew itself, as well as all the abundant software and apps that come with it.

Nevertheless, I believe having the experience like such is also a good thing, through it, people get wiser and the community and more mature, and more importantly, we have a happy ending :-)

Thank you!

vitorgalvao:
> About the token reference guidelines, I actually think it's a good thing, because naming is hard, really hard. When people are adding new casks and if they struggle with finding a proper name, having the token reference as a guideline, or even a tool to provide some naming recommendations is something that it's really helpful.

That’s not why it exists. The token reference is there to avoid duplicates, which used to be a major problem. @miccal wasn't blindly following arbitrary rules, he was following a proven useful rule with [a clear, documented purpose](https://docs.brew.sh/Cask-Cookbook#purpose).

The points [you mentioned above](https://github.com/Homebrew/homebrew-cask/pull/114668#issuecomment-974763780) are not relevant. I appreciate you trying to contribute to the discussion and I wholeheartedly believe your intentions are good and honest, but *context* matters, as does an understanding of why things are the way they are. It’s the principle of [Chesterton’s fence](https://wiki.lesswrong.com/wiki/Chesterton%27s_Fence): to criticise and change a rule, you should first understand it.

However you look at it, this situation is unfortunate. We either break existing user installations or future user expectations and have yet another hole for duplicates. There’s no clear best path.

[Having said that](http://youtu.be/ENHVQ2gslp8), I’ll be the first to agree the token reference is stricter than it perhaps need be today, and I do think a tad of pragmatism was a good choice here. But with every one of these we make, we weaken the foundation. You shouldn’t see this a victory, because it isn’t one neither for users nor maintainers. It is instead an increase in the pressing need for [a feature we sorely miss in casks](https://wiki.lesswrong.com/wiki/Chesterton%27s_Fence). PRs are welcome.

carlocab:
> I wholeheartedly believe your intentions are good and honest

This is important. The term "wrongdoing" carries with it connotations of malice that I find disrespectfully uncharitable.

PR #114916, opened 2021-11-25
eclipse-*: delete `:R` suffix in version string

FnControlOption:

#95207


bevanjkay:
I restarted CI a 4 or 5 times yesterday and it seems like at least one audit will run into a network error on one of the casks (can be different ones). 

I have seen most go green (except maybe *-javascript), so they may all be functional it's just temporary timeouts. 

May need to split the PR in half?

PR #114971, opened 2021-11-27
Update netron to 5.3.9

lutzroeder:


bevanjkay:
Continuing in #114986 as CI doesn't seem to want to run checks properly on this PR.

PR #115047, opened 2021-11-28
platypus: fix livecheck

neutric:

### DIFF
--- a/Casks/platypus.rb
+++ b/Casks/platypus.rb
@@ -16,8 +16,8 @@ cask "platypus" do
   app "Platypus.app"
 
   zap trash: [
-    "~/Library/Application Support/Platypus",
     "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/org.sveinbjorn.platypus.sfl2",
+    "~/Library/Application Support/Platypus",
     "~/Library/Caches/org.sveinbjorn.Platypus",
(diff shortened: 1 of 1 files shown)


Casks/platypus.rb
```diff
@@ -8,17 +8,16 @@
...
   end
 
   auto_updates true
 
   app "Platypus.app"
 
   zap trash: [
-    "~/Library/Application Support/Platypus",
     "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/org.sveinbjorn.platypus.sfl2",
+    "~/Library/Application Support/Platypus",
```
neutric, line 20: @miccal `sort`, which I had introduced to `createzap` recently, orders items like this by default. As there is no right or wrong with this stuff: which rules are you using so that we can minimise the need for corrections?

miccal: I am just using one built-in to the Shortcuts.app on Monterey, and sometimes I use the one in Vim.

neutric: Ok, let me phrase it differently: you prefer sorting alphabetically ignoring the case in mixed-case scenarios, correct?

miccal: I do not have a preference, really -- depends on my mood and which way the wind is blowing that day :)

neutric: Got it!

PR #115208, opened 2021-12-02
Update Semulov from 2.4.1 to 2.6

premist:

---

Seems like author of this application stopped updating download page on their personal website, however releasing a stable version on the official [GitHub Releases](https://github.com/kainjow/Semulov/releases).

The Sparkle update feed may not be actively maintained anymore — it still says 2.4.1 as a latest version.

 I changed the website URL to point to GitHub (which is being updated) and changed livecheck strategy to `:github_latest`.

### DIFF
--- a/Casks/semulov.rb
+++ b/Casks/semulov.rb
@@ -1,11 +1,11 @@
 cask "semulov" do
-  version "2.4.1"
-  sha256 :no_check
+  version "2.6.0"
+  sha256 "fc25b015251562df79c89143b2fb123fc793d780e3f1eb9990d6d5723c870dc6"
 
-  url "https://kainjow.com/downloads/Semulov.zip"
+  url "https://github.com/kainjow/Semulov/releases/download/v2.6/Semulov.zip"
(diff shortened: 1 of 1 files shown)


Casks/semulov.rb
```diff
@@ -1,15 +1,15 @@
...
   name "Semulov"
   desc "Access mounted and unmounted volumes from the menubar"
-  homepage "https://kainjow.com/"
+  homepage "https://github.com/kainjow/Semulov"
 
   livecheck do
-    url "https://kainjow.com/updates/semulov.xml"
-    strategy :sparkle
+    url :url
+    strategy :github_latest
```
carlocab, line 12: Do we need to specify this? Does `brew livecheck --debug semulov` produce wrong output without `:github_latest`?

We usually try to avoid `:github_latest` since it can get rate-limited.

premist: Thanks for checking, I just ran livecheck and it seems to detect version strings fine. 

Removed `livecheck` block.

image

PR #115673, opened 2021-12-10
Update opencore-configurator from 2.54.0.0 to 2.54.1.0

scorpig:
Update opencore-configurator from 2.54.0.0 to 2.54.1.0


miccal:
Thank you @scorpig -- we have had issues with this Cask in the past when adding the version to the `url`.

PR #115739, opened 2021-12-11
Oxygen XML Editor: Add URL for Apple Silicon build

adunning:

### DIFF
--- a/Casks/oxygen-xml-editor.rb
+++ b/Casks/oxygen-xml-editor.rb
@@ -1,9 +1,12 @@
 cask "oxygen-xml-editor" do
+  arch = Hardware::CPU.intel? ? "" : "-aarch64"
+
   version :latest
   sha256 :no_check
 
-  url "https://www.oxygenxml.com/InstData/Editor/MacOSX/VM/oxygen.dmg"
+  url "https://www.oxygenxml.com/InstData/Editor/MacOSX/VM/oxygen-openjdk#{arch}.dmg"
   name "oXygen XML Editor"
+  desc "Tools for XML editing, including Oxygen XML Developer and Author"
   homepage "https://www.oxygenxml.com/xml_editor.html"
 
   suite "Oxygen XML Editor"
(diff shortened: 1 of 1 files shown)


joewiz:
@adunning This PR introduced two changes not described in the PR description:

1. It deleted the `version :latest` and `sha256 :nocheck` directives in favor of a named version and hash values. The reason the previous directives were in place is that the oXygen folks often issue new builds without incrementing the version number. Thus, homebrew has no way to associate a version number with a particular hash. As a result, as soon as the oXygen folks issue a new build, homebrew will not recognize the binary at the original location. See @vitorgalvao's original advice that led to the current setting at https://github.com/Homebrew/homebrew-cask/pull/28632#discussion_r94833062.

2. It switched from using the default oXygen XML Editor build, which includes Java SE 8u202, to an alternate option which includes OpenJDK 17—presumably because, as [the oXygen download page](https://www.oxygenxml.com/xml_editor/download_oxygenxml_editor.html) states, the Apple Silicon-specific build needs to use OpenJDK 17. But this switches the default for users of Intel Macs.

Would it be possible to submit a new PR that is limited to facilitation the installation of the Apple Silicon build, without introducing these other changes?

vitorgalvao:
> The reason the previous directives were in place is that the oXygen folks often issue new builds without incrementing the version number.

The reason was the URL was unversioned, thus all changes happened in place. This PR changed it to a versioned URL so it will work unless they regularly do patch releases on the same URL.

miccal:
We can switch back to using `version :latest` and `sha256 :nocheck` using the following links for Intel:
```
https://www.oxygenxml.com/InstData/Editor/MacOSX/VM/oxygen.dmg
https://www.oxygenxml.com/InstData/Editor/MacOSX/VM/oxygen-openjdk.dmg
```
and for Arm:
```
https://www.oxygenxml.com/InstData/Editor/MacOSX/VM/oxygen-openjdk-aarch64.dmg
```
But which version should we use for Intel?

adunning:
It's bizarre that versioned URLs are not stable. I've submitted an unversioned PR in #115845.

neutric:
To add to the back and forth: since there are version number and **build id** published on the downloads page, wouldn't it make sense to carry both along even if `sha256 :no_check` is used in order to signal to users when an update is available?

PR #115747, opened 2021-12-12
oxygen-xml-editor: add livecheck

miccal:


joewiz:
@miccal I agree that switching from appcast to livecheck makes sense for oXygen. But looking at the diffs for this PR, I wonder whether this will pick up changes when the oXygen developers make new releases? Often, between major and minor releases, the oXygen developers issue new builds that include minor fixes but are not included in the version number. Your livecheck regex in https://github.com/Homebrew/homebrew-cask/pull/115747/files#diff-4f97dddc2a34a38d590f8d41e49f62c352b0a9a1f84f5a902a8dc34a085e7d41R19 will trigger a new download if the string `Oxygen XML Editor version 24.0` at https://www.oxygenxml.com/rssBuildID.xml changes to 24.1 or 25.0, but not if the oXygen folks issue a new build. Perhaps the livecheck strategy should be altered to look for any change to the RSS file, since this feed only changes when there is a new build?

Cc: @raducoravu from the oXygen team. Radu, is there a good way that an app distribution system like Homebrew can find out when there's a new version of oXygen available, so that it can fetch the new version and ensure the new version is distributed to users?

raducoravu:
@joewiz Oxygen's own version checking mechanism relies on a couple of XML documents we keep on our web site:

https://www.oxygenxml.com/allVersions.xml

and for each version there is a reference to another XML file which has details about each latest minor version or minor bug fix release in that version.

joewiz:
@raducoravu Thank you for this info! It's very helpful!

PR #115764, opened 2021-12-12
punto-switcher: remove

neutric:
```
==> Analytics
install: 0 (30 days), 0 (90 days), 2 (365 days)
```
Was added 25 Mar 2014.


bevanjkay:
@neutric I merged the first one and now the rest are showing as conflicts. I'll send rebase commands but hopefully it doesn't keep happening. 😞

PR #115844, opened 2021-12-14
Update my-image-garden from 3.6.5 to 3.6.6,04

rgoldberg:
Upgraded `my-image-garden` to version `3.6.6,04`.

Changed version to use comma to handle variable in url (was `02`, is now `04`, and might change in future) so we can easily use `brew bump-cask-pr` in the future.

Added `desc` stanza.

### DIFF
--- a/Casks/my-image-garden.rb
+++ b/Casks/my-image-garden.rb
@@ -1,13 +1,14 @@
 cask "my-image-garden" do
-  version "3.6.5"
-  sha256 "d40260bb5786738a40560b1a6834c4b284c7c9986ae7ce2d96c90e820d7f8f73"
+  version "3.6.6,04"
+  sha256 "1b83b4695cd8d84c7807a06a5ba4c7e46cc5cfd73bdaf606a327e7341f573bb2"
 
-  url "https://gdlp01.c-wss.com/gds/2/0200006062/02/mmig-mac-#{version.dots_to_underscores}-ea11.dmg",
+  url "https://gdlp01.c-wss.com/gds/2/0200006062/#{version.csv.second}/mmig-mac-#{version.csv.first.dots_to_underscores}-ea11.dmg",
       verified: "c-wss.com/"
   name "Canon My Image Garden"
+  desc "Photo editing and printing tool"
   homepage "https://support-asia.canon-asia.com/?personal"
 
-  pkg "My Image Garden V#{version.no_dots}.pkg"
+  pkg "My Image Garden V#{version.csv.first.no_dots}.pkg"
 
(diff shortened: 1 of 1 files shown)


miccal:
@bevanjay the only `livecheck` i can find is using the `url`
```
https://pdisp01.c-wss.com/gdl/WWUFORedirectTarget.do?id=MDIwMDAwNjA2MjA0&cmp=ABR&lang=EN
```
via:
```
|-> curl -sLI "https://pdisp01.c-wss.com/gdl/WWUFORedirectTarget.do?id=MDIwMDAwNjA2MjA0&cmp=ABR&lang=EN"
HTTP/2 302 
date: Tue, 14 Dec 2021 11:39:15 GMT
content-type: text/html; charset=UTF-8
location: https://gdlp01.c-wss.com/gds/2/0200006062/04/mmig-mac-3_6_6-ea11.dmg
server: Apache
x-frame-options: SAMEORIGIN
cache-control: no-cache
pragma: no-cache
set-cookie: JSESSIONID=Z7q4vVK6QsQ0cIBuUihKdB1fjuFFZ4IwEuy156nKSz5ARo3EUcON!1543167859; path=/; HttpOnly

HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Type: application/octet-stream
ETag: "2f79a504bd832ec43be7f784bd456ac6:1638941478.419477"
Last-Modified: Wed, 08 Dec 2021 05:31:24 GMT
Server: AkamaiNetStorage
Content-Length: 314857795
Date: Tue, 14 Dec 2021 11:39:15 GMT
Connection: keep-alive
```
Can you have a look and see of you may be able to find a more suitable one?

bevanjkay:
@miccal What you have found would also be the best that I can do, in the future we may be able to query the driver search page, however this won't be possible until POST requests are available within livecheck.

rgoldberg:
What is the livecheck?

Why is it causing a problem now yet not before?

Why can't it just use the url from which the install is downloaded?

bevanjkay:
@rgoldberg The `livecheck` is used to check that the version in the cask is the latest version that is available upstream. It ensures that versions that are merged in are the correct versions, and you can use the command `brew livecheck <token>` to check if there are newer versions available.
[See the docs for more info.](https://docs.brew.sh/Cask-Cookbook#stanza-livecheck)

bevanjkay:
@miccal We can give this a go, but I'm not sure if the redirect ID changes when the package updates.

rgoldberg:
I assume it's way too late, but wouldn't current-release make more sense than livecheck?

The latter sounds like a check if the install can be downloaded.

PR #116000, opened 2021-12-17
Added acrobat font pack

henryhchchc:

### DIFF
--- a/Casks/adobe-acrobat-font-pack.rb
+++ b/Casks/adobe-acrobat-font-pack.rb
@@ -1,0 +1,19 @@
+cask "adobe-acrobat-font-pack" do
+  version "21.007.20091"
+  sha256 "3fb67cc3140de22bde3beb9483fce484992c810ff4cf63c7f0c943cf861a3dd8"
+
+  url "https://ardownload3.adobe.com/pub/adobe/reader/mac/AcrobatDC/misc/FontPack#{version.no_dots}_XtdAlf_Lang_DC.dmg"
+  name "Adobe Acrobat Extended Language Font Pack"
+  desc "Extended Language Font Pack For Adobe Acrobat"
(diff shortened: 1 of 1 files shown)


henryhchchc:
Hi maintainers, I found that the pkg installation will work only if `adobe-acrobat-pro` or `adobe-acrobat-reader` is installed. Is it possible to make it depends on **either of the two packages**?

bevanjkay:
@henryhchchc Thank you for your contribution. A couple of things to note here. This cask should be within the [`homebrew-cask-fonts`](https://github.com/Homebrew/homebrew-cask-fonts/) repository, and regarding the `depends_on`, the challenge there is that it will install the depended upon casks if this cask is installed - perhaps a caveat could be used instead.

henryhchchc:
Thanks for the suggestions and I have the following concerns.
> This cask should be within the [`homebrew-cask-fonts`](https://github.com/Homebrew/homebrew-cask-fonts/) repository,

The casks in `homebrew-cask-fonts` are purely fonts (i.e., they install nothing but fonts). However this package will install some text processing plugins to Acrobat in addition to fonts. So I suggest keeping it at the `homebrew-casks` repository.

> and regarding the `depends_on`, the challenge there is that it will install the depended upon casks if this cask is installed - perhaps a caveat could be used instead.

I do agree that using a caveat is good, but not including the dependencies will fall the Homebrew CI (and this is the reason why the CI is failing now). I am wondering if there a workaround for it.

miccal:
@henryhchchc we cannot add a conditional `depends_on cask` for `adobe-acrobat-pro` or `adobe-acrobat-reader`, so a `caveat` would be best here.

miccal:
Looking at this further, it appears that you also need to install a specific version depending on the version of Adobe Reader one has installed.

Thank you @henryhchchc, but I am afraid that this is not suitable for inclusion as a Cask.

henryhchchc:
Thank you for the information.

PR #116137, opened 2021-12-20
Fix URL for autofirma

adruiza:

The URL structure for this cask has changed and is now versioned. The previous URL, pointing to just "currentversion", now gives a 404 status code. I updated the URL according to the new download link provided at https://firmaelectronica.gob.es/Home/Descargas.html, which is the official place provided by the spanish government to download this application.

I also added the SHA256 and a description for the cask.

### DIFF
--- a/Casks/autofirma.rb
+++ b/Casks/autofirma.rb
@@ -2,9 +2,10 @@ cask "autofirma" do
   version "1.6.5"
   sha256 "4564893f65a1566ea82c131a63bb56b44dbd2bcc218b5ba501ebb6dc22634c79"
 
-  url "https://estaticos.redsara.es/comunes/autofirma/#{major}/#{minor}/#{patch}/AutoFirma_Mac.zip",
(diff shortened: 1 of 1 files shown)


adruiza:
I also added a livecheck stanza, which was missing in the current Cask.

adruiza:
The tests do not complete due to an interactive window that is shown during the installation of the pkg.

The installation does not finish until you click 'Ok' in this window, making brew wait forever. This behavior was the same when installing the Cask before the changes. I'll check the Cookbook just in case there is an option to bypass this window.

adruiza:
I couldn't find an easy way to fix the installation stalling in the tests due to the interactive dialog I mentioned in the previous comment. The pkg contains a post-install script that requires the user to click an "Ok" button in order to continue.

I believe this should be merged anyway (the current Cask doesn't work due to a 404 status code and it has the same issue with the interactive dialog). I'll try to contact the developers of the app to find if there is some way we can make this installation fully unattended, and I'll open a new PR.

miccal:
This Cask has failed the install checks in the past.

miccal:
Thank you @adruiza.

PR #116217, opened 2021-12-21
Update pgadmin4 from 6.2 to 6.3

egilkh:


egilkh:
I'm too late and I'm sorry about making the PR. My brew update did not catch the one that is out already.

PR #116247, opened 2021-12-22
Workaround for GitKraken Big Sur and later Performance Issues

fedotov-as:

### DIFF
--- a/Casks/gitkraken.rb
+++ b/Casks/gitkraken.rb
@@ -18,6 +18,20 @@ cask "gitkraken" do
 
   app "GitKraken.app"
 
+  # Workaround for GitKraken Big Sur and later Performance Issues
+  # See https://www.gitkraken.com/blog/workaround-gitkraken-big-sur-issues
+  if MacOS.version >= :big_sur
+    postflight do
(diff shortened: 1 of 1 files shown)


bevanjkay:
@fedotov-as 

We don’t delete the quarantine, *we add it* and go to some lengths to do so. A download via `curl` (which Homebrew uses) does not add the quarantine.

To install something without quarantining, always at your own risk, use the `--no-quarantine` flag.

---

We may consider a caveat instead.

fedotov-as:
@bevanjkay

Official instruction:

> Consider running the following command in your terminal: 
> `codesign --remove-signature /Applications/GitKraken.app/Contents/Frameworks/GitKraken\ Helper\ \(Renderer\).app`

But if you run codesign while the application is in quarantine, then it will break the application.

Is it possible to solve this problem?

bevanjkay:
@fedotov-as We may consider adding a [`caveat`](https://docs.brew.sh/Cask-Cookbook#stanza-caveats). We should not programmatically be removing quarantine, this is something that the user should do on their own volition.

PR #116404, opened 2021-12-24
paraview 5.10.0

miccal:


gerlero:
Thanks. I was myself just working on a manual PR for this update.

By the way, Paraview 5.10.0 now offers binaries for ARM, so it'd be great to support those. I was doing something like this:

```ruby
  arch = Hardware::CPU.intel? ? "x86_64" : "arm64"
  min_macos_version = Hardware::CPU.intel? ? "10.13" : "11.0"
  
  version "5.10.0"
  
  if Hardware::CPU.intel?
    sha256 "346833f38ef82d9313fcb203396b901da41e968a8d7078e2f128a345d2d6bafc"
  else
    sha256 "d2c4d45e1c173d50e2400e76bacc17ae535c4d9e6532bd80e61bc001a88843a7"
  end

  url "https://www.paraview.org/paraview-downloads/download.php?submit=Download&version=v#{version.major_minor}&type=binary&os=macOS&downloadFile=ParaView-#{version}-MPI-OSX#{min_macos_version}-Python3.9-#{arch}.dmg",
      user_agent: :fake
```

miccal:
@gerlero I notices that too, but it seems that the Arm version has a problem:

Screen Shot 2021-12-24 at 09 39 13

gerlero:
That's right.. the dmg itself seems to be broken due to a failing CRC check. I guess ARM support will have to wait for a fixed dmg.

PR #116408, opened 2021-12-24
far2l 2.3.2

yurikoles:

---
I was unable to write a regex for a new tag scheme.


yurikoles:
I think that we should keep using double versions with comma, since upstream is known to use random tag format.

miccal:
We will need to keep an eye on this for the future, as the `livecheck` changes I have made may not work if they revert to their old versioning scheme.

PR #116409, opened 2021-12-24
change mqttx app to universal app

jianyun8023:
Modify the mqttx application to support universal version.


miccal:
Thank you @jianyun8023.

PR #116558, opened 2021-12-28
vlcstreamer: add desc

alexreg:
This seems to be the "official" description on the website.

### DIFF
--- a/Casks/vlcstreamer.rb
+++ b/Casks/vlcstreamer.rb
@@ -9,8 +9,7 @@ cask "vlcstreamer" do
 
   livecheck do
     url "https://hobbyistsoftware.com/vlcstreamer-versions"
-    strategy :page_match
-    regex(%r{href=.*?/VLCStreamer_(\d+(?:\.\d+)*)\.dmg}i)
+    regex(%r{href=.*?/VLCStreamer_(\d+(?:\.\d+)+)\.dmg}i)
   end
 
   app "VLCStreamer.app"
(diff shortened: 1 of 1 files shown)


Casks/vlcstreamer.rb
```diff
@@ -4,6 +4,7 @@
 
   url "https://hobbyistsoftware.com/Downloads/VLCStreamer/Versions/VLCStreamer_#{version}.dmg"
   name "VLC Streamer"
+  desc "Streaming by the power of VLC"
```
neutric, line 7: Not familiar with the app, but from cursory glance it does not have functionality by itself but enables the Android/iOS app "VLC Streamer". I'd thus suggest `desc "VLC Streamer companion app"`.

alexreg: Your right, the official description is a bit vague and more of a slogan. I've updated it to something more descriptive. (This app actually does the streaming to the mobile devices.)

neutric: I was imprecise, sorry. What I meant: does this "helper app", as the homepage calls it, provide functionality if you do not use the VLC Streamer mobile app?

alexreg: Not as far as I'm aware.

neutric: Then I'd personally go with describing it a companion app as your desc sounds like it exposes a stream url that one can access with a media player of ones choosing. Feel free to leave it as is, though. Maintainers have the last word anyway. :)

alexreg: Yeah, I do see your point... will see what one of them says, I suppose. Obviously happy to change it if desired.

PR #116579, opened 2021-12-29
Update MacLoggerDX livecheck to use possesive quantifiers

macauley:

### DIFF
--- a/Casks/macloggerdx.rb
+++ b/Casks/macloggerdx.rb
@@ -9,7 +9,7 @@ cask "macloggerdx" do
 
   livecheck do
     url :homepage
-    regex(/Download:\s*v?(\d+(?:\.\d+)+)[^a-z]+/i)
+    regex(/Download:\s*v?(\d+(?:\.\d++)++)[^b]+/i)
   end
 
   conflicts_with cask: "homebrew/cask-versions/macloggerdx-beta"
--- a/Casks/macloggerdx.rb
+++ b/Casks/macloggerdx.rb
@@ -9,7 +9,7 @@ cask "macloggerdx" do
 
   livecheck do
     url :homepage
-    regex(/Download:\s*v?(\d+(?:\.\d++)++)[^b]+/i)
+    regex(/Download:\s*v?(\d+(?:\.\d+)+)[\s<]/im)
   end
 
   conflicts_with cask: "homebrew/cask-versions/macloggerdx-beta"


miccal:
Ping @samford.

Casks/macloggerdx.rb
```diff
@@ -9,7 +9,7 @@
 
   livecheck do
     url :homepage
-    regex(/Download:\s*v?(\d+(?:\.\d+)+)[^a-z]+/i)
+    regex(/Download:\s*v?(\d+(?:\.\d++)++)[^b]+/i)
```
samford, line 12:
```suggestion
    regex(/Download:\s*v?(\d+(?:\.\d+)+)[\s<]+?/im)
```

In my view, both `[^a-z]+` and `[^b]+` feel a bit too loose in this context. The goal is to only match a version like `6.36` while omitting unstable versions like `6.37b14`, so I think it may make more sense to use `[\s<]+?` to anchor the end and restrict matching to versions like `6.36`.

Since this is matching loosely formatted text that can potentially span more than one line, this suggestion also uses the `m` flag.

samford (changes requested):
Throughout the Homebrew repositories, we're currently only using possessive quantifiers in one regex. The `v?(\d+(?:\.\d+)+)` and `v?(\d+(?:\.\d+)*)` patterns are used in thousands of `livecheck` block regexes in formulae/casks (as well as in some strategies), so this would be a notable change.

From some simple testing in isolation (i.e., comparing the performance of the regex with/without possessive quantifiers over many iterations when `scan`ning a string of the `homepage` HTML), I don't see any notable performance improvement from using possessive quantifiers in this regex. Assuming this is representative and can be extrapolated to similar regexes, I would prefer to simply maintain consistency with the existing regexes throughout Homebrew and avoid possessive quantifiers for the time being.

That said, do you have any compelling evidence to support using possessive quantifiers in this particular context (e.g., benchmarks that show an improvement)?

Casks/macloggerdx.rb
macauley: The change to using possessive quantifiers wasn't done for performance, but like you guessed was to avoid the unstable/beta version strings like '6.37b14' from matching the original regex mistakenly as '6.3'. 

> The goal is to only match a version like 6.36 while omitting unstable versions like 6.37b14, so I think it may make more sense to use [\s<] to anchor the end.

Agree, this approach is better, will make this change.

PR #116702, opened 2021-12-31
jellyfin-media-player: switch livecheck strategy

neutric:
Before: `jellyfin-media-player : 1.6.1 ==> 1.7.0-pre2`
After: `jellyfin-media-player : 1.6.1 ==> 1.6.1`


bevanjkay:
@neutric There's some stray changes in this one 😊

neutric:
> @neutric There's some stray changes in this one 😊

Indeed, sry. Let me fix that real quick.

PR #116747, opened 2022-01-01
Update plexamp from 3.5.0 to 3.9.0

disrupted:
Plex is now releasing ARM builds

### DIFF
--- a/Casks/plexamp.rb
+++ b/Casks/plexamp.rb
@@ -1,5 +1,6 @@
 cask "plexamp" do
   version "3.9.0"
+
   if Hardware::CPU.intel?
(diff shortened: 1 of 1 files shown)


boggydigital:
I was about to submit PR to do that. Quick question - perhaps we can use this opportunity to simplify cask? The only difference between downloads would be URL (-arm64 suffix) and sha256 - perhaps outer if could be removed and replaced with:

arch = Hardware::CPU.intel? ? "" : "-arm64"
sha256 = ... (similarly, two values based on arch)

and url "https://plexamp.plex.tv/plexamp.plex.tv/desktop/Plexamp-#{version}#{arch}.dmg",

I think the rest of the cask is the same for Intel vs Apple Silicon and we don't really need larger if. I could be wrong though.

PR #116879, opened 2022-01-03
Delete finereader.rb

joewiz:
The cask contained an outdated link to ABBYY FineReader Pro v12.1.13. Unfortunately, no public link to the current version, ABBYY FineReader PDF v15.1.0, is available.


neutric:
```
==> Analytics
install: 14 (30 days), 41 (90 days), 116 (365 days)
```

> Unfortunately, no public link to the current version, ABBYY FineReader PDF v15.1.0, is available.

That is no problem as long as the download URL is not both behind a login/registration form and from a host that differs from the homepage as per https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask.

joewiz:
@neutric Judging by the categories on that page, the current stable version (v15) belongs in the "Walled" category. The download URL is behind a login/registration form. Since Walled apps are not accepted, I believe this PR is still valid. If users still wish to access the legacy version (v12), then this cask would be a candidate for a PR to `cask-versions`.

neutric:
I wasn't clear about what you meant by "no public link". If you can't download it without handing the server a non-permanent token, then the rest is indeed moot.

joewiz:
@neutric Hmm... I dug into this a little more, and maybe this cask can be serviced!? 

While the ABBYY folks clearly go to some effort to obscure the download URL from casual users, it *is* public. The registration form at https://pdf.abbyy.com/finereader-pdf-for-mac/trial/ prompts users to complete a form to download a trial copy, which can then be unlocked after purchase (no new download required). The actual download link to the current version is in the HTML source of that page - https://downloads.abbyy.com/fr/fr_mac/current/ABBYY_FineReader_PDF.dmg?secure=5c11NlPrQmIZeGn6dK2PZA==. I assume that the `secure` parameter changes with each release. There's no release notes page advertising new versions. Even the app's own "Check for Updates" command constructs a highly specific query to the abbyy.com servers, e.g., https://go.abbyy.com/?Product=FineReaderMac&Edition=Retail&Serial=DELETED_BY_ME&Language=en&Version=15.1.0.429960&Part=1402.7&Target=CheckUpdate. So ABBYY make it very tricky to get the current download URL. (Perhaps it's because the app is so darned big - nearly 1 GB - that they want to limit downloads.)

Could we somehow engineer a [livecheck](https://docs.brew.sh/Cask-Cookbook#stanza-livecheck) directive to spot future changes to the `secure` parameter as published to the "trial" form page?

joewiz:
p.s. Checking the archive.org version of the trial form from April, the download link was:

https://downloads.abbyy.com/fr/fr_mac/current/ABBYY_FineReader_PDF.dmg?secure=5c11NlPrQmIZeGn6dK2PZA==

Even though the "Check for Updates" URL from my previous post shows new versions have been released since April, the `secure` link did not change.

So perhaps instead of deleting the cask we could just update the URL and its `secure` parameter? There's no guarantee it won't change with a new release... So the livecheck approach might be best? It's beyond my abilities though.

joewiz:
p.p.s. If we keep the cask, I believe the rules at https://docs.brew.sh/Cask-Cookbook#finding-the-simplified-name-of-the-vendors-distribution mean it should be renamed from `finereader` to `abbyy-finereader-pdf`. The "exact name of the Application bundle on disk" is "ABBYY FineReader PDF.app."

neutric:
Good detective work! Given the above information, I'd assume the secure parameter stays the same across at least minor versions, which means we can't checksum the resulting file and also that we don't need to involve the parameter in `livecheck`. I'd use the url that ends with the part number for `livecheck` purposes and anchor the regex to "Update Part Number:".

joewiz:
@neutric Great, I'll close this PR and proceed in that direction. Thanks for your feedback and guidance!

PR #116881, opened 2022-01-03
Add Creality Slicer v4.8.2-96

vahrennd:

### DIFF
--- a/Casks/creality-slicer.rb
+++ b/Casks/creality-slicer.rb
@@ -1,5 +1,5 @@
 cask "creality-slicer" do
-  version "4.8.2-96"
+  version "4.8.2,96"
   sha256 :no_check
 
   url "https://file2-cdn.creality.com/website/b049ecf6-2295-4fac-8d10-377dc67a494c"
@@ -9,9 +9,9 @@ cask "creality-slicer" do
 
   livecheck do
     url :homepage
-    regex(/Creality_Slicer-((?:\d+(?:\.\d+)*)-build-(?:\d+(?:\.\d+)*))-Darwin\.zip/i)
+    regex(/Creality[._-]Slicer[._-](\d+(?:\.\d+)+)[._-]build[._-](\d+)[._-]Darwin\.zip/i)
     strategy :page_match do |page, regex|
-      page.scan(regex).map { |match| match[0].sub("-build-", "-") }
+      page.scan(regex).map { |match| "#{match[1]},#{match[2]}" }
     end
   end
 
@@ -19,8 +19,8 @@ cask "creality-slicer" do
 
   zap trash: [
     "~/Library/Application Support/Creality Slicer",
-    "~/Library/Saved Application State/com.creality.crealityslicer.savedState",
-    "~/Library/Preferences/com.creality.crealityslicer.plist",
     "~/Library/Logs/Creality Slicer",
+    "~/Library/Preferences/com.creality.crealityslicer.plist",
+    "~/Library/Saved Application State/com.creality.crealityslicer.savedState",
   ]
 end
--- a/Casks/creality-slicer.rb
+++ b/Casks/creality-slicer.rb
@@ -11,7 +11,7 @@ cask "creality-slicer" do
     url :homepage
     regex(/Creality[._-]Slicer[._-](\d+(?:\.\d+)+)[._-]build[._-](\d+)[._-]Darwin\.zip/i)
     strategy :page_match do |page, regex|
-      page.scan(regex).map { |match| "#{match[1]},#{match[2]}" }
+      page.scan(regex).map { |match| "#{match[0]},#{match[1]}" }
     end
   end
 
--- a/Casks/creality-slicer.rb
+++ b/Casks/creality-slicer.rb
@@ -1,0 +1,19 @@
+cask "creality-slicer" do
+  version "4.8.2-96"
+  sha256 :no_check
+
+  url "https://file2-cdn.creality.com/website/b049ecf6-2295-4fac-8d10-377dc67a494c"
+  name "creality-slicer"
+  desc "Slicer for all Creality FDM 3D printers"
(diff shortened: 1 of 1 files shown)


Casks/creality-slicer.rb
```diff
@@ -0,0 +1,19 @@
+cask "creality-slicer" do
+  version "4.8.2-96"
+  sha256 :no_check
```
vahrennd, line 3: Just wanted to check if this was desirable here. `brew audit` said to use no sha check if there was no version in the download URL, but there is what looks like a unique ID that may change if the app version changes.

neutric:
```
%  curl -ILs https://file2-cdn.creality.com//website/b049ecf6-2295-4fac-8d10-377dc67a494c | grep -E "(Location|Disposition)"
Content-Disposition: attachment;filename="Creality_Slicer-4.8.2-build-96-Darwin.zip"
```
If the hex string is static across versions, then we need `sha256 :no_check`. If it changes with every new version, than we'd have to carry it along as part of `version` but would gain checksumming. archive.org has no history on that page, so I guess we'd have to wait for an update to ultimately decide.

Casks/creality-slicer.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  url "https://file2-cdn.creality.com/website/b049ecf6-2295-4fac-8d10-377dc67a494c"
+  name "creality-slicer"
```
neutric, line 6: `name "Creality Slicer"`

Casks/creality-slicer.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  livecheck do
+    url :homepage
+    regex(/Creality_Slicer-((?:\d+(?:\.\d+)*)-build-(?:\d+(?:\.\d+)*))-Darwin\.zip/i)
+    strategy :page_match do |page, regex|
+      page.scan(regex).map { |match| match[0].sub("-build-", "-") }
+    end
+  end
+
+  app "Creality Slicer.app"
```
neutric, line 18: Please try to add a [`zap` stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap). The linked documentation provides some recommended methods to create it.

Casks/creality-slicer.rb
```diff
@@ -0,0 +1,19 @@
...
+  desc "Slicer for all Creality FDM 3D printers"
+  homepage "https://www.creality.com/download/"
+
+  livecheck do
+    url :homepage
+    regex(/Creality_Slicer-((?:\d+(?:\.\d+)*)-build-(?:\d+(?:\.\d+)*))-Darwin\.zip/i)
+    strategy :page_match do |page, regex|
+      page.scan(regex).map { |match| match[0].sub("-build-", "-") }
+    end
+  end
```
neutric, line 16: Not sure much is won by stripping out "-build" from the version. The resulting version scheme looks different to what you find on `homepage` and thus might cause mental overhead without benefit that I can see. Wait for a maintainer's comment on that, though.

neutric (commented):
Thank you for your contribution to `homebrew-cask` @vahrennd! My review is in the code. Take it with a grain of salt as I am not a maintainer.

Casks/creality-slicer.rb
vahrennd: The reason I did so is that everywhere I can find references version 4.8.2-96 except the website. For example, in the app itself:

image

That being said, if the consensus is to just leave the `-build-` part in then I can go with that.

Casks/creality-slicer.rb
vahrennd: Maybe we can just leave it without the check for now and see what happens next time they release a new version.

Casks/creality-slicer.rb
neutric: In that case 4.8.2-96 is to be preferred.

Casks/creality-slicer.rb
```diff
@@ -9,18 +9,18 @@
 
   livecheck do
     url :homepage
-    regex(/Creality_Slicer-((?:\d+(?:\.\d+)*)-build-(?:\d+(?:\.\d+)*))-Darwin\.zip/i)
+    regex(/Creality[._-]Slicer[._-](\d+(?:\.\d+)+)[._-]build[._-](\d+)[._-]Darwin\.zip/i)
     strategy :page_match do |page, regex|
-      page.scan(regex).map { |match| match[0].sub("-build-", "-") }
+      page.scan(regex).map { |match| "#{match[1]},#{match[2]}" }
```
neutric: `page.scan(regex).map { |match| "#{match[0]},#{match[1]}" }`

miccal:
@vahrennd I am assuming that you need access to a Creality FDM 3D printer for this softeware to be useful?

If so, this needs to be submitted to the [`homebrew-cask-drivers` repository](https://github.com/Homebrew/homebrew-cask-drivers).

vahrennd:
You would need access to *a* 3D printer to make much use of it, sure. It technically has preset profiles for other brands of printers as well as the ability to define your own. [ultimaker-cura](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/ultimaker-cura.rb) is basically an equivalent piece of software available in homebrew-cask which is what made me think it belonged here.

Would it help if I updated the description to more accurately reflect that? I mostly just took that from their download page.

miccal:
> You would need access to _a_ 3D printer to make much use of it, sure. It technically has preset profiles for other brands of printers as well as the ability to define your own. [ultimaker-cura](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/ultimaker-cura.rb) is basically an equivalent piece of software available in homebrew-cask which is what made me think it belonged here.
> 
> Would it help if I updated the description to more accurately reflect that? I mostly just took that from their download page.

Thanks for the explanation -- I think actually it will be fine in this repository.

PR #117223, opened 2022-01-11
Update Firefox from 95.0.2 to 96.0

brianvanburken:

### DIFF
--- a/Casks/firefox.rb
+++ b/Casks/firefox.rb
@@ -1,108 +1,108 @@
 cask "firefox" do
-  version "95.0.2"
+  version "96.0"
 
   language "cs" do
-    sha256 "3581f1658a505ea2876f179bc31f820c16ebb5bf3c4c6bfe900f0b07f4f97216"
+    sha256 "14acd5e2d7831e4bfe4f0df69d37735a40aae0600bc51426b38c3e48badfddd4"
     "cs"
   end
   language "de" do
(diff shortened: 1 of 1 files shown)


brianvanburken:
It seems I'm a bit too early with this PR. The [changelog](https://www.mozilla.org/en-US/firefox/96.0/releasenotes/) hasn't been updated yet. Also, locally all steps succeed, but the CI `livecheck` fails.

bevanjkay:
@brianvanburken We will need to wait until the livecheck passes, you can check this locally with `brew livecheck firefox` or by including `--online` when you run `brew audit firefox --online` (this is what CI uses) - but in this instance it will download all the files in the update (which may be cached for you), so may not be very economical.

brianvanburken:
@bevanjkay Thank you for the information.  I updated the cask using `update_multilangual_casks` and ran the steps mentioned in the PR template. 

I guess, we will have to wait a bit before trying again with the CI step.

bevanjkay:
@brianvanburken Firefox is on my livecheck watchlist so if I notice that `96.0` goes live I'll restart CI.

default50:
I think it's live: https://hg.mozilla.org/releases/mozilla-release/rev/8fae7cc94cfb

😄

brianvanburken:
Yes, it is finally live. I checked, and the hashes are still correct. The PR is ready to be merged 🎉

PR #117250, opened 2022-01-11
Add Cask: CleanMyMac X Chinese

LeileiChui:
Cask cleanmymac-ch is cleanmymac's Chinese special edition.
And they conflict with each other.

### DIFF
--- a/Casks/cleanmymac-ch.rb
+++ b/Casks/cleanmymac-ch.rb
@@ -6,7 +6,7 @@ cask "cleanmymac-ch" do
       verified: "dl.devmate.com/com.macpaw.zh.CleanMyMac"
   name "CleanMyMac X Chinese"
   desc "Tool to remove unnecessary files and folders from disk. China Special Edition"
-  homepage "https://macpaw.com/cleanmymac"
+  homepage "https://www.mycleanmymac.com/"
 
   livecheck do
     url "https://updates.devmate.com/com.macpaw.zh.CleanMyMac#{version.major}.xml"
--- a/Casks/cleanmymac-ch.rb
+++ b/Casks/cleanmymac-ch.rb
@@ -1,0 +1,74 @@
+cask "cleanmymac-ch" do
+  version "4.10.0,41000.0.2112201815"
+  sha256 "c2c63fe19526706cff20d42262c9f737bd424508b913deb5fd3a0d0536b0579c"
+
+  url "https://dl.devmate.com/com.macpaw.zh.CleanMyMac#{version.major}/CleanMyMacXChinese.dmg",
+      verified: "dl.devmate.com/com.macpaw.zh.CleanMyMac"
+  name "CleanMyMac X Chinese"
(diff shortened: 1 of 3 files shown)


bevanjkay:
@LeileiChui Homebrew has a[ language stanza](https://docs.brew.sh/Cask-Cookbook#stanza-language) that should be able to be used here instead of duplicating the cask.

Casks/cleanmymac-ch.rb
```diff
@@ -0,0 +1,74 @@
+cask "cleanmymac-ch" do
+  version "4.10.0,41000.0.2112201815"
+  sha256 "c2c63fe19526706cff20d42262c9f737bd424508b913deb5fd3a0d0536b0579c"
+
+  url "https://dl.devmate.com/com.macpaw.zh.CleanMyMac#{version.major}/CleanMyMacXChinese.dmg",
+      verified: "dl.devmate.com/com.macpaw.zh.CleanMyMac"
+  name "CleanMyMac X Chinese"
+  desc "Tool to remove unnecessary files and folders from disk. China Special Edition"
+  homepage "https://macpaw.com/cleanmymac"
```
miccal, line 9: Is there a Chinese specific `homepage`? I cannot find the `url` you have provided from the current `homepage`.

miccal:
> @LeileiChui Homebrew has a[ language stanza](https://docs.brew.sh/Cask-Cookbook#stanza-language) that should be able to be used here instead of duplicating the cask.

I am not sure if is possible to use the `language` stanza here because it does not use standard language codes in a consistent manner.

Casks/cleanmymac-ch.rb
LeileiChui: > @LeileiChui Homebrew has a[ language stanza](https://docs.brew.sh/Cask-Cookbook#stanza-language) that should be able to be used here instead of duplicating the cask.

CleanMyMac X Chinese is a special version, which is not common with the license of CleanMyMac X. And their application id are different, which relates to zap operation when uninstalling. So I think it should be separated and not as a Chinese language version of CleanMyMac X

> Is there a Chinese specific `homepage`? I cannot find the `url` you have provided from the current `homepage`.

They don't have a Chinese specific homepage

miccal: We need some way to verify the `url` -- where then does it come from?

LeileiChui: > We need some way to verify the `url` -- where then does it come from?

苏州思杰马克丁软件有限公司 is the distributor of MacPaw in China, and CleanMyMac X Chinese is the company's distribution in China. The official website of CleanMyMac X Chinese is https://www.mycleanmymac.com/

miccal:
Thank you @LeileiChui.

miccal:
PS: I re-named this to `cleanmymac-zh` since the `zh` region code is used throughout.

LeileiChui:
OK

PR #117441, opened 2022-01-15
Added scroll, renamed hookshot to rectangle-pro, and updated middle to 1.6.6

Aaron-Rumpler:
Also includes miscellaneous improvements for `rectangle`, `rectangle-pro`, `multitouch`, `middle`, `hightop`, and `hyperkey`.

Livecheck for `hyperkey` is only returning the short version (i.e. `0.11`) instead of both the short version and version (i.e. `0.11,11`), whereas the other casks return both. Is this due to the version matching the short version's minor version number?


neutric:
Can you please at least separate the rename and the addition of a cask into their own pull requests? The remaining ones ideally also get their own pull request unless what you change is related somehow.

Aaron-Rumpler:
The relation is that they're all from the same developer. Should I still separate out the changes to `hightop`, `hyperkey`, `multitouch` and `rectangle` into separate PRs?

Aaron-Rumpler:
`middle`: #117446
`scroll`: #117447
`rectangle-pro`: #117448

neutric:
> The relation is that they're all from the same developer. Should I still separate out the changes to `hightop`, `hyperkey`, `multitouch` and `rectangle` into separate PRs?

That'd be nice. With related I mean something like https://github.com/Homebrew/homebrew-cask/pull/117256. Being from the same developer does not qualify in this case. In other cases, it would, e.g. if multiple casks needed a change of `url` scheme because they all are on the same domain and that domain changed.

PR #117442, opened 2022-01-15
Replace version.before|after_comma with version.csv

fedotov-as:


bevanjkay:
@SMillerDev I'll need to add more rubocops or adjust the existing one to check more than just the URL. But this will be good to merge. Probably better to just do a syntax only CI if we confirm that only the syntax has changed.

bevanjkay (approved):

PR #117446, opened 2022-01-15
Update middle from 1.6.2,65 to 1.6.6,69

Aaron-Rumpler:

### DIFF
--- a/Casks/middle.rb
+++ b/Casks/middle.rb
@@ -1,6 +1,6 @@
 cask "middle" do
-  version "1.6.2,65"
-  sha256 "28fa9dd31686c51b56cee4cecee84c86f5d6cb29bc3d60d0b481ce42d3bed6b4"
+  version "1.6.6,69"
+  sha256 "fe9d1dba5e8e5ac080d3794d453537450cd26e21ae64bcfc3de32eff7a276fac"
 
   url "https://middleclick.app/downloads/Middle#{version.csv.first}.dmg"
   name "Middle"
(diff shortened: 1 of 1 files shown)


Casks/middle.rb
```diff
@@ -20,10 +20,14 @@
   uninstall quit: "com.knollsoft.Middle"
 
   zap trash: [
+    "~/Library/Application Scripts/com.knollsoft.MiddleLauncher",
     "~/Library/Application Support/Middle",
     "~/Library/Caches/com.knollsoft.Middle",
     "~/Library/Containers/com.knollsoft.MiddleLauncher",
-    "~/Library/Cookies/com.knollsoft.Middle.binarycookies",
+    "~/Library/Cookies/com.knollsoft.Middle",
+    "~/Library/Cookies/com.knollsoft.Middle.*",
```
neutric, line 28: If you are using wildcards, why not condense the above two further down to `"~/Library/Cookies/com.knollsoft.Middle*",`? Same with the two right below.

Aaron-Rumpler: I did consider it, but this way matches only `com.knollsoft.Middle`, and not something of the form `com.knollsoft.Middle[^.]+`, not that such a bundle id is ever likely to exist.

Aaron-Rumpler: Considering that `~/Library/Cookies` only has `$BUNDLE_ID.binarycookies` files, and `~/Library/HTTPStorages` only has `$BUNDLE_ID` directories and `$BUNDLE_ID.binarycookies` files, I'll probably change it to just that instead of using wildcards.

neutric: > I did consider it, but this way matches only `com.knollsoft.Middle`, and not something of the form `com.knollsoft.Middle[^.]+`, not that such a bundle id is ever likely to exist.

Good point.

> Considering that `~/Library/Cookies` only has `$BUNDLE_ID.binarycookies` files, and `~/Library/HTTPStorages` only has `$BUNDLE_ID` directories and `$BUNDLE_ID.binarycookies` files, I'll probably change it to just that instead of using wildcards.

I'd do it that way. Using a wildcard to save one line doesn't seem like a good cost-benefit trade-off.

PR #117496, opened 2022-01-16
openshot-video-editor: add postflight

miccal:
See https://github.com/OpenShot/openshot-qt/pull/4586 and https://github.com/Homebrew/discussions/discussions/2793

### DIFF
--- a/Casks/openshot-video-editor.rb
+++ b/Casks/openshot-video-editor.rb
@@ -9,4 +9,8 @@ cask "openshot-video-editor" do
   homepage "https://openshot.org/"
 
   app "OpenShot Video Editor.app"
+
+  postflight do
+    set_permissions "#{appdir}/OpenShot Video Editor.app/Contents", "a+rx"
+  end
 end


miccal:
@hmijail are you able to confirm that this change fixes the issues you are seeing?

hmijail:
Sorry, doesn't work. I did `brew update`, which showed the cask being updated, and `brew install openshot-video-editor`. But the resulting app still fails in the same way as before.

neutric:
Same here, app crashes without displaying GUI.

miccal:
Thank you both.

I have now tested this myself and tried many ways to get this to work, but all were unsuccessful.

Are you both on Monterey?

neutric:
> Are you both on Monterey?

Yes, I am on `12.1` and @hmijail said in the discussion thread

> the installed app is broken in (at least) MacOS 12.0 and 12.1

so he most likely is, too.

neutric:
Daily releases are [here](https://www.openshot.org/download/) and the links seem to not expire for a loooooooong time as the asset count is at 5000+, should we consider using a pre-release in this case.

miccal:
https://github.com/Homebrew/homebrew-cask-versions/pull/13057

miccal:
@hmijail the daily build of this which works on Monetery has been added the the `homebrew-cask-versions` repository as `openshot-video-editor-daily`.

hmijail:
It works, thank you.
However, maybe the non-daily cask should be stopped from installing in Monterey? (no idea if it works elsewhere either)

miccal:
@hmijail we go by what upstream does -- until they officially state that the version they provide is not compatible with Monterey, we keep this as-is.

PR #117615, opened 2022-01-18
Update macupdater from 2.1.2.12780 to 2.1.2.12880

dalisoft:

### DIFF
--- a/Casks/macupdater.rb
+++ b/Casks/macupdater.rb
@@ -1,6 +1,6 @@
 cask "macupdater" do
-  version "2.1.2,12780"
-  sha256 "a6b4f0edc62c51bba437635c33638fa36fb74362ad85ef9c2c37259364d8ed66"
+  version "2.1.2,12880"
+  sha256 "8c11689131cee194de5c1b5d9f1cd1151f287fb21f9512040e12890702ba0a9b"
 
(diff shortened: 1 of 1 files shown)


neutric:
@miccal: I emailed the author and he told me to "fully expect our CFBundleVersion and sparkle:version never to be in-sync" as this is necessary to stage the rollout of new versions / control the update experience.

Unfortunately, releases with the same major.minor.patch are updated in place, which leads to warnings regarding invalid checksums and probably prompted this PR.

I just now replied and suggested using URL redirects instead of updating in place. Until then, `strategy :extract_plist` might be necessary.

neutric:
Update: The author doesn't want to change his procedure, so `sha256 :no_check` has to stay.

Reverting to checking the sparkle file in order to avoid a 13,7 MB download is probably fine. The latest version displayed there will be close behind the version as per `https://www.corecode.io/downloads/macupdater_latest.dmg`, any deviations are just in order to facilitate staged rollouts. Also, new installs wouldn't be affected by a reversion as they'd get the latest release of the same patch-level and existing installations don't upgrade via Homebrew anyway since `auto_updates true`.

If we stay on `strategy :extract_plist`, then the `url` for livecheck needs to be changed to `https://www.corecode.io/downloads/macupdater_latest.dmg`.

PR #117621, opened 2022-01-18
Add Surge XT version 1.0.0

baconpaul:
Surge XT is the next version of surge (already packagted as
surge-synthesizer). There are a variety of legitimate reasons to install
both bits of software simultaenously so we have split the version identity
and subsequently split the homebrew identities too.

The documentation for the synth is at https://surge-synthesizer.github.io/

Please note this version fails the 'git relevance' test because we
use a different repository for our binaries than our source. The source
repo https://github.com/surge-synthesizer/surge/issues has 279 forks
and 1.8k stars; and Surge 1.9 was downloaded more than 100k times from
our various distributions. I hope this technical split of source from
binary repo doesn't avoid inclusion of XT in HomeBrew.

Thank you as always for your consideration.

### DIFF
--- a/Casks/surge-xt-synthesizer.rb
+++ b/Casks/surge-xt-synthesizer.rb
@@ -18,10 +18,12 @@ cask "surge-xt-synthesizer" do
     "com.surge-synth-team.surge-xt-fx.component.pkg",
     "com.surge-synth-team.surge-xt-fx.app.pkg",
     "com.surge-synth-team.surge-xt.resources.pkg",
-  ]
+  ],
+            delete:
+                     [
+                       "/Applications/Surge XT Effects.app",
+                       "/Applications/Surge XT.app",
+                     ]
 
-  zap trash: [
-    "/Applications/Surge XT Effects.app",
-    "/Applications/Surge XT.app",
-  ]
+  zap trash: []
 end
--- a/Casks/surge-xt-synthesizer.rb
+++ b/Casks/surge-xt-synthesizer.rb
@@ -14,11 +14,9 @@ cask "surge-xt-synthesizer" do
     "com.surge-synth-team.surge-xt.app.pkg",
     "com.surge-synth-team.surge-xt.component.pkg",
     "com.surge-synth-team.surge-xt.vst3.pkg",
-
     "com.surge-synth-team.surge-xt-fx.app.pkg",
     "com.surge-synth-team.surge-xt-fx.component.pkg",
     "com.surge-synth-team.surge-xt-fx.vst3.pkg",
-
     "com.surge-synth-team.surge-xt.resources.pkg",
   ],
             delete:
--- a/Casks/surge-xt-synthesizer.rb
+++ b/Casks/surge-xt-synthesizer.rb
@@ -19,4 +19,6 @@ cask "surge-xt-synthesizer" do
     "com.surge-synth-team.surge-xt-fx.app.pkg",
     "com.surge-synth-team.surge-xt.resources.pkg",
   ]
+
+  zap trash: []
 end
--- a/Casks/surge-xt-synthesizer.rb
+++ b/Casks/surge-xt-synthesizer.rb
@@ -20,5 +20,8 @@ cask "surge-xt-synthesizer" do
     "com.surge-synth-team.surge-xt.resources.pkg",
   ]
 
-  zap trash: []
+  zap trash: [
+    "/Applications/Surge XT Effects.app",
+    "/Applications/Surge XT.app",
+  ]
 end
--- a/Casks/surge-xt-synthesizer.rb
+++ b/Casks/surge-xt-synthesizer.rb
@@ -11,12 +11,14 @@ cask "surge-xt-synthesizer" do
   pkg "surge-xt-macOS-#{version}.pkg"
 
   uninstall pkgutil: [
-    "com.surge-synth-team.surge-xt.vst3.pkg",
-    "com.surge-synth-team.surge-xt.component.pkg",
     "com.surge-synth-team.surge-xt.app.pkg",
-    "com.surge-synth-team.surge-xt-fx.vst3.pkg",
-    "com.surge-synth-team.surge-xt-fx.component.pkg",
+    "com.surge-synth-team.surge-xt.component.pkg",
+    "com.surge-synth-team.surge-xt.vst3.pkg",
+
     "com.surge-synth-team.surge-xt-fx.app.pkg",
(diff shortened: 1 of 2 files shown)


SMillerDev (commented):

baconpaul:
> Please add a [`zap` stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap) here. The [CreateZap](https://github.com/nrlquaker/homebrew-createzap) tool can help you with this.

Happy to. But surge doesn't install anything in user directories *except* for user created files (the only one is `~/Documents/Surge XT` which just contains user authored patches) so I think the zap stanza should be empty.  I've just pushed an update that does that.

Thanks!

baconpaul:
ahh sorry now i see the CI failure and our uninstaller didn't remove the applications. let me add those to the zap and try again

baconpaul:
OK great; thanks for the review and i now have it so the only CI step it fails is the relevance on the repo one, where if possible I would love an exception because, as i mentioned, it's a binary-only repo we use separately from the (very active) source repo.

And thanks for maintaining such a great system and set of documentation!

Casks/surge-xt-synthesizer.rb
```diff
@@ -0,0 +1,29 @@
...
+    "com.surge-synth-team.surge-xt-fx.app.pkg",
+    "com.surge-synth-team.surge-xt.resources.pkg",
+  ],
+            delete:
+                     [
+                       "/Applications/Surge XT Effects.app",
+                       "/Applications/Surge XT.app",
+                     ]
+
+  zap trash: []
```
neutric, line 28: Please use `# No zap stanza required` instead.

Casks/surge-xt-synthesizer.rb
```diff
@@ -0,0 +1,29 @@
...
+  pkg "surge-xt-macOS-#{version}.pkg"
+
+  uninstall pkgutil: [
+    "com.surge-synth-team.surge-xt.vst3.pkg",
+    "com.surge-synth-team.surge-xt.component.pkg",
+    "com.surge-synth-team.surge-xt.app.pkg",
+    "com.surge-synth-team.surge-xt-fx.vst3.pkg",
+    "com.surge-synth-team.surge-xt-fx.component.pkg",
+    "com.surge-synth-team.surge-xt-fx.app.pkg",
+    "com.surge-synth-team.surge-xt.resources.pkg",
```
neutric, line 20: Please sort those alphabetically unless dependencies force a certain order.

neutric (commented):
Thanks for your contribution to `homebrew-cask`, @baconpaul! My review is in the code. Take it with a grain of salt as I am not a maintainer. Don't worry about the notability check, the source repo serves as proof that adding this as a cask is most likely not going to be a fruitless exercise.

Casks/surge-xt-synthesizer.rb
baconpaul: Done

Casks/surge-xt-synthesizer.rb
baconpaul: Done. Also added spaces for the three product 'group' (xt, xt-fx and xt.resources)

miccal:
Thank you @baconpaul.

Casks/surge-xt.rb
```diff
@@ -11,19 +11,18 @@
...
-                       "/Applications/Surge XT Effects.app",
-                       "/Applications/Surge XT.app",
-                     ]
+            delete:  [
+              "/Applications/Surge XT Effects.app",
+              "/Applications/Surge XT.app",
+            ]
 
-  # no zap stanza required
+  zap rmdir: "~/Documents/Surge XT"
```
baconpaul, line 27: Hi!

I am pretty sure this is wrong

The zap document says "don't remove any user generated content" and the *only* thing in ~/Documents/Surge XT is user generated content. Can we please not merge with this in place? It will make the zap unusable.

It's kinda like if uninstalling gcc did an "rm -rf ~/dev"

baconpaul:
Hi @miccal 

I think the zap is wrong! 

From https://docs.brew.sh/Cask-Cookbook#stanza-zap you say

```
zap stanzas should not remove:

Files created by the user directly.
```

but your change does that.

Should I push a pull request which removes it?

Casks/surge-xt.rb
miccal: The way I have done it will remove that folder ONLY if it is empty.

miccal:
@baconpaul `rmdir` will remove the folder ONLY if it is empty.

baconpaul:
Surge will create it only if it is not empty, though!

I put in a PR to change the line to not do this; I really think it's not the kinda thing your zap wanted. But if you want to reject that PR I understand. It really makes me uncomfortable though. Buckets of intellectual property in that directory and we very strictly don't touch it destructively without a user prompt in the app.

baconpaul:
(Also thank you for the merge! :) )

neutric:
> Buckets of intellectual property in that directory and we very strictly don't touch it destructively without a user prompt in the app.

Meaning what? Can a user delete a project which would then clear the corresponding files in said directory? If so, it is conceivable that you end up with an empty `~/Documents/Surge XT`, no? Only then would it be removed by uninstall and only if the user explicitly calls `--zap` during uninstall, as per https://docs.brew.sh/Cask-Cookbook#summary-of-keys.

baconpaul:
Oh just meaning "that's where the user saves all their patches and waves. We've made it so none of our installers ever touch that directory and we only create it if needed on user action".

Like I said I'm fine either way (and @miccal since merged a no zap change - thanks!) but when I saw the diff I had a screaming heebie-jeebies because it is exactly the one directory our users care about (other than their DAW projects).

(By the way, wouldn't 'trash' be safer than 'rmdir' in this case? Just learning though)

And again, appreciate the diligence here. It's a load of work to make this thing work and you folks do a great job.

neutric:
> (By the way, wouldn't 'trash' be safer than 'rmdir' in this case? Just learning though)

`rmdir` only removes empty directories. `trash` moves directories with whatever content to the trash. So, no.

baconpaul:
gotcha thanks.

Well if the directory exists it's not empty, so it's moot. If its important to your infra happy to put it back! Just let me know.

neutric:
> Well if the directory exists it's not empty, so it's moot.

The user might take manual action, though, and end up with an empty folder. At the current state of the cask, this dir then remains after an uninstall. This being said, I am fine with what was merged by the maintainers.

PR #117823, opened 2022-01-21
Remove Docker Desktop conflict for Rancher Desktop

sanyer:

### DIFF
--- a/Casks/rancher.rb
+++ b/Casks/rancher.rb
@@ -16,10 +16,7 @@ cask "rancher" do
   homepage "https://rancherdesktop.io/"
 
   auto_updates true
-  conflicts_with cask:    %w[
-    docker
-  ],
-                 formula: %w[
+  conflicts_with formula: %w[
                    docker
                    helm
                    kubernetes-cli
--- a/Casks/rancher.rb
+++ b/Casks/rancher.rb
@@ -23,11 +23,6 @@ cask "rancher" do
   ]
 
   app "Rancher Desktop.app"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/docker"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/helm"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/kim"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/kubectl"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/nerdctl"
 
   uninstall delete: [
     "/opt/rancher-desktop",
--- a/Casks/rancher.rb
+++ b/Casks/rancher.rb
@@ -17,10 +17,10 @@ cask "rancher" do
 
   auto_updates true
   conflicts_with formula: %w[
-                   docker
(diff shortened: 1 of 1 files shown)


neutric:
Could you please provide some context to your changes?

sanyer:
@neutric yeah, sure.
Docker Desktop and Rancher Desktop are not mutually exclusive, although similar in function.

Situation is further improved by Rancher not installing supporting binaries (docker, kubectl, helm, etc) by default and allowing user to chose which ones to install.

miccal:
Some background:
https://github.com/rancher-sandbox/rancher-desktop/issues/1081
https://github.com/rancher-sandbox/rancher-desktop/issues/1221
https://github.com/rancher-sandbox/rancher-desktop/issues/1300

commitay:
This shouldn't be merged. 

The casks do conflict depending on which options are enabled in Rancher Desktop and as Homebrew doesn't have any knowledge about those options they need to be handled as if they have been enabled.

Casks/rancher.rb
```diff
@@ -16,21 +16,13 @@
...
+    helm
+    kubernetes-cli
+  ]
 
   app "Rancher Desktop.app"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/docker"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/helm"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/kim"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/kubectl"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/nerdctl"
```
commitay, line 33: These binaries still need to be managed by Homebrew so they are removed if the cask is uninstalled, see https://github.com/rancher-sandbox/rancher-desktop/issues/272#issuecomment-931028377.

miccal:
Thank you @sanyer but considering all of the above I have decided that this change should not be merged. If users want to install both this and Docker, I would suggest they install one of them manually and make their own choices to ensure that they do not conflict.

sanyer:
@miccal interesting, thanks for the background.
I will check again how they both work for me.

sanyer:
@commitay the issues you mentioned do exist, but they affect only the execution of these container management tools, not installation.

sanyer:
Anyways, thank you both for the fast and useful feedback. I'll think about creating my own Tap for these both. Looks like a good candidate for trying out Taps :-)

commitay:
Just mentioning this here in advance as I'm sure someone will open another PR for this ...

> https://docs.rancherdesktop.io/faq/
>
> Q: Can I have Docker Desktop installed alongside Rancher Desktop?
>
> A: Yes, but they cannot be run at the same time as both Rancher Desktop and Docker Desktop use the same Docker socket (/var/run/docker.sock). Before starting one, be sure to stop the other first.

Despite what their FAQ says it still conflicts as far as Homebrew is concerned as uninstalling/reinstalling/upgrading one cask may remove files that are used by the other cask.

The other option would be adding a `caveat` explaining all the above but people usually don't bother reading them anyways ...

sanyer:
Here is my personal Tap for Docker Desktop and Rancher Desktop: https://github.com/sanyer/homebrew-containers-desktop

PR #117919, opened 2022-01-22
Update klayout from 0.26.8 to 0.27.7

adavidzh:
Notes:
- This is a draft PR for a couple of reasons:
  - I have not tested anything other than Monterey even if I got the `sha256`s are correct for all 3 options.
  - This version `0.27.7` has removed support for older macOS versions and I am not sure how to deal with that.
  - The new `dmg`s available are not `qt5MP-RsysPsys` (system Ruby and system Python) but `qt5Bre-RsysPhb38` so I added some `depends_on` stanzas.
- ~~`brew audit` complains about the lack of `desc` that was already missing.~~

Formal part:

### DIFF
--- a/Casks/klayout.rb
+++ b/Casks/klayout.rb
@@ -1,7 +1,7 @@
 cask "klayout" do
   version "0.27.7"
 
-if MacOS.version <= :catalina
+  if MacOS.version <= :catalina
     sha256 "7e7b515f2e0a735ce57f247a1a51e4f1b37b71d73dac5fb931c4e428ed38ff0b"
 
     url "https://www.klayout.org/downloads/MacOS/HW-klayout-#{version}-macOS-Catalina-1-qt5Brew-RsysPhb38.dmg",
--- a/Casks/klayout.rb
+++ b/Casks/klayout.rb
@@ -5,22 +5,27 @@ cask "klayout" do
     sha256 "7e7b515f2e0a735ce57f247a1a51e4f1b37b71d73dac5fb931c4e428ed38ff0b"
 
     url "https://www.klayout.org/downloads/MacOS/HW-klayout-#{version}-macOS-Catalina-1-qt5Brew-RsysPhb38.dmg",
-        verified: "klayout.org/"
+        verified: "www.klayout.org/downloads/MacOS/"
   elsif MacOS.version <= :big_sur
     sha256 "398524660a4fab288190791f57d34262b120faf6bb19c024b438826edb0a7d28"
 
     url "https://www.klayout.org/downloads/MacOS/HW-klayout-#{version}-macOS-BigSur-1-qt5Brew-RsysPhb38.dmg",
-        verified: "klayout.org/"
+        verified: "www.klayout.org/downloads/MacOS/"
   else
     sha256 "a5047a1058a8795f06ae7b26cce417ac093eee42b0fdd6d0bac86efaf5a003f4"
 
     url "https://www.klayout.org/downloads/MacOS/HW-klayout-#{version}-macOS-Monterey-1-qt5Brew-RsysPhb38.dmg",
-        verified: "klayout.org/"
+        verified: "www.klayout.org/downloads/MacOS/"
   end
 
-  appcast "https://www.klayout.de/development.html"
   name "KLayout"
+  desc "IC design layout viewer and editor"
   homepage "https://www.klayout.de/"
+
+  livecheck do
+    url "https://github.com/KLayout/klayout/"
+    strategy :github_latest
+  end
 
   depends_on macos: ">= :catalina"
   depends_on formula: "python@3.8"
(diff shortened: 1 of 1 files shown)


neutric:
> * This is a draft PR for a couple of reasons

Not that it matters much, but there is https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/

> * `brew audit` complains about the lack of `desc` that was already missing.

My suggestion based on https://www.klayout.de/intro.html: `desc "IC design layout viewer and editor"`

More [info on macOS support](https://github.com/KLayout/klayout/tree/master/macbuild) and [this](https://github.com/KLayout/klayout/issues/587) sheds some light on the file naming scheme.

Casks/klayout.rb
```diff
@@ -1,28 +1,35 @@
...
 
-  depends_on macos: ">= :mojave"
+  livecheck do
+    url "https://github.com/KLayout/klayout/"
+    strategy :github_latest
+  end
+
+  depends_on macos: ">= :catalina"
+  depends_on formula: "python@3.8"
+  depends_on formula: "qt@5"
```
miccal, line 32: The link https://github.com/KLayout/klayout/issues/587 states that both Qt5 and Python3.7 frameworks are included in the heavyweight (HW) versions of KLayout, so are these formula dependents necessary?

Casks/klayout.rb
```diff
@@ -5,23 +5,28 @@
...
   end
 
-  appcast "https://www.klayout.de/development.html"
   name "KLayout"
+  desc "IC design layout viewer and editor"
   homepage "https://www.klayout.de/"
 
+  livecheck do
+    url "https://github.com/KLayout/klayout/"
+    strategy :github_latest
```
neutric, line 27: Why not use the implicit strategy? Leaving `strategy` out worked fine locally.

miccal: Habit!

adavidzh: Included by @miccal.

Casks/klayout.rb
adavidzh: Right. Thanks to @neutric, the whole story on dependencies is much more clear and for `0.27.7` the relevant issue is https://github.com/KLayout/klayout/issues/976.

Studying https://github.com/KLayout/klayout/issues/976 in more detail, it seems that we have a couple of out-of-the-box options here:
- Stick with the `HW-klayout-0.27.7-macOS-{Catalina,BigSur,Monterey}-1-qt5Brew-RsysPhb38.dmg` and drop the dependencies (simple).
- It seems to be possible to run the same `ST-klayout-0.27.7-macOS-Catalina-1-qt5MP-RsysPsys.dmg` with Catalina, Big Sur, and Monterey. IIUC, Qt5 is embedded and Ruby and Python are from the OS so there would also be no dependencies (simpler?).
- Roll with `LW-klayout-0.27.7-macOS-{Catalina,BigSur,Monterey}-1-qt5Brew-Rhb27Phb38.dmg` that, IIUC would require pulling in `qt@5`, `ruby@2.7` and `python@3.8`.

Thanks to @neutric's pointer to [`klayout`'s `macbuild`](https://github.com/KLayout/klayout/tree/master/macbuild), I was also wondering if it is possible/desirable to build `klayout`. E.g., since it supports Qt6 and Python 3.9, those of us already using `Python@3.9` and `qt6`/`qt` would not end up pulling in `Python@3.8` and `qt5`/`qt@5`. I have no idea how the logic for the this would look like and/or if it is even feasible/desirable. I just thought it would be worth mentioning.

adavidzh:
> > * This is a draft PR for a couple of reasons
> 
> Not that it matters much, but there is https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/

Cool. I don't have write access :-)

Thanks a bunch for the pointers on how to decode the `klayout` macOS package names and build system!

neutric:
> Cool. I don't have write access :-)

I don't have write access, either. It works on PRs you created, at least it does for me. Search on page for "draft", it is rather low key in the right sidebar.

Casks/klayout.rb
neutric: Thanks for finding the more relevant issue, it was late and I took a shortcut. ;) I asked myself the same questions yesterday but was hoping that you might have some more insight derived from actually using the software. It might make sense to involve the [main developer](https://github.com/klayoutmatthias).

adavidzh:
> Search on page for "draft", it is rather low key in the right sidebar.

Eek. It's really low key and I missed it.

Casks/klayout.rb
adavidzh: I don't use it a lot so it could indeed a good idea to thread in Matthias, esp. if @miccal also thinks so.

miccal: Let's go with the first option and remove the dependencies, less issue that way especially considering Qt and Python are large dependencies with a a lot of further dependencies that users may not want installed.

Building would require this to be submitted to `homebrew-core`.

miccal:
Thank you @adavidzh and @neutric for all your work here.

PR #117983, opened 2022-01-23
Update roboform from 9.2.1 to 9.2.8

NicoHaase:


Casks/roboform.rb
```diff
@@ -1,5 +1,5 @@
 cask "roboform" do
-  version "9.2.1"
+  version "9.2.8"
   sha256 "2027bc48a1214097e59c36c0b496dd9c75d3f295643baf642731abb8ac1a4c63"
```
neutric, line 3: CI complains that it derived a differing checksum, as one would expect when the file differs. Did you manually override the `sha256` when using `brew bump-cask-pr`?

NicoHaase: No, that's pretty strange. I just ran `brew bump-cask-pr --version 9.2.8 roboform`. Sadly, my knowledge of Ruby scripting is not that profound, such that I could not debug this problem on my own :( In my last PR [113538](https://github.com/Homebrew/homebrew-cask/pull/113538), this worked as it should

neutric: Do you still have the shell window open? If so, post the command's output.

NicoHaase: Sorry, I closed it already :( Let's hope this happens once more on the next update, such that I can share it directly ;)

PR #118012, opened 2022-01-23
4peaks: add livecheck

bevanjkay:


miccal:
Weird `homepage` for this one:
```
|-> curl https://nucleobytes.com/4peaks/index.html
Warning: Binary output can mess up your terminal. Use "--output -" to tell 
Warning: curl to output it to your terminal anyway, or consider "--output 
Warning: <FILE>" to save to a file.
```
and adding `--output -` prints jibberish.

Perhaps an `:extract_plist` might be better?

neutric:
The livecheck is fine. Remove the `appcast` stanza...

PR #118020, opened 2022-01-24
eloston-chromium: various updates

miccal:
Continuation of https://github.com/Homebrew/homebrew-cask/pull/117980

I did not immediately see what @bevanjkay was driving at in his review -- sorry @bevanjkay !


neutric:
That looks neater, thanks!

frodera:
Yeah, including the architecture in the version string is mostly redundant in this case. Thanks for the cleanup, @miccal

PR #118099, opened 2022-01-25
navicat-premium 16.0.7

miccal:


miccal:
The `zh-CN` language version is giving me trouble, so I have removed it.

PR #118258, opened 2022-01-27
Add PanWriter.app

DrJume:

### DIFF
--- a/Casks/panwriter.rb
+++ b/Casks/panwriter.rb
@@ -15,11 +15,6 @@ cask "panwriter" do
   desc "Markdown editor with pandoc integration and paginated preview"
   homepage "https://panwriter.com/"
 
-  livecheck do
-    url :url
-    strategy :github_latest
-  end
-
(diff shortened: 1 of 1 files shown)


SMillerDev (commented):

DrJume:
Done

neutric:
This would most likely also benefit from `depends_on formula: "pandoc"` as `homepage` says:

> You also have to install pandoc to export to most formats.

DrJume:
Yes, you are right. Thanks for the tip!

neutric:
Question for the maintainers:
```
livecheck do
  url :url
  strategy :github_latest
end
```
seems to be slightly less common than the above but with `url :url`, as it is the implicit default, omitted. Which version do you prefer?

miccal:
@neutric if the `url` is a GitHub release we try first to omit it entirely and see if the `livecheck` is automatic (which it is in this case).

miccal:
Thank you @DrJume.

PR #118296, opened 2022-01-28
Update rancher binaries/uninstall

commitay:

Rancher Desktop will create these symlinks in `/usr/local/bin` so using `binary` stanzas to manage them only works when `HOMEBREW_PREFIX` is `/usr/local`. 

rd

Moving the `binaries` to `delete` like is used in the `docker` cask.

https://github.com/Homebrew/homebrew-cask/blob/34f2f56291def96d84f7e9dc1f30d0b9cc394a6f/Casks/docker.rb#L49-L61

### DIFF
--- a/Casks/rancher.rb
+++ b/Casks/rancher.rb
@@ -26,16 +26,16 @@ cask "rancher" do
                  ]
 
   app "Rancher Desktop.app"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/docker"
-  binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/helm"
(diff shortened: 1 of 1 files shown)


miccal:
Why not use `target` instead? For example:
```
binary "#{appdir}/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/docker", target: "/usr/local/bin/docker"
```

commitay:
> Why not use `target` instead?

I didn't see any other casks using `target` to link in `/usr/local/bin`.

IIRC using `target` in `/usr/local` / `HOMEBREW_PREFIX` was a bit finicky. Not sure how well `target` would handle non-Homebrew conflicts or if always linking binaries into `/usr/local/bin` is pattern Homebrew would want to introduce as it makes `HOMEBREW_PREFIX/bin` less clear?

miccal:
Agreed.

If users want the binaries, they should manage them from within the app.

PR #118396, opened 2022-01-30
nozbe: rename to nozbe-personal

miccal:


rdasgup:
Greate!!!

even at my site it was runing well and I have even Nozbe under revision control of home brew from now on.

Many Thanks for that
Roni

PR #118669, opened 2022-02-04
Update autofirma from 1.6.5 to 1.7.1

adruiza:


adruiza:
Installation fails in CI due to an interactive window, it happened in the previous version too.

PR #118757, opened 2022-02-06
Sage assets changed. Update the hashes.

culler:

### DIFF
--- a/Casks/sage.rb
+++ b/Casks/sage.rb
@@ -4,9 +4,9 @@ cask "sage" do
   version "9.5,1.3"
 
   if Hardware::CPU.intel?
-    sha256 "21606aa7476b70a01de6594e37d74625436ac6307ccafe19ff1146d2315189fd"
+    sha256 "035a882ec96f190e4ec6929913bdd8880035494f9ffe27e109c828eefb35945b"
   else
-    sha256 "4640428a7b41b829a1a18d6f5ecf9d22d2ae12c50d4b791fcebb532a369d70e0"
+    sha 256 "a134fe93a5fa61de58bf6d512f9867663b565bb2b72eea5c77b8cf8a0fc8cb81"
   end
 
   url "https://github.com/3-manifolds/Sage_macOS/releases/download/v#{version.csv.second}/SageMath-#{version.csv.first}_#{arch}.dmg",
--- a/Casks/sage.rb
+++ b/Casks/sage.rb
@@ -6,7 +6,7 @@ cask "sage" do
   if Hardware::CPU.intel?
     sha256 "035a882ec96f190e4ec6929913bdd8880035494f9ffe27e109c828eefb35945b"
   else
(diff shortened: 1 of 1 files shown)


culler:
I discovered a small bug immediately after converting from prerelease to release and uploaded new disk images.

neutric:
> I discovered a small bug immediately after converting from prerelease to release and uploaded new disk images.

Thanks for providing context!

culler:
I have now placed the md5 and sha256 hash values in architecture-specific json files in the assets.  Would it be possible for the cask script to read the sha256 hash from the appropriate json file instead of embedding it as a constant?

bevanjkay:
@culler The hashes are static in the cask file as a way of verifying the artifact contents at the time the cask is updated. It helps to stop the assets being updated upstream by bad actors and then being distributed via Homebrew (when the hash mismatches the install does not proceed). If the hash was dynamic it could also be manipulated by a bad actor upstream.

culler:
The problem is that I cannot take on the job of monitoring all third parties that are redistributing binaries that we distribute as github releases, and notifying them all with a PR every time a release asset changes.  (In this case, an issue was posted on our github site, not the Homebrew-cask site, to report that the hash in the cask did not match.)  While I do want the Homebrew-cask installs to work, there are limits to how much of my time I can devote to that.  I see that one option available in a cask is to just skip the hash check.  Maybe that would be a better idea in this case, since we are not able at this point to guarantee that release assets will never change after a release is created.

culler:
One other aspect of this is that, while it would be bad if a bad actor were able to impersonate a github release page, a much more likely scenario (especially with a 1GB download) is that a user receives a corrupt file.  Checking a hash which is published as an asset within the release does detect corrupt files, while specifying nocheck for the hash does not.

Also this particular app is signed and has been notarized by Apple.  The Apple gatekeeper would not allow a tainted app created by a bad actor to run.  So the corrupt file check is valuable while the bad actor check is redundant.

neutric:
You signed and notarized the disk image, which makes the checksum also redundant for file corruption purposes as opening the *.dmg will produce a `[...] is damaged and can't be opened. You should move it to the Bin.`

All the above could have been avoided by not retagging / changing the releases assets but instead releasing a new version, as laid out in the [Semantic Versioning Specification](https://semver.org/).

culler:
Thanks.  That is a good point about notarizing the disk image.  It was not
clear to me why it was important to do that extra (time-consuming) step.
Now I see.

I realize that a policy of never changing release assets would have
prevented this problem.  But, as I said, we were not prepared to enforce
that policy yet and, given the current state of the project, enforcing it
would make it much more likely for the cask to be out of date.  We may be
tightening up the policy soon - the reasons for doing so are becoming
clearer.  But, in the meantime, my conclusion from this very helpful
conversation is that it would work better and be no more dangerous if the
cask were to just skip the hash test.  That said, it appears very unlikely
to me that there will be any further changes to the assets for this
release, and we may well be following the versioning rules strictly before
the next SageMath release appears.

- Marc

PR #118763, opened 2022-02-06
Update zap list for private-internet-access

ColinHebert:
Add more configuration files to the private-internet-access zap list


bevanjkay:
Than you @ColinHebert

PR #118993, opened 2022-02-11
Rename dotnet to dotnet-runtime

ghost:
`dotnet` is ambiguous name


miccal:
Thank you @Uguisu32J but this is far too popular a Cask for us to be re-naming it now.

Uguisu32J:
I'm thinking of sending the same PR to `cask-versions`.

PR #119014, opened 2022-02-11
Update xamarin-mac.rb depends_on

oliverchaplin:
The version of mono-mdk in homebrew/homebrew-cask and homebrew/homebrew-cask-versions are currently the same. It would be good to be able to use the mono-mdk from homebrew/homebrew-cask


miccal:
Doesn't this Cask specifically require the `mono-mdk` version specific for Visual Studio users?

Although `mono-mdk` and `mono-mdk-for-visual-studio` are currenlty pointing to the same version, that may change in the future.

miccal:
Ping @robertleeblairjr for their thoughts also.

oliverchaplin:
I believe you actually may be right after some investigation. My bad I will close this pull request.

PR #119083, opened 2022-02-12
Add ujam

aaronk6:
- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.
- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [x] `brew audit --new-cask <cask>` worked successfully.
- [x] `brew install --cask <cask>` worked successfully.
- [x] `brew uninstall --cask <cask>` worked successfully.

PR #119134, opened 2022-02-13
Remove finisher-voodoo

aaronk6:
I suggest that you remove the `finisher-voodoo` cask. With #119083, I’ve added the UJAM App which can install all of UJAM’s products. Before the UJAM App existed, I made an attempt to add all UJAM products as individual casks, but that was rejected due to limited popularity. Only some of the Finisher products were added back then. However, in the meantime, `finisher-fluxx`, `finisher-fluxx`, and `finisher-neo` have already been removed due to the lack of installs:

https://github.com/Homebrew/homebrew-cask/pull/110501
https://github.com/Homebrew/homebrew-cask/pull/110500
https://github.com/Homebrew/homebrew-cask/pull/110499

Only `finisher-voodoo` is left now, which seems inconsistent to me.

Thanks for your work!


neutric:
```
==> Analytics
install: 0 (30 days), 0 (90 days), 2 (365 days)
```
@aaronk6 It didn't get removed until now because the analytics still showed interest above the threshold of "at least a couple of installs per year" last time around somebody checked.

PR #119151, opened 2022-02-14
Update fpc-laz from 3.2.0,2.0.12 to 3.2.2,2.2.0

samford:

---

This updates `fpc-laz` to the respective 3.2.2 and 2.2.0 versions (I'll create a similar `fpc-src-laz` PR in a bit). I removed the `pkg` specification, as this is using a dmg file. I'm not well-versed with the cask DSL, so do let me know if we need to keep this.

Besides updating this to the latest version(s), this PR reworks the `livecheck` block to:

* Check the RSS feed for the `Lazarus macOS x86-64` subdirectory, so we only work with the version information we're interested in (i.e., we don't have to worry about this information getting pushed out of the main project RSS feed).
* Move the regex out of the `strategy` block, as we should establish it using `#regex` and pass it in instead.
* Replace the explicit `\.intel-macosx\.dmg` ending in the regex with a looser `[^"' >]+?\.(?:dmg|pkg)`. The suffix and file type have fluctuated a bit over time and this looseness may help us to not miss versions in the future.
* Use the `Sourceforge` strategy instead of `PageMatch`. Granted, when we're using a custom URL, regex, and `strategy` block, this doesn't make a functional difference but it's potentially useful for tracking which strategies we're using and where.
* Use the `page.scan(regex).map` approach in the `strategy` block, which will collect all matches in the RSS feed, not just the first match.

### DIFF
--- a/Casks/fpc-laz.rb
+++ b/Casks/fpc-laz.rb
@@ -17,6 +17,8 @@ cask "fpc-laz" do
   end
 
   conflicts_with formula: "fpc"
+  
+  pkg "fpc-#{version.csv.first}-intelarm64-macosx.mpkg/Contents/Packages/fpc-#{version.csv.first}-intelarm64-macosx.pkg"
 
   uninstall pkgutil: [
     "org.freepascal.freePascalCompiler320.fpcinst386",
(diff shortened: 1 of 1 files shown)


bevanjkay:
Thanks @samford 👍🏻

PR #119176, opened 2022-02-14
Update fpc-src-laz from 3.2.0-2,2.0.12 to 3.2.2-20210709,2.2.0

samford:

---

This updates `fpc-src-laz` to the respective 3.2.2-20210709 and 2.2.0 versions. I updated the `pkg` argument to align with the `fpc-laz` changes in #119151 but please check my work. The `uninstall` stanza may need changes but I'm not familiar with how to approach it.

Besides updating this to the latest version(s), this PR reworks the `livecheck` block to:

* Check the RSS feed for the `Lazarus macOS x86-64` subdirectory, so we only work with the version information we're interested in (i.e., we don't have to worry about this information getting pushed out of the main project RSS feed).
* Move the regex out of the `strategy` block, as we should establish it using `#regex` and pass it in instead.
* Replace the (potentially mistaken?) use of the match anything `.` in the regex with `[.-]`, which is appropriate for the repeating part of the version (and will handle versions like 3.2.0, 3.2.0-2, 3.2.2-20210709, etc.).
* Replace the explicit `-laz\.pkg` ending in the regex with a looser `[^"' >]+?\.(?:dmg|pkg)`. The suffix and file type have fluctuated a bit over time and this looseness may help us to not miss versions in the future.
* Use the `Sourceforge` strategy instead of `PageMatch`. Granted, when we're using a custom URL, regex, and `strategy` block, this doesn't make a functional difference but it's potentially useful for tracking which strategies we're using and where.
* Use the `page.scan(regex).map` approach in the `strategy` block, which will collect all matches in the RSS feed, not just the first match.

### DIFF
--- a/Casks/fpc-src-laz.rb
+++ b/Casks/fpc-src-laz.rb
@@ -1,24 +1,22 @@
 cask "fpc-src-laz" do
-  version "3.2.0-2,2.0.12"
-  sha256 "8f85c873c3a224961a98fd24e97320a2134bbcc7f91c6d8153e53b36beb7fe74"
+  version "3.2.2-20210709,2.2.0"
+  sha256 "c0ed6b9261679ba040cdf07f4f5d13d915184a4164b1addf5a81e19b0738e87a"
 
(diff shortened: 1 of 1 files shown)


bevanjkay:
Thank you @samford

PR #119190, opened 2022-02-15
knox: remove

neutric:
Current version is from mid-2015. In this [tweet](https://twitter.com/joi/status/735526426749849600) vendor says that `knox` has been discontinued since 2015 and that it "doesn't run the greatest on El Capitan".

```
==> Analytics
install: 4 (30 days), 13 (90 days), 43 (365 days)
```

Remove instead?


miccal:
Yes, let's just remove it please.

PR #119175, opened 2022-02-14
Remove old Coriolis Systems software.

al45tair:
I'd prefer that this only be distributed through the https://coriolis-systems.com website directly. That way people get to read what's written there, whereas installing using Homebrew doesn't do that.

To be clear, I'm not 100% opposed to Homebrew being able to install it, but if it's going to do that it should probably fetch the appropriate license file and install that at the same time, and it would be nice if it also told people they should read about it on the website.

(It would also have been nice had whoever added it in the first place actually asked whether it was OK, or even mentioned it at all. I realise Homebrew is just downloading it and isn't doing anything wrong by doing that, but it would still have been nice had someone got in touch.)

PR #119343, opened 2022-02-17
Add licenses to Coriolis Casks

ed-flanagan:

- [x] `brew audit --cask i{defrag,partition}` is error-free.
- [x] `brew style --fix i{defrag,partition}` reports no offenses.

---

Was reading https://github.com/Homebrew/homebrew-cask/pull/119175 yesterday/today, figured I'd take a crack at adding the licenses described.

Added a `caveat` block with `discontinued`/`free_license`. Cargo-culted from https://github.com/Homebrew/homebrew-cask/pull/9255/files. `brew info idefrag` includes this now:
Screen Shot 2022-02-17 at 2 59 10 PM

I also updated/added a `depends_on` for the macOS version since site states upper bound is 10.13.

### DIFF
--- a/Casks/idefrag.rb
+++ b/Casks/idefrag.rb
@@ -17,7 +17,12 @@ cask "idefrag" do
     regex(%r{href=.*?/iDefrag[._-]v?(\d+(?:\.\d+)+)\.dmg}i)
   end
 
-  depends_on macos: ">= :yosemite"
+  depends_on macos: "<= :high_sierra"
 
   app "iDefrag.app"
+
+  caveats do
+    discontinued
+    free_license "https://coriolis-systems.com/downloads/iDefrag.png"
+  end
 end
--- a/Casks/ipartition.rb
+++ b/Casks/ipartition.rb
@@ -17,7 +17,14 @@ cask "ipartition" do
     regex(%r{href=.*?/iPartition[._-]v?(\d+(?:\.\d+)+)\.dmg}i)
   end
 
+  depends_on macos: "<= :high_sierra"
+
   app "iPartition.app"
 
   zap trash: "~/Library/Preferences/com.coriolis-systems.iPartition.plist"
+
+  caveats do
+    discontinued
+    free_license "https://coriolis-systems.com/downloads/iPartition.png"
+  end
 end


Casks/idefrag.rb
```diff
@@ -17,7 +17,12 @@
     regex(%r{href=.*?/iDefrag[._-]v?(\d+(?:\.\d+)+)\.dmg}i)
   end
 
-  depends_on macos: ">= :yosemite"
+  depends_on macos: "<= :high_sierra"
```
neutric, line 20: The requirement `depends_on macos: ">= :yosemite"` still has to stay as `5.1.3` only works with macOS 10.10–10.11. I believe this then has to be written as
```
depends_on macos: [
  :yosemite,
  :el_capitan,
  :sierra,
  :high_sierra
]
```

Casks/ipartition.rb
```diff
@@ -17,7 +17,14 @@
     regex(%r{href=.*?/iPartition[._-]v?(\d+(?:\.\d+)+)\.dmg}i)
   end
 
+  depends_on macos: "<= :high_sierra"
```
neutric, line 20: See above.
```
depends_on macos: [
  :yosemite,
  :el_capitan,
  :sierra,
  :high_sierra
]
```

Casks/idefrag.rb
ed-flanagan: I changed to a `<=` since https://docs.brew.sh/Cask-Cookbook#depends_on-macos says 10.10 is listed as the oldest supported version, thinking the `<=` implicit range would be 10.10-10.13.
I also saw this pattern in other Casks like: https://github.com/Homebrew/homebrew-cask/blob/ca26efe2e47ed6a13ad9f22637c18806f6e26fd6/Casks/deeper.rb#L39

neutric: Using `<=` is totally fine, but if `10.10` were not implicitly enforced, which I am not sure of, then we'd lose this requirement in the switch. Right now it explicitly says ` depends_on macos: ">= :yosemite"`, which either means that there is no implicit enforcement or that it just wasn't updated to reflect the moving list of symbol-to-version-mappings.

ed-flanagan: Sounds like a maintainer needs to confirm behavior and desired approach.

* PR's based on the assumption constraint is implicitly enforced, if not, then yeah, would need the explicit set.
More things that lead to my assumption (best guesses, I'm not familiar with `brew`'s code base):
    * Symbol definition usage https://github.com/Homebrew/brew/blob/89d31a3504479f009c4a374fd962d65d492efba2/Library/Homebrew/os/mac/version.rb#L16-L25
    * `DependsOn` for macOS https://github.com/Homebrew/brew/blob/9c03493774500cf16ced8938e1eb4eeae8216b20/Library/Homebrew/cask/dsl/depends_on.rb#L57
    * Which uses https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/macos_requirement.rb#L18
* The existing version constraints were introduced in 2017 here: https://github.com/Homebrew/homebrew-cask/commit/34a3e9497082170a824d1b2bce7cf34034429756. Guessing the constraints just haven't been updated recently.

miccal:
Thank you @ed-flanagan -- since MacOS 10.10 is the oldest one we "officially" support in Cask, I am fine with this.

PR #119807, opened 2022-03-01
Add Playdate SDK 1.9.0

ngs:

### DIFF
--- a/Casks/playdate-simulator.rb
+++ b/Casks/playdate-simulator.rb
@@ -1,18 +1,16 @@
-cask "playdate-sdk" do
+cask "playdate-simulator" do
   version "1.9.0"
   sha256 "c61e3ba45b65034f89690e5ad22b2ebd24e03df6d376b72e0fae62b003d104ff"
 
   url "https://download-keycdn.panic.com/playdate_sdk/PlaydateSDK-#{version}.pkg",
-      verified: "panic.com"
+      verified: "panic.com/playdate_sdk/"
   name "Playdate SDK"
-  desc "Playdate Lua and C APIs, docs and SSimulator for local development"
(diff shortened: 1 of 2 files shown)


miccal:
The `pkg` installs an `app` called `Playdate Simulator.app`, hence the name change.

ngs:
@miccal This .pkg installer contains Lua + C Library, Docs and CLI tools. So I think it's better to name this cask `playdate-sdk` .

miccal:
@ngs the naming follow our [strict token reference](https://docs.brew.sh/Cask-Cookbook#simplified-names-of-pkg-based-installers):

> If a pkg installs an App, then use that App name

Casks/playdate-simulator.rb
```diff
@@ -1,18 +1,16 @@
-cask "playdate-sdk" do
+cask "playdate-simulator" do
   version "1.9.0"
   sha256 "c61e3ba45b65034f89690e5ad22b2ebd24e03df6d376b72e0fae62b003d104ff"
 
   url "https://download-keycdn.panic.com/playdate_sdk/PlaydateSDK-#{version}.pkg",
-      verified: "panic.com"
+      verified: "panic.com/playdate_sdk/"
```
neutric, line 6: Shouldn't this be `verified: "download-keycdn.panic.com/playdate_sdk/"`?

miccal: Yes it should, thanks.

neutric:
> @miccal This .pkg installer also contains Lua + C Library, Docs and CLI tools.
> 
> So I think it's better to name this cask `playdate-sdk` .

This information is not lost as it is contained in `name` and `desc`. The `token` fulfills another [purpose](https://docs.brew.sh/Cask-Cookbook#purpose).

PR #119875, opened 2022-03-02
gcc-*-embedded: fix zap

neutric:
---
Not sure why `zap delete` was chosen but using `trash` instead should allow for easier zap error recovery.


pthariensflame:
Thanks!  This looks better.  I only used `zap delete:` because I was transforming it from an `uninstall delete:`, so I have no reason to keep it that way.

SMillerDev (approved):

PR #119778, opened 2022-02-28
Add Rambox version 2.0.1

paour:
A new cask is created, since Rambox (CE) 0.x still exists, and no migration is handled to version 2.

**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making all changes to a cask, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [x] `brew audit --new-cask <cask>` worked successfully.
- [x] `brew install --cask <cask>` worked successfully.
- [x] `brew uninstall --cask <cask>` worked successfully.

PR #119915, opened 2022-03-02
rambox 2.0.1

miccal:
Continuation of https://github.com/Homebrew/homebrew-cask/pull/119778

### DIFF
--- a/Casks/rambox.rb
+++ b/Casks/rambox.rb
@@ -1,12 +1,14 @@
 cask "rambox" do
-  version "0.7.9"
-  sha256 "237315fd3d5790e84b1d531024d44fd8fb29066123b02f2580840d069edf1dc2"
+  version "2.0.1"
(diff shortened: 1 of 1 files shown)


miccal:
Or just install `rambox-ce` from the `homebrew-cask-versions` repository.

ric2b:
I lost all my settings and configured services after the last update, is there a way to recover it or at least prevent it from happening to other people?

seebi:
Interesting - I thought I lost everything as well but just uninstalling rambox and installing rambox-ce healed everything ...

ric2b:
This update should've added as a separate application until an automatic migration was in place, as it is the update is incredibly disruptive. At minimum a message should be displayed on startup with the above workaround.

timkgh:
Disappointing that you are forcing the Pro version on CE users, no questions asked.

PR #119919, opened 2022-03-02
the-unarchiver: bump version 4.3.4, update livecheck, update url

bevanjkay:


owine:
I don't think the new livecheck is working as expected:

`the-unarchiver : 4.3.4,1646238910 ==> 137,`

bevanjkay:
@owine It seems like the change in download URL formatting must have been temporary, I am reverting to the previous now.

PR #75000, opened 2020-01-02
Update radarr.rb

Mactastic1-5:
Added dependency requirement as indicated by https://github.com/Radarr/Radarr/wiki/Installation and set permissions to run. Version wasn't updated.

After making all changes to the cask:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [  ] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #120043, opened 2022-03-05
Add Prowlarr 0.2.0.1448

alexreg:
The only small issue is that it's beta software, though it's already evidently popular and in widespread use.

### DIFF
--- a/Casks/prowlarr.rb
+++ b/Casks/prowlarr.rb
@@ -1,0 +1,23 @@
+cask "prowlarr" do
+  arch = Hardware::CPU.intel? ? "x64" : "arm64"
+
+  version "0.2.0.1448"
+
+  if Hardware::CPU.intel?
+    sha256 "ee499053a9e0a3dd1e9520a33c0dbb378cc64d2e93e31228d1c9c1f5b73288ed"
+  else
+    sha256 "e48498a72902298b857bf223efccb7d8fe5d5192db7141f0821b3de0b2d44b83"
+  end
+
+  url "https://github.com/Prowlarr/Prowlarr/releases/download/v#{version}/Prowlarr.develop.#{version}.osx-app-core-#{arch}.zip",
+      verified: "github.com/Prowlarr/Prowlarr/"
+  name "Prowlarr"
+  desc "Indexer manager/proxy for various PVR apps"
+  homepage "https://prowlarr.com/"
+
+  depends_on macos: ">= :high_sierra"
+
+  app "Prowlarr.app"
+
+  # No zap stanza required
+end
--- a/audit_exceptions/github_prerelease_allowlist.json
+++ b/audit_exceptions/github_prerelease_allowlist.json
@@ -23,6 +23,7 @@
   "nuclear": "all",
   "pock": "all",
   "protonmail-bridge": "all",
+  "prowlarr": "all",
   "ql-ansilove": "all",
   "profilecreator": "all",
   "qdesktop": "all",


neutric:
> The only small issue is that it's beta software, though it's already evidently popular and in widespread use.

This still belongs in `homebrew-cask` until it is out of beta. To get CI to agree, add the app to [`github_prerelease_allowlist.json`](https://github.com/Homebrew/homebrew-cask/blob/master/audit_exceptions/github_prerelease_allowlist.json).

Casks/prowlarr.rb
```diff
@@ -0,0 +1,30 @@
...
+  url "https://github.com/Prowlarr/Prowlarr/releases/download/v#{version}/Prowlarr.develop.#{version}.osx-app-core-#{arch}.zip",
+      verified: "github.com/Prowlarr/Prowlarr/"
+  name "Prowlarr"
+  desc "Indexer manager/proxy built on popular *arr .NET/ReactJS base stack"
+  homepage "https://prowlarr.com"
+
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
```
neutric, line 21: This can be dropped as `strategy :github_latest` isn't needed here and we thus can rely on the implicit default for GitHub URLs.

Casks/prowlarr.rb
```diff
@@ -0,0 +1,30 @@
...
+  if Hardware::CPU.intel?
+    sha256 "ee499053a9e0a3dd1e9520a33c0dbb378cc64d2e93e31228d1c9c1f5b73288ed"
+  else
+    sha256 "e48498a72902298b857bf223efccb7d8fe5d5192db7141f0821b3de0b2d44b83"
+  end
+
+  url "https://github.com/Prowlarr/Prowlarr/releases/download/v#{version}/Prowlarr.develop.#{version}.osx-app-core-#{arch}.zip",
+      verified: "github.com/Prowlarr/Prowlarr/"
+  name "Prowlarr"
+  desc "Indexer manager/proxy built on popular *arr .NET/ReactJS base stack"
```
neutric, line 15: "built on popular *arr .NET/ReactJS base stack" is irrelevant imho. Instead, I'd add "Torrent" and "Usenet" and maybe "integration with PVR apps" in an elegant way.

Casks/prowlarr.rb
```diff
@@ -0,0 +1,30 @@
...
+    strategy :github_latest
+  end
+
+  depends_on macos: ">= :high_sierra"
+
+  app "Prowlarr.app"
+
+  preflight do
+    set_permissions "#{staged_path}/Prowlarr.app", "0755"
+  end
```
neutric, line 29: Just out of interest: what happens if you don't set it to 0755?

neutric (commented):
Thanks for your contribution to `homebrew-cask`, @alexreg! My review is in the code. Take it with a grain of salt as I am not a maintainer.

Casks/prowlarr.rb
alexreg: It's not totally irrelevant, as it suggests compatibility with Radarr/Sonarr/etc.. but I know what you meant. I just took the first part of the official description... which isn't the best.

Casks/prowlarr.rb
alexreg: To be fair, I was copying this from the Radarr cask just to be safe (same authors, built on the same stack). It seems to require a restart to work anyway.

Casks/prowlarr.rb
alexreg: Ah right, good to know. Presumably can be removed for *radarr* too?

alexreg:
> Thanks for your contribution to `homebrew-cask`, @alexreg! My review is in the code. Take it with a grain of salt as I am not a maintainer.

Thanks for the review!

Casks/prowlarr.rb
neutric: You could stick that information into the "integration with PVR apps" bit.

Casks/prowlarr.rb
neutric: No, there they differentiate between release/stable (the most current one being designated as latest) and pre-release/develop. This isn't the case here as far as I could see.

Casks/prowlarr.rb
```diff
@@ -0,0 +1,30 @@
...
+  homepage "https://prowlarr.com"
+
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
+
+  depends_on macos: ">= :high_sierra"
+
+  app "Prowlarr.app"
```
neutric, line 25: Please try to add a [`zap` stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap). The linked documentation provides some recommended methods to create it.

Casks/prowlarr.rb
neutric: It was added to `radarr` in https://github.com/Homebrew/homebrew-cask/pull/75000. Not sure this is necessary here or there, can you please test?

Casks/prowlarr.rb
alexreg: Yeah, I suppose because this is still pre-1.0. I see your point, though maybe it's worth leaving in for when it reaches v1.0 and the release model changes?

Casks/prowlarr.rb
alexreg: I looked into this, but it seems Prowlarr (like Radarr) puts absolutely nothing in ~/Library... to my surprise too!

Casks/prowlarr.rb
neutric: We can put it back then. `strategy :github_latest` costs an API call and they are limited, that's why we remove it if not necessary.

alexreg: Got it.

Casks/prowlarr.rb
neutric: Same here. Please add `# No zap stanza required` on new line just above the final `end`.

Casks/prowlarr.rb
neutric: I installed it without the `preflight` stanza and didn't find any obvious need for this. Please remove until a need for it reveals itself.

Casks/prowlarr.rb
alexreg: Done.

alexreg:
So, it seems the one outstanding issue is that latest macOS (or maybe just on M1 machines?) does not like unsigned apps, and doesn't even give me the option in the Security pane to allow it. How odd. Is there a workaround for this?

Casks/prowlarr.rb
bevanjkay: @alexreg Any thoughts here?

alexreg: Ah sorry, missed this. You could be right, I'm just a bit loath to remove it because it seems to be required for Radarr, which is built on the same stack and by the same devs. Also because I was already getting a security issue to do with it being unsigned, though that could be independent. In any case, I see your point. I'll remove it if you really want.

neutric:
> So, it seems the one outstanding issue is that latest macOS (or maybe just on M1 machines?) does not like unsigned apps, and doesn't even give me the option in the Security pane to allow it. How odd. Is there a workaround for this?

Is [this](https://wiki.servarr.com/prowlarr/faq#help-my-mac-says-prowlarr-cannot-be-opened-because-the-developer-cannot-be-verified) your issue?

Casks/prowlarr.rb
neutric:
```
%  ls -al /Applications/Prowlarr.app/Contents/MacOS/Prowlarr
-rwxr-xr-x 1 neutric admin 113256 2022-01-31 19:22 /Applications/Prowlarr.app/Contents/MacOS/Prowlarr
```
The reason why it was added to `radarr` does not apply here. Permissions are fine without using the `preflight` stanza (that is, they are already `0755`).

The same is true for the current version of `radarr`, installed via Homebrew with the `preflight` stanza removed beforehand:

```
ls -al /Applications/Radarr.app/Contents/MacOS/Radarr
-rwxr-xr-x 1 neutric admin 113032 2022-03-01 02:34 /Applications/Radarr.app/Contents/MacOS/Radarr
```

At least on Monterey all is fine.

alexreg:
@neutric Afraid not quite (although running `codesign` on it does indeed fix it). I don't get that dialogue box; it just crashes on start-up. I can only see what's going on in Console.

Casks/prowlarr.rb
alexreg: Okay, so let's remove those preflight stanzas on both casks?

neutric:
> I can only see what's going on in Console.

What does it say there?

alexreg:
@neutric https://github.com/Prowlarr/Prowlarr/issues/886

neutric:
The crash only happens when launching it through Alfred. Launching it the "normal" way gives you the expected dialog window. Right-click and open once and it'll not bother you again.

alexreg:
@neutric Weird. That didn't happen for me. But I think we can put this down to a macOS bug

Casks/prowlarr.rb
alexreg: @neutric ?

bevanjkay:
Thank you @alexreg

alexreg:
Thanks for the reviews, both of you.

Casks/prowlarr.rb
neutric: Sorry, didn't see this until now. Retroactive 👍 on removing them.

PR #120154, opened 2022-03-07
dropbox: update livecheck

samford:

---

This is a follow-up to #120131. Same as that PR, I noticed that the check for `dropbox` wasn't working properly on ARM64 and I created a fix last night with the intention of opening a PR in the morning 😅

The regex in this PR is a bit looser, as it doesn't account for text in the filename after the version (e.g., `.dmg`, `.arm64.dmg`). With this approach, we only use one regex regardless of the execution architecture and we may not need to update the regex if the filename suffix changes further in the future (i.e., the check would continue to return version information instead of an error) . Since this regex is matching within the `location` header string, using a looser regex isn't as much of a concern in this context (unlike most body content like HTML, JSON, etc.).

Besides that, `#regex` should come before `#strategy` in a `livecheck` block and this PR also addresses that. For what it's worth, the existing regex intends to match a dot in text like `.arm64` but the dot isn't escaped like it should be (i.e., `\.`), so it matches any character in that location. This isn't an issue with this PR but it's something to be careful about in the future.

@bevanjkay Thoughts on this approach?

### DIFF
--- a/Casks/dropbox.rb
+++ b/Casks/dropbox.rb
@@ -1,6 +1,5 @@
 cask "dropbox" do
   arch = Hardware::CPU.intel? ? "" : "&arch=arm64"
-  livecheck_folder = Hardware::CPU.intel? ? "" : "arm64"
 
   version "143.4.4161"
   sha256 :no_check
@@ -12,8 +11,8 @@ cask "dropbox" do
 
   livecheck do
     url :url
+    regex(%r{/Dropbox(?:%20|[._-])v?(\d+(?:\.\d+)+)}i)
     strategy :header_match
-    regex(/Dropbox%20v?(\d+(?:\.\d+)+)(?:.#{livecheck_folder})?\.dmg/i)
   end
 
   auto_updates true


samford:
> > Besides that, #regex should come before #strategy in a livecheck block and this PR also addresses that.
> 
> Could you make a rubocop check for this? Because currently it just comes down to reviewer preference

I certainly agree with the idea. Long story short, it's been on my to-do list for a while but there are obstacles that stand in the way.

The order of components in formulae/casks isn't handled in a universal way and instead we have separate implementations for each. Also, the related formula and cask Rubocops only account for the order of top-level components at the moment and don't check component order within a block (see [`ast_constants.rb`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/ast_constants.rb) and [`rubocops/cask/constants/stanza.rb`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/rubocops/cask/constants/stanza.rb)). I haven't looked into whether the existing implementations already support that on a technical level but, worst case scenario, we would have to modify both to allow for this possibility.

Ideally, we would only have one way of handling component order that works for both formulae and casks. Both of the existing approaches have functionality that the other is lacking (e.g., the formula component order setup distinguishes between method calls and blocks (casks do not) and the cask setup groups components into stanzas with whitespace in between (formulae only seem to care about the order of components)), so it would be nice to have the best of both in one cohesive setup, if possible.

No doubt this type of Rubocop is technically possible but it may require a notable amount of work regardless of how we approach it (e.g., maintaining the status quo by implementing everything twice or putting in the work to synthesize these setups).

I'll try to find a moment to look into this enough to understand how much work would be involved but, if it's significant, I may not have the time to work on it for a while. If I don't end up having time, I can create a "help-wanted" issue in Homebrew/brew to explain the situation, in case someone else wants to take it on (it isn't something that requires livecheck experience).

bevanjkay:
Thanks @samford - I erred on the side of being more verbose, but I like this approach more anyway.

PR #119932, opened 2022-03-03
update rescuetime to 2.16.7.1

sschemer:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making all changes to a cask, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

PR #120282, opened 2022-03-09
rescuetime: bump version, update livecheck, update url

bevanjkay:
See https://github.com/Homebrew/homebrew-cask/pull/119932#issuecomment-1063206142

### DIFF
--- a/Casks/rescuetime.rb
+++ b/Casks/rescuetime.rb
@@ -1,14 +1,18 @@
 cask "rescuetime" do
-  version "2.16.7.1"
+  version "3.0.13.2"
   sha256 :no_check
 
-  url "https://www.rescuetime.com/installers/RescueTimeInstaller.pkg"
-  appcast "https://www.rescuetime.com/release-notes/mac"
+  url "https://www.rescuetime.com/installers/RescueTimeInstaller_X.pkg"
   name "RescueTime"
   desc "Time optimising application"
   homepage "https://www.rescuetime.com/"
 
-  pkg "RescueTimeInstaller.pkg"
+  livecheck do
+    url :url
+    strategy :extract_plist
+  end
+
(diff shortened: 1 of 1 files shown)


SMillerDev (approved):

k4rtik:
@bevanjkay @SMillerDev this update should not have been merged without careful reading of what RT v3 is about. It led to all sorts of problems for me. See: https://community.rescuetime.com/t/whats-with-the-new-sudden-unannounced-changes/126/3

RescueTime v3 seems like a completely different product. There should perhaps be two different casks for the classic RT v2.16.7.1 and another for RTX starting with v3.0.13.2.

If you all agree, I can send a PR to make that happen.

bevanjkay:
@k4rtik the cask was updated to version 3 on advice from the upstream vendor. 

A cask for version 2 could be submitted to `homebrew-cask-versions`.

k4rtik:
> the cask was updated to version 3 on advice from the upstream vendor.

Who is the upstream vendor here? 

RT does not recognize brew as a source of distributing itself, see https://community.rescuetime.com/t/whats-with-the-new-sudden-unannounced-changes/126/6?u=kartik

k4rtik:
Oh, I see the comment linked in the OP now. I don't know if we should consider a cask in `homebrew-cask-versions`. This is a breaking change for existing users (like it was for me) when we upgrade, that too without any notice!

The user-facing product is called RescueTime X (v3) and they are calling the older product RescueTime Classic.

Would it be acceptable to keep this cask to continue to be v2.x and send a PR for another called `rescuetime-x`?

SMillerDev:
> Would it be acceptable to keep this cask to continue to be v2.x and send a PR for another called rescuetime-x?

I would prefer a new cask for rescuetime-classic. This version is the future of rescuetime according to upstream so that should be the default.

PR #120283, opened 2022-03-10
Update scribus.rb from 1.4.8 to 1.5.6.1

jsoref:
This reverts #109745

https://docs.brew.sh/Installation#macos-requirements says that the system requirements are:
> macOS Catalina (10.15) (or higher)

The change appears to have taken the wrong side of the `if` resulting in something that only works on Mojave (or lower).

### DIFF
--- a/Casks/scribus.rb
+++ b/Casks/scribus.rb
@@ -1,25 +1,13 @@
 cask "scribus" do
-  version "1.4.8"
-  sha256 "9626c35ca5de5da59ac983efac3572318d327b3a921522c9f80a525b039a0af5"
+  version "1.5.6.1"
+  sha256 "5a79350884a405e223789bc65038d7c40f41c97bf352e8da0d96a9444a0bddd5"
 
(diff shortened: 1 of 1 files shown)


neutric:
> https://docs.brew.sh/Installation#macos-requirements says that the system requirements are:
> 
> > macOS Catalina (10.15) (or higher)

And the footnote says

> 10.15 or higher is recommended, while 10.10–10.14 are supported on a best-effort basis.

Also, these requirements are regarding Homebrew itself, not the software being served.

The change you are trying to revert was done because `homebrew-cask` is for stable versions, if available. The more up-to-date, devel-version is now in `homebrew-cask-versions`.

If this here doesn't run on anything above Mojave, add a `depends_on macos: "<= :mojave"`.

jsoref:
I can't speak with certainty for things between mojave and Big Sur, but based on the fact that it didn't run here and the construction of the code that was removed by that PR, that seems likely.

So, I should revert my changes and _just_ add that one field? (Seems like I should make a new pr to do that)

PR #120321, opened 2022-03-10
Update blender-benchmark version and livecheck

sanyer:

### DIFF
--- a/Casks/blender-benchmark.rb
+++ b/Casks/blender-benchmark.rb
@@ -9,11 +9,9 @@ cask "blender-benchmark" do
 
   livecheck do
     url :homepage
-    strategy :page_match do |page|
-      match = page.match(%r{.*Benchmark(\d+(?:\.\d+)+)/.*benchmark[._-]launcher[._-](\d+(?:\.\d+)+)[._-]macos\.dmg}i)
-      next if match.blank?
-
-      "#{match[2]},#{match[1]}"
+    regex(%r{["'][^"']+?Benchmark(\d+(?:\.\d+)+)/[^"']*?benchmark[._-]launcher[._-](\d+(?:\.\d+)+)[._-]macos\.dmg}i)
+    strategy :page_match do |page, regex|
+      page.scan(regex).map { |match| "#{match[1]},#{match[0]}" }
     end
   end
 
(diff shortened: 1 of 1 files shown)


Casks/blender-benchmark.rb
```diff
@@ -1,15 +1,20 @@
...
 
   livecheck do
-    url "https://opendata.blender.org/cdn/BlenderBenchmark#{version.major_minor}/launcher/"
-    regex(/href=.*benchmark[._-]launcher[._-](\d+(?:\.\d+)+)[._-]macos\.dmg/i)
+    url :homepage
+    strategy :page_match do |page|
+      match = page.match(%r{.*Benchmark(\d+(?:\.\d+)+)/.*benchmark[._-]launcher[._-](\d+(?:\.\d+)+)[._-]macos\.dmg}i)
+      next if match.blank?
+
+      "#{match[2]},#{match[1]}"
```
samford, line 16:
```suggestion
    regex(%r{["'][^"']+?Benchmark(\d+(?:\.\d+)+)/[^"']*?benchmark[._-]launcher[._-](\d+(?:\.\d+)+)[._-]macos\.dmg}i)
    strategy :page_match do |page, regex|
      page.scan(regex).map { |match| "#{match[1]},#{match[0]}" }
```

This suggestion addresses a few areas:

* Use of `.*` should be strengthened to something contextually-appropriate, to reduce the risk of the regex matching something it shouldn't. The URL is found within JSON data in this instance, so we can restrict matching to the string literal using `[^"']+?`/`[^"']*?`.
* The regex should be established outside of the `strategy` block using `#regex` and passed in (i.e., not defined within the `strategy` block).
* The `strategy` block should use the `page.scan(regex).map { ... }` approach to format version strings. This will handle all matches on the page instead of only the first match. It's less of a concern in this specific case since there's currently only one available version on the homepage but it's better to standardize on this more flexible approach.

samford:
I'll leave final review to our cask maintainers. If the overall approach in this PR is acceptable, the `livecheck` block is good to go at this point.

SMillerDev (approved):

SMillerDev:
Thanks @sanyer !

PR #120360, opened 2022-03-11
Update sha265 sum of syntax-hightlight 2.1.10

ffuugoo:
The [`Syntax.Highlight.zip` for the 2.1.10 release has been updated "in-place", without pushing new release](https://github.com/sbarex/SourceCodeSyntaxHighlight/issues/154#issuecomment-1065009070), so the hash sum doesn't match.


SMillerDev (approved):
Thanks @ffuugoo  ! Without contributions like yours it'd be impossible to keep homebrew going with the high standards that users have come to expect from the project. You can feel good knowing that you've made the world a tiny bit better for homebrew users around the world! 👍 🎉

PR #120407, opened 2022-03-12
Update Latest from 0.7.3 to 0.8

mangerlahn:

### DIFF
--- a/Casks/latest.rb
+++ b/Casks/latest.rb
@@ -1,8 +1,8 @@
 cask "latest" do
-  version "0.7.3,462"
-  sha256 :no_check
+  version "0.8,487"
+  sha256 "7740a16afde28845c5f1dee951208e2fadf6a1e214668e980164394c8ad34169"
 
-  url "https://max.codes/latest/download"
+  url "https://max.codes/latest/#{version.csv.first}.zip"
   name "Latest"
   desc "Utility that shows the latest app updates"
   homepage "https://max.codes/latest"


Casks/latest.rb
```diff
@@ -1,8 +1,8 @@
 cask "latest" do
-  version "0.7.3,462"
+  version "0.8,487"
   sha256 :no_check
 
-  url "https://max.codes/latest/download"
+  url "https://max.codes/latest/Latest.zip"
```
neutric, line 5:
```suggestion
  url "https://max.codes/latest/#{version}.zip"
```
The sparkle feed revealed https://max.codes/latest/0.8.zip, so we can switch to the versioned `url`. This requires setting `sha256` as well.

mangerlahn: I am sorry for my late response. `version` currently includes the build number as well, which is not part of the URL. 0.8.zip and Latest.zip point to the same zip file, so I _think_ it would be fine to keep the generic url? I can add the sha256 string nonetheless.

neutric: Sorry, I must have been distracted. Change suggestion is updated. Keeping the `url` pointing to `Latest.zip` would produce a checksum mismatch as soon as you release a new version...

mangerlahn: That makes sense, thanks for the updated suggestion!

SMillerDev (approved):

PR #120463, opened 2022-03-13
signal: use only the universal url

lpla:

Signal.org only shows the universal link in the "Get Signal" page.

### DIFF
--- a/Casks/signal.rb
+++ b/Casks/signal.rb
@@ -1,15 +1,9 @@
 cask "signal" do
-  arch = Hardware::CPU.intel? ? "x64" : "arm64"
-
   version "5.35.0"
 
-  if Hardware::CPU.intel?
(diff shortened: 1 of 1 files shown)


neutric:
> Signal.org only shows the universal link in the "Get Signal" page.

The `url` used for `livecheck` purposes has links for architecture-optimized builds, though. Is there any documentation pointing to the deprecation of those in favor of the universal build?

I downloaded the universal build of `5.34`directly from the homepage and, after the initial run and automatic update to the latest version `5.35`, the Signal.app Info (cmd+i) shows "Application (Intel)" where it showed "Application (Universal)" before. 

Signal.org probably avoids a few support tickets by not giving end users the choice of downloading the wrong binary for their architecture and instead serving the universal build. Homebrew can reliably detect the architecture, though, so no need for the much larger universal build.

lpla:
Well reasoned. Closing PR.

PR #120530, opened 2022-03-15
Remove handling of deleted 02_feature_request.yml

loganrosen:
This template was removed in #120491, but this file wasn't modified accordingly, so the sync flow has been failing.


SMillerDev (approved):
Thanks @loganrosen

PR #120552, opened 2022-03-16
Add Balena Etcher builds for Apple Silicon

nilsstreedain:

- [ x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [ x] `brew audit --cask <cask>` is error-free.
- [ x] `brew style --fix <cask>` reports no offenses.

### DIFF
--- a/Casks/balenaetcher.rb
+++ b/Casks/balenaetcher.rb
@@ -1,15 +1,14 @@
 cask "balenaetcher" do
-  
   version "1.7.7"
-  
+
   if Hardware::CPU.intel?
     sha256 "61a8835661e22854204cbd0c9948312f2641be8e02eea43d5c507f369b167972"
     url "https://github.com/balena-io/etcher/releases/download/v#{version}/balenaEtcher-#{version}.dmg",
-      verified: "github.com/balena-io/etcher/"
+        verified: "github.com/balena-io/etcher/"
   else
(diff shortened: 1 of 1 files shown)


bevanjkay:
@nilsstreedain Thank you for your submissions but we should not be distributing a community build to users expecting to receive the application from the actual authors.

nilsstreedain:
> @nilsstreedain Thank you for your submission but we should not be distributing a community build to users expecting to receive the application from the actual authors.

This is built using the codebase of BalenaEtcher (just recompiled for Apple Silicon) which is written to support Apple Silicon and is specifically referenced by BalenaEtcher maintainers in [balena/etcher#3595](https://github.com/balena-io/etcher/issues/3593). They went this route because they do not currently have an Apple Silicon Mac in their build pipeline.

bevanjkay:
@nilsstreedain I do understand that, but `homebrew-cask` should not be distributing a community build in a cask where a user would be expecting to receive the vendor build.


See this note from upstream's issue: https://github.com/balena-io/etcher/issues/3714

>     Note: These versions are not built, maintained, or supervised by balena, so proceed with caution.

nilsstreedain:
It has been pointed to by Balena maintainers as what to use while they determine how to move forward with their build structure but I understand. Would it be reasonable for me to create a separate balenaetcher-community cask in the meantime?

PR #120825, opened 2022-03-22
Add Cask for Free-GPGMail 2021.3.1

jpbernius:
[`Free-GPGMail`](https://github.com/Free-GPGMail/Free-GPGMail) is a fork of GPG Mail (as part of `gpg-suite`):
> A modification of the Apple Mail plugin for GnuPG encrypted e-mails, so it does not require a support plan.

This PR adds the cask prepared by @EivindArvesen as diskussed in [Free-GPGMail#41](https://github.com/Free-GPGMail/Free-GPGMail/issues/41) as an official Cask.

---

- [X] `brew audit --cask free-gpgmail` is error-free.
- [X] `brew style --fix free-gpgmail` reports no offenses.

- [X] `brew audit --new-cask free-gpgmail` worked successfully.
- [X] `brew install --cask free-gpgmail` worked successfully.
- [X] `brew uninstall --cask free-gpgmail` worked successfully.

---

Many Thanks to @EivindArvesen for his work on this cask!

### DIFF
--- a/Casks/free-gpgmail.rb
+++ b/Casks/free-gpgmail.rb
@@ -1,0 +1,48 @@
+cask 'free-gpgmail' do
+  name "Free-GPGMail"
+  desc "A modification of the Apple Mail plugin for GnuPG encrypted e-mails, so it does not require a support plan."
+  homepage "https://github.com/Free-GPGMail/Free-GPGMail"
+  auto_updates false
+  depends_on macos: ">= :mojave"
+  depends_on cask: "gpg-suite-no-mail"
+
+  preflight do
+    system "rm", "#{ENV['HOME']}/Library/Mail/Bundles/Free-GPGMail_*.mailbundle"
+  end
+
+  if MacOS.version <= :mojave
+    version '2021.3'
+    sha256 '7b4eaf2d52ddc1519dfec4087f09bad018657fde87a87021904075f55ac00377'
+    url "https://github.com/Free-GPGMail/Free-GPGMail/releases/download/v#{version}/Free-GPGMail_3-#{version}.mailbundle.zip"
+    artifact "Free-GPGMail_3.mailbundle", target: "#{ENV['HOME']}/Library/Mail/Bundles/Free-GPGMail_3.mailbundle"
+    uninstall delete: "~/Library/Mail/Bundles/Free-GPGMail_3.mailbundle"
+  elsif MacOS.version <= :catalina
+    version '2021.3'
+    sha256 'c2fda8cc9a52825cf0c0a9d954036e215076db5ec8b898c35d16d306aa6b1ea4'
+    url "https://github.com/Free-GPGMail/Free-GPGMail/releases/download/v#{version}/Free-GPGMail_4-#{version}.mailbundle.zip"
+    artifact "Free-GPGMail_4.mailbundle", target: "#{ENV['HOME']}/Library/Mail/Bundles/Free-GPGMail_4.mailbundle"
+    uninstall delete: "~/Library/Mail/Bundles/Free-GPGMail_4.mailbundle"
+  elsif MacOS.version <= :big_sur
+    version '2021.3'
+    sha256 '438519fa8b5ed04081553efd58b5edf52b6847ea3719fe9b1331e97c8021ca43'
+    url "https://github.com/Free-GPGMail/Free-GPGMail/releases/download/v#{version}/Free-GPGMail_5-#{version}.mailbundle.zip"
+    artifact "Free-GPGMail_5.mailbundle", target: "#{ENV['HOME']}/Library/Mail/Bundles/Free-GPGMail_5.mailbundle"
+    uninstall delete: "~/Library/Mail/Bundles/Free-GPGMail_5.mailbundle"
+  elsif MacOS.version >= :monterey
+    version '2021.3'
+    sha256 'a3e9d799ee332093303f4bf7ecc03d084e0bd3e28954cedc7d9149f660ae72e5'
+    url "https://github.com/Free-GPGMail/Free-GPGMail/releases/download/v#{version}/Free-GPGMail_6-#{version}.mailbundle.zip"
+    artifact "Free-GPGMail_6.mailbundle", target: "#{ENV['HOME']}/Library/Mail/Bundles/Free-GPGMail_6.mailbundle"
+    uninstall delete: "~/Library/Mail/Bundles/Free-GPGMail_6.mailbundle"
+  end
+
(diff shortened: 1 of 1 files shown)


Casks/free-gpgmail.rb
```diff
@@ -0,0 +1,41 @@
+cask "free-gpgmail" do
+  if MacOS.version <= :catalina
+    version "5,2021.3.1,-unsigned"
+    sha256 "3a3c9290622cc3fe8d4761050363917d009f9fa923eb32d436364f0ae9a161b9"
+  elsif MacOS.version <= :big_sur
+    version "5,2021.3.1,-signed"
+    sha256 "ee0b1a313afedb5cdf8a9703b00f35040b3109877247bb342c92a60b53944c9a"
+  elsif MacOS.version >= :monterey
+    version "6,2021.3.1,"
```
jpbernius, line 9: Is there a good way of live checking only part of the version?
The `:github_latest` strategy can produce `version.csv.second` easily.
If this value changes, manual work is needed to determine which version is now compatible with each OS based on the table in the README/Release Notes.

Casks/free-gpgmail.rb
```diff
@@ -0,0 +1,51 @@
+cask "free-gpgmail" do
+  if MacOS.version <= :catalina
+    version "5,2021.3.1,-unsigned"
```
SMillerDev, line 3: Can't this use the signed version?

Casks/free-gpgmail.rb
SMillerDev: There is a regex method in the livecheck block.

Casks/free-gpgmail.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  artifact "Free-GPGMail_#{version.csv.first}.mailbundle", target: "#{ENV["HOME"]}/Library/Mail/Bundles/Free-GPGMail_#{version.csv.first}.mailbundle"
+
+  preflight do
+    system "rm", "#{ENV["HOME"]}/Library/Mail/Bundles/Free-GPGMail_*.mailbundle"
+  end
+
+  uninstall delete: "~/Library/Mail/Bundles/Free-GPGMail_#{version.csv.first}.mailbundle"
+
+  caveats "Apple Mail plugin for GnuPG has been updated."
```
SMillerDev, line 40: Unless this is specific to the homebrew install it shouldn't be a caveat.

Casks/free-gpgmail.rb
jpbernius: According to the [README](https://github.com/Free-GPGMail/Free-GPGMail/blob/master/ReadMe.md), Mojave and Catalina should use the unsigned version. I personally only run Monterey, so I cannot verify this and trust the README.
image

Casks/free-gpgmail.rb
jpbernius: I've come up with this:
```rb
  livecheck do
    url :url
    strategy :git do |tags|
      tags.map do |tag|
        match = tag.match(/^v(\d+(?:\.\d+)*)/i)
        "#{version.csv.first},#{match[1]},#{version.csv.third}" if match
      end.compact
    end
  end
```
is this the way to go, especially the use of `#{version.csv.first}`?

Casks/free-gpgmail.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  auto_updates false
+  depends_on macos: ">= :mojave"
+  depends_on cask: "gpg-suite-no-mail"
+
+  artifact "Free-GPGMail_#{version.csv.first}.mailbundle", target: "#{ENV["HOME"]}/Library/Mail/Bundles/Free-GPGMail_#{version.csv.first}.mailbundle"
+
+  preflight do
+    system "rm", "#{ENV["HOME"]}/Library/Mail/Bundles/Free-GPGMail_*.mailbundle"
```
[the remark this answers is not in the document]
jpbernius: Removed the preflight block.

Casks/free-gpgmail.rb
jpbernius: Removed the caveats.

jpbernius:
Thanks for the quick review @SMillerDev!

Casks/free-gpgmail.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  livecheck do
+    url :url
+    strategy :git do |tags|
+      tags.map do |tag|
+        match = tag.match(/^v(\d+(?:\.\d+)*)/i)
+        "#{version.csv.first},#{match[1]},#{version.csv.third}" if match
+      end.compact
+    end
+  end
```
samford, line 26:
```suggestion
  # This restricts matching to new releases that use the same major as the
  # cask `version` (based on the execution environment). As such, this won't
  # surface a new major version and that will need to be handled manually.
  livecheck do
    url :url
    regex(/href=.*?Free-GPGMail[._-]v?(\d+)-(\d+(?:\.\d+)+)(-[^"' >]+?)?[._-]mailbundle\.zip/i)
    strategy :github_latest do |page, regex|
      page.scan(regex).map do |match|
        next if match[0] != version.csv.first

        if match[2].present?
          next if version.csv.third.present? && match[2] != version.csv.third

          "#{match[0]},#{match[1]},#{match[2]}"
        else
          "#{match[0]},#{match[1]}"
        end
      end
    end
  end
```

Checking the Git tags and combining it with parts of the `version` will be a problem if a given release doesn't provide a file for a given version (e.g., `6`). It's better not to assume, as we can simply use the `GithubLatest` strategy and match version information from the download files for the release.

So long as we restrict matching to the major version (with respect to the environment where `brew livecheck` is run), we won't be able to detect when a new major version appears. However, based on how releases are done, we should still get a new version for a previous major version and folks will just need to make sure to check the release before creating a version bump PR.

To keep the regex sane, this suggestion restricts matching to a given major (e.g., `6`) and suffix (e.g., `-unsigned`) in the `strategy` block. The regex is intended to match all the existing formats and is relatively straightforward compared to what we would have if we used the regex to restrict matches in this context (i.e., creating a long regex that uses `Regexp.escape(version.csv.first)` and `Regexp.escape(version.csv.third || '')`).

---

Looking to the future, this would benefit from being able to specify the version to use for comparison within livecheck. If we only used `version.csv.second`, we could potentially simplify this drastically (checking the Git tags may be fine in that context). I know of at least one other cask that would benefit from this, so it's been on my to-do list for a while but it will be a bit before I get to it. The suggestion above should work in the interim time, at least.

jpbernius:
I checked and merged the `livecheck` from @samford. I hope this is ready for merge now.

PR #121022, opened 2022-03-26
Add Espanso 2.1.4

lkrms:
The developer has explained the beta tag in v2.1.4's release notes:

> With this release, the v2 will officially become the stable version,
> while version 0.7.3 (legacy version) will be deprecated. This
> decision comes after months of testing, and we're confident the v2 is
> ready for general adoption!

Further to the above, it is my understanding that this release meets the "But There Is No Stable Version!" criteria, and I can personally attest that despite officially being a "beta" product since I started relying on it in late 2019, Espanso has been more stable than most of the "stable" software I use.

### DIFF
--- a/Casks/espanso.rb
+++ b/Casks/espanso.rb
@@ -1,0 +1,39 @@
+cask "espanso" do
+  arch = Hardware::CPU.intel? ? "Intel" : "M1"
+
+  version "2.1.5"
+
+  if Hardware::CPU.intel?
+    sha256 "6058ea8b851073ab97a6ffd257127ef7cfd367ccc6c290648217b3b657319539"
+  else
+    sha256 "8e71bf8904f3418ddc54c8faeb9e1f40fe4a57bef81d6c93468dbf0c55bcb210"
+  end
+
+  url "https://github.com/federico-terzi/espanso/releases/download/v#{version}-beta/Espanso-Mac-#{arch}.zip",
+      verified: "github.com/federico-terzi/espanso/"
+  name "Espanso"
+  desc "Cross-platform Text Expander written in Rust"
+  homepage "https://espanso.org/"
+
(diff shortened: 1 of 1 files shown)


Casks/espanso.rb
```diff
@@ -0,0 +1,41 @@
...
+  homepage "https://espanso.org/"
+
+  livecheck do
+    url "https://github.com/federico-terzi/espanso/releases/latest"
+    strategy :page_match
+    regex(%r{href=.*?/v(\d+(?:\.\d+)*)(?:-beta)?/Espanso-Mac-(?:Intel|M1)\.zip}i)
+  end
+
+  conflicts_with formula: "espanso"
+
```
bevanjkay, line 25: @lkrms There is no formula with this token, hence why CI is currently failing.

bevanjkay (changes requested):

Casks/espanso.rb
lkrms: Thanks for letting me know. This cask would deprecate a legacy formula provided by the developer in their own tap. Would including the tap prefix work here, or should I just remove the directive?

lkrms: After finding existing `conflicts_with` declarations that used explicit tap references, I did the same, updated the cask to 2.1.5 and rebased. Hopefully CI will pass now. Thanks again.

lkrms: Third-party taps don't seem to be supported by CI, so I've removed `conflicts_with` entirely. The cookbook indicates it hasn't been implemented for casks yet anyway, so this isn't much of a loss.

lkrms: Hi @bevanjkay, just checking in to see if there's something more I should be doing here? Thank you.

bevanjkay (approved):

PR #121314, opened 2022-04-01
ganttproject 3.2.3230 (botched upstream release)

domq:
It looks like the upstream maintainer was not thorough in re-releasing 3.2.3200.
- Add `version_of_github_release` to paper over this
- Use it to construct `url`
- Update sha256

- [X] `brew audit --cask ganttproject` is error-free.


domq:
The CI checks detect the same problem that I'm trying to work around. Please merge regardless.

bevanjkay:
@domq We should not be merging changes when the CI checks are detecting problems, they are there for a reason. In this instance we can detect the correct version by adjusting the `livecheck` stanza.

PR #121365, opened 2022-04-02
Add Accord.app v1.0.2

mariobrostech:

### DIFF
--- a/Casks/accord.rb
+++ b/Casks/accord.rb
@@ -10,7 +10,6 @@ cask "accord" do
   livecheck do
     url :url
     strategy :github_latest
-    regex(%r{href=.*?/tag/v?([\w._-]+)["' >]}i)
   end
 
   depends_on macos: ">= :big_sur"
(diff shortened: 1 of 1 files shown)


Casks/accord.rb
```diff
@@ -0,0 +1,24 @@
...
+  name "accord"
+  desc "Discord client written in Swift for modern Macs"
+  homepage "https://github.com/evelyneee/accord"
+
+  livecheck do
+    url :url
+    strategy :github_latest
+    regex(%r{href=.*?/tag/v?([\w._-]+)["' >]}i)
+  end
+
```
bevanjkay, line 15: @mariobrostech The `livecheck` stanza should be able to be omitted here as the default `:git` strategy will be automatically inferred from the url.

bevanjkay (changes requested):

Casks/accord.rb
mariobrostech: I tried without having the `livecheck` stanza initially, but the audit failed since it was expecting version `1.0-2` instead of `1.0.2`, as it should be. Is there a way to make it have the correct version number without using the `livecheck` stanza? That's the only reason I included it.

bevanjkay: Thanks for checking @mariobrostech 
In that case, just the URL and strategy without the `regex` should suffice. I have pushed the change already 👍🏻

mariobrostech: Awesome, thank you!

bevanjkay (approved):

PR #121473, opened 2022-04-04
Add PolyMC.app v1.1.1

kthchew:

### DIFF
--- a/Casks/polymc.rb
+++ b/Casks/polymc.rb
@@ -1,0 +1,25 @@
+cask "polymc" do
+  version "1.1.1"
+  sha256 "5f40a9db742b1910825f88980c24e5bb40673b127e5794a4d09729356e5a9778"
+
+  url "https://github.com/PolyMC/PolyMC/releases/download/#{version}/PolyMC-macOS-#{version}.tar.gz",
+      verified: "github.com/PolyMC/PolyMC/"
+  name "PolyMC"
+  desc "Minecraft launcher"
+  homepage "https://polymc.org/"
+
+  livecheck do
+    url :url
+  end
+
+  depends_on macos: ">= :sierra"
+
+  app "PolyMC.app"
+
+  zap trash: [
+    "~/Library/Application Support/PolyMC/metacache",
+    "~/Library/Application Support/PolyMC/PolyMC-*.log",
+    "~/Library/Application Support/PolyMC/polymc.cfg",
+    "~/Library/Saved Application State/org.polymc.PolyMC.savedState",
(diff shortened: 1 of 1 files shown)


Casks/polymc.rb
```diff
@@ -0,0 +1,25 @@
...
+  url "https://github.com/PolyMC/PolyMC/releases/download/#{version}/PolyMC-macOS-#{version}.tar.gz",
+      verified: "github.com/PolyMC/PolyMC/"
+  name "PolyMC"
+  desc "Minecraft launcher"
+  homepage "https://polymc.org/"
+
+  livecheck do
+    url :url
+  end
+
```
bevanjkay, line 14: @kthchew This can be omitted as it will be automatically inferred from the url.

Casks/polymc.rb
```diff
@@ -0,0 +1,25 @@
...
+  depends_on macos: ">= :sierra"
+
+  app "PolyMC.app"
+
+  zap trash: [
+    "~/Library/Application Support/PolyMC/metacache",
+    "~/Library/Application Support/PolyMC/PolyMC-*.log",
+    "~/Library/Application Support/PolyMC/polymc.cfg",
```
bevanjkay, line 22: @kthchew Can we trash `~/Library/Application Support/PolyMC` instead?

bevanjkay (changes requested):

Casks/polymc.rb
kthchew: Done, thanks.

Casks/polymc.rb
kthchew: I don't think so. That directory contains user instances (game data, world files, etc.) in `~/Library/Application Support/PolyMC/instances` that are directly created by the user. AFAIK those shouldn't be deleted by a zap. There's also other directories in there that usually contain auto-generated files, but may have been modified or created by a user manually, so I wanted to be safe and not delete them.

kthchew: For reference when I open PolyMC for the first time and then create an instance, I see the following in `~/Library/Application Support/PolyMC`:

```
icons (directory, contains image files a user places there)
instances (directory, contains user data)
PolyMC-0.log (file)
polymc.cfg (file)
themes (directory, contains a default theme and/or user customizations)
translations (directory)
meta (directory, contains info about game/library versions but these can be edited by the user if they need something custom)
metacache (file)
```

bevanjkay (approved):

bevanjkay:
Thank you @kthchew

PR #121925, opened 2022-04-13
[google-chat-electron] add apple silicon support

AdamCoulterOz:

### DIFF
--- a/Casks/google-chat-electron.rb
+++ b/Casks/google-chat-electron.rb
@@ -1,8 +1,15 @@
 cask "google-chat-electron" do
   version "2.15.0"
-  sha256 "a757dc557990f32a9be8b21fd81995748e8a2ecc56c772681a3cb427000986fb"
 
-  url "https://github.com/ankurk91/google-chat-electron/releases/download/#{version}/google-chat-electron-#{version}-darwin-x64.zip"
+  if Hardware::CPU.intel?
+    sha256 "a757dc557990f32a9be8b21fd81995748e8a2ecc56c772681a3cb427000986fb"
+    arch = "darwin-x64"
+  else
+    sha256 "aadb3e0db0053330fa0543556781d28dc6ca4e32e0a5060b18778e6d5c346cbd"
+    arch = "darwin-arm64"
+  end
+
+  url "https://github.com/ankurk91/google-chat-electron/releases/download/#{version}/google-chat-electron-#{version}-#{arch}.zip"
   name "google-chat-electron"
   desc "Standalone app for Google Chat"
   homepage "https://github.com/ankurk91/google-chat-electron"
--- a/Casks/google-chat-electron.rb
+++ b/Casks/google-chat-electron.rb
@@ -23,5 +23,4 @@ cask "google-chat-electron" do
     "~/Library/Preferences/com.electron.google-chat-electron.plist",
     "~/Library/Saved Application State/com.electron.google-chat-electron.savedState",
   ]
-
 end
--- a/Casks/google-chat-electron.rb
+++ b/Casks/google-chat-electron.rb
@@ -23,4 +23,10 @@ cask "google-chat-electron" do
     "~/Library/Preferences/com.electron.google-chat-electron.plist",
     "~/Library/Saved Application State/com.electron.google-chat-electron.savedState",
   ]
+
+  caveats <<~EOS
+    If Apple Silicon macOS says the package is corrupted or not trusted, remove app quarantine:
+    - Run `xattr -d com.apple.quarantine /Applications/google-chat-electron.app`
+    - or; install using `brew install --cask --no-quarantine google-chat-electron`
+  EOS
 end
--- a/Casks/google-chat-electron.rb
+++ b/Casks/google-chat-electron.rb
@@ -24,12 +24,4 @@ cask "google-chat-electron" do
     "~/Library/Saved Application State/com.electron.google-chat-electron.savedState",
   ]
 
-  caveats <<~EOS
-    According to the vendor (https://github.com/ankurk91/google-chat-electron#installation-mac),
-    the quarantine attribute breaks the app and needs to be removed. To do so, install this cask with:
-
-      brew install --cask --no-quarantine #{token}
-
-    WARNING: quarantining is a macOS security feature and you disable it at your own risk!
-  EOS
 end
--- a/Casks/google-chat-electron.rb
+++ b/Casks/google-chat-electron.rb
@@ -25,9 +25,10 @@ cask "google-chat-electron" do
   ]
 
   caveats <<~EOS
-    The quaratine attribute breaks the app (macOS will say the package is corrupted or not trusted), to remove it:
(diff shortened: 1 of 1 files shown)


bevanjkay:
@AdamCoulterOz We should not be encouraging users to remove the quarantine attribute in this way as Homebrew goes to great lengths to ensure that downloaded files are quarantined.

Here is an example of an alternative caveat that could apply - https://github.com/Homebrew/homebrew-cask/pull/98527/files#diff-df256926e1a55a622db61a0ff695bd5da69b8edafef0dacc43f31c868281eb21R18-R25

Casks/google-chat-electron.rb
```diff
@@ -25,8 +25,10 @@
...
 
   caveats <<~EOS
-    If Apple Silicon macOS says the package is corrupted or not trusted, remove app quarantine:
-    - Run `xattr -d com.apple.quarantine /Applications/google-chat-electron.app`
-    - or; install using `brew install --cask --no-quarantine google-chat-electron`
+    The quaratine attribute breaks the app (macOS will say the package is corrupted or not trusted), to do so:
+      - install this cask with `brew install --cask --no-quarantine #{token}`
+      - or; run `xattr -d com.apple.quarantine /Applications/#{token}.app` after the install
+
+    WARNING: quarantining is a macOS security feature and you disable it at your own risk!
```
bevanjkay, line 32: @AdamCoulterOz Please use the same structure as the referenced commit - https://github.com/Homebrew/homebrew-cask/pull/98527/files#diff-df256926e1a55a622db61a0ff695bd5da69b8edafef0dacc43f31c868281eb21R18-R25

1. Include a link to documentation upstream
2. Do not suggest running the `xattr -d com.apple.quarantine` command.

bevanjkay (changes requested):

Casks/google-chat-electron.rb
AdamCoulterOz: done

bevanjkay (dismissed):

Casks/google-chat-electron.rb
```diff
@@ -16,4 +23,13 @@
     "~/Library/Preferences/com.electron.google-chat-electron.plist",
     "~/Library/Saved Application State/com.electron.google-chat-electron.savedState",
   ]
+
+  caveats <<~EOS
+    According to the vendor (https://github.com/ankurk91/google-chat-electron#installation-mac),
```
SMillerDev, line 28: The issue the developer is describing here is documented in our FAQ: https://docs.brew.sh/FAQ#why-cant-i-open-a-mac-app-from-an-unidentified-developer and doesn't actually need the quarantine removed. This caveat is therefore not needed.

SMillerDev (changes requested):

Casks/google-chat-electron.rb
AdamCoulterOz: @SMillerDev not quite what's happening in this case, macOS doesn't say it's from an unknown developer, it says the package is corrupted.

AdamCoulterOz: Also my I suggest that having a user need to go look at the FAQ in order to install an app where this is probably the most common problem with all cask installs isn't a great user experience. Perhaps add an additional field (or two, if you also want the developer doc reference) to the cask schema which can be marked true if the cask requires quarantine removal in order to function correctly. This way you can standardise the message to users and prompt them to look at the FAQ or whatever.

SMillerDev: Users don't need to be reminded of documentation on every install though, once they know how to solve this issue they can use that knowledge everywhere.

Additionally, re-reading the upstream documentation there don't seem to be any issues with `brew install --cask` so that would suggest this isn't needed at all.

AdamCoulterOz: > Users don't need to be reminded of documentation on every install though, once they know how to solve this issue they can use that knowledge everywhere.

I'm not suggesting printing out the documentation in the console, just a flag that informs them of the additional step they need to take to make it work.

> Additionally, re-reading the upstream documentation there don't seem to be any issues with `brew install --cask` so that would suggest this isn't needed at all.

No idea what you are talking about. I've just spent the last 3 hours testing this, it won't open without changing the attribute. Either it will say `unknown developer` if it's running on `darwin_x64` or `package corrupted` if its on `darwin_arm64`.

AdamCoulterOz: Screen Shot 2022-04-13 at 5 52 59 pm

Screen Shot 2022-04-13 at 5 53 05 pm

SMillerDev: > No idea what you are talking about. This is the line in the instructions in the upstream link I provided:

That line is for manual downloads though, not for the `brew cask` install method.

> The vendor has been getting heaps of people reporting the package corrupted issue and there needs to be a prompt for them.

All 3 of the reports on GitHub are for manual downloads though, not for brew.

AdamCoulterOz: This rb file _is_ the definition of the brew install script. The issue of corrupted package is new as of this release which now includes darwin_arm64 for the first time. Brew casks _are a wrapper_ around the vendors manual install method. So the manual install problems _are the only relevant reference_.

We are meeting the homebrew interface contract for publishing a brew cask. You should _only_ be manually checking we've met that where you can't automate it.

AdamCoulterOz: Also, how is that logical? Is there some magical unicorns who make the brew script running _the exact same command and package as the manual method_ make it somehow work while quarantined?

AdamCoulterOz: > The issue the developer is describing here is documented in our FAQ: https://docs.brew.sh/FAQ#why-cant-i-open-a-mac-app-from-an-unidentified-developer and doesn't actually need the quarantine removed. This caveat is therefore not needed.

I have proven this is not the case with my screenshots, and you've now moved the goal posts. Can you please qualify what you are trying to achieve by blocking this?

SMillerDev: > Can you please qualify what you are trying to achieve by blocking this?

Make sure bad security advice given by brew only happens when it is absolutely required.

---

After downloading the binary to check the signature:
```
$ codesign -v -f  google-chat-electron.app
> google-chat-electron.app: code has no resources but signature indicates they must be present
```

So it seems it's actually incorrectly signed and the command only disables the check Apple does before opening apps.

AdamCoulterOz: Thanks for checking that @SMillerDev, I can follow up with the vendor now.

https://github.com/ankurk91/google-chat-electron/issues/85#issuecomment-1099881091

AdamCoulterOz: @SMillerDev I've removed the caveat regarding quarantine completely now, but there should still be a better user experience for unknown developers than this.

SMillerDev (approved):
Thanks @AdamCoulterOz  ! Without contributions like yours it'd be impossible to keep homebrew going with the high standards that users have come to expect from the project. You can feel good knowing that you've made the world a tiny bit better for homebrew users around the world! 👍 🎉

PR #122019, opened 2022-04-14
Rebranding Ferdi to Ferdium

amangalampalli:

- [X ] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [X ] `brew audit --cask <cask>` is error-free.
- [X ] `brew style --fix <cask>` reports no offenses.

### DIFF
--- a/Casks/ferdium.rb
+++ b/Casks/ferdium.rb
@@ -1,4 +1,4 @@
-cask "ferdi" do
+cask "ferdium" do
   arch = Hardware::CPU.intel? ? "" : "-arm64"
 
   version "5.8.1"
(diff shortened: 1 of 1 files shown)


otkd:
This should be marked as a draft until the uncertainty regarding the status of the project is resolved

bevanjkay:
Given that the vendor is changing, **once/if there is some more stability in the project**, it would be more appropriate to add this as a new cask, and we should remove the old cask in a separate PR if it is indeed confirmed as being unsupported upstream.

Alpheron:
@bevanjkay This change was supposed to be seen as temporary as since the vendor is changing, the previous Github repositories no longer exist. So for now, it was supposed to be a temporary change to host the project for people to download and then change the repository in the future when the next version is released. This also means that the existing `ferdi.rb` formula no longer exists since the initial repository was deleted. Should I submit a PR to remove the initial `ferdi.rb` formula?

PR #122077, opened 2022-04-16
Add wpsoffice-cn 4.2.0

wjlmelody:

### DIFF
--- a/Casks/wpsoffice-cn.rb
+++ b/Casks/wpsoffice-cn.rb
@@ -1,5 +1,6 @@
 cask "wpsoffice-cn" do
   version "4.2.0,6717"
+  
   arch = Hardware::CPU.intel? ? "x64" : "arm64"
 
   if Hardware::CPU.intel?
--- a/Casks/wpsoffice-cn.rb
+++ b/Casks/wpsoffice-cn.rb
@@ -25,7 +25,7 @@ cask "wpsoffice-cn" do
     end
   end
 
-  conflicts_with cask: "wps-office"
+  conflicts_with cask: "wpsoffice"
   depends_on macos: ">= :sierra"
 
   app "wpsoffice.app"
--- a/Casks/wpsoffice-cn.rb
+++ b/Casks/wpsoffice-cn.rb
@@ -20,11 +20,11 @@ cask "wpsoffice-cn" do
   livecheck do
     url :homepage
     strategy :page_match do |page|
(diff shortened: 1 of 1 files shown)


Casks/wpsoffice-cn.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  name "WPS Office"
+  desc "All-in-one office service platform in Chinese"
+  homepage "https://mac.wps.cn/"
+
+  livecheck do
+    url "https://mac.wps.cn/"
+    strategy :page_match do |page|
+      v = page.match(%r{(\d+\.\d+\.\d+)(?=\(\d+\)/)}i)
+      id = page.match(%r{\d+(?=\)/)}i)
```
bevanjkay, line 24: @wjlmelody Can these regexes be updated so that they are anchored to something static on the page, if the content of the page changes there could easily be false positives for these.

wjlmelody: The latest software version displayed by this web page is relatively simple that there's no need to update regexes. In addition, I failed to find any other web pages that can be used by livecheck.

bevanjkay: @wjlmelody Please have a read of the [livecheck docs](https://docs.brew.sh/Brew-Livecheck#regex-guidelines) as there are a couple of points that are relevant here. If you need further help let us know.

Casks/wpsoffice-cn.rb
```diff
@@ -18,14 +18,8 @@
...
-    url "https://mac.wps.cn/"
-    strategy :page_match do |page|
-      v = page.match(/data[._-]href=["']?.*?WPS_Office[._-](\d+(?:\.\d+)+)(?=\(\d+\))/i)
-      id = page.match(%r{\d+(?=\)/)}i)
-      next if v.blank? || id.blank?
-
-      "#{v},#{id}"
-    end
+    url :url
+    strategy :extract_plist
```
bevanjkay, line 22: @wjlmelody This strategy is only useful when the `url` is unversioned. Because this cask uses a versioned URL this strategy will always return the current version because it checks `.plist` files in the download.

wjlmelody: All right

wjlmelody:
I don't know how to fix this error "Use // around regular expression". @bevanjkay

bevanjkay:
@wjlmelody I believe this should be good to go now. Thanks for your hard work on this one.

wjlmelody:
> @wjlmelody I believe this should be good to go now. Thanks for your hard work on this one.

You're welcome.

PR #122081, opened 2022-04-16
Add i4tools 1.13.010

wjlmelody:

### DIFF
--- a/Casks/i4tools.rb
+++ b/Casks/i4tools.rb
@@ -9,7 +9,9 @@ cask "i4tools" do
   homepage "https://www.i4.cn/"
 
   livecheck do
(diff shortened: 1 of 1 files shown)


Loyalsoldier:
According to https://github.com/Homebrew/homebrew-cask/pull/69182 , this cask contained malware. Is it secure now?

ivaquero:
This software can be used to jailbreak and change virtual location of iPhone, I guess this is the reason for raising malware warning.

bevanjkay:
@Loyalsoldier @ivaquero Thank you for picking this up. If the cask was removed previously for security reasons then we shouldn't be adding it back in. I will revert the commit.

PR #122097, opened 2022-04-17
Update qiyimedia from 13.3.5 to 13.4.0

wjlmelody:

### DIFF
--- a/Casks/qiyimedia.rb
+++ b/Casks/qiyimedia.rb
@@ -9,8 +9,7 @@ cask "qiyimedia" do
   homepage "https://app.iqiyi.com/mac/player/index.html"
 
   livecheck do
-    url "https://app.iqiyi.com/mac/player/index.html"
-    regex(/最新版本：(\d+(?:\.\d+)+)/i)
+    skip "No latest version information available"
   end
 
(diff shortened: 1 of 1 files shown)


Casks/qiyimedia.rb
```diff
@@ -9,19 +9,15 @@
...
 
   depends_on macos: ">= :catalina"
 
   app "爱奇艺.app"
 
   zap trash: [
-    "~/Library/Application Scripts/com.iqiyi.player.QYWidget",
     "~/Library/Application Scripts/com.iqiyi.player",
-    "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.iqiyi.player.sfl*",
-    "~/Library/Containers/com.iqiyi.player.QYWidget",
```
bevanjkay, line 24: @wjlmelody Are you certain that these files should be removed from the zap? They may have just not been created yet on your machine - there's no harm in leaving files from older versions in the zap stanza.

wjlmelody: Under your suggestion, I realized that there may be a mistake in deleting these files. I have put these files back to the zap.

PR #122408, opened 2022-04-23
Fix URL for ZEsarUX release & add desc to cask file

floehopper:
It appears that the URL has changed slightly and this was preventing me from installing the cask.

I've also added a `desc` string to the cask file as recommended by `brew audit`.

I confirm that I have checked the following:


floehopper:
Sorry that the build failed for this PR initially - I think I've now fixed everything!

bevanjkay:
Thank you @floehopper

PR #122379, opened 2022-04-22
Update r from 4.1.3 to 4.2.0

noamross:
Created with `brew bump-cask-pr`.

PR #122414, opened 2022-04-23
Fix for updating r from 4.1.3 to 4.2.0

noamross:
This fixes a mistake I made in https://github.com/Homebrew/homebrew-cask/pull/122379. I had changed the livecheck URL, but now realize that this URL can point to binaries built but not yet signed, and the delay in updating the original, correct livecheck URL is in signing the binaries. This PR reverts to the original livecheck URL and updates the hashes to the now-signed updated binaries for 4.2.0


chenrui333:
fixes https://github.com/orgs/Homebrew/discussions/3208

PR #122521, opened 2022-04-26
Kitty: Add link to terminfo

henryhchchc:

Using the [cask code for alacritty](https://github.com/Homebrew/homebrew-cask/blob/fc6f5981818e7eb1d9215e623b4cc5d58494c860/Casks/alacritty.rb#L23) as a reference

### DIFF
--- a/Casks/kitty.rb
+++ b/Casks/kitty.rb
@@ -13,6 +13,8 @@ cask "kitty" do
   # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
   shimscript = "#{staged_path}/kitty.wrapper.sh"
   binary shimscript, target: "kitty"
+  binary "#{appdir}/kitty.app/Contents/Resources/terminfo/78/xterm-kitty",
+         target: "#{ENV.fetch("TERMINFO", "#{ENV["HOME"]}/.terminfo")}/78/xterm-kitty"
 
   preflight do
     File.write shimscript, <<~EOS


miccal (approved):

miccal:
Thank you @henryhchchc.

mitchblank:
I am not sure why this was added, but it seems to break cask upgrade for me.

When running a shell under kitty itself, it sets `$TERMINFO` to point inside of its /Application directory.  So if you then run `brew upgrade` from inside of a kitty shell, this causes it to try to create a symlink into itself:

```
==> Upgrading 1 outdated package:
kitty 0.25.0 -> 0.25.1
==> Upgrading kitty
[...]
==> Moving App 'kitty.app' to '/Applications/kitty.app'
==> Linking Binary 'kitty.wrapper.sh' to '/usr/local/bin/kitty'
Error: kitty: It seems there is already a Binary at '/Applications/kitty.app/Contents/Resources/kitty/terminfo/78/xterm-kitty'.
```

I was able to upgrade kitty successfully by running `brew upgrade` from inside of an iTerm2 window instead:

```
==> Moving App 'kitty.app' to '/Applications/kitty.app'
==> Linking Binary 'kitty.wrapper.sh' to '/usr/local/bin/kitty'
==> Linking Binary 'xterm-kitty' to '/Users/mitch/.terminfo/78/xterm-kitty'
```

In the past I haven't had any issue upgrading kitty even from kitty itself.  Obviously I didn't get the new version until I quit and relaunched, but it never caused any problems for me even if it's a little dubious swapping out all of the files in `/Application/kitty.app` from underneath its feet.

henryhchchc:
It is added based on https://sw.kovidgoyal.net/kitty/faq/#keys-such-as-arrow-keys-backspace-delete-home-end-etc-do-not-work-when-using-su-or-sudo

It broke the upgrade for me as well. Given that the `TERMINFO` is already set. We may no longer need this link.

mitchblank:
> We may no longer need this link.

FWIW, it seems that kitty has been setting `$TERMINFO` since 2016 https://github.com/kovidgoyal/kitty/commit/cca33a555040cd93aa52d9b0d0b529836a4db3d7

mitchblank:
I guess it *may* make sense to drop a symlink into `~/.terminfo` so that "xterm-kitty" is available even if `$TERMINFO` isn't set (as can happen inside a `su` or `sudo` session, per that link you provided.

What *isn't* a good idea is to decide where to put the link based on the current value of `$TERMINFO`, since that just produces the circular link when running as non-root inside of kitty.

PR #122621, opened 2022-04-28
Added vscodium arm64

vineryap:

### DIFF
--- a/Casks/vscodium.rb
+++ b/Casks/vscodium.rb
@@ -1,8 +1,10 @@
 cask "vscodium" do
+  arch = Hardware::CPU.intel? ? "x64" : "arm64"
+
   version "1.66.2"
   sha256 "7163ee27353cba1d8477da3a39001db0c67a71bef19799af3e16ff9367c93e10"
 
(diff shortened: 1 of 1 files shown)


Casks/vscodium.rb
```diff
@@ -1,8 +1,10 @@
 cask "vscodium" do
+  arch = Hardware::CPU.intel? ? "x64" : "arm64"
+
   version "1.66.2"
   sha256 "7163ee27353cba1d8477da3a39001db0c67a71bef19799af3e16ff9367c93e10"
```
bevanjkay, line 5: @vineryap The `sha256` will need to be inside of an `if/else` block as it will be different for the two architectures.

See the `visual-studio-code` cask as an example https://github.com/Homebrew/homebrew-cask/commit/269c16d642b5cc5d7619018819836fe6137fc315#diff-bd913c80ea9d950602a99afcadb61adf132f96a1731b57505ecca52d72e6063cR9-R12

vineryap: Thank you for the guide!
I've added the `sha256`.

bevanjkay (approved):

PR #122757, opened 2022-04-30
Update xemu formulae to v0.6.5

mavethee:
Since 0.6.5 replaced gh-release/* with xemu-v*, I updated ruby file for new Xemu releases.


mavethee:
So brew audit works, reinstalling cask with local rb goes without issues, I guess thats it. No idea about previously encoded url, but without encode works as well.

bevanjkay:
@mavethee Can you please remove the livecheck block. It will be automatically  inferred from the url.

mavethee:
> @mavethee Can you please remove the livecheck block. It will be automatically inferred from the url.

Done!

bevanjkay (approved):

PR #123789, opened 2022-05-15
Fix Cave Story app, name stanza

inoue22:

---

Some notes:

Different app names are set for the English version and the Japanese version.
So I separated app stanza into English and Japanese.
Before this change, the Japanese version of app name was not consistent.
I also removed name stanza "Doukutsu", because it's not app's official name.

### DIFF
--- a/Casks/cave-story.rb
+++ b/Casks/cave-story.rb
@@ -46,6 +46,7 @@ cask "cave-story" do
   end
 
   name "Cave Story"
+  name "Doukutsu"
   name "洞窟物語"
   desc "Action-adventure game reminiscent of classic 8- and 16-bit games"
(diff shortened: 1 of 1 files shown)


bevanjkay:
Is "Doukutsu" a name for the application though? Casks can have multiple names, and it looks like that is how the `app` artifact is named?

toco8:
@bevanjkay Thank you very much for reviewing my PR.
"Doukutsu" is a name for the `app` artifact.
I understood it needs to be registered.
I reverted my change. (remove "Doukutsu" line)
I thought that part of "Doukutsu Monogatari" was omitted.

miccal (approved):

PR #123807, opened 2022-05-16
publii: update livecheck

samford:

-----

The existing `livecheck` block for `publii` checks the latest release from GitHub but the cask doesn't use any files from the repository, so the check isn't aligned with the source of the cask `url`. This PR updates the `livecheck` block to check the first-party download page, which links to the dmg file used in the URL.

### DIFF
--- a/Casks/publii.rb
+++ b/Casks/publii.rb
@@ -14,13 +14,8 @@ cask "publii" do
   homepage "https://getpublii.com/"
 
   livecheck do
-    url "https://github.com/GetPublii/Publii/releases/latest"
-    strategy :page_match do |page|
-      match = page.match(%r{href=.*?/v\.?(\d+(?:\.\d+)+).zip}i)
-      next if match.blank?
-
-      match[1].to_s
-    end
+    url "https://getpublii.com/download/"
+    regex(/href=.*?Publii[._-]v?(\d+(?:\.\d+)+)(?:[._-]#{arch})?\.dmg/i)
   end
 
   app "Publii.app"


bevanjkay:
@samford We probably shouldn't run the livecheck updates with the `ci-syntax-only` tag as it doesn't run livecheck as part of the checks. The CI overhead is much smaller in the `homebrew-cask-*` repos than `homebrew-core` so there it's fine to run the full checks.

bevanjkay (approved):

miccal (approved):

samford:
@bevanjkay This is an acknowledged shortcoming of using `CI-syntax-only` for `livecheck` PRs and we've just lived with it in homebrew-core because the overhead is otherwise unnecessary and often steep. The ideal would be something like syntax-only with the addition of running livecheck but I haven't yet investigated how that could work (e.g., modifying test-bot to have an `--only-livecheck` option). In the interim time, I manually test `livecheck` blocks locally as part of reviewing PRs, though I've seen two formulae where a check worked locally but failed only on CI and that type of situation would be missed until the next normal CI run.

Part of the reason behind using `CI-syntax-only` for `livecheck` PRs in cask repositories is because `Cask::Audit#check_livecheck_version` will error if the livecheck version is newer than the current cask version (i.e., it requires a `version` bump to pass). It's not uncommon for a cask to have a new version available when I'm adding/updating a `livecheck` block but my experience with casks is limited and I usually don't have extra time at the moment to update a cask and handle any issues that arise. It's fine when a version bump is very simple and painless but it can become a time sink when it's not. Either way, I would prefer for a version bump to be optional in a `livecheck` block PR (i.e., it's better suited for a follow-up PR).

Past that, I'm not fond of the idea of unnecessarily downloading application data when we don't need to, as some casks involve files that are hundreds of MB (or even more than a GB). homebrew-cask CI doesn't have to compile applications like homebrew-core CI but there's still unnecessary overhead involved. I don't feel that it's a worthwhile tradeoff just to run livecheck on CI when 99% of the time we would see the same result as running `brew livecheck` locally. It _is_ more convenient for a reviewer not to have to manually apply and test the changes locally but I don't think the benefit justifies the cost at the moment.

That said, we can try this out if you feel strongly about it but I think we're better off continuing to use `CI-syntax-only` until there's something appropriate like `CI-livecheck-only`. At the very least, `CI-syntax-only` may still be necessary for bulk PRs (I'm working on one that will modify hundreds of `livecheck` blocks), as one flaky check in the bunch would cause CI to fail.

-----

Looking to the future, I have some stashed work that I need to revisit to add a `--skip-livecheck` option in test-bot (allowing CI to pass in very exceptional circumstances when a `livecheck` block only fails on CI and we don't have a fix or it requires changes that will take time and shouldn't hold up a core/cask PR). I'll try to find some time to revisit this and look into adding an `--only-livecheck` option as well (though it may be tricky to achieve).

bevanjkay:
Thanks @samford - that is a perfectly reasonable explanation, and thank you for being thorough. Having read through, I am happy to continue as-is. We can easily follow up with separate `bump` PRs where necessary once livecheck updates have been merged.

PR #123881, opened 2022-05-17
mega 11.0.11

miccal:
Hopefully changing the token and updating from version 10 (or "x") to version 11 won't cause too many issues:
```
==> Analytics
install: 4 (30 days), 20 (90 days), 61 (365 days)
```


Casks/mega.rb
```diff
@@ -0,0 +1,35 @@
...
+            pkgutil: "com.igem.pkg.MEGA#{version.major}",
+            delete:  [
+              "/Applications/MEGA#{version.major}.app",
+              "/usr/local/bin/megacc",
+              "/usr/local/share/man/man1/mega.1",
+              "/usr/local/share/man/man1/megacc.1",
+            ]
+
+  zap trash: [
+    "~/Library/Preferences/com.company.MEGA#{version.major} Helper.plist",
```
bevanjkay, line 31: I prefer using globs in zaps instead of `version.major` - because it will clear up leftovers from the previous version too (if they still exist).

bevanjkay (approved):

PR #124077, opened 2022-05-20
coarchi: remove

neutric:
```
==> Analytics
install: 0 (30 days), 0 (90 days), 5 (365 days)
```
Was added 2021-10-11T04:44:22+03:00


neutric:
Added too recently to give it the boot, sorry for only catching that after opening the PR.

PR #124145, opened 2022-05-22
Rename imgotv.rb to mangotv.rb

wjlmelody:
"MangoTV" is the official name of this app. It is necessary to change the name of this rb file.
Reference website:
https://apps.apple.com/us/app/mangotv/id1295034410


miccal:
https://github.com/Homebrew/homebrew-cask/pull/124144#issuecomment-1133842785

PR #124146, opened 2022-05-22
Rename adrive.rb to aliyundrive.rb and Update with version 3.3.0

wjlmelody:
"Aliyundrive" is the official name of this app. It is necessary to change the name of this rb file.
Reference website:
https://www.aliyundrive.com/


miccal:
https://github.com/Homebrew/homebrew-cask/pull/124144#issuecomment-1133842785

PR #124408, opened 2022-05-27
Update version, URL, OS dependency, zap paths for paragon-extfs

p-linnane:

### DIFF
--- a/Casks/paragon-extfs.rb
+++ b/Casks/paragon-extfs.rb
@@ -1,10 +1,14 @@
 cask "paragon-extfs" do
-  version :latest
+  version "11.6.237"
   sha256 :no_check
 
   url "https://dl.paragon-software.com/demo/trial_extfs.dmg"
-  name "Paragon ExtFS"
-  homepage "https://www.paragon-software.com/ufsdhome/extfs-mac/"
+  name "extFS for Mac by Paragon Software"
+  desc "Read/write support for ext2/3/4 formatted volumes"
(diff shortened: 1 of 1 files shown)


p-linnane:
@miccal RE: livecheck: Paragon Software publishes version information on pages that require sign in. I am going to reach out and see if they can publish this information on the public page.

miccal:
Thanks @p-linnane.

Ping @samford for their insight.

samford:
I looked around and didn't find any public source for version information either. The application also requires you to sign in to check for updates. I also searched the app's strings and came up empty handed.

The installer's `Info.plist` contains the current version but `UnversionedCaskChecker` doesn't appear to handle this situation as-is, so I agree with skipping this unless/until upstream starts publishing public version information in some fashion. I was in the process of adding a `skip` suggestion but Bevan committed exactly what I was going to suggest seconds before 😄

Seeing as there's no available version information, is it still appropriate to switch this from `version :latest` to a version string?

bevanjkay:
@samford Given that we are now hash checking `version :latest` artifacts for changes, I feel it's probably more beneficial to change it back.

Particularly, because there's no livecheck we will only know of any update if a user bumps the version after checking it manually. Where `version :latest` would at least check the download for differences if `brew upgrade --greedy` or `brew upgrade --greedy-latest` is used. Additionally, there's no easy way for a maintainer to verify an updated version is correct.

p-linnane:
Thank you all for the input and collaboration on this. This has been a solid learning experience for me!

miccal (approved):

PR #124506, opened 2022-05-30
cmd/lib/ci_matrix: only use macos-10.15 when needed; prefer macos-11

Bo98:
GitHub will be winding down macos-10.15 support over the coming weeks, removing it finally at the end of August.

We however still schedule 90% of PRs to run on it. I've reduced this to 0 - now it's 90% macos-11 and 10% macos-12 (since tha latter is still in beta and will probably have reduced capacity). The exception is when there is a `MacOS.version == :catalina` in the cask, in which case it's beneficial to still test with it and the selection code will do (even with 0 weight).

### DIFF
--- a/cmd/lib/ci_matrix.rb
+++ b/cmd/lib/ci_matrix.rb
@@ -8,9 +8,9 @@ module CiMatrix
   MAX_JOBS = 256
 
   RUNNERS = {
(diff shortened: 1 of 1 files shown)


bevanjkay (approved):
Thanks @Bo98

PR #124512, opened 2022-05-30
Update openemu with livecheck

wjlmelody:


bevanjkay:
@wjlmelody The `:github_latest` strategy is only required when the default `:git` strategy does not work. `:github_latest` requires a higher API usage overhead so we use it sparingly to avoid hitting rate limits. 

When the `url` is a github.com URL, the `:git` strategy is used by default so a livecheck is not required - as Homebrew automatically detects it.

PR #124563, opened 2022-05-31
sublime-text: update version format

samford:

-----

This is a follow-up to #124526, where I noticed that the `sublime-text` cask was using its own unique version instead of respecting the current version format upstream. In the past, [Sublime Text 2](https://www.sublimetext.com/2) initially used build numbers but switched to versions like `2.0.2` after build 2181. [Sublime Text 3](https://www.sublimetext.com/3) still provided a version like `3.2.2` but started referencing the build as well (e.g., `3.2.2 (build 3211)`) and the current version is listed as "Build 3211".

For [Sublime Text 4](https://www.sublimetext.com/download), the build number alone has been used as the version. The version from the app itself is simply `Build 4126`.

With this in mind, it doesn't feel appropriate for the cask to create its own version scheme from the build number (i.e., `4.126`) and we should simply use the build number instead. The major version (`4`) is used in parts of the cask, so this PR uses a version format like `4,4126`, where the major version and build numbers can be easily accessed.

### DIFF
--- a/Casks/sublime-text.rb
+++ b/Casks/sublime-text.rb
@@ -1,8 +1,8 @@
 cask "sublime-text" do
-  version "4.126"
+  version "4,4126"
   sha256 "06e47f404b3e1c14d505bfbd0b884f8d5a62c9f376a62f735a693dc31d9cc212"
 
-  url "https://download.sublimetext.com/sublime_text_build_#{version.no_dots}_mac.zip"
+  url "https://download.sublimetext.com/sublime_text_build_#{version.csv.second}_mac.zip"
   name "Sublime Text"
   desc "Text editor for code, markup and prose"
   homepage "https://www.sublimetext.com/"
@@ -11,10 +11,7 @@ cask "sublime-text" do
     url "https://www.sublimetext.com/download_thanks?target=mac"
     regex(/href=.*?v?(\d+)_mac\.zip/i)
     strategy :page_match do |page, regex|
-      match = page.match(regex)[1]
-      next if match.blank?
-
-      "#{match[0]}.#{match[1..]}"
+      page.scan(regex).map { |match| "#{match[0][0]},#{match[0]}" }
     end
   end
 


bevanjkay (approved):

bevanjkay:
Thanks @samford

bevanjkay:
@reagle This is an unfortunate side effect of changing to a different (but more correct) version number in the cask. Future updates to the cask should now behave for you how they have previously.

samford:
I'm not a cask expert either but if `brew-cask-upgrade` is only using the first part of the version, it may always display as `4` (though hopefully they're using the full version for comparison). `brew info` displays the full version but cask versions can be lengthy, so I can understand why `brew cu` would only use the first part of the version for the sake of brevity.

Since the first digit of Sublime Text build numbers is the major version, including the major version as the first part of the cask `version` was just a way to easily extract the major version. We could switch to using only the build number as the `version` and replace `version.major` in the cask with `version[0]`. I was on the fence about which approach to use but, seeing this shortcoming, it may be better to simply use the build number as the version (I can open a follow-up PR if there's support for the idea). What do you think, @bevanjkay?

bevanjkay:
@samford I don't have a strong preference here. But I'm not aware of the `brew cu` command (looks to be external).
FWIW we do have other casks using similar versioning for a variety of reasons, and they upgrade and format appropriately - example: https://github.com/Homebrew/homebrew-cask/pull/120519

samford:
Yes, [`brew cu`](https://github.com/buo/homebrew-cask-upgrade) is a third-party command with a somewhat surprising number of users. It came on my radar in the recent past when I made a Homebrew/brew change that unfortunately broke some things and they let me know.

> FWIW we do have other casks using similar versioning for a variety of reasons, and they upgrade and format appropriately - example: #120519

If Sublime Text wasn't reliable about the first digit of the build number being the major version (it has been true from at least version 2), then I think this approach would make more sense (i.e., the `basecamp` version contains differing parts). I'm fine with using the build number as the `version`, so I'll create another PR.

reagle:
Confirming, as I just got the latest update via brew cu:

    ==> Found outdated apps
         Cask          Current  Latest  A/U    Result
    1/1  sublime-text  4        4126     Y   [ FORCED ]

PR #124626, opened 2022-05-31
mockoon: update livecheck

samford:

-----

The existing `livecheck` block for `mockoon` includes a redundant `#strategy` call, as the `Git` strategy is already used for the `url` by default. We only use the `#strategy` method when we need to override the default strategy for a given URL or we're using a `strategy` block. Additionally, the regex is unbounded at the end, so it could potentially match more than just stable versions (the standard regex for Git tags like `1.2.3`/`v1.2.3` is `/^v?(\d+(?:\.\d+)+)$/i`). Lastly, the preceding comment about needing a regex to skip tags isn't necessary, as it's implied by the use of `#regex` (i.e., this is common for `Git` checks).

That said, this should be checking GitHub releases since the cask `url` is a download file from a release (i.e., in this context we want livecheck to report a version when it's released, not when it's tagged). This PR resolves the aforementioned issues by simply using the `GithubLatest` strategy. [If this becomes a problem in the future, we can switch to checking the [first-party download page](https://mockoon.com/download/).]

### DIFF
--- a/Casks/mockoon.rb
+++ b/Casks/mockoon.rb
@@ -8,11 +8,9 @@ cask "mockoon" do
   desc "Create mock APIs in seconds"
   homepage "https://mockoon.com/"
 
-  # a regex is required to skip tags that relate to mockoon cli
   livecheck do
     url :url
-    strategy :git
-    regex(/^v?(\d+(?:\.\d+)+)/i)
+    strategy :github_latest
   end
 
   auto_updates true


bevanjkay (approved):

bevanjkay:
@samford Thanks for the fix. A goof on my part there, I shouldn't code after my bedtime 😅

PR #124702, opened 2022-06-01
itools: remove livecheck

samford:

-----

The `itools.hk` page in the `itools` `livecheck` block now returns a 403 (Forbidden) response with the `:default` user agent and resolves to a parked domain page with a `:browser` user agent. It appears that the `itools.hk` and `itools.cn` websites have disappeared.

I'm not sure how we want to handle this situation in the grand scheme of things but at the very least we need to remove this `livecheck` block. If any further changes need to be made here (e.g., adding `discontinued` to the `caveats`), just let me know or feel free to update this accordingly.

### DIFF
--- a/Casks/itools.rb
+++ b/Casks/itools.rb
@@ -7,10 +7,5 @@ cask "itools" do
   name "iTools"
   homepage "https://pro.itools.cn/mac/english"
 
-  livecheck do
-    url "http://dl2.itools.hk/update/iTools64ForMacCast.xml"
-    strategy :sparkle
-  end
-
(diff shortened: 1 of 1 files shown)


bevanjkay:
Let's go with removal, I'll follow up in a separate PR. Thanks @samford.

In general, when we have dead homepages, we look at the analytics for the cask. In this case the URLs have been broken for at least 90 days (0 downloads in that timeframe), and no users have created PRs or mentioned an issue. This combined with the fact that the previous version bump was in 2018 means that it's a good candidate for removal.

PR #124708, opened 2022-06-02
docker: uninstall zsh/site-functions artifacts

RandyMcMillan:
Remove install/reinstall blocking files to ensure a clean configuration.


RandyMcMillan:
```
$ brew install --force --cask docker
```

Screen Shot 2022-06-01 at 8 23 48 PM

The entire install shouldn't be backed out of because these files exist...
Uninstalling doesn't remove these...

bevanjkay:
@RandyMcMillan Have you installed docker in another way previously? I could be wrong, but I doubt that this is a cask issue, as those binaries have been included for 12 months, and in that time there have been ~270,000 successful installs of `docker`.

RandyMcMillan:
You may be correct on the statistics - but many CI sequences completely blow out docker and install clean. This alone would inflate your numbers...

RandyMcMillan:
A completely fresh install is ideal. Since these files are already included in the download - it isnt adding overhead and is what people should assume a forced install/reinstall is doing anyway.

miccal:
Thank you @RandyMcMillan.

miccal (approved):

PR #124814, opened 2022-06-03
tweetbot 3.5.8,35800

n-thumann:
This PR improves the tweetbot livecheck and bumps the version to 3.5.8.

Currently, the `extract_plist` strategy is used, which causes the livecheck URL to be ignored. Instead, the app itself is downloaded and the version is extracted from there.
With this patch applied, the latest version will be extracted from the `update.plist`. In comparison to the current implementation, this will actually return the latest available version instead of only returning the current version of the cask.

---

### DIFF
--- a/Casks/tweetbot.rb
+++ b/Casks/tweetbot.rb
@@ -10,8 +10,13 @@ cask "tweetbot" do
 
   livecheck do
     url "https://tapbots.net/tweetbot4/update.plist"
-    strategy :extract_plist do |version|
-      "#{version.values.map(&:short_version).compact.first},#{version.values.map(&:version).compact.first}"
+    regex(%r{
+      <key>shortVersion</key>.*\n.*<string>(\d+(?:\.\d+)+)</string>
+      (?:.*\n){3}.*
+      <key>version</key>.*\n.*<integer>(\d+)</integer>
+    }ix)
+    strategy :page_match do |page, regex|
+      page.scan(regex).map { |match| "#{match[0]},#{match[1]}" }
     end
   end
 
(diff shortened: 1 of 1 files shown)


Casks/tweetbot.rb
```diff
@@ -10,8 +10,13 @@
 
   livecheck do
     url "https://tapbots.net/tweetbot4/update.plist"
-    strategy :extract_plist do |version|
-      "#{version.values.map(&:short_version).compact.first},#{version.values.map(&:version).compact.first}"
+    regex(%r{
+      <key>shortVersion</key>.*\n.*<string>(\d+(?:\.\d+)+)</string>
+      (?:.*\n){3}.*
+      <key>version</key>.*\n.*<integer>(\d+)</integer>
+    }ix)
```
samford, line 17:
```suggestion
    regex(%r{
      <key>shortVersion</key>.*?<string>(\d+(?:\.\d+)+)</string>
      .*?
      <key>version</key>.*?<integer>(\d+)</integer>
    }imx)
```

We can simplify this a little if we use the multiline (`m`) flag and make the `.*` parts non-greedy (i.e., `.*?`).

n-thumann: Sure! Would you prefer this as an additional commit or have the first commit amended?

samford: Since you agree with the changes, I'll just commit it and it'll get squashed by BrewTestBot when it's merged.

samford (approved):
Thanks, @n-thumann! It's always nice to see a `livecheck` block move away from the `ExtractPlist` strategy.

PR #116308, opened 2021-12-22
GitKraken caveats

fedotov-as:
After making all changes to a cask, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

PR #124889, opened 2022-06-05
Sign Radarr per install instructions

indirect:
Radarr apparently doesn't sign their packages, and expects end-users to
do that after putting the .app in /Applications. This PR adds a
postflight step to run the `codesign` command, following the
instructions at https://radarr.video/#downloads-v3-macos.

Without this step, macOS refuses to launch the app, with the usual
unsigned code message "cannot be opened".

### DIFF
--- a/Casks/radarr.rb
+++ b/Casks/radarr.rb
@@ -20,6 +20,10 @@ cask "radarr" do
     strategy :github_latest
   end
 
+  postflight do
+    `codesign --force --deep -s - /Applications/Radarr.app`
+  end
+
   depends_on macos: ">= :high_sierra"
 
   app "Radarr.app"
--- a/Casks/radarr.rb
+++ b/Casks/radarr.rb
@@ -20,13 +20,13 @@ cask "radarr" do
     strategy :github_latest
   end
 
-  postflight do
-    `codesign --force --deep -s - /Applications/Radarr.app`
-  end
-
   depends_on macos: ">= :high_sierra"
 
   app "Radarr.app"
+
+  postflight do
+    `codesign --force --deep -s - /Applications/Radarr.app`
+  end
 
   zap trash: [
     "~/.config/Radarr",


bevanjkay:
@indirect We don't want to add functions to systematically work around security features - however a `caveat` stanza may be considered.
See https://github.com/Homebrew/homebrew-cask/pull/116308 for an example.

indirect:
This isn't a performance fix like GitKraken, this is literally "brew install completes and leaves you with something that can't run". The security risk here is basically the same as installing a pkg file, and brew does that automatically.

It doesn't make sense to me to say "the user explicitly ran `brew install radarr`, we shouldn't execute the installation instructions provided by the Radarr project. That's like refusing to mount the dmg, or refusing to install the pkg. Yes, it's safer, but it's not what the user asked for.

bevanjkay:
@indirect I understand that this isn't a performance fix like GItkraken, however, we shouldn't systematically work around a security feature that is provided by MacOS.

This exact situation is in the Brew documentation, and codesigning shouldn't actually be necessary to run the application - 
https://docs.brew.sh/FAQ#why-cant-i-open-a-mac-app-from-an-unidentified-developer

indirect:
@bevanjkay ahh, I see, we seem to have been talking about two completely different things. Sorry about that.

This PR is _not_ about the "unidentified developer" problem. I am familiar with that problem, and I know the usual workaround. This problem is about the upstream vendor including no signature whatsoever, which means it is impossible to open the program.

You can run `brew install radarr`, and then follow the instructions from the URL you linked, but when you do, it doesn't matter. The OS will never let the program run, and simply shows this error:

image

(This might only be on M1 machines? I'm not entirely sure about the signature rules on macOS 12.)

Anyway, I don't want to remove the quarantine flag from Radarr, I want it to be possible to open the program at all. This PR allows that, following the instructions on the upstream vendor's website, which are there because they are required to get the OS to let the program open, ever.

bevanjkay:
@indirect Thanks for clarifying. 

In saying that, it still doesn't give reason to systematically work around MacOS security features (regardless of what upstream recommend) so we would only consider a caveat. Brew should not be making these decisions on behalf of the user.

indirect:
I see. I was under the impression that the goal of `brew install` was to install the stated program, automating any manual steps like running scripts. I still hold that creating a local code signature for the downloaded .app is no more dangerous than running a .pkg installer, which can execute literally any shell script, with root permissions! Since you don't want the cask to be able to install a working program, I'll just close this and stop using the cask from this repo.

bevanjkay:
@indirect FWIW I am able to install Radarr on both intel and silicon macs without needing to codesign. It just requires following the instructions to "open anyway" from the security & privacy prefpane.

indirect:
Please tell me your secrets. Have you by any chance run `sudo spctl --master-disable`? I have not, and so there is no option to "Allow apps downloaded from anywhere".

image

Then, after I run `brew install radarr`, I can go to /Applications, right click on Radarr, and choose "Open", and it pops up two dialogue boxes at the same time:

image

If I go back and right-click and choose "Open" a second time, I get the option to "Open":

image


Unfortunately, clicking the button "Open" does nothing. Right-clicking and choosing "Open" yet a third time produces no warning dialogues, and only our old friend:

image

Even after having chosen the "Open" button from the dialogue box, it is also impossible to run the executable from the command line:

```
❯ /Applications/Radarr.app/Contents/MacOS/Radarr
zsh: killed     /Applications/Radarr.app/Contents/MacOS/Radarr
```

Any ideas why my experience is so different from yours?

indirect:
I found [this article about required code signing on Apple Silicon starting with Big Sur](https://eclecticlight.co/2020/08/22/apple-silicon-macs-will-require-signed-code/), which provides this quote from Apple's Big Sur beta release notes:

> the operating system will enforce that any executable must be signed with a valid signature before it’s allowed to run. There isn’t a specific identity requirement for this signature: a simple ad-hoc signature issued locally is sufficient, which includes signatures which are now generated automatically by the linker. This new behavior doesn’t change the long-established policy that our users and developers can run arbitrary code on their Macs, and is designed to simplify the execution policies on Apple silicon Mac computers and enable the system to better detect code modifications. This new policy doesn’t apply to translated x86 binaries running under Rosetta, nor does it apply to macOS 11 running on Intel platforms.

When I check, the .app installed by homebrew does in fact have no signature at all, and so I should not be able to run it on Apple Silicon under Apple's stated rules:

```
❯ codesign --verify --verbose /Applications/Radarr.app
/Applications/Radarr.app: code object is not signed at all
In architecture: arm64
```

Given that Apple explicitly states this is not a security issue, and is simply to allow the OS to detect maliciously modified applications, it seems correct, and more secure, for Homebrew to immediately sign the .app at install-time, allowing the OS to detect if malware ever modifies it in the future.

bevanjkay:
@indirect Apologies, I was mistaken I am seeing the same error on my M1 machine.

If you would still like to manage the installation with brew, what I would recommend you do is host the cask in a personal tap, which is very easy to do: https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap
Then you can have the system command run upon installation.

indirect:
Now that you can see that this cask does not actually work on M1 machines, would you be open to accepting this PR?

Thank you for the link, I do already maintain my own tap, and I will add the cask there if you ultimately decide against this change.

bevanjkay:
@indirect As I have already explained, manually codesigning is absolutely a decision that should be made **by the user** and not something that brew should do automatically.
A caveat would provide the required instructions to the user.

indirect:
Okay. Let me take this moment to say that, as a user of homebrew, my decision is that I very much want the `brew install` command to ad-hoc sign things I have explicitly asked to have installed.

I would not use the `brew install` command unless I wanted the program to be able to run on my machine.

I hope Homebrew will change this policy (specifically and only for ad-hoc signatures on M1 machines) in the future.

PR #124757, opened 2022-06-02
Update docker from 4.8.2 to 4.9.0

p-linnane:
Created with `brew bump-cask-pr`.

PR #124900, opened 2022-06-05
docker: add missing completions and uninstall them

darklore:

Add missing shell completion scripts installation removed in https://github.com/Homebrew/homebrew-cask/pull/124757
Fix uninstallation of shell completions.


kidonng:
Hah, you beat me by a couple of minutes, I was looking at #124757 as well.

Since we are at it, could you correct those completions in `uninstall` to use `#{HOMEBREW_PREFIX}`? It was done incorrectly in #124708.

Casks/docker.rb
```diff
@@ -33,8 +33,12 @@
   ]
 
   app "Docker.app"
+  binary "#{appdir}/Docker.app/Contents/Resources/etc/docker",
+         target: "#{HOMEBREW_PREFIX}/etc/bash_completion.d/docker"
```
bevanjkay, line 37: @darklore This is different to the previously linked bash_completion for docker that was removed in #124757 as it no longer existed. Are you intentionally linking the main `docker` binary in the bash_completion folder?

darklore: No.
This file ("#{appdir}/Docker.app/Contents/Resources/etc/docker") is not main docker binary.
This is shell completion script.

darklore:
@kidonng Thank you for your advice.
I modified the patch to use `HOMEBREW_PREFIX` in uninstallation too.

miccal (approved):

PR #124970, opened 2022-06-07
Update start from 0.301.6 to 0.301.19

wjlmelody:

### DIFF
--- a/Casks/start.rb
+++ b/Casks/start.rb
@@ -1,5 +1,6 @@
 cask "start" do
   arch = Hardware::CPU.intel? ? "" : "m1/"
+  livecheck_arch = Hardware::CPU.intel? ? "" : "-arm"
 
   if Hardware::CPU.intel?
     version "0.301.19-183"
@@ -16,8 +17,8 @@ cask "start" do
   homepage "https://start.qq.com/"
 
   livecheck do
-    url :homepage
-    regex(%r{.*installer/#{arch}START-Installer[._-]v?(\d+(?:[.-]\d+)+)\.dmg}i)
+    url "https://api.start.qq.com/cfg/get?biztypes=macos-update-info#{livecheck_arch}"
+    regex(/START-Installer[._-]v?(\d+(?:[.-]\d+)+)\.dmg/i)
   end
 
   auto_updates true
--- a/Casks/start.rb
+++ b/Casks/start.rb
@@ -15,9 +15,18 @@ cask "start" do
   desc "Tencent cloud gaming platform"
   homepage "https://start.qq.com/"
 
-  livecheck do
-    url :homepage
-    regex(%r{.*installer/#{arch}START-Installer[._-]v?(\d+(?:[.-]\d+)+)\.dmg}i)
+  # Upstream is replacing the intel version of the app on the homepage with javascript after page load.
+  # The arm64 version is not being replaced - so a split livecheck is required.
+  if Hardware::CPU.intel?
+    livecheck do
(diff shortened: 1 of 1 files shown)


bevanjkay (dismissed):

bevanjkay:
@wjlmelody I have had to split the livecheck by architecture for now, however, we will need to keep an eye on it, as it may not be consistent moving forward.

bevanjkay:
@wjlmelody It looks like it was a transient issue. So I'll give it a little bump and we should be good to go.

wjlmelody:
> @wjlmelody‎这看起来像是一个暂时性的问题。所以我会给它一点颠簸，我们应该很好。‎‎编辑：实际上看起来英特尔已经恢复到以前的版本，但arm64版本已经更新。‎

OK. Thanks for your work.

miccal:
@samford can you please have a look at the `livecheck` here, thanks.

samford:
> @samford can you please have a look at the `livecheck` here, thanks.

The pre-existing approach (i.e., checking `api.start.qq.com`) appears to be correct from my point of view. The HTML of the homepage (and [download page](https://start.qq.com/download.html)) links to old files (i.e., `START-Installer-0.299.1-139.dmg` for x86 and `START-Installer-0.299.1-26.dmg` for ARM), so this proposed `livecheck` block in this PR surfaces those older versions instead of the newest ones. The links you see in a browser come from separate requests to `api.start.qq.com` (one for x86 and another for ARM) and this is what the pre-existing `livecheck` block checks. The files given by the API URLs are `START-Installer-0.301.6-170.dmg` for x86 and `START-Installer-0.301.17-65.dmg` for ARM.

I don't see where the `0.301.19-183` (x86) and `0.301.19-67` (ARM) versions in this PR come from. Unless there's a reliable source for these versions, I'm in favor of keeping the pre-existing approach and simply bumping the ARM version to `0.301.17-65` (to align the cask with the advertised versions on the homepage and download page).

bevanjkay:
@samford I've had a look at this in the past. From what I've seen the API returns different versions on different requests. 

The html returns a version, and sometimes the API replaces the value, but sometimes it doesn't (returns an empty object instead). I'm not sure if this is intentional or not.

miccal:
I suggest we just skip the `livecheck` here with an explanation of the issue.

samford:
> I've had a look at this in the past. From what I've seen the API returns different versions on different requests.

Does this vary based on something like headers/geolocation/etc. or does the API randomly return different version information for identical requests (e.g., some sort of A/B testing)? In my limited testing over the past two days, the pre-existing check has reliably given me `0.301.17-65` each time on ARM (though this isn't the version in this PR). I also used a VPN to test it in a few different countries (albeit, all outside of China) and the version still stayed the same.

That said, the main questions for me are whether the versions in this PR are the correct latest versions and how they were identified. If the only checkable version information is outdated, then I think skipping the check (and relying on user-contributed version bumps) may make sense, as Miccal suggested.

Otherwise, if `0.301.6-170` (Intel) and `0.301.17-65` (ARM) are the correct latest versions, then it seems like we should maintain the pre-existing setup and simply bump the ARM version here.

bevanjkay:
@samford There was a time where the API was not returning a different version, so the version embedded in the HTML was the one being used.

This would require us to check the HTML, and check the API, merge the versions.
But perhaps this was just a transient issue at the time, and moving forward the API will be reliable.

At the same time as this occurrence we were seeing different version returned from the API on sequential requests, but again this may have been transient.

I think we could reinstate the livecheck from the API URL and see how it goes.

bevanjkay (approved):

PR #124798, opened 2022-06-03
Update Librewolf from v100.0.2-1 to v101.0-2

fxbrit:
After making all changes to a cask, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [ ] `brew audit --cask --online <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

PR #125385, opened 2022-06-10
Update LibreWolf from v101.0-2 to v101.0.1-1

fxbrit:

### DIFF
--- a/Casks/librewolf.rb
+++ b/Casks/librewolf.rb
@@ -2,11 +2,11 @@ cask "librewolf" do
   arch = Hardware::CPU.intel? ? "x86_64" : "aarch64"
 
   if Hardware::CPU.intel?
-    version "101.0,2,3dfd2f7e3647dfd6d40c7564acdaaa1e"
-    sha256 "1f0b31fdbabcf92359a14e300988745a6c27c6790dc61fafea0cbd4e1f2586c6"
+    version "101.0.1,1,7e71ebf6a602644d41b0abc72050485d"
+    sha256 "b64ef8fcb5357fa949f471334ebf1dd3bcd335e18bc4aaf0c4aa796c3caa73e4"
   else
-    version "101.0,2,1f2e33257adc82b080cfd6cc4b6bc61a"
-    sha256 "82cb70f87ccf423d8f97d83ee37d975a1a124d9e269a47ce7f40cf18267a3fd2"
+    version "101.0.1,1,6afeeb127dd2ca609ebacb161143c778"
+    sha256 "8da76424422d988c41d6108f4dba175fad8817b1927e1fd687f824f9935101ed"
   end
 
   url "https://gitlab.com/librewolf-community/browser/macos/uploads/#{version.csv.third}/librewolf-#{version.csv.first}-#{version.csv.second}.en-US.mac.#{arch}.dmg",


fxbrit:
@bevanjkay as a follow up from https://github.com/Homebrew/homebrew-cask/pull/124798#issuecomment-1146000909: I install don't fully understand the need to run the `--online` check. I mean, isn't the livecheck expected to fail because the newly released version is of course newer than the one in brew? 

I run `brew livecheck --debug --cask librewolf`:

```
Cask:             librewolf
Livecheckable?:   Yes

URL:              https://gitlab.com/api/v4/projects/13853965/releases
Strategy:         PageMatch
Regex:            /\/(\w+)\/librewolf[._-](\d+(?:\.\d+)+)-(\d+)\.en-US\.mac\.x86_64\.dmg[\s"]/i

Matched Versions:
101.0.1,1,7e71ebf6a602644d41b0abc72050485d, 101.0,2,3dfd2f7e3647dfd6d40c7564acdaaa1e, 100.0.2,1,6b8461deccb01d1f6b626106b8608067, 100.0.1,1,09e6256595bec55f433cb7c0c10a009a, 100.0,3,e7345295904e99340fff3b7a07e0cc68, 100.0,1,80b24f36c1618f7d9655c1bbc180653f, 99.0.1,4,ce11b221ea2eedc393ffbc66d204a354, 99.0.1,1,659a83b4fdf2cb6b7e1f7bb0382cd1b4, 99.0,1,d6a6b221192d4042226fda02abd5dc6d, 98.0.2,1,9a776608fca00d9313ed271769d672de, 98.0,1,c3f59ce3926026cd1786008fcf09f69f, 97.0.2,1,fecd166b8f84dc13ceba6acfc77797fe, 97.0.1,1,d6906f34d5b20b5ca889d7062378d9b6, 97.0,1,45070518c865ab1af2cd2c40fe9c3143

librewolf: 101.0,2,3dfd2f7e3647dfd6d40c7564acdaaa1e ==> 101.0.1,1,7e71ebf6a602644d41b0abc72050485d
```

it looks right, but for the check to be useful wouldn't it need to match upstream with the version in my local branch, rather then the one already live in brew?

bevanjkay:
The `--online` flag doesn't use the cask version from the online tap. It runs some additional audits such as `brew livecheck` with your local version to ensure that the changes are correct. 

These audits are run in CI with your changes applied for the PR to go green before merging.

fxbrit:
thanks for clarifying. does it run the livechecks against a particular local branch then (eg. master, and not the one I'm working in)? as you can see the matched version is `101.0.1` and my cask code is of course pointing to the same one, as the PR shows.

bevanjkay:
It uses whatever branch you are currently using

fxbrit:
then I'm probably missing something, because the output I posted in https://github.com/Homebrew/homebrew-cask/pull/125385#issuecomment-1152365829 shouldn't look that way, and `brew audit --cask --online <cask>` should be error free :-(

fxbrit:
mmmh, actually I run it after a `git checkout v101.0.1-1-librewolf`. I double-checked again and re-run it just to make sure, which is why I asked if the branch matters; same result as above.

thanks for the suggestion btw.

bevanjkay (approved):

fxbrit:
well I have no idea then, at least we know it works for everyone but me I guess :-) thanks again!

PR #23360, opened 2016-08-02
Delete gedit.rb

vitorgalvao:
See https://github.com/Homebrew/homebrew-core/pull/3331.

PR #125791, opened 2022-06-16
double-commander: Discontinued

p-linnane:

### DIFF
--- a/Casks/double-commander.rb
+++ b/Casks/double-commander.rb
@@ -8,5 +8,10 @@ cask "double-commander" do
   desc "File manager"
   homepage "https://doublecmd.sourceforge.io/"
 
+  livecheck do
+    url "https://sourceforge.net/projects/doublecmd/rss"
+    regex(%r{url=.*?/doublecmd/files/OldFiles/.*?[-_/](\d+(?:[-.]\d+)+)[-_/%.].}i)
+  end
+
   app "Double Commander.app"
 end
--- a/Casks/double-commander.rb
+++ b/Casks/double-commander.rb
@@ -8,10 +8,9 @@ cask "double-commander" do
   desc "File manager"
   homepage "https://doublecmd.sourceforge.io/"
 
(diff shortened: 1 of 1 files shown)


p-linnane:
Development on macOS seems to have stopped. This appears to be the final release.

bevanjkay:
@p-linnane We can instead add a discontinued caveat block and remove the livecheck. 

https://docs.brew.sh/Cask-Cookbook#caveats-mini-dsl

p-linnane:
@bevanjkay Updated with the discontinued caveat.

bevanjkay:
@samford Is the current failure on the CI run an issue in the livecheck logic? Having a `discontinued` caveat should mean the livecheck block can (and should) be omitted, but the audit is failing because the Sourceforge URL requires a livecheck?

samford:
> Is the current failure on the CI run an issue in the livecheck logic? Having a `discontinued` caveat should mean the livecheck block can (and should) be omitted, but the audit is failing because the Sourceforge URL requires a livecheck?

I didn't create the livecheck-related cask audits but, as you discovered, this is a shortcoming in the `Cask::Audit#check_hosting_with_livecheck` logic and I agree with the brew PR 👍

bevanjkay (approved):

decadent:
BTW, Double Commander can actually be compiled on OSX—dunno what's Homebrew's policy in regard to compiling apps, as opposed to console stuff.

p-linnane:
@decadent I believe that would fit better in the core tap. Per the [Rejected Casks](https://docs.brew.sh/Acceptable-Casks#rejected-casks) section under docs:

> The app is open-source and has a GUI but no compiled versions (or only old ones) are provided. It’s better to have them in [Homebrew/core](https://github.com/Homebrew/homebrew-core) so users don’t get perpetually outdated versions. See [gedit](https://github.com/Homebrew/homebrew-cask/pull/23360) for example.

PR #126079, opened 2022-06-20
better-window-manager: Switch to HTTP

p-linnane:
TLS certificate expired 16 days ago. Livecheck and install are failing.


p-linnane:
@SMillerDev Similar to #126084, this application hasn't been updated since 2017

miccal (approved):

PR #126122, opened 2022-06-20
Revert Playcover 1.0.1 Update

hayatofujii:
Playcover 1.0.1 dmg is shipping a non-working version (requires a `sudo xattr -cr /Applications/PlayCover.app` to fix it). Upstream is already aware of this problem.

Let's revert to the 1.0.0 version for the time being.

This reverts commit f66ffba9c2cae2ef40d2e81bce59e60b4dbd47ae.

### DIFF
--- a/Casks/playcover.rb
+++ b/Casks/playcover.rb
@@ -11,9 +11,9 @@ cask "playcover" do
 
     app "PlayCover #{version}/PlayCover.app"
   else
-    version "1.0.1"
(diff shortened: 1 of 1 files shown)


bevanjkay:
@hayatofujii Generally we would downgrade when upstream pull the release, but it looks like they are currently still shipping the previous version.

hayatofujii:
@bevanjkay Doesn't seem like upstream is retracting or fixing the image. Will close this down for the time being...
Be advised that people may complain about this :(

PR #126162, opened 2022-06-21
mysteriumvpn: Fix livecheck

p-linnane:

### DIFF
--- a/Casks/mysteriumvpn.rb
+++ b/Casks/mysteriumvpn.rb
@@ -8,6 +8,11 @@ cask "mysteriumvpn" do
   desc "VPN client"
   homepage "https://www.mysteriumvpn.com/"
 
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
+
   app "MysteriumVPN.app"
 
   zap trash: [


bevanjkay (approved):

otkd:
@bevanjkay why was the same livecheck stanza edited out of the [initial PR](https://github.com/Homebrew/homebrew-cask/pull/119364/commits/fadc71f2e0d7d80ca12025af67151c14075a5589#diff-9c4391335c79b502a67fe595d2cbec638e05ea98c65a47f9acf341a42f197fc8L11)?

bevanjkay:
@otkd Generally the `:github_latest` strategy isn't required except in a couple of instances. 1) Upstream uses GitHub prereleases. 2) Upstream use tags that do not have a matching release. 3) Upstream has legacy tags that are picked up as newer than the correct latest version.

In this case, it doesn't presently look like `mysteriumvpn` requires the livecheck stanza, as the `:git` strategy is inferred from the URL by default. Do we know if upstream uses GitHub prereleases or creates tags before releases? 
If not then this PR should be reverted, we avoid `:github_latest` as much as possible as it uses additional API calls that can result in ratelimiting.

otkd:
> @otkd Generally the `:github_latest` strategy isn't required except in a couple of instances. 1) Upstream uses GitHub prereleases. 2) Upstream use tags that do not have a matching release. 3) Upstream has legacy tags that are picked up as newer than the correct latest version.
> 
> 
> 
> In this case, it doesn't presently look like `mysteriumvpn` requires the livecheck stanza, as the `:git` strategy is inferred from the URL by default. Do we know if upstream uses GitHub prereleases or creates tags before releases? 
> 
> If not then this PR should be reverted, we avoid `:github_latest` as much as possible as it uses additional API calls that can result in ratelimiting.
> 
> 
> 
> EDIT: Sorry, I can see now that `prereleases` are used upstream, so the livecheck block is necessary.

@bevanjkay thank you for the detailed clarification, much appreciated!

PR #126453, opened 2022-06-24
winzip: Bump version, add livecheck

p-linnane:

### DIFF
--- a/Casks/winzip.rb
+++ b/Casks/winzip.rb
@@ -1,10 +1,16 @@
 cask "winzip" do
-  version "7.0.4565"
-  sha256 "e65aab9f4619a190da6fc29a01864049cf409e5ae6c962d840d96bbd25d98899"
+  version "9.0.5520"
+  sha256 "2b7f2b5322ffe709151b9790aaba6231ce54e423ceca5ea9433977e3079da528"
 
   url "https://download.winzip.com/winzipmacedition#{version.major}.dmg"
   name "WinZip"
+  desc "File archiving tool"
   homepage "https://www.winzip.com/mac/en/winzip.html"
+
+  livecheck do
+    url :url
+    strategy :extract_plist
+  end
 
   app "WinZip.app"
 end
--- a/Casks/winzip.rb
+++ b/Casks/winzip.rb
@@ -8,9 +8,16 @@ cask "winzip" do
   homepage "https://www.winzip.com/mac/en/winzip.html"
 
   livecheck do
-    url :url
-    strategy :extract_plist
-  end
+    url "https://www.winzip.com/en/download/"
+    regex(/href=.*?winzipmacedition[._-]?v?(\d+)\.dmg/i)
+    strategy :page_match do |page, regex|
+      major_version = page[regex, 1]
+      next if major_version.blank?
 
+      cask = CaskLoader.load("winzip")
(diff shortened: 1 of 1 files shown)


Casks/winzip.rb
```diff
@@ -1,10 +1,16 @@
...
   url "https://download.winzip.com/winzipmacedition#{version.major}.dmg"
   name "WinZip"
+  desc "File archiving tool"
   homepage "https://www.winzip.com/mac/en/winzip.html"
 
+  livecheck do
+    url :url
+    strategy :extract_plist
+  end
+
```
bevanjkay, line 14: Albeit a bit convoluted, when https://github.com/Homebrew/brew/pull/13346 is merged we could do the below which will detect the full version, even when the major version changes.

```suggestion
  livecheck do
    url "https://www.winzip.com/mac/en/winzip.html"
    regex(/href=.*?winzipmacedition[._-]?v?(\d+)\.dmg/i)
    strategy :page_match do |page, regex|
      major_version = page.scan(regex).flatten.first
      next if major_version.blank?
      
      download_url = "https://download.winzip.com/winzipmacedition#{major_version}.dmg"

      cask = CaskLoader.load('winzip')

      Homebrew::Livecheck::Strategy::ExtractPlist.find_versions(cask: cask, url: download_url)[:matches].first
    end
  end
```

@samford There may be a better way to initiate the second half, and to pass the current cask object into the function (without reloading it with Cask::CaskLoader).

samford: As it stands, I don't think there's a way for us to reference the cask in this context other than what you're already doing. We could potentially modify the `livecheck` DSL to add `:formula_or_cask` to `attr_reader`, which would allow us to use it like `find_versions(cask: formula_or_cask, ...)`. There are also some `strategy` blocks in homebrew/core where we need to access formula values (e.g., `stable`, `homepage`, etc.), which is a somewhat similar situation. I'll have to have a proper think about how we may want to approach this when my brain's a bit less tired.

In the interim time, I'm okay with this approach but would suggest a few tweaks:

* The `homepage` and download page (below) both provide the same information but the latter is a few KB smaller. [Though, if we used the homepage, we would want to use `url :homepage` here.]
* When we only need the first match on a page, we normally use `page[regex, 1]`. This works for now but if the download page ever includes more than one dmg link, we would want to use `page.scan(regex).flatten.uniq.map { |v| Version.new(v) }.max` to properly identify the newest version (instead of assuming the first will be the newest).
* I generally try to return all matches from a `strategy` block and let livecheck sort out the newest version. We do have a small number of `strategy` blocks where it's necessary to only use the first match (e.g., if a page contains a link for a stable version followed by a development version, both using a stable version format) but they're exceptions.

```suggestion
  livecheck do
    url "https://www.winzip.com/en/download/"
    regex(/href=.*?winzipmacedition[._-]?v?(\d+)\.dmg/i)
    strategy :page_match do |page, regex|
      major_version = page[regex, 1]
      next if major_version.blank?

      cask = CaskLoader.load("winzip")
      download_url = "https://download.winzip.com/winzipmacedition#{major_version}.dmg"
      Homebrew::Livecheck::Strategy::ExtractPlist.find_versions(cask: cask, url: download_url)[:matches].values
    end
  end
```

bevanjkay (approved):

PR #126673, opened 2022-06-28
jdownloader: Change URL & split arch

p-linnane:

### DIFF
--- a/Casks/jdownloader.rb
+++ b/Casks/jdownloader.rb
@@ -1,11 +1,17 @@
 cask "jdownloader" do
-  version "46316"
-  sha256 :no_check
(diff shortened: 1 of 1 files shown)


p-linnane:
`Error: Failure while executing; '/usr/local/Caskroom/jdownloader/46316/JDownloader\ 2\ Installer.app/Contents/MacOS/JavaApplicationStub -dir /Applications -q -Dinstall4j.suppressStdout=true -Dinstall4j.debug=false -VcreateDesktopLinkAction\$Boolean=false -VaddToDockAction\$Boolean=false' exited with 22.`

Running this command outside of brew works correctly. I'm unable to determine why running this in the cask is causing failures, especially since prior versions ran it with no issue. Any assistance would be appreciated.

PR #127514, opened 2022-07-10
add sha256 for Ringcentral

yslpn:


bevanjkay:
@yslpn This cask cannot have a `sha256` value because every time the vendor updates the application the cask will stop functioning. We are only able to add a shasum check to casks that have versioned downloads, as generally speaking each version's download will continue working for an indefinite amount of time.

PR #127789, opened 2022-07-13
firefly: Fix livecheck & bump to 1.6.3

p-linnane:

### DIFF
--- a/Casks/firefly.rb
+++ b/Casks/firefly.rb
@@ -10,7 +10,8 @@ cask "firefly" do
 
   livecheck do
     url :url
-    regex(/^desktop[._-]v?(\d+(?:\.\d+)+)$/i)
+    strategy :github_latest
+    regex(/desktop[._-]v?(\d+(?:\.\d+)+).dmg/i)
   end
 
   auto_updates true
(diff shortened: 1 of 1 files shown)


samford (changes requested):
Using the `GithubLatest` strategy is appropriate here (i.e., the cask uses a release artifact, so checking the tags isn't appropriate and could surface a version before the related dmg file is available). Past that, there are a few areas to address:

* When a `livecheck` block contains both a `#regex` and `#strategy` call, `#regex` should come before `#strategy`. The idea is that the regex can be passed into a `strategy` block and having the `#regex` call before the `#strategy` call helps to make it clear that the `#regex` argument is the same regex being passed into the `strategy` block (one person was confused about this in the past).  This particular `livecheck` block doesn't contain a `strategy` block but this stylistic norm makes it easier to add a `strategy` block in the future, if necessary. [Long term, I intend to add an audit that catches this but I need to finish reworking livecheck audits to make them apply to casks as well (I have a local implementation but it still needs some work). This is currently blocked by an open Homebrew/brew PR that refactors existing audits and will likely require significant rework of my existing approach.]
* When matching versions from filenames on an HTML page, we match within `href` attributes whenever possible (though there are rare exceptions to this).
* Parts of a regex that intend to match a dot need to use an escaped dot (`\.`), as `.` means "match anything" in regex syntax. That is to say, `.dmg` should be `\.dmg` in this regex.

p-linnane:
Thank you as always for the guidance @samford.

samford (approved):
Thanks, @p-linnane!

PR #127813, opened 2022-07-13
tigervpn: Discontinued

p-linnane:
Service has shutdown

https://www.tigervpn.com


SMillerDev:
Can you add a `discontinued` caveat?

p-linnane:
@SMillerDev The binaries are no longer available. Is the `discontinued` caveat still the right move here?

p-linnane:
@SMillerDev ready for re-review

PR #127942, opened 2022-07-15
name-mangler: Remove cask

p-linnane:
Site is now behind CloudFlare. Cannot fetch binaries or perform livecheck.
```
==> Analytics
install: 4 (30 days), 23 (90 days), 108 (365 days)


zgracem:
[The site outage was temporary and due to a DDoS attack against the dev](https://twitter.com/manytricks/status/1547950477228158983). Perhaps this should be reinstated?

p-linnane:
@zgracem Thank you for the heads up! #128046

PR #128076, opened 2022-07-18
Add new cask `Growtopia`

Adrthegamedev:

### Problems
- [ ] There is a second version displayed within the app that I'm not sure how to write a livecheck for. It's currently `V3.95`. I think it can be grabbed from [the apple appstore page for it](https://apps.apple.com/US/app/id590495115), but trying to figure out how to combine livechecks with `csv` versions is making my head spin.

### DIFF
--- a/Casks/growtopia.rb
+++ b/Casks/growtopia.rb
@@ -1,0 +1,20 @@
+cask "growtopia" do
+  version "56596"
+  sha256 "e267e37af3c08427b7d19f37848c817edfc8f1990b7755622337d58b82c943fd"
+
+  url "https://ubistatic-a.akamaihd.net/0098/#{version}/Growtopia.dmg",
+      verified: "ubistatic-a.akamaihd.net/0098/" # Growtopia is owned by Ubisoft, 0098 is where Growtopia-related contents are usually stored on their website
+  name "Growtopia"
+  desc "Massively multiplayer online sandbox game"
+  homepage "https://growtopiagame.com/"
+
+  livecheck do
+    url "https://growtopiagame.com/Growtopia-mac.dmg"
+    strategy :header_match
+    regex(%r{0098/v?(\d+)/Growtopia\.dmg}i)
+  end
+
+  app "Growtopia.app"
+
+  zap trash: "~/Library/Application Support/Growtopia"
+end
--- a/Casks/growtopia.rb
+++ b/Casks/growtopia.rb
@@ -3,7 +3,7 @@ cask "growtopia" do
   sha256 "e267e37af3c08427b7d19f37848c817edfc8f1990b7755622337d58b82c943fd"
 
   url "https://ubistatic-a.akamaihd.net/0098/#{version}/Growtopia.dmg",
-      verified: "ubistatic-a.akamaihd.net/0098/" # Growtopia is owned by Ubisoft, 0098 is where Growtopia-related contents are usually stored on their website
+      verified: "ubistatic-a.akamaihd.net/0098/"
   name "Growtopia"
   desc "Massively multiplayer online sandbox game"
   homepage "https://growtopiagame.com/"


Adrthegamedev:
Odd, the Github Action seemingly doesn't have access to the growtopia website, as it both 403s and doesn't get any version returned. It worked when I tested it (multiple times, infact) so i'm not sure what's up here.

Adrthegamedev:
I think I know why; it's [robots.txt](https://growtopiagame.com/robots.txt) likely blocks the github actions bot from reaching it.
Other than that small problem with the checks, there's still the problem of the other version (V3.95) that needs to be solved somehow.

bevanjkay:
@Adrthegamedev To me it looks like the version provided as a direct download is only the `beta` and that stable versions are only provided through the Mac App Store?

Adrthegamedev:
> @Adrthegamedev To me it looks like the version provided as a direct download is only the `beta` and that stable versions are only provided through the Mac App Store?
Thing is, it's not available on the **Mac** App Store, only the iPhone and iOS ones, but that does come into question if the cask should be renamed to "growtopia-beta.rb".

bevanjkay:
@Adrthegamedev In that case it will be fine as-is. We will just need to sort something out regarding the livecheck.

p-linnane:
Our issue here is the configuration of the AWS Elastic Load Balancer. The response headers contain:
```
Access-Control-Allow-Origin: https://growtopiagame.com
Access-Control-Allow-Origin: https://www.growtopiagame.com
```
Which prevents livecheck's standard `curl` configuration from working. Specifying the origin works fine:

```
curl -LH "Origin: https://growtopiagame.com" https://growtopiagame.com/Growtopia-mac.dmg --output ~/Desktop/Growtopia-mac.dmg
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   112  100   112    0     0    305      0 --:--:-- --:--:-- --:--:--   306
100  144M  100  144M    0     0  12.7M      0  0:00:11  0:00:11 --:--:-- 15.9M
```
@bevanjkay Does livecheck have a way for us to pass an origin?

bevanjkay:
@p-linnane Not at the moment, but in the future we should be able to have a bit more control. @samford has been chipping away at an `options` API for livecheck.

Adrthegamedev:
A trouble I just noticed is that the internal livecheck version doesn't seem to be alphabetically consistent; I originally commited this with a version of "56596", but, if you go download it right now, it downloads as "52720".

samford:
`curl --location --head --request GET --include  https://growtopiagame.com/Growtopia-mac.dmg` (what `Strategy#page_headers` uses internally) works fine for me without an `Origin` header. This `livecheck` block also works for me locally, granted the current URL in the location header of the response is `https://ubistatic-a.akamaihd.net/0098/52720/Growtopia.dmg`, as mentioned above (though this appears to be an upstream issue, not a livecheck issue). If the version doesn't remain consistent, then that would be another problem for livecheck.

The issue for me is that the server gives a 403 (Forbidden) response when the request is made from a blocked IP address range. For example, I tested this on a VPN and received a 403 response even when an `Origin` header was provided in the request. The 403 response on CI when checking the homepage would suggest to me that our CI machines may also be in a blocked IP address range (it's not uncommon for known cloud providers' servers to also be blocked when VPNs are blocked). It's worth confirming but, if true, it seems like that would be the bigger issue here.

Unless we can come up with a way to work around this, this cask may always fail on CI and I'm not sure how we would handle this type of situation. Our cask maintainers would probably have a better idea about whether there are other casks with this sort of issue (e.g., which have to be merged despite CI failures) and how they're handled.

-----

> @samford has been chipping away at an `options` API for livecheck.

For what it's worth, I've had the `options` feature implemented locally for quite some time but I've been working towards creating PRs for some other changes first, time permitting (I've been busier with work/life/etc. recently). The internal implementation is likely pretty settled but I'll need to discuss approaches to the API (and the tradeoffs) and that may end up taking a bit of time. At the moment, our GSoC project is currently modifying some of livecheck's internals, so I'm trying to be careful about PRs that may create conflicts for our student. That said, there are a handful of casks that explicitly need `options` and I also have some other livecheck changes planned that depend on it, so it's closer to the top of my to-do list.

unitof:
Works for me. Just livechecked and then updated to `49124` and it downloaded & installed successfully (though app is not notarized).

p-linnane:
Given that the app is not signed, we would not be able to accept it into the main repo.

Please consider hosting this [in your own Tap](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap) and if the developers sign their app in the future we would be happy to consider adding this to `homebrew-cask`.

Adrthegamedev:
> Given that the app is not signed, we would not be able to accept it into the main repo.

Thing is, `codesign -dv` says it's adhoc signed, and not simply not signed at all.

Adrthegamedev:
@p-linnane I'm not sure if that means it's not signed, but it's certainly an indication of something, and definitely not the reason that the roblox casks have been rejected in the past for.

p-linnane:
Gatekeeper cannot run checks on an ad-hoc signature, so essentially it is unsigned. If you choose to host this in your own tap you can pass `--no-quarantine` at install to strip the `com.apple.quarantine` attribute or run `xattr -dr com.apple.quarantine /path/to/app`.

Adrthegamedev:
Ah, makes sense.

PR #128734, opened 2022-07-27
voov-meeting: Skip livecheck

p-linnane:
The livecheck for `voov-meeting` no longer works. @samford implemented a novel solution in #120458, but `bonjour.swoosh.run` no longer resolves.

I know the `options` implementation for `livecheck` is still being worked on, so this should be a temporary measure. Hoping for some discussion here.


samford (approved):
This is in line with how we're currently handling other casks that need to make a `POST` request in the `livecheck` block, so it makes sense to me. Per my previous comment, I still have a tentative implementation of this `livecheck` block stashed locally and I'll create a PR for it (and related casks) once I can make the `options` feature available (i.e., sometime after our GSoC project is past the point where we're modifying livecheck's internals).

Thanks, @p-linnane!

PR #128893, opened 2022-07-29
mightymike: update homepage, rename to mighty-mike

EricFromCanada:
This set of game ports now has its own site. Also trims an unnecessary zap path and renames the cask to match the app package name.

---


bevanjkay:
@EricFromCanada We tend to avoid changing tokens once a cask is already established, as unlike `homebrew-core` we do not have a way to alias the old token to the new.

This instance potentially could be okay as the cask is not super popular, so I'm interested to hear your thoughts here.

```
mightymike
30 days: n/a
90 days: 2 (#6718)
365 days: 20 (#5510)
Age: 280 days (added 2021, October 23)
```

miccal (approved):

PR #128899, opened 2022-07-30
Add Roblox v0.539.0.5390403

aaronliu0130:

- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues). It appears that the refused ones were due to the submitter not correctly using an `installer` stanza and closed to stale.
- [ ] `brew audit --new-cask <cask>` worked successfully. It turns out the installer is signed but the app itself isn't signed.

### DIFF
--- a/Casks/roblox.rb
+++ b/Casks/roblox.rb
@@ -1,0 +1,21 @@
+cask "roblox" do
+  version "1.2.2.5360459,6e3d2e7c823142ae"
+  sha256 "eefc95fb85c888cbefaa16212779954c46482197442a82555f2e622135a28f06"
+
+  url "https://setup.rbxcdn.com/mac/version-#{version.csv.second}-Roblox.dmg",
+      verified: "setup.rbxcdn.com/"
+  name "Roblox"
+  desc "Online multiplayer game platform"
+  homepage "https://www.roblox.com/"
+
+  installer script: "Roblox.app/Contents/MacOS/Roblox"
+
+  uninstall quit:   "com.roblox.RobloxPlayer",
+            delete: "/Applications/Roblox.app"
+
+  zap trash: [
+    "~/Library/Preferences/com.Roblox.Roblox.plist",
+    "~/Library/Preferences/com.roblox.RobloxPlayer.plist",
+    "~/Library/Preferences/com.roblox.RobloxPlayerChannel.plist",
+  ]
+end
--- a/Casks/roblox.rb
+++ b/Casks/roblox.rb
@@ -1,8 +1,8 @@
 cask "roblox" do
   version "0.539.0.5390403,d937a525133a45cc"
-  sha256 "ec64a7a49192aaaa113cc2fe160f2198c1552805cee8cdbbfe910107d9d312c6"
+  sha256 "f68cfbccc5cf82d2acad713feeb9423b78e5668143500a44e63d00ab419ae48c"
 
-  url "https://setup.rbxcdn.com/mac/version-#{version.csv.second}-Roblox.dmg",
+  url "https://setup.rbxcdn.com/mac/version-#{version.csv.second}-RobloxPlayer.zip",
       verified: "setup.rbxcdn.com/"
   name "Roblox"
   desc "Online multiplayer game platform"
@@ -20,7 +20,7 @@ cask "roblox" do
 
   auto_updates true
 
-  installer manual: "Roblox.app"
+  app "RobloxPlayer.app", target: "Roblox.app"
 
   uninstall quit:   "com.roblox.RobloxPlayer",
             delete: "/Applications/Roblox.app"
--- a/Casks/roblox.rb
+++ b/Casks/roblox.rb
@@ -8,6 +8,22 @@ cask "roblox" do
   desc "Online multiplayer game platform"
   homepage "https://www.roblox.com/"
 
+  livecheck do
+    url "https://clientsettingscdn.roblox.com/v1/client-version/MacPlayer"
+    regex(/version[._-]([^"]+)/i)
+    strategy :page_match do |page, regex|
+      client_version = page[regex, 1]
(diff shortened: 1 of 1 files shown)


bevanjkay:
@aaronliu0130 Unfortunately the script installer is interactive, so this would need to use  [`installer manual`](https://docs.brew.sh/Cask-Cookbook#installer-manual) unless you can find a way to suppress the GUI.

A maintainer or another contributor will be able to help with the livecheck.

bevanjkay:
@aaronliu0130 Unfortunately, the installer is interactive so this cask would need to use [`installer manual`](https://docs.brew.sh/Cask-Cookbook#installer-manual) unless you can find a way to suppress the GUI. 

A maintainer or another contributor will be able to help with the livecheck.

p-linnane:
I'm looking into this. We can get the latest URL from the location header in "https://www.roblox.com/download/client". I haven't been able to find anything that references the half of the version before the comma. Their release notes aren't helpful.

@bevanjkay Are we ok with just using what can be pulled from those headers for the version number?

aaronliu0130:
>> Also, the history of roblox.rb shows that it was removed and attempted to be readded a few times since. Not sure if we're opening a can of worms here.

> It appears that the refused ones were due to the submitter not correctly using an installer stanza and closed to stale.

After a lot of searching for a silent mode, I went ahead and changed the installer to manual.

I also realized that the installer gives you a dock icon i.e. pins the roblox icon to dock. Which file should I delete to remove this icon?

aaronliu0130:
>> We can get the latest URL from the location header in "https://www.roblox.com/download/client".

After looking at the delete pr, it seems like the download string can be parsed from https://clientsettingscdn.roblox.com/v1/client-version/MacPlayer, in the key `clientVersionUpload`, after the `version-`. However, the version string appears to be incorrect.

aaronliu0130:
I think the awaiting user reply label can be removed now.

aaronliu0130:
Is the part of the version before the comma important? The installer never seems to change for each version bump, and the app will auto-update itself. I also couldn't find a page that changes with it.

Casks/roblox.rb
```diff
@@ -0,0 +1,21 @@
+cask "roblox" do
+  version "1.2.2.5360459,6e3d2e7c823142ae"
+  sha256 "eefc95fb85c888cbefaa16212779954c46482197442a82555f2e622135a28f06"
+
+  url "https://setup.rbxcdn.com/mac/version-#{version.csv.second}-Roblox.dmg",
+      verified: "setup.rbxcdn.com/"
+  name "Roblox"
+  desc "Online multiplayer game platform"
+  homepage "https://www.roblox.com/"
+
```
bevanjkay, line 10: Pending https://github.com/Homebrew/brew/pull/13658 - here is a livecheck that will retrieve everything that we need.

```suggestion
  livecheck do
    url "https://clientsettingscdn.roblox.com/v1/client-version/MacPlayer"
    regex(/version[._-]([^"]+)/i)
    strategy :page_match do |page, regex|
      client_version = page[regex, 1]
      next if client_version.blank?

      cask = CaskLoader.load("roblox")
      download_url = "https://setup.rbxcdn.com/mac/version-#{client_version}-Roblox.dmg"
      main_version = Homebrew::Livecheck::Strategy::ExtractPlist.find_versions(cask: cask, url: download_url)[:matches].values.first
      next if main_version.blank?

      "#{main_version},#{client_version}"
    end
  end
```

aaronliu0130: Thanks! I will merge it all into one commit after the ci passes.

aaronliu0130: Thanks! Do I still have to merge them into one commit?

aaronliu0130: I'll probably merge the commits after the relevant pull is merged.

aaronliu0130:
Something really strange just happened. Roblox updated and now the version is the version we see at the appcast url
image
{"version":"0.538.0.5380364","clientVersionUpload":"version-17f793e683ec4fd2","bootstrapperVersion":"5380364"}

Casks/roblox.rb
aaronliu0130: I just updated the version and added a currently working livecheck, is this good?

aaronliu0130:
It has been 15 days, I don't see what's blocking this pull other than squashing

bevanjkay:
@aaronliu0130 I am seeing the latest version`1.2.0.5390403` in app.

```
	<key>CFBundleShortVersionString</key>
	<string>1.2.0.5390403</string>
```

aaronliu0130:
image
It's this version for me.

bevanjkay:
@aaronliu0130 Because the download is an installer, not the actual application, and we need to use `installer manual`, the installer is downloaded to the caskroom, and then must be run manually by the user.

aaronliu0130:
But the file in the screenshot IS the app itself

bevanjkay:
@aaronliu0130 Roblox is a bit tricky because we are only downloading the installer, not the actual application. I've had a look at trying to download the actual application but I'm not sure that it is working correctly. I will push the changes, and can you give it a test?

bevanjkay:
@aaronliu0130 The issue has been traced and will hopefully be fixed soon

aaronliu0130:
Wait, actually there's a Roblox.app inside the downloaded app's MacOS folder which IS signed.
image

aaronliu0130:
Nevermind, it was the installer.

aaronliu0130:
This probably isn't something a contributor can fix without reverting to the "run the installer app" way. The problem is that the upstream devs didn't bother to sign the actual app, they only signed the installer.

miccal:
Thank you for your submission @aaronliu0130, but we cannot accept it at this time because it does not pass our signature verification check:

PR #129220, opened 2022-08-05
Add Cider Cask

a1eaiactaest:

- [ ] `brew audit --cask --online <cask>` is error-free. **livecheck detects something else.**

- [ ] `brew audit --new-cask <cask>` worked successfully.  **same error as above**

### DIFF
--- a/Casks/cider.rb
+++ b/Casks/cider.rb
@@ -1,14 +1,13 @@
 cask "cider" do
   version "1.14.3"
-  sha256 "bc99ba04122f17cd0a8f4cd6413b13d83ec369d4c3a21cedb75fd6990144618f"
-   
-  url "https://github.com/ciderapp/Cider/releases/download/macos-beta/Cider.dmg"
+  sha256 :no_check
+
+  url "https://github.com/ciderapp/Cider/releases/download/macos-beta/Cider.dmg", verified: "https://github.com/ciderapp/Cider"
   name "Cider"
-  desc "An open source and community oriented Apple Music client for Windows, Linux, macOS, and more."
-  homepage "https://cider.sh"
+  desc "Open source and community oriented Apple Music"
+  homepage "https://cider.sh/"
 
   auto_updates true
 
   app "Cider.app"
 end
-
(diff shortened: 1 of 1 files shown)


miccal:
Thank you for your contribution to `homebrew-cask` @a1eaiactaest.

Could you please try to add a [`zap` stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap), thanks.

a1eaiactaest:
Added `zap` stanza. Sorry for a late reply. @miccal

Casks/cider.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  url "https://github.com/ciderapp/cider-releases/releases/download/v#{version}/Cider-#{version}-universal.dmg",
+      verified: "github.com/ciderapp/cider-releases"
+  name "Cider"
+  desc "Open source and community oriented Apple Music client"
+  homepage "https://cider.sh/"
+
+  livecheck do
+    url :url
+    strategy :github_latest
```
samford, line 13:
```suggestion
    url "https://github.com/ciderapp/cider-releases/releases?q=prerelease%3Afalse&expanded=true"
    regex(/href=.*?Cider[._-]v?(\d+(?:\.\d+)+[^"' >]*?)(?:[._-]universal)?\.dmg/i)
    strategy :page_match
```

The default `GithubLatest` regex only works with simple tags like `1.2.3`/`v1.2.3` and it won't work with the trailing beta suffix in this context. It's possible to resolve this issue by adding a regex but there's a little more to this situation.

Looking through recent releases, I saw at least two where files weren't provided for macOS (1.5.4-beta.96, 1.5.4-beta.89), despite releases before/after providing a dmg. Since we can't 100% guarantee that a release will provide a dmg file, there's something to be said for checking the first page of releases for matching dmg files instead of only the "latest" release in this particular context. We try to avoid this approach under normal circumstances but it makes sense here.

The regex in the suggestion above has to be a little loose about the optional suffix matching (e.g., `-beta.8`), so it may unexpectedly end up matching more than it should in the future but this works at the moment. It may need to be updated in the future if the situation changes but that's to be expected.

miccal:
Thank you @a1eaiactaest.

miccal (approved):

a1eaiactaest:
You're welcome @miccal.

PR #129381, opened 2022-08-09
gcc-arm-embedded: Update to 11.3.rel1 and package structure changes

Kreeblah:

Note: I removed the zap stanza because there doesn't seem to be a reasonable way to do it for this package without potentially impacting an installed copy of the gcc-aarch64-embedded cask once it's updated.  I believe the updated package for that one will also install into /Applications/ArmGNUToolchain/#{version}, but in a different subdirectory.

### DIFF
--- a/Casks/gcc-arm-embedded.rb
+++ b/Casks/gcc-arm-embedded.rb
@@ -52,7 +52,7 @@ cask "gcc-arm-embedded" do
             delete:  [
               "/Applications/ArmGNUToolchain/#{version}/arm-none-eabi",
             ],
-            rmdir: [
+            rmdir:   [
               "/Applications/ArmGNUToolchain/#{version}",
               "/Applications/ArmGNUToolchain",
             ]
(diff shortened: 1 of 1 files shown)


reitermarkus (dismissed):

reitermarkus:
> I removed the zap stanza because there doesn't seem to be a reasonable way to do it for this package without potentially impacting an installed copy of the gcc-aarch64-embedded cask once it's updated.

What exactly is left behind after uninstalling?

Kreeblah:
What's left behind is `/Applications/ArmGNUToolchain/#{version}/` and `/Applications/ArmGNUToolchain/`.  However, the `aarch64` package also installs into a different subdirectory within `/Applications/ArmGNUToolchain/#{version}/`, which is my concern.

reitermarkus:
> different subdirectory within `/Applications/ArmGNUToolchain/#{version}/

Can we remove only the subdirectory in the `#{version}` directory then? And `rmdir` the `/Applications/ArmGNUToolchain/#{version}` directory.

reitermarkus:
Ah, the subdirectory is already in `uninstall delete:`. So just add `uninstall rmdir: "/Applications/ArmGNUToolchain/#{version}"`.

reitermarkus (dismissed):

reitermarkus (approved):

PR #129877, opened 2022-08-17
defold: Add livecheck

p-linnane:

### DIFF
--- a/Casks/defold.rb
+++ b/Casks/defold.rb
@@ -8,6 +8,12 @@ cask "defold" do
   desc "Game engine for development of desktop, mobile and web games"
   homepage "https://defold.com/"
 
+  livecheck do
+    url "https://github.com/defold/defold/releases/"
+    regex(%r{href=.*?/tag/(\d+(?:\.\d+)+).*?stable</a>}i)
+    strategy :page_match
+  end
+
   auto_updates true
   conflicts_with cask: [
     "homebrew/cask-versions/defold-beta",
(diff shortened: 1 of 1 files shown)


Casks/defold.rb
```diff
@@ -8,6 +8,12 @@
   desc "Game engine for development of desktop, mobile and web games"
   homepage "https://defold.com/"
 
+  livecheck do
+    url "https://github.com/defold/defold/releases/"
+    regex(%r{href=.*?/tag/(\d+(?:\.\d+)+).*?stable</a>}i)
```
samford, line 13:
```suggestion
    url "https://github.com/defold/defold/releases?q=prerelease%3Afalse"
    regex(%r{href=.*?/tag/v?(\d+(?:\.\d+)+)["' >]}i)
```

When checking the first page of releases on GitHub, we should use the `prerelease:false` query to improve our chances of finding stable releases. That is to say, release pages only contain ten entries but currently there are three pre-release entries on the first page. Beta releases aren't marked as pre-release, so the ten releases are split between stable and beta even with this search filter. It's not perfect in this instance but it's better than the default (i.e., it's unlikely that the first page will be nothing but beta releases).

Additionally, the stable tags in this repository use a format like `1.2.3`, so we can simply use the typical regex for matching versions from tag URLs on GitHub release pages (i.e., the default regex from the `GithubLatest` strategy).

p-linnane: Thanks as always for the info @samford! I'll incorporate that moving forward.

samford (approved):
I forgot to add an explanatory comment in my previous suggestion (since this is an unusual situation) but I pushed a commit to address it. This should be good to go now. Thanks, @p-linnane!

SMillerDev (approved):

PR #130042, opened 2022-08-19
julia: fix arch conditional

miccal:


jan-xyz:
nice! Thanks for the quick fix ❤️

Casks/julia.rb
```diff
@@ -1,6 +1,6 @@
 cask "julia" do
   arch arm: "aarch64", intel: "x64"
-  midfix = on_arch_conditional arm: arch
```
jan-xyz, line 3: I think this needs to be `aarch` (double `a`) from the link: https://julialang-s3.julialang.org/bin/mac/aarch64/1.8/julia-1.8.0-macaarch64.dmg

adavidzh:
I wonder if this also fixes what I tried to address in https://github.com/Homebrew/homebrew-cask/pull/130044.

miccal:
Fixed by https://github.com/Homebrew/homebrew-cask/pull/130045

PR #130261, opened 2022-08-23
Use arch and macOS version DSLs

Rylan12:
This PR is a follow-up to https://github.com/Homebrew/homebrew-cask/pull/130260 and the other DSL change PRs. This PR updates the casks that contain either unusual conditionals or both arch _and_ macOS version conditionals. These casks couldn't be updated with `brew style --fix` (or, if they could be updated, needed to be manually checked).

I've used `brew info --json=v2` on all of these casks to ensure that their JSON files remain the same before and after this PR. I tested on Monterey (ARM and Intel), Mojave (Intel), High Sierra (Intel) and Sierra (Intel) to ensure that there were no unexpected changes.

### DIFF
--- a/Casks/anki.rb
+++ b/Casks/anki.rb
@@ -2,22 +2,19 @@ cask "anki" do
   arch arm: "apple", intel: "intel"
 
   version "2.1.54"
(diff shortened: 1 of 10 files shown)


Rylan12:
The style issues will be fixed by https://github.com/Homebrew/brew/pull/13749. Once that's merged, CI can be re-run here and this will be ready

miccal (approved):

PR #130366, opened 2022-08-24
Add BookWright v2.4.1

thom4parisot:

### DIFF
--- a/Casks/bookwright.rb
+++ b/Casks/bookwright.rb
@@ -1,0 +1,26 @@
+cask "bookwright" do
+  version "2.4.1"
+  sha256 "a013943d327be16da1b8028ffbe3a1fc683eacce238f6ee10b4decb0768fb508"
+
+  url "https://software.blurb.com/bookwright_v#{version.major}/#{version}/BookWright.dmg"
+  name "bookwright"
+  desc "Make a book with this tool and the Blurb printing service"
+  homepage "https://www.blurb.com/bookwright"
+
+  livecheck do
+    # url "https://www.blurb.com/bookwright-install-mac?platform=MacOS&bwversion=#{version}&download=true"
+    # regex(/href=.+www.blurb.com\/booksmart_download\/bookwright?platform=mac&bwversion=(\d+\.\d+\.\d+)/i)
+
+    skip "Unable to get a direct link, it is added via JavaScript within the page."
+  end
+
+  app "BookWright.app"
+
+  zap trash: [
+    "~/Library/Application Support/com.blurb.bookwright",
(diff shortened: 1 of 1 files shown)


Casks/bookwright.rb
```diff
@@ -0,0 +1,26 @@
...
+  url "https://software.blurb.com/bookwright_v#{version.major}/#{version}/BookWright.dmg"
+  name "bookwright"
+  desc "Make a book with this tool and the Blurb printing service"
+  homepage "https://www.blurb.com/bookwright"
+
+  livecheck do
+    # url "https://www.blurb.com/bookwright-install-mac?platform=MacOS&bwversion=#{version}&download=true"
+    # regex(/href=.+www.blurb.com\/booksmart_download\/bookwright?platform=mac&bwversion=(\d+\.\d+\.\d+)/i)
+
+    skip "Unable to get a direct link, it is added via JavaScript within the page."
```
samford, line 14:
```suggestion
  # If this URL only provides information for a specific major version, this
  # check will be unable to detect new major versions and these will need to
  # be identified manually outside of livecheck. In that scenario, the check
  # should theoretically work as expected again after updating the cask version.
  livecheck do
    url "https://www.blurb.com/bookwright_v#{version.major}/versions.json"
    regex(/["']version["']:\s*?["']v?(\d+(?:\.\d+)+)["']/i)
```

The version information can be found in the `versions.json` file that's loaded on the homepage (this is what's used to generate the link to the download page for the latest version). The only issue is that the link to `versions.json` may be tied to a major version (i.e., if `bookwright_v2` refers to the current major version of 2 and not something else). If this is the case, then this check won't be able to identify new major versions and we'll have to rely on a human to manually identify the latest version and create a version bump PR. Assuming the next URL uses `bookwright_v3` for a major version of 3, we would only need to update the cask version and the check should work as expected again without requiring any changes to the `livecheck` block.

For what it's worth, `/bookwright_v2/versions.json` is hardcoded in the `main` JavaScript file (currently `main.min-d27d6d0d4358e394bbae302eeae33581afa76722df56d56749d460524d4c84b3.js`) but its filename contains a trailing hash that will change when it's updated. It's unfortunately not possible for us to catch major version changes by checking that JS file for the current `versions.json` URL and then running the normal check [in a `strategy` block] using the matched URL. The suggested approach above seems to be the best we can do for now, unless there's an alternative that I've overlooked.

miccal:
Thank you @thom4parisot.

miccal (approved):

Casks/bookwright.rb
thom4parisot: Oh, smart. Thanks 🙂

PR #130563, opened 2022-08-28
Add two new labels.

reitermarkus:
About 200 casks are still using `appcast`. This should make them easier to spot if they come up in PRs.

Also, a bunch of casks are still missing descriptions, this should make it a bit more obvious to contributors.


p-linnane:
This will be helpful. I know I've personally missed adding descriptions when updating older casks.

miccal (approved):

PR #130623, opened 2022-08-29
Fix android-file-transfer zap

malob:
Removed `rmdir: "~/Library/Application Support/Google/"` from `zap`, since this folder is used by way more things than `android-file-transfer`, including things like Google Chrome and Google Drive.

### DIFF
--- a/Casks/android-file-transfer.rb
+++ b/Casks/android-file-transfer.rb
@@ -18,9 +18,8 @@ cask "android-file-transfer" do
   app "Android File Transfer.app"
 
   zap trash: [
-        "~/Library/Application Support/Google/Android File Transfer",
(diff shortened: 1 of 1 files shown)


miccal:
Thank you @malob, but `rmdir` will only remove the folder if it is empty, so this is correct.

malob:
I disagree. This caused a bunch of folder to be deleted from my Google Drive. Folders that, although empty, I very much wanted to be there.

It was also scary since the Google Drive app on my computer started repeatedly reporting that files were being deleted. Initial I had no idea why, and it was very confusing and concerning. None of this should happen because I uninstalled on unrelated application.

miccal:
I see the problem.

miccal (approved):

malob:
Thanks! 

Perhaps there should be an automated check of some kind for this? Counting on humans to catch a trailing slash that shouldn’t be there is a dangerous game 😛

PR #131072, opened 2022-09-05
Fix regex match in list_login_items_for_app

jawshooah:
Prior to the fix:

```
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/developer/bin/list_login_items_for_app:37:in `process_args': undefined method `match?' for /^-+h(?:elp)?$/:Regexp (NoMethodError)
Did you mean?  match
	from /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/developer/bin/list_login_items_for_app:65:in `<main>'
```

### DIFF
--- a/developer/bin/list_login_items_for_app
+++ b/developer/bin/list_login_items_for_app
@@ -34,7 +34,7 @@ def usage
 end
 
 def process_args
-  if /^-+h(?:elp)?$/.match?(ARGV.first)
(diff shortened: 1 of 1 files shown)


jawshooah:
Turns out I was running an ancient 2.3.x version of Ruby, and `Regexp#match?` was introduced in 2.4. Disregard

PR #131226, opened 2022-09-09
Add Rokoko Studio v1.20

ethanproia:
Real-time motion capture solution


p-linnane:
Please ensure the cask file is located in the `Casks` directory.

p-linnane:
CI is failing because the application is not `codesigned`. Please contact the upstream developer and ask them to `codesign` their application. Otherwise this will prevent the application from running on Apple Silicon machines without disabling Gatekeeper (which is not ideal).

miccal:
Thank you for your submission @ethanproia, but we cannot accept it at this time because it does not pass our signature verification check:

PR #131227, opened 2022-09-09
Add AutoRemesher v1.0.0-beta.3

ethanproia:
Automatic quad geometry retopology/remesh.
`Livecheck` stanza may need fixing

### DIFF
--- a/Casks/autoremesher.rb
+++ b/Casks/autoremesher.rb
@@ -1,6 +1,6 @@
 cask "autoremesher" do
   version "1.0.0-beta.3"
-  sha256 :no_check
+  sha256 "be8450f69d8c52e1b7cd9d3e7ecacfe6bd64fa7420c6fa3944949509338a85a7"
 
   url "https://github.com/huxingyi/autoremesher/releases/download/#{version}/autoremesher-#{version}.dmg"
   name "AutoRemesher"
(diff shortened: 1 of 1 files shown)


Casks/autoremesher.rb
```diff
@@ -0,0 +1,16 @@
+cask "autoremesher" do
+  version "1.0.0-beta.3"
+  sha256 :no_check
```
p-linnane, line 3:
```suggestion
  sha256 "be8450f69d8c52e1b7cd9d3e7ecacfe6bd64fa7420c6fa3944949509338a85a7"
```

Casks/autoremesher.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  url "https://github.com/huxingyi/autoremesher/releases/download/#{version}/autoremesher-#{version}.dmg"
+  name "AutoRemesher"
+  desc "Your gateway to Megascans and Metahumans"
+  homepage "https://github.com/huxingyi/autoremesher/"
+
+  livecheck do
+    url "https://github.com/huxingyi/autoremesher/releases/"
+    strategy :header_match
```
p-linnane, line 12:
```suggestion
    url :url
    regex(/^v?(\d+(?:\.\d+)+(?:-beta\.?\d*)?)$/i)
```

p-linnane:
CI is failing because the application is not `codesigned`. Please contact the upstream developer and ask them to `codesign` their application. Otherwise this will prevent the application from running on Apple Silicon machines without disabling Gatekeeper (which is not ideal).

miccal:
Thank you for your submission @ethanproia, but we cannot accept it at this time because it does not pass our signature verification check:

PR #131385, opened 2022-09-10
kindle-previewer: Change livecheck

p-linnane:
Existing livecheck is broken due to changes on Amazon's side. The page requests you contact Amazon about accessing the page in an automated manner instead of serving the content.

I've changed the strategy to `:extract_plist` which is obviously not ideal, but works.

### DIFF
--- a/Casks/kindle-previewer.rb
+++ b/Casks/kindle-previewer.rb
@@ -1,5 +1,5 @@
 cask "kindle-previewer" do
-  version "3.67.0"
+  version "3.67"
   sha256 :no_check
 
   url "https://d2bzeorukaqrvt.cloudfront.net/KindlePreviewerInstaller.pkg",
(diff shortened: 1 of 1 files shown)


p-linnane:
Would love some insight from @samford here.

samford (commented):
The existing check works if the user agent is set to `:browser` but this won't be possible until after I create a PR for the `options` feature (along with the ancillary work to handle a `user_agent` option). I have everything implemented locally but I've been waiting for the GSoC `livecheck` PR to wrap up to rebase my work and create a PR (though I should simply create a draft PR in the interim time, for the sake of discussion).

Regarding an `ExtractPlist` approach, the pkg file is 500+ MB and the install count isn't too high (30 days: 47, 90 days: 134, 365 days: 622), so I kind of lean towards just skipping this for the time being. The GSoC project should be over by the beginning of October (and the related PR should be finished sometime before then), so the change in this PR will likely only matter for the next few weeks (i.e., we would only miss 1-2 `kindle-previewer` versions, if any).

However, one can also make the argument that we would only be using `ExtractPlist` for the same amount of time, so I'll defer to the cask maintainers whether it would be worth it to them. I've added a suggestion if we want to skip this in the end but I would wait to do anything until a cask maintainer weighs in on this.

PR #131403, opened 2022-09-10
thonny-xxl: Change livecheck

p-linnane:

### DIFF
--- a/Casks/thonny-xxl.rb
+++ b/Casks/thonny-xxl.rb
@@ -9,7 +9,9 @@ cask "thonny-xxl" do
   homepage "https://thonny.org/"
 
   livecheck do
-    cask "thonny"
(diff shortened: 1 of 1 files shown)


Casks/thonny-xxl.rb
```diff
@@ -9,7 +9,9 @@
   homepage "https://thonny.org/"
 
   livecheck do
-    cask "thonny"
+    url "https://github.com/thonny/thonny/releases"
```
samford, line 12: @p-linnane Just as a heads up for the future, stable casks where it's necessary to check the first page of releases for versions should also use the `prerelease:false` query (i.e., `/releases?q=prerelease%3Afalse`). There may be rare situations where we need to omit it but it should otherwise be the default.

PR #131423, opened 2022-09-11
Update Apifox from 2.1.34 to 2.1.35

fangzhengjin:

### DIFF
--- a/Casks/apifox.rb
+++ b/Casks/apifox.rb
@@ -1,9 +1,9 @@
 cask "apifox" do
   arch arm: "-macOS-arm64"
 
-  version "2.1.34"
-  sha256 arm:   "c8a5741801c15dbf078a5d99f29e3884aa9c03d249f54bed6dd5136a11d33532",
-         intel: "4b6ebce6da81ab38c84413dea010ef7e4b9432e82b9bba7d89c1ba351f069d0a"
+  version "2.1.35"
+  sha256 arm:   "2a8b286c22b7ecbab71ee91801187b0b3caf798597b8c70ca7af6a68197b5b9c",
+         intel: "a7ebff4967490dd17ca768319684dfb2efdeef09249a396ea0873a7d3eb8b224"
 
   url "https://cdn.apifox.cn/download/#{version}/Apifox#{arch}-#{version}.dmg"
   name "Apifox"
--- a/Casks/apifox.rb
+++ b/Casks/apifox.rb
@@ -11,7 +11,7 @@ cask "apifox" do
   homepage "https://www.apifox.cn/"
 
   livecheck do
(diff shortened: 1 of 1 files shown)


Casks/apifox.rb
```diff
@@ -1,17 +1,17 @@
...
+         intel: "a7ebff4967490dd17ca768319684dfb2efdeef09249a396ea0873a7d3eb8b224"
 
   url "https://cdn.apifox.cn/download/#{version}/Apifox#{arch}-#{version}.dmg"
   name "Apifox"
   desc "Platform for API documentation, debugging, Mock and testing"
   homepage "https://www.apifox.cn/"
 
   livecheck do
-    url "https://cdn.apifox.cn/download/mac/latest-mac.yml"
+    url "https://cdn.apifox.cn/download/mac/latest-mac.yml?noCache=" + (Time.new.to_f * 1000).to_i.to_s
```
samford, line 14:
```suggestion
    url "https://cdn.apifox.cn/download/mac/latest-mac.yml?noCache=#{Time.new.to_i * 2}"
```

Is there a reason behind this specific approach?

Assuming the `noCache` value is just an epoch date and this code is trying to create a far-future date, this approach seems excessive (i.e., it leads to a date around 54680-06-27). Since the number is so large, I also don't see any notable value in the precision of converting the time to a float and then to an integer in this particular scenario. It seems like we could get away with something like `Time.new.to_i * 2` (i.e., this gave the same result in testing).

Past that, we can avoid the explicit `to_s` call if we simply use string interpolation here.

miccal (approved):

Casks/apifox.rb
samford: @fangzhengjin Thanks for the follow-up. Does this specifically need to be a 13-digit value or is it simply a Unix epoch integer + three trailing digits from the fractional time? [This can vary in length based on the length of the Unix epoch integer, so I assume this wouldn't always be a fixed 13-digit value.]

From what I'm seeing in testing, a `noCache` value of `Time.new.to_i` works the same as `(Time.new.to_f * 1000).to_i.to_s` and they both return the latest version. Unless the upstream server expects a specific `noCache` value format, it seems like the value simply needs to be unique to avoid caching and `Time.new.to_i` would be fine.

Is the `noCache` parameter something that's supported by the upstream server (i.e., does it expect a value with a specific format) or was this simply something you added? I tested out a different parameter (e.g., `latest-mac.yml?aTest=1`) and it also returns the latest version, which seems to suggest that this is just about needing a unique value to avoid caching.

Could you explain a bit more about how you found this URL and query string parameter and if there are any specific requirements we need to meet? I'm in favor of meeting upstream requirements when necessary but I would like to avoid unnecessary complexity otherwise.

PR #131543, opened 2022-09-13
spotify: Add sha256

timvisher:
This was marked as `:no_check` [by automation][1] rather than by a human so it seems reasonable to convert it to sha256 sum now.

[1]: https://github.com/timvisher/homebrew-cask/commit/c8cd897435157aa7d79767c667012b3f14f5895b

- [x] `brew audit --cask --online spotify` is error-free.
- [x] `brew style --fix spotify` reports no offenses.

~Additionally, **if adding a new cask**:~ N/A

### DIFF
--- a/Casks/spotify.rb
+++ b/Casks/spotify.rb
@@ -2,7 +2,7 @@ cask "spotify" do
   arch arm: "ARM64"
 
   version "1.1.93.896,3ae3b4f3,10"
-  sha256 :no_check
+  sha256 "354427a2e113bc6318362ea70f81ae09b8e84205ebb9f1acc7bc9a292adf52d1"
 
   url "https://download.scdn.co/Spotify#{arch}.dmg",
       verified: "download.scdn.co/"


timvisher:
@bevanjkay Oops! Thought I'd done my due diligence but I did miss this. Thanks for taking the time to explain it.

Would you be open to:

1. A change to the README in the Style Guide section that explicitly calls out the requirement for Versioned Artifacts when providing a sha256? I would've seen that and maybe future contributors would be saved the effort.
2. A change to `brew audit --cask` that would attempt to catch when a sha256 sum is provided for an unversioned Artifact? Since I ran `brew audit --cask` and it passed I had assumed I was good and given that this is a somewhat common situation (719 matches at present) it would probably add some value over the lifetime of the project.

SMillerDev:
Additionaly, I think adding an audit for unversioned artifacts not using `:latest` or at the very least `:no_check` would be a good idea.

bevanjkay:
@SMillerDev @timvisher 
There will be a number of casks using `:latest` that are versioned. Usually these are combined with a `livecheck` strategy of `:extract_plist` or sometimes `:header_match` can find the version using the `Content-Disposition` header, but a direct link cannot be provided.
In these instances, you may receive the wrong version when you first download the cask, however `brew outdated` or `brew upgrade` will correctly notify users/upgrade the application when there is a newer version installed than what they currently have (even if the exact version is incorrect).
In saying this, there has been an update to Brew recently that stores a `sha256` value for casks with `version :latest` that will do a download and comparison when running `brew outdated --greedy-latest` or `brew upgrade --greedy-latest` - so there is a mechanism to know that a cask with an unversioned `url` has been updated upstream.

PR #131597, opened 2022-09-14
arduino: various updates

miccal:
See https://github.com/Homebrew/homebrew-cask/pull/131594


wilkmaciej:
Probably should add new cask arduino-ide.
Version 2 of arduino uses different config file locations.

bevanjkay (approved):

PR #132167, opened 2022-09-23
abricotine: Change livecheck

p-linnane:


p-linnane:
This is an attempt to fix https://github.com/Homebrew/brew/issues/13853. GitHub support has confirmed that release pages will render their asset list collapsed by default, so our existing approach no longer works. This is 1 of 2 examples I'm opening utilizing this new method. Please poke holes in this so we can hopefully figure out a new solution if this doesn't work.

p-linnane:
Closing this one as after additional testing it's essentially just a roundabout way of utilizing the `:git` strategy.

PR #132168, opened 2022-09-23
aleph-one: Change livecheck

p-linnane:

### DIFF
--- a/Casks/aleph-one.rb
+++ b/Casks/aleph-one.rb
@@ -11,7 +11,7 @@ cask "aleph-one" do
   livecheck do
     url :url
     strategy :github_latest
-    regex(%r{href=.*?/AlephOne-(\d+)-Mac\.dmg}i)
+    regex(%r{expanded_assets/release[._-](\d+(?:\d+)+)}i)
   end
 
   app "Aleph One.app"
--- a/Casks/aleph-one.rb
+++ b/Casks/aleph-one.rb
@@ -10,7 +10,7 @@ cask "aleph-one" do
 
   livecheck do
     url :homepage
-    regex(%r{expanded_assets/release[._-](\d(?:\d+))}i)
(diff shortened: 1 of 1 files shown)


p-linnane:
See https://github.com/Homebrew/homebrew-cask/pull/132167#issuecomment-1255710488. This example utilizes `:github_latest`.

Casks/aleph-one.rb
```diff
@@ -11,7 +11,7 @@
   livecheck do
     url :url
     strategy :github_latest
```
samford, line 13:
```suggestion
    url :homepage
```

Casks/aleph-one.rb
```diff
@@ -11,7 +11,7 @@
...
-    regex(%r{href=.*?/AlephOne-(\d+)-Mac\.dmg}i)
+    regex(%r{expanded_assets/release[._-](\d(?:\d+))}i)
```
samford, line 14:
```suggestion
    regex(%r{href=.*?/AlephOne[._-]v?(\d+(?:\.\d+)*)[._-]Mac\.dmg}i)
```

samford (commented):
We have a couple options in this particular case. The homepage links to the latest dmg file on GitHub, so we can easily check that. Otherwise, we could continue using the `GithubLatest` strategy by using a modified version of the default strategy regex that accounts for the `release-` prefix and date-based version format in tags (i.e., `%r{href=["']?[^"' >]*?/tag/release[._-]v?(\d+(?:\.\d+)*)["' >]}i`). That said, I've added suggestions that update the `livecheck` block to check the homepage (and bring the existing regex in line with prevailing standards).

In general, the `expanded_assets` URL simply uses the Git tag, so it's effectively no different from matching the version from the tag URL like the `GithubLatest` strategy does by default. I'll explain this a bit more in the related issue, though.

samford (approved):
Thanks!

PR #132245, opened 2022-09-25
Update qq from 6.8.3 to 6.8.5

wjlmelody:


samford (commented):
The homepage HTML only provides text like "v 6.8.3" (outdated but also incomplete with respect to the dmg version). The up to date information (e.g., used for the download button) comes from a secondary request to `/rainbow/ntQQDownload/` after the page loads. That response contains some JavaScript that includes the URL to the dmg file.

With this in mind, we only need to update the `livecheck` block URL, as the existing regex works fine in this context. I've pushed a commit that takes care of this, so it should pass CI this time around.

miccal (approved):

PR #132567, opened 2022-09-29
sage 9.7,1.5

miccal:


samford (approved):
I'm going to try to go through some more of these `livecheck` blocks tomorrow but I'm good with skipping this for the moment (for the sake of the version bump) 👍

PR #132610, opened 2022-09-30
far2l 2.4.1

yurikoles:

---
Livecheck broke again :(


samford (approved):
The upstream situation continues to be unsatisfactory but I've pushed a commit to update the `livecheck` block to check versions from tags instead. The tags don't mention whether a given version is alpha/beta/etc. (e.g., `v_2.4.1` for 2.4.1 beta) and anyone updating this will need to continue to keep that in mind but this will at least let us know that a new version is available.

PR #132607, opened 2022-09-30
Update hydrus-network from 499 to 500

Amorymeltzer:
Created with `brew bump-cask-pr`.

PR #132706, opened 2022-10-02
Update tabby to 1.0.184

fblupi:

### DIFF
--- a/Casks/tabby.rb
+++ b/Casks/tabby.rb
@@ -13,8 +13,8 @@ cask "tabby" do
   homepage "https://eugeny.github.io/tabby/"
 
   livecheck do
-    url :url
-    strategy :github_latest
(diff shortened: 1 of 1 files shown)


fblupi:
I can't understand why the live check is complaining about the version. The latest version is 1.0.184, the version 5.0.0 doesn't exist.

miccal (approved):

samford:
As mentioned in https://github.com/Homebrew/homebrew-cask/pull/132607#pullrequestreview-1127580680, I intend to make checking the `location` header the default behavior for the `GithubLatest` strategy in the future but there's a bit of technical work to get through first. [Some existing checks identify the version from the release page HTML and I would like to optionally support that use case but I'll have to look into how many actually require this. If it's not very many, I could just switch those to `PageMatch` instead.]

PR #132960, opened 2022-10-06
Fix 'Cask definition error' in "appcleaner" Cask

bastienlimbour:

-----

I propose this solution because I encountered this error when searching cask with RayCast on Mac :

Error : Cask 'appcleaner' definition is invalid: 'livecheck' stanza may only appear once.

Hope this is a good solution..


p-linnane:
Fixed in https://github.com/Homebrew/homebrew-cask/commit/28613b5c947dc859a2e3ed9fbfd3be8b81594071.

PR #133313, opened 2022-10-12
switchhosts: update livecheck

samford:

-----

The existing `livecheck` block for `switchhosts` identified the version from dmg files in the "latest" GitHub release's assets list. However, GitHub now omits the asset list HTML from release pages and it's fetched separately when the assets list is expanded (see https://github.com/Homebrew/brew/issues/13853), so this check is currently broken.

In this case, unfortunately the upstream repository doesn't include the build number in the tag, so it's necessary to match the version information from the filename. This PR updates the `livecheck` block to find the tag for the latest release (using the `HeaderMatch` strategy to identify it in the `location` header), fetch the asset list HTML for that release, and match versions in the assets list HTML. This works for now but I intend to come up with a better approach (e.g., creating a strategy for this instead of needing a hefty `strategy` block for each of these) after all of these broken checks are working again.

### DIFF
--- a/Casks/switchhosts.rb
+++ b/Casks/switchhosts.rb
@@ -12,9 +12,21 @@ cask "switchhosts" do
   homepage "https://oldj.github.io/SwitchHosts/"
 
   livecheck do
-    url :url
-    strategy :github_latest
+    url "https://github.com/oldj/SwitchHosts/releases/latest"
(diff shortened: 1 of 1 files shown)


p-linnane:
There are a number of affected casks that still don't have a viable means of livechecking. Any issue if I utilize this for the ones where matching from the asset list is absolutely required?

PR #133600, opened 2022-10-15
Update uvtools.rb: Add live check and dependencies

sn4k3:

### DIFF
--- a/Casks/uvtools.rb
+++ b/Casks/uvtools.rb
@@ -9,6 +9,16 @@ cask "uvtools" do
   name "UVtools"
   desc "MSLA/DLP, file analysis, calibration, repair, conversion and manipulation"
   homepage "https://github.com/sn4k3/UVtools"
+  license "AGPL-3.0"
+   
+  livecheck do
+    url "https://api.github.com/repos/sn4k3/UVtools/releases/latest"
+	strategy :page_match
+    regex(/"tag_name":\s*"v?(\d+\.\d+\.\d+)"/im)
+  end
+  
+  depends_on macos: ">= :catalina"
+  depends_on formula: "mono-libgdiplus"
 
   app "UVtools.app"
 
--- a/Casks/uvtools.rb
+++ b/Casks/uvtools.rb
@@ -11,9 +11,8 @@ cask "uvtools" do
   homepage "https://github.com/sn4k3/UVtools"
 
   livecheck do
-    url "https://api.github.com/repos/sn4k3/UVtools/releases/latest"
-    strategy :page_match
-    regex(/"tag_name":\s*"v?(\d+\.\d+\.\d+)"/im)
+    url :url
+    strategy :github_latest
   end
 
   auto_updates true
--- a/Casks/uvtools.rb
+++ b/Casks/uvtools.rb
@@ -10,13 +10,13 @@ cask "uvtools" do
   desc "MSLA/DLP, file analysis, calibration, repair, conversion and manipulation"
   homepage "https://github.com/sn4k3/UVtools"
   license "AGPL-3.0"
-   
+
   livecheck do
     url "https://api.github.com/repos/sn4k3/UVtools/releases/latest"
     strategy :page_match
     regex(/"tag_name":\s*"v?(\d+\.\d+\.\d+)"/im)
   end
-  
+
   depends_on macos: ">= :catalina"
   depends_on formula: "mono-libgdiplus"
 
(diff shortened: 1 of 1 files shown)


Casks/uvtools.rb
```diff
@@ -10,6 +10,15 @@
...
   homepage "https://github.com/sn4k3/UVtools"
 
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
+
+  auto_updates true
+  depends_on macos: ">= :catalina"
+  depends_on formula: "mono-libgdiplus"
```
miccal, line 20: Why is this dependency required for the `app`? This is only required when building from source. no?

sn4k3: App runs fine without `mono-libgdiplus` but a little function (rarely used by users) require that lib in order to run (**Export layers to GIF**). If not installed app will not crash but will show an error message that a type is missing and prevent the execution of the tool, it's really not critical. You think it's best to remove it so it not bloat the install and users who want to execute that i can instruct them to install `mono-libgdiplus`?
(Also not required to build)

image

image

miccal: `mono-libgdiplus` is a heavy dependency, so I think it is best not to add it:
```
|-> brew info mono-libgdiplus
==> mono-libgdiplus: stable 6.1 (bottled)
GDI+-compatible API on non-Windows operating systems
https://www.mono-project.com/docs/gui/libgdiplus/
Not installed
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/mono-libgdiplus.rb
License: MIT
==> Dependencies
Build: pkg-config ✔
Required: cairo ✔, fontconfig ✔, freetype ✔, gettext ✔, giflib ✔, glib ✔, jpeg-turbo ✔, libexif ✘, libpng ✔, libtiff ✔, pango ✔
==> Analytics
install: 2,009 (30 days), 7,489 (90 days), 21,406 (365 days)
install-on-request: 2,012 (30 days), 7,492 (90 days), 21,410 (365 days)
build-error: 0 (30 days)
```

sn4k3: That's ok for me 👍

miccal (approved):

sn4k3:
@miccal One question, i notice livecheck has been removed on your [commit](https://github.com/Homebrew/homebrew-cask/pull/133600/commits/43f860112a66f2a89096dce5a3bc7ba789b50a46) ?

miccal:
Yes, for GitHub releases the `livecheck` is automatic and not needed.

sn4k3:
> Yes, for GitHub releases the `livecheck` is automatic and not needed.

Ok, didn't know about that. Thanks!

sn4k3:
@miccal Sorry to bring the topic here but I have a few questions to see if i can improve the deployment of UVtools to the ease of use.

My app is not codesigned, this means:

- Security flag to trigger and users must go security and allow it to run (some users fail to do this), on my auto installer script I bypass that by [removing the quarantine attribute](https://github.com/sn4k3/UVtools/blob/843c3e817e812cfb291c24d65063ee9e099e4f63/Scripts/install-uvtools.sh#L56).  
**Question:** Is possible to run such line on the post install cask or it is against the deployment rules?
- `arm64` app doesn't run due not codesign app. I bypass that by [creating a run file on user desktop](https://github.com/sn4k3/UVtools/blob/843c3e817e812cfb291c24d65063ee9e099e4f63/Scripts/install-uvtools.sh#L59-L85) it end in calling `dotnet UVtools.dll` instead which is able to run the app just fine.  
**Question:** Is possible to do the same on this cask if arm64? Also it would require `dotnet-sdk` dependency on `arm64` only

miccal:
> @miccal Sorry to bring the topic here but I have a few questions to see if i can improve the deployment of UVtools to the ease of use.
> 
> My app is not codesigned, this means:
> 
> * Security flag to trigger and users must go security and allow it to run (some users fail to do this), on my auto installer script I bypass that by [removing the quarantine attribute](https://github.com/sn4k3/UVtools/blob/843c3e817e812cfb291c24d65063ee9e099e4f63/Scripts/install-uvtools.sh#L56).
>   **Question:** Is possible to run such line on the post install cask or it is against the deployment rules?

We do not allow this, but users can install with the `--no-quarantine` flag if they wish.

> * `arm64` app doesn't run due not codesign app. I bypass that by [creating a run file on user desktop](https://github.com/sn4k3/UVtools/blob/843c3e817e812cfb291c24d65063ee9e099e4f63/Scripts/install-uvtools.sh#L59-L85) it end in calling `dotnet UVtools.dll` instead which is able to run the app just fine.
>   **Question:** Is possible to do the same on this cask if arm64? Also it would require `dotnet-sdk` dependency on `arm64` only

I am afraid that we would not allow this either -- we do not bypass the built-in macOS security features in Casks.

PR #133704, opened 2022-10-18
Improve CI coverage for Casks using `on_os` blocks

Bo98:
Should hopefully reduce the current high frequency of broken cask readability on older OS versions.

* Run `brew readall` in CI to make sure casks are actually readable on all OS versions.
* Adjust the CI matrix logic to consider `on_os` blocks rather than just `MacOS.version` (for full install testing).
* While we're here, adjust the runner allocation to always prioritise `macos-12` over `macos-11` (now that capacity has increased).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,16 +117,21 @@ jobs:
         id: gems
         run: brew install-bundler-gems
 
+      - name: Run brew readall ${{ matrix.tap }}
+        id: readall
+        run: brew readall '${{ matrix.tap }}'
+        if: always() && steps.gems.outcome == 'success'
+
       - name: Run brew style ${{ matrix.tap }}
         run: brew style '${{ matrix.tap }}'
-        if: always() && steps.gems.outcome == 'success' && !matrix.cask
+        if: always() && steps.readall.outcome == 'success' && !matrix.cask
 
       - name: Run brew fetch --cask ${{ matrix.cask.token }}
         id: fetch
         run: |
           brew fetch --cask --retry --force '${{ matrix.cask.path }}'
         timeout-minutes: 30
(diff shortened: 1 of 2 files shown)


Bo98:
The extra step adds ~45s.

Rylan12 (approved):
Thanks, @Bo98!

Looks good to me. I think there's still the potential for an edge case with `:or_older` and `:or_newer` since OS versions affected by those might not be named directly.

One way to combat this might be just to run every OS version whenever we see any `on_*` block (except `on_arm` and `on_intel`). That way we're always being thorough.

On the other hand, I do think that the `brew readall` fix will help a lot so it may be fine to just stick with this for now.

I'm going to go ahead and merge and we can continue to discuss if desired

SMillerDev (approved):

MikeMcQuaid (approved):
Great work!

Bo98:
> Looks good to me. I think there's still the potential for an edge case with `:or_older` and `:or_newer` since OS versions affected by those might not be named directly.
> 
> One way to combat this might be just to run every OS version whenever we see any `on_*` block (except `on_arm` and `on_intel`). That way we're always being thorough.

This currently checks `on_monterey` and `on_bigsur`.

* `on_catalina :on_newer` - not an issue since it should be same result on Monterey and Big Sur so only need one runner.
* `on_catalina :or_older` - not an issue since we can't test those anyway.
* `on_bigsur :or_older` - will test `macos-11` and `macos-12`, as desired
* `on_bigsur :or_newer` - overtests by doing both `macos-11` and `macos-12`
* `on_monterey :or_older` - overtests by doing both `macos-11` and `macos-12`
* `on_monterey :or_newer` - will test `macos-11` and `macos-12`, as desired

So a couple of overtest edge cases, though it's better to overtest rather than undertest.

Is there an edge case I'm missing?

PR #133961, opened 2022-10-21
Add livecheck to ui.rb

nilsstreedain:
This is the first livecheck I've added to a Cask so I apologize if my style needs correction beyond what is detected by the automations. Here are the changes:
- Updated to unversioned URL for livecheck
- Set sha256 to no check for unversioned URL
- Merged version array into single version for livecheck
- Added live check
- Opened package using new version

### DIFF
--- a/Casks/ui.rb
+++ b/Casks/ui.rb
@@ -1,14 +1,15 @@
 cask "ui" do
-  version "98a0-macOS-0.56.3-79aba98f601b49818f3dae0e3d34b754"
-  sha256 :no_check
+  version "b755-macOS-0.56.3-b2d6dcc1ca1b470db93a710da52de392"
+  sha256 "d3ce491d37b6aa184247359d1b4f1a5b2e48f6054f521a53d40f4de22b513884"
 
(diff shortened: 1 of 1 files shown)


SMillerDev:
Why move to the unversioned download though?

nilsstreedain:
> Why move to the unversioned download though?

I did that because it has the version for livecheck in the location header. I'm new to this so I'm happy to receive feedback, is there a preferred implementation?

SMillerDev:
Unversioned downloads are much worse to maintain (they can change without notice) and to use (you don't know which version you're actually using). If you need a specific url in the livecheck block, specify it using `url "some-url"`

nilsstreedain:
I've switched back to a versioned URL and updated the cask to the latest version but I have a question about builds within versions. It appears UI desktop often has later builds released using the same version number, for example `UI 0.56.3 (51)` and `UI 0.56.3 (58)`. However the version numbering appears differently in the URL such as `98a0-macOS-0.56.3-79aba98f601b49818f3dae0e3d34b754` and `b755-macOS-0.56.3-b2d6dcc1ca1b470db93a710da52de392`. Should the Cask and livecheck be updated for just the version `0.56.3` or for the individual build?

p-linnane (approved):

PR #133974, opened 2022-10-21
Oversight 2.0.2  - Added Livecheck

andrewcrook:
Added LiveCheck using GitHub repo as source

- [ X] `brew audit --cask --online <cask>` is error-free.
- [X ] `brew style --fix <cask>` reports no offenses.

### DIFF
--- a/Casks/oversight.rb
+++ b/Casks/oversight.rb
@@ -7,7 +7,12 @@ cask "oversight" do
   name "OverSight"
   desc "Monitors computer mic and webcam"
   homepage "https://objective-see.com/products/oversight.html"
-
+  
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
+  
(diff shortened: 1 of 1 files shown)


p-linnane:
The [`Git`](https://rubydoc.brew.sh/Homebrew/Livecheck/Strategy/Git.html) strategy is used by `livecheck` automatically when the binary is pulled from a GitHub URL. In this case, we do not need to specify `:github_latest` as it can cause rate limiting errors, and the `Git` strategy is working.

andrewcrook:
> The [`Git`](https://rubydoc.brew.sh/Homebrew/Livecheck/Strategy/Git.html) strategy is used by `livecheck` automatically when the binary is pulled from a GitHub URL.

@p-linnane   Ah okey I didn’t see that fact it makes sense as to why it didn’t have it in the first place.
I will remember that thanks.

PR #134261, opened 2022-10-25
iTerm2: use older version on Mojave as well as High Sierra

jamiew:
The latest versions of iTerm2 no longer work on Mojave (10.14), but the older version 3.3.12 does. This patch expands the "old OS?" check to include Mojave as well

The audit failed the livecheck, but that is expected and I'm not sure how to indicate that is expected.

Checklist:

- [x] `brew audit --cask --online <cask>` is error-free (failed livecheck, but per above, that is expected)

Not adding a new cask, so skipped remainder of template

### DIFF
--- a/Casks/iterm2.rb
+++ b/Casks/iterm2.rb
@@ -1,6 +1,6 @@
 cask "iterm2" do
   # NOTE: "2" is not a version number, but an intrinsic part of the product name
-  if MacOS.version <= :high_sierra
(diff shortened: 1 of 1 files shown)


p-linnane:
The website and application itself indicate that `v3.4.17` is supported on Mojave. Please open an issue [upstream](https://gitlab.com/gnachman/iterm2/-/issues) if you are having problems on Mojave.

PR #134443, opened 2022-10-27
virtualbox: add arm64 version

yurikoles:

### DIFF
--- a/Casks/virtualbox.rb
+++ b/Casks/virtualbox.rb
@@ -1,8 +1,15 @@
 cask "virtualbox" do
-  version "7.0.2,154219"
-  sha256 "2027a2334d4ffb32820864378667fd5277be9290b863517da785036a70325768"
+  on_intel do
+    version "7.0.2,154219"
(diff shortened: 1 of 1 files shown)


p-linnane:
This should be submitted to `homebrew-cask-versions` since it is a beta.

yurikoles:
>But There Is No Stable Version!
>When software is only available as a beta, development, or unstable version, its cask can go in the main repo. When >stable versions become available, only those will be accepted as subsequent updates.

There is no stable version of VirtualBox for ARM64 macOS.

p-linnane:
Correct, but I don't think we should mix a beta arm64 version with a stable x86_64 version.

yurikoles:
> Correct, but I don't think we should mix a beta arm64 version with a stable x86_64 version.

Because why? And alternative is to create a beta ARM64-only cask, that will be eventually silently deleted and users will need to figure out themselves that it's time to switch to back to stable cask?

p-linnane:
My position is that I do not think it's wise to mix stable and beta in a cask. You can open a PR to update the one in `homebrew-cask-versions`. There is no more productive conversation to be had on this matter, so I'll be locking this.

PR #135504, opened 2022-11-10
Add restfox cask to Homebrew-Cask

Tha14:
A new cask for this project [https://github.com/flawiddsouza/Restfox](https://github.com/flawiddsouza/Restfox) with token: `restfox`
Version: 0.0.6
Currently only x64, will try to add support for M1(We need to do tests on a physical device first).
Thank you.

EDIT: Sorry for the many commits.
---

### DIFF
--- a/Casks/restfox.rb
+++ b/Casks/restfox.rb
@@ -1,6 +1,4 @@
 cask "restfox" do
-  arch intel: "osx64"
-
   version "0.0.6"
   sha256 "c855011459615b181efde6d88a82ce7c4c0a5b257a4b8a35e41d2a2de35aba6e"
 
--- a/Casks/restfox.rb
+++ b/Casks/restfox.rb
@@ -10,11 +10,6 @@ cask "restfox" do
   desc "Offline-first web HTTP client"
   homepage "https://restfox.dev/"
 
-  livecheck do
-    url :url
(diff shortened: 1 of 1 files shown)


Casks/restfox.rb
```diff
@@ -0,0 +1,23 @@
...
+  url "https://github.com/flawiddsouza/Restfox/releases/download/v#{version}/Restfox-darwin-x64-#{version}.zip",
+      verified: "github.com/flawiddsouza/Restfox/releases/download/"
+  name "Restfox"
+  desc "Offline-first web HTTP client"
+  homepage "https://restfox.dev/"
+
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
```
bevanjkay, line 16: @Tha14 
The :github_latest livecheck strategy should only be used in instances where the standard :git strategy won't work (if upstream uses pre-releases, or has tags without corresponding releases, or an older versioning system). Do you know if it is required here? If not the livecheck block can be omitted as :git will automatically be inferred from the url.

Tha14: No, it appears to not be needed in this case. I will remove that the livecheck block.

miccal:
Thank you @Tha14.

miccal (approved):

PR #135505, opened 2022-11-10
Add Icon Shelf v0.1.17

juliennicouleaud:
- [x] The submission is for a stable version

Additionally:

### DIFF
--- a/Casks/icon-shelf.rb
+++ b/Casks/icon-shelf.rb
@@ -1,5 +1,5 @@
 cask "icon-shelf" do
-  arch arm: "-arm64", intel: ""
+  arch arm: "-arm64"
 
   version "0.1.17"
   sha256 arm:   "7a5bc6f4af0658d9ae6a6449946b95aea9a5a19a5fd779973d8c3a6590347147",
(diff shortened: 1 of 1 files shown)


Casks/icon-shelf.rb
```diff
@@ -0,0 +1,28 @@
...
+  url "https://github.com/Icon-Shelf/icon-shelf/releases/download/v#{version}/Icon-Shelf-#{version}#{arch}.dmg",
+      verified: "github.com/Icon-Shelf/icon-shelf/"
+  name "Icon Shelf"
+  desc "Icon manager for web developers"
+  homepage "https://icon-shelf.github.io/"
+
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
```
bevanjkay, line 17: @juliennicouleaud
The `:github_latest` livecheck strategy should only be used in instances where the standard  `:git` strategy won't work (if upstream uses pre-releases, or has tags without corresponding releases, or an older versioning system). Do you know if it is required here? If not the livecheck block can be omitted as `:git` will automatically be inferred from the url.

juliennicouleaud: Sorry I hadn't reelized that. I've removed the block in the latest commit.

miccal (approved):

PR #135689, opened 2022-11-13
Update affinity-designer/affinity-photo/affinity-publisher to 2.0.0

unitof:
Serif has released a major [new "v2"](https://affinity.serif.com/en-us/whats-new/) for their suite of three apps:

- `affinity-designer`
- `affinity-photo`
- `affinity-publisher`

I believe these should not be new casks, as the current v1 versions [will no longer receive updates](https://affinity.serif.com/en-us/affinity-2-faq/), and v2 is backwards-compatible.

However, it **is** a paid upgrade (new license required), and the current apps **do not** auto-update or prompt user to update (at least not yet).

Could you some advice if I should remove `auto_update` then, but guessing we should leave `true`, at let the software take care of eventually alerting users to new v2 in their own auto-update or notification mechanisms.

The current v1 versions could then move to `cask-versions` as `affinity-[app]-1`, which I'm happy to do if this is accepted.

- [x] `brew audit --cask --online affinity-photo affinity-publisher affinity-designer  ` is error-free.
- [x] `brew style --fix affinity-photo affinity-publisher affinity-designer` reports no offenses.

### DIFF
--- a/Casks/affinity-designer.rb
+++ b/Casks/affinity-designer.rb
@@ -22,7 +22,7 @@ cask "affinity-designer" do
     "~/Library/Caches/com.seriflabs.affinitydesigner2",
     "~/Library/HTTPStorages/com.seriflabs.affinitydesigner2",
     "~/Library/Preferences/com.seriflabs.affinitydesigner2.plist",
-    "~/Library/WebKit/com.seriflabs.affinitydesigner2",
     "~/Library/Saved Application State/com.seriflabs.affinitydesigner2.savedState",
+    "~/Library/WebKit/com.seriflabs.affinitydesigner2",
   ]
 end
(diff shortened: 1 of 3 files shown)


p-linnane:
Given that `v1` will not receive updates anymore, I do not think they need to be added to `homebrew-cask-versions`. We only officially support the latest version, so updating these casks is the right move. I will take a look at these after CI finishes.

unitof:
Confirmed `brew upgrade affinity-publisher` **does** properly backup and trash current `/Applications/Affinity [Name].app` and installs `/Applications/Affinity [Name] 2.app`. Inherits v1 settings smoothly.

Nice! Didn't realize Homebrew intuits a greedy upgrade if cask name specified.

I had the App Store versions of Photo and Designer, so can only test installing them fresh, but that also works smoothly.

v1 & v2 Affinity apps can run simultaneously without issue.

p-linnane:
Please add `depends_on macos: ">= :catalina"` and alphabetize the `zap` stanzas. After that we're good to merge.

p-linnane (approved):
Thank you @unitof!

PR #126036, opened 2022-06-19
betterdummy: Rename cask & bump version

p-linnane:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making all changes to a cask, verify:

- [X] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [X] `brew audit --cask --online <cask>` is error-free.
- [X] `brew style --fix <cask>` reports no offenses.

PR #135895, opened 2022-11-16
paw: Rename to rapidapi, bump to 4.0.0

p-linnane:
Paw has [undergone a rebranding](https://twitter.com/luckymarmot/status/1592625527109857280) to RapidAPI. Requesting input from @Homebrew/cask due to the analytics numbers for this cask.

```
==> Analytics
install: 286 (30 days), 801 (90 days), 4,184 (365 days)
```

### DIFF
--- a/Casks/rapidapi.rb
+++ b/Casks/rapidapi.rb
@@ -1,16 +1,21 @@
-cask "paw" do
-  version "3.4.0,3004000012"
-  sha256 "d393c6fb175d8e5f671361e1e2e4578fbab0eb6a03a2226507cad953a8a819da"
+cask "rapidapi" do
+  version "4.0.0,4000000007"
+  sha256 "581f0a769247c0b7ba633fe4c5f4b6cb647014d6f400a8bcd5e9ff8ca6fbab9d"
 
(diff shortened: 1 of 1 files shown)


bevanjkay:
Until we have a way of handling aliases/renames, for popular casks what we have done is to keep the existing cask but make it `discontinued` with a `caveat` to install the new cask. And then add the new cask separately. The only to watch for is that we don't do it too often, and ideally have a mechanism to remove the old casks at a certain point.

bevanjkay (dismissed):

p-linnane:
@bevanjkay I'm going to copy the `zoomus`/`zoom` method here.

bevanjkay:
The other option is to handle it the same way we with `betterdummy/betterdisplay` which doesn't automatically install the updated version.
https://github.com/Homebrew/homebrew-cask/pull/126036

p-linnane:
Should be good now

bevanjkay:
@p-linnane Let's leave this open for a day or so to see if there's some more maintainer feedback before making a decision. I'm happy with either the Zoom method or the current method, but I would mean lightly toward the current example.

miccal (approved):

PR #136562, opened 2022-11-26
Jump desktop connect cask fix

viking1304:
Temporary fix until we are waiting for https://github.com/Homebrew/brew/pull/14123

This is a new workaround since the previous one stopped working with the new macOS.


p-linnane (approved):
Thank you @viking1304.

PR #136564, opened 2022-11-27
Update Klokki to 1.3.6

nihaals:

- [ ] `brew style --fix <cask>` reports no offenses

### DIFF
--- a/Casks/klokki.rb
+++ b/Casks/klokki.rb
@@ -1,19 +1,27 @@
 cask "klokki" do
-  version "1.3.3,62"
+  version "1.3.6,85"
   sha256 :no_check
 
-  url "https://storage.yandexcloud.net/klokki/Klokki.dmg",
-      verified: "storage.yandexcloud.net/klokki/"
+  url "https://klokki.com/download/Klokki.dmg"
   name "Klokki"
   desc "Automatic time-tracking solution or freelancers and makers"
   homepage "https://klokki.com/"
 
+  # The dates in the appcast are sometimes out of order (i.e., a newer version
+  # may have an older `pubDate`), so we have to override the default `sparkle`
+  # strategy logic to identify the latest version in this case.
   livecheck do
-    url "https://storage.yandexcloud.net/klokki/appcast.xml"
-    strategy :sparkle
+    url "https://klokki.com/download/appcast.xml"
+    strategy :sparkle do |items|
+      items.map do |item|
+        next if item.short_version.blank? || item.version.blank?
+
+        "#{item.short_version},#{item.version}"
+      end
+    end
   end
 
-  depends_on macos: ">= :mojave"
+  depends_on macos: ">= :high_sierra"
 
   app "Klokki.app"
 end


nihaals:
The appcast reports 10.13 being the minimum but the website reports 10.14.

nihaals:
Unsure why the livecheck is giving 1.3.1 since I see 1.3.6 in the appcast and the client also parses it correctly

Casks/klokki.rb
```diff
@@ -1,15 +1,14 @@
...
   sha256 :no_check
 
-  url "https://storage.yandexcloud.net/klokki/Klokki.dmg",
-      verified: "storage.yandexcloud.net/klokki/"
+  url "https://klokki.com/download/Klokki.dmg"
   name "Klokki"
   desc "Automatic time-tracking solution or freelancers and makers"
   homepage "https://klokki.com/"
 
   livecheck do
```
samford, line 10:
```suggestion
  # The dates in the appcast are sometimes out of order (i.e., a newer version
  # may have an older `pubDate`), so we have to override the default `Sparkle`
  # strategy logic to identify the latest version in this case.
  livecheck do
```

Casks/klokki.rb
```diff
@@ -1,15 +1,14 @@
...
-    url "https://storage.yandexcloud.net/klokki/appcast.xml"
+    url "https://klokki.com/download/appcast.xml"
     strategy :sparkle
```
samford, line 12:
```suggestion
    strategy :sparkle do |items|
      items.map do |item|
        next if item.short_version.blank? || item.version.blank?

        "#{item.short_version},#{item.version}"
      end
    end
```

samford (commented):
The `pubDate` values in the appcast don't appear to be in chronological order (i.e., newer versions may have an older `pubDate`) and that's why the `Sparkle` strategy reports an older version. Basically, the `Sparkle` strategy sorts the items by `pubDate` and bundle version and picks the first one.

For example, version 1.3.6 has a `pubDate` of `2022-07-10 20:35:00 +0300` but version 1.3.1 has a `pubDate` of `2022-11-24 20:05:00 +0300`, so 1.3.1 is returned as newest. The default approach is fine for most appcasts but in this scenario we have to override the default `Sparkle` behavior using a `strategy` block.

I tried to push a commit to your PR branch but received a permission denied error, so I've added suggested changes here.

p-linnane:
Thank you @samford. I didn't realize `pubDate` was the initial sort, I thought it was the highest version number. That's good to know.

@nihaals In order to pass the audit, you will need to commit the above suggestions and change the `depends_on macos:` line to `:high_sierra` to match their appcast. Just as a heads up, when opening a PR it's helpful if you allow maintainers to push to your branch so we can add these fixes in and get this merged.

nihaals:
> In order to pass the audit, you will need to commit the above suggestions and change the `depends_on macos:` line to `:high_sierra` to match their appcast.

@p-linnane I assume the appcast is always given priority since that's what `audit` checks? This isn't the only time they've conflicted (https://github.com/Homebrew/homebrew-cask/pull/136525#issuecomment-1327941418).

p-linnane:
Correct. It's up to the devloper to make sure their appcast is accurate. We can only audit against what it reports. Thanks!

p-linnane (approved):

PR #136777, opened 2022-11-30
Fix Hyperkey minimum macOS version

Aaron-Rumpler:
`brew audit` throws the error `Upstream defined 10.12 as minimal OS version and the cask defined 10.13`, however both the `Info.plist` and Sparkle appcast list 10.13 as the minimum version. I assume it's the lack of a `sparkle:shortVersionString` in the appcast?

### DIFF
--- a/Casks/hyperkey.rb
+++ b/Casks/hyperkey.rb
@@ -13,7 +13,7 @@ cask "hyperkey" do
   end
 
   auto_updates true
-  depends_on macos: ">= :sierra"
+  depends_on macos: ">= :high_sierra"
 
   app "Hyperkey.app"
 


samford:
> I assume it's the lack of a `sparkle:shortVersionString` in the appcast?

The macOS version requirement in the appcast is specified using [`sparkle:minimumSystemVersion`](https://sparkle-project.org/documentation/publishing/#minimum-system-version-requirements). The older appcast items have a `minimumSystemVersion` of 10.12 but it's 10.13 for version 0.22, so this is a bug in the audit.

Looking at the [related audit](https://github.com/Homebrew/brew/blob/aea140c0447916dd3321ef45a3ea95c3a3a93d8a/Library/Homebrew/cask/audit.rb#L548), it appears to simply take the first item from the appcast instead of using livecheck's `Sparkle` strategy to identify the newest version in the appcast. In this case, the newest version is *last* in the appcast, so the audit is incorrectly using the `minimumSystemVersion` for the first item (version 0.1) and that's why we receive the error about 10.12.

Long story short, this audit needs to be reworked to use the `Sparkle` strategy to properly identify the newest item instead of naively assuming the first item in the appcast will be the newest version. It should be an easy fix, so I'll try to get a brew PR up later when I have a moment.

miccal (approved):

PR #136962, opened 2022-12-02
megasync: Add apple silicon download

sataset:

### DIFF
--- a/Casks/megasync.rb
+++ b/Casks/megasync.rb
@@ -1,8 +1,11 @@
 cask "megasync" do
-  version "4.8.1"
-  sha256 :no_check
+  arch arm: "Arm64", intel: ""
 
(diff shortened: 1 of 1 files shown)


p-linnane:
Thank you for your contribution to `homebrew-cask` @sataset. Hash checks are only supported for URLs that contain the application version information. In this case, `:no_check` is appropriate since there is no version info in the URL.

sataset:
@p-linnane I suppose there should be no problem if checks are added with version specified in Cask code?

p-linnane:
If the URL was something like `https://mega.nz/MEGAsyncSetup#{version}-#{arch}.dmg` or something to that effect we can support hash checks. It just needs to have the version in the URL.

p-linnane (approved):

PR #137009, opened 2022-12-03
kekaexternalhelper: update livecheck

p-linnane:

---
@samford Not sure what I'm missing with this one, hoping a second set of eyes will help.

### DIFF
--- a/Casks/kekaexternalhelper.rb
+++ b/Casks/kekaexternalhelper.rb
@@ -9,9 +9,10 @@ cask "kekaexternalhelper" do
   homepage "https://github.com/aonez/Keka/wiki/Default-application"
 
   livecheck do
-    url "https://www.keka.io/en/"
-    strategy :page_match do |page|
-      match = page.match(%r{href=.*?/v?(\d+(?:\.\d+)*)/KekaExternalHelper-v?(\d+(?:\.\d+)*)\.zip}i)
+    url "https://d.keka.io/helper/"
+    regex(%r{v?(\d+(?:\.\d+)*)/KekaExternalHelper-v?(\d+(?:\.\d+)*)\.zip}i)
+    strategy :header_match do |headers, regex|
+      match = headers["location"]&.match(regex)
       next if match.blank?
 
       "#{match[2]},#{match[1]}"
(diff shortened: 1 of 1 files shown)


samford (commented):
Basically, the issue is that 1) [`#page_content`/`#page_headers` always follow redirections](https://github.com/Homebrew/brew/blob/001bacee18c512657e1535dff757c8cfdaefc116/Library/Homebrew/livecheck/strategy.rb#L51) and 2) headers from all the responses are merged together in [`HeaderMatch#find_versions`](https://github.com/Homebrew/brew/blob/001bacee18c512657e1535dff757c8cfdaefc116/Library/Homebrew/livecheck/strategy/header_match.rb#L91) before being passed to the `strategy` block, so only the last instances of a given header is present. In this case, the first response contains a `location` header that gives us the GitHub URL we need but this is lost/overwritten because the next response contains another `location` header (objects.githubusercontent.com) and it's the latter one that appears in the `headers` hash.

To address this we would need to be able to either selectively disable use of the `--location` curl arg (it's used by default) or update how `HeaderMatch` handles `strategy` block arguments to allow an alternative `all_headers` argument that provides the unmerged headers array (where we could iterate through the header hash for each response to find the URL we need).

Being able to add/remove curl args in a `livecheck` block depends on the `options` feature that I *still* need to create a PR for (it's near the top of the list but I need to get through some smaller PRs first). This will involve some discussion around the shape of the public-facing interface, so it's not going to be a quick fix. [That said, I noted this cask in my related to-do for the future.]

Modifying `HeaderMatch` to be able to selectively pass all the headers to the `strategy` block [depending on the first block parameter name] is easier and something that we can accomplish fairly quickly. I've already done something similar in `Sparkle#versions_from_content`, where the block receives a different argument based on whether the first argument to the `strategy` block is named `item` or `items`.

I'm working on something else right now but I can create a PR for the `HeaderMatch` changes when I have a moment (maybe tomorrow). Past that, I don't see any way of creating a working `livecheck` block until then, so we may just want to skip this for now. If you're fine with that, I can push a commit to add an explanatory comment before the `livecheck` block along with an appropriate `#skip` call.

p-linnane:
Sounds great! Thank you!

samford:
I pushed a commit to update the `livecheck` block to skip (along with an explanatory comment) for the time being. If you're good with the shape of this, I'll go ahead and approve it.

p-linnane:
LGTM. Thanks again!

samford (approved):

PR #137289, opened 2022-12-07
qlvideo 2.00

Marginal:
I've changed QLVideo from being packaged as an installer package to an app in order to work around various problems with registering QuickLook and Spotlight plugins on Ventura.

  `brew audit --cask --online qlvideo` fails with `Version '2.00' differs from '1.96' retrieved by livecheck`.
  `brew audit --cask --strict qlvideo` passes.

### DIFF
--- a/Casks/qlvideo.rb
+++ b/Casks/qlvideo.rb
@@ -8,13 +8,8 @@ cask "qlvideo" do
   homepage "https://github.com/Marginal/QLVideo"
 
   livecheck do
-    url "https://api.github.com/repos/Marginal/QLVideo/releases"
-    strategy :page_match do |page|
-      match = page.match(/QLVideo[._-]v?(\d+?)(\d+)\.dmg/i)
-      next if match.blank?
-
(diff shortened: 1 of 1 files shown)


bevanjkay:
@Marginal Are those files removed when the `.pkg` is uninstalled? If so, then the files will be removed for the user if they use `brew upgrade`, as brew will uninstall the package (based on the previously installed version cask), before upgrading to the new version.

Marginal:
@bevanjkay 
> Are those files removed when the .pkg is uninstalled? If so, then the files will be removed for the user if they use brew upgrade, as brew will uninstall the package (based on the previously installed version cask), before upgrading to the new version.

Yes, both `brew upgrade` and `brew reinstall` correctly remove the old files. Sorry to have wasted your time.

I think this is ready for merge.

miccal:
Thank you @Marginal.

miccal (approved):

Marginal:
@miccal Thanks

PR #137293, opened 2022-12-07
mplabx-ide: leave other Microchip packages alone.

smammy:
Other software packages from Microchip may be installed in other
subdirectories of /Applications/microchip, and other versions of MPLABX
IDE may be installed in other versioned subdirectories of
/Applications/microchip/mplabx. Prior to this commit, installing this
cask would cause the entire contents of the /Applications/microchip
directory to have its owner and group changed, and uninstalling this
cask would cause the entire /Applications/microchip directory to be
removed.

This commit restricts this cask to operating only on
/Applications/microchip/mplabx/#{version}, so that other packages are
left alone.

### DIFF
--- a/Casks/mplabx-ide.rb
+++ b/Casks/mplabx-ide.rb
@@ -41,5 +41,16 @@ cask "mplabx-ide" do
               input:      ["y", 3],
               sudo:       true,
             },
-            delete: "/Applications/microchip/mplabx/#{version}"
+            delete: [
+              "/Applications/microchip/mplabx/#{version}",
+              # The below version number needs to be updated
+              # manually each time this Cask is updated
+              "/Applications/microchip/mplabcomm/3.51.00",
+            ],
+            rmdir:  [
+              "/Applications/microchip/mplabx",
+              "/Applications/microchip/mplabcomm",
+            ]
+
+  zap trash: "/Applications/microchip"
 end
--- a/Casks/mplabx-ide.rb
+++ b/Casks/mplabx-ide.rb
@@ -32,7 +32,7 @@ cask "mplabx-ide" do
 
   postflight do
     set_ownership staged_path.to_s
-    set_ownership "/Applications/microchip/mplabx"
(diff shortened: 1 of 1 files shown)


smammy:
Based on CI result I see that the MPLABX installer also installs some software to a versioned directory under `/Applications/microchip/mplabcomm`, which the uninstaller ignores.

mplabcomm is a set of libraries that Microchip software uses to talk to JTAGs. It's possible that the user might have other Microchip software installed that depends on mplabcomm, so I don't think this cask should uninstall mplabcomm automatically.

I'm not sure that this is the right call, but I figure it's better to be conservative than to potentially break the user's system.

miccal:
@smammy are you aware of any way we can obtain the version number `3.51.00` of `mplabcomm` before installation?

smammy:
@miccal I haven't been able to find one.

Once MPLABX is installed, the mplabcomm version number can be found in a couple of places inside `/Applications/microchip/mplabx/v6.05/mplab_platform/mplablibs/modules/com-microchip-mplab-comm.jar`:

* The first line of `./com/microchip/mplab/comm/MPLABCommVersion.txt` is `MPLABComm 3.51.00 2022-09-13`.
* The method `getMPLABCommVersion` in `./com/microchip/mplab/comm/MPLABCommVersionHandler.class` returns `3.51.00`.

Would it work to have the cask to read `MPLABCommVersion.txt` in postflight to decide which version of mplabcomm to remove on uninstall?

Another option would be to remove `/Applications/microchip/mplabcomm` only when `/Applications/microchip` is otherwise empty.

I think it would be ideal if the Microchip installer could be made to install both mplabx and mplabcomm to `staged_path`, but that doesn't seem possible.

A more manual approach would be to hardcode the version and bump it manually every time mplabx is updated.

Let me know what you think.

miccal:
Thank you for the detailed explanation @smammy.

p-linnane (approved):

PR #137257, opened 2022-12-07
Update launchcontrol from 2.0.3,2304 to 2.0.3,2310

solzard:
Created with `brew bump-cask-pr`.

PR #137301, opened 2022-12-08
launchcontrol: fix depends_on for legacy version

p-linnane:
I broke the legacy version in #137257. This method allows us to set `depends_on macos` based off the user's OS.

### DIFF
--- a/Casks/launchcontrol.rb
+++ b/Casks/launchcontrol.rb
@@ -2,9 +2,22 @@ cask "launchcontrol" do
   if MacOS.version <= :catalina
     version "1.52.7,1959"
     sha256 "16c3d89e41a99cbf43e6996681358e8e7a4bc63fa770b9f8c0bc72c5356a0b8a"
+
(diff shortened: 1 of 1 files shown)


miccal (approved):

Casks/launchcontrol.rb
```diff
@@ -2,23 +2,30 @@
   if MacOS.version <= :catalina
```
MikeMcQuaid, line 2: @p-linnane @miccal Can we just remove this block?
- it broke formulae.brew.sh generation again: https://github.com/Homebrew/formulae.brew.sh/actions/runs/3647250876/jobs/6160003257
- it's not actually tested with any CI in any way
- we don't support Catalina any more anyway
- it's not using the new @Rylan12 DSLs (e.g. https://github.com/Homebrew/homebrew-cask/pull/130392, `on_catalina`) but instead the old `if`-based approach that keeps breaking

Thanks ❤️

PR #137491, opened 2022-12-11
Ved: Remove livecheck, Cloudflare is preventing it from working

Adrthegamedev:

Livecheck gives a 403, likely because the cloudflare botcheck is blocking it. This PR fixes that.


Adrthegamedev:
livecheck still returns something, even after the livecheck is gone??? weird. i'll try to sort something out with the devs of this app though

Issue #137512, opened 2022-12-12
Convert `if MacOS.version` blocks to `on_system` blocks


p-linnane:
### Description of issue

This is a tracking issue to monitor the progress of converting all `if MacOS.version` blocks into the new `on_system` blocks. We have better coverage with `brew readall` using these new blocks.

- [x] amethyst #137487
- [x] apparency #137516
- [x] appcleaner #137517
- [x] appgate-sdp-client #137518
- [x] araxis-merge #137536 
- [x] artisan #137537
- [x] bartender #137538
- [x] bbedit #137539 
- [x] birdfont #137625
- [x] boxcryptor #137626
- [x] calibre #137627
- [x] cave-story #137628
- [x] cd-to #137629
- [x] choosy #137630
- [x] clean-me #137631
- [x] cockatrice #137632
- [x] cocktail #137661 
- [x] contexts #137662 
- [x] controllermate #137803
- [x] copyq #137804
- [x] coteditor #137805
- [x] couchbase-server-enterprise #137807 
- [x] dash #137808 
- [x] datweatherdoe #137809 
- [x] day-o #137810 
- [x] deeper #137811 
- [x] diskmaker-x #137812 
- [x] drawpile #137813 
- [x] duet #137515
- [x] evernote #138309
- [x] f-bar #138310
- [x] far2l #138312
- [x] flrig #138311
- [x] fluor #138313 
- [x] freac #138314 
- [x] free-gpgmail #138315 
- [x] geekbench https://github.com/Homebrew/homebrew-cask/pull/138578
- [x] gobdokumente https://github.com/Homebrew/homebrew-cask/pull/138579
- [x] hammerspoon https://github.com/Homebrew/homebrew-cask/pull/138580
- [x] horos https://github.com/Homebrew/homebrew-cask/pull/138581
- [x] idefrag https://github.com/Homebrew/homebrew-cask/pull/138582
- [x] inkstitch https://github.com/Homebrew/homebrew-cask/pull/138583
- [x] ipartition https://github.com/Homebrew/homebrew-cask/pull/138584
- [x] ipsecuritas https://github.com/Homebrew/homebrew-cask/pull/138585
- [x] isubtitle https://github.com/Homebrew/homebrew-cask/pull/138586
- [x] isyncr https://github.com/Homebrew/homebrew-cask/pull/138587
- [x] iterm2 #138307 
- [x] itsycal #138034
[...]

PR #137518, opened 2022-12-12
appgate-sdp-client: convert to `on_system` blocks

p-linnane:
https://github.com/Homebrew/homebrew-cask/issues/137512


bevanjkay:
One option would be to have the oldest version include `:or_older`, so that the cask can be read, but keep the depends_on so that it will show as unsupported on attempted install.

p-linnane:
I think that is the preferred approach for now, as discussed in Slack. I am going to open an issue to track this behavior per request.

miccal (approved):

PR #137611, opened 2022-12-13
suspicious-package: convert to `on_system` blocks

miccal:


bevanjkay:
@miccal @p-linnane In these instances I think we have two options with livecheck. We could keep the livecheck block as a separate instance in each `on_{system}` block, or we could move it to where livecheck usually is and just split it in half. It adds two additional `on_{system}` blocks, but it reduces the repeated `livecheck skip` blocks.

p-linnane:
@bevanjkay @miccal I am a proponent of keeping everything arch/OS specific in the top `on_system` blocks, and all shared lines below those blocks. I think it's cleaner because it lets you see what is arch/OS specific in one spot. I'm open to the alternative, my main concern is that we make sure we're consistent with whatever we choose.

One thing that helped me make up my mind was a comment from @Rylan12 earlier:
>The point of `on_system` is simply to apply certain stanzas on certain os versions, not to determine which versions are supported.

p-linnane (approved):

PR #137662, opened 2022-12-14
contexts: convert to `on_system` blocks

bevanjkay:
#137512

### DIFF
--- a/Casks/contexts.rb
+++ b/Casks/contexts.rb
@@ -1,25 +1,30 @@
 cask "contexts" do
-  if MacOS.version <= :mojave
+  on_mojave :or_older do
     version "3.7.1"
     sha256 "de5e4a660cc30276155606b539d1ae58684115a3983d69598f1505fcad499a87"
-  else
+
+    livecheck do
+      skip "Legacy version"
+    end
+  end
+  on_catalina :or_newer do
     version "3.9.0"
     sha256 "c86a214d5ee7718453d0fb6442b3ef9a7fc02c5c1582b6d28ef0c2e621a6b6aa"
+
+    livecheck do
+      url "https://contexts.co/appcasts/stable.xml"
+      strategy :sparkle do |item|
+        item.version.chars.join(".")
+      end
+    end
+
+    depends_on macos: ">= :catalina"
   end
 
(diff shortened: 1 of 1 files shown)


Casks/contexts.rb
```diff
@@ -1,26 +1,31 @@
...
     sha256 "c86a214d5ee7718453d0fb6442b3ef9a7fc02c5c1582b6d28ef0c2e621a6b6aa"
+
+    livecheck do
+      url "https://contexts.co/appcasts/stable.xml"
+      strategy :sparkle do |item|
+        item.version.chars.join(".")
+      end
+    end
+
+    depends_on macos: ">= :catalina"
```
bevanjkay, line 21: @p-linnane @miccal The `depends_on` statements on some of these are going to be unnecessary, except for passing the audit that compares the `depends_on` with the required version from sparkle.

p-linnane: I think we are using the combination of `:or_older` on the first block and the `depends_on` to prevent older OS's that don't have a version from throwing errors and breaking regeneration.

p-linnane (approved):

Casks/contexts.rb
bevanjkay: For this specific cask, because the livecheck uses `:sparkle` the `on_{system}` block has to have a `depends_on macos:` that matches the minimum version in sparkle.

etiennepellegrini:
Downgrading to match the livecheck broke Contexts for me - it flashes for a second and exits when started.
```console
$ brew uninstall contexts
$ cd $(dirname $(brew edit --print-path contexts))
$ git checkout a367607a6e -- contexts.rb
$ brew install contexts
```
fixed it.

(FYI:
```
OS: macOS 12.6.1 21G217 x86_64
Host: MacBookPro16,1
Kernel: 21.6.0
Shell: bash 5.2.15
CPU: Intel i9-9980HK (16) @ 2.40GHz
GPU: Intel UHD Graphics 630, AMD Radeon Pro 5500M
Memory: 24336MiB / 65536MiB
```
)

bevanjkay:
@etiennepellegrini It looks like upstream is providing a manual link to `3.9.0`, but their Sparkle updating mechanism may be broken at the moment.
https://contexts.co/whats-new/

PR #137805, opened 2022-12-15
coteditor: convert to `on_system` blocks

p-linnane:
https://github.com/Homebrew/homebrew-cask/issues/137512

### DIFF
--- a/Casks/coteditor.rb
+++ b/Casks/coteditor.rb
@@ -1,16 +1,52 @@
 cask "coteditor" do
-  if MacOS.version <= :el_capitan
+  on_el_capitan :or_older do
     version "3.5.4"
     sha256 "0b2cbf38cc531268e3691f307445e05ae5da64b48ceaf86c4d16b993c9be3e9f"
-  elsif MacOS.version <= :mojave
+
+    livecheck do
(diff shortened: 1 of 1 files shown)


p-linnane:
@Rylan12 @bevanjkay This feels kind of weird. There is a version for El Capitan, and a version for Sierra through Mojave. I thought `on_mojave :or_older` would account for 10.12 through 10.14 by reading top down, but that's not the case. I had to put in conditionals for Sierra, High Sierra, and Mojave seperately.

bevanjkay:
@p-linnane I had made a similar assumption. It may be too much of an edge case, but it could be useful to target multiple systems at once `on_sierra, on_high_sierra, on_mojave do`

Rylan12:
If we find enough[^1] cases where such a thing would be useful, it's probably worth implementing. I don't think it would be very difficult, we just have to decide on the best DSL to use that's consistent, clear, and easy to use

[^1]: Not sure what the threshold is, though, and I'm not sure we're there yet

bevanjkay (approved):

p-linnane:
Agreed that we are not there yet. Will continue to monitor as we progress through the rest of the casks.

PR #138032, opened 2022-12-18
proxifier: convert to `on_system` blocks

EricFromCanada:
#137512

### DIFF
--- a/Casks/proxifier.rb
+++ b/Casks/proxifier.rb
@@ -1,16 +1,19 @@
 cask "proxifier" do
-  if MacOS.version <= :catalina
+  on_catalina :or_older do
     version "2.26"
-    sha256 :no_check
-    url "https://www.proxifier.com/download/ProxifierMac.dmg"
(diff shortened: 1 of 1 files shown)


Casks/proxifier.rb
```diff
@@ -1,17 +1,20 @@
...
     end
-  else
+  end
+  on_big_sur :or_newer do
     version "3.7"
-    sha256 "ed96cb643aa52292410109850997165d4c80f2782d596e018f53b33e45415ed3"
-    url "https://www.proxifier.com/download/ProxifierMac#{version.major}.dmg"
+    sha256 :no_check
+
+    url "https://www.proxifier.com/download/ProxifierMac.dmg"
```
bevanjkay, line 17: @EricFromCanada Does the versioned link no longer exist?

EricFromCanada: They've switched it so that versioned links are legacy and the unversioned link is current (see https://www.proxifier.com/download/).

bevanjkay (approved):

Casks/proxifier.rb
bevanjkay: Thanks for checking @EricFromCanada 👍🏻

PR #138213, opened 2022-12-20
add sha256 checksum for sync.com cask

tillydray:
**Why:**
Sync.com is an end-to-end encrypted cloud storage solution. As such, users are more likely to want a checksum included to ensure that the app they downloaded is the app sync.com made

### DIFF
--- a/Casks/sync.rb
+++ b/Casks/sync.rb
@@ -1,6 +1,7 @@
 cask "sync" do
   version "2.1.4"
-  sha256 :no_check
+  # https://sync.com/download/sync-packages/osx/VERSION_NUMBER/cur/Sync.dmg.sha256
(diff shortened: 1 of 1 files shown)


p-linnane:
We can only support checksums on URLs that contain the version number. Since this application is updated in place, it would cause checksum errors on the next update and prevent installation. If a versioned URL is available, we can add that and the checksum.

tillydray:
Oh I didn’t realize that, sorry. It’s annoying to have to turn off the sha check to update this, or install it outside of brew. Was hoping I could fix it, and I’m willing to maintain it but I realize that’s going to be annoying for everyone. Oh well :)

p-linnane:
You can try to ask upstream to provide versioned links.

PR #138217, opened 2022-12-20
Remove `aceinstaller` from Loopback

nihaals:
This removes `aceinstaller` from `loopback`'s `installer` and `uninstall` blocks. This is because:
- None of the other Rogue Amoeba casks that use ACE run it
- This only needs to be ran once for a new ACE version, not for each cask
- ACE should not be uninstalled when uninstalling a single cask as many other applications require ACE:
  - Airfoil
  - Audio Hijack
  - Piezo
  - SoundSource
  - Some [third-party](https://rogueamoeba.com/licensing/ace/#current-licensees) applications

### DIFF
--- a/Casks/loopback.rb
+++ b/Casks/loopback.rb
@@ -16,17 +16,6 @@ cask "loopback" do
   depends_on macos: ">= :catalina"
 
   app "Loopback.app"
-  installer script: {
-    executable: "Loopback.app/Contents/Resources/aceinstaller",
-    args:       ["install", "--silent"],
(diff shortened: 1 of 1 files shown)


bevanjkay (approved):

bevanjkay:
Thanks @nihaals - I had originally added those lines as it was breaking some of my automated installs with the driver getting out of sync with the app.
But I had not realised it's effect on other software. I'll move the changes to my own tap 👍🏻

PR #138218, opened 2022-12-20
updated logos-cask

dirksierd:

### DIFF
--- a/Casks/logos.rb
+++ b/Casks/logos.rb
@@ -1,8 +1,11 @@
 cask "logos" do
-  version "9.17.0.0020"
-  sha256 "e0e904043eab8eb8624e19261617daa197d50b561dd22afcc3e638a1720ae888"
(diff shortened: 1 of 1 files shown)


Casks/logos.rb
```diff
@@ -23,7 +26,5 @@
 
   zap trash: [
     "~/Library/Preferences/com.logos.Logos.plist",
-    "~/Library/Preferences/com.logos.LogosCEF.plist",
-    "~/Library/Preferences/com.logos.LogosIndexer.plist",
```
bevanjkay, line 27: @dirksierd I would prefer to keep the zap lines, as it doesn't hurt, and if the program leaves behind files from previous versions they are cleaned up.

dirksierd: Allright, I restored them 👌

p-linnane (approved):
Thank you @dirksierd.

PR #138308, opened 2022-12-22
touch-bar-simulator: convert to on_system blocks

viking1304:

### DIFF
--- a/Casks/touch-bar-simulator.rb
+++ b/Casks/touch-bar-simulator.rb
@@ -3,6 +3,8 @@ cask "touch-bar-simulator" do
     version "1.2.0"
     sha256 "4abe55de716ae56a41031cdb1d3b27bf6b1efae18b33b80bb0419669a9a76aa1"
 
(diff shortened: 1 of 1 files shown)


bevanjkay:
@viking1304 Thanks! This one needs to keep the `url` in the different blocks because the formatting is different on the pre-mojave version.

viking1304:
> @viking1304 Thanks! This one needs to keep the `url` in the different blocks because the formatting is different on the pre-mojave version.

Oooops! Now I see the difference in the URL. Thank you very much for pointing that out.

bevanjkay:
@viking1304 We've been adding the `livecheck skip` blocks where necessary, to make sure `brew bump` and `brew livecheck` wouldn't return incorrect versions when on older hardware. Sometimes an actual livecheck block can be used, but not often.

viking1304:
> @viking1304 We've been adding the `livecheck skip` blocks where necessary, to make sure `brew bump` and `brew livecheck` wouldn't return incorrect versions when on older hardware. Sometimes an actual livecheck block can be used, but not often.

Make sense. Thanks for the explanation.

bevanjkay (approved):

PR #138470, opened 2022-12-24
threads: remove

p-linnane:
Binary downloads and `livecheck` are both returning 503 now. Everything appears to be walled now.


nhooey:
> @p-linnane: Binary downloads and livecheck are both returning 503 now. Everything appears to be walled now.

It looks like that's no longer true, the same version: 2.2.4 is now available for download:
```bash
$ VERSION=2.2.4 && curl -L -I \
    "https://starupdate.threads.com/download/version/${VERSION}/Threads-darwin-${VERSION}.zip"
HTTP/2 200
```

PR #138579, opened 2022-12-27
gobdokumente: convert to `on_system` blocks

Rylan12:
Convert to `on_system` blocks

See #137512

### DIFF
--- a/Casks/gobdokumente.rb
+++ b/Casks/gobdokumente.rb
@@ -1,31 +1,20 @@
 cask "gobdokumente" do
-  if MacOS.version <= :catalina
-    version "1.6.8"
-    sha256 :no_check
-    url "https://download.moapp.software/GoBDokumente.zip",
-        verified: "download.moapp.software/"
+  version "2.1"
+  sha256 :no_check
 
-    livecheck do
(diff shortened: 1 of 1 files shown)


Rylan12:
Remind me what we decided to do about `Error: Upstream defined 10.16 as minimal OS version and the cask defined 10.11`?

p-linnane:
For the 10.16 thing specifically we are setting it to 11 on the cask and sending a message to upstream asking them to change to 11. The audit fails, but we merge anyway.

Rylan12:
Okay, thanks!

Think we should just add a case to the audit to treat 10.16 as the same as 11?

Also, because of this, it's safe to remove the legacy version info, right? Since no one on those OS versions could install anyway because of the `depends_on`

p-linnane:
There is currently no logic in the audit for 10.16. I would implement it, but I'm not sure how 😄

Rylan12:
> There is currently no logic in the audit for 10.16. I would implement it, but I'm not sure how 😄

https://github.com/Homebrew/brew/pull/14297

bevanjkay (approved):

PR #138606, opened 2022-12-28
microsoft-word: convert to `on_system` blocks

Rylan12:
Convert to `on_system` blocks

See https://github.com/Homebrew/homebrew-cask/issues/137512

### DIFF
--- a/Casks/microsoft-word.rb
+++ b/Casks/microsoft-word.rb
@@ -1,17 +1,21 @@
 cask "microsoft-word" do
-  if MacOS.version <= :el_capitan
+  on_el_capitan :or_older do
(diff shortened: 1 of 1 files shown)


Casks/microsoft-word.rb
```diff
@@ -1,17 +1,21 @@
-cask "microsoft-word" do
-  if MacOS.version <= :el_capitan
+eask "microsoft-word" do
```
Rylan12, line 1:
```suggestion
cask "microsoft-word" do
```

Not sure what happened here 😅

bevanjkay (approved):

PR #138802, opened 2022-12-31
workflows: update autopublish to work on all homebrew-cask-* repositories

bevanjkay:
This is an attempt to setup the autopublish workflow to work across all of our `homebrew-cask-*` repositories.

CC: @p-linnane


p-linnane (approved):

bevanjkay:
@p-linnane We will see if it still works soon 👍🏻

PR #138875, opened 2023-01-01
calhash: restore

p-linnane:
Titanium Software has brought this application back. Updated for `on_system` blocks.

### DIFF
--- a/Casks/calhash.rb
+++ b/Casks/calhash.rb
@@ -1,0 +1,45 @@
+cask "calhash" do
+  on_catalina do
+    version "1.0.5"
+    sha256 "bb15c18839837bdeda203a8a916b09b4f2b971e5bba10b603f290130939d3bf7"
+
+    depends_on macos: :catalina
+  end
+  on_big_sur do
+    version "1.1.1"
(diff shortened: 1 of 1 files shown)


Casks/calhash.rb
```diff
@@ -0,0 +1,45 @@
...
+    depends_on macos: :monterey
+  end
+  on_ventura do
+    version "1.2.0"
+    sha256 "b09df7b7c93b0bd0476a0e5c2dd665f94b0eeec5dbfe64b593765a5de002c537"
+
+    depends_on macos: :ventura
+  end
+
+  url "https://www.titanium-software.fr/download/#{MacOS.version.to_s.delete(".")}/CalHash.dmg"
```
bevanjkay, line 27: Is this syntax compatible with the API? @Rylan12 

We can merge for now and come back to it if we need to

bevanjkay (approved):

Casks/calhash.rb
p-linnane: This is the same syntax that's in all the other Titanium Software casks.

PR #139287, opened 2023-01-09
Create peazip.rb

3xp0rt:

### DIFF
--- a/Casks/peazip.rb
+++ b/Casks/peazip.rb
@@ -1,0 +1,20 @@
+cask "peazip" do
+  arch arm: "aarch64", intel: "x86_64"
+
+  version "9.0.0"
+  sha256 arm:   "556db5cc95a229dbb16c9c29085e1978fbae2138c0ddb7f935a572047bd855e8",
+         intel: "65f184f2a79be59096231b9f18daf4a22018c4f47f99c4b503278b86bb60af9a"
+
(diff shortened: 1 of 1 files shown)


miccal:
Thank you for your submission @3xp0rt, but we cannot accept it at this time because it does not pass our signature verification check that are a requirement for new submissions:

mariobrostech:
@miccal Does this policy also apply to Intel versions of apps, since they can be ran on Intel Macs without fully disabling Gatekeeper? If so, can this pull request be modified to only contain the Intel build of PeaZip, allowing users of Intel Macs still obtain PeaZip via Homebrew Cask while we wait for the developer to sign future builds for Apple Silicon users?

miccal:
> @miccal Does this policy also apply to Intel versions of apps, since they can be ran on Intel Macs without fully disabling Gatekeeper? If so, can this pull request be modified to only contain the Intel build of PeaZip, allowing users of Intel Macs still obtain PeaZip via Homebrew Cask while we wait for the developer to sign future builds for Apple Silicon users?

Yes, this policy applies to all new submission to Cask.

mariobrostech:
Ok, thanks for the clarification! I feel like this should be emphasized a little more in the FAQ since it seemed like this policy only applies to Apple Silicon apps from what's stated there (at least that was my interpretation when I read it initially).

PR #139334, opened 2023-01-10
Reflects changes in download files in lyx-2.3.7

kojiOnHill:
Current url points to a binary for older versions of MacOS, which is a limited version such that SyncTeX does not work. Modified so that newer MacOS points to a full version (https://www.lyx.org/Download#toc4).

MacOS does not include python since Monterey on which LyX depends. Adding dependencies for users' easy access.
See https://wiki.lyx.org/Mac/Mac#toc2

### DIFF
--- a/Casks/lyx.rb
+++ b/Casks/lyx.rb
@@ -1,11 +1,12 @@
 cask "lyx" do
   version "2.3.7"
 
-  if MacOS.version <= :monterey
+  on_monterey :or_older do
     sha256 "aaaaa005c5ec4bf574534de31cbd93ab4908dfa655da7535044185874a285c52"
     url "https://ftp.lip6.fr/pub/lyx/bin/#{version.major_minor_patch}/LyX-#{version}+qt5-x86_64-cocoa.dmg",
         verified: "ftp.lip6.fr/pub/lyx/"
-  else
+  end
+  on_ventura :or_newer do
     sha256 "4a0e5d9ad2d08f2b379892816934b64b99d815eaeede14157c3219f80fe039d2"
     url "https://ftp.lip6.fr/pub/lyx/bin/#{version.major_minor_patch}/LyX-#{version}+qt5-x86_64-arm64-cocoa.dmg",
         verified: "ftp.lip6.fr/pub/lyx/"
@@ -20,10 +21,10 @@ cask "lyx" do
     regex(/LyX-(\d+(?:\.\d+)*)\+qt5/i)
   end
 
-  if MacOS.version >= :monterey
+  on_monterey :or_newer do
     depends_on formula: "python"
   end
-  
+
   app "LyX.app"
   binary "#{appdir}/LyX.app/Contents/MacOS/inkscape", target: "lyx-inkscape"
   binary "#{appdir}/LyX.app/Contents/MacOS/lyx"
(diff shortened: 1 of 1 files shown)


kojiOnHill (commented):
Download failed simply because the server is slow

kojiOnHill:
Check failed because of the server speed. Retry.

miccal:
According to https://wiki.lyx.org/Mac/Mac#toc2:

> There are three options to install Python:
Apples Python 3 - install the command line tools of Xcode
Original Python 3 - download and install the software package from python.org
Open source package manager - use Macports from macports.org or Homebrew from brew.sh

Since the CLT are a requirement for `brew`, i would suggest that the `depends_on formula: "python"` is not required?

kojiOnHill:
> Since the CLT are a requirement for `brew`, i would suggest that the `depends_on formula: "python"` is not required?

Yes, I agree. Having CLT solves it.

miccal:
Thanks @kojiOnHill.

Next question -- the file `LyX-2.3.7+qt5-x86_64-arm64-cocoa.dmg` appears to be a universal binary and I believe from the LyX website will work on all versions of macOS from 10.4, so is there a reason to bother with continuing to offer the `LyX-2.3.7+qt5-x86_64-cocoa.dmg` version?

kojiOnHill:
> Next question -- the file `LyX-2.3.7+qt5-x86_64-arm64-cocoa.dmg` appears to be a universal binary and I believe from the LyX website will work on all versions of macOS from 10.4, so is there a reason to bother with continuing to offer the `LyX-2.3.7+qt5-x86_64-cocoa.dmg` version?

I haven't tested the former on older OS's so I'm not sure. Since the webpage says SyncTeX doesn't work with the latter file, I suspect that the former is more than a simple bundle. We need more information before removing it. I think I can ask someone.

miccal:
> > Next question -- the file `LyX-2.3.7+qt5-x86_64-arm64-cocoa.dmg` appears to be a universal binary and I believe from the LyX website will work on all versions of macOS from 10.4, so is there a reason to bother with continuing to offer the `LyX-2.3.7+qt5-x86_64-cocoa.dmg` version?
> 
> I haven't tested the former on older OS's so I'm not sure. Since the webpage says SyncTeX doesn't work with the latter file, I suspect that the former is more than a simple bundle. We need more information before removing it. I think I can ask someone.

@kojiOnHill that would be very helpful indeed if you are able to ask upstream about this, thank you.

kojiOnHill:
@miccal One of the LyX developers says:

- LyX-2.3.7+qt5-x86_64-arm64-cocoa.dmg
* build on BigSur for Intel and Apple chips (minimum macOS 10.14 with Qt 5.15.7).

- LyX-2.3.7+qt5-x86_64-cocoa.dmg
* build on Mojave for Intel chips (minimum macOS 10.12 with Qt 5.12.9).

So the former is for MacOS later than 10.14. 

Another point is that the latter also requires MacOS 10.12, which means that users of older OS than that need to compile from the source. In that case, things should move from cask to homebrew. Should we take care of this? Is it better have a non-cask version too?

miccal:
Thank you for chasing this up @kojiOnHill.

miccal (approved):

PR #139730, opened 2023-01-18
Install mysqlworkbench 8.0.32 just on intel due to critical upstream bug

rbubley:
https://bugs.mysql.com/bug.php?id=109671

- [ X] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [ X] `brew audit --cask --online <cask>` is error-free.
- [ ~] `brew style --fix <cask>` reports no offenses.
No offenses other than ones in the cask before this PR.

### DIFF
--- a/Casks/mysqlworkbench.rb
+++ b/Casks/mysqlworkbench.rb
@@ -1,3 +1,6 @@
+# typed: false
+# frozen_string_literal: true
+
 cask "mysqlworkbench" do
   on_sierra :or_older do
     version "6.3.10"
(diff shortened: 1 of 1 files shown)


bevanjkay:
@rbubley Thanks for putting this together. We generally serve the version as provided upstream and the download page is still serving `8.0.32` for both Intel and ARM without any warning of a bug, so in this case we should just wait until `8.0.33`

PR #139807, opened 2023-01-20
Add `obs-advanced-scene-switcher` cask

sambostock:
This adds a cask for https://obsproject.com/forum/resources/advanced-scene-switcher.395

### DIFF
--- a/Casks/obs-advanced-scene-switcher.rb
+++ b/Casks/obs-advanced-scene-switcher.rb
@@ -1,0 +1,43 @@
+cask "obs-advanced-scene-switcher" do
+  arch arm: "arm64", intel: "universal"
+
+  version "1.20.1"
+  sha256 arm:   "54f520a84a885b4611ad2f2f336fedb30880d7d1719c731b2cd9498842094d25",
+         intel: "dd5022325b7e16ed9ec954c1be67e1c101eb604dce5255da47f7605514b9b904"
+
+  url "https://github.com/WarmUpTill/SceneSwitcher/releases/download/#{version}/advanced-scene-switcher-macos-#{arch}.pkg",
+      verified: "https://github.com/WarmUpTill/SceneSwitcher"
+  name "OBS Advanced Scene Switcher"
+  desc "Automated scene switcher for OBS Studio"
+  homepage "https://obsproject.com/forum/resources/advanced-scene-switcher.395"
+
+  depends_on cask: "obs"
+
+  # FIXME: pkg puts this plugin in wrong directory...
+  #        Homebrew moves OBS into
+  #          /Applications/OBS.app
+  #        However, it continues to use
+  #          ~/Library/Application\ Support/obs-studio
+  #        Therefore, the plugin should be put in
+  #          ~/Library/Application\ Support/obs-studio/plugins/advanced-scene-switcher.plugin
+  #        but it gets put in the root library instead
+  #          /Library/Application\ Support/obs-studio/plugins/advanced-scene-switcher.plugin
+  #        How to solve this?
+  #          Should Homebrew be passing something to the installer? "install for me, not all users"?
+  #          Is it the package's fault for assuming the app's settings are in the root?
+  #          Is it the OBS cask's fault for leaving the settings in ~/Library instead of moving them?
+  #          Didn't find any "choices" to set...
+  pkg "advanced-scene-switcher-macos-#{arch}.pkg"
+  # NOTE: Attempting to fix this with artifact stanza does not work
+  #       ==> Moving Generic Artifact 'advanced-scene-switcher.plugin' to
+  #           '/Users/sambostock/Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin'
+  #       ==> Purging files for version 1.20.1 of Cask obs-advanced-scene-switcher
+  #       Error: Permission denied @ rb_file_s_rename - (
+  #         /Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin,
+  #         /Users/sambostock/Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin,
+  #       )
+  plugin_path = "/Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin"
+  artifact plugin_path, target: Dir.home + plugin_path
+
+  uninstall pkgutil: "com.warmuptill.advanced-scene-switcher"
+end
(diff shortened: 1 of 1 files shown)


Casks/obs-advanced-scene-switcher.rb
```diff
@@ -0,0 +1,43 @@
...
+  #           '/Users/sambostock/Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin'
+  #       ==> Purging files for version 1.20.1 of Cask obs-advanced-scene-switcher
+  #       Error: Permission denied @ rb_file_s_rename - (
+  #         /Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin,
+  #         /Users/sambostock/Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin,
+  #       )
+  plugin_path = "/Library/Application Support/obs-studio/plugins/advanced-scene-switcher.plugin"
+  artifact plugin_path, target: Dir.home + plugin_path
+
+  uninstall pkgutil: "com.warmuptill.advanced-scene-switcher"
```
sambostock, line 42: I'm having trouble figuring out a resolution to this. As far as I can tell, the `obs` cask moves `OBS.app` into `/Application`, but it continues to use `~/Library/Application\ Support/obs-studio` for its config files, notably the `plugins` subdirectory.

Unfortunately, the `.pkg` installer installs the plugin under the wrong `Library` directory

```diff
-/Library/Application\ Support/obs-studio/plugins/advanced-scene-switcher.plugin
+~/Library/Application\ Support/obs-studio/plugins/advanced-scene-switcher.plugin
```

This results in the plugin not being found when the OBS app is restarted.

If I manually run the `.pkg` installer, I have the option to change the install location:

Change install location
Install for me only

This results in the `.plugin` being put in the correct location (`~/Library/Application\ Support/obs-studio/plugins/advanced-scene-switcher.plugin`), and in `OBS.app` finding it upon reopening.

Is there a way to give the `.pkg` installer this information? I tried my hand at investigating "choices" as per the docs, but I didn't find anything.

sambostock: Based on `installer -help`, I think what we want is to be able to specify the `target` as in the example

```sh
installer -pkg InstallMe.pkg -target CurrentUserHomeDirectory
```

I noticed in the debug logs that Homebrew specifies `-target /`

```
brew reinstall --cask obs-advanced-scene-switcher --debug
...
==> Installing artifact of class Cask::Artifact::Pkg
==> Running installer for obs-advanced-scene-switcher; your password may be necessary.
Package installers may write to any location; options such as `--appdir` are ignored.
/usr/bin/sudo -E -- /usr/bin/env LOGNAME=me USER=me USERNAME=me /usr/sbin/installer -pkg /opt/homebrew/Caskroom/obs-advanced-scene-switcher/1.20.1/advanced-scene-switcher-macos-arm64.pkg -target /
...
```

sambostock: Looks like that's controlled by this line: https://github.com/Homebrew/brew/blob/31ed6906f927ec7c63b72c6f5c7d72ea6baa2fb4/Library/Homebrew/cask/artifact/pkg.rb#L57

sambostock: I've opened https://github.com/Homebrew/brew/pull/14393 to add this, but it will definitely require review from someone with more context about MacOS installers 😅

sambostock: I've confirmed that updating the formula to

```ruby
pkg "advanced-scene-switcher-macos-#{arch}.pkg", target: "CurrentUserHomeDirectory"
```

works (on my branch of Homebrew), and the plugin is detected by OBS. I can update the branch if someone confirms this is the right direction.

sambostock: I've also investigated adding the StreamFX OBS plugin, but it looks like its `.pkg` does not support configuring the install location, so perhaps the approach should be to allow `artifact` to move a file from `/Library/...` to another directory? Currently, that fails because `sudo` is required.

Casks/obs-advanced-scene-switcher.rb
```diff
@@ -0,0 +1,43 @@
...
+  #        Homebrew moves OBS into
+  #          /Applications/OBS.app
+  #        However, it continues to use
+  #          ~/Library/Application\ Support/obs-studio
+  #        Therefore, the plugin should be put in
+  #          ~/Library/Application\ Support/obs-studio/plugins/advanced-scene-switcher.plugin
+  #        but it gets put in the root library instead
+  #          /Library/Application\ Support/obs-studio/plugins/advanced-scene-switcher.plugin
+  #        How to solve this?
+  #          Should Homebrew be passing something to the installer? "install for me, not all users"?
```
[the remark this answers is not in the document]
sambostock: In case it's useful, I found https://obsproject.com/kb/plugins-guide#install-or-remove-plugins which discusses the OBS plugin folder:

> Install or remove plugins
Many of OBS Studio's most popular plugins come with installers. In general, they install plugins to the locations listed below.
> 
> If you need to install a plugin manually, the relevant files likely need to go in one these folders. Please check the plugin's documentation first.
> 
> ...
> <table><tr><th>macOS</th><td><code>~/Library/Application Support/obs-studio/plugins</code></td></tr></table>
> ...
> 
> If you install OBS Studio to a custom location or use portable mode, you will need to change the installer's target folder to your custom location's data/plugins folder.

Casks/obs-advanced-scene-switcher.rb
carlocab: Specifying `-target` doesn't seem reliable because it's up to the upstream `.pkg` to decide whether to support it (and how to treat it in cases where it is supported).

Maybe `advanced-scene-switcher` should use `~/Library` by default, since that's what `OBS.app` uses?

Casks/obs-advanced-scene-switcher.rb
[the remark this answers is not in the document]
carlocab: > Homebrew Cask generally does not install stuff into `~/Library` when it can use `/Library` instead because the installing and running users are not necessarily the same.

This seems like a good reason to use `/Library` where possible, but it's not clear that that's the case here. Maybe we need an answer to @reitermarkus's [question](https://github.com/Homebrew/homebrew-cask/pull/139807/files#r1098893569) here:

> Can we install it globally into `/Library/Application Support` and symlink it to `~/Library/Application Support`?
> 
> Does symlinking the plugin not work?

As it seems like a solution that won't require changes to the cask DSL.

miccal (approved):

PR #139910, opened 2023-01-24
Create livecheck for rwts-pdfwriter

myzhang1029:

- [ ] `brew audit --cask --online <cask>` is error-free. (awaiting #139909)


p-linnane:
This is not necessary. GitHub URL's are automatically parsed by `livecheck`, so we do not need to manually specify if like this.

PR #140395, opened 2023-01-29
Add more cache file location for Google Drive

Ninzero:
It seems like that Google Drive will cache file structure at `~/Library/CloudStorage/GoogleDrive-abc@example.com/`. Therefore even if you have uninstalled Google Drive, Spotlight will still be able to index these files, result in difficulty in find actual files you need.


bevanjkay:
@Ninzero a zap stanza should not remove user-created files. The cache folder stores the files that Google Drive is syncing, they are not preference or configuration files. There is a chance that this could result in data loss for users, so it would be best to  leave it out.

Ninzero:
Though personally I do not think it can cause data loss after you uninstalled Google Drive, I agree with the conclusion to not remove these files automatically. Is there any stanza I can use to notify users that remove this location might be helpful?

PR #140448, opened 2023-01-31
Update microsoft-excel from 16.69.23010700 to 16.69.23011600

CrawX:
Created with `brew bump-cask-pr`.

PR #140447, opened 2023-01-31
Update microsoft-outlook from 16.69.23010700 to 16.69.23011802

CrawX:
Created with `brew bump-cask-pr`.

PR #140490, opened 2023-02-01
Update microsoft-outlook from 16.69.23010700 to 16.69.23011802

CrawX:

---
This is a continuation of #140447.
The previously used livecheck seems outdated as [this official resource](https://learn.microsoft.com/en-us/officeupdates/update-history-office-for-mac) lists 16.69.23011802 as the most recent version. I updated the livecheck to use that resource.

If this is deemed acceptable, I would also align #140448, #140449 and #140450 similarly. Either in this PR or in another one, whatever works better.


miccal:
Thank you @CrawX, but our experience with Microsoft Casks in the past means that we find it more convenient to use the `:header_match` strategy for `livecheck`'s.

 It should not take too long for these to be updated.

Issue #140273, opened 2023-01-27
Onyx 4.3.5 SHA256 mismatch error on macOS 13.2


gbstan:
### Verification

- [X] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [X] I have retried my command with `--force`.
- [X] I ran `brew update-reset && brew update` and retried my command.
- [X] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [X] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [X] I made doubly sure this is not a [checksum does not match](https://docs.brew.sh/Common-Issues#cask---checksum-does-not-match) error.

### Description of issue

I know this is a checksum does not match error. I have checked the version numbers on brew cask and on the website of the application and they are the same. I have tried to do an update of the cask via https://docs.brew.sh/Common-Issues#cask---checksum-does-not-match but got stuck. I would be grateful if someone could help.

The command `brew install --cask onyx` produces the following output:

```
Error: SHA256 mismatch
Expected: dd8c81e30a03f3e4cdcbc015d9207b6036bd40200dc51528dfe5d3d32817e3a3
  Actual: 5d4bd7ca16fbd2bd70a383f74e6c254d281776ae0a012d1484c392cc4dc5ee28
```

### Command that failed

brew install --cask onyx

### Output of command with `--verbose --debug`

```shell
/usr/local/Homebrew/Library/Homebrew/shims/shared/curl --disable --cookie /dev/null --globoff --show-error --user-agent Homebrew/3.6.20-143-g10845a1\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 13.2\)\ curl/7.86.0 --header Accept-Language:\ en --fail --max-time 5 --retry 3 --location --remote-time --output /Users/guy-bartstan/Library/Caches/Homebrew/api/formula.json --time-cond /Users/guy-bartstan/Library/Caches/Homebrew/api/formula.json --compressed --silent https://formulae.brew.sh/api/formula.json
[...]

PR #140548, opened 2023-02-03
onyx: fix URL handling for API installs

Rylan12:
See https://github.com/Homebrew/homebrew-cask/issues/140273

Unfortunately, there isn't a great way to do this other than duplicating the URL with each unique version number. Otherwise, the URL isn't included in the JSON variations and so users who try to install `onyx` will always get the macOS 12 version but the `sha256` for the correct version based on their OS version.

### DIFF
--- a/Casks/onyx.rb
+++ b/Casks/onyx.rb
@@ -1,56 +1,50 @@
 cask "onyx" do
+  sha256 :no_check
+
   on_el_capitan do
     version "3.1.9"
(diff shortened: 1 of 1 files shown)


bevanjkay (dismissed):

SMillerDev (approved):

carlocab (commented):
Wait, why did we have to get rid of all of the checksums?

miccal:
> Wait, why did we have to get rid of all of the checksums?

Because now the `url`'s are un-versioned and it was failing our CI checks with them in.

PR #140573, opened 2023-02-03
Calibre stopped supporting catalina on version 6.11.0

danielnetop:
According to calibre changelog (https://github.com/kovidgoyal/calibre/blob/master/Changelog.txt) version 6.12.0 now only supports big sur and newer versions of macOS

> - Update bundled Qt to 6.4 this means calibre on macOS is now only supported on Big Sur and newer

### DIFF
--- a/Casks/calibre.rb
+++ b/Casks/calibre.rb
@@ -15,7 +15,15 @@ cask "calibre" do
       skip "Legacy version"
     end
   end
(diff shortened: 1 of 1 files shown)


miccal (approved):

miccal:
Thank you @DanielNetoP.

PR #140639, opened 2023-02-06
ci: add NO_INSTALL_FROM_API env

bevanjkay:
I believe this should stop the issue we are seeing with audits failing, because the audit is being completed with the version present in the API, not the PR - I am not 100% sure if this is the long-term solution.


SMillerDev (approved):

bevanjkay:
Merging to see if this unblocks the PRs, @Rylan12 feel free to revert if there is a better solution.

PR #141290, opened 2023-02-13
docker: various updates

bevanjkay:
See discussion in https://github.com/Homebrew/homebrew-cask/pull/141286

### DIFF
--- a/Casks/docker.rb
+++ b/Casks/docker.rb
@@ -1,9 +1,29 @@
 cask "docker" do
   arch arm: "arm64", intel: "amd64"
 
-  version "4.16.2,95914"
-  sha256 arm:   "838eabe6cc42fe7e4be2cdb4d73924c61fc7982366dac2a9467793845851cb2e",
-         intel: "22eecb8ea07f10892d81cde07d614fb8b342163106133a75c4280a8e28787643"
+  on_catalina :or_older do
+    version "4.15.0,93002"
+    sha256 arm:   "fc8609d57fb8c8264122f581c0f66497e46e171f8027d85d90213527d6226362",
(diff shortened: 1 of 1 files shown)


p-linnane (approved):

ddelange:
formula wizard 🧙

many thanks!

Casks/docker.rb
```diff
@@ -1,9 +1,29 @@
...
+    livecheck do
+      skip "Legacy version"
+    end
+  end
+  on_big_sur :or_newer do
+    version "4.16.2,95914"
+    sha256 arm:   "838eabe6cc42fe7e4be2cdb4d73924c61fc7982366dac2a9467793845851cb2e",
+           intel: "22eecb8ea07f10892d81cde07d614fb8b342163106133a75c4280a8e28787643"
+
+    depends_on macos: ">= :catalina"
```
glensc, line 20: shouldn this be "> catalina"? as "=" in mathematics includes catalina as well

ddelange: afaik that value automatically propagates from the upstream (incorrect) appcast xml. no need to worry about it due to L15 above

bevanjkay: We have a CI check that requires the `depends_on` stanza to match the result of the appcast. Upstream currently have a discrepancy where the appcast is presenting that the latest version is compatible with catalina where the bundle actually isn't.

Because the `depends_on` is inside of a `on_big_sur :or_newer` block, it is only read on big sur or newer, so this causes the cask to pass CI, but also provide the correct latest version for catalina or older.

This solution is by no means perfect, and will be updated when upstream fix their appcast to match the compatibility of the bundle.

PR #141357, opened 2023-02-14
Update firefox from 109.0.1 to 110.0

p-linnane:
Created with `brew bump-cask-pr`.

PR #141373, opened 2023-02-15
ci_matrix: add `ci-skip-fetch` label

bevanjkay:
This will allow us to skip the fetching (by association the `sha256` check is also skipped). Each URL will still be checked to ensure they do not 404. Perhaps a better approach is to decide whether we test all languages in audit (unless specifically requested with a  flag)? Maybe when there are more than a particular number we choose a set at random?

Could be useful in the instance of Firefox - where the downloads timeout https://github.com/Homebrew/homebrew-cask/pull/141357

CC: @p-linnane


bevanjkay:
Closing in favour of changes to audit.

PR #141475, opened 2023-02-17
Update geekbench from 5.5.1 to 6.0.0

5ouma:

### DIFF
--- a/Casks/geekbench.rb
+++ b/Casks/geekbench.rb
@@ -3,9 +3,13 @@ cask "geekbench" do
     version "4.4.2"
     sha256 "3c46e630a28a0752afd702fc1cd379edd2420001be22302c932e61751284c0cc"
   end
-  on_high_sierra :or_newer do
+  on_catalina :or_older do
     version "5.5.1,503987"
     sha256 "04b06cb642e51230a3dfd07ce2d3a4ea696cb349583737622749174dc8747313"
+  end
+  on_big_sur :or_newer do
+    version "6.0.0,602754"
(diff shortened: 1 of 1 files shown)


Casks/geekbench.rb
```diff
@@ -3,10 +3,14 @@
     version "4.4.2"
     sha256 "3c46e630a28a0752afd702fc1cd379edd2420001be22302c932e61751284c0cc"
   end
-  on_high_sierra :or_newer do
+  on_catalina :or_older do
```
bevanjkay, line 6: @5ouma Unfortunately this conditional won't work when there are two `:or_older` options as it will read as having two version stanzas. 

If this version should apply to Catalina and High Sierra you will need to have two blocks with the same version listed. `on_high_sierra` and `on_catalina`

bevanjkay (dismissed):

Casks/geekbench.rb
5ouma: @bevanjkay Thank you for your reply.
How can I set multiple macOS versions to one cask version?

- Sierra or older -> `4.4.2`
- High Sierra to Catalina -> `5.5.1`
- Big Sur or newer -> `6.0.0`

bevanjkay: You would need to add a block for each MacOS version for High Sierra -> Catalina. With the same version information if necessary. 

```
on_high_sierra do
...
end
on_catalina do
...
end
```

5ouma: OK, Thank you! I'll do so.

miccal:
Thank you @5ouma.

miccal (approved):

PR #141495, opened 2023-02-17
mathpix-snipping-tool: update `livecheck`

miccal:


bevanjkay:
@miccal It looks like there's actually versioned downloads available here :)

p-linnane (approved):

PR #141496, opened 2023-02-17
whatsapp: update `conflicts_with`

miccal:


miccal:
I noticed that all the `microsoft-` Casks I updated recently were this way, and though it was neater.

bevanjkay:
@miccal It looks like we generally allow for both. I found reference to a discussion around forcing [""] style arrays from a number of years ago. https://github.com/Homebrew/brew/pull/2768#issuecomment-307720444
Benefit of [""] is that it is more globally understood as an array of strings, by non-rubyists.

Moulick:
As a non-ruby dev, `[""]` is definitely much easier to understand at first glance. Even people who might only know json will immediately understand `["a", "b"]` as an array of strings

bevanjkay:
My preference would be [""]

@Homebrew/cask

PR #141506, opened 2023-02-18
fix: command-tab-plus ("source not there" err.)

chrisgrieser:
This fixes the `source not there` error due to the app file name having changed since version 2. Also bumped version number and updated `zap` files.

I tried following [the contribution guidelines](https://github.com/Homebrew/homebrew-cask/blob/HEAD/CONTRIBUTING.md#getting-set-up-to-contribute), but it seems that since homebrew 4.0 made the casks tap unnecessary, I do not have a `Taps/homebrew/homebrew-cask` directory anymore. Trying `brew tap brew/cask` also was not successful, since it prompts for github username and password which are not supported by github anymore I have setup SSH for github and it works correctly when pushing to my private repos, so I am not sure what the problem here is. 

Due to these troubles, I wasn't able to follow the contribution guide and therefore could not check for `brew audit/style`. Sorry, probably bad timing to make the first contribution right after the 4.0 upgrade 😅 

---

### DIFF
--- a/Casks/command-tab-plus.rb
+++ b/Casks/command-tab-plus.rb
@@ -1,5 +1,5 @@
 cask "command-tab-plus" do
-  version "1.130,380"
+  version "2.6"
   sha256 :no_check
 
   url "https://noteifyapp.com/download/Command-Tab%20Plus.dmg"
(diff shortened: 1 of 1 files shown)


miccal:
Thank you @chrisgrieser.

miccal (approved):

PR #141514, opened 2023-02-18
Bump imhex to 1.27.1

gsingh93:
Update `imhex` cask to 1.27.1.

This was updated manually as the URL has changed (`x86_64` is appended). Should I be doing `arch intel: "x86_64"` instead and using `#{arch}` in the URL, or should I just leave it as is? There is no ARM build at the moment, so would that prevent ARM users from using it with Rosetta 2 (don't have an M1 to test this)? Or does this mean I should add `requires_rosetta`?


p-linnane (approved):
Since there is no dedicated arm64 version right now, the way you did it is correct.

Thanks for your first contribution to Homebrew @gsingh93!

PR #141537, opened 2023-02-19
Add CodeEdit.app

devicarus:

### DIFF
--- a/Casks/codeedit.rb
+++ b/Casks/codeedit.rb
@@ -1,6 +1,6 @@
 cask "codeedit" do
-  version "0.0.1-alpha.25,8e438ba"
-  sha256 "b4c05d73faacc6f70d4db1e06205653e855692455b016e4955ba27355611ba7f"
+  version "0.0.2-alpha.26,de40742"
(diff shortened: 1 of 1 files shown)


p-linnane:
Thank you for your submission to `homebrew-cask` @devicarus. I've added a `livecheck` stanza and adjusted the version & url to do what we discussed.

Please try to add a [`zap` stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap) for this cask.

samford (commented):
The dmg filename seems to reliably use the short commit hash of the release and we can technically match it (and the release tag) without having to check the release asset list HTML. However, it's likely safer to check the filename itself (in case upstream doesn't something weird), so I'm fine with using the latest release asset list approach. [I'll work on turning it (or a different approach that meets the same need) into a strategy after I work through some of my promised brew PRs.]

I've added a commit to tweak the `livecheck` block regex. Namely, upstream presumably won't create alpha releases forever, so we want to be able to match something like `1.2.3-beta.4` or `1.2.3` as well.

miccal:
Thank you @devicarus.

miccal (approved):

PR #141585, opened 2023-02-20
ghidra: Remove quarantine xattr from binaries

fabianfreyer:
Remove the quarantine flag on binaries executed by ghidra.


bevanjkay:
@fabianfreyer Quarantine support is a security feature of MacOS that we have gone to great lengths to support, casks should not be systematically circumventing quarantine support. If you personally choose to, you can pass the `--no-quarantine` flag to install/upgrade without quarantine.

PR #141646, opened 2023-02-21
Add livechecks for unversioned casks.

reitermarkus:
Depends on https://github.com/Homebrew/brew/pull/14745 and https://github.com/Homebrew/brew/pull/14758.


reitermarkus:
I think we need to tweak the audit:

```
Casks with a `livecheck` should not use `version :latest`.
```

That is true if the livecheck is actually functional, in this case there is a `skip` which is basically just documenting why `:extract_plist` cannot be used.

miccal (approved):

PR #141681, opened 2023-02-22
Rename mongodb-realm-studio to realm-studio

nrlquaker:

As of [this commit](https://github.com/Homebrew/homebrew-cask/pull/91079/commits/ace7943035245b728cad4c3f7c8b4d7ae5c71e60) it was renamed to `mongodb-realm-studio`, but in [version 12](https://github.com/Homebrew/homebrew-cask/commit/13210ab519479f27cdbe28a8203ce24145d19a54) MongoDB is not part of the the app name anymore. And there is no mentioning MongoDB on [the offsite](https://studio-releases.realm.io) for latest versions. One more example of the name [in the offsite](https://www.mongodb.com/docs/realm/studio/).

### DIFF
--- a/Casks/realm-studio.rb
+++ b/Casks/realm-studio.rb
@@ -1,4 +1,4 @@
-cask "mongodb-realm-studio" do
+cask "realm-studio" do
   version "13.0.2"
   sha256 "39f8a9bc620084190e511c86d1f3aafac1abb18234d803b8a6573838a6541b1a"
 


miccal:
```
==> Analytics
install: 57 (30 days), 136 (90 days), 501 (365 days)
```

bevanjkay:
We always approach renames for existing casks with hesitancy, because it results in existing users no longer receiving updates (renames are seen as deleted/added casks).
This will be the case until such a time that we have a working version of `formula_renames` for cask.

In this instance, I don't have a strong opinion either way, however the least disruptive method may be to leave the token until we can support renaming the correct way.

PR #141715, opened 2023-02-23
Upgrade signal from 6.6.0 to 6.7.0

gronke:


SMillerDev (approved):
Thanks @gronke   ! Without contributions like yours it'd be impossible to keep homebrew going with the high standards that users have come to expect from the project. You can feel good knowing that you've made the world a tiny bit better for homebrew users around the world! 👍 🎉

PR #141720, opened 2023-02-23
Update shottr from 1.6.1 to 1.6.2

janosorcsik:

### DIFF
--- a/Casks/shottr.rb
+++ b/Casks/shottr.rb
@@ -7,10 +7,16 @@ cask "shottr" do
   desc "Screenshot measurement and annotation tool"
   homepage "https://shottr.cc/"
 
+  # `version.json` provides a `latestVersion` field but it can reference an
+  # older version while the `package` field points to a file with a newer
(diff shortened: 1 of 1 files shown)


SMillerDev (approved):

samford (commented):
* The download button on the homepage lists 1.6 as the current version but this links to a dmg for 1.6.1 and the website also references 1.6.1 in the release notes.
* `version.json` contains a `latestVersion` field which is currently `1.6.0` but the `package` field is `https://shottr.cc/dl/Shottr-1.6.2.pkg`.

Unfortunately, we're presented with inconsistency in the version information, so we have to use our judgment. If we reference the website, 1.6.1 is the newest released version. If we reference the `package` field in `version.json`, 1.6.2 is newest.

It's unclear to me which is correct but if we simply want the newest available version, checking the `package` field seems to be the way to go (for now). I've pushed a commit that updates the `livecheck` block accordingly but I can modify this approach if we want to go in a different direction based on this commentary (e.g., checking the download URL on the homepage).

PR #141798, opened 2023-02-24
orion: various updates

miccal:
https://github.com/Homebrew/brew/pull/14804


Bo98:
This is in a `on_mojave` so you'll need to make it `on_mojave :or_older` for it to apply.

Bo98 (approved):

PR #141893, opened 2023-02-25
a-better-finder-attributes: move to a/a-better-finder-attributes

cho-m:

---

Experimenting with moving single cask to shard.

I think it will fail style assuming CI runs:
```console
❯ brew style a-better-finder-attributes
a/a-better-finder-attributes.rb:1:1: C: [Correctable] Missing frozen string literal comment.
cask "a-better-finder-attributes" do
^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```


p-linnane:
@cho-m Do we still need this open?

cho-m:
We should start sharding the repo.

Some things that may need to be done:
- [ ] Agree on sharding approach, e.g. the `a/` - `z/` directories? leave `[0-9].*` casks alone? etc.
- [ ] I mentioned on Slack some URLs that won't work for any Cask we migrate. Not sure if we want to handle before or just leave them broken (assuming low impact), e.g.
  - https://github.com/Homebrew/brew/blob/master/Library/Homebrew/api.rb#L123 (we could still migrate with this broken as I think it will take the fallback using formulae.brew.sh)
    ```rb
          raw_endpoint = "#{git_head}/Casks/#{name}.rb"
    ```
  - https://github.com/Homebrew/formulae.brew.sh/blob/master/_layouts/cask.html#L23 (I think this is only used for website)
    ```html
    <p><a target="_blank" href="{{ site.taps.cask.remote }}/blob/{{ c.tap_git_head }}/Casks/{{ token }}.rb">Cask code</a> on GitHub</p>
    ```
- [ ] Commit a small number of Cask migrations to test everything is working.
- [ ] Potential mass migration or migrate in phases
- [ ] Add some checks for new Casks.

carlocab (approved):
Just testing something. Feel free to ignore.

razvanazamfirei:
@reitermarkus, approach 2 makes a lot of sense to me. Last I checked, we would end up with ~500 directories. 

I've been playing around with md5 hashing the file name and taking a modulo of the hash to control the number of directories. If we use it to assign cask to directories, the distribution of casks/shard is reasonable. It would be decently easy to implement and audit, but very user-unfriendly.

PR #141821, opened 2023-02-25
Update bankid from 7.13.1 to 7.14.0

p-linnane:
Created with `brew bump-cask-pr`.

PR #141987, opened 2023-02-25
CI: Support ci-skip-livecheck label

samford:
This adds support for a `ci-skip-livecheck` label to cask repository CI, as a way of skipping livecheck audits when a check works locally and exclusively fails in the CI environment (e.g., https://github.com/Homebrew/homebrew-cask/pull/141821#pullrequestreview-1314545282). This is a rare situation and the label should not be used for any other purpose at this time.

I previously implemented support for this label in homebrew/core (and test-bot) to handle this situation (with the same restrictions on its use), so this commit aligns the core/cask repositories in this regard. The practical effects of the label differ but the fundamental idea remains the same.

My knowledge of cask CI is somewhat limited, so I would appreciate review from anyone with more experience. It seems fairly straightforward from what I've seen in the cask CI workflows and the related `audit` code in Homebrew/brew but it would be good to make sure I haven't overlooked anything.

### DIFF
--- a/cmd/lib/ci_matrix.rb
+++ b/cmd/lib/ci_matrix.rb
@@ -125,6 +125,10 @@ module CiMatrix
         "--appcast"
       end
 
+      if labels.include?("ci-skip-livecheck")
+        audit_exceptions << ["hosting_with_livecheck", "livecheck_version", "livecheck_min_os"]
+      end
+
       audit_args << "--except" << audit_exceptions.join(",") if audit_exceptions.any?
 
       runners(path).map do |runner|
--- a/.github/workflows/rerun-workflow.yml
+++ b/.github/workflows/rerun-workflow.yml
@@ -26,6 +26,7 @@ jobs:
         github.event.label.name == 'ci-retry' ||
         github.event.label.name == 'ci-skip-appcast' ||
         github.event.label.name == 'ci-skip-install' ||
+        github.event.label.name == 'ci-skip-livecheck' ||
         github.event.label.name == 'ci-syntax-only' ||
         !github.event.label.name
       )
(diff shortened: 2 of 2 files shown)


cmd/lib/ci_matrix.rb
```diff
@@ -125,6 +125,8 @@ def self.generate(tap, labels: [])
         "--appcast"
       end
 
+      audit_exceptions << "livecheck" if labels.include?("ci-skip-livecheck")
```
samford, line 128: This will simply skip all audits that have `livecheck` in the name, so we don't have to worry as much about keeping an explicit list of audits up to date if we add/remove them. However, it may skip more than we want.

For example, it's fine for `audit_appcast_and_livecheck` and `audit_latest_with_appcast_or_livecheck` to run, since they don't actually run livecheck. We primarily need to skip `audit_livecheck_version` and `audit_livecheck_min_os` but an explicit list would need to be kept up to date as audits are added/removed/renamed or this functionality would break.

Thoughts on this?

bevanjkay: @samford The `--except` options for `brew audit` for cask currently requires the audit `def` to be passed explicitly, this could be a bug in the [implementation](https://github.com/Homebrew/brew/blob/e86a8c2477032bda22c075f948f2c526c7ce4e02/Library/Homebrew/cask/audit.rb#L60-L66). 

```
      audit_exceptions << ["livecheck_version", "hosting_with_livecheck", "livecheck_min_os"] if labels.include?("ci-skip-livecheck")
```

samford: Oh shoot, that's right. I quickly skimmed that code and assumed the matching may be loose because of the use of `#include?` but, on closer inspection, it does require the explicit method name without the leading `audit_`. That's one way to resolve this question 😆

Thanks for the catch!

bevanjkay: I learnt this the other day when I added and fixed another label :) LGTM now

bevanjkay (approved):

PR #142047, opened 2023-02-27
Remove unecessary `rubocop:disable`

Rylan12:
Needed for https://github.com/Homebrew/brew/pull/14825


Rylan12:
Oof, looks like a better fix is probably to restrict the check on the Homebrew/brew side

reitermarkus:
> Oof, looks like a better fix is probably to restrict the check on the Homebrew/brew side

Yeah, I don't see a reason why this should suddenly not be style-checked.

Rylan12:
Since the PR this was intended to unblock has been merged, there's no need for this anymore.

PR #142063, opened 2023-02-27
Update screens from 4.12.7 to 4.12.8

miccal:
Created by `brew bump`

---

Created with `brew bump-cask-pr`.

PR #142080, opened 2023-02-28
ci_matrix: add label to skip livecheck_min_os audit

bevanjkay:
Adds the ability to skip the `livecheck_min_os` audit when upstream's sparkle feed reports an incorrect minimum OS.
Note that in these cases the issue should also be reported to upstream.

Unblocks: #142063

### DIFF
--- a/.github/workflows/rerun-workflow.yml
+++ b/.github/workflows/rerun-workflow.yml
@@ -27,6 +27,8 @@ jobs:
         github.event.label.name == 'ci-skip-appcast' ||
         github.event.label.name == 'ci-skip-install' ||
         github.event.label.name == 'ci-skip-livecheck' ||
+        github.event.label.name == 'ci-skip-livecheck-min-os' ||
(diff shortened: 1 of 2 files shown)


SMillerDev:
I think the website might not be updated? Why not just follow the livecheck? If people complain they should be redirected upstream.

bevanjkay:
@SMillerDev It is only the sparkle feed that is incorrect in the instance of #142063 - the bundle itself is still referencing the old minimum system version.

bevanjkay:
I am personally okay with holding PRs until upstream fixes the issue, or updating to match the sparkle feed in these instances (even if it may be incorrect). As users would experience the same treatment from the in-app sparkle updater.

SMillerDev:
I'm still inclined to follow upstream in this mistake

bevanjkay:
Looks like we're on the same page. 👍

miccal (approved):

PR #142311, opened 2023-03-05
Add MultiViewer for F1.app v1.14.0

malob:

### DIFF
--- a/Casks/multiviewer-for-f1.rb
+++ b/Casks/multiviewer-for-f1.rb
@@ -1,0 +1,36 @@
+cask "multiviewer-for-f1" do
+  arch arm: "arm64", intel: "x64"
+
+  on_arm do
+    version "1.14.0,98081841"
+    sha256 "0af0e0245ef417f0423bb6beba09184a2626d8d83c338f3e7e0419929e19ef9e"
+  end
+  on_intel do
+    version "1.14.0,98082227"
+    sha256 "1caa2f7e901d20d87b01919b93bdf1c7f45b858b0f57f6d55f567cbfa67ffd04"
+  end
+
+  url "https://releases.multiviewer.app/download/#{version.csv.second}/MultiViewer.for.F1-#{version.csv.first}-#{arch}.dmg"
+  name "MultiViewer for F1"
+  desc "Unofficial desktop client for F1TV"
+  homepage "https://multiviewer.app/"
+
+  livecheck do
+    url "https://api.multiviewer.dev/api/v1/releases/latest"
+    regex(/v?(\d+(?:\.\d+)+)/i)
+    strategy :json do |json|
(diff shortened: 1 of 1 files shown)


Casks/multiviewer-for-f1.rb
```diff
@@ -0,0 +1,36 @@
...
+  name "MultiViewer for F1"
+  desc "Unofficial desktop client for F1TV"
+  homepage "https://multiviewer.app/"
+
+  livecheck do
+    url "https://api.multiviewer.dev/api/v1/releases/latest"
+    regex(/v?(\d+(?:\.\d+)+)/i)
+    strategy :json do |json|
+      json["downloads"].select { |item| (item["platform"] == "darwin_#{arch}") && (item["type"] == "download") }
+                       .map { |item| "#{item["url"][regex, 1]},#{item["id"]}" }
```
samford, line 23:
```suggestion
    regex(%r{/([^/]+?)/MultiViewer[._-]for[._-]F1[._-]v?(\d+(?:\.\d+)+)[._-]#{arch}\.dmg}i)
    strategy :json do |json|
      json["downloads"].map do |item|
        match = item["url"]&.match(regex)
        next if match.blank?

        "#{match[2]},#{match[1]}"
      end
```

The `version` values that we want are interpolated into the cask `url`,so  there's maybe something to be said for matching these values from the `url` field of the `downloads` objects. This suggestion uses a regex to extract these values, which leads to a more common `strategy` block pattern (for casks). This may be a bit easier to reason about compared to the relative complexity of the existing approach in this PR.

The only notable difference is that this isn't restricted to only the objects where `type` is `"download"`, as this may not ever matter for dmg files.

miccal:
Thank you @malob and @samford.

miccal (approved):

Casks/multiviewer-for-f1.rb
```diff
@@ -0,0 +1,40 @@
...
+  on_arm do
+    version "1.14.0,98081841"
+    sha256 "0af0e0245ef417f0423bb6beba09184a2626d8d83c338f3e7e0419929e19ef9e"
+  end
+  on_intel do
+    version "1.14.0,98082227"
+    sha256 "1caa2f7e901d20d87b01919b93bdf1c7f45b858b0f57f6d55f567cbfa67ffd04"
+  end
+
+  url "https://releases.multiviewer.app/download/#{version.csv.second}/MultiViewer.for.F1-#{version.csv.first}-#{arch}.dmg"
```
f1multiviewer, line 13: @miccal if you want, there's another API to get the latest release:

`https://api.multiviewer.dev/api/v1/releases/latest/darwin_arm64/download`

bevanjkay: @f1multiviewer Thanks. We prefer to use versioned URLs for download where possible as it allows us to have a shasum check.

PR #142344, opened 2023-03-05
openzfs: update `on_#{arch}` block order

miccal:


bevanjkay:
@miccal If this is something we want to enforce we should setup a style audit for it.

SMillerDev (approved):

PR #142356, opened 2023-03-06
Update netron to 6.7.2

lutzroeder:
Update version and sha256


bevanjkay:
@lutzroeder Can you please update your script to include the `zap` stanza?

PR #142509, opened 2023-03-07
Added Thorium.app M109.0.5414.120-2

mietzen:


mietzen:
Just a remark, the main Repo containing the source code has +400 Stars: https://github.com/Alex313031/thorium

> An app from a code repository that is not notable enough (under 30 forks, 30 watchers, 75 stars).

I hope that this is a OR conjunction.

miccal:
The naming of this Cask does not conform to our [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference), and we already have a Cask called [`thorium`](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/thorium.rb), so to follow [our rules](https://docs.brew.sh/Acceptable-Casks#forks-and-apps-with-conflicting-names) this should be named `alex313031-thorium`.

You will also need to add a line `conflicts_with cask: "thorium"`.

mietzen:
> The naming of this Cask does not conform to our [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference), and we already have a Cask called [`thorium`](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/thorium.rb), so to follow [our rules](https://docs.brew.sh/Acceptable-Casks#forks-and-apps-with-conflicting-names) this should be named `alex313031-thorium`.
> 
> You will also need to add a line `conflicts_with cask: "thorium"`.

Thanks, I implemented your feedback 👍

Casks/alex313031-thorium.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  livecheck do
+    url "https://api.github.com/repos/Alex313031/Thorium-Special/releases"
+    regex(%r{/([^/]+?)/Thorium[._-]MacOS[._-]#{arch}\.dmg}i)
+    strategy :json do |json, regex|
+      json.map do |release|
+        curr_asset = release["assets"]&.find { |asset| asset["browser_download_url"]&.match(regex) }
+        next if curr_asset.blank?
+
+        curr_asset["browser_download_url"][regex, 1]
```
samford, line 22:
```suggestion
  # NOTE: This approach involves multiple requests and should be avoided
  # whenever possible. If upstream starts reliably providing `wine-stable` zip
  # files in every release, we should switch to `url :url` with
  # `strategy :github_latest`.
  livecheck do
    url "https://github.com/Alex313031/Thorium-Special/releases?q=prerelease%3Afalse"
    regex(%r{/([^/]+?)/Thorium[._-]MacOS[._-]#{arch}\.dmg}i)
    strategy :page_match do |page, regex|
      # Collect the release tags on the page
      tags = page.scan(%r{href=["']?[^"' >]*?/releases/tag/([^"' >]*?)["' >]}i)&.flatten&.uniq

      max_reqs = 4
      tags.each_with_index do |tag, i|
        break if i >= max_reqs

        # Fetch the assets list HTML for the tag and match within it
        assets_page = Homebrew::Livecheck::Strategy.page_content(
          @url.sub(%r{/releases/?.+}, "/releases/expanded_assets/#{tag}"),
        )
        matches = assets_page[:content]&.scan(regex)&.map { |match| match[0] }

        break matches if matches.present?
```

samford (commented):
Though [I technically created this `livecheck` block](https://github.com/mietzen/homebrew-thorium-cask/pull/1), we specifically avoid checks that use the GitHub API in first-party taps for now. As mentioned in the [related discussion](https://github.com/orgs/Homebrew/discussions/4299#discussioncomment-5207708), we have to worry about the API rate limit due to the number of checks we run (i.e., this was historically a problem in the old homebrew-livecheck repository before I got rid of the GitHub API checks).

This is one of those scenarios where it's unfortunately necessary to use the pattern where we iterate through release asset lists until we find a release with a matching asset. The GitHub API approach is more efficient and simple but there's some research to do and groundwork to lay before we could consider using it in our own taps (e.g., we need to be able to optionally make authenticated requests to increase the rate limit and we also would need to make sure that it couldn't conflict with the other GitHub API requests we make in CI, etc.). It's something that I plan to look into when I have the time but there's some other livecheck work that I'm pushing forward at the moment.

That said, this suggested check could potentially be made more efficient if we searched the release description text to identify the macOS releases in the `strategy` block but it may be a challenge because the related text varies. However, I've added "macos" to the release search query (in the `livecheck` block URL), so that should help a little (granted, the descriptions for non-macOS releases also appear in the results because they link to a macOS release).

miccal (approved):

PR #142618, opened 2023-03-09
Add irpf2023 1.0

lucasrodcosta:
App for income declaration and tax calculation made for Brazilian tax payers.

**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making any changes to a cask, existing or new, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask --online <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [x] `brew audit --new-cask <cask>` worked successfully.
- [x] `brew install --cask <cask>` worked successfully.
- [x] `brew uninstall --cask <cask>` worked successfully.

PR #142653, opened 2023-03-09
Delete `irpf2021` and `irpf2022`

lucasrodcosta:
These casks refer to older versions of `irpf`. [Another PR](https://github.com/Homebrew/homebrew-cask/pull/142618) is already under review to add the latest version.


bevanjkay (approved):

diegocoxta:
@lucasrodcosta sometimes we need to download older versions of this software to send "rectifications". IMO we need to maintain the last six versions of this cask because this is the rule to flag this software as deprecated by Receita Federal. Do you think it makes sense?

Screenshot 2023-03-18 at 16 07 22

lucasrodcosta:
Absolutely, @diegocoxta. I think you're right.
It's a good idea to revert this commit and open a new PR.

PR #142862, opened 2023-03-10
aleph-one: remove

p-linnane:
This has been having certificate errors for an extended period of time.

```
30 days: n/a
90 days: 6 (#4618)
365 days: 22 (#5088)
```


miccal (approved):

Amorymeltzer:
@p-linnane Sorry for my not knowing, but what does "certificate errors" mean in this context?  

----

I can open a separate issue to discuss it if preferred (cc @miccal) but I'd strongly urge reverting this if there's something moderately transient or repairable.  It's part of a set of casks (`marathon`, `marathon2`, `marathon-infinity`) all from the same place and repo, are they having the same issue?  Indeed, having any of those without `aleph-one` would be weird given the nature of the game/engine.  I brought this up a few years ago in #88604; most of that rationale holds up, with the major difference being the project has moved to a more regular release cadence.

p-linnane:
@Amorymeltzer Those other casks are also having the same issue, but I didn't get around to removing them. Basically the certificate chain of the homepage is incomplete. See [here](https://www.ssllabs.com/ssltest/analyze.html?d=alephone.lhowon.org&latest). This is causing livecheck to fail. 

I am open to reverting this if you'd like to open a PR and reach out to upstream about their certificate. They can use https://whatsmychaincert.com to generate a correct certificate chain.

Amorymeltzer:
Thanks for that.  I hadn't realized there was an issue: all four are in my (brief) `.brew_livecheck_watchlist` and I haven't run into any problems with livchecking them on my end.

The livecheck was changed most recently in afdf3c0c9c to get around the GitHub issues, am I correct that those got worked out on GitHub's end?  I might have my press releases mixed up.

p-linnane:
No, that was a permanent change on GitHub's end. If you want to open the PR to restore `aleph-one` I can write a new set of livechecks for those 4 casks using an alternative method of checking GitHub.

Amorymeltzer:
Thanks @p-linnane, I appreciate it.  If that works for you it works for me! #142900

Hopper262:
Certificate chain errors at alephone.lhowon.org have now been fixed. Thanks @p-linnane for the info, and @Amorymeltzer for bringing this to my attention!

PR #142873, opened 2023-03-10
gdat: update livecheck

p-linnane:
Download page URL has changed.

### DIFF
--- a/Casks/gdat.rb
+++ b/Casks/gdat.rb
@@ -9,7 +9,7 @@ cask "gdat" do
   homepage "https://www.getgdat.com/"
 
   livecheck do
-    url "https://sites.google.com/view/genealogical-dna-analysis-tool/home/download"
+    url "https://sites.google.com/view/genealogical-dna-analysis-tool/home_1/download"
     regex(/Genealogical DNA Analysis Tool (\d+(?:.\d+)+)/i)
   end
 


reitermarkus (approved):

reitermarkus:
@p-linnane, why `ci-syntax-only`? Clearly it isn't just a syntax update.

p-linnane:
We typically use that label when only updating livechecks.

reitermarkus:
That's not what it should be used for though. I don't see a reason why we shouldn't run CI on livecheck URL updates.

The description of the label is accurate: Use only for bulk changes. This label mostly exists for RuboCop changes affecting hundreds of casks.

p-linnane:
I'm not sure. It's used in `homebrew-core` too when making livecheck changes. Tagging @samford to provide some insight here.

samford:
I can only speak for homebrew/core but we use `CI-syntax-only` there because we don't have a way of running only syntax and livecheck tests and CI can be pretty heavy over there by default (with compilation, testing dependents, etc. across various macOS versions and Linux).

Basically, we decided years ago when we migrated the homebrew-livecheck tap into core (and brew) that the overhead of running CI wasn't worth it simply to test livecheck changes (i.e., they're easy to test locally when reviewing and usually there's not a notable difference on CI). When we made that decision, I was also churning out `livecheck` block PRs in core fairly aggressively (there were times where I did 40-50 in one day) and those PRs would have had a notable affect on CI at the time. It was initially intended as a short-term workaround but it's been necessary for longer than anticipated.

I imagine CI isn't as heavy in the cask repositories but I'm not very familiar with how long a typical PR takes to run on CI (or if we end up with long queues when there are a lot of PRs in a short period of time), so I defer to the cask maintainers on what's preferable. I agree that we ideally want to be testing livecheck changes on CI, so if the folks reviewing cask PRs think it's fine to reserve `ci-syntax-only` for bulk `livecheck` block PRs only, then it makes sense to me.

All that said, our original goal was to have a label [in homebrew-core] that would only check syntax and run livecheck on CI (probably `CI-livecheck`). It was always something we were going to implement but we just haven't gotten around to it yet. I'll have to bump it up my to-do list for when I finish up the livecheck stuff I'm working on in Homebrew/brew. I don't have any in-progress work for this (and have plenty else to work on), so if anyone else wants to work on it, feel free.

p-linnane:
That makes sense. There are no queues in the cask repos, and the PR's run quickly since there's no compilation, so there should be no issue running livecheck updates without `ci-syntax-only`. Thanks for bringing it up @reitermarkus!

PR #142918, opened 2023-03-12
Speed up CI.

reitermarkus:
Two improvements:

- We are populating cache on macOS 11, but running mostly macOS 12, resulting in gems taking ~1m to install even on a cache hit. Fix caching by including the macOS version in the cache key. (~1m improvement)
- Run `generate-matrix` and `syntax` on `macos-latest`, since `homebrew/core` and `homebrew/cask` are already tapped and don't need to be cloned from scratch. (~4m improvement per task, i.e. ~9m improvement overall given they run sequentially).

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -19,10 +19,11 @@ jobs:
     if: startsWith(github.repository, 'Homebrew/')
     strategy:
       matrix:
-        os:
-          - macos-11.0
+        runner:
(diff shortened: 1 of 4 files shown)


miccal (dismissed):

reitermarkus:
One more small improvement: Skip `readall` in the `syntax` job if we're already running it in a job for a cask.

miccal (approved):

PR #143065, opened 2023-03-14
Update microsoft-outlook from 16.70.23021201 to 16.71.23031200

bevanjkay:
Created by `brew bump`

---

Created with `brew bump-cask-pr`.

PR #143098, opened 2023-03-15
Try fixing spurious `automerge` cancellations.

reitermarkus:
Not sure if this will work, if not, remove it altogether.

See https://github.com/Homebrew/homebrew-cask/pull/143065#issuecomment-1469004126.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -14,7 +14,7 @@ on:
       - ready_for_review
 
 concurrency:
-  group: automerge-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


miccal (dismissed):

Bo98:
For `pull_request_target`, `github.ref` will point to the state of `master` rather than the PR itself so I feel like this won't fix it.

`github.event.pull_request.number` (or similar) seems relevant.

miccal (dismissed):

miccal (approved):

PR #143139, opened 2023-03-16
Add Rewind

nihaals:

### DIFF
--- a/Casks/rewind.rb
+++ b/Casks/rewind.rb
@@ -2,7 +2,7 @@ cask "rewind" do
   version :latest
   sha256 :no_check
 
-  url "https://download.rewind.ai/Rewind.pkg"
+  url "https://download.rewind.ai/Rewind.dmg"
   name "Rewind"
   desc "Record and search your screen and audio"
   homepage "https://www.rewind.ai/"
@@ -10,7 +10,7 @@ cask "rewind" do
   depends_on arch:  :arm64,
              macos: ">= :monterey"
 
-  pkg "Rewind.pkg"
+  app "Rewind.app"
 
   postflight do
     # Description: Ensure console variant of postinstall is non-interactive.
--- a/Casks/rewind.rb
+++ b/Casks/rewind.rb
@@ -12,21 +12,6 @@ cask "rewind" do
 
   app "Rewind.app"
 
-  postflight do
-    # Description: Ensure console variant of postinstall is non-interactive.
-    # This is because `open "${app_dir}"` is called from the postinstall
-    # script of the package and we don't want any user intervention there.
-    retries ||= 3
-    ohai "The Rewind package postinstall script launches the Rewind app" unless retries < 3
-    ohai "Attempting to close Rewind.app to avoid unwanted user intervention" unless retries < 3
-    return unless system_command "/usr/bin/pkill", args: ["-f", "/Applications/Rewind.app"]
-
-  rescue RuntimeError
-    sleep 1
-    retry unless (retries -= 1).zero?
-    opoo "Unable to forcibly close Rewind.app"
-  end
-
   zap trash: [
     "~/Documents/rewind_logs_*.zip",
     "~/Library/Application Support/com.memoryvault.MemoryVault",
--- a/Casks/rewind.rb
+++ b/Casks/rewind.rb
@@ -12,6 +12,8 @@ cask "rewind" do
 
   app "Rewind.app"
 
+  uninstall quit: "com.memoryvault.MemoryVault"
+
   zap trash: [
     "~/Documents/rewind_logs_*.zip",
     "~/Library/Application Support/com.memoryvault.MemoryVault",
--- a/Casks/rewind.rb
+++ b/Casks/rewind.rb
@@ -7,6 +7,9 @@ cask "rewind" do
   desc "Record and search your screen and audio"
   homepage "https://www.rewind.ai/"
 
+  depends_on arch:  :arm64,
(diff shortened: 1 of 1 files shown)


nihaals:
Failed due to being Apple silicon only

miccal:
> Failed due to being Apple silicon only

The `.app` is a Universal binary, so it is not arm-only.

Screenshot 2023-03-16 at 11 09 46

nihaals:
> > Failed due to being Apple silicon only
> 
> The `.app` is a Universal binary, so it is not arm-only.
> 
> Screenshot 2023-03-16 at 11 09 46

I noticed the same thing but they [officially and explicitly say](https://help.rewind.ai/en/articles/6705951-what-hardware-is-required-to-run-rewind) they only support Apple silicon and I've seen screenshots of the pkg failing on Intel Macs (although this was on old versions and I haven't been able to test the cask myself on an Intel Mac).

miccal:
The version has changed five times since you opened this PR!

bevanjkay:
The `sparkle` feed provides a versioned download URL - https://updates.rewind.ai/builds/main/b8514.1-main-e1b1c5a.zip - which contains the `.app` instead of the `.pkg`.
I'm not sure if the `.pkg` affords any extra that the versioned does not. But if they result in the same files installed, then the versioned download would be preferred.

nihaals:
> The version has changed five times since you opened this PR!

That was something I wanted to mention, should this be `:latest`? They seem to deploy every time there's a commit to `main` and some days there can be a lot of updates, I would definitely say it's unreasonable to expect Homebrew to stay up to date based on the update rate I've seen for the past few months.

> The `sparkle` feed provides a versioned download URL - https://updates.rewind.ai/builds/main/b8514.1-main-e1b1c5a.zip - which contains the `.app` instead of the `.pkg`.
> 
> I'm not sure if the `.pkg` affords any extra that the versioned does not. But if they result in the same files installed, then the versioned download would be preferred.

I had a similar thought but assumed the pkg would be preferred due to being the official installation method (and could be updated in the future). There is a postinstall script but when I looked at it I was only checking it opening after install. I'm not very experienced with pkgs so it would be good to get someone else's thoughts on what the pkg does other than copying and opening the app. The main thing I would be weary of is the pkg getting updated in the future, but right now I can't think of anything the app does which would require a pkg over a simple app.

p-linnane:
We have historically dropped support for these kinds of casks that update with every commit (see https://github.com/Homebrew/homebrew-cask/pull/135413). If that is how this application operates, I am not in favor of adding it to `homebrew-cask`.

nihaals:
Wouldn't this cask be fairly stable with `:latest` and match the example in the cookbook of "for example, [`chromium.rb`](https://github.com/Homebrew/homebrew-cask/blob/f3e9de24ba57d7b1d949132504e581759725d0c5/Casks/chromium.rb#L4) releases multiple versions a day"? What would be the disadvantage of having the cask in the repo? Since the cask is `auto_updates` it feels like having it in the repo would still provide the value of easy install/uninstall and `--zap` while not needing many changes over time.

bevanjkay:
If this cask is to be added, `:latest` would be the best option. It is worth noting that the behaviour for `:latest` has changed in the last few months, when installed, a `shasum` is recorded, and then if `--greedy` is passed to `brew upgrade` the cask will be downloaded and the `shasum` compared with the installed version.

For the cask to have `:latest` as version, we would need to also utilise an unversioned url.

razvanazamfirei:
The download link has changed to `https://download.rewind.ai/Rewind.dmg`. I've updated the cask and it passes local audits on arm. My questions is if the uninstall stanza is still required.

razvanazamfirei:
For checks without the arch limitations: https://github.com/Homebrew/homebrew-cask/pull/144244/checks

SMillerDev:
This looks to me like it's changing way too fast to be maintainable here. Maybe a tap would be a better fit. Especially the changing download location is a big red flag to me.

razvanazamfirei:
Agreed, @SMillerDev. @nihaals, Thank you for opening the PR and your work on this. As it stands, the unpredictability in development, as well as the potential maintenance effort required to address the changes suggest that perhaps now is not the best time for the cask to be included. Putting aside the frequent version changes, silently changing the download link and format in which the app is delivered are red flags (they lead to signature verification failures in the past).

If the development trajectory becomes more predictable, we're more than happy to take another look. In the meantime, you can consider creating a [tap](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap). The cask will require minimal modifications.

Thank you again for your contribution!

nihaals:
And thanks for spotting the switch to a dmg, I wouldn't have noticed and it feels much nicer than using a pkg.

nihaals:
@unitof I did make a [tap](https://github.com/nihaals/homebrew-rewind), although I didn't try to make it as user-friendly and public as possible and am mostly using it for personal use. I knew about the appcast and we spoke about it above. Now that Rewind uses a dmg, using the URLs in the appcast would be fine, but it wouldn't be viable to keep it up to date in Homwbrew (sometimes they merge multiple times in an hour) and I wanted to keep it as close to a Homebrew cask as possible (so avoiding things like using GitHub Actions to automatically update it every x minutes). It sounds like Rewind is currently working towards 1.0 which could reduce the number of releases and make it more maintainable though so I think it would be best to see what happens.

unitof:
@nihaals Ah! My bad, missed you'd had the appcast earlier. Makes sense, thanks!

PR #143196, opened 2023-03-16
Add New Cask Zed

razvanazamfirei:

The download path is difficult to find; it's in a JSON disguised as a .dmg file, which redirects to a objects.githubusercontent.com path. It will fail for lack of a verified url, but otherwise everything looks ok.

### DIFF
--- a/Casks/zed.rb
+++ b/Casks/zed.rb
@@ -1,23 +1,36 @@
 cask "zed" do
-  version "0.77.3"
+  version "0.77.3,95639680"
   sha256 "7a10a5d6131fb8578908fb19d3a8c0dd0babecfeda81e57abc7c823ce57db7ae"
 
-  url "zed"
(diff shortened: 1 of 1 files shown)


p-linnane:
I'm not crazy about adding `objects.githubusercontent.com` as `verified`. Kind of defeats the purpose since we can't give it more than that...Not sure how else we could handle this.

razvanazamfirei:
Not wild about this either. The `objects.githubusercontent.com` URL is coming from the verified domain, I guess we could dynamically set the verified URL to the URL parsed from `https://zed.dev/api/releases/#{version.csv.second}/Zed.dmg`. This seems silly and I haven't been able to get it to work. I'll reach out to the devs and see if they're willing to make changes, but I'd also be okay with not including the cask for now.

razvanazamfirei (commented):
Should be ready now! @p-linnane @miccal

miccal (approved):

PR #143431, opened 2023-03-20
Add symlinks to AWS Session Manager cask

carlfugate:


miccal:
Thank you @carlfugate.

miccal (approved):

bevanjkay:
@miccal Should the `system` command be inside of a postflight install block here?

miccal:
No, the `system` block is not necessary.

Both Intel and Arm packages install into `/usr/local` and having arch-specific downloads automatically solves the problem of the symlinks.

carlfugate:
> @miccal Should the `system` command be inside of a postflight install block here? 

I must confess this was my very first time even working with formulas/casks and couldn't find a good example of how to do this.  I completely agree it should be a post install process. If you can help point me to an example I'm happy to change it.

bevanjkay:
@miccal Oh I didn't see you had removed the `system` command. I must have been seeing a cached version of the file. Thanks!

carlfugate:
First, thanks for all of the improvements @miccal as this is in much better shape.  I'm curious about the removal of the symlink creation (which was my whole initial purpose of the change) as well as leaving the binary install location to /usr/local instead of using the Homebrew prefix.

PR #143516, opened 2023-03-22
Various: Fix `arch` ordering

miccal:


Casks/biscuit.rb
```diff
@@ -1,21 +1,14 @@
...
-  end
+  version "1.2.27"
+  sha256 arm:   "b2bd9101f9eec59430db3586387cf5bcd077b06115f244ecdff9de1f4f329ac4",
+         intel: "4eff2c48e3ce24af3ffeb814ed46f5a21406516b0c094763f5896e73bc82bd9b"
 
+  url "https://github.com/agata/dl.biscuit/releases/download/v#{version}/Biscuit-#{version}#{arch}.dmg",
+      verified: "github.com/agata/dl.biscuit/"
   name "Biscuit"
-  desc "Browser to organize apps"
+  desc "Browser to organise apps"
```
bevanjkay, line 11: @miccal I think that generally we should be using American English, as it is standard for coding conventions.

miccal: I disagree:

https://github.com/Homebrew/brew/blob/master/docs/Prose-Style-Guidelines.md#guidelines

bevanjkay: Thanks, I hadn't seen that style guide, but saw someone mention Americanised descriptions the other day.

bevanjkay (approved):

PR #143634, opened 2023-03-23
gcc-arm-embedded 12.2.mpacbti-rel1

pthariensflame:


pthariensflame:
This failure is due to livecheck, but I'm not sure how to resolve it.  The new version is [genuinely succeeding the old one on the page](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads); I take it livecheck doesn't like the fact that the alphabet places `m` before `r`.  Any help would be appreciated greatly!

miccal (approved):

PR #143646, opened 2023-03-23
rubocop: Fix `Cask/NoOverrides` offenses

issyl0:
- This is a companion PR to https://github.com/Homebrew/brew/pull/15013, which itself is a prerequisite for https://github.com/Homebrew/brew/pull/14976.
- Fix the offenses that the new `Cask/NoOverrides` RuboCop found.

-----

### DIFF
--- a/Casks/ableton-live-intro.rb
+++ b/Casks/ableton-live-intro.rb
@@ -7,7 +7,8 @@ cask "ableton-live-intro" do
     sha256 "68ae4aa7c1a4286f77de340f5e2556442ea17a73976c3f90c9dce1c75dc8234a"
   end
   on_mojave :or_newer do
-    sha256 "70fb6f33c7e03f7c6290c7f5d5a6bee4454df272527c5fdd5dfd045b7d76f1eb"
+    sha256 arm:   "70fb6f33c7e03f7c6290c7f5d5a6bee4454df272527c5fdd5dfd045b7d76f1eb",
+           intel: "ad24ed3cecd392464efd8b6bd3a0b65f8f7b624fdd07b0315e266050d97f77e2"
   end
 
   url "https://cdn-downloads.ableton.com/channels/#{version}/ableton_live_intro_#{version}_#{arch}.dmg"
--- a/Casks/ableton-live-intro.rb
+++ b/Casks/ableton-live-intro.rb
@@ -1,10 +1,12 @@
 cask "ableton-live-intro" do
-  arch = "intel"
-  version "11.2.11"
-  sha256 "68ae4aa7c1a4286f77de340f5e2556442ea17a73976c3f90c9dce1c75dc8234a"
+  arch arm: "universal", intel: "intel"
 
+  version "11.2.11"
+
+  on_high_sierra :or_older do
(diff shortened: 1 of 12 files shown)


Casks/blockblock.rb
```diff
@@ -1,22 +1,23 @@
...
+      args:       ["-install"],
+      sudo:       true,
+    }
 
+    uninstall script: {
+      executable: "#{staged_path}/BlockBlock Installer.app/Contents/MacOS/BlockBlock Installer",
+      args:       ["-uninstall"],
+      sudo:       true,
+    }
+  end
```
Rylan12, line 20: These seem like they should be in the top-level and not specific to `catalina :or_newer`, no?

Casks/ableton-live-intro.rb
```diff
@@ -1,8 +1,11 @@
 cask "ableton-live-intro" do
   arch = "intel"
```
Rylan12, line 2: This should be replaced with an `arch` stanza (this also applies to the other `ableton-live-*` casks):

```suggestion
  arch arm: "universal", intel: "intel"
```

Casks/transcribe.rb
```diff
@@ -3,11 +3,12 @@
 
   livecheck_version = "11"
```
Rylan12, line 4: What are our current thoughts on using variables like this? Sometimes, we'll just put a different `livecheck` block in each `on_{system}` block to avoid this pattern

Rylan12 (commented):
I think there's some inconsistency about the ordering of the `on_{system}` blocks. Probably worth getting that order consistent now to avoid more PRs later since that's the ultimate goal

Casks/transcribe.rb
issyl0: This has gone now thanks to Miccal! #143675.

issyl0:
Rebased on `master` after #143675. I'll deal with the rest of the review comments and CI failures soon.

Casks/ableton-live-intro.rb
reitermarkus: Wouldn't `universal` work for both?

Casks/blockblock.rb
issyl0: Oooh yes, this was from before I realised that not every stanza is overrideable.

Casks/ableton-live-intro.rb
issyl0: The `on_high_sierra` stanza has an arch of `intel` so I don't think so?

issyl0:
Thanks for the reviews! I've been through and:

- Made the `on_system` blocks go in order of oldest => newest OS.
- Ensured that all the Casks pass `brew readall`.
- Checked that any unnecessarily overridden stanzas (eg `install` and `uninstall`) are not anymore.
- Ran the new `Cask/NoOverrides` cop to check there's no new offenses since my rebase on `master` compared to yesterday.

I think this should be good to go? 🙏🏻

miccal (approved):

reitermarkus:
@Homebrew/cask, are the `on_high_sierra` blocks on the `ableton` casks even valid? The `version` and `url` are the same regardless of the `on_high_sierra` block it seems.

MikeMcQuaid:
Nice work getting this over the line @issyl0 and thanks for all the help, cask folks!

PR #143711, opened 2023-03-24
odrive: fix uninstall, add zap

unitof:

### DIFF
--- a/Casks/odrive.rb
+++ b/Casks/odrive.rb
@@ -16,6 +16,13 @@ cask "odrive" do
 
   pkg "odrivesync.#{version}.pkg"
 
-  uninstall quit:    "com.oxygencloud.odrive",
+  uninstall quit:    "com.oxygen.odriveapp",
             pkgutil: "com.oxygen.odrive.*"
+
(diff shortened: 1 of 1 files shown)


miccal (approved):

bevanjkay:
If not removing `.odrive` on uninstall causes a reinstall to fail, does that cause the `brew upgrade` command to not work? @unitof

unitof:
I'm not actually sure; I haven't tested it or see how it handles self-updates.

What's weird/annoying is that this app stores its main background-running binary in `~/.odrive/bin/#{version}/odriveapp.app`. `/Applications/odrive.app` is just a launcher.

However in my testing for this PR, it seems the install PKG simply checks if `~/.odrive` directory exists, and doesn't touch it if it does. So you end up with a nonfunctional launcher if it exists but is missing the binary.

But `~/.odrive` also contains user preferences and logs so I didn't want to delete it outside of a zap.

bevanjkay:
Sounds like it's good to go as-is, and we can update if we see any future issues. Thanks.

PR #143739, opened 2023-03-25
Fix ableton casks.

reitermarkus:
See https://github.com/Homebrew/homebrew-cask/pull/143646#issuecomment-1483739028.

This is a weird case: `on_high_sierra` should have the `intel` URL, while later versions can use the `universal` URL for both ARM and Intel.


reitermarkus:
The `universal` version also works for High Sierra, so let's just use the `universal` URL for simplicity.

bevanjkay (approved):

PR #143791, opened 2023-03-26
Add airvpn/hummbingbird-1.2.1

sebkraemer:
This PR adds hummingbird cask, a multiplatform vpn client cli program provided by AirVPN team using openvpn 3.

### DIFF
--- a/Casks/hummingbird.rb
+++ b/Casks/hummingbird.rb
@@ -7,7 +7,7 @@ cask "hummingbird" do
 
   url "https://eddie.website/repository/hummingbird/#{version}/hummingbird-macos-#{arch}-notarized-#{version}.zip",
       verified: "eddie.website/repository/hummingbird/"
-  name "hummingbird"
+  name "Hummingbird"
   desc "OpenVPN 3 client"
   homepage "https://airvpn.org/hummingbird"
 
--- a/Casks/hummingbird.rb
+++ b/Casks/hummingbird.rb
@@ -8,7 +8,7 @@ cask "hummingbird" do
   url "https://eddie.website/repository/hummingbird/#{version}/hummingbird-macos-#{arch}-notarized-#{version}.zip",
       verified: "eddie.website/repository/hummingbird/"
   name "hummingbird"
-  desc "OpenVPN 3 client based on AirVPN's OpenVPN 3 library fork"
+  desc "OpenVPN 3 client"
   homepage "https://airvpn.org/hummingbird"
 
   livecheck do
--- a/Casks/hummingbird.rb
+++ b/Casks/hummingbird.rb
@@ -1,0 +1,42 @@
+cask "hummingbird" do
+  version "1.2.1"
+
+  on_intel do
+    sha256 "89be6ad9a198ce8589e19dd236881d24b6a1d415b748c5d4ef768ff94786fe9d"
+
+    url "https://eddie.website/repository/hummingbird/#{version}/hummingbird-macos-x86_64-notarized-#{version}.zip",
+        verified: "eddie.website"
+  end
+  on_arm do
+    sha256 "3ac509d0b8831b827e19f9945b588703c4f107e87e1ad8dbc19421fb2c7683c0"
+
+    url "https://eddie.website/repository/hummingbird/#{version}/hummingbird-macos-arm64-notarized-#{version}.zip",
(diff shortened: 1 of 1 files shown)


sebkraemer:
Ok so the auditing fails but I cannot change that. Does that mean that the package cannot be added? It is the official source of an established service provider (AirVPN).

I'm leaving this as a draft however I do not plan any changes until discussion on how to proceed. Be kind, this is my first homebrew submission :)

Casks/hummingbird.rb
```diff
@@ -0,0 +1,28 @@
...
+    regex(/eddie[._-]selected[._-]version">v?(\d+(?:\.\d+)+)</i)
+  end
+
+  depends_on macos: ">= :high_sierra"
+
+  binary "hummingbird-macos-#{arch}-#{version}/hummingbird"
+
+  postflight do
+    set_ownership("#{staged_path}/hummingbird-macos-#{arch}-#{version}/hummingbird", user: "root")
+  end
```
miccal, line 25:
```
warning: Unexpected method 'arch' called during postflight on Cask hummingbird.
```

sebkraemer: Hi @miccal, thanks for the review and the improvements regarding DSL usage!

> ```
> warning: Unexpected method 'arch' called during postflight on Cask hummingbird.
> ```

Looks to me as if that was introduced by your commit. I hoped to avoid if-else-casing to postflight but didn't see a way to use `#{arch}`. Actually I think I got an error, it did not find the method. 🤔

bevanjkay: @sebkraemer I have opened https://github.com/Homebrew/brew/pull/15064 to hopefully fix the issue.

sebkraemer: I see. I'll happily rebase once a fix gets merged. (y)

Casks/hummingbird.rb
```diff
@@ -0,0 +1,28 @@
...
+    url "https://airvpn.org/macos/hummingbird/"
+    regex(/eddie[._-]selected[._-]version">v?(\d+(?:\.\d+)+)</i)
+  end
+
+  depends_on macos: ">= :high_sierra"
+
+  binary "hummingbird-macos-#{arch}-#{version}/hummingbird"
+
+  postflight do
+    set_ownership("#{staged_path}/hummingbird-macos-#{arch}-#{version}/hummingbird", user: "root")
```
reitermarkus, line 24: Why does this need to be owned by `root`?

sebkraemer: I submitted a ticket to AirVPN support and they replied the following (shortened):

>  Hummingbird is linked against tthe OpenVPN3-AIrVPN library and is used to connect a system to some Virtual
> Private Network. It also adds features to check DNS settings (to respect the VPN DNS push, if any), and to
> prevent leaks via packet filtering tool rules (pf, iptables and nftables are supported).
> Therefore, root privileges are required to:
> -  add and remove packet filtering rules,
> - modify and restore DNS settings,
> - modify and restore routing table entries,
> - modify and restore default gateway,
> - bring up and configure a tun (virtual network interface).
>
>Hummingbird doesn't strictly need to be owned by root, although it is a good security practice for peculiar cases.
> For example Eddie, which may run Hummingbird, will refuse to run it with root privileges if Hummingbird is not
> owned by root. Some privilege escalation attacks are efficiently prevented when you run something with root
> privileges only if that something is owned by root.
> [..]

In practice, the hummingbird will refuse to run if not owned by root.

reitermarkus (dismissed):

reitermarkus (approved):

FredrikWendt:
What the heck happened here? There was a program called Hummingbird that is now silently replaced with a completely different application? Is this spyware that was approved and now being rolled out?

FredrikWendt:
Sorry for the noise - I had this other tap installed and the name collision obviously prioritized silently installing a different application than the one I currently had installed.

https://github.com/finestructure/Hummingbird#installation

sebkraemer:
It seems the issue can be overcome by using the full tap+cask name, however I don't see why one shouldn't avoid the problem in the first place and proposed a rename in PR #145837.
Actually I had initially named the cask that way but then saw that another airvpn application, "eddie", was already casked without the prefix.

PR #143806, opened 2023-03-27
orbstack: Enable auto_updates

kdrag0n:
OrbStack includes an auto-updater, which can be triggered from either the GUI (Check for Updates) or CLI (`orb update`). Update notifications are sent automatically.

### DIFF
--- a/Casks/orbstack.rb
+++ b/Casks/orbstack.rb
@@ -16,6 +16,7 @@ cask "orbstack" do
     strategy :header_match
   end
 
+  auto_updates true
   depends_on macos: ">= :monterey"
 
   app "OrbStack.app"


kdrag0n:
I'm the developer of OrbStack and I'm trying to add first-class support for brew cask. I had a question related to this: Does `auto_updates` mean that Homebrew will never update the app by itself?

The reason I'm asking is because if Homebrew will update the app anyway, then I'd have to add a postflight script to complete the update process for users who only use the Docker CLI (and hence never interact with OrbStack's CLI tools or GUI). If `auto_updates` prevents brew from updating it, then this isn't an issue because the GUI will complete the update.

bevanjkay:
@kdrag0n When `auto_updates` is set to `true` - Homebrew won't list the app as outdated when running `brew outdated` or upgrade the app when running `brew upgrade`. However, if the user passes the `--greedy` or `--greedy-auto-updates` flag to either command then it will be listed/upgraded.

kdrag0n:
Got it, thanks. In that case I think I'll add the postflight script to make sure the update is completed even with `--greedy`.

bevanjkay (approved):

PR #143807, opened 2023-03-27
orbstack: Add hooks and more complete uninstall/zap

kdrag0n:
- Add postflight hook to complete update as discussed in https://github.com/Homebrew/homebrew-cask/pull/143806. Does nothing on first install.
- Add uninstall and zap hooks for VM cleanup (does nothing in v0.5.0 and v0.5.1, but already implemented for future versions)
- Clean up GUI app state in ~/Library on zap. (`macvirt` is the project's codename)

**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making any changes to a cask, existing or new, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask --online <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

PR #143867, opened 2023-03-27
fig: Remove broken zap script

kdrag0n:
The script is part of the app but runs after the zap, not before, so it no longer exists:

```
==> Running uninstall script /Applications/Fig.app/Contents/MacOS/fig-darwin-universal Error: uninstall script /Applications/Fig.app/Contents/MacOS/fig-darwin-universal does not exist.
```

Zapping implies uninstalling, which happens first, so the script isn't necessary anyway.

Discussed in #143807.

### DIFF
--- a/Casks/fig.rb
+++ b/Casks/fig.rb
@@ -34,21 +34,17 @@ cask "fig" do
               "io.fig.cursor",
             ]
 
-  zap script: {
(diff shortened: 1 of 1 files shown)


reitermarkus (approved):

bevanjkay:
Can be reinstated once https://github.com/Homebrew/brew/issues/15070 if/when is completed

PR #143972, opened 2023-03-29
Add livechecks for unversioned casks.

reitermarkus:


bevanjkay:
@reitermarkus We will need to add the colorpickers to UnversionedCaskChecker so that it can find the `.plist` files. The errors in `google-japanese-ime` will need to be ignored, because they cannot be uninstalled without zap (shared amongst Google apps).

reitermarkus:
Yes, I have a branch for `Homebrew/brew` but haven't pushed it yet. Will do tomorrow.

p-linnane (approved):

PR #143981, opened 2023-03-29
orbstack: Link orb and orbctl CLI binaries

kdrag0n:
`orbctl` and the shorter `orb` command are CLI tools for using and managing OrbStack and its Linux machines. They're usually symlinked at first launch, but managing them with Homebrew should be better for cask users. The app will detect brew's symlinks and leave them alone.

OrbStack also includes Docker tools similar to Docker Desktop, but it's probably better to let the app decide what it needs to link at launch. There's no requirement to use any particular versions of the Docker tools, so for example, users with Docker installed via Homebrew may want to keep using their existing ones.

### DIFF
--- a/Casks/orbstack.rb
+++ b/Casks/orbstack.rb
@@ -20,6 +20,8 @@ cask "orbstack" do
   depends_on macos: ">= :monterey"
 
   app "OrbStack.app"
+  binary "#{appdir}/OrbStack.app/Contents/MacOS/bin/orb", target: "orb"
+  binary "#{appdir}/OrbStack.app/Contents/MacOS/bin/orbctl", target: "orbctl"
 
   postflight do
     system_command "#{appdir}/OrbStack.app/Contents/MacOS/bin/orbctl",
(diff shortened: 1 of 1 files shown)


Casks/orbstack.rb
```diff
@@ -20,6 +20,8 @@
   depends_on macos: ">= :monterey"
 
   app "OrbStack.app"
+  binary "#{appdir}/OrbStack.app/Contents/MacOS/bin/orb", target: "orb"
```
bevanjkay, line 23: @kdrag0n I don't believe that `target` is required here, as it will already be set. It's only used when the binary itself is renamed.

bevanjkay (changes requested):

Casks/orbstack.rb
kdrag0n: Removed `target`, thanks.

bevanjkay (dismissed):

bevanjkay (approved):

PR #144269, opened 2023-03-31
prusaslicer 2.5.2

altf4arnold:
Updating to the latest version


SMillerDev (approved):
Thanks @altf4arnold   ! Without contributions like yours it'd be impossible to keep homebrew going with the high standards that users have come to expect from the project. You can feel good knowing that you've made the world a tiny bit better for homebrew users around the world! 👍 🎉

PR #144310, opened 2023-03-31
smartgit 22.1.5

g0t4:
Currently, `v22.1.4` sha256 values don't match up to the hosted files, so currently cannot install/upgrade `smartgit` cask.


reitermarkus (approved):

reitermarkus:
Thanks!

PR #144319, opened 2023-04-01
phew: remove

miccal:
This will fail during `uninstall`, and I do not know why. Adding various `uninstall` stanzas does not seem to help.

May not be worth the trouble:
```
==> Analytics
install: 0 (30 days), 3 (90 days), 26 (365 days)
```


razvanazamfirei:
Last updated in [2017](https://github.com/sveinbjornt/Phew/releases/tag/1.0). Would not be against removing.

p-linnane (approved):

PR #144354, opened 2023-04-01
avast-security 15.6.0

miccal:
May need to be changed to a manual install:

Screenshot 2023-04-02 at 12 01 06


razvanazamfirei:
It gets stuck here. This is cannot be bypassed by sudo, AFAIK.
CleanShot 2023-04-02 at 10 46 55@2x

p-linnane:
Looks like we need to change it to manual install.

razvanazamfirei (approved):

PR #144444, opened 2023-04-03
Add Mullvad Browser v12.0.4 cask

jdorn-eu:

### DIFF
--- a/Casks/mullvad-browser.rb
+++ b/Casks/mullvad-browser.rb
@@ -1,0 +1,30 @@
+cask "mullvad-browser" do
+  version "12.0.4"
+  sha256 "f468e4d875f46f9ed486e2f1225e5c3cd6304a44d3843bf8b20aad5505099b4c"
+
+  url "https://cdn.mullvad.net/browser/#{version}/MullvadBrowser-#{version}-macos_ALL.dmg"
+  name "Mullvad Browser"
+  desc "Firefox fork with anti-fingerprinting features from Tor Browser"
+  homepage "https://mullvad.net/en/browser"
+
+  livecheck do
+    url "https://cdn.mullvad.net/browser/update_responses/update_1/release/downloads.json"
+    strategy :json do |json|
+      json["version"]
+    end
+  end
+
+  auto_updates true
+  depends_on macos: ">= :big_sur"
+
+  app "Mullvad Browser.app"
+
+  uninstall quit: "net.mullvad.mullvadbrowser"
+
+  zap trash: [
+    "~/Library/Application Support/MullvadBrowser",
+    "~/Library/Caches/MullvadBrowser",
+    "~/Library/Preferences/net.mullvad.mullvadbrowser.plist",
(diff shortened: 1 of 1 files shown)


razvanazamfirei (changes requested):

razvanazamfirei (approved):

razvanazamfirei:
@jeifour, this looks great! Thank you for your contribution!

pospeselr:
@yanagibashi-mt do you have an email we can use to notify you of updates and changes potential affecting packaging?

bevanjkay:
@pospeselr Anyone can submit version bumps and change to Homebrew Casks. There are some built in commands to make this easy `brew bump` for example.

razvanazamfirei:
Hi @pospeselr, that's essentially it. We use the feed at `https://cdn.mullvad.net/browser/update_responses/update_1/release/downloads.json` to check for version updates and the file at `https://cdn.mullvad.net/browser/#{version}/MullvadBrowser-#{version}-macos_ALL.dmg` gets downloaded and installed. If they change anyone can submit a PR.

razvanazamfirei:
> We will be updating Mullvad Browser at least once a month along with the upstream Firfox ESR security updates, so the version value will need to be changed each month.
> 
> Is it possible for you to grab the current version and download URL from the downloads.json file directly? If you can do that then the package won't need to change unless we have a watershed update (an update which forces all users to first update to a particular version before progressing to latest; this can happen if some settings migration logic needs to run for example). In that scenario the latest stuff would be under update_2, update_3, etc.

We grab the version from the `downloads.json` file and use it to fill in the version into the download URL. For the most part, we try not to grab download URLs directly.

> If this is using the ordinary `/Applications` installation method, then the browser's built-in updater mechanism should also work fine out of the box. Is there a potential problem of the browser's version getting out-of-sync (eg you install 12.0.4 through homebrew, then the browser auto-updates itself to 12.0.5 while the package manager think its still on 12.0.4).

The auto_updates stanza in the cask addresses this and `brew` will not upgrade the cask unless specifically told to -- this is to prevent interfering with the app's autoupdate mechanism.

> @bevanjkay indeed it is open to a pull request, but as you can imagine a lot of 3rd party/volunteer packages are popping up and we cannot maintain them all or necessarily have the domain knowledge to do so properly. It is much more sustainable for us to send out an email notification to maintainers and let the experts take care of things.

I don't believe we have the expectation that the developers will be maintaining the packages. Applications in Cask tend to be updated reasonably quickly by either the users themselves or the maintainers (as long as there's a relatively stable way to check for new versions). Given the total number of apps and the number of daily PRs, it would be impractical for maintainers to respond to email notifications and it would create an unsustainable precedent. 

You can look at the commit history for both a very popular app https://github.com/Homebrew/homebrew-cask/commits/master/Casks/google-chrome.rb and a niche app https://github.com/Homebrew/homebrew-cask/commits/master/Casks/puzzles.rb to see how the process works. The most helpful thing is having a stable way for brew to determine the latest software version and (ideally) a versioned download link.

PR #144479, opened 2023-04-04
spatial: delete

reitermarkus:
Download is gone/walled.


reitermarkus:
Seems only the Chinese URL is gone.

PR #144505, opened 2023-04-04
gdat: add rosetta caveat

miccal:


Casks/gdat.rb
```diff
@@ -3,7 +3,7 @@
   sha256 "5079b12f9941de7a9631bba08d9e3470e794553f7d5a044969eafa95febda46c"
 
   url "https://drive.google.com/uc?export=download&id=#{version.csv.second}",
-      verified: "drive.google.com/uc?export=download&id=#{version.csv.second}"
+      verified: "drive.google.com/"
```
razvanazamfirei, line 6: I feel like this defeats the purpose of verified urls.

Casks/gdat.rb
```diff
@@ -23,5 +23,6 @@
 
   caveats do
     requires_rosetta
+    license "https://www.getgdat.com/home/download/terms-of-use"
```
razvanazamfirei, line 26: Adding based on strong wording on upstream website. Will ask them to consider moving to a sane download link.

Casks/gdat.rb
reitermarkus: Yeah, anyone can upload anything to Google Drive.

In this case it technically doesn't matter since the ID is also verified by having the livecheck, but this should still be reverted.

razvanazamfirei (approved):

Casks/gdat.rb
reitermarkus: We shouldn't interpolate anything in `verified` URLs if there is no livecheck though.

PR #144623, opened 2023-04-05
nozbe 2023.14 (new cask)

razvanazamfirei:

### DIFF
--- a/Casks/nozbe.rb
+++ b/Casks/nozbe.rb
@@ -9,7 +9,7 @@ cask "nozbe" do
 
   livecheck do
     url "https://nozbe.help/general/release-notes"
-    regex(%r{newest["']>version\s*?(\d+(?:\.\d+)*).*?</}i)
+    regex(%r{id=["']newest["'][^>]*?>\s*version\s*(\d+(?:\.\d+)+)}i)
   end
 
   app "Nozbe.app"
--- a/Casks/nozbe.rb
+++ b/Casks/nozbe.rb
@@ -9,8 +9,10 @@ cask "nozbe" do
 
   livecheck do
     url "https://nozbe.help/general/release-notes"
-    regex(%r{id=["']newest["'][^>]*?>\s*version\s*(\d+(?:\.\d+)+)}i)
+    regex(/id=["']newest["'][^>]*?>\s*version\s*(\d+(?:\.\d+)+)/i)
   end
+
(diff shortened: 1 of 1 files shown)


bevanjkay:
It looks like there is an in-app check for latest version, however it's not up to date https://builds.nozbe.com/updates.mac.json
The https://nozbe.com/download/mac returns a page with the latest version embedded, but livecheck curl can't seem to pick it up.

Casks/nozbe.rb
```diff
@@ -0,0 +1,25 @@
...
+  sha256 "92cd7aa7855567ac3a563c481caf1bf0f795cd9efc96f20516ea77186448c990"
+
+  url "https://builds.nozbe.com/mac/#{version}/Nozbe.app.zip"
+  name "Nozbe"
+  desc "Project management app"
+  homepage "https://nozbe.com/"
+
+  livecheck do
+    url "https://nozbe.help/general/release-notes"
+    regex(%r{newest["']>version\s*?(\d+(?:\.\d+)*).*?</}i)
```
samford, line 12:
```suggestion
    regex(%r{id=["']newest["'][^>]*?>\s*version\s*(\d+(?:\.\d+)+)}i)
```

* We're specifically targeting the `#newest` element, so there's something to be said for being explicit about the `id` attribute (i.e., we don't want to unexpectedly match an element with a different attribute value that happens to end with `newest`).
* There may be additional attributes on this element in the future, so we should account for those (i.e., `[^>]*?>`) to help keep the regex from breaking.
* In this particular case, we don't really benefit from anchoring the end using `.*?</`, so it's probably fine to leave it off for now.

samford (commented):
> The https://nozbe.com/download/mac returns HTML with the latest version embedded, but livecheck curl can't seem to pick it up.

I did some digging and this is because `#curl_args` uses the `--fail` option by default, so `stdout` ends up empty in this scenario (presumably because it's a 404 response?). The HTML ends up in `stdout` if we use `show_output: true` to avoid that flag.

Off hand, I'm not sure if setting that in livecheck's `DEFAULT_CURL_OPTIONS` would notably affect other checks, so I'll have to do some comparative tests before creating a brew PR. It's late over here but I can run tests in the background tomorrow while I work on other things.

That said, checking the release notes seems fine for the time being (after some tweaks to the `regex`).

reitermarkus (approved):

PR #144666, opened 2023-04-06
Add TransIP STACK v2.11.1-20220906

rkrieger:
Although previously removed in 9d975dd, links are available again and do not require login. Livecheck currently narrowed to 2.x.y versions per vendor recommendation.

Checked the basics from contributing instructions (token, both install/uninstall and audit).

### DIFF
--- a/Casks/stack.rb
+++ b/Casks/stack.rb
@@ -9,7 +9,7 @@ cask "stack" do
 
   livecheck do
     url "https://filehosting-client.transip.nl/packages/stack/"
-    regex(%r{href=["'].*?stack/v?((?:2)(?:\.\d+){2}(?:[-_]\d+))["' >]}i)
+    regex(%r{href=["'](?:.*?/)?v?(2(?:[.-]\d+)+)["' >]}i)
   end
 
   app "stack.app"
--- a/Casks/stack.rb
+++ b/Casks/stack.rb
@@ -1,0 +1,25 @@
+cask "stack" do
+  version "2.11.1-20220906"
+  sha256 "d5b92e5f0070fbd1f92c36be58e0c979efcceb79b685dd7e665ebf165d70a82c"
+
+  url "https://filehosting-client.transip.nl/packages/stack/v#{version}/macos/stack-v#{version}.dmg"
+  name "STACK"
+  desc "Personal online hard drive to store, view and share files"
+  homepage "https://www.transip.nl/stack/"
+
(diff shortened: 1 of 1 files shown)


p-linnane:
No concers readding this if CI passes. I'd like @samford to weigh in the on the regex before we merge though.

rkrieger:
> No concerns readding this if CI passes. I'd like @samford to weigh in the on the regex before we merge though.

The current regex is to ensure that it matches vendor-recommend version, rather than v3.2.0-20230307, which is also available at the url. Once that version becomes the recommended one, I'd swap the literal '2' for a '\d+'.

Or is your concern different?

Casks/stack.rb
```diff
@@ -0,0 +1,25 @@
...
+  sha256 "d5b92e5f0070fbd1f92c36be58e0c979efcceb79b685dd7e665ebf165d70a82c"
+
+  url "https://filehosting-client.transip.nl/packages/stack/v#{version}/macos/stack-v#{version}.dmg"
+  name "STACK"
+  desc "Personal online hard drive to store, view and share files"
+  homepage "https://www.transip.nl/stack/"
+
+  livecheck do
+    url "https://filehosting-client.transip.nl/packages/stack/"
+    regex(%r{href=["'].*?stack/v?((?:2)(?:\.\d+){2}(?:[-_]\d+))["' >]}i)
```
samford, line 12:
```suggestion
    regex(%r{href=["'](?:.*?/)?v?(2(?:[.-]\d+)+)["' >]}i)
```

This can be simplified a bit and brought more in line with prevailing regex standards.

* This is a directory listing page dedicated to stack, so we don't need to specifically match the leading `stack/` in the `href` attribute URLs
* Using `(?:2)` in the capture group is no different than `2`
* Restricting versions to three parts (e.g., 1.2.3) makes this more brittle (e.g., it would miss a version with fewer or more version parts), so we should avoid it unless there's a specific reason
* In this case, we can simply use `(?:[.-]\d+)+` to match a format like `1.2.3-20230406` (it will also match `1.2.3` without the date suffix, if a version like that ever appears)

p-linnane:
Thanks @samford!

p-linnane (approved):

Casks/stack.rb
rkrieger: Agree, this makes more sense and looks better. Thanks for clarifying.

PR #144732, opened 2023-04-08
double-commander 1.1.0

rich2014:
Update Double-Commander to 1.1.0.

1. Double Commander for MacOS has regained official support

2. audit / sytle / install / uninstall have been tested

3. both AArch64 and x86_64 versions have beed inclued in official down page:
https://sourceforge.net/p/doublecmd/wiki/Download/

4. offical homepage:
https://doublecmd.sourceforge.io

### DIFF
--- a/Casks/double-commander.rb
+++ b/Casks/double-commander.rb
@@ -8,17 +8,20 @@ cask "double-commander" do
   url "https://downloads.sourceforge.net/doublecmd/doublecmd-#{version.csv.first}-#{version.csv.second}.cocoa.#{arch}.dmg",
       verified: "downloads.sourceforge.net/doublecmd/"
   name "Double Commander"
-  desc "File Manager with double panels"
+  desc "File manager with two panels"
   homepage "https://doublecmd.sourceforge.io/"
 
   livecheck do
-    skip "No version information available"
+    url "https://sourceforge.net/projects/doublecmd/rss"
+    regex(/doublecmd[._-](\d+(?:\.\d+)+)[._-](\d+)\.cocoa/i)
+    strategy :page_match do |page, regex|
+      page.scan(regex).map { |match| "#{match[0]},#{match[1]}" }
+    end
   end
 
   app "Double Commander.app"
 
-  preflight do
-    system_command "xattr",
-                   args: ["-cr", "#{staged_path}/Double Commander.app"]
-  end
+  zap trash: [
+    "~/Library/Caches/doublecmd",
+  ]
 end
--- a/Casks/double-commander.rb
+++ b/Casks/double-commander.rb
@@ -11,10 +11,14 @@ cask "double-commander" do
   desc "File Manager with double panels"
   homepage "https://doublecmd.sourceforge.io/"
 
+  livecheck do
+    skip "No version information available"
+  end
+
   app "Double Commander.app"
 
   preflight do
     system_command "xattr",
-                   args: ["-d", "com.apple.quarantine", "#{staged_path}/Double Commander.app"]
+                   args: ["-cr", "#{staged_path}/Double Commander.app"]
   end
 end
--- a/Casks/double-commander.rb
+++ b/Casks/double-commander.rb
@@ -21,7 +21,5 @@ cask "double-commander" do
 
   app "Double Commander.app"
 
-  zap trash: [
-    "~/Library/Caches/doublecmd",
-  ]
+  zap trash: "~/Library/Caches/doublecmd"
 end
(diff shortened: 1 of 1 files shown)


Casks/double-commander.rb
```diff
@@ -1,16 +1,24 @@
...
+  end
+
   app "Double Commander.app"
 
-  caveats do
-    discontinued
+  preflight do
+    system_command "xattr",
+                   args: ["-cr", "#{staged_path}/Double Commander.app"]
   end
```
bevanjkay, line 23: @rich2014 Why are the attributes being removed from the file?

bevanjkay (dismissed):

Casks/double-commander.rb
rich2014: @bevanjkay

1. Apps downloaded from the Internet will be added with attributes by MacOS.

2. if the app has not been submitted to Apple for review by the Apple Developer ID, when the user opens the app, they will be warned and cannot open it directly.

3. many open source free software have the same problem, such as CudaText:
https://cudatext.github.io/download.html

bevanjkay: @rich2014 we have gone to great lengths to support Apple's quarantine attribute for cask downloads, so we do not support removing of these attributes.

reitermarkus: Homebrew adds these attributes to quarantine the download. This defeats the quarantining.

rich2014: So can we prompt the user some message like this?

```
On macOS Catalina or later, you may get a warning that "the developer cannot be verified". you can:
1. Ctrl+Click and open the APP
2. or disabled in the "Security & Privacy" preferences panel
3. or by running the following command:
    xattr -cr "#{target_dir}"
```

reitermarkus: I don't think we need a message.

Users are supposed to know what they are downloading. This message would need to be displayed for virtually every cask.

You would need to follow the same steps if you install this manually.

Casks/double-commander.rb
```diff
@@ -1,16 +1,24 @@
...
+  url "https://downloads.sourceforge.net/doublecmd/doublecmd-#{version.csv.first}-#{version.csv.second}.cocoa.#{arch}.dmg",
       verified: "downloads.sourceforge.net/doublecmd/"
   name "Double Commander"
-  desc "File manager"
+  desc "File Manager with double panels"
   homepage "https://doublecmd.sourceforge.io/"
 
+  livecheck do
+    skip "No version information available"
+  end
```
razvanazamfirei, line 16:
```suggestion
  livecheck do
    url "https://sourceforge.net/projects/doublecmd/rss"
    regex(/doublecmd[._-](\d+(?:\.\d+)+)[._-](\d+)\./i)
    strategy :page_match do |page, regex|
      page.scan(regex).map { |match| "#{match[0]},#{match[1]}" }
    end
  end
```

Casks/double-commander.rb
```diff
@@ -1,16 +1,24 @@
...
-  url "https://downloads.sourceforge.net/doublecmd/doublecmd-#{version}.qt.x86_64.dmg",
+  version "1.1.0,10668"
+  sha256 arm:   "4727e9a4a75cb6b455d010cbd01bf8b82f1a5557a996774a98ababae3c902d34",
+         intel: "22196b11a5943ad833d5adc6df34deb3a98ea1d1d80138be34bed2ce5bda2300"
+
+  url "https://downloads.sourceforge.net/doublecmd/doublecmd-#{version.csv.first}-#{version.csv.second}.cocoa.#{arch}.dmg",
       verified: "downloads.sourceforge.net/doublecmd/"
   name "Double Commander"
-  desc "File manager"
+  desc "File Manager with double panels"
```
razvanazamfirei, line 11:
```suggestion
  desc "File manager with double panels"
```

Casks/double-commander.rb
```diff
@@ -1,16 +1,24 @@
...
+
   app "Double Commander.app"
 
-  caveats do
-    discontinued
+  preflight do
+    system_command "xattr",
+                   args: ["-cr", "#{staged_path}/Double Commander.app"]
   end
 end
```
razvanazamfirei, line 24: Please include a [zap stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap).

razvanazamfirei (dismissed):

Casks/double-commander.rb
razvanazamfirei: I agree! I think the entire preflight stanza should be removed.

rich2014:
@reitermarkus @razvanazamfirei 

thx.

issue fixed, check it please.

reitermarkus (approved):

reitermarkus:
Thanks, @rich2014!

Issue #143596, opened 2023-03-22
google-cloud-sdk zsh autocompletion doesn't load automatically


hayatofujii:
### Verification

- [X] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [ ] I have retried my command with `--force`. -> not really applicable
- [X] I ran `brew update-reset && brew update` and retried my command.
- [X] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [X] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [X] I made doubly sure this is not a [checksum does not match](https://docs.brew.sh/Common-Issues#cask---checksum-does-not-match) error.

### Description of issue

For some reason, `google-cloud-sdk` zsh autocompletion is linked to `$(brew --prefix)/share/zsh/site-functions`, as shown in `brew info google-cloud-sdk`:

```
==> Artifacts
[snip for brevity]
google-cloud-sdk/completion.zsh.inc -> /opt/homebrew/share/zsh/site-functions/_google-cloud-sdk (Binary)
[snip for brevity]
```

Following the [instructions about shell completion](https://docs.brew.sh/Shell-Completion), logging out and relogin doesn't autocomplete `gcloud` commands.

I'm fully aware of the caveat, and, by doing that, makes gcloud autocompletion work (thou I prefer sourcing from `$(brew --prefix)/share/zsh/site-functions`):

```
==> Caveats
To add gcloud components to your PATH, add this to your profile:
[snip]
  for zsh users
    source "/opt/homebrew/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/path.zsh.inc"
[snip]
```

In conclusion, what's the point of the `google-cloud-sdk/completion.zsh.inc` artifact?

### Command that failed

`gcloud <Tab>`

### Output of command with `--verbose --debug`

```shell
not applicable
```

### Output of `brew doctor` and `brew config`

```shell
$ brew doctor
Your system is ready to brew.

$ brew config
HOMEBREW_VERSION: 4.0.9-35-g65d858d
ORIGIN: https://github.com/Homebrew/brew
HEAD: 65d858da123dda54784a7b374605b6eb1a699882
Last commit: 4 hours ago
Core tap origin: https://github.com/Homebrew/homebrew-core
[...]

PR #144834, opened 2023-04-10
google-cloud-sdk 425.0.0 (caveat)

razvanazamfirei:

See #143596 and https://github.com/Homebrew/homebrew-cask/issues/143596#issuecomment-1502136855
This does not apply to bash or fish users. 

For brevity, I'm adding the completion file to the existing caveat as opposed to adding a separate completion-specific section.


hayatofujii:
I think that the binary stanza on line 49 (the `_google_cloud_sdk` symlink) can be dropped, since it doesn't work.

p-linnane (approved):

PR #144970, opened 2023-04-12
mucommander 1.2.0-2

ahadas:
Update mucommander from 1.2.0-1 to 1.2.0-2

### DIFF
--- a/Casks/mucommander.rb
+++ b/Casks/mucommander.rb
@@ -1,8 +1,8 @@
 cask "mucommander" do
   version "1.2.0-1"
-  sha256 "c77c3d0105de57dcb2e2641a23266aef576b9874a61096935d961427eece636c"
+  sha256 "d5e386d62b01ed7614b15a3fb807b21f8c3e650ddbd2b5b8d08c6949ffcd779d"
 
-  url "https://github.com/mucommander/mucommander/releases/download/#{version}/muCommander-#{version}.dmg",
+  url "https://github.com/mucommander/mucommander/releases/download/#{version}/muCommander-1.2.0-2.dmg",
       verified: "github.com/mucommander/mucommander/"
   name "muCommander"
   desc "File manager with a dual-pane interface"
--- a/Casks/mucommander.rb
+++ b/Casks/mucommander.rb
@@ -8,5 +8,15 @@ cask "mucommander" do
   desc "File manager with a dual-pane interface"
   homepage "https://www.mucommander.com/"
 
+  livecheck do
+    url :url
+    regex(%r{href=.*?/tag/v?(\d+(?:[.-]\d+)+)["' >]}i)
+    strategy :github_latest
+  end
+
   app "muCommander.app"
+
+  zap trash: [
+    "~/Library/Preferences/muCommander",
+  ]
 end
--- a/Casks/mucommander.rb
+++ b/Casks/mucommander.rb
@@ -1,18 +1,12 @@
 cask "mucommander" do
-  version "1.2.0-1"
+  version "1.2.0-2"
   sha256 "d5e386d62b01ed7614b15a3fb807b21f8c3e650ddbd2b5b8d08c6949ffcd779d"
 
-  url "https://github.com/mucommander/mucommander/releases/download/#{version}/muCommander-1.2.0-2.dmg",
+  url "https://github.com/mucommander/mucommander/releases/download/#{version}/muCommander-#{version}.dmg",
       verified: "github.com/mucommander/mucommander/"
   name "muCommander"
   desc "File manager with a dual-pane interface"
   homepage "https://www.mucommander.com/"
-
-  livecheck do
-    url :url
-    regex(%r{href=.*?/tag/v?(\d+(?:[.-]\d+)+)["' >]}i)
-    strategy :github_latest
-  end
 
   app "muCommander.app"
 


razvanazamfirei:
Thank you for your submission to homebrew-cask, @ahadas. Please try to add a [zap stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap) for this cask.

ahadas:
@razvanazamfirei sure, posted #144978 
about the failing job in this PR - is it a result of leaving mucommander-1.2.0-1.dmg under the same 'release' in github?

ahadas:
filed #144979 instead so someone will pick it up or remove the cask, thanks!

ahadas:
@razvanazamfirei sorry, I didn't notice your changes before closing

razvanazamfirei:
Hi @ahadas, no worries. Yes, would you be able to move the 1.2.0-2 version into a 1.2.0-2 release? Otherwise it will be very difficult to parse the url. Once that's done, my latest commit should pass CI.

ahadas:
> Hi @ahadas, no worries. Yes, would you be able to move the 1.2.0-2 version into a 1.2.0-2 release? Otherwise it will be very difficult to parse the url. Once that's done, my latest commit should pass CI.

That's problematic because it's not really a new release, just another build of the dmg of 1.2.0 with a fix for https://github.com/mucommander/mucommander/issues/898 - I prefer not to introduce another release because then I'll need to create also the other artifacts (msi, rpm, deb, ...). I've removed the 1.2.0-1 dmg, isn't it enough?

razvanazamfirei:
We don't control the the versioning policies of upstream repos, but for this cask to be updated we will need a url that does not contain mismatched (`1.2.0-1` and `1.2.0-2`) versions. Currently, the download URL is `https://github.com/mucommander/mucommander/releases/download/1.2.0-1/muCommander-1.2.0-2.dmg`.  If you're able to either release `1.2.0-2` or change the filename to `muCommander-1.2.0-1.dmg` we'd be happy to take a look.

ahadas:
@razvanazamfirei I see, so I don't know what's our options here - either to disable the cask if possible or to remove it and add it back when the next version is released..

razvanazamfirei:
Waiting for @samford to take a look at this!

reitermarkus:
> just another build of the dmg of 1.2.0 with a fix

So this is includes a fix which isn't actually included in the `1.2.0-1` tag? If that's the case, probably best to remove it form that release entirely.

> because then I'll need to create also the other artifacts

Are they not created automatically via CI?

ahadas:
> > because then I'll need to create also the other artifacts
> 
> Are they not created automatically via CI?

not all of them, and since other artifacts besides the dmg didn't change, I don't want users that use these artifacts to be notified about having a new release that they should upgrade to in order to use the latest version

reitermarkus:
> I don't want users that use these artifacts to be notified about the availability of a new release that they should upgrade to in order to use the latest version

Not every upgrade will fix something for every user, so I don't think it is a good approach to optimize towards the worst case.

If the release notes only state a single fix for macOS, users can decide for themselves if the want to update or wait for the next release.

ahadas:
> > I don't want users that use these artifacts to be notified about the availability of a new release that they should upgrade to in order to use the latest version
> 
> Not every upgrade will fix something for every user, so I don't think it is a good approach to optimize towards the worst case.
> 
> If the release notes only state a single fix for macOS, users can decide for themselves if the want to update or wait for the next release.

imagine yourself using muCommander on Windows, ok?
you get this notification that there's a newer version available, which points you to github and let's say that you see the note that is suggested above and realize that you don't need this release, what then?
you have two options:
1. to keep the notifications for the availability of a new version enabled - and then for some time you keep checking but if you don't see that another version is relevant for you, you either stop checking these notifications or disable these notifications, which takes us to the next option..
2. disable the notifications for the availability of a new version - and then you can't tell when a new version is available without checking the website/twitter

Guys, really, I appreciate the feedback on this but if there's an issue in homebrew to support an artifact that includes a different build version in its name then no need to change how upgrades are done in mucommander or do major changes in homebrew - it's really fine to drop the cask from my side

reitermarkus:
I see your point, but a `1.2.0-1` tag containing a `1.2.0-2` asset is confusing.

> if there's an issue in homebrew to support an artifact that includes a different build version in its name

We can certainly work around this, but it just makes it harder than it needs to be for us to maintain. Note that this is not the first release naming scheme we encountered that is hard to maintain, we just wanted to let you know since you are a maintainer of the project.

The main problem is for the current version we would have to use `version "1.2.0-1,1.2.0-2"`, change the interpolation in the `url` and a use very workaround-y `livecheck`. And then we'd likely have to revert all this again for the next release.

So, I would say the options here are (from most to least preferable):

1. Create a new release for `1.2.0-2`.
2. If you think this could happen again in the future, create tags without the build number, e.g. `1.2.1`, so that we can at least simplify the version to only use e.g. `1.2.1-1`. (Note that this would still require us to have a custom `livecheck`.)
3. Keep everything as is.

ahadas:
> 2. If you think this could happen again in the future, create tags without the build number, e.g. `1.2.1`, so that we can at least simplify the version to only use e.g. `1.2.1-1`. (Note that this would still require us to have a custom `livecheck`.)

yeah, it's been a while since I published different builds for the same versions (back in 0.9.3 and 0.9.4) and back then this approach also crossed my mind but since it didn't happen since then, I didn't take into account the possibility of publishing another build for 1.2.0, otherwise I would have named it 1.2.0 rather than 1.2.0-1. I think that's the way to go with the current versioning in muCommander

razvanazamfirei:
Hey @ahadas, happy to take another look at this once there's more clarity about the versioning plan! Hopefully that will resolve #144979 more sustainably. Thanks for submitting and working with us!

PR #145010, opened 2023-04-13
cask-drivers: migrate casks to homebrew/cask

razvanazamfirei:
Merge at the same time as: https://github.com/Homebrew/homebrew-cask-drivers/pull/3416


bevanjkay:
We'll need to add `skip` labels to bypass the signing checks if we want to keep some of these.

razvanazamfirei:
```
❯ codesign -vv ~/Downloads/LogiMgr\ Installer\ 10.00.75.app/
/Users/johnqpublic/Downloads/LogiMgr Installer 10.00.75.app/: valid on disk
/Users/johnqpublic/Downloads/LogiMgr Installer 10.00.75.app/: satisfies its Designated Requirement
```

bevanjkay (approved):

PR #145223, opened 2023-04-16
reminders-menubar 1.17.0 (new cask)

DamascenoRafael:
Adding [Reminders MenuBar](https://github.com/DamascenoRafael/reminders-menubar) for the first time in homebrew-cask

The cask was available in a personal tap: [DamascenoRafael/homebrew-tap](https://github.com/DamascenoRafael/homebrew-tap)
Here is the PR removing from the previous tap and adding the `tap_migrations.json`:
https://github.com/DamascenoRafael/homebrew-tap/pull/2

### DIFF
--- a/Casks/reminders-menubar.rb
+++ b/Casks/reminders-menubar.rb
@@ -18,4 +18,11 @@ cask "reminders-menubar" do
       system_command "/usr/bin/open", args: ["-a", application], must_succeed: false
     end
   end
+
+  zap trash: [
+    "~/Library/Application Scripts/br.com.damascenorafael.reminders-menubar",
+    "~/Library/Application Scripts/br.com.damascenorafael.RemindersLauncher",
+    "~/Library/Containers/br.com.damascenorafael.reminders-menubar",
+    "~/Library/Containers/br.com.damascenorafael.RemindersLauncher",
+  ]
 end
--- a/Casks/reminders-menubar.rb
+++ b/Casks/reminders-menubar.rb
@@ -1,0 +1,21 @@
+cask "reminders-menubar" do
+  version "1.17.0"
+  sha256 "e3fdf46b3fc339bde71a14296f544c74ca099058d3845e86aca0d2fdcf8557fe"
+
+  url "https://github.com/DamascenoRafael/reminders-menubar/releases/download/v#{version}/reminders-menubar.zip"
+  name "Reminders MenuBar"
+  desc "Simple menu bar app to view and interact with reminders"
+  homepage "https://github.com/DamascenoRafael/reminders-menubar"
+
+  depends_on macos: ">= :big_sur"
+
+  app "Reminders MenuBar.app"
+
+  postflight do
+    application = "#{appdir}/Reminders MenuBar.app"
+    if system_command("ps", args: ["x"]).stdout.match?(application)
+      system_command "/usr/bin/pkill", args: ["-f", application], must_succeed: false
+      system_command "/usr/bin/open", args: ["-a", application], must_succeed: false
+    end
+  end
+end
--- a/Casks/reminders-menubar.rb
+++ b/Casks/reminders-menubar.rb
@@ -12,7 +12,7 @@ cask "reminders-menubar" do
   app "Reminders MenuBar.app"
 
   postflight do
-    application = "#{appdir}/Reminders MenuBar.app"
+    application = "#{staged_path}/#{token}"
     if system_command("ps", args: ["x"]).stdout.match?(application)
       system_command "/usr/bin/pkill", args: ["-f", application], must_succeed: false
       system_command "/usr/bin/open", args: ["-a", application], must_succeed: false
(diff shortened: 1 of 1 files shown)


razvanazamfirei:
Thank you for your submission to homebrew-cask! Please try to add a [zap stanza](https://docs.brew.sh/Cask-Cookbook#stanza-zap) for this cask.

DamascenoRafael:
@razvanazamfirei I added the zap stanza according to nrlquaker’s CreateZap and AppCleaner.

Casks/reminders-menubar.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  app "Reminders MenuBar.app"
+
+  postflight do
+    application = "#{appdir}/Reminders MenuBar.app"
+    if system_command("ps", args: ["x"]).stdout.match?(application)
+      system_command "/usr/bin/pkill", args: ["-f", application], must_succeed: false
+      system_command "/usr/bin/open", args: ["-a", application], must_succeed: false
+    end
+  end
```
razvanazamfirei, line 20: Is this necessary for installing the cask?

Casks/reminders-menubar.rb
```diff
@@ -0,0 +1,28 @@
...
+  name "Reminders MenuBar"
+  desc "Simple menu bar app to view and interact with reminders"
+  homepage "https://github.com/DamascenoRafael/reminders-menubar"
+
+  depends_on macos: ">= :big_sur"
+
+  app "Reminders MenuBar.app"
+
+  postflight do
+    application = "#{appdir}/Reminders MenuBar.app"
```
razvanazamfirei, line 15: Not all users install the apps in #{appdir}, this should cover more use cases. 
```suggestion
    application = "#{staged_path}/#{token}"
```

razvanazamfirei (dismissed):
Hi @DamascenoRafael, this looks great! I have one question below.

Casks/reminders-menubar.rb
DamascenoRafael: @razvanazamfirei It's not exactly necessary. But it seems useful when updating the app.

As the App is in the menu bar, it is likely to be running when the user updates it.
Without this script, after an update, the user will continue using the old version, and will only start using the new version after manually closing and reopening the app.
The postflight first checks if the application is already running, and **only if it is running** it will close the app and reopen it. If the app is not running, nothing happens.

Does this sound ok for an cask?

Casks/reminders-menubar.rb
DamascenoRafael: Thanks, I hadn't thought about the possibility of installations outside #{appdir}.

DamascenoRafael:
when using `application = "#{staged_path}/#{token}"` the script doesn't work for me.
The result of `application` is `/usr/local/Caskroom/reminders-menubar/1.17.0/reminders-menubar`.

Casks/reminders-menubar.rb
```diff
@@ -0,0 +1,22 @@
...
+  url "https://github.com/DamascenoRafael/reminders-menubar/releases/download/v#{version}/reminders-menubar.zip"
+  name "Reminders MenuBar"
+  desc "Simple menu bar app to view and interact with reminders"
+  homepage "https://github.com/DamascenoRafael/reminders-menubar"
+
+  depends_on macos: ">= :big_sur"
+
+  app "Reminders MenuBar.app"
+
+  uninstall quit: "br.com.damascenorafael.reminders-menubar"
```
razvanazamfirei, line 14: What do you think about this? Should achieve the same thing as it would be called during upgrade.

DamascenoRafael: It will not be exactly the same because if the app is already open this will not reopen the app, just close it. So the user would still have to manually reopen the app.

However, if the previous script is something that should be avoided I'm ok with going with that option.

Do you think "reopen after update" could be a possible function in homebrew in the future?

reitermarkus: Personally I find `uninstall quit:` very annoying to the point that I cannot `brew upgrade` while I'm doing anything because it might just quit the program I'm currently using.

I feel like `uninstall quit:` should only be called on `uninstall`, not on `upgrade`.

But I guess it depends on the application. In this case it's a menubar application, which probably isn't going to be used “actively”.

razvanazamfirei: I agree that calling `uninstall` stanzas during `brew upgrade` is at the very least irritating and counterintuitive. 

In terms of this cask, I like the `postflight` solution. It's clean and unobtrusive, but since this approach is uncommon I'll defer to @reitermarkus and the other maintainers.

reitermarkus: I think for this let's leave it with `uninstall quit:`. If we want to improve this, we should have a proper solution and discuss further in an issue on what that solution should look like.

razvanazamfirei (approved):

DamascenoRafael:
@razvanazamfirei thanks for approving PR!
I'll take a deeper look at brew's implementations and will open an issue later about the postflight feature.

PR #145600, opened 2023-04-21
CI: add prerelease exceptions to ci-skip-repository

razvanazamfirei:
`ci-skip-repository` does not skip all repository checks. There is currently no way to skip checks for pre-release versions without using `ci-syntax-only`. This is an issue when files are served from github, but the labeling of releases does not reflect the version info provided on the upstream website (i.e. stable versions are marked as pre-release on github). Adding the cask to the prerelease allow-list does not work when the github latest version matches the version provided upstream.

### DIFF
--- a/cmd/lib/ci_matrix.rb
+++ b/cmd/lib/ci_matrix.rb
@@ -141,9 +141,9 @@ module CiMatrix
       audit_exceptions << "livecheck_min_os" if labels.include?("ci-skip-livecheck-min-os")
 
       if labels.include?("ci-skip-repository")
-        audit_exceptions << ["github_repository", "github_prerelease_version",
-                             "gitlab_repository", "gitlab_prerelease_version",
-                             "bitbucket_repository"]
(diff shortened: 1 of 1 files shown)


razvanazamfirei:
That's very reasonable and I agree. I still think a change along those lines would be useful -- I found it very counterintuitive that skipping repository checks did not skip a repository-related check. I'm happy to hear if there are any downsides to this I haven't considered!

SMillerDev (approved):

reitermarkus:
The reason why it fails when it's not a pre-release is so we don't forget to remove it from that list once it becomes stable.

razvanazamfirei:
That's fair and makes sense if versions are consistently marked as pre-release! For casks like https://github.com/standardnotes/app/releases, where prereleases and releases are interspersed, the functionality breaks. If the former is more common among the casks on the list it makes sense.

reitermarkus:
For standardnotes we should just not update it to a pre-release. But I think the website links to the pre-release, which is the problem in this case, right?

PR #145388, opened 2023-04-18
mysqlworkbench 8.0.33

TsanBoi:
Created with `brew bump-cask-pr`.

PR #145722, opened 2023-04-24
mysqlworkbench 8.0.33 (rework)

razvanazamfirei:

[Latest updated dropped support for macOS <= 12. 
](https://github.com/Homebrew/homebrew-cask/pull/145388#issuecomment-1519586679)

### DIFF
--- a/Casks/mysqlworkbench.rb
+++ b/Casks/mysqlworkbench.rb
@@ -1,4 +1,6 @@
 cask "mysqlworkbench" do
+  arch arm: "arm64", intel: "x86_64"
+
   on_sierra :or_older do
     version "6.3.10"
     sha256 "29857bf84bebb7c4442ce147e44602d00f8c001e3c09b3a6e3af356767e08d2c"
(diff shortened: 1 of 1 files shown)


Vulxan:
https://bugs.mysql.com/bug.php?id=110726
According to the reply from the MySQL team, it can be confirmed that version 8.0.33 requires macOS 13 or later, although the official website download page has not updated the system version requirements.

sebastian2209:
Can you go back to version 8.0.31? There is a bug in 8.0.32 when trying to run a select.
https://bugs.mysql.com/bug.php?id=109671
https://bugs.mysql.com/bug.php?id=109686
https://bugs.mysql.com/bug.php?id=109729

razvanazamfirei:
Hi @sebastian2209! You can use a [previous version](https://github.com/Homebrew/homebrew-cask/commit/208c89f72974f469e5d8235490c03db138d7ce01) of the cask or download `8.0.31` from the developers. 

We don't revert versions due to bugs in the upstream software, we just try to offer the latest compatible version of the software. If a newer version exists that fixes the bug introduced in `8.0.32`, we're happy to include that.

miccal (approved):

PR #145908, opened 2023-04-26
captain (add livecheck)

krehel:
* Add livecheck stanza

### DIFF
--- a/Casks/captain.rb
+++ b/Casks/captain.rb
@@ -11,6 +11,11 @@ cask "captain" do
   desc "Manage Docker containers from the menu bar"
   homepage "https://getcaptain.co/"
 
+  livecheck do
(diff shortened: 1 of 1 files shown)


razvanazamfirei:
Hi @krehel, thank you for submitting this! Is there a reason for using the `github_latest` strategy? We [avoid using it](https://docs.brew.sh/Brew-Livecheck#general-guidelines) unless the “latest” release is different than the newest version inferred from the tags.

If the cask uses a github `url`, there's an implied `git` strategy and no need to specify a separate `livecheck`.

krehel:
@razvanazamfirei Thanks for the explanation.  Did not know that, and overlooked that part in the doc.  Noted.  

We can close this then.

PR #146035, opened 2023-04-27
sfdx (fix sha256 no_check, update zap)

krehel:
* Revert back to sha256 no_check due to unversioned URL

* Add path to zap


razvanazamfirei:
Thanks for all the great work @krehel! 😄

razvanazamfirei (approved):

PR #145132, opened 2023-04-14
Migrate casks from homebrew/cask-drivers.

reitermarkus:
Merge with https://github.com/Homebrew/homebrew-cask-drivers/pull/3421.

PR #146071, opened 2023-04-28
Add Red Canary Mac Monitor.app v1.0.3

Brandon7CC:

_In the following questions `red-canary-mac-monitor` is the token of the cask you're submitting._

  - **Red Canary Mac Monitor does have the platform in the description because it's a "macOS threat research tool"** 


Casks/red-canary-mac-monitor.rb
```diff
@@ -0,0 +1,31 @@
+cask "red-canary-mac-monitor" do
+  version "1.0.3"
+  sha256 "41a0265630b63d66af65b59f11c8ff11daffd002849e068bcf25b83223a1ef29"
+
+  url "https://github.com/redcanaryco/mac-monitor/releases/download/v#{version}/Red-Canary-Mac-Monitor-GoldCardinal-#{version.tr(".", "-")}.pkg",
+      verified: "github.com/redcanaryco/mac-monitor/"
+  name "Red Canary Mac Monitor"
+  name "red-canary-mac-monitor"
+  name "rc-mac-monitor"
+  desc "Dynamic analysis tool for macOS security research and malware triage"
```
bevanjkay, line 10:
```suggestion
  desc "Analysis tool for security research and malware triage"
```

`homebrew-cask` only supports MacOS so including the platform in the description is redundant.

bevanjkay (changes requested):

Casks/red-canary-mac-monitor.rb
```diff
@@ -0,0 +1,31 @@
...
+  homepage "https://redcanary.com/mac-threat-analysis-tool/"
+
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
```
bevanjkay, line 16: @Brandon7CC The `:github_latest` livecheck strategy should only be used where the default `:git` strategy cannot be used due to the use of pre-releases. If the default `:git` strategy can be used here, the entire `livecheck` block can omitted as it will be inferred automatically from the `url`.

bevanjkay:
Also from the test failure -
```
Some packages are still installed, add them to
uninstall pkgutil: com.redcanary.agent
```

Casks/red-canary-mac-monitor.rb
Brandon7CC: @bevanjkay thank you! I went ahead and removed this.

Casks/red-canary-mac-monitor.rb
Brandon7CC: Thank you :)

Casks/red-canary-mac-monitor.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  pkg "Red-Canary-Mac-Monitor-GoldCardinal-#{version.tr(".", "-")}.pkg"
+
+  uninstall script: [
+    executable: "#{appdir}/Red Canary Mac Monitor.app/Contents/SharedSupport/uninstall.sh",
+    sudo:       false,
+  ]
+  uninstall pkgutil: [
+    "com.redcanary.agent",
+  ]
```
bevanjkay, line 24: @Brandon7CC These should be combined in to one `uninstall` stanza.

```suggestion
  uninstall pkgutil: "com.redcanary.agent",
                 script:   {
                   executable: "#{appdir}/Red Canary Mac Monitor.app/Contents/SharedSupport/uninstall.sh",
                   sudo:            false,
                 }
```

This may fail style.

Brandon7CC:
Please note that we're using `/usr/bin/osascript` to uninstall the app. We're doing this because the app hosts a System Extension which requires user authentication to approve. By "moving the application bundle" to the trash with AppleScript the user is able to authenticate and remove the System Extension.

razvanazamfirei:
@Brandon7CC, thank you for submitting this! 
This is not directly relevant to the cask, but issues relating to managing system extensions/ `systemextensionsctl uninstall` not working with SIP came up before. 

We've briefly talked about using `/usr/bin/osascript` as a potential strategy, but I figured I would check whether or not this has been a reliable uninstall strategy for you and if you had any issues dealing with orphaned extensions.

Brandon7CC:
@razvanazamfirei Thank you for asking about this! The OSA method works well -- given that the user accepts the uninstall request. If they don't macOS blocks the operation and the System Extension (SE) will be left installed unfortunately. Using `systemextensionsctl` like you said isn't helpful with SIP enabled, but the other option would be to use the `SystemExtensions` API directly with a "privileged helper binary" -- the user can again deny this operation though and the SE will be left installed. I found the OSA implementation to be the most "elegant" to help with this research tool case.

Casks/red-canary-mac-monitor.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  uninstall pkgutil: "com.redcanary.agent",
+            script:  {
+              executable: "#{appdir}/Red Canary Mac Monitor.app/Contents/SharedSupport/uninstall.sh",
+              sudo:       false,
+            }
+
+  zap trash: [
+    "~/Library/Application Support/Red Canary Mac Monitor",
+  ]
```
razvanazamfirei, line 26:
```suggestion
  zap trash: [
    "~/Library/Application Support/Red Canary Mac Monitor",
    "~/Library/Preferences/com.redcanary.agent.plist",
    "~/Library/Saved Application State/com.redcanary.agent.savedState",
  ]
```

Casks/red-canary-mac-monitor.rb
```diff
@@ -0,0 +1,27 @@
+cask "red-canary-mac-monitor" do
+  version "1.0.3"
+  sha256 "41a0265630b63d66af65b59f11c8ff11daffd002849e068bcf25b83223a1ef29"
+
+  url "https://github.com/redcanaryco/mac-monitor/releases/download/v#{version}/Red-Canary-Mac-Monitor-GoldCardinal-#{version.tr(".", "-")}.pkg",
+      verified: "github.com/redcanaryco/mac-monitor/"
+  name "Red Canary Mac Monitor"
+  name "red-canary-mac-monitor"
+  name "rc-mac-monitor"
```
razvanazamfirei, line 9: I'm not sure this is necessary unless these are commonly-used names for the app.

razvanazamfirei (changes requested):
Added a few more things to the zap stanza and a comment. Besides this, I think it looks good!

Note: Locally, it uninstalls cleanly, but does require additional user input. It will always time-out during CI.

razvanazamfirei:
Thank you, @Brandon7CC! This is very helpful.

Casks/red-canary-mac-monitor.rb
Brandon7CC: 👍  Took these out

Brandon7CC:
One more question... when we submit PRs for updates and the CI fails due to user interaction should we specify that at the next update?

Casks/red-canary-mac-monitor.rb
```diff
@@ -0,0 +1,27 @@
...
+      verified: "github.com/redcanaryco/mac-monitor/"
+  name "Red Canary Mac Monitor"
+  desc "Analysis tool for security research and malware triage"
+  homepage "https://redcanary.com/mac-threat-analysis-tool/"
+
+  auto_updates true
+  depends_on macos: ">= :ventura"
+
+  pkg "Red-Canary-Mac-Monitor-GoldCardinal-#{version.tr(".", "-")}.pkg"
+
```
razvanazamfirei, line 15:
```suggestion

  # The uninstall script requires user input to remove an installed system
  # extension. It is expected that the uninstall CI will time out and fail.
```

razvanazamfirei:
Adding the comment in the cask should streamline version bumps. It will fail initially, but hopefully pass once the install test is skipped. We don't want to do this by default in case install problems show up.

razvanazamfirei (dismissed):

razvanazamfirei:
@Brandon7CC, since we don't have write access to the fork, can you rebase this? Thank you

Brandon7CC:
@razvanazamfirei should be good to go! Let me know if there's anything else funky going on

razvanazamfirei (approved):

bevanjkay:
Is there no way for this to pass CI without an exception. Ideally we shouldn't be using exceptions except in very rare circumstances.

Brandon7CC:
@bevanjkay  the app hosts a System Extension (https://developer.apple.com/system-extensions/) to communicate with the Endpoint Security subsystem. System Extensions are the replacement for Kexts and are protected by SIP (and in turn from the `root` user — “rootless”). To remove a System Extension theres a few different options (the ones relevant here are):

- User interaction and authentication deleting the app from the Finder (i.e. in the OSA case here with: `/usr/bin/osascript`).
- Using the System Extensions API directly which also requires user authentication and must be done from a “privileged app/binary”.

I’ve gone ahead and implemented the first option here and have found it to be the most elegant / helpful in this context. There are **no** unprivileged / entitled binaries on macOS to “remove a System Extension” without SIP disabled — to be expected as this is a security relevant operation. The CI job is failing due to the AppleEvent timeout waiting for the user to accept the System Extension uninstall request. Please let me know if I can provide any additional details!


System Extension Removal Methods

razvanazamfirei:
We had a similar discussion in #145132 with respect to `ftdi-vcp-driver`. It is possible to have this pass CI at the expense of orphaning the system extension which would then require disabling SIP to remove. Turning this into a method is on my to-do list as it seems to be relevant to other casks as well.

bevanjkay:
@Brandon7CC My concern is that `homebrew-cask` users expect not to be presented with GUI inputs, and can be used successfully using a remote shell.
We have tested adding a `must_succeed: false` argument in #146256 which allows CI to pass, with the downside that Homebrew will consider the uninstall complete even if the script fails.

Without further digging as for the reason, the script did error out for me when I tested the install/uninstall process.

<details><summary>Error</summary>
<p>
```
29:125: execution error: Finder got an error: The operation can't be completed right now because another operation is in progress, such as moving or copying an item or emptying the Bin. (-8065)
Error: Failure while executing; `/usr/bin/env /Applications/Red\ Canary\ Mac\ Monitor.app/Contents/SharedSupport/uninstall.sh` exited with 1. Here's the output:
29:125: execution error: Finder got an error: The operation can't be completed right now because another operation is in progress, such as moving or copying an item or emptying the Bin. (-8065)
```
</p>
</details>

PR #146129, opened 2023-04-28
various: add zaps

krehel:
Add zaps

* 010-editor
* sapmachine-jdk
* warsow
* zeronet

Update zap

* arduino-ide
* clone-hero
* whatsize

### DIFF
--- a/Casks/010-editor.rb
+++ b/Casks/010-editor.rb
@@ -13,4 +13,11 @@ cask "010-editor" do
   end
 
   app "010 Editor.app"
+
+  zap trash: [
+    "~/.config/SweetScape",
+    "~/Documents/SweetScape",
+    "~/Library/Application Support/SweetScape",
+    "~/Library/Saved Application State/com.SweetScape.010Editor.savedState",
+  ]
 end
--- a/Casks/010-editor.rb
+++ b/Casks/010-editor.rb
@@ -15,9 +15,9 @@ cask "010-editor" do
   app "010 Editor.app"
 
   zap trash: [
-    "~/.config/SweetScape",
-    "~/Documents/SweetScape",
-    "~/Library/Application Support/SweetScape",
-    "~/Library/Saved Application State/com.SweetScape.010Editor.savedState",
-  ]
+        "~/.config/SweetScape",
+        "~/Library/Application Support/SweetScape",
+        "~/Library/Saved Application State/com.SweetScape.010Editor.savedState",
+      ],
+      rmdir: "~/Documents/SweetScape"
 end
--- a/Casks/arduino-ide.rb
+++ b/Casks/arduino-ide.rb
@@ -22,5 +22,6 @@ cask "arduino-ide" do
   zap trash: [
     "~/.arduinoIDE",
     "~/Library/Application Support/arduino-ide",
+    "~/Library/Saved Application State/cc.arduino.IDE2.savedState",
   ]
 end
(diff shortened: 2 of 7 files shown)


krehel:
Hi team - updated multiple zaps.  Let me know if the preference is to break these into their own pulls and I'll resubmit separately.

Thanks

Casks/whatsize.rb
```diff
@@ -20,13 +20,13 @@
             launchctl: "com.id-design.v#{version.major}.whatsizehelper"
 
   zap trash: [
-    "/Library/LaunchDaemons/com.id-design.v7.whatsizehelper.plist",
-    "/Library/PrivilegedHelperTools/com.id-design.v7.whatsizehelper",
+    "/Library/LaunchDaemons/com.id-design.v#{version.major}.whatsizehelper.plist",
+    "/Library/PrivilegedHelperTools/com.id-design.v#{version.major}.whatsizehelper",
```
razvanazamfirei, line 24: Are these files actually getting deleted? Some of them may need to be added to `zap delete:`

Casks/arduino-ide.rb
```diff
@@ -22,5 +22,6 @@
   zap trash: [
     "~/.arduinoIDE",
     "~/Library/Application Support/arduino-ide",
+    "~/Library/Saved Application State/cc.arduino.IDE2.savedState",
```
razvanazamfirei, line 25:
```suggestion
    "~/Library/Saved Application State/cc.arduino.IDE#{version.major}.savedState",
```

razvanazamfirei (commented):
Thanks for submitting this and all the hard work! 😄  We generally don't encourage submitting multiple casks in a PR, but if the number is kept reasonably low and the casks are not problematic, it should be fine. If there's a possibility that CI would fail, it should be in a separate pull request.  A few comments:

Casks/010-editor.rb
```diff
@@ -13,4 +13,11 @@
   end
 
   app "010 Editor.app"
+
+  zap trash: [
+    "~/.config/SweetScape",
+    "~/Documents/SweetScape",
```
bevanjkay, line 19: Does this folder contain user-created files. If so we should not trash it but could add it to `zap rmdir` do that it is removed if empty.

krehel: @bevanjkay, I thought perhaps the same thing.  Both are populated with application files, but I am happy to move both to an rmdir for safety.

Casks/arduino-ide.rb
krehel: Thanks @razvanazamfirei,

Submitting changes.  Little too excited with the zap on whatsize, we can remove those files you noted as they are removed as part of uninstall.

krehel: Pushed commit for this!

Casks/whatsize.rb
krehel: Pushed a fix to remove these files

Casks/010-editor.rb
krehel: To add more context...

By default -

~/.config/Sweetscape contains an editor .ini file
~/Documents/SweetScape contains binary template files and 010-editor script files.  Given users could make their own and place into at least the ~/Documents folder, I agree we should move at least that to rmdir.

razvanazamfirei (approved):

krehel:
@razvanazamfirei and @bevanjkay - thank you both for your patience and guidance.  I'll continue to strive to improve the quality of my commits.

Appreciate it greatly!

razvanazamfirei:
@krehel, you're doing great! :)

PR #146183, opened 2023-04-30
nuclear 0.6.21

krehel:
* Bump version to 0.6.21

* Adjust url to support latest versioning scheme

* Add livecheck

### DIFF
--- a/Casks/nuclear.rb
+++ b/Casks/nuclear.rb
@@ -10,7 +10,7 @@ cask "nuclear" do
 
   livecheck do
     url :url
-    regex(/^v?(\d+(?:\.\d+)+)$/i)
+    regex(/v?(\d+(?:\.\d+)+)/i)
   end
 
   app "nuclear.app"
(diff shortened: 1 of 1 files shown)


krehel:
Adding a note here -

This cask for a while was using a standard x.y.z numbering scheme, and then switched to using a partial commit hash for the version for a period of many months.  The latest few releases are using x.y.z, so this cask has been modified to support that.

Added a livecheck, but please note that releases for this app are marked as "pre-release" - including the version this cask originally pulled.  As such, this looks for latest version and not the latest tag.

The livecheck also assumes sticking to x.y.z numbering format, and if the app switches back to commit hashing versions this will need to be revisited.

razvanazamfirei:
Hi @krehel, thank you for submitting this and adding the live check. `nuclear` is on our pre-released exception list, so that should not be an issue. I made a small change to the regex; since the cask uses the `Git` strategy and not `PageMatch` it still only runs the regex on the tags. Hopefully they'll stick to a sane versioning strategy!

razvanazamfirei (approved):

krehel:
Cheers @razvanazamfirei! Thanks for the fix. Noted for future. Many tricks to learn!

PR #146256, opened 2023-05-01
red-canary-mac-monitor

razvanazamfirei:
More background discussion in https://github.com/Homebrew/homebrew-cask/pull/146071. This is a a cleaner version of the cask.


miccal (approved):

Brandon7CC:
Hey @bevanjkay, @razvanazamfirei, and @miccal!

## A *slightly* better solution
I spent some more time thinking on the uninstall process and decided to go ahead and add an additional uninstall feature to the app. The System "Security" Extension can now be unloaded directly with the [API](https://developer.apple.com/documentation/systemextensions/ossystemextensionrequest/3295273-deactivationrequest) to circumvent the "trash" issue you ran into:
`/usr/bin/open -a "/Applications/Red Canary Mac Monitor.app" --args --deactivate-security-extension`. Since System Extensions are generally considered to be "security relevant" they will need authentication to remove.

## Authentication
However, we're still dealing with user authentication. We have a few options and I'd love your input to make this as seamless as possible for future Homebrew deployment.

The uninstall shell script will now run:
```sh
# Open the app with the System Extension deactivation request
/usr/bin/open -a "/Applications/Red Canary Mac Monitor.app" --args --deactivate-security-extension

# We'll need to do some waiting here... the user is entering their password

# After the user has entered their password kill the app
/usr/bin/killall "Red Canary Mac Monitor"

# After the app is killed - delete it
/bin/rm -rf "/Applications/Red Canary Mac Monitor.app"
```

## Our two options
1. Wait forever until the user enters their password
  a. Will this be a problem for people automating Homebrew?
3. Bail after some amount of time (60 seconds?)
  b. The System Extension will be left installed and to remove it the user will need to reinstall the app (not a problem if upgrading)

Thank you for the merge as wll!

PR #146307, opened 2023-05-02
pingid (add zap, remove livecheck)

krehel:
* Add zap stanza

### DIFF
--- a/Casks/pingid.rb
+++ b/Casks/pingid.rb
@@ -16,4 +16,10 @@ cask "pingid" do
   pkg "PingID.pkg"
 
   uninstall pkgutil: "com.pingidentity.pingid.pcclient"
+
+  zap trash: [
+    "~/Library/Logs/PingID",
+    "~/Library/Preferences/com.pingidentity.pingid.plist",
+    "~/Library/Saved Application State/com.pingidentity.pingid.pcclient.savedState",
+  ]
 end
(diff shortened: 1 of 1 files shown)


razvanazamfirei (dismissed):

Casks/pingid.rb
```diff
@@ -9,11 +9,17 @@
   homepage "https://www.pingidentity.com/"
 
   livecheck do
-    url "https://prod-ping-sfc.zoominsoftware.io/api/bundle/pingid/toc?language=enus"
-    regex(/"title"\s*:\s*"Desktop\s+app\s+v?(\d+(?:\.\d+)+)/i)
+    url :url
+    strategy :extract_plist
```
razvanazamfirei, line 13: We avoid doing this, but can be useful sometimes. There isn't extensive documentation, but it's mentioned [here](https://docs.brew.sh/Cask-Cookbook#stanza-livecheck).

razvanazamfirei (approved):

Casks/pingid.rb
krehel: Thanks @razvanazamfirei.  Livecheck is still one thing I need to figure the ins and outs of.  Apologies on doing this incorrectly - I didn't like removing this.

I'll read these docs a bit more closely.

Thanks for chipping in the fix.!

razvanazamfirei: Really no reason to apologize! :)

PR #146338, opened 2023-05-02
DOSBox Staging 0.80.1

Burrito78:

### DIFF
--- a/Casks/dosbox-staging.rb
+++ b/Casks/dosbox-staging.rb
@@ -1,21 +1,13 @@
 cask "dosbox-staging" do
   version "0.80.1"
-  sha256 "46a256645255e8345981ea357f1416b8ce4bc60a2aba9a86b5122d5075aa7fa"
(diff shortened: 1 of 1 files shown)


razvanazamfirei:
Hi @Burrito78, thank you for submitting your first PR! 😄  Unfortunately, we're not able to accept it for several reasons. The first one is that a formula already exists for dosbox-staging. The second one is that it does not pass our signature verification check that are a requirement for new submissions:

> macOS on ARM requires applications to be signed. Please contact the upstream developer to let them know they should notarize their package.

Hopefully the existing formula will be sufficient. If not, please consider hosting this [in your own Tap](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap).

Burrito78:
@razvanazamfirei Hi! Thanks for the kind reply and the cask changes!

How would we best resolve that we could have both DOSBox Staging CLI as well as GUI (cask) in Homebrew?

bevanjkay:
@Burrito78 if the application is different to the formula we would be able to accept it as a cask. 

However, the application would need to be signed to be accepted, as per the failing audit above.

PR #146447, opened 2023-05-03
hiarcs-chess-explorer (rework)

krehel:
* Adjust livecheck url to support optional [a-z] at version end

* Add zap stanza

* Add legacy versions, remove depends_on macOS version

### DIFF
--- a/Casks/hiarcs-chess-explorer.rb
+++ b/Casks/hiarcs-chess-explorer.rb
@@ -17,7 +17,7 @@ cask "hiarcs-chess-explorer" do
   end
   on_mojave :or_newer do
     version "1.12.2a"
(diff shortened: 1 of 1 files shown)


bevanjkay:
@krehel If the [a-z] matching at the version end is part of the download URL, then it should it be captured as part of the version?

krehel:
@bevanjkay I thought about that also, my bad.  I've made the changes.  It's only reflected in the .pkg download, not listed as an update so I'm guessing it was an update in place scenario.

krehel:
Sorry about that with the sha256.  Hopefully this passes now...

bevanjkay (approved):

PR #146489, opened 2023-05-04
xtorrent (mark discontinued)

krehel:
* Update livecheck url and regex

### DIFF
--- a/Casks/xtorrent.rb
+++ b/Casks/xtorrent.rb
@@ -7,15 +7,11 @@ cask "xtorrent" do
   desc "Torrent client"
   homepage "http://www.xtorrent.com/"
 
-  livecheck do
-    url :homepage
-    strategy :page_match do |page|
-      match = page.match(/version\s*(\d+(?:\.\d+)*)\s*\(v?(\d+(?:\.\d+)*)\)/i)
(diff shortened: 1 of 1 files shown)


krehel:
Thanks @razvanazamfirei - I've made those changes and pushed.

At what point do you decide to stop supporting an outdated cask like this? (just for my reference)

As you've mentioned it's been about 10 years, which is also when that PubSub framework was deprecated.

Eventually this will fail to work on any reasonably supported OS version.

razvanazamfirei:
Hi @krehel, 
I think this is the general guideline we're using:

> We generally haven't removed casks unless they are troublesome (maintainer burden), the download no longer exists, the homepage no longer works, issues have been reported for the cask and it isn't popular enough to fix or the annual downloads are very low (below 15-20).
> For the most part there isn't any reason to remove casks if they are still usable without issue - even if they have been discontinued.
From https://github.com/Homebrew/homebrew-cask/pull/142262#issuecomment-1453503013

krehel:
Thanks for that note.  Fair enough!

Thanks for your help and inputs!

razvanazamfirei (approved):

PR #146668, opened 2023-05-07
tuxera-ntfs (zap, livecheck)

krehel:
* Update livecheck url and regex to pull latest version

* Add files to uninstall delete

* Add zap stanza

### DIFF
--- a/Casks/tuxera-ntfs.rb
+++ b/Casks/tuxera-ntfs.rb
@@ -9,7 +9,7 @@ cask "tuxera-ntfs" do
 
   livecheck do
     url "https://ntfsformac.tuxera.com/support"
-    regex(/Release\s*v?(\d+(?:\.\d+)*)/i)
+    regex(/Release\s*?v?(\d+(?:\.\d+)*)/i)
   end
 
   auto_updates true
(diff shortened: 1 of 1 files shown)


Casks/tuxera-ntfs.rb
```diff
@@ -8,8 +8,8 @@
   homepage "https://ntfsformac.tuxera.com/"
 
   livecheck do
-    url :homepage
-    regex(/href=.*?tuxerantfs[._-]?v?(\d+(?:\.\d+)*)\.dmg/i)
+    url "https://ntfsformac.tuxera.com/support"
+    regex(/Release\s*v?(\d+(?:\.\d+)*)/i)
```
razvanazamfirei, line 12:
```suggestion
    regex(/Release\s*?v?(\d+(?:\.\d+)+)/i)
```

Casks/tuxera-ntfs.rb
```diff
@@ -23,5 +23,14 @@
             pkgutil: [
               "com.tuxera.pkg.Tuxera_NTFS",
               "com.tuxera.pkg.Tuxera_NTFS_compat",
+            ],
+            delete:  [
+              "/Library/Application Support/Tuxera NTFS",
+              "/Library/Preferences/com.tuxera.NTFS.plist",
```
razvanazamfirei, line 29: Shouldn't this be in zap delete?

razvanazamfirei (changes requested):

Casks/tuxera-ntfs.rb
krehel: Hey @razvanazamfirei - these are in the system library and may require sudo to delete.  My understanding has been those should go into delete and not zap.  Am I incorrect in this thought?

Casks/tuxera-ntfs.rb
krehel: I can accept the first change to `\s*?` but the second change wouldn't match as the current version is `2022`, which would cause a non-match.

razvanazamfirei: Gotcha! Should be changed now.

Casks/tuxera-ntfs.rb
razvanazamfirei: They're the type of files we would have in zap. 

You can use all the uninstall directives in zap stanzas, including delete. See below: https://github.com/Homebrew/homebrew-cask/blob/72fcbda1d5b16f8220c9c528cff89ee9c838ca46/Casks/powerpanel.rb#L27-L34

krehel: Gotcha!  My miss, I'll correct both and push changes.

razvanazamfirei (approved):

PR #146677, opened 2023-05-08
cloudmounter: update `livecheck`

miccal:


bevanjkay:
Should we consider adding a method to `:sparkle` strategy to surface this?
`strategy :sparkle, &:short_version_without_build

reitermarkus:
Doesn't seem that common. Only 15 casks in total use `.short_version.` and only four of those use `.short_version.sub`.

reitermarkus (approved):

PR #146844, opened 2023-05-09
jira-cli: 1.4.0 (new cask)

idleberg:

### DIFF
--- a/Casks/jira-cli.rb
+++ b/Casks/jira-cli.rb
@@ -1,0 +1,14 @@
+cask "jira-cli" do
+  arch arm: "arm64", intel: "x86_64"
+
+  version "1.4.0"
+  sha256 arm:   "9503d084743d96d45ec7799f40b32895eaf06135051dfc47e835724582f36e83",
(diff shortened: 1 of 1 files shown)


razvanazamfirei:
Hi @idleberg, thank you for your first submission! In cases like this, we recommend submitting it to homebrew/core first as per below.

> Common reasons to reject a cask entirely:
> ...
> App is both open-source and CLI-only (i.e. it only uses the binary artifact). In that case, and [in the spirit of deduplication](https://github.com/Homebrew/homebrew-cask/issues/15603), submit it first to [homebrew/core](https://github.com/Homebrew/homebrew-core) as a formula that builds from source. If it is rejected, you may then try again as a cask (link to the issue from your pull request so we can see the discussion and reasoning for rejection).

razvanazamfirei:
To add to that, we cannot accept it at this time because it does not pass our signature verification check that are a requirement for new submissions:

> macOS on ARM requires applications to be signed. Please contact the upstream developer to let them know they should notarize their package.

You can also consider hosting this [in your own Tap](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap) and if the developers sign their app in the future we would then be happy to consider adding this to `homebrew-cask`.

PR #147109, opened 2023-05-13
vagrant (update livecheck)

krehel:
* Update livecheck to :github_latest to avoid pre-releases

I updated the livecheck for Vagrant to use `:github_latest` instead of `:git` as Vagrant has been publishing pre-release versions which have been getting picked up by the livecheck.


razvanazamfirei (approved):

razvanazamfirei:
Great! Thank you!

PR #147129, opened 2023-05-13
chessx (update livecheck)

krehel:
* Update livecheck to use rss feed instead of :header_match

I noticed this cask has been pulling version `1.5.6` for the last couple updates rather than the most recent version despite using `:header_match`.  

This cask has an rss feed for versions, so updated livecheck to check that instead.

### DIFF
--- a/Casks/chessx.rb
+++ b/Casks/chessx.rb
@@ -9,8 +9,8 @@ cask "chessx" do
   homepage "https://chessx.sourceforge.io/"
 
   livecheck do
-    url "https://sourceforge.net/projects/chessx/files/latest/download"
-    strategy :header_match
+    url "https://sourceforge.net/projects/chessx/rss?path=/chessx"
+    regex(%r{chessx/(\d+(?:\.\d+)+).*?\.dmg}i)
   end
 
   pkg "chessx-installer.mpkg"
--- a/Casks/chessx.rb
+++ b/Casks/chessx.rb
@@ -9,8 +9,8 @@ cask "chessx" do
   homepage "https://chessx.sourceforge.io/"
 
   livecheck do
-    url "https://sourceforge.net/projects/chessx/rss?path=/chessx"
-    regex(%r{chessx/(\d+(?:\.\d+)+).*?\.dmg}i)
+    url "https://sourceforge.net/projects/chessx/rss"
+    regex(%r{url=.*?/chessx[._-]v?(\d+(?:\.\d+)+)\.dmg}i)
   end
 
   pkg "chessx-installer.mpkg"


Casks/chessx.rb
```diff
@@ -9,8 +9,8 @@
   homepage "https://chessx.sourceforge.io/"
 
   livecheck do
-    url "https://sourceforge.net/projects/chessx/files/latest/download"
-    strategy :header_match
+    url "https://sourceforge.net/projects/chessx/rss?path=/chessx"
+    regex(%r{chessx/(\d+(?:\.\d+)+).*?\.dmg}i)
```
samford, line 13:
```suggestion
    regex(%r{url=.*?/chessx[._-]v?(\d+(?:\.\d+)+)\.dmg}i)
```

Matching the version from the directory name works (and accommodates their older setup, where they used a generic `chessx.dmg` file name) but they've used a consistent file format like `chessx-1.2.3.dmg` since ~2013. The only exceptions are the unstable files like `chessx-1.5.7a.dmg` in the `beta` directory, which we don't want to match.

There's something to be said for matching the version from the filename, as this would use a more standard pattern for `SourceForge` checks and is stricter than the regex this PR is proposing. My thinking is that this looser regex would match an unstable version if upstream used a directory name like `1.2.3` with unstable `chessx-1.2.3a.dmg` files inside (instead of something like `beta`).

Casks/chessx.rb
```diff
@@ -9,8 +9,8 @@
   homepage "https://chessx.sourceforge.io/"
 
   livecheck do
-    url "https://sourceforge.net/projects/chessx/files/latest/download"
-    strategy :header_match
+    url "https://sourceforge.net/projects/chessx/rss?path=/chessx"
```
razvanazamfirei, line 12:
```suggestion
    url "https://sourceforge.net/projects/chessx/rss
```
In addition to the great suggestions below, I think the url can also be simplified. 

Not an issue in this specific cask, but may be worth highlighting some of the potential considerations when using the implied [Sourceforge](https://rubydoc.brew.sh/Homebrew/Livecheck/Strategy/Sourceforge.html) strategy vs PageMatch.

Casks/chessx.rb
krehel: Thanks @samford!  Valid points.  I'll work to be more strict with regexes in the future.

Appreciate the inputs.  Made changes as suggested.

Casks/chessx.rb
krehel: Thank you @razvanazamfirei!  Valid point.  Should've seen that.  😄

Made the change

razvanazamfirei (approved):

Casks/chessx.rb
samford: @razvanazamfirei In this case, the original URL (with the `path` parameter) was preferable here. This restricts the RSS feed to only the `chessx` directory, which is where the release files are kept.

Basically, the idea is that we do this when appropriate/possible because the RSS feed contains a limited number of entries. The URL without the path will also contain entries from the `chessx-extras` directory and, theoretically, the files we're interested in could be pushed out of the feed by files we're not interested in.

Hope this helps to explain this URL pattern across other casks but let me know if you have questions.

razvanazamfirei: @samford, thank you for making the clarification regarding the url. On first glance, the files looked identical and my local tests gave the same results, which is why I suggested the simpler version. 

>It may also be worth mentioning that this livecheck block still uses the Sourceforge strategy with these changes (not PageMatch directly)

Yeah, this is exactly what I was trying to point out; thank you for stating it much more clearly than I could and for all your detailed explanations. I'm always low-key excited reading your comments.

krehel: Thanks @samford and @razvanazamfirei for your time and explanation of the nuances for how this works. This level of education is superb.

I understand your primary goal is stability and making correct changes, and as someone new to submitting PR's for the project I appreciate the specificity, kindness, and willingness to work to make everything correct.

Thank you!

PR #147144, opened 2023-05-14
kiibohd-configurator: migrate from `homebrew/cask-drivers`

Cellane:
Similar to #147012, this PR migrates a deleted cask from Homebrew/homebrew-cask-drivers#3443.
I realized during setting up a new laptop that I can’t install `kiibohd-configurator`, so hoping it can be re-added, even if it’s not a super popular application.

Related issue: Homebrew/homebrew-cask-drivers#3414.


Cellane:
Hmm… the failing test is interesting… I don’t know much about the requirements imposed by Apple, but kiibohd-configurator previously installed from Homebrew is running fine on my old M1 Pro Max…

razvanazamfirei (approved):

razvanazamfirei:
Thank you for submitting this! Looks great! 👍

PR #147156, opened 2023-05-14
Make Mixxx livecheck regex more robust

fwcd:
This makes the regex less dependent on the order of the links on the website by only considering release download links in the livecheck.

---

### DIFF
--- a/Casks/mixxx.rb
+++ b/Casks/mixxx.rb
@@ -8,7 +8,7 @@ cask "mixxx" do
   homepage "https://www.mixxx.org/"
 
   livecheck do
-    url "https://www.mixxx.org/download/"
+    url "https://mixxx.org/download/"
     regex(%r{href=["']?https://downloads\.mixxx\.org/releases/.*/mixxx[-_.]v?(\d+(?:\.\d+)+)[-_.]macos(?:intel|arm)\.dmg}i)
   end
 
--- a/Casks/mixxx.rb
+++ b/Casks/mixxx.rb
@@ -12,6 +12,8 @@ cask "mixxx" do
     regex(%r{href=.*?/mixxx[._-]v?(\d+(?:\.\d+)+)[._-]macos(?:intel|arm)\.dmg}i)
   end
 
+  conflicts_with cask: "homebrew/cask-versions/mixxx-snapshot"
+
   app "Mixxx.app"
 
   zap trash: [
--- a/Casks/mixxx.rb
+++ b/Casks/mixxx.rb
@@ -9,7 +9,7 @@ cask "mixxx" do
 
   livecheck do
     url "https://mixxx.org/download/"
-    regex(%r{href=["']?https://downloads\.mixxx\.org/releases/.*/mixxx[-_.]v?(\d+(?:\.\d+)+)[-_.]macos(?:intel|arm)\.dmg}i)
+    regex(%r{href=.*?/mixxx[._-]v?(\d+(?:\.\d+)+)[._-]macos(?:intel|arm)\.dmg}i)
   end
 
   app "Mixxx.app"
--- a/Casks/mixxx.rb
+++ b/Casks/mixxx.rb
@@ -9,7 +9,7 @@ cask "mixxx" do
 
   livecheck do
     url "https://www.mixxx.org/download/"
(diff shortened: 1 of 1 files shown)


Casks/mixxx.rb
```diff
@@ -9,7 +9,7 @@
 
   livecheck do
     url "https://www.mixxx.org/download/"
-    regex(%r{href=.*?/mixxx[._-]v?(\d+(?:\.\d+)+)[._-]macosintel\.dmg}i)
+    regex(%r{href=["']?https://downloads\.mixxx\.org/releases/.*/mixxx[-_.]v?(\d+(?:\.\d+)+)[-_.]macos(?:intel|arm)\.dmg}i)
```
razvanazamfirei, line 12:
```suggestion
    strategy :github_latest
```

Casks/mixxx.rb
```diff
@@ -9,7 +9,7 @@
 
   livecheck do
     url "https://www.mixxx.org/download/"
```
razvanazamfirei, line 11:
```suggestion
    url "https://github.com/mixxxdj/mixxx/"
```

razvanazamfirei (dismissed):
Hi @fwcd, thank you for all the effort you put in. This looks good, but as I was checking the regex I noticed that the developers have an active Github repository. We have a robust strategy for Github repository. That's why I switched this from PageMatch to GithubLatest.

Casks/mixxx.rb
fwcd: That's an option sure, since the binaries themselves aren't hosted on GitHub though, I felt like it makes sense to query the download page which always points to a release that is guaranteed to be valid (this could fail if Mixxx were to e.g. release on their website, but not on GitHub or similar). Of course that's an unlikely scenario, but the current way ensures that the livecheck is always in sync with the download server.

I'm open to this too though, so if querying GitHub is generally preferential, we can of course do so.

samford, line 12:
```suggestion
    regex(%r{href=.*?/mixxx[._-]v?(\d+(?:\.\d+)+)[._-]macos(?:intel|arm)\.dmg}i)
```

A few items:

* Matching the first part of the URL before `/releases/` makes the regex more brittle without providing any particular benefit (i.e., both stable and unstable releases use the same domain). In this scenario, we only want to explicitly match the parts that make a meaningful difference. The existing regex only matches the filename but also matching the `/releases/` part of the path could theoretically provide some benefit (in terms of distinguishing stable releases from snapshots), so that's really all that we would need to add here. However, it isn't necessary right now because snapshot files use a different filename format that isn't matched by the existing regex (e.g., `https://downloads.mixxx.org/snapshots/main/mixxx-2.5-alpha-33-g3e500b1abc-macosintel.dmg`), so the standard `href=.*?` is fine for now.
* I do agree with using `(?:intel|arm)`, as that could be helpful in the future. At the moment there's only an Intel version but there are ARM snapshots, so it's conceivable that they may release both in the future. When that happens, this will probably be updated to use something like `arch arm: "arm", intel: "intel"` and we would use `#{arch}` in place of this added group in the regex.
* `[._-]` is the standard order for this particular character class. I originally put them in that order kind of based on frequency. It's functionally the same if they're in a different order but standardization sometimes makes it easier to `grep` for particular regex patterns across formulae/casks.

samford (commented):
Unless I'm missing something, most of the changes to the regex are unnecessary (i.e., matching the URL scheme/domain/path doesn't provide a meaningful benefit but the specificity makes the regex more brittle). I've added a suggestion to pare it down to the useful change (matching both Intel and ARM files).

fwcd:
> most of the changes to the regex are unnecessary (i.e., matching the URL scheme/domain/path doesn't provide a meaningful benefit but the specificity makes the regex more brittle)

My intent was to avoid matching snapshot download links (which I stumbled on while working on https://github.com/Homebrew/homebrew-cask-versions/pull/16681). This currently relies on the fact that release links are placed before snapshot links on the page, which probably is a bit more fragile than just matching against `/release`.

Casks/mixxx.rb
fwcd: > However, it isn't necessary right now because snapshot files use a different filename format

Perhaps still useful since the templated download link hardcodes the `/release` anyway? My feeling was the closer we get the regex to that, the less fragile.

miccal (approved):

PR #147458, opened 2023-05-18
smooze-pro: added conflicts_with "smooze"

cawaltrip:

### DIFF
--- a/Casks/smooze-pro.rb
+++ b/Casks/smooze-pro.rb
@@ -13,6 +13,7 @@ cask "smooze-pro" do
   end
 
   auto_updates true
+  conflicts_with cask: "smooze"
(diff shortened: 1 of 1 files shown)


cawaltrip:
This installs Smooze Pro, not Smooze.  They are two distinct applications.

reitermarkus:
Yes, so why does it conflict with `smooze`?

cawaltrip:
Running one of the programs kills the other.  I guess there's nothing that technically prevents you from having both installed, but only one of them can be used at a time.  Since that's the case, should I go ahead and close this request?

reitermarkus:
> should I go ahead and close this request?

Yes, as long as they don't install the same stuff, we don't need to mark it as a conflict.

PR #147459, opened 2023-05-18
smooze: corrected to work with "smooze-pro"

cawaltrip:
Updated version back to 1.9.26 and removed the sparkle livecheck because it is being used to redirect users to Smooze Pro.  Added a conflict with Smooze Pro because the developer has stated the two applications cannot work with each other.  Added more directories to the zap list.  Added a caveat that this application is in maintenance mode and pointed users to the "smooze-pro" cask.

### DIFF
--- a/Casks/smooze.rb
+++ b/Casks/smooze.rb
@@ -12,7 +12,6 @@ cask "smooze" do
   end
 
   auto_updates true
-  conflicts_with cask: "smooze-pro"
   depends_on macos: ">= :sierra"
 
   app "Smooze.app"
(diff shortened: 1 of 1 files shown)


cawaltrip:
There are two products: Smooze and Smooze Pro.  This is for Smooze.  From discussions with the developer (see attachment), Smooze won't be receiving new feature updates, but may still be updated if OS changes require it.  We can't rely on the version number reported by Sparkle because it doesn't match the version number of the application itself (it's being used to point users to Smooze Pro).

[SmoozeUpdates.pdf](https://github.com/Homebrew/homebrew-cask/files/11510576/SmoozeUpdates.pdf)

I thought about adding this to the [Homebrew/homebrew-cask-versions](https://github.com/Homebrew/homebrew-cask-versions) repo, but seeing the discussions around trying to deprecate it, I thought I'd just update this one instead.  More than happy to change things around as needed!

miccal (approved):

PR #148435, opened 2023-06-03
xmind 22.11.3771-202305311748

WangEdward:
https://github.com/orgs/Homebrew/discussions/4553

Created with `brew bump-cask-pr`.

PR #148476, opened 2023-06-04
add caveats xmind.rb

WangEdward:
https://github.com/Homebrew/homebrew-cask/pull/148435#issuecomment-1575415219


bevanjkay:
Thanks for this PR @WangEdward. Caveats should only be used for Homebrew-specific installation instructions - https://docs.brew.sh/Cask-Cookbook#stanza-caveats
So in this case we will pass on adding the caveat here.

PR #148477, opened 2023-06-04
Remove ql-ansilove

nrlquaker:
[Repo](https://github.com/ansilove/QLAnsilove) is archived.
And the app no longer works.

From original README:
>QLansilove is being moved to archive/read-only. Apple has deprecated quicklook generators in Big Sur in favor of including previews in app extensions. I will look to see if it would be possible to include this functionality in Ascension which is a more appropriate and Apple-like place for this functionality to land. -AK


razvanazamfirei:
This should probably be set to discontinued instead, but few people are actually using it. 

```
ql-ansilove
30 days: n/a
90 days: n/a
365 days: 62 (#3278)
```

razvanazamfirei (approved):

PR #125996, opened 2022-06-18
add dockutil v3.0.2 cask

ei9h7:
Verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions).
- [x] `brew audit --cask --online homebrew/cask/dockutil` is error-free.
- [x] `brew style --fix homebrew/cask/dockutil` reports no offenses.

Additionally,

- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [ ] `brew audit --new-cask homebrew/cask/dockutil` worked successfully.
  - says conflicts with formula but this is expected.
- [x] `brew install --cask dockutil` worked successfully.
- [x] `brew uninstall --cask dockutil` worked successfully.

PR #148504, opened 2023-06-05
Create dockutil.rb

mofneko:

### DIFF
--- a/Casks/dockutil.rb
+++ b/Casks/dockutil.rb
@@ -1,0 +1,13 @@
+cask "dockutil" do
+  version "3.0.2"
+  sha256 "175137ea747e83ed221d60b18b712b256ed31531534cde84f679487d337668fd"
+
+  url "https://github.com/kcrawford/dockutil/releases/download/#{version}/dockutil-#{version}.pkg"
+  name "Dockutil"
+  desc "Command-line utility for managing macOS dock items"
+  homepage "https://github.com/kcrawford/dockutil"
+
+  pkg "dockutil-#{version}.pkg"
+
+  uninstall pkgutil: "dockutil.cli.tool"
+end


bevanjkay:
Thank you for your submission @mofneko but `dockutil` is already available in [`homebrew-core`](github.com/homebrew/homebrew-core)

mofneko:
But core is not the latest version.

bevanjkay:
Then the formula should be updated in `homebrew-core` to be the latest version.

mofneko:
Since dockutil is provided in pkg format, it is appropriate to provide it in cask.

bevanjkay:
See cask docs here regarding rejected casks - https://docs.brew.sh/Acceptable-Casks#rejected-casks

> App is both open-source and CLI-only (i.e. it only uses the binary artifact). In that case, and in the spirit of deduplication, submit it first to homebrew/core as a formula that builds from source. If it is rejected, you may then try again as a cask (link to the issue from your pull request so we can see the discussion and reasoning for rejection).

mofneko:
https://github.com/Homebrew/homebrew-core/pull/132838
I also created a PR for core.
If rejected here, please consider merging into cask.

chenrui333:
> [Homebrew/homebrew-core#132838](https://github.com/Homebrew/homebrew-core/pull/132838) I also created a PR for core. If rejected here, please consider merging into cask.

it can be submitted into core, but it has to be built from source

mofneko:
After all, since the official is provided as pkg, I think that it is correct to distribute it with cask.
In that case, core packages will issue a PR for removal.
do you agree with my idea?

bevanjkay:
@mofneko Because the submission is for an open source CLI package, it should be submitted to `homebrew-core` and built from source regardless of how it is packaged for distribution by upstream.

p-linnane:
See also:
https://github.com/Homebrew/homebrew-cask/pull/125996
https://github.com/Homebrew/homebrew-cask/pull/135196

We will not be accepting this into `homebrew-cask`.

PR #148556, opened 2023-06-05
various: formatting for macOS 14

p-linnane:
Fixes various formatting issues for macOS 14 preliminary support.

### DIFF
--- a/Casks/transcribe.rb
+++ b/Casks/transcribe.rb
@@ -10,6 +10,11 @@ cask "transcribe" do
     livecheck do
       skip "Legacy version"
     end
+
+    zap trash: [
+      "~/Library/Preferences/com.seventhstring.transcribe.plist",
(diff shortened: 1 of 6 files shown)


Casks/transcribe.rb
```diff
@@ -61,9 +81,4 @@
   app "Transcribe!.app"
 
   uninstall quit: "com.seventhstring.transcribe"
-
-  zap trash: [
-    "~/Library/Preferences/com.seventhstring.transcribe.plist",
-    "~/Library/Preferences/Transcribe!#{version.major} Preferences",
-  ]
```
Bo98, line 68: Does this need to move?

Bo98 (approved):
Thanks!

Casks/transcribe.rb
p-linnane: Putting it outside the `on_${os}` blocks causes the cask to become unreadable, since the version information stays within those blocks.

Issue #148497, opened 2023-06-04
powerphotos: Download failed on Cask 'powerphotos' with message: Failure while executing; The requested URL returned error: 404


Danrancan:
### Verification

- [X] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [X] I have retried my command with `--force`.
- [X] I ran `brew update-reset && brew update` and retried my command.
- [X] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [X] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [X] I made doubly sure this is not a [checksum does not match](https://docs.brew.sh/Common-Issues#cask---checksum-does-not-match) error.

### Description of issue

When updating casks with homebrew, I am having issues downloading the Powerphotos app for Mac OS. It seems the the download url is incorrect.

Here is my log:
```
powerphotos: Download failed on Cask 'powerphotos' with message: Failure while executing; `/usr/bin/env /usr/local/Homebrew/Library/Homebrew/shims/shared/curl --disable --cookie /dev/null --globoff --show-error --user-agent Homebrew/4.0.20-20-ge9a1cd4\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 11.7.7\)\ curl/7.87.0 --header Accept-Language:\ en --retry 3 --fail --location --silent --head --request GET https://www.fatcatsoftware.com/powerphotos/PowerPhotos_217.zip` exited with 22. Here's the output:
curl: (22) The requested URL returned error: 404
HTTP/1.1 404 Not Found
Date: Sun, 04 Jun 2023 18:34:59 GMT
Server: Apache/2.4.38 (Debian)
Content-Length: 285
Content-Type: text/html; charset=iso-8859-1
```

### Command that failed

brew update --cask && brew upgrade --cask

### Output of command with `--verbose --debug`

```shell
==> Upgrading 2 outdated packages:
plistedit-pro 1.9.5,933 -> 1.9.6
powerphotos 2.1.7,1955 -> 2.1.7
==> Upgrading plistedit-pro
==> Downloading https://www.fatcatsoftware.com/plisteditpro/downloads/PlistEditPro_196.zip
[...]

PR #148643, opened 2023-06-06
powerphotos: update url

razvanazamfirei:
Looks like the link changed again.

https://github.com/Homebrew/homebrew-cask/issues/148497#issuecomment-1579399156


p-linnane (approved):

krehel:
Hey @razvanazamfirei - sidebar question here - is there a way to test and validate the legacy versions within a cask?  I would've thought the checks run when updating the `on_big_sur` stanza would check that block?    

If so I could've caught this the first time when validating locally.

razvanazamfirei:
Hi @krehel, using `brew audit --os=` should work!

razvanazamfirei:
@krehel, this is a new feature. It was added in the last month; no reason to worry!

PR #149256, opened 2023-06-15
opencpn 5.8.4

krehel:
* Update to version 5.8.4

* Update url to latest format

* Update livecheck to parse latest release artifact format

Hi maintainers,

This cask has been reworked for both livecheck and latest version of this cask to bring it up to the latest format.  

Of note for this cask, the `.pkg` artifact naming scheme for downloads has changed some versions back from `OpenCPN_x.y.z` to `OpenCPN_x.y.z-0+build.commithash` (extrapolating a bit from the pkg name).

I've gotten the updates to work, but the url feels a bit fragile. If the format changes or goes to the previous style of `x.y.z`, this will break because of the `-` char in the download url.  I suppose that could be said for any cask, though.  😄 

This can be avoided and both old and new formats can be supported if the second part of the csv starts with a `-` char, as the url string can be updated to `...#{version.csv.first}#{version.csv.second}` in lieu of `...#{version.csv.first}-#{version.csv.second}`, but starting a csv with a `-` seems a bit clunky. 

However, starting with a `-` would also yield being able to install the file with `pkg "OpenCPN_#{version}.pkg"` instead of `pkg "OpenCPN_#{version.first.csv}-#{version.second.csv}"`

I am mentioning this as maybe there is a preferred implementation from your deeper knowledge and style guide.

I intentionally left the `+` in the version number, rather than replacing it with a `,` and therefore leaving the version as 2 csv segments instead of 3.

Livecheck scans the assets in the latest release, as the asset name is different than the release version page.  I borrowed this from another cask and updated accordingly.

Your review and any updates are appreciated.

Thank you for your time!

### DIFF
--- a/Casks/opencpn.rb
+++ b/Casks/opencpn.rb
@@ -1,19 +1,32 @@
 cask "opencpn" do
-  version "5.6.2"
-  sha256 "dd093babd8b565cced08c0fd6467c70e6af4da707caf5ddc47e64b601944467a"
+  version "5.8.4,0+637.1637c28"
+  sha256 "2c3dc08908c002dcfe61a67b8e594a14d4de7ec486f6e44fc4ddc18219b65b44"
 
-  url "https://github.com/OpenCPN/OpenCPN/releases/download/Release_#{version}/OpenCPN_#{version}.pkg",
+  url "https://github.com/OpenCPN/OpenCPN/releases/download/Release_#{version.csv.first}/OpenCPN_#{version.csv.first}-#{version.csv.second}.pkg",
       verified: "github.com/OpenCPN/OpenCPN/"
   name "OpenCPN"
   desc "Full-featured and concise ChartPlotter/Navigator"
   homepage "https://www.opencpn.org/"
 
   livecheck do
-    url :url
-    regex(/v?(\d+(?:\.\d+)+)$/i)
+    url "https://github.com/OpenCPN/OpenCPN/releases/latest"
+    regex(%r{href=.*?/OpenCPN[._-]?v?(\d+(?:\.+\d+)+)[._-]?([0-9]?[+-]?\d*\.[0-9a-z]*)\.pkg}i)
+    strategy :header_match do |headers, regex|
+      next if headers["location"].blank?
+
+      # Identify the latest tag from the response's `location` header
+      latest_tag = File.basename(headers["location"])
+      next if latest_tag.blank?
+
+      # Fetch the assets list HTML for the latest tag and match within it
+      assets_page = Homebrew::Livecheck::Strategy.page_content(
+        @url.sub(%r{/releases/?.+}, "/releases/expanded_assets/#{latest_tag}"),
+      )
+      assets_page[:content]&.scan(regex)&.map { |match| "#{match[0]},#{match[1]}" }
+    end
   end
 
-  pkg "OpenCPN_#{version}.pkg"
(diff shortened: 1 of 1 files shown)


krehel:
I am adding a zap as a separate commit shortly.

razvanazamfirei:
Hi, @krehel! Thank you for submitting this. Their naming scheme seems decently stable for now. The livecheck looks good, but I hope @samford can chime in regarding the robustness of the versioning strategy.

krehel:
Thanks @razvanazamfirei!  Just looking to keep it as simple as possible and error free!

samford (commented):
The `GithubLatest` strategy was updated fairly recently to check the GitHub API, so we can iterate through the assets in a simpler way now (instead of my previous approach that you've used). [For what it's worth, we've also added a separate `GithubReleases` strategy for the instances where it's absolutely necessary to iterate through multiple releases.] This reminds me that I need to get back to updating these `strategy` blocks to use `GithubLatest` 😅

I've added a suggestion that also modifies the regex to match all the existing filename formats (and simplify it a little), as I'm not convinced upstream will stick to one format. The current pkg filenames (with their release date) are:

* 2023-06-15: `OpenCPN_5.8.4-0+637.1637c28.pkg`
* 2023-05-01: `OpenCPN_5.8.2-0+458.9f64845.pkg`
* 2023-04-13: `OpenCPN_5.8.0-0+405.0b3b417.pkg`
* 2022-04-22: `OpenCPN_5.6.2.pkg`
* 2021-12-05: `OpenCPN_5.6.0.pkg`
* 2020-10-17: `OpenCPN_5.2.4+1.6b314e6.pkg`
* 2020-07-16: `OpenCPN_5.2.0+1.4d956e1.pkg`

It's worth mentioning that the regex in the suggestion also makes the second capture group optional (to account for filenames like `OpenCPN_5.6.0.pkg`) and adapts the returned version string in the `strategy` block based on the presence of `match[2]`.

I've also updated the regex to match both dmg and pkg files. Historically, they've only published pkg files but the newest release also provides a dmg file, so I want to make sure that the check doesn't break if they only publish dmg files sometime in the future.

-----

In terms of the `-0` part in some filenames (e.g., `OpenCPN_5.8.4-0+637.1637c28.pkg`), I think we should simply include the `-` delimiter in the second capture group. There were previously releases like `OpenCPN_5.2.0+1.4d956e1.pkg` that don't have this part, so hardcoding the `-` in the cask `url` wouldn't work for that format. With the variability in the filename over time, this could become a headache for contributors/maintainers.

If we include `-` in the second part of the `version` and remove the hardcoded `-` in the `url`, then `OpenCPN_#{version.csv.first}#{version.csv.second}.pkg` will work for all the filenames mentioned above. For example, for a version like `5.6.2`, `version.csv.second` would be `nil`, so nothing would be interpolated for the second part and the filename in the `url` would correctly be `OpenCPN_5.6.2.pkg`.

A version like `5.8.4,-0+637.1637c28` does feel a _little_ weird but the `free-gpgmail` cask can have a version like `5,2022.2,-unsigned` or `5,2022.2,-signed` in some cases, so it isn't unprecedented. I think the improved functionality more than makes up for the syntactical weirdness. However, that's just my two cents [as someone who has spent more time in homebrew/core than homebrew/cask] and the cask maintainers may have a different view.

miccal (approved):

krehel:
Thanks @miccal for updating the Caskfile with the changes per @samford!

@samford - thank you as always for imparting valued wisdom on best approaches!

PR #149389, opened 2023-06-18
Update microsoft-edge 114.0.1823.51 and change download url

WangEdward:
This change can allow autobump and brew-bump-pr to work ("folder" will be changed upon every update)

### DIFF
--- a/Casks/microsoft-edge.rb
+++ b/Casks/microsoft-edge.rb
@@ -1,19 +1,17 @@
 cask "microsoft-edge" do
-  folder = on_arch_conditional arm:   "dcaed03b-bd41-40c0-bd42-9c4ab774c5f9",
-                               intel: "99b54385-40f7-4275-8c97-386da51bf238"
   linkid = on_arch_conditional arm: "2093504", intel: "2069148"
 
-  version "114.0.1823.43"
-  sha256 arm:   "c4d0db7bd50971c6dea451823d19b7c6bc04e1f62fa357464871ad2032656283",
-         intel: "e43aa0cf804a796078cc379dac0f22fcacfa569e6fca279522455a2bd50d7908"
+  version "114.0.1823.51"
+  sha256 arm:   "79da7a86a8e18b2eb62a703ef4fc43533abdc0c71afc39b93ccbe88bbc1affc2",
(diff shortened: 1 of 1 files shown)


Casks/microsoft-edge.rb
```diff
@@ -1,19 +1,17 @@
...
 
-  version "114.0.1823.43"
-  sha256 arm:   "c4d0db7bd50971c6dea451823d19b7c6bc04e1f62fa357464871ad2032656283",
-         intel: "e43aa0cf804a796078cc379dac0f22fcacfa569e6fca279522455a2bd50d7908"
+  version "114.0.1823.51"
+  sha256 arm:   "79da7a86a8e18b2eb62a703ef4fc43533abdc0c71afc39b93ccbe88bbc1affc2",
+         intel: "bc4448f6a13224d6077afa0be5bc660b8730ce1ae3354912abb2d36bcc8e9d19"
 
-  url "https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/#{folder}/MicrosoftEdge-#{version}.pkg"
+  url "https://go.microsoft.com/fwlink/?linkid=#{linkid}"
```
bevanjkay, line 8: @WangEdward Because the URL is unversioned if we go down this route we would need to make the `sha256` be `:no_check` which is less than ideal. If the folder changes we can include it in the livecheck.

WangEdward: the `folder` for two arch is different, which will make the version string look awful if we store it in the version string

bevanjkay: @WangEdward It's the standard procedure for `url` parts that change dynamically from version to version. Replacement PR incoming.

WangEdward: If we don't change the `folder` automatically, autobump and brew-bump-pr may not work  https://github.com/Homebrew/homebrew-cask/commit/4a00f3f0b4d93e63d467135ea5d21196ae7934d7

bevanjkay: It's included in the livecheck - https://github.com/Homebrew/homebrew-cask/pull/149396

bevanjkay: Autobump may have issues yet, but this is a separate issue that @razvanazamfirei is looking into with the `brew bump` and `brew bump-cask-pr` commands not simulating the alternative architectures.

PR #149788, opened 2023-06-24
sonic-robo-blast-2 2.2.11

krehel:
* Update to latest version 2.2.11

* Update url and livecheck

For the latest release, it looks like the naming scheme was changed for the url.  It is not clear if this will remain the case, but seeing as this app has updated very infrequently over the past couple years, it can be adjusted back if it changes in the next iteration.

Switched `livecheck` to use `:github_latest` instead of the website downloads page, which also points to the GitHub artifact.


krehel:
Note the homepage is accessible locally and passes `brew audit` locally so I think this may just be a CI issue.

reitermarkus:
Likely the website is protected by CloudFlare.

reitermarkus (approved):

PR #149827, opened 2023-06-25
Elgato Casks: various updates

miccal:


bevanjkay:
The downloads page does not embed the latest version in the page. They are replaced with Javascript from a GrapQL API query, so the livechecks are not returning the latest version, `elgato-wave-link` being an example of this, the correct latest version is `1.8.0.448` as per the previously open PR.

miccal:
> The downloads page does not embed the latest version in the page. They are replaced with Javascript from a GraphQL API query, so the livechecks are not returning the latest version, `elgato-wave-link` being an example of this, the correct latest version is `1.8.0.448` as per the previously open PR.
> 
> Here's the ugly query again for reference...
> 
> ```
> https://www.elgato.com/graphql?query=query%20contentJson(%24identifier%3A%5BString%5D%24contentType%3AString%24options%3AContentJsonOptionsInput)%7BcontentJson(identifiers%3A%24identifier%20contentType%3A%24contentType%20options%3A%24options)%7Bidentifier%20entries%7D%7D&operationName=contentJson&variables=%7B%22contentType%22%3A%22downloads%22%2C%22identifier%22%3A%5B%22downloads%22%5D%2C%22options%22%3A%7B%22level%22%3A1%7D%7D&locale=en-US
> ```

I am aware, but the `livecheck` for many of these were failing since `https://www.elgato.com/sites/default/files/downloads.json` is 404'ing.

So let's merge this as-is to update the rest, and I suspect the download link for `elgato-wave-link` will be updated in due course.

bevanjkay:
The query URL is being used to serve the URLs for all of the casks that are returned by the API. So it is possible that the versions may progress for all casks beyond the ones that are apart of the static website build.

bevanjkay (approved):

PR #149832, opened 2023-06-25
Housekeeping: fix found zap and uninstall errors

krehel:

Updates an uninstall-delete and a zap stanza which contained typos for the word "Application(s)" which would cause removal of files to not work as expected.

### DIFF
--- a/Casks/blender-benchmark.rb
+++ b/Casks/blender-benchmark.rb
@@ -19,6 +19,6 @@ cask "blender-benchmark" do
 
   zap trash: [
     "~/Library/Caches/blender-benchmark-launcher",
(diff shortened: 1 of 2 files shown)


reitermarkus (approved):

reitermarkus:
@krehel, this isn't syntax-only. It changes how `uninstall`/`zap` works, so should definitely run CI.

krehel:
Thanks @reitermarkus. Wasn't sure, that's why I broke these out to their own commit away from other changes.

Apologies on that.

reitermarkus:
No worries. `ci-syntax-only` should only ever be applied to bulk changes, as the label description suggests, which is mostly only when a new RuboCop style change needs to be done.

PR #150034, opened 2023-06-28
Update and rename ossia-score to score

miccal:
The original name did not follow our Token Reference.

### DIFF
--- a/Casks/score.rb
+++ b/Casks/score.rb
@@ -1,4 +1,4 @@
-cask "ossia-score" do
+cask "score" do
   version "3.1.11"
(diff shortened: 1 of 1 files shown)


Casks/score.rb
```diff
@@ -8,10 +8,7 @@
   desc "Interactive sequencer for intermedia art"
   homepage "https://ossia.io/"
 
-  livecheck do
-    url :url
-    strategy :github_latest
-  end
```
bevanjkay, line 14: @miccal Upstream uses/has used GitHub for alpha/pre-releases so the strategy will need to stay

bevanjkay (approved):

bevanjkay:
Thanks @miccal

PR #150346, opened 2023-07-04
mpc: new cask

roolo:

- [ ] `brew audit --cask --online mpc-software` is error-free.
  - Audit reported error, which I wasn't able to fix / debug / understand. Pls throw a link at me which explains this
    ```
       $ brew audit --cask --online mpc-software
      ==> Downloading https://cdn.inmusicbrands.com/akai/MPC2/2121_sufbchdj/Install-MPC-v2.12-2.12.1.35-release-Mac.zip
      Already downloaded: /Users/mailo/Library/Caches/Homebrew/downloads/5383e709546f93c1b5442669070fb55fc4d770f460142b566d6f0e01aae83ff0--Install-MPC-v2.12-2.12.1.35-release-Mac.zip
      homebrew/core
        * audit_exceptions/versioned_dependencies_conflicts_allowlist.json references
          formulae or casks that are not found in the homebrew/core tap.
          Invalid formulae or casks: predictionio
        * pypi_formula_mappings.json references
          formulae or casks that are not found in the homebrew/core tap.
          Invalid formulae or casks: rtv
      Error: 2 problems in 1 tap detected.
    ```
- [x] `brew style --fix mpc-software` reports no offenses.

- [ ] `brew audit --new-cask mpc-software` worked successfully.
- [ ] `brew install --cask mpc-software` worked successfully.
- [ ] `brew uninstall --cask mpc-software` worked successfully.


Casks/mpc-software.rb
```diff
@@ -0,0 +1,22 @@
+cask "mpc-software" do
+  version "2.12.1.35"
+  sha256 "4619858980c148d31038fc9e352ad7b08b524c646242dfc56dd543914c2a56da"
+
+  url "https://cdn.inmusicbrands.com/akai/MPC2/2121_sufbchdj/Install-MPC-v2.12-#{version}-release-Mac.zip",
```
razvanazamfirei, line 5: Is there a difference between this url and
`https://cdn.inmusicbrands.com/akai/MPC/2121_kdhebxog/MPC%202.12.1%20Updater.app.zip`?

Casks/mpc.rb
```diff
@@ -0,0 +1,48 @@
+cask "mpc" do
+  version "2.12.1.35"
+  sha256 "4619858980c148d31038fc9e352ad7b08b524c646242dfc56dd543914c2a56da"
+
+  url "https://cdn.inmusicbrands.com/akai/MPC2/#{version.major_minor_patch.no_dots}_sufbchdj/Install-MPC-v#{version.major_minor}-#{version}-release-Mac.zip",
```
razvanazamfirei, line 5: Is there a difference between this url and
`https://cdn.inmusicbrands.com/akai/MPC/2121_kdhebxog/MPC%202.12.1%20Updater.app.zip`?

Casks/mpc.rb
```diff
@@ -0,0 +1,48 @@
+cask "mpc" do
```
razvanazamfirei, line 1: The app installed is called MPC. I think this should be the cask name.

Casks/mpc.rb
```diff
@@ -0,0 +1,48 @@
...
+    "com.airmusictechnology.hybrid.content",
+    "com.akaipro.mpc.aax.pkg",
+    "com.akaipro.mpc.appsupport.pkg",
+    "com.akaipro.mpc.audiounit.pkg",
+    "com.akaipro.mpc.standalone.pkg",
+    "com.akaipro.mpc.vst.pkg",
+  ]
+
+  zap delete:    [
+        "/usr/local/bin/iloktool",
```
bevanjkay, line 28: I'm not sure about including the `ilok` related items, even in zap. As it is likely that they are relevant to other applications, unless the bundle ID change between ilok-locked software. I'm familiar with ilok, but not how it installs.

Casks/mpc.rb
razvanazamfirei, line 5: Slightly concerned that `#{version.major_minor_patch.no_dots}_sufbchdj` is not stable. If that becomes a problem, we could capture it in the regex (what is now non-capturing group 1.

```
regex(%r{href=.*?/MPC2/\d+[._-](?:\w+)/Install[._-]MPC[._-]v?[\d.]+[_-](\d+(?:\.\d+)+)[_-]release[_-]Mac\.zip"}i)`
```

Casks/mpc.rb
bevanjkay: Further to this, is [`ilok-license-manager`](https://github.com/homebrew/homebrew-cask/tree/master/Casks/ilok-license-manager.rb) bundled with this software, or is it separate?

Even if it's bundled I don't think we should zap anything anyway (you can see that the files are relevant to the `ilok` package independently from it's own cask) - as it would break other software that uses `ilok` for licensing.

Casks/mpc.rb
```diff
@@ -0,0 +1,48 @@
...
+  uninstall pkgutil: [
+    "com.airmusictechnology.hybrid.content",
+    "com.akaipro.mpc.aax.pkg",
+    "com.akaipro.mpc.appsupport.pkg",
+    "com.akaipro.mpc.audiounit.pkg",
+    "com.akaipro.mpc.standalone.pkg",
+    "com.akaipro.mpc.vst.pkg",
+  ]
+
+  zap delete:    [
```
razvanazamfirei, line 27: There are probably more things that should be included in the zap that I can't find since it requires a license. 

Of note, the cask also installs `ilok`, which is why all relevant files are included in zap.

razvanazamfirei (changes requested):

Casks/mpc.rb
bevanjkay: These also shouldn't be included in the `uninstall` stanza to satisfy CI, as they will break other applcation.

razvanazamfirei: Didn't realize you already commented! :) 

It's bundled with the software and I see the concerns. Ideally, we could notify users in the caveates that they may have to run 'brew uninstall ilok-license-manager --zap --force`. I don't think caveats show up on uninstall, though. Perhaps in a postflight block?

bevanjkay: Yeah, that could be a good workaround, can we use `ohai` or similar in a `postflight_uninstall`?
It's a fairly unprecedented case. But in terms of CI we do use `ci-skip-install` at the moment for `google-drive` because it installs common software.

Casks/mpc.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  depends_on macos: ">= :catalina"
+
+  pkg "MPC-Installer-#{version}/Install MPC #{version}.app/Contents/Resources/InstallApp.pkg"
+
+  uninstall_postflight do
+    oh1 "Caveats"
+    print <<~EOS
+      #{token} installs ilok-license-manager for license management.
+      Please note that other casks may depend on it and it will not be
```
bevanjkay, line 24: @razvanazamfirei I think it might be better to say other "applications" or similar as it could affect more than just casks.

razvanazamfirei: Done.

Casks/mpc.rb
razvanazamfirei: @roolo, could you address this?

PR #150535, opened 2023-07-07
viscosity 1.10.6: update sha256

krehel:

This cask seems to have updated its `sha256` again.  See #150503.  The `sha256` being updated is reflected directly from the applications release notes [here](https://www.sparklabs.com/viscosity/releasenotes/mac).

If this continues to change, we'll have to make further amends.


krehel:
Yeah, this is certainly a possibility.  I do see where this has occurred a couple times over the past several years.

I was chalking this up to a release issue caused an in place update.

I think I'm ok leaving it for now since this is the documented `sha256` and if it changes again, we'll remove the check.

razvanazamfirei (approved):

PR #151456, opened 2023-07-21
tencent-meeting 3.18.5.455

stevenlele:
Closes #151359

Used a different endpoint for livecheck.

This endpoint returns no information if version is latest, where the script returns the current version and CDN hash. If this behavior is not favored, a similar workaround to #145346 (pretty hacky; that one had to use this because that endpoint returns HTTP 204) is to use `version.major_minor_patch` in the livecheck URL, which ensures the API returns something (as long as they don't trim the revision code). Don't use any lower version as that might not return the latest update.

Also updated zap trash stanza.

---


samford (commented):
I haven't looked into this myself but I can at least confirm that this works as expected when I tested it. I've pushed a commit that reworks the `strategy` block to achieve the same goal in a less verbose fashion. I'm good with the `livecheck` change but I'll leave general cask review (version update, zap changes) to our cask maintainers.

miccal (approved):

PR #151614, opened 2023-07-24
Housekeeping: various qlplugin updates (add no zap, add desc)

krehel:
Doing some housekeeping, and found there are a good number of `casks` that are `qlplugins` without a `zap` stanza and do not need one.

This submission is only for casks that can be marked `no zap stanza required` and are a `qlplugin` to help get them off the `missing zap` list.

Some of these due to their age may also not have a `desc` block, so that will be added as needed.


razvanazamfirei (dismissed):

miccal (dismissed):

krehel:
That should be most of the QuickLook plugins.  I have a couple removals which will be submitted as separate PR's.

miccal (approved):

PR #152176, opened 2023-08-02
Stats requires MacOS 10.15 starting with v2.9

decadent:

Hopefully, changing the OSX version doesn't trigger errors.

### DIFF
--- a/Casks/stats.rb
+++ b/Casks/stats.rb
@@ -18,7 +18,6 @@ cask "stats" do
   homepage "https://github.com/exelban/stats"
 
   auto_updates true
-
   depends_on macos: ">= :mojave"
 
   app "Stats.app"
--- a/Casks/stats.rb
+++ b/Casks/stats.rb
@@ -1,6 +1,16 @@
 cask "stats" do
-  version "2.9.3"
-  sha256 "b70d0b9caa09c7fe8eeadfe8022a40af87f32eec662401f002af14ecc979f8d0"
+  on_mojave :or_older do
+    version "2.8.26"
+    sha256 "1a4b44ba02520683b0a6c192388f593c36dde4d15c784a22dccf0caefe81e8b7"
+
+    livecheck do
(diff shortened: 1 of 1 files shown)


krehel:
Thanks for the submission @decadent!

Are you able to update the Caskfile to include Mojave support on the legacy version?  If not, let me know and I can help make the updates.

Thanks!

decadent:
@krehel Honestly, I just have no idea what that entails, as I'm basically a passerby in this business. Do I copy the file for the previous version into `homebrew-cask-versions`? It sounds like you're proposing adding something in the current file. If there's some document or an existing cask that I could follow, perhaps I could do that. I would come in handy since I have a few more casks that likewise don't work on old versions of OSX, so I could update those as well.

krehel:
No worries @decadent!  I pushed the update I was referring to.  (Let's hope I got it right)

You can refer to the [Cask Cookbook](https://docs.brew.sh/Cask-Cookbook) for more information on how to contribute and regarding the Cask DSL.

decadent:
@krehel I've seen the cookbook, but alas, it misses precisely the directives that you added to the file.

I have to ask: with your changes, is there any warning when installing the cask that the newest app version doesn't work on the older OSX and is thus not installed? Perhaps a `caveats` should be added under `on_mojave`?

krehel:
@decadent - no it won't.  It will install that specific version which is listed in the upstream releases as the last version to support Mojave.

`caveats` are used sparingly to convey important information, and sometimes is used to mention an app is `discontinued` from the upstream provider as a whole, but not that it is not the most recent version.

I (personally) think it's a reasonable expectation to allow the install of older software versions as long as the binaries exist and the effort is not cumbersome to provide them.  But, broadly speaking, understanding the versions of software being used and being aware of EOL cycles should be up to the end user.  We'll simply provide the means.

miccal (approved):

PR #152440, opened 2023-08-07
o*: move to o/ subdirectory.

MikeMcQuaid:
Continuing the sharding work now that `orka.rb` was moved without issue.


MikeMcQuaid:
@p-linnane typing is hard, single characters even moreso

(will fix)

p-linnane (approved):

PR #153307, opened 2023-08-19
Enforce sha256 on Packages

denizt:
Currently _Packages_ is hosted on free.fr (HTTP) and no shasum checking is enforced. Add the sha256 for the current version, to add integrity verification.

I also got in touch with the developer to move it to secure host with versioning.

### DIFF
--- a/Casks/p/packages.rb
+++ b/Casks/p/packages.rb
@@ -1,6 +1,7 @@
 cask "packages" do
   version "1.2.10"
-  sha256 :no_check
+  # required as the host has HTTP (insecure) transport only, see PR for details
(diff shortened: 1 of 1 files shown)


krehel:
Hi @denizt - thank you for the submission!

No `sha256` is enforced because the URL is unversioned, not due to being HTTP.  Accepting a `sha256` on an unversioned URL would cause the download to break at every update.

If the download can be moved to a versioned URL, we would be more than happy to accept a `sha256` change.

Thank you!

bevanjkay:
@denizt `homebrew-cask` only serves binaries/packages from upstream (downloaded directly from source), we don't have the ability to build and host.

PR #153675, opened 2023-08-24
1password-cli v2.20.0

AndyTitu:
This PR was created automatically.


krehel:
@AndyTitu - can you please update your automated jobs to support the now sharded repository?  Your file should be pushed to `Casks/1/1password-cli.rb` and not the root of `Casks/`

Thanks!

PR #153774, opened 2023-08-27
viso 6,8.9.1

Ninzero:

-------

I have two questions about this pull request:

1. I am not sure what is the right format of `version` stanza. This app have different short version and version, It's version is "8.9.1" but short version is "6". The about page of this app says "Viso v6 (8.9.1)". I chose "8.9.1" to be the version, but I am not sure if this is the right decision.
2. When I am testing this cask with `brew install --cask viso`. It tells me that cask viso does not exist. I suspect this is because Homebrew only use the .json file to seach for casks. How to tell Homebrew to check local repository so I can install this cask for testing?

### DIFF
--- a/Casks/v/viso.rb
+++ b/Casks/v/viso.rb
@@ -17,8 +17,6 @@ cask "viso" do
   app "Viso.app"
 
   zap trash: [
-    "/Library/Preferences/com.apple.P2PSupervisor.plist",
-    "~/Desktop/viso_appcast.xml",
     "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.ephod.viso.sfl2",
     "~/Library/Caches/com.ephod.Viso",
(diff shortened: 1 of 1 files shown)


Casks/v/viso.rb
```diff
@@ -0,0 +1,28 @@
...
+    url "https://getviso.app/files/visoappcast.xml"
+    strategy :sparkle, &:version
+  end
+
+  auto_updates true
+
+  app "Viso.app"
+
+  zap trash: [
+    "/Library/Preferences/com.apple.P2PSupervisor.plist",
```
bevanjkay, line 20: This doesn't look correct

bevanjkay (changes requested):

Casks/v/viso.rb
Ninzero: You are right. The zap part is generated by createzap, forgot to check it.

bevanjkay (approved):

PR #154279, opened 2023-09-03
Add clickhouse-client 23.8.1.2992-lts

orf:

### DIFF
--- a/Casks/c/clickhouse.rb
+++ b/Casks/c/clickhouse.rb
@@ -1,0 +1,23 @@
+cask "clickhouse" do
+  arch arm: "-aarch64"
+
+  version "23.8.1.2992-lts"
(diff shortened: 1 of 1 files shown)


razvanazamfirei (approved):

bevanjkay:
@orf Is there a particular reason that we require the `:github_releases` strategy here, we should use it sparingly as it utilises the Github API and can cause rate limit issues. Does the standard `:git` strategy not work in this case?

orf:
> @orf Is there a particular reason that we require the `:github_releases` strategy here, we should use it sparingly as it utilises the Github API and can cause rate limit issues. Does the standard `:git` strategy not work in this case?

Yeah I think that would work, this is my first cask so I wasn’t sure what strategy to use.

PR #154503, opened 2023-09-06
calhash: use static URLs for each system block

EricFromCanada:
Match style of other titanium-software.fr casks - see https://github.com/Homebrew/brew/pull/15956#issuecomment-1706959261

Also adds extra system blocks so that each macOS release gets either a URL or a `depends_on` stanza - see https://github.com/Homebrew/brew/pull/15956#issuecomment-1707579462

### DIFF
--- a/Casks/c/calhash.rb
+++ b/Casks/c/calhash.rb
@@ -1,30 +1,38 @@
 cask "calhash" do
+  sha256 :no_check
+
+  on_mojave :or_older do
+    depends_on macos: ">= :catalina"
+  end
   on_catalina do
     version "1.0.5"
-    sha256 "bb15c18839837bdeda203a8a916b09b4f2b971e5bba10b603f290130939d3bf7"
+
+    url "https://www.titanium-software.fr/download/1015/CalHash.dmg"
 
     depends_on macos: :catalina
   end
   on_big_sur do
     version "1.1.1"
-    sha256 "66b2de6f514119b3ddf915310b94486d24d3cc6733c74af97e367b76cb2d7478"
+
+    url "https://www.titanium-software.fr/download/11/CalHash.dmg"
 
     depends_on macos: :big_sur
   end
   on_monterey do
     version "1.1.9"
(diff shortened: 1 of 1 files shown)


bevanjkay:
@EricFromCanada Do we need the `depends_on` block inside of each `on_{system}` block here?
And we should be able to put the `depends_on macos: "<= :ventura"` inside of an `on_ventura :or_newer do` block?

EricFromCanada:
> @EricFromCanada Do we need the `depends_on` block inside of each `on_{system}` block here? 

This is a pattern followed by other apps from the same provider (see `deeper`, `maintenance`, and `onyx`). My assumption is that this ensures that each mentioned version runs only on the system release it's associated with, say, to ensure it's marked incompatible and gets upgraded if a user upgrades their system. Is that actually necessary compared to e.g. a list of allowed system versions (see `horndis`)? I don't know. Waiting for @apainintheneck to chime in.

> And we should be able to put the `depends_on macos: "<= :ventura"` inside of an `on_ventura :or_newer do` block?

Looks like that's what the other casks do, so sure.

bevanjkay:
> My assumption is that this ensures that each mentioned version runs only on the system release it's associated with
I guess that make sense, but I would have assumed the `on_{system}` part takes care of that.

EricFromCanada:
These are the related PRs that add the individual `depends_on` lines: https://github.com/Homebrew/homebrew-cask/pull/138876, https://github.com/Homebrew/homebrew-cask/pull/137513, https://github.com/Homebrew/homebrew-cask/pull/142380 (cc @p-linnane @jlec)

miccal (approved):

Bo98:
iirc there was something about you not being able to specify `depends_on` twice and those formulae basically needed a `depends_on macos: ">= :lower && <= :upper"` type restriction

EricFromCanada:
It is possible to specify a list of allowed macOS releases: https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/h/horndis.rb#L10-L15

Bo98:
Sounds like we could maybe do that instead then

PR #154556, opened 2023-09-06
calhash, deeper, maintenance, onyx: use array for allowed macOS releases

EricFromCanada:
An improved take on the issue addressed in #154503, relying on brew changes made in Homebrew/brew#15977. Applies the same changes to all apps from the same vendor, using [syntax](https://docs.brew.sh/Cask-Cookbook#requiring-an-exact-macos-release) that [wasn't originally available](https://github.com/Homebrew/homebrew-cask/pull/154503#issuecomment-1707621049) to define an exact list of allowed macOS releases. Each retains the Ventura block's `:or_newer` to avoid issues with CI once macOS Sonoma testing starts, or once `brew audit` is updated to [fully support SimulateSystem](https://github.com/Homebrew/brew/pull/15977).

### DIFF
--- a/Casks/c/calhash.rb
+++ b/Casks/c/calhash.rb
@@ -1,36 +1,25 @@
 cask "calhash" do
   sha256 :no_check
 
-  on_mojave :or_older do
-    depends_on macos: ">= :catalina"
-  end
   on_catalina do
     version "1.0.5"
 
     url "https://www.titanium-software.fr/download/1015/CalHash.dmg"
-
-    depends_on macos: :catalina
   end
   on_big_sur do
     version "1.1.1"
 
     url "https://www.titanium-software.fr/download/11/CalHash.dmg"
-
-    depends_on macos: :big_sur
   end
   on_monterey do
     version "1.1.9"
 
     url "https://www.titanium-software.fr/download/12/CalHash.dmg"
-
-    depends_on macos: :monterey
   end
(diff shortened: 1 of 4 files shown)


Casks/c/calhash.rb
```diff
@@ -1,38 +1,22 @@
...
   on_ventura :or_newer do
     version "1.2.1"
-
-    url "https://www.titanium-software.fr/download/13/CalHash.dmg"
-
-    depends_on macos: "<= :ventura"
+    sha256 "5372c281ab6edbd10f4a6a237a7e9afb529222b62f39c3b0357d779672cbaa6d"
   end
 
+  url "https://www.titanium-software.fr/download/#{MacOS.version.to_s.delete(".")}/CalHash.dmg"
```
bevanjkay, line 19: @EricFromCanada Is `MacOS.version` now available when using SimulateSystem?

apainintheneck: `MacOS.version` always returns the actual OS version and ignores `SimulateSystem` completely right now.

p-linnane (dismissed):

Casks/c/calhash.rb
bevanjkay: It was my understanding that we shouldn't be using `MacOS.version` anymore because it cannot be simulated where necessary now or in future.

miccal (approved):

PR #154571, opened 2023-09-06
huiontablet: remove

chenrui333:
download page got changed somewhere between May 27[^1] and June 2[^2], and there is no driver download page for tablet anymore.

[^1]: https://web.archive.org/web/20230527002507/https://www.huion.com/download/
[^2]: https://web.archive.org/web/20230602132708/https://www.huion.com/download/


razvanazamfirei (approved):

sxn:
I noticed the page has been updated, but it looks like the download links have maintained the same format. Is there another reason behind why this would be removed, that I'm missing? 🤔 

image

chenrui333:
@sxn that looks like a good find, thanks! Feel free to create a new PR to bring it life.

PR #153968, opened 2023-08-29
Rename Remotion to Multiapp

rtharston:
We've renamed Remotion to Multi, but multi is already taken so I used multiapp, referring to our website Multi.app.

I removed the Remotion cask because it points to an old version we no longer support and that cask can't be updated anymore. If there is a preferred way to deprecate casks, please let me know.

Verified:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask --online <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- **NOTE: The token rules say to remove app but there is already a cask named Multi, so I used 'multiapp' because our website is multi.app.** 
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [x] `brew audit --new-cask <cask>` worked successfully.
- [x] `brew install --cask <cask>` worked successfully.
- [x] `brew uninstall --cask <cask>` worked successfully.

PR #154585, opened 2023-09-07
Replace Remotion with Multiapp in autobump

rtharston:
When I renamed Remotion to Multiapp in #153968 I didn't know about the auto bump script. (Looking back I should have done a search for Remotion in the repo. 🙄)


bevanjkay (changes requested):
Can you please keep the list sorted alphabetically?

PR #154652, opened 2023-09-07
updated latest box-drive version number

s0wx:
in case sha256 checksum is also required (but only one version exists covering both, Intel and M1):

`shasum -a 256 BoxDrive.pkg`
`561f976d3e97f966f5cbd24ee69da2b31b995cd8bcb24e2c5b09b126196f135b`  BoxDrive.pkg

---

### DIFF
--- a/Casks/b/box-drive.rb
+++ b/Casks/b/box-drive.rb
@@ -1,5 +1,5 @@
 cask "box-drive" do
-  version "2.34.84"
+  version "2.35.95"
   sha256 :no_check
 
   url "https://e3.boxcdn.net/desktop/releases/mac/BoxDrive.pkg",


bevanjkay (approved):

s0wx:
@bevanjkay @miccal as the `brew audit --cask box-drive` job in the CI test part always seem to fail, would it resolve the issue if the sha256 checksum would be added to the file as well?

Just in case, if the `livecheck` of the job uses some kind of _"if no checksum is defined, compare version number"_-checking. Which would always fail, as I only updated the version number. Causing an infinite error if the the job uses the logic I just described. 

I only didn't add the checksum value (first comment at the top) to the commit, as I'm not familiar with the brew behaviour if a checksum is added if none was stored before.
If I can somehow assist you resolving this issue, please let me know.

bevanjkay:
@s0wx The livecheck is failing because the version inside the package file being downloaded does not match the version in your updated cask file.

razvanazamfirei:
Closing for now given no progress.

PR #154801, opened 2023-09-10
koodo-reader: add livecheck

chenrui333:


bevanjkay:
@chenrui333 We should actually be getting the livecheck from the `homepage`. There is a note even on version `1.5.7` to use the homepage as it is a "developer release".

miccal (approved):

PR #155138, opened 2023-09-14
sogouinput: remove

miccal:
As found in https://github.com/Homebrew/homebrew-cask/pull/154321, the `url` is now dynamic and as such is not suitable as a Cask.


krehel (approved):

chenrui333:
👍

gibfahn:
This is sad, hopefully we can find a URL that doesn't have an opaque hash in it to download this from.

JounQin:
Hi, I just wrote a tiny service https://github.com/un-ts/rabbit-linker to resolve the random url issue, shall we use it?

The followings version and short version are both fine.

https://rabbit-linker.vercel.app/gtimg/sogou_mac/1688992049
https://rabbit-linker.vercel.app/gtimg/sogou_mac/614d
https://rabbit-linker.vercel.app/gtimg/sogou_mac

I'm currently maintaining `sogouinput` cask at https://github.com/homebrew-x/homebrew-x/blob/master/Casks/sogouinput.rb#L5

miccal:
Great work @JounQin, but unfortunately we cannot use an external `url` resolver in Homebrew.

Thank you for maintaining a tap, though :)

gibfahn:
@JounQin could we do the link rewriting locally in the formula definition without using the service? I don't know if homebrew would accept it, but it would avoid using an external service.

JounQin:
@gibfahn Sure. I had a similar proposal at https://github.com/Homebrew/homebrew-cask/issues/107354, but I had limited ruby knowledge so I rejected my PR.

PR #155313, opened 2023-09-15
localdev: remove cask

krehel:
Cask was previously marked as discontinued and is no longer supported upstream.

Download links are no longer functional.


chenrui333:
image

chenrui333 (approved):

PR #156230, opened 2023-09-26
revert anka-virtualization to 3.3.4

NorseGaud:
Need to revert to 3.3.4


krehel:
Hi @NorseGaud - do you know why this version was pulled?  I went looking and couldn't find evidence, but I don't use this software.

Thanks!

NorseGaud:
Hi @krehel, it has several bugs that are harmful to production environments that we didn't catch in our automated testing. Our customers have rolled back, but now they're complaining about the homebrew cask 404ing :)

krehel (approved):

krehel:
Thanks @NorseGaud!  Just good to know why since it's a revert.  Totally understood!

PR #156395, opened 2023-09-28
Update AlDente sources and improve uninstall routine

PascalOchel:
The developers do not use the davidwernhart identifiers any longer. The github repo _davidwernhart/AlDente_ does not exist anymore and points to _AppHouseKitchen/AlDente-Charge-Limiter_.

The paid features are of course not available on github. The now used sources from the developers' website include the free version as well as the premium features unlock-able by license key.

Additionally I improved the uninstall routine a bit. Some preference files have not been removed and the quit command ran into empty space.

### DIFF
--- a/Casks/a/aldente.rb
+++ b/Casks/a/aldente.rb
@@ -1,15 +1,15 @@
 cask "aldente" do
   version "1.22.2"
-  sha256 :no_check
+  sha256 "6d1108d114b8c6f11c732c37c9e779683f281b9c260f98721bcdaa5b6b6a9440"
 
-  url "https://apphousekitchen.com/aldente/AlDente-Pro.dmg"
+  url "https://apphousekitchen.com/aldente/AlDente#{version}.dmg"
   name "AlDente"
   desc "Menu bar tool to limit maximum charging percentage"
   homepage "https://apphousekitchen.com/"
 
   livecheck do
-    url "https://github.com/AppHouseKitchen/AlDente-Charge-Limiter"
-    strategy :github_latest
+    url "https://apphousekitchen.com/aldente/aldenteproappcast.xml"
+    strategy :sparkle, &:short_version
   end
 
   auto_updates true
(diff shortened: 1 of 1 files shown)


Casks/a/aldente.rb
```diff
@@ -1,14 +1,14 @@
 cask "aldente" do
   version "1.22.2"
-  sha256 "6d1108d114b8c6f11c732c37c9e779683f281b9c260f98721bcdaa5b6b6a9440"
+  sha256 :no_check
 
-  url "https://github.com/davidwernhart/AlDente/releases/download/#{version}/AlDente.dmg"
+  url "https://apphousekitchen.com/aldente/AlDente-Pro.dmg"
```
bevanjkay, line 5: @PascalOchel It is likely better to use the GitHub source as we can keep a versioned `url` and allow a shasum check.
`https://github.com/AppHouseKitchen/AlDente-Charge-Limiter/releases/download/1.22.2/AlDente.dmg`

PascalOchel: In my configuration, I now only use github to retrieve the current version number. I see the advantages of downloading the software from github too. However, this option has a significant disadvantage.

The developers provide basic features for free and offer advanced features with paid license. **The paid functionality is not made public by the developers, it is simply not included in the public github repo** - neither in the source code nor in the executable.

**I bought the pro license, so the software from github in the existing cask does not work for me.** The developers therefore provide a separate compiled pro version on their website for download, which I reference. This pro version is suitable for all users. There is no compulsion to enter a license key there. Users who have not purchased a license can simply use the Pro version with the basic features.

An alternative solution to the problem could be to keep the existing cask _aldente_ and create a new cask _aldente-pro._ I don't see any advantage in this, because the _aldente-pro_ cask would be suitable for both license types anyway. Why would you need the limited cask _aldente_ then?

miccal:
Thank you @PascalOchel.

miccal (approved):

PR #156519, opened 2023-09-29
virtualbuddy: update livecheck

samford:

-----

This is a follow-up to #156512, to replace the older GitHub asset-checking approach that we used before the `GithubLatest` strategy was modified to use the JSON API (I wasn't able to catch that PR before it was merged). Nowadays, `GithubLatest` can pass the release JSON data to a `strategy` block, so we can easily iterate through the assets [without having to make multiple requests].

Besides the refactoring related to the strategy, this simplifies the regex and modifies how we handle the match, so it will continue to return a version if there is ever a dmg file that doesn't include a numeric suffix (e.g., `1.2.3` instead of the usual `1.2.3-4`).

### DIFF
--- a/Casks/v/virtualbuddy.rb
+++ b/Casks/v/virtualbuddy.rb
@@ -8,20 +8,15 @@ cask "virtualbuddy" do
   homepage "https://github.com/insidegui/VirtualBuddy"
 
   livecheck do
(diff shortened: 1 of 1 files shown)


p-linnane (approved):
Thank you @samford!

PR #156889, opened 2023-10-04
Dev Tunnel: New Cask

setaskin:

### DIFF
--- a/Casks/d/devtunnel.rb
+++ b/Casks/d/devtunnel.rb
@@ -3,7 +3,7 @@ cask "devtunnel" do
 
   version "1.0.922+fa34179138"
   sha256 arm:   "3d81eaa2bd1d5be99489ac378c2925f2076f01d1a505044b0e51902ce052a868",
-         intel: "d9875469d96c563302cb68d1bd6834bde6ba949ace57714a86ecd0108073da1e"
+         intel: "c48f72ec265cf152929143672ccc090ff9739b925bbf56e54621ea55657e8f54"
   
   url "https://tunnelsassetsprod.blob.core.windows.net/cli/#{version}/osx-#{arch}-devtunnel-zip",
     verified: "tunnelsassetsprod.blob.core.windows.net/cli/"
--- a/Casks/d/devtunnel.rb
+++ b/Casks/d/devtunnel.rb
@@ -7,9 +7,11 @@ cask "devtunnel" do
 
   url "https://tunnelsassetsprod.blob.core.windows.net/cli/#{version}/osx-#{arch}-devtunnel-zip",
       verified: "tunnelsassetsprod.blob.core.windows.net/cli/"
-  name "Microsoft devtunnel"
-  desc "Provides developers to secure tunnels"
+  name "Microsoft Dev Tunnels"
+  desc "Provides developers secure tunnels to share local services"
   homepage "https://aka.ms/devtunnels/docs"
 
   binary "devtunnel"
+
+  # No zap stanza required
 end
--- a/Casks/d/devtunnel.rb
+++ b/Casks/d/devtunnel.rb
@@ -1,11 +1,10 @@
 cask "devtunnel" do
   arch arm: "arm64", intel: "x64"
 
-  version "1.0.922+fa34179138"
-  sha256 arm:   "065e86312e3513aa2a5bd6ea2f0822ca3a50ad6981e497043ded6b1630044737",
-         intel: "d9875469d96c563302cb68d1bd6834bde6ba949ace57714a86ecd0108073da1e"
+  version :latest
+  sha256 :no_check
 
-  url "https://tunnelsassetsprod.blob.core.windows.net/cli/#{version}/osx-#{arch}-devtunnel-zip",
+  url "https://tunnelsassetsprod.blob.core.windows.net/cli/osx-#{arch}-devtunnel-zip",
       verified: "tunnelsassetsprod.blob.core.windows.net/cli/"
   name "Microsoft Dev Tunnels"
   desc "Provides developers secure tunnels to share local services"
--- a/Casks/d/devtunnel.rb
+++ b/Casks/d/devtunnel.rb
@@ -2,7 +2,7 @@ cask "devtunnel" do
   arch arm: "arm64", intel: "x64"
 
   version "1.0.922+fa34179138"
-  sha256 arm:   "3d81eaa2bd1d5be99489ac378c2925f2076f01d1a505044b0e51902ce052a868",
+  sha256 arm:   "065e86312e3513aa2a5bd6ea2f0822ca3a50ad6981e497043ded6b1630044737",
          intel: "d9875469d96c563302cb68d1bd6834bde6ba949ace57714a86ecd0108073da1e"
 
   url "https://tunnelsassetsprod.blob.core.windows.net/cli/#{version}/osx-#{arch}-devtunnel-zip",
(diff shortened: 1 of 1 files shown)


krehel:
Hi @setaskin! 👋 

Thanks for opening this PR!  I have some comments I'll post in-line and we can get this moving along. 😄 

One thing I did notice is this software is in a public preview.  We try to accept only software considered as stable.  Mentioning for others who may review and have their own commentary.

Casks/d/devtunnel.rb
```diff
@@ -4,12 +4,12 @@
   version "1.0.922+fa34179138"
```
krehel, line 4: I think this may be better as `version,commit_hash` so we can use `#{version.csv.first}+#{version.csv.second}` in the URL.  This may also help with the `livecheck`

Casks/d/devtunnel.rb
```diff
@@ -0,0 +1,15 @@
+cask "devtunnel" do
+  arch arm: "arm64", intel: "x64"
+
+  version "1.0.922+fa34179138"
+  sha256 arm:   "3d81eaa2bd1d5be99489ac378c2925f2076f01d1a505044b0e51902ce052a868",
+         intel: "d9875469d96c563302cb68d1bd6834bde6ba949ace57714a86ecd0108073da1e"
+
+  url "https://tunnelsassetsprod.blob.core.windows.net/cli/#{version}/osx-#{arch}-devtunnel-zip",
+      verified: "tunnelsassetsprod.blob.core.windows.net/cli/"
+  name "Microsoft devtunnel"
```
krehel, line 10: Would this be better as `Microsoft Dev Tunnels` ?

Casks/d/devtunnel.rb
```diff
@@ -0,0 +1,15 @@
...
+  desc "Provides developers to secure tunnels"
```
krehel, line 11: Would a more apt description be `desc "Provides developers secure tunnels to share local services"` ?

Casks/d/devtunnel.rb
```diff
@@ -0,0 +1,15 @@
...
+  homepage "https://aka.ms/devtunnels/docs"
+
```
krehel, line 13: We need a `livecheck` here. Is there a good source of truth we can pull version information from? We can help you create it.

Casks/d/devtunnel.rb
```diff
@@ -0,0 +1,15 @@
...
+  binary "devtunnel"
+end
```
krehel, line 15: If this only installs a `binary`, we can add a comment stating no `zap` is needed.  Can you confirm this is `binary` only?

setaskin:
> Hi @setaskin! 👋
> 
> Thanks for opening this PR! I have some comments I'll post in-line and we can get this moving along. 😄
> 
> One thing I did notice is this software is in a public preview. We try to accept only software considered as stable. Mentioning for others who may review and have their own commentary.

Hi @krehel! thank for the prompt response. I will fix the parts you commented but I just want to discuss being public preview. This is an official Microsoft product. We  had been added to other package managers. We also want to be part of Homebrew as well for our Mac users. Can we still move forward with this Pr for insertion?

krehel:
@setaskin - I would agree we can move forward, given it's an official product.  I'll tag the rest of @Homebrew/cask for additional opinions.

It's always on a case by case basis, and at the end of the day part of it is we want to avoid user experience issues, both for the sake of the user and to avoid fielding spurious bug reports.

For some of the comments, feel free to take them as suggestions (especially those posed as questions, as you're in a better position than I am on some verbiage given your position).

Cheers!

Casks/d/devtunnel.rb
setaskin: I confirm it is binary only

krehel: Great!  Can you add a comment directly above the last `end` stating:

```# No zap stanza required```

Then we can skip the `Missing zap` check.  Thanks!

p-linnane:
This is fine to include.

Per [Acceptable Casks](https://docs.brew.sh/Acceptable-Casks#stable-versions):

> When software is only available as a beta, development, or unstable version, its cask can go in the main homebrew/cask repository. When stable versions become available, only those will be accepted as subsequent updates.

setaskin:
> This is fine to include.
> 
> Per [Acceptable Casks](https://docs.brew.sh/Acceptable-Casks#stable-versions):
> 
> > When software is only available as a beta, development, or unstable version, its cask can go in the main homebrew/cask repository. When stable versions become available, only those will be accepted as subsequent updates.

Hello @p-linnane , thank you for the response. I am glad we can move forward. I want to make sure I understand this quote you mentioned above correctly. So, when we check this Pr in,  until we become Ga product we cannot update the version?  is that correct? So the version in this Pr will be the only available one until GA ?

p-linnane:
It just means that once the product is GA we will not accept beta/development releases in this repo. They would need to go in Homebrew/homebrew-cask-versions.

Casks/d/devtunnel.rb
samford: I'm happy to add a suggestion for a `livecheck` block if there's a source for the version information (ideally something that links to the cask `url`). I took a cursory glance at the documentation, etc. and wasn't able to find anything.

setaskin: Hi,  @krehel , @samford, yes we don't actually have a mechanism to check the upstream I believe. In azure blob location we keep different versions in different folders and folder name is the version number. We do have a generic url which always points to the latest released version but that doesn't have version indicator.  I have been reading the docs but couldn't find what is best for us in this situation

setaskin: I also have been reading the docs for auto updating the casks but couldn't find what I am looking for. Do we have to manually send a new pr everytime we have a new version ?  or is there a automated way ?

bevanjkay: @setaskin If there is currently no way of checking for the latest version with `livecheck` then we would need to use the unversioned URL (with the downside of not being able to have a `sha256` value), coupled with `version :latest`.

https://docs.brew.sh/Cask-Cookbook#version-latest

In future, we could move to a versioned `url` when there is a reliable source for finding the latest version.
I did find https://global.rel.tunnels.api.visualstudio.com/api/version - but I believe this is different to the latest version of the cli tool itself?

setaskin: > @setaskin If there is currently no way of checking for the latest version with `livecheck` then we would need to use the unversioned `url` (with the downside of not being able to have a `sha256` value), coupled with [`version :latest`](https://docs.brew.sh/Cask-Cookbook#version-latest).
> 
> In future, we could move to a versioned `url` when there is a reliable source for finding the latest version. I did find https://global.rel.tunnels.api.visualstudio.com/api/version - but I believe this is different to the latest version of the cli tool itself?

Hi,  @bevanjkay , thank you for the response.  Yes the one you indicated above is not the cli version number. It is our service version number which is unrelated here. We were thinking to add a .txt file that has the latest version number in our blob location . Would it work?

krehel: @setaskin - you may want to look at some other Microsoft casks.  Based on what you are describing regarding the url, does it match a `version` scheme similar to `visual-studio` or `microsoft-edge`?

The `livecheck` will allow anyone to check versions and update the Cask file using the `brew` built tooling.

As @bevanjkay mentions, if we don't have a version number we can point to `:latest`, but we drop the `sha256` due to update in place.

bevanjkay: @setaskin Yes that would work perfectly.

setaskin: @krehel , thanks for all the help.  I checked bunch of other Microsoft products before I submit  but their versioning is quite different than ours.  So I will move forward with generic url for now with no version number.  Once we have the version checkable url  I will update the cask for versioned urls and sha number

miccal:
Thank you @setaskin.

miccal (approved):

PR #156962, opened 2023-10-05
languagetool 1.6.0

piijai:
bump-cask languagetool to 1.6.0

audit fails, because livecheck receives version 1.6 instead of 1.6.0. Is there a way to solve this?


krehel:
Looks like version should be `1.6`.  Application `Info.plist` contains the same.

```
<key>CFBundleName</key>
<string>LanguageTool for Desktop</string>
<key>CFBundlePackageType</key>
<string>APPL</string>
<key>CFBundleShortVersionString</key>
<string>1.6</string>
```

samford:
> should livecheck be flagging this as not equivalent versions? Locally, I made the change and it didn't seem to complain when running livecheck, (i.e, doesn't show as version upgrade or downgrade) but fails audit the same as here.

@krehel livecheck does version comparison by creating `Version` objects (from version strings). 1.6 and 1.6.0 being equal is part of how `Version` comparison works (https://github.com/Homebrew/brew/blob/7456789fb8319b4b9a838a480f6a4fc53f0e8fce/Library/Homebrew/version.rb#L117), so this is expected behavior in livecheck.

`Cask::Audit#audit_livecheck_version` does string comparison instead (https://github.com/Homebrew/brew/blob/7456789fb8319b4b9a838a480f6a4fc53f0e8fce/Library/Homebrew/cask/audit.rb#L536), which is why "1.6" and "1.6.0" aren't treated as equal. Casks interpolate the version into the `url`, so a missing `.0` makes a difference and the strictness of string comparison is warranted in this context.

The appropriate fix here is to modify the `livecheck` block, so it returns 1.6.0 (as in the file name) instead of 1.6 (from the `shortVersionString`). Bevan already beat me to it in https://github.com/Homebrew/homebrew-cask/pull/156995, though 😄

PR #157069, opened 2023-10-07
evernote 10.63.3

WangEdward:
evernote [10.63.3](https://evernote.com/release-notes) & Update livecheck

### DIFF
--- a/Casks/e/evernote.rb
+++ b/Casks/e/evernote.rb
@@ -30,14 +30,14 @@ cask "evernote" do
     end
   end
   on_mojave :or_newer do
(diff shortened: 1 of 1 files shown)


bevanjkay:
@WangEdward I believe the url change to "stage" is because Evernote conducts staggered rollouts. We should wait until livecheck returns the new version with the GA url.

WangEdward:
The only version available at the [official website](https://evernote.com/download) is [10.63.3](https://mac.desktop.evernote.com/builds/Evernote-latest.dmg), 10.58.8 don't seems to be the latest version that vendor intend user to install.

bevanjkay (approved):

PR #157796, opened 2023-10-16
new cask: DevEco Studio

cyxmon:

### DIFF
--- a/Casks/d/deveco.rb
+++ b/Casks/d/deveco.rb
@@ -1,0 +1,38 @@
+cask "deveco" do
+  version "3.1.0.501"
+
+  on_arm do
+    sha256 "f3e77ba60e596c9e49cd5fc3ab67f3f944efd235f2fa7b298b501abcfc668f04"
+
+    url "https://contentcenter-vali-drcn.dbankcdn.cn/pvt_2/DeveloperAlliance_package_901_9/b7/v3/ONY_R_ihQ6-6k8Fnu8Kl2Q/devecostudio-mac-arm-#{version}.zip?HW-CC-KV=V1&HW-CC-Date=20230621T074841Z&HW-CC-Expire=315360000&HW-CC-Sign=FE223D9B6AB5C722096FF8C7BA574555E63638374A3D3543C72D7D9E869AB934",
+        verified: "contentcenter-vali-drcn.dbankcdn.cn/"
+
+    container nested: "devecostudio-mac-arm-#{version}/deveco-studio-#{version}-aarch64.dmg"
+  end
+  on_intel do
+    sha256 "1a380b8b4a172b0f00af476b3bdcd83ee2dab24937c00b72d20d9121db99f5b7"
+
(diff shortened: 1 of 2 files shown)


bevanjkay:
@cyxmon Can you please explain the issue you are trying to resolve here, as we may be able to help?

cyxmon:
Sorry about the messy commit naming. Not quite much problems left now. If you would, could you tell me how to pass that commit style check about rebasing?

bevanjkay:
@cyxmon We can fix the Commit style issue at the end. What is the reason for including a `preflight` block?

cyxmon:
This cask is downloaded as a zip file, and the dmg file is in the directory of what's unarchived. Adding a preflight block is what I could think of to install the cask normally. Though, I think there should be some better way to do this but is currently beyond my knowledge.

bevanjkay:
@cyxmon I'll have a look at fixing it now

bevanjkay:
@cyxmon Have a look at the `container nested:` stanza, it should do what you are looking for - https://docs.brew.sh/Cask-Cookbook#optional-stanzas

cyxmon:
Thanks, it works elegantly now.

Casks/d/deveco.rb
```diff
@@ -0,0 +1,38 @@
+cask "deveco" do
+  version "3.1.0.501"
+
+  on_arm do
+    sha256 "f3e77ba60e596c9e49cd5fc3ab67f3f944efd235f2fa7b298b501abcfc668f04"
+
+    url "https://contentcenter-vali-drcn.dbankcdn.cn/pvt_2/DeveloperAlliance_package_901_9/b7/v3/ONY_R_ihQ6-6k8Fnu8Kl2Q/devecostudio-mac-arm-#{version}.zip?HW-CC-KV=V1&HW-CC-Date=20230621T074841Z&HW-CC-Expire=315360000&HW-CC-Sign=FE223D9B6AB5C722096FF8C7BA574555E63638374A3D3543C72D7D9E869AB934",
```
bevanjkay, line 7: @cyxmon It looks like there may be additional parts of this `url` that change between releases? I am also concerned about the use of query strings that may expire when time passes?

cyxmon: You're right. This is what I am concerning as well. They just distribute releases in this way that makes it hard to follow up, which may require manual maintenance. How do you often deal with this situation? I'm not sure whether it's possible to find another stabler link.

bevanjkay: @cyxmon If the `url` always requires a token that is refreshed then unfortunately this may not be able to be added to `homebrew-cask`. Acceptable casks should be easily-maintainable, and not require additional intervention on version updates than what can be achieved with the `brew bump` or `brew bump-cask-pr` commands.

cyxmon: Make sense. I'll try to reach out to the official to see if they could provide a better solution.

miccal:
I am going to merge this and we will see how stable the `url`'s are in time.

Thank you @cyxmon.

miccal (approved):

PR #157896, opened 2023-10-17
manuskript 0.15.0

samford:

-----

This PR updates `manuskript` to the latest release, 0.15.0.

This also updates the `livecheck` block to use the `GithubReleases` strategy and make the regex match additional file types beyond zip (as the 0.15.0 release uses a dmg instead of a zip file).

Currently, the `livecheck` block is checking the GitHub releases page and relying on the fact that release descriptions contain a link to the macOS release asset, as the release asset HTML is now fetched asynchronously on load (and when opening the assets list). Switching to the `GithubReleases` strategy allows us to check the release assets for recent releases instead of hoping that upstream always includes a link in the release description text.

To be clear, I'm using the `GithubReleases` strategy here because not every release contains an asset for macOS. If upstream becomes consistent about providing macOS files on every release, then we could switch to the `GithubLatest` strategy (though we would need to modify the `strategy` block in the process).

### DIFF
--- a/Casks/m/manuskript.rb
+++ b/Casks/m/manuskript.rb
@@ -1,18 +1,38 @@
 cask "manuskript" do
-  version "0.10.0"
-  sha256 "f48f9c96af19b42c61a1ab6119e581e9398c33a8795ba5051cc0d44add1d2d7f"
+  version "0.15.0"
+  sha256 "59cd751d2010e661e4b3fc38bbbf49801dcf9b9317d7a62aff0826f8bcbac085"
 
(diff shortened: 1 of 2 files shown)


p-linnane (dismissed):

p-linnane (dismissed):

samford:
Upstream has not marked the past few releases as pre-release, so it seems like we can remove `manuskript` from the GitHub prerelease allowlist.

razvanazamfirei (approved):

PR #157950, opened 2023-10-18
Update genymotion from 3.5.0 to 3.5.1

yurikoles:

### DIFF
--- a/Casks/g/genymotion.rb
+++ b/Casks/g/genymotion.rb
@@ -1,6 +1,6 @@
 cask "genymotion" do
-  version "3.5.0"
-  sha256 "c4be3f53a85908027340b7529dcdc79e7a6b19572056b354fc94d9688c0c10f7"
+  version "3.5.1"
+  sha256 "cc4bc949b8e5744272dbb77c798035ff009740d0eb8bd251b2b5d82daa2113fb"
 
   url "https://dl.genymotion.com/releases/genymotion-#{version}/genymotion-#{version}.dmg"
   name "Genymotion"
@@ -9,7 +9,7 @@ cask "genymotion" do
 
   livecheck do
     url "https://www.genymotion.com/download/"
-    regex(/Genymotion\s*Desktop\s*(\d+(?:\.\d+)+)/i)
+    regex(/genymotion-(\d+(?:\.\d+)+)\.dmg/i)
   end
 
   app "Genymotion.app"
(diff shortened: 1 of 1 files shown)


yurikoles:
My regex-fu is weak, so `livecheck` may be suboptimal.

Casks/g/genymotion.rb
```diff
@@ -9,9 +9,11 @@
 
   livecheck do
     url "https://www.genymotion.com/download/"
-    regex(/Genymotion\s*Desktop\s*(\d+(?:\.\d+)+)/i)
+    regex(/genymotion-(\d+(?:\.\d+)+)\.dmg/i)
```
samford, line 12:
```suggestion
    regex(/href=.*?Genymotion[._-]v?(\d+(?:\.\d+)+)\.(?:dmg|pkg)/i)
```

This suggestion is a variation on our typical pattern for matching the version from a filename in an `href` attribute. When it comes to casks, I sometimes include alternative filename extensions if it's using something like `dmg`, as the regex may fail to match a new version in the future if upstream switches to a different file format (e.g., pkg, zip, etc.). Upstream is using only the file extension to distinguish between files for macOS and Windows, so I've opted to just use the macOS-specific file extensions (dmg, pkg) for now.

yurikoles:
According to `otool` even `3.5.1` is built with deployment target `10.13`, thanks to [stackoverflow](https://stackoverflow.com/q/17143373/978568).
```sh
otool -l /Volumes/Genymotion\ Desktop/Genymotion.app/Contents/MacOS/genymotion | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.13
      sdk 10.14
Load command 10
```

In [release notes](https://docs.genymotion.com/desktop/Get_started/012_Macos_install/) they state it was **_tested_** on Ventura and Sonoma. So I'm unsure does it actually work on pre-Ventura.

Casks/g/genymotion.rb
yurikoles: Thank you for the explanation, I'll try to remember it.

yurikoles: Thank you for the explanation, I'll try to remember it.

bevanjkay (approved):

PR #157952, opened 2023-10-18
Update cinebench from 23 to 2024

yurikoles:

---

I can't fix `livecheck` and it seems like download from CLI ends with `403`.

### DIFF
--- a/Casks/c/cinebench.rb
+++ b/Casks/c/cinebench.rb
@@ -2,7 +2,8 @@ cask "cinebench" do
   version "2024"
   sha256 "dcc840852c85c2f9ea39f55e78b23efbb59a27960555c0fb319d5bce7f51fdb6"
 
-  url "https://mx-app-blob-prod.maxon.net/mx-package-production/website/macos/maxon/cinebench/Cinebench#{version}_macOS.dmg"
+  url "https://mx-app-blob-prod.maxon.net/mx-package-production/website/macos/maxon/cinebench/Cinebench#{version}_macOS.dmg",
+      referer: "https://www.maxon.net/en/downloads/cinebench-#{version}-downloads"
   name "Cinebench"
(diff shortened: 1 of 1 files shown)


samford (commented):
This `livecheck` block should continue to check the JS file for the https://www.maxon.net/en/downloads page (currently https://www.maxon.net/_nuxt/static/1697644834/en/downloads/payload.js). https://www.maxon.net/_nuxt/static/1697633267/en/downloads/cinebench-2024-downloads/payload.js is tied to a specific Cinebench version, which is something we avoid in checks (as it may not surface subsequent versions).

The existing check returns `23` and the only difference between the existing `livecheck` block URL and the current JS file URL is the numeric portion in the middle. From what I'm seeing, the numeric part may change when the website is updated, so checking a fixed `payload.js` URL like this may always return the same content for a specific revision of the website (making the check effectively useless in terms of surfacing new versions).

It seems that if we want to properly get the version information, we'll have to find the `payload.js` file URL in the [Downloads page](https://www.maxon.net/en/downloads) HTML, fetch the file, and match within it (i.e., this is one of those checks that requires two HTTP requests).

The changes are a little too involved for a suggestion, so I'll push a commit in a moment to update the `livecheck` block (edit: done).

samford:
Hm, I get a 403 (Forbidden) response when trying to fetch the dmg file locally, too. It only seems to work if the `Referer` is set (e.g., `curl -IL -H "Referer: https://www.maxon.net/en/downloads/cinebench-2024-downloads" https://mx-app-blob-prod.maxon.net/mx-package-production/website/macos/maxon/cinebench/Cinebench2024_macOS.dmg`).

bevanjkay (approved):

samford:
Thanks, @yurikoles!

PR #158044, opened 2023-10-18
lightkey: update livecheck, bump legacy version

bevanjkay:
Updates the `livecheck` to find the highest version number, and bumps legacy version.

@samford there is currently no way using the`:sparkle` strategy to sort by the minimum system version is there? This could be useful in these cases where we could easily filter the legacy/modern versions by system version comparison.

### DIFF
--- a/Casks/l/lightkey.rb
+++ b/Casks/l/lightkey.rb
@@ -1,7 +1,7 @@
 cask "lightkey" do
   on_monterey :or_older do
-    version "4.4.4"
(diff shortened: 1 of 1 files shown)


samford:
> there is currently no way using the`:sparkle` strategy to sort by the minimum system version is there? This could be useful in these cases where we could easily filter the legacy/modern versions by system version comparison.

`minimumSystemVersion` isn't included in the `Item` struct, so it's not something we can do in a `strategy` block at the moment. Adding it makes sense to me, so I'll create a brew PR in a bit.

miccal (approved):

PR #158320, opened 2023-10-23
guitar-pro: update livecheck

samford:

-----

The existing `livecheck` block for `guitar-pro` checks the homepage but only returns 8 as the version instead of 8.1. This is because the homepage only includes text with the major version (e.g., "Guitar Pro 8").

Upstream only seems to list new versions on their blog, so this PR updates the `livecheck` block to check the related section. The blog post title format varies (and there are other posts that aren't related to new versions) but they at least all include text like "Guitar Pro 8.1", "Guitar Pro 7.5.5", etc., so this check works for now.

### DIFF
--- a/Casks/g/guitar-pro.rb
+++ b/Casks/g/guitar-pro.rb
@@ -8,8 +8,8 @@ cask "guitar-pro" do
   homepage "https://www.guitar-pro.com/"
 
   livecheck do
-    url :homepage
-    regex(/Guitar\s*Pro\s*(\d+(?:\.\d+)*)/i)
+    url "https://www.guitar-pro.com/blog/c/2163-guitar-pro-news-tips"
+    regex(/Guitar\s+Pro\s+v?(\d+(?:\.\d+)*)/i)
   end
 
   pkg "guitar-pro-#{version.major}-setup.pkg"


bevanjkay (approved):

bevanjkay:
@samford Is this a better livecheck URL?
https://updates.guitar-pro.com/gp8?os=macOS&channel=stable

samford:
> Is this a better livecheck URL?
> https://updates.guitar-pro.com/gp8?os=macOS&channel=stable

I was hoping that there was an appcast but that appears to only cover 8.x versions. Outside of versioned formulae/casks, we avoid limited checks where we know that livecheck won't be able to find newer versions. Thanks for the suggestion, though.

bevanjkay:
@samford It's tricky with sparkle style feeds because some seem to lock into the major version, where some will bump to the next major on release. The feed is from the in-app updated, but the url does seem to point to the version with the sound files included, not the base installer.

samford:
> It's tricky with sparkle style feeds because some seem to lock into the major version, where some will bump to the next major on release.

Oh yeah, true. However, considering how unorganized the upstream website is when it comes to version information, it feels like a gamble as to whether upstream would add an item for the next major version when the time comes.

I think let's just keep this in our back pocket for now, for if/when our current approach becomes untenable in the future (and there isn't any alternative).

PR #158614, opened 2023-10-26
Update qingg from 2.14 to 2.15

Igotit:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions)

### DIFF
--- a/Casks/q/qingg.rb
+++ b/Casks/q/qingg.rb
@@ -1,6 +1,6 @@
 cask "qingg" do
-  version "2.14"
-  sha256 "26680dbc411f15c460866f07b2f4a7849d11f2717983688592d2c406cccc0e31"
+  version "2.15"
+  sha256 "f06747f985146b041df0883c12256adbdfe8e8608db0ac8a0d7a1ef12aa24f85"
 
   url "https://qingg.im/download/Qingg-#{version}.dmg"
   name "QinggIM"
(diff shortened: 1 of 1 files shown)


asins:
@Igotit  在这看到你在你也在提交版本啊~，我也刚提交了

bevanjkay (dismissed):

samford (commented):
I updated the `livecheck` block to match the version from the filename in the enclosure URL, as the `shortVersionString` in the appcast currently contains more text than just the version (i.e., `2.15 十周年纪念版` instead of only `2.15`). We can switch back to checking `item.short_version` if/when the `shortVersionString` is guaranteed to only be a numeric version but this will address the livecheck failure for now.

-----

Beyond that, it looks like there's an issue with installing the software on CI:

```
==> Running installer for qingg with sudo; the password may be necessary.
installer: Package name is 清歌输入法
installer: Installing at base path /
installer: The install failed. (The Installer encountered an error that caused the installation to fail. Contact the software manufacturer for assistance. An error occurred while running scripts from the package “安装包.pkg”.)
```

miccal:
@Igotit it appears that the `pkg` for this version included a `postinstall` script that only works on an Arm-based Mac. Are you able to confirm if this is correct, and that this can now not be installed on an Intel-based Mac?

Igotit:
> @Igotit it appears that the `pkg` for this version included a `postinstall` script that only works on an Arm-based Mac. Are you able to confirm if this is correct, and that this can now not be installed on an Intel-based Mac?

Thanks for figuring out this problem. I will update and re-submit.

miccal (approved):

PR #158830, opened 2023-10-30
luminance-hdr: update livecheck

samford:

-----

Prior to #158828, the existing `livecheck` block for `luminance-hdr` was giving an `Unable to get versions` error, as it was checking a page that no longer exists (and now simply redirects to the SourceForge project page).

This PR is an alternative fix for this issue, using the `Sourceforge` strategy to check the RSS feed for the `luminance` subdirectory, as this is the typical check for this scenario.

@bevanjkay Was there any particular reason for checking the project activity page that I'm missing? We don't have any other SourceForge formulae/casks using this type of check and it doesn't seem necessary here (not to mention that the activity page contains quite a bit of extraneous information like notifications for tickets, comments, blog posts, etc.).

### DIFF
--- a/Casks/l/luminance-hdr.rb
+++ b/Casks/l/luminance-hdr.rb
@@ -9,9 +9,8 @@ cask "luminance-hdr" do
   homepage "https://qtpfsgui.sourceforge.io/"
 
   livecheck do
-    url "https://sourceforge.net/p/qtpfsgui/activity"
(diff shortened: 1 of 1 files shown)


bevanjkay:
@samford No there wasn't any particular reason, but it looked like the previous `livecheck` block was using the activity feed. Due to new activity the links had pushed off the page, so I just maintained the same, but allowed it to find the same url as previously. Thanks for the correct fix.

bevanjkay (approved):

PR #159106, opened 2023-11-01
Set auto_updates true for Reflect

mfilej:
Reflect always automatically installs updates and has no way of opting out.

Additionally, in order to pass `brew audit`, I also did the following:
- updated the formula to Reflect 2.1.2
- moved the formula to `./r/reflect.rb`.

Note: `brew oudated` still lists `reflect (2.0.9) != 2.1.2` for me — I'm assuming it's not using the formula in `/opt/homebrew/` that I just edited and submitted?

***


p-linnane:
Your base branch is very out of date, which is why you are experiencing these issues.

p-linnane (approved):

PR #159213, opened 2023-11-02
music-decoy: new cask

alin23:

### DIFF
--- a/Casks/m/music-decoy.rb
+++ b/Casks/m/music-decoy.rb
@@ -2,17 +2,20 @@ cask "music-decoy" do
   version "1.0"
   sha256 "e4fa4670e24522ac5e0999ed6c483201d257e858fc644d3490edee6887b847eb"
 
-  url "https://files.lowtechguys.com/MusicDecoy-#{version}.zip"
+  url "https://github.com/FuzzyIdeas/MusicDecoy/releases/download/v#{version}/MusicDecoy.zip",
+      verified: "github.com/FuzzyIdeas/MusicDecoy/"
   name "Music Decoy"
-  desc "Stop launching the Music app whenever you press Play"
+  desc "Music app blocker utility"
   homepage "https://lowtechguys.com/musicdecoy"
 
   livecheck do
-    url "https://lowtechguys.com/musicdecoy/"
-    regex(%r{href="https://files.lowtechguys.com/MusicDecoy-(\d+\.\d+)\.zip"}i)
+    url :url
+    strategy :github_latest
   end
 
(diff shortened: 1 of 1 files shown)


Casks/m/music-decoy.rb
```diff
@@ -0,0 +1,21 @@
+cask "music-decoy" do
+  version "1.0"
+  sha256 "e4fa4670e24522ac5e0999ed6c483201d257e858fc644d3490edee6887b847eb"
+
+  url "https://github.com/FuzzyIdeas/MusicDecoy/releases/download/v#{version}/MusicDecoy.zip",
```
razvanazamfirei, line 5: I'll defer to you on what makes most sense distribution-wise. Using the github repo simplifies the livecheck, but if the website is more reliable I'm happy to revert it.

Casks/m/music-decoy.rb
```diff
@@ -0,0 +1,21 @@
...
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
+
+  depends_on macos: ">= :high_sierra"
+
+  app "Music Decoy.app"
+
+  # No zap stanza required
```
razvanazamfirei, line 20: Since the bundle ID is identical to the native app, I wanted to double check whether there are any other files saved that would need to be removed on uninstall.

razvanazamfirei (commented):
Salut Alin, 

I have a couple of questions I was hoping you could answer. Everything else looks great!

Casks/m/music-decoy.rb
alin23: I wanted to use the repo as well but it's a very new repo and it does not have enough stars 😅 at least that is what the audit said. If you can make an exception, then sure the Github repo makes more sense.

Casks/m/music-decoy.rb
alin23: Nope, the app does not create any other file. It does not use UserDefaults, Caches etc. Thanks for double checking!

Also the system Music app is sandboxed and has its own container so there is no danger of removing a file owned by the real Music app since my app is not sandboxed.

razvanazamfirei:
Repository exception:
> app that has its own website but the developers use GitHub for hosting the binaries

razvanazamfirei (approved):

alin23:
> Repository notability exception:
> 
> > app that has its own website but the developers use GitHub for hosting the binaries

Ooh interesting, I missed that. Thanks!

razvanazamfirei:
Awesome! Looks great! Thank you for the submission.

PR #159326, opened 2023-11-03
autodesk-fusion360: fix install and uninstall

bevanjkay:
This PR fixes the install stanza by making it truly unattended with no GUI, and also fixes the uninstall to completely uninstall the app.


Casks/a/autodesk-fusion360.rb
```diff
@@ -10,6 +10,9 @@
 
   installer script: {
     executable: "#{staged_path}/Install Autodesk Fusion 360.app/Contents/MacOS/Fusion 360 Client Downloader",
+    args:       [
+      "--quiet",
+    ],
```
miccal, line 15:
```suggestion
    args:       ["--quiet"],
```

miccal (approved):

PR #159370, opened 2023-11-04
Update URLs to resolve unnecessary redirections, part 2

samford:

-----

This is a follow-up to #159369. This PR updates URLs to resolve the same types of redirections as the aforementioned PR (see the description for more information) but this PR also involves changes to the main cask `url`. I created a separate PR for these to run CI without the `ci-syntax-only` label, so we will properly test cask `url` changes. To be clear, I've already tested/verified the `livecheck` block changes locally.

I've also added descriptions for `cursorcerer`, `filemon`, `isolator`, and `monity-helper`, to address the related audit.

### DIFF
--- a/Casks/a/airdisplay.rb
+++ b/Casks/a/airdisplay.rb
@@ -2,13 +2,13 @@ cask "airdisplay" do
   version "3.4.2"
   sha256 "272d14f33b3a4a16e5e0e1ebb2d519db4e0e3da17f95f77c91455b354bee7ee7"
 
-  url "https://www.avatron.com/updates/software/airdisplay/ad#{version.no_dots}.zip"
+  url "https://avatron.com/updates/software/airdisplay/ad#{version.no_dots}.zip"
   name "Air Display"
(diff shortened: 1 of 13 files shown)


samford:
The `livecheck` block for `aladin` works fine for me locally (macOS14, ARM), so I'm not sure why it's failing on CI:

```
audit for aladin: failed
 - The livecheck URL https://aladin.cds.unistra.fr/java/nph-aladin.pl?frame=downloading is not reachable
Error: 1 problem in 1 cask detected.
```

I'll re-run `aladin` again in a bit and remove it from this PR if it fails again.

samford:
Ah, `#curl_check_http_content` fails with the `aladin` `livecheck` block URL (both with and without these changes) because the upstream server incorrectly uses `\n\n` to separate the response headers instead of `\r\n\r\n`. As a result, `#parse_curl_output` treats the response as the body instead, doesn't parse the status code from the response headers, and this causes the audit to fail.

I'll remove `aladin` from this PR while I think about this.

p-linnane (approved):

PR #159377, opened 2023-11-04
xscreensaver: add uninstall_postflight and zap

krehel:
Background: 

The `pkg` that installs `xscreensaver` does not contain an uninstall script nor does it appear to keep receipts of the installed artifacts.  As a consequence, running a `brew uninstall` does not remove the actual screensaver artifacts stored in `/Library/Screen Savers` as of this commit.

This commit introduces a brief `uninstall_postflight` to look through all screensavers in `/Library/Screen Savers` and identify the bundle identifier associated with the screensavers in this package (`org.jwz`), gather the `CFBundleExecutable` name, and remove any whitespace so it can be used as the identified artifact for removal.  

As of this commit there are 252 screensavers, so while it's not particularly effective to add each one to `zap trash`, it can be done and would require ongoing maintenance with each release.

Certainly welcome to changes or a better way to do this.  Thanks!

### DIFF
--- a/Casks/x/xscreensaver.rb
+++ b/Casks/x/xscreensaver.rb
@@ -14,9 +14,30 @@ cask "xscreensaver" do
 
   pkg "Install Everything.pkg"
 
+  # There is no uninstall script for this Cask, so a manual uninstall is performed
+  # Loop through all screensaver plist files, looking for "org,jwz" in the bundle identifier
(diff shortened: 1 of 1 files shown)


bevanjkay:
Seems to work okay locally for me. Soft approval, will give some time for other to respond.

miccal (approved):

PR #159432, opened 2023-11-06
Hash update on Lunar 6.3.0 because of hotfix

alin23:

---

**[Lunar developer here](https://github.com/alin23/Lunar)**: I had to do an emergency hotfix for 6.3.0, the last build had a dangerous bug on sending monitor commands too fast.

I hope you can accept a hash change. I can provide any verification you need if you're worried if someone took over my account and is doing this.

The commits are signed with my GPG key, in case that would be enough of a verification:

Firefox Developer Edition-004730-Monday-11-39

### DIFF
--- a/Casks/l/lunar.rb
+++ b/Casks/l/lunar.rb
@@ -1,6 +1,6 @@
 cask "lunar" do
   version "6.3.0"
-  sha256 "ee20dbdafd88e20a6693aeb6d243bf646346bcc4c891eca5e58433c69cce8d7f"
+  sha256 "ecb2117b5105cefbf3917b4d5da31b18b775d3a153e3c6472fdb9e3e3b216aae"
 
   url "https://static.lunar.fyi/releases/Lunar-#{version}.dmg"
   name "Lunar"


bevanjkay:
@alin23 Yes we could update this PR, but for Homebrew users they will not be notified that there is an update available unless the version is incremented. In these cases it is best to issue a new release with an update minor version so that your user-base is fully aware (even in the case of a hotfix).

alin23:
> @alin23 Yes we could update this PR, but for Homebrew users they will not be notified that there is an update available unless the version is incremented. In these cases it is best to issue a new release with an update minor version so that your user-base is fully aware (even in the case of a hotfix).

That's ok, I would not have done this if there were users that needed the update. I double checked my analytics dashboard and there are very few users that updated to 6.3.0 already. Based on the tags, none of them have setups affected by the bug.

bevanjkay (approved):

alin23:
Thank you @bevanjkay ! Appreciate the fast resolution \^\_\^

PR #159436, opened 2023-11-06
hdfview: fix url and livecheck

bevanjkay:
Fixes broken `url` and additionally the `livecheck` block.

### DIFF
--- a/Casks/h/hdfview.rb
+++ b/Casks/h/hdfview.rb
@@ -5,11 +5,11 @@ cask "hdfview" do
   url "https://support.hdfgroup.org/ftp/HDF5/releases/HDF-JAVA/hdfview-#{version}/bin/HDFView-#{version}-macos11_64.tar.gz"
   name "HDFView"
   desc "Tool for browsing and editing HDF files"
-  homepage "https://confluence.hdfgroup.org/display/support"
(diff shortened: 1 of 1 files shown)


Casks/h/hdfview.rb
```diff
@@ -5,11 +5,11 @@
...
   name "HDFView"
   desc "Tool for browsing and editing HDF files"
-  homepage "https://confluence.hdfgroup.org/display/support"
+  homepage "https://www.hdfgroup.org/downloads/hdfview/"
 
   livecheck do
-    url "https://confluence.hdfgroup.org/display/support/Download+HDFView"
-    regex(/HDFView\+(\d+(?:\.\d+)+)/i)
+    url :homepage
+    regex(/href=.*?HDFView[._-]v?(\d+(?:\.\d+)+)[._-]macos11_64\.(?:t|zip)/i)
```
samford, line 12:
```suggestion
    regex(/href=.*?HDFView[._-]v?(\d+(?:\.\d+)+)[._-]macos[^"' >]*?\.(?:t|zip)/i)
```

This will fail to match if/when upstream updates the `macos` suffix (e.g., `macos12_64` instead of `macos11_64`), so we may be better off glossing over the specifics after `macos`.

samford (approved):

PR #159442, opened 2023-11-06
Improve MS Teams ZAP

kiblik:

### DIFF
--- a/Casks/m/microsoft-teams.rb
+++ b/Casks/m/microsoft-teams.rb
@@ -64,6 +64,7 @@ cask "microsoft-teams" do
         "~/Library/Containers/com.microsoft.teams2.launcher",
         "~/Library/Containers/com.microsoft.teams2.notificationcenter",
         "~/Library/Cookies/com.microsoft.teams.binarycookies",
+        "~/Library/Group Containers/*.com.microsoft.teams",
         "~/Library/HTTPStorages/com.microsoft.teams",
         "~/Library/HTTPStorages/com.microsoft.teams.binarycookies",
         "~/Library/Logs/Microsoft Teams",


krehel:
I am marking this as draft because it would be valuable to go through the `zap` items and ensure the only items being captured are from the "New" Teams, so there is not accidentally breakage of the other version.

miccal:
@samford the versions for the new Microsoft Teams (and the classic one also) are listed here:

https://learn.microsoft.com/en-us/officeupdates/teams-app-versioning

Any chance you can work your `livecheck` magic and use this link instead of the current `:extract_plist` strategy?

FYI: I also maintain this in my tap and use a different `livecheck`: https://github.com/miccal/homebrew-miccal/blob/master/Casks/m-microsoft-teams.rb#L9-L12

samford:
> Any chance you can work your `livecheck` magic and use this link instead of the current `:extract_plist` strategy?

Thanks for the information. It may be a little tricky to manually parse the HTML tables (i.e., we don't use `nokogiri` in brew, since it uses native gem extensions) but I'm always glad when we can avoid `ExtractPlist`. I've run out of time today but I'll try to take a pass at it tomorrow.

Bo98:
There's also https://config.teams.microsoft.com/config/v1/MicrosoftTeams/1415_1.0.0.0?environment=prod&audienceGroup=general&teamsRing=general&agent=TeamsBuilds

bevanjkay:
@Bo98 Is there any significance to the `1415_1.0.0.0` in the URL?

Bo98:
Combination of client versions I believe. The API is used for more than update checking. It seems that bit doesn't really matter for this, but must be something numerical.

samford:
> It seems that bit doesn't really matter for this, but must be something numerical. Can probably check what the actual client uses.

It would be good to know what the client does but I can at least say that using the cask version in the URL seems to work fine:

```ruby
  livecheck do
    url "https://config.teams.microsoft.com/config/v1/MicrosoftTeams/#{version}?environment=prod&audienceGroup=general&teamsRing=general&agent=TeamsBuilds"
    strategy :json do |json|
      json.dig("BuildSettings", "WebView2", "macOS", "latestVersion")
    end
  end
```

The response seems to provide the latest version information regardless of what version you use in the URL, so that seems promising. If we agree on this approach, I'll push a commit in the morning.

miccal:
Closing in favour of https://github.com/Homebrew/homebrew-cask/pull/159807

PR #159483, opened 2023-11-06
3dgenceslicer: add zap and livecheck update

krehel:
Adding zap stanza

### DIFF
--- a/Casks/3/3dgenceslicer.rb
+++ b/Casks/3/3dgenceslicer.rb
@@ -9,8 +9,7 @@ cask "3dgenceslicer" do
 
   livecheck do
     url "https://support.3dgence.com/software.html"
-    regex(/3DGence(?:[._-]|\s)?Slicer(?:[._-]|\s)?(\d+(?:\.\d+)+)(?:[._-]|\s)?
-          \((?:version|v)?(?:[._-]|\s)?(\d+(?:\.\d+)+)\)(?:[._-]|\s+)*system(?:[._-]|\s)MacOS/ix)
+    regex(/3DGence\s+SLICER\s+v?(\d+(?:\.\d+)+)\s+\((?:version\s+|v)?(\d+(?:\.\d+)+)\)\s+-\s+system\s+MacOS/im)
     strategy :page_match do |page, regex|
       page.scan(regex).map do |match|
         "#{match[1]},#{match[0]}"
--- a/Casks/3/3dgenceslicer.rb
+++ b/Casks/3/3dgenceslicer.rb
@@ -9,9 +9,8 @@ cask "3dgenceslicer" do
 
   livecheck do
     url "https://support.3dgence.com/software.html"
-    regex(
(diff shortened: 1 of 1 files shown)


chenrui333 (dismissed):

krehel:
When trying to add a `zap`, the CI picked up a broken livecheck.  The information for the version is in a string like follows.  DMG file does not appear to be linked, though the download still works.

`<p>3DGence SLICER 4.0 (version 3.4.0) - system MacOS</p>`

Maybe went a bit too far with the livecheck, but tried to match spaces and [._-] for robustness.  The `system MacOS` is in the livecheck to avoid picking up the Windows version, which wouldn't matter too much if they stay in sync.  If they diverge it'll break again, so rather be explicit.

@samford expertise always appreciated! 😄

bevanjkay:
@krehel It's odd that upstream are not providing a link to download that version, perhaps it is intentional?

krehel:
I thought about that.  It could be accidental as the Windows version is available, and all other download links on that page are functional.  Additionally, `brew fetch/install` etc. does work so the link itself isn't blocked or redirecting.

Though maybe we're better off pointing `livecheck` at something else.

Open to suggestions!

Casks/3/3dgenceslicer.rb
```diff
@@ -9,9 +9,8 @@
 
   livecheck do
     url "https://support.3dgence.com/software.html"
-    regex(
-      %r{href=.*?/3DGence(?:[._-]|%20)?Slicer(?:[._-]|%20)(\d+(?:\.\d+)+)(?:[._-]|%20)\(v?\.?(\d+(?:\.\d+)+)\)\.dmg}i,
-    )
+    regex(/3DGence(?:[._-]|\s)?Slicer(?:[._-]|\s)?(\d+(?:\.\d+)+)(?:[._-]|\s)?
+          \((?:version|v)?(?:[._-]|\s)?(\d+(?:\.\d+)+)\)(?:[._-]|\s+)*system(?:[._-]|\s)MacOS/ix)
```
samford, line 13:
```suggestion
    regex(/3DGence\s+SLICER\s+v?(\d+(?:\.\d+)+)\s+\((?:version\s+|v)?(\d+(?:\.\d+)+)\)\s+-\s+system\s+MacOS/im)
```

I think a simpler version with respect to whitespace is probably fine. I usually don't see whitespace turn into something like `[._-]` in loose text like this (filenames are a different story) but I have seen regexes break from using a literal space or `\s` instead of `\s+` (or `\s*`, when appropriate).

samford (commented):
I happened to be going through checks yesterday and saw that this one was broken (due to the missing dmg link). I was on the fence about whether to just wait and see if they fix the dmg link with the next version or update the `livecheck` block to match the related text.

Since we're already going down this path, I'm fine with updating the `livecheck` block to match the related text. If we run into issues with this in the future, we can always revisit it.

samford (approved):

PR #159619, opened 2023-11-08
timely: remove cask

krehel:
This application looks to have been replaced by [Memory](https://github.com/Homebrew/homebrew-cask/blob/4171cd8e66bccb6801d2bfa6df2a70823d4940f5/Casks/m/memory.rb).

Though the upstream [repository](https://github.com/Timely/desktop-releases) still exists, it has not seen an update in over a year (at time of this comment) despite having relatively frequent dev activity for a while previously.  It appears to have transitioned to a web app with the Memory app as a desktop companion.

Additionally, the upstream webpage (below) references using the Timely Desktop app as Memory.

Screenshot 2023-11-07 at 6 42 33 PM


bevanjkay:
@krehel This app seems to still be receiving a few downloads so maybe we should consider adding a `discontinued` caveat first?

```
timely
30 days: 9 (#3244)
90 days: 20 (#3573)
365 days: 48 (#3622)
Age: 2612 days (added 2016, September 13)
```

krehel:
We can.  I could not find references to this desktop app on the upstream site any longer and appears to be superseded.

If we do keep, we should at least update to the latest version as this is several years out of date.

bevanjkay:
@krehel I considered we could use a migration, but the apps seem to be considerably different.
Given how long ago this was last updated, I am happy to go ahead with the removal.

bevanjkay (approved):

krehel:
Cheers, thanks @bevanjkay.  We can always add it back in if needed, also.

PR #159625, opened 2023-11-08
keymapp: add `auto_updates true`

bevanjkay:
App has in-app update mechanism. Was missed in initial PR.

### DIFF
--- a/Casks/k/keymapp.rb
+++ b/Casks/k/keymapp.rb
@@ -15,6 +15,7 @@ cask "keymapp" do
     end
   end
 
+  auto_updates true
   depends_on macos: ">= :monterey"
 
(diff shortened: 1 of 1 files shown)


krehel:
I saw this and originally included it, but decided not to because it update checks seemed optional.  Which lead me to believe it was not automatically downloading and installing versions in the background similar to how VSCode or Chrome would.

bevanjkay:
If the app has an inbuilt mechanism to **both** download **and** install the update (whether it can be toggled or not) then we should set `auto_updates` to `true`.

krehel (approved):

krehel:
We'll do it 🚀

PR #159702, opened 2023-11-08
forklift: add `conflicts_with` for version 3

bevanjkay:
Ref: https://github.com/Homebrew/homebrew-cask-versions/pull/18325

### DIFF
--- a/Casks/f/forklift.rb
+++ b/Casks/f/forklift.rb
@@ -13,6 +13,7 @@ cask "forklift" do
   end
 
   auto_updates true
+  conflicts_with cask: "homebrew/cask-versions/forklift3"
   depends_on macos: ">= :sierra"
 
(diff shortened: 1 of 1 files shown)


krehel:
For consideration -

I reached out to the vendor a while back because their Sparkle feed minimum OS requirements don't align with the stated requirements on their website.

Since Forklift 4 is for 12 (Monterey) and newer according to their site, would it make sense just to keep Forklift 3 in this Cask also and just use `big_sur :or_older`, foregoing the `cask-versions` file?

bevanjkay:
@krehel The reason for the other cask would be more due to requiring a new license for version 4.

krehel:
Got it @bevanjkay!

krehel (approved):

PR #159935, opened 2023-11-12
Bump OBS to 30.0.0 

wblondel:
It should be an easy bump but OBS now uses "Intel" and "Apple" as architecture names in the download link. The architecture names in the livecheck URL remain the same.

The livecheck still returns 29.1.3 as the latest version, even though the latest stable version on the OBS website is 30.0.0, so I cannot check checkbox 2.

No offense detected by brew style --fix, checking checkbox 3.

-----

### DIFF
--- a/Casks/o/obs.rb
+++ b/Casks/o/obs.rb
@@ -9,7 +9,7 @@ cask "obs" do
   url "https://cdn-fastly.obsproject.com/downloads/obs-studio-#{version}-macos-#{arch}.dmg"
   name "OBS"
   desc "Open-source software for live streaming and screen recording"
-  homepage "https://obsproject.com/forum/list/test-builds.20/"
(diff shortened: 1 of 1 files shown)


derrod:
The Sparkle URL is outdated, the old ones are supposed to be still updated but it seems that was missed, it'll be done soon:tm:.

Our current URLs are:
- ARM: https://obsproject.com/osx_update/updates_arm64_v2.xml
- x86: https://obsproject.com/osx_update/updates_x86_64_v2.xml

Note that we're making use of Sparkle's "channel" feature to allow users to opt-in to beta updates, I'm not sure how well supported that is here.

miccal:
Thank you @wblondel and @derrod.

miccal (approved):

PR #159975, opened 2023-11-13
raze: use `github_releases` livecheck strategy

bevanjkay:
Use `github_releases` livecheck strategy to find the latest version that includes a suitable binary.

### DIFF
--- a/Casks/r/raze.rb
+++ b/Casks/r/raze.rb
@@ -7,6 +7,8 @@ cask "raze" do
   desc "Build engine port backed by GZDoom tech"
   homepage "https://github.com/coelckers/Raze"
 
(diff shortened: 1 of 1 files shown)


samford (approved):
The situations where we need to use `GithubReleases` may change in the future (e.g., upstream could start reliably providing a macOS file with every release), such that we could potentially switch to `GithubLatest` at some point. With that in mind, I try to always document the reason why we need to use the `GithubReleases` strategy, so it's easy to reevaluate if we can switch away from it at a later date (as it naturally involves fetching more data than `GithubLatest`).

I pushed a commit to add an explanatory comment before the `livecheck` block, so this should be good to go.

PR #158514, opened 2023-10-25
Fix sf cask hashes

CarsonF:
After making any changes to a cask, existing or new, verify:

- [ ] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [ ] `brew audit --cask --online <cask>` is error-free.
- [ ] `brew style --fix <cask>` reports no offenses.

PR #156346, opened 2023-09-27
sf: add cask

krehel:
- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask --online <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [x] `brew audit --new-cask <cask>` worked successfully.
- [x] `brew install --cask <cask>` worked successfully.
- [x] `brew uninstall --cask <cask>` worked successfully.

Salesforce has released a `v2` of their CLI, simply named `sf` which replaces the now obsolete `sfdx` (another PR to be submitted to end-of-life that Cask, and add a caveat to point to `sf`).

Details can be found here => https://github.com/forcedotcom/cli/tree/main/releasenotes

```
IMPORTANT: Are you still using sfdx (v7)? If so, we recommend that you move to sf (v2). It's easy: simply uninstall sfdx and then install sf. 
```

PR #160014, opened 2023-11-13
sf: updates for `sha256` issues

krehel:
* Revert to `sha256 :no_check` to avoid in place update hash issues

* Parse `README.md` on GitHub for [stable] version

This cask has been having some issues with `sha256` mismatches (#159993, #158514, #156404, #156385).  There have also been several updates to this Cask, sometimes on a daily basis, when this should be updating once per week as a `stable` release from Salesforce on Wednesdays, US time.  Since these incremental updates are being pulled in, it seems as a regression and not providing an actual `stable` version as intended.

This PR switches to a method that was used in its predecessor cask, `sfdx`, by parsing the authoritative source of updates from the Salesforce repository [`README.md`](https://raw.githubusercontent.com/forcedotcom/cli/main/releasenotes/README.md) and extracting the update beside the `[stable]` tag.  This also pulls the latest official `stable` download from Salesforce directly for install.  Unfortunately, this is not a versioned `url` so dropping the `sha256` will be needed, but should help improve overall stability.

### DIFF
--- a/Casks/s/sf.rb
+++ b/Casks/s/sf.rb
@@ -1,25 +1,22 @@
 cask "sf" do
   arch arm: "arm64", intel: "x64"
 
-  version "2.16.10,38f3e4f"
-  sha256 arm:   "f6845bb2d329bdae9ed835f53390b8ad60e335f0cc15cd368cef63803e09b93a",
-         intel: "18d09197e7cd31519589d7f8117a369b1b0ffb26cc921ce53a9e17ac1165b454"
+  version "2.16.7"
+  sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


krehel:
Note: the proposed changes are in line with the original submission for this Cask found at https://github.com/Homebrew/homebrew-cask/pull/156346/commits/414cc08a0fa7f94db2908e2f72dbc8b85dd981ad

bevanjkay:
@krehel Changes look good. 
The previous was a false assumption of mine that upstream's update method would be stable. But as they update in place a lot, it does seem to be best to just use the non-versioned approach

krehel:
No issues at all @bevanjkay.  I appreciate you were able to improve upon my submission and make it a more rigorous check.  Just unfortunate about the updates in place that we didn't know about.

bevanjkay (approved):

PR #160186, opened 2023-11-14
kindle: delete unsupported cask

dkav:
Kindle for Mac (now called the Kindle Classic app) has been replaced with a new app available from the App Store. See https://www.amazon.com/kindle-dbs/arp/B0C9PRPV28 for more info.


bevanjkay:
While the download is still available we could add a `discontinued` caveat with a reference that the latest versions are only available on the Mac App Store first.

dkav:
I guess you could but it probably should be renamed. The renamed App Store version of this app is still available (with end of support noted). See https://apps.apple.com/us/app/kindle-classic/id405399194?mt=12

krehel:
I second to support the discontinued, but not removal.  Normally I'd support removal since it's effectively being abandoned and requires Rosetta, which means it's unlikely to come to Apple Silicon in its current iteration outside of App Store, never mind any security implications.

However this is relatively popular download.

```
30 days: 412 (#420)
90 days: 1,455 (#351)
365 days: 4,073 (#305)
```

Steps to take imho -

1. Mark as `discontinued` with a note pointing at the "official successor" to this Cask (such as App Store url, see `sfdx` or `zsa-wally` for examples)
2. Add `requires_rosetta` as it is not listed currently, and is needed.

PR #160479, opened 2023-11-18
Fix Calibre version compatibility

Funk66:
Calibre 7.0 is not compatible with older MacOS releases. See https://calibre-ebook.com/download_osx
Note: I wasn't able to verify the changes with `brew audit --cask --online calibre`. It failes with `No Cask with this name exists`. Providing a path to my updated formula, instead of the cask token, fails with `Calling brew audit [path ...] is disabled!`. I read the docs and found them to be unclear about what command I'm supposed to execute. In any case, I tested that (un)installing this formula works on BigSur.

### DIFF
--- a/Casks/c/calibre.rb
+++ b/Casks/c/calibre.rb
@@ -23,7 +23,23 @@ cask "calibre" do
       skip "Legacy version"
     end
   end
-  on_big_sur :or_newer do
+  on_big_sur do
+    version "6.29.0"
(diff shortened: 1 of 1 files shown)


razvanazamfirei (approved):

razvanazamfirei:
Hi, @Funk66! Thanks so much for your first contribution! This looks great! 👍

PR #160519, opened 2023-11-19
lightkey: update livecheck

samford:

-----

As proposed in https://github.com/Homebrew/homebrew-cask/pull/158044, this updates the `livecheck` block for the legacy version of `lightkey` to match items based on their `minimumSystemVersion`. This wasn't previously possible but I recently modified the `Sparkle` strategy to surface more `item` values (see https://github.com/Homebrew/brew/pull/16158 and https://github.com/Homebrew/brew/pull/16196), so we can now use `item.minimum_system_version` to filter `items` in a `strategy` block.

This type of approach should only be used when upstream is actively releasing new legacy versions (alongside the newer, primary version). This isn't appropriate for when upstream simply bumps the `minimumSystemVersion` going forward and does not publish new versions with a lower major/minor/etc. version.

With that in mind, this should probably switch back to using `skip "Legacy version"` if/when upstream stops publishing new legacy versions. I didn't see any clear guidance around how long legacy versions will be maintained but I added a comment about this (to keep it in mind).

One caveat about this approach is that the conditions in the `strategy` block may need to be updated as the minimum system version of releases changes (though only as needed to correctly match legacy versions).

The other caveat is that this will omit any item without a `minimumSystemVersion` value. Since we need this information to restrict matching to certain macOS versions, omitting those items is more reasonable than naively including them (potentially leading to an inappropriate version being reported as newest).

### DIFF
--- a/Casks/l/lightkey.rb
+++ b/Casks/l/lightkey.rb
@@ -3,8 +3,21 @@ cask "lightkey" do
     version "4.4.5"
     sha256 "457df4bb2d2f21a52eec9b9b05830eb082014ce66fd79f91544a0838d54a3241"
 
+    # This check should only return legacy versions and the conditions may need
+    # to be updated as the minimum system version of releases changes. If/when
+    # upstream stops publishing new legacy versions, this should be updated to
+    # use `skip` instead.
     livecheck do
-      skip "Legacy version"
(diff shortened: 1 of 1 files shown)


miccal (approved):

bevanjkay:
Thanks @samford

PR #160547, opened 2023-11-19
standard-notes: update livecheck

samford:

-----

This updates the `livecheck` block for `standard-notes` to use the typical regex for matching tags on the GitHub releases page. We've mostly migrated to use the `GithubLatest` and `GithubReleases` strategy instead but this one makes sense to continue checking the releases page with the `prerelease:false` query.

Basically, upstream creates a lot of pre-release versions but GitHub doesn't provide a way to omit them from the API results, so there's a real chance that the newest stable release for desktop will be pushed out of the most recent API data (as it's limited to 30 recent releases by default). Until the GitHub API provides this functionality, we're better off maintaining the existing check.

That said, the existing check isn't perfect, as there's a chance that non-desktop releases could push the latest desktop release beyond the first page of results (limited to 10 recent releases). However, this is arguably less likely to occur than the previously-described situation with the GitHub API.

Past that, this adds `depends_on macos: ">= :catalina"`, to address the related audit error.

### DIFF
--- a/Casks/s/standard-notes.rb
+++ b/Casks/s/standard-notes.rb
@@ -11,16 +11,21 @@ cask "standard-notes" do
   desc "Free, open-source, and completely encrypted notes app"
   homepage "https://standardnotes.org/"
 
-  # Using only desktop releases that are not marked as pre-release to align
-  # with the auto-updater strategy used by the app.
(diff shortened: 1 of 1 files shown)


razvanazamfirei (approved):

razvanazamfirei:
Thanks as always, @samford!

PR #160605, opened 2023-11-20
Update some livecheck blocks to use :url

samford:

-----

The casks in this PR use a GitHub release asset and have a `livecheck` block that uses a GitHub URL string. This updates these `livecheck` blocks to simply use `url :url` instead.

We only use a GitHub URL string in a `livecheck` block when the cask `url` is not a GitHub URL. In those cases, we should add a comment before the `livecheck` block to explain why we can't create a check that aligns with the cask `url` source and have to check GitHub instead. [I'm working through GitHub-related `livecheck` blocks and will be adding comments for existing exceptions in a later PR.]

### DIFF
--- a/Casks/c/cellprofiler.rb
+++ b/Casks/c/cellprofiler.rb
@@ -9,7 +9,7 @@ cask "cellprofiler" do
   homepage "https://cellprofiler.org/"
 
   livecheck do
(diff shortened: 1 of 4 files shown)


p-linnane (approved):

samford:
Whoops, looks like I had a typo in the commit messages 😅

PR #160823, opened 2023-11-23
ci: add ability to manually audit a list of casks

bevanjkay:
This PR adds the ability for us to manually audit a list of casks with a workflow dispatch.


razvanazamfirei:
I'm not comfortable enough with the CI workflows to review this, but hopefully someone else can chime in.

p-linnane:
@Bo98 & @carlocab are the ones most familiar with GitHub Actions AFAIK.

bevanjkay:
For additional context, the idea here is to be able to pass a list of casks to run audits on CI manually to check existing casks for issues with downloads, mismatches etc.. without having to run locally for maintainers. 

May need an additional option to run "stricter" tests by optionally passing `--new-cask`

krehel:
This is a key addition towards being able to move away from unsigned Casks as well.  Auditing "the world" locally is a pretty tedious and network-intensive process.

We can also catch a fair number of issues with stricter checking on version bump, but obviously doesn't encapsulate all Casks and avoids during-PR issues.

Bo98:
Would be good to combine `cmd/lib/manual-generate-matrix.rb` with `cmd/lib/generate-matrix.rb` more. Bit too much copy-pasting between the two I reckon.

bevanjkay:
I'll do a refactor hopefully tomorrow

bevanjkay:
@bo98 I have had a go at a refactor.

Bo98:
Thanks!

Workflow also looks like a copy-paste job. It looks like the only differences are the addition of `workflow_dispatch` and a different `generate-matrix` line.

Could you instead add `workflow_dispatch` to `ci.yml` and then in the `generate-matrix` step you can chose which args to pass via:
```bash
if [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]]
then
  # ...
else
  # ...
fi
```

bevanjkay:
@Bo98 I think I have done this correctly, and thank you for your guidance to clean this up 🙇‍♂️

Bo98 (approved):
Thanks!

PR #161024, opened 2023-11-26
rode-connect 1.3.33 (update version)

jerryibrahim:

- [ X ] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [ X ] `brew audit --cask --online <cask>` is error-free.
- [ X ] `brew style --fix <cask>` reports no offenses.

### DIFF
--- a/Casks/r/rode-connect.rb
+++ b/Casks/r/rode-connect.rb
@@ -1,5 +1,5 @@
 cask "rode-connect" do
-  version "1.3.31"
+  version "1.3.33"
   sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


bevanjkay:
This is not an ideal scenario, upstream's updater is still showing 1.3.31 - which is an greater concern than normal because the version is required to install the cask (due to it formulating part of the `.pkg` file). It also means this cask will break on every update, because the download is not versioned.

PR #161048, opened 2023-11-26
aria2gui: remove

chenrui333:
last release was like six years ago
image

image


bevanjkay:
I'm not convinced that this should be removed yet, while it is discontinued, it is still relatively popular.
So if it is still functional we should keep it.

chenrui333:
(on one side, I agree with you, on the other side, I think this is good to be deleted so that people can look for alternative choices)

chenrui333:
I will leave this PR open for a bit

chenrui333:
Just like this PR, #161053 (fortunately, the repo is not archived, people have a central place to discuss the issues, IMO, this cask should be removed at this point)

chenrui333:
cc @Homebrew/cask

razvanazamfirei:
> I'm not convinced that this should be removed yet, while it is discontinued, it is still relatively popular. So if it is still functional we should keep it.

Agree with this.

krehel:
Think we all agree on keeping this due to popularity and pending future implementation of `deprecate/disable`.

Closing as unmerged.

PR #161051, opened 2023-11-27
clover: remove cask

chenrui333:
image

- https://toolfinder.co/news/clover-shutdown

----

image


bevanjkay:
@chenrui333 I think we should remove this one, as it does not look to be functional anymore - existing data cannot even be exported.

Screenshot 2023-11-27 at 11 04 58 am

chenrui333:
> @chenrui333 I think we should remove this one, as it does not look to be functional anymore - existing data cannot even be exported.
> 
> Screenshot 2023-11-27 at 11 04 58 am

totally, removing now.

krehel:
Agreed with @bevanjkay

krehel (approved):

chenrui333:
yeah, removed :)

PR #161059, opened 2023-11-27
update cask cloudflare-warp

shikiroot:
simple version bump


bevanjkay:
Upstream are not serving this as the latest version yet, so we will need to wait for that to be updated before this passes CI.

PR #161189, opened 2023-11-28
Add mdb-viewer

samford:

-----

This adds a cask for MDB Viewer, which is an app I use by the same author as the existing `postico*` and `table-tool` casks. The [app homepage](https://eggerapps.at/mdbviewer/) refers to the app as both "MDB Viewer" and "MDB/ACCDB Viewer", so I used the longer version as the `name` but the shorter version as the cask name. If we would prefer to simply use "MDB Viewer" as the `name`, I can update this.

The download link (https://eggerapps.at/mdbviewer/download/) on the app homepage redirects to the latest versioned zip file, so I've used that URL in the `livecheck` block with the `HeaderMatch` strategy. `HeaderMatch` can identify the version from the filename fine, so we don't need a regex for now.

### DIFF
--- a/Casks/m/mdb-viewer.rb
+++ b/Casks/m/mdb-viewer.rb
@@ -1,0 +1,22 @@
+cask "mdb-viewer" do
+  version "2.3.1"
+  sha256 "8f6ee157d169182c26f2f3b8cffe1387010203a33a9d3bddaf71706e40783493"
+
+  url "https://eggerapps-downloads.s3.amazonaws.com/mdbviewer-#{version}.zip",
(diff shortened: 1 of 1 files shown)


p-linnane (approved):

bevanjkay:
The token here really should be the longer name. The token reference is usually quite strict for uniformity, for `app` based installs it always follows the artifact.

bevanjkay:
@samford It only has one noted installation at the moment, so we can probably skip the cask_renames if we do it soon enough. Thanks.

PR #161302, opened 2023-11-29
google-drive 84.0.6

bevanjkay:
Created by `brew bump`

---

Created with `brew bump-cask-pr`.

PR #161140, opened 2023-11-28
chrome-remote-desktop-host 120.0.6099.8

chenrui333:
Created by `brew bump`

---

Created with `brew bump-cask-pr`.

PR #161306, opened 2023-11-30
`cmd/lib/check` replace `system_command!` with `system_command`

bevanjkay:
Required for #161302 and #161140

A `system_command` error was being passed to `Plist.parse_xml()` causing the workflow to error out.

### DIFF
--- a/cmd/lib/check.rb
+++ b/cmd/lib/check.rb
@@ -29,11 +29,10 @@ module Check
       format_launchjob = lambda { |file|
         name = file.basename(".plist").to_s
 
-        xml, = system_command! "plutil", args:         ["-convert", "xml1", "-o", "-", "--", file],
-                                         print_stderr: false,
-                                         sudo:         true
+        result = system_command "plutil", args: ["-convert", "xml1", "-o", "-", "--", file], sudo: true
+        return name unless result.success?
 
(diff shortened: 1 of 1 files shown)


cmd/lib/check.rb
```diff
@@ -29,7 +29,9 @@ module Check
       format_launchjob = lambda { |file|
         name = file.basename(".plist").to_s
 
-        xml, = system_command! "plutil", args: ["-convert", "xml1", "-o", "-", "--", file], sudo: true
+        xml, = system_command! "plutil", args:         ["-convert", "xml1", "-o", "-", "--", file],
+                                         print_stderr: false,
+                                         sudo:         true
 
         label = Plist.parse_xml(xml)["Label"]
```
Bo98, line 36:
```suggestion
        result = system_command "plutil", args: ["-convert", "xml1", "-o", "-", "--", file], sudo: true
        return name unless result.success?

        label = result.plist["Label"]
```

Bo98 (commented):
`system_command` is already splitting stdout from stderr. The issue you are having is that you are telling `system_command` to error on a non-zero exit code by using `system_command!` with the exclamation mark

bevanjkay:
Thank you @Bo98

Bo98 (approved):

PR #161371, opened 2023-11-30
Wiso steuer 2024

piijai:

### DIFF
--- a/Casks/w/wiso-steuer-2024.rb
+++ b/Casks/w/wiso-steuer-2024.rb
@@ -15,7 +15,8 @@ cask "wiso-steuer-2024" do
   auto_updates true
   depends_on macos: ">= :catalina"
 
-  app "WISO Steuer 2024.app"
+  # Renamed for consistency: app name differs in Finder to shell
+  app "SteuerMac 2024.app", target: "WISO Steuer 2024.app"
 
   zap trash: [
     "~/Library/Application Support/BuhlData.com/WISOsteuerMac2024",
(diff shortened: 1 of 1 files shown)


bevanjkay:
@krehel This discussion has come up before, and I believe it is common for people to require access to the apps for previous years.

krehel:
@bevanjkay - fair enough!  Wanted to pose the question.

bevanjkay:
How many years we should keep is probably a good question though.
@piijai Do you have any thoughts here?

piijai:
Hi,
your previous point @bevanjkay is correct.

Like it says in the description, WISO Steuer is a software to create your tax declaration. Maybe we could also add, that it is only for the tax declaration for Germany, because it only supports the German tax declaration. 
And yes, every version only supports the tax declaration for the year before. So the current WISO Steuer 2024 only supports the tax declaration for the fiscal year 2023. If you want to create your tax declaration for the fiscal year 2022 you can't do that with WISO Steuer 2024, you have to use WISO Steuer 2023. 

In Germany you can declare your taxes up to 4 years back. So this year, you can still declare your taxes for the year 2019.
But yes, I assume only a few people will use the older versions like 2 years back, or more. As far as I know Buhl, the software company, supports the older versions with security and bug fixes for those 4 years.

Currently the audit is failing, because the livecheck retrieves an empty version string. I looked in the dmg file and the Info.plist file has the same structure as WISO Steuer 2023. So I have no idea, why the livecheck retrieves an empty version string.

I hope with those insights help you, how to handle the casks for WISO Steuer in the future.

bevanjkay:
@piijai The `livecheck` was failing because the `app` is named differently in shell to finder, so the app was not being found.

miccal (approved):

PR #161415, opened 2023-12-01
Update qq.rb

sitdownkevin:
Upgrade to v6.9.23.18968


sitdownkevin:
livecheck fail to fetch the latest version. The official website has not updated the latest version.

PR #161466, opened 2023-12-01
hapigo 2.11.1,230 (new cask)

zzsqwq:
HapiGo is a application launcher and productivity software

### DIFF
--- a/Casks/h/hapigo.rb
+++ b/Casks/h/hapigo.rb
@@ -1,0 +1,30 @@
+cask "hapigo" do
+  version "2.11.1,230"
+  sha256 :no_check
+
+  url "https://dl.hapigo.com/HapiGo.dmg"
+  name "HapiGo"
+  desc "Application launcher and productivity software"
+  homepage "https://www.hapigo.com/"
+
+  livecheck do
+    url :url
+    strategy :extract_plist
+  end
+
+  auto_updates true
+  depends_on macos: ">= :mojave"
+
+  app "HapiGo.app"
+
+  zap trash: [
+    "~/Library/Application Support/Hapigo",
+    "~/Library/Caches/Hapigo",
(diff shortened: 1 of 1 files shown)


bevanjkay:
We can use this for `livecheck` so that we can avoid `extract_plist`, but it doesn't look like there's versioned download links available. Will check again on the next update.

```ruby
  livecheck do
    url "https://hapigo.com/update/cast.plist"
    regex(%r{<key>version</key>\s*\n\s*<string>(\d+(?:\.\d+)+)</string>}i)
  end
```

zzsqwq:
> We can use this for `livecheck` so that we can avoid `extract_plist`, but it doesn't look like there's versioned download links available. Will check again on the next update.
> 
> ```ruby
>   livecheck do
>     url "https://hapigo.com/update/cast.plist"
>     regex(%r{<key>version</key>\s*\n\s*<string>(\d+(?:\.\d+)+)</string>}i)
>   end
> ```

Now I have switched to using feed and regex for livecheck. Could you please help me check again? Also, since I am not the developer of this software, I currently cannot change the content in the feed. However, I will try to communicate with them later.

By the way, may I ask how you found the links related to updates? I did not find them in Info.plist.

bevanjkay:
I used an application called Proxyman to intercept the requests the app makes when it checks for updates.

zzsqwq:
> I used an application called Proxyman to intercept the requests the app makes when it checks for updates.

CleanShot 2023-12-02 at 16 33 57@2x

Oh, right. I forgot that I can use it to capture the request for checking updates. Thank you!

bevanjkay:
When there is an update next it will be worth seeing if it points to a versioned download or just the same static 'latest' version.

miccal:
Thank you @zzsqwq.

miccal (approved):

PR #161484, opened 2023-12-02
ci: fix manual auditing

bevanjkay:
Fixes a couple of issues with manual auditing.

`generate-matrix:` Was not passing through the `labels` parameter because of a bad `if` statement.
`ci:` Lines up the `workflow_dispatch` inputs with the arguments in the `generate_matrix` call.


Rylan12 (approved):
Makes sense to me!

PR #161549, opened 2023-12-03
Update the download URL and versioning scheme of command-x

knu:

### DIFF
--- a/Casks/c/command-x.rb
+++ b/Casks/c/command-x.rb
@@ -26,10 +26,7 @@ cask "command-x" do
   ]
 
   caveats <<~EOS
-    This package is the non-App Store Version and claimed to be updated only once a year.
-
-    Use the App Store version whenever possible:
-
-      https://apps.apple.com/app/id6448461551
+    This variant of #{token} is only updated annually or when security vulnerabilities are discovered.
+    For a more frequently updated versio, install via the Mac App Store.
   EOS
 end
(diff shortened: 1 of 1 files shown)


krehel:
I have mixed feelings about this.  If we're making a software available that's only updated once per year and the other alternative is the MAS with frequent updates, I'd argue there's a potential quality of life issue with using the Homebrew provided version.

I can understand why people might not want to use MAS, but I would wonder if its known the software being installed does not receive updates.

I would almost be in favor of not providing this software through Homebrew and instead recommending MAS.

knu:
It's a need rather than a want.  MacBooks provided by your company may not have access to the App Store due to employment regulations and/or forced by device management software like jamf.  The way I see it, the author stating it is only updated once a year does not necessarily mean there'd be no timely update if a serious (security) problem were found.

krehel:
Sure, I get that!  100%.  I've been subject to the same conditions from employers.  And yes, it does not say an update wouldn't be made if a security issue exists.

I think my comment is some notes and thoughts on the matter, but also a loose agreement with @bevanjkay on the value of a `caveat` or comment.  Others may come across these comments in the future as well for understanding.

I'm thinking, very broadly, if I had to personally choose between the same software that is updated once per year or regular updates that bring fixes, enhancements, etc. I would pick the most up-to-date version.  I don't speak for everyone of course.  But I think there's value in knowing the "once per year" update so everyone is better informed for their own needs.

bevanjkay:
@knu 
A caveat indicating that the cask version is only updated annually or when security flaws are found would be ideal. And a note that the MAS version is updated more frequently.

knu:
I see, I've added a caveats message.  Does it look OK?

bevanjkay (approved):

bevanjkay:
Thank you @knu

PR #161724, opened 2023-12-05
Replace undocumented flag `--new-cask` in PR template

muzimuzhi:
Flag `--new-cask` is not listed in manpage for `brew audit`, and from implementation it's the same as `--new` and `--new-formula`.

For cask tap, `--cask --new` is better.

See
- https://docs.brew.sh/Manpage#audit-options-formulacask-
- https://github.com/Homebrew/brew/blob/4.1.23/Library/Homebrew/dev-cmd/audit.rb


reitermarkus:
> Not sure this change is strictly needed.

I actually think we should deprecate the `--new-formula`/`--new-cask` options, given that they do the same thing as just `--new` by itself.

I don't think we had `--formula`/`--cask` yet when these were added.

krehel:
Fair take @reitermarkus, if there's no difference.  I saw it as having additional aliases which I viewed as no harm.  But simplicity also good 👍

reitermarkus (approved):

PR #161963, opened 2023-12-07
gitkraken-cli: fix. Don't delete .gitkraken folder

sergiolms:
Since the `.gitkraken` folder is shared with the GitKraken Client (Desktop), I'm removing this line so the folder doesn't get deleted after uninstalling `gitkraken-cli` because a user could have both products installed.

### DIFF
--- a/Casks/g/gitkraken-cli.rb
+++ b/Casks/g/gitkraken-cli.rb
@@ -11,6 +11,4 @@ cask "gitkraken-cli" do
   homepage "https://github.com/gitkraken/gk-cli"
 
   binary "gk"
-
-  zap trash: "~/.gitkraken"
 end


bevanjkay:
@sergiolms The `zap` stanza is optional during uninstallation and only applies if the `--zap` flag is passed. Based on the documentation for `--zap` and the `brew help uninstall` output, it is expected that using the `--zap` flag may remove files that are shared between applications.

```
brew help uninstall
Usage: brew uninstall, remove, rm [options] installed_formula|installed_cask
[...]

Uninstall a formula or cask.

  -f, --force                      Delete all installed versions of formula.
                                   Uninstall even if cask is not installed,
                                   overwrite existing files and ignore errors
                                   when removing files.
      --zap                        Remove all files associated with a cask.
                                   May remove files which are shared between
                                   applications.
```

sergiolms:
Oh okay, so if you run `brew uninstall` without the `--zap` flag that folder shouldn't be removed. :+1: Gotcha.

Then I guess we're good, I'm closing this PR. Thanks for the clarification!

PR #162049, opened 2023-12-08
add TruHu (new cask)

0xdevalias:

- [x] `brew audit --cask --new <cask>` worked successfully.

### DIFF
--- a/Casks/t/truhu.rb
+++ b/Casks/t/truhu.rb
@@ -4,46 +4,20 @@ cask "truhu" do
 
   url "https://truhu-mobile-builds.s3.us-west-2.amazonaws.com/TruHu+Desktop+#{version}.dmg"
   name "TruHu"
-  desc "Uses a Desktop and Mobile app together to measure your display’s colors and proceed with calibration"
+  desc "Display calibration utility"
   homepage "https://truhu.app/"
 
-  # NOTE: This method isn't stable since the build hash will change
-  # livecheck do
-  #   url "https://truhu.app/static/js/main.9202b0c2.js"
-  #   regex(/\{no:(\d+),date:"[^"]*"/i)
-  # end
-
-  # NOTE: This gets an error while trying to run it:
-  #   Error: truhu: private method `open' called for URI:Module
-  # Is there a better way to check/use content from 2 different URLs in a livecheck?
   livecheck do
     url "https://truhu.app/downloads/index.html"
-    regex(%r{/static/js/main\.[a-f0-9]+\.js})
-
+    regex(/\{no:(\d+)\s*,date/i)
     strategy :page_match do |page, regex|
-      # Find the script URL with the hash
-      script_path = page.match(regex)
-      next if script_path.blank?
+      js_file = page[%r{src=["']?/(static/js/main\.[a-f0-9]+\.js)}, 1]
+      next if js_file.blank?
 
-      # Form the complete URL for the script
-      script_full_url = "https://truhu.app#{script_path}"
-
-      # Fetch the script content
-      script_content = URI.open(script_full_url).read
-
-      # Extract version and date
-      script_content.scan(/\{no:(\d+),date:"([^"]+)"/i).map do |match|
-        version, date = match
-        # You can format the version and date as needed here
-        "#{version},#{date}"
-      end
+      version_page = Homebrew::Livecheck::Strategy.page_content("https://truhu.app/#{js_file}")
+      version_page[:content].scan(regex).flatten
     end
   end
-
-  # NOTE: This is a last resort
-  # livecheck do
-  #   skip "No version information available"
-  # end
 
   app "TruHu Mac Desktop.app"
 
--- a/Casks/t/truhu.rb
+++ b/Casks/t/truhu.rb
@@ -6,7 +6,7 @@ cask "truhu" do
       verified: "truhu-mobile-builds.s3.us-west-2.amazonaws.com/"
   name "TruHu"
   desc "Display calibration utility"
-  homepage "https://truhu.app"
+  homepage "https://truhu.app/"
 
   livecheck do
     url :homepage
--- a/Casks/t/truhu.rb
+++ b/Casks/t/truhu.rb
@@ -22,7 +22,7 @@ cask "truhu" do
 
   depends_on macos: ">= :catalina"
 
-  app "TruHu Mac Desktop.app"
+  app "TruHu Mac Desktop.app", target: "TruHu.app"
 
   zap trash: [
     "~/Library/Application Support/TruHuDesktop",
--- a/Casks/t/truhu.rb
+++ b/Casks/t/truhu.rb
@@ -1,0 +1,57 @@
+cask "truhu" do
+  version "289"
+  sha256 "a5c1de3ddbf3d001fa88ac745e075b894964898388db1da7bb854c922cec792f"
+
+  url "https://truhu-mobile-builds.s3.us-west-2.amazonaws.com/TruHu+Desktop+#{version}.dmg"
+  name "TruHu"
+  desc "Uses a Desktop and Mobile app together to measure your display’s colors and proceed with calibration"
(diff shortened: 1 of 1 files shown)


Casks/t/truhu.rb
```diff
@@ -0,0 +1,57 @@
...
+        # You can format the version and date as needed here
+        "#{version},#{date}"
+      end
+    end
+  end
+
+  # NOTE: This is a last resort
+  # livecheck do
+  #   skip "No version information available"
+  # end
```
0xdevalias, line 46: Not sure of the best way to implement the `livecheck` here; as it seems the only way I can find of figuring out the version number is by extracting it from the bundled code, which uses a hash in it's filename.

The first attempt using the hardcoded URL works, but is brittle.

The second attempt using a dynamic way to get the correct URL+hash doesn't work due to the following error:

```
⇒ brew livecheck --cask truhu --debug
/usr/local/Homebrew/Library/Homebrew/brew.rb (Cask::CaskLoader::FromDefaultTapPathLoader): loading truhu

Cask:             truhu
Livecheckable?:   Yes

URL:              https://truhu.app/downloads/index.html
Strategy:         PageMatch
Regex:            /\/static\/js\/main\.[a-f0-9]+\.js/
Error: truhu: private method `open' called for URI:Module
Warning: Removed Sorbet lines from backtrace!
Rerun with --verbose to see the original backtrace
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Casks/t/truhu.rb:32:in `block (3 levels) in load'
/usr/local/Homebrew/Library/Homebrew/livecheck/strategy/page_match.rb:57:in `versions_from_content'
/usr/local/Homebrew/Library/Homebrew/livecheck/strategy/page_match.rb:109:in `find_versions'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:752:in `block in latest_version'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:684:in `each'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:684:in `each_with_index'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:684:in `latest_version'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:280:in `block in run_checks'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:220:in `map'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:220:in `with_index'
/usr/local/Homebrew/Library/Homebrew/livecheck/livecheck.rb:220:in `run_checks'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/livecheck.rb:133:in `livecheck'
/usr/local/Homebrew/Library/Homebrew/brew.rb:86:in `<main>'
```

I couldn't find a pattern for extracting from 2 URLs on the livecheck docs:

- https://docs.brew.sh/Brew-Livecheck

Casks/t/truhu.rb
```diff
@@ -0,0 +1,31 @@
...
+  sha256 "a5c1de3ddbf3d001fa88ac745e075b894964898388db1da7bb854c922cec792f"
+
+  url "https://truhu-mobile-builds.s3.us-west-2.amazonaws.com/TruHu+Desktop+#{version}.dmg"
+  name "TruHu"
+  desc "Display calibration utility"
+  homepage "https://truhu.app/"
+
+  livecheck do
+    url "https://truhu.app/downloads/index.html"
+    regex(/\{no:(\d+)\s*,date/i)
```
bevanjkay, line 12: I am not 100% happy with this regex here, we may need to monitor this to make sure it remains stable.

0xdevalias: > But the version number does seem to be returned in the function matched by this regex.

This is reading it from the section of the website bundle that constructs the download link used by the 'Download Desktop Mac' button. While the main bundle code is minified, they also ship the sourcemaps, so we can easier see what the code is meant to be doing in that area.

`pages/DownloadBeta/index.tsx`:

image

While that code file/variables in it refer to 'beta', I believe that may just be leftover from old code patterns, as we can see that the `hasBetaAccess` check has been commented out, and the companion app is on the main Apple appstore, which didn't require me to be part of any beta programs.

The `{ no:` part of the `betaVersions` array seems to stay the same even through the minification, so I would suspect that this should stay stable between builds.

---

> There is an in-app update mechanism using a Firebase API

@bevanjkay Interesting. How did you find that one?

I used this script, but it didn't turn up anything:

```
$(brew --repository homebrew/cask)/developer/bin/find-appcast '/path/to/application.app'
```

bevanjkay: > @bevanjkay Interesting. How did you find that one?

I used an application called "Proxyman" to intercept the requests being made by the app.

0xdevalias: @bevanjkay Ah, true. That makes sense :)

I thought you might have had another cool 'static analysis' tool I could add to my workflow.

bevanjkay:
@0xdevalias We shouldn't be changing the app name unless there is an inconsistency with how it appears in shell, or a couple of other exceptions that I do not believe apply here.

0xdevalias:
> We shouldn't be changing the app name unless there is an inconsistency with how it appears in shell, or a couple of other exceptions that I do not believe apply here.

@bevanjkay Ah ok, fair enough. I figured that since 'Mac Desktop' isn't really relevant to the name of it that it would make sense to remove it; in a similar way that those sorts of things are removed from the cask name/etc.

I can revert it back to how it was if that's a homebrew requirement; but given the app is just called "TruHu", this seemed to make more sense when showing up in my installed apps/etc.

samford (commented):
I pushed a commit with some `livecheck` block changes:

* Using `url :url` with the `PageMatch` strategy will cause livecheck to download the dmg file instead of some page on their website (i.e., `PageMatch` isn't one of the strategies that generates a new URL from `:stable` in formulae or `:url` in casks). The `main` JS file looks to be the same on all pages, so `url :homepage` will work fine (I assume that's what was intended here).
* I tweaked the regex to allow for a little more flexibility (e.g., where `no:` appears in the object, allowing for whitespace if the file doesn't end up being minified, etc.) while still only matching the value that we want. It's still not great but there's only so much we can do with what upstream provides and this works for now.

bevanjkay:
Thanks @samford - I definitely intended to use `:homepage` thanks 🤦🏼‍♂️

bevanjkay:
> I can revert it back to how it was if that's a homebrew requirement; but given the app is just called "TruHu", this seemed to make more sense when showing up in my installed apps/etc.

@0xdevalias We only really change the app name with target if there is a functional reason to do so. In this case we should keep it consistent with how upstream provide the files.

0xdevalias:
> I pushed a commit with some `livecheck` block changes

@samford Looks good, thanks :)

> We only really change the app name with target if there is a functional reason to do so. In this case we should keep it consistent with how upstream provide the files.

@bevanjkay Noted, shall revert the change.

bevanjkay (dismissed):

0xdevalias:
@bevanjkay RE: https://github.com/Homebrew/homebrew-cask/pull/162049/commits/1a6e3ed27de0e219f253d851b2dab256279c3b1f

```
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Casks/t/truhu.rb:9:12: C: [Correctable] Cask/HomepageUrlTrailingSlash: 'https://truhu.app/' must have a slash after the domain.
  homepage "https://truhu.app/"
           ^^^^^^^^^^^^^^^^^^^

4347 files inspected, 1 offense detected, 1 offense autocorrectable
Error: Cask/HomepageUrlTrailingSlash: 'https://truhu.app/' must have a slash after the domain.
```

bevanjkay (approved):

0xdevalias:
Thanks for your assistance in getting this fixed up and merged @bevanjkay / @samford 🎉

PR #162174, opened 2023-12-11
heimdall-suite: update `uninstall`

miccal:


bevanjkay (approved):

bevanjkay:
For reference, this is because `pkg` is always run as `root` with `-target /`

PR #162176, opened 2023-12-11
insta360-studio: update `uninstall`

miccal:


bevanjkay:
For reference, this is because `pkg` is always run as `root` with `-target /`

bevanjkay (approved):

PR #162382, opened 2023-12-13
Update PULL_REQUEST_TEMPLATE.md

0xdevalias:

- [ ] `brew audit --cask --new <cask>` worked successfully.

---

Every time I open a new Cask PR, the way I read this template (before the changes in this PR) makes me think I need to run both:

- `brew audit --cask --online <cask>`
- `brew audit --cask --new <cask>`

But `--new` implies both `--strict` and `--online`.

This was my attempt at re-arranging the checklist to be a bit more clear.

I wasn't sure if install/uninstall checks should be included in the 'always' section, or if it's only relevant for new casks. If my assumption was wrong, that can easily be put back.

---

This is part of my attempts to remove ambiguity/make things simpler for new contributors, alongside changes such as the following:

- https://github.com/Homebrew/brew/pull/16327

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,8 +6,6 @@ After making any changes to a cask, existing or new, verify:
 
 - [ ] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
 - [ ] `brew style --fix <cask>` reports no offenses.
-- [ ] `brew install --cask <cask>` worked successfully.
-- [ ] `brew uninstall --cask <cask>` worked successfully.
 
 **If updating an existing cask**:
 
@@ -19,3 +17,5 @@ After making any changes to a cask, existing or new, verify:
 - [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
 - [ ] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
 - [ ] `brew audit --cask --new <cask>` worked successfully.
+- [ ] `brew install --cask <cask>` worked successfully.
+- [ ] `brew uninstall --cask <cask>` worked successfully.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,7 @@ After making any changes to a cask, existing or new, verify:
 
 **If updating an existing cask**:
 
-- [ ] `brew audit --cask --strict --online <cask>` is error-free.
+- [ ] `brew audit --cask --online <cask>` is error-free.
 
 **If adding a new cask**:
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,14 +5,17 @@ _In the following questions `<cask>` is the token of the cask you're submitting.
 After making any changes to a cask, existing or new, verify:
 
 - [ ] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
-- [ ] `brew audit --cask --online <cask>` is error-free.
 - [ ] `brew style --fix <cask>` reports no offenses.
+- [ ] `brew install --cask <cask>` worked successfully.
+- [ ] `brew uninstall --cask <cask>` worked successfully.
 
-Additionally, **if adding a new cask**:
+**If updating an existing cask**:
+
+- [ ] `brew audit --cask --strict --online <cask>` is error-free.
+
+**If adding a new cask**:
 
 - [ ] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
 - [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask/search?q=is%3Aclosed&type=Issues).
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,14 +5,17 @@ _In the following questions `<cask>` is the token of the cask you're submitting.
...
 - [ ] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
-- [ ] `brew audit --cask --online <cask>` is error-free.
 - [ ] `brew style --fix <cask>` reports no offenses.
+- [ ] `brew install --cask <cask>` worked successfully.
+- [ ] `brew uninstall --cask <cask>` worked successfully.
+
+**If updating an existing cask**:
 
-Additionally, **if adding a new cask**:
+- [ ] `brew audit --cask --strict --online <cask>` is error-free.
```
bevanjkay, line 14:
```suggestion
- [ ] `brew audit --cask --online <cask>` is error-free.
```

bevanjkay:
@0xdevalias I think the previous workflow is actually more correct. We don't actually suggest running `brew install` and `brew uninstall` on every update, as generally this doesn't
 change, and on the occasion it does it will be picked up by the CI run anyway.

> - [] brew audit --cask --new <cask> worked successfully.

I think the additional point is okay, as it just runs a few additional audits that are only applied to new casks.

It is under a heading that reads "Additionally, **if adding a new cask:**"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,14 +5,17 @@ _In the following questions `<cask>` is the token of the cask you're submitting.
 After making any changes to a cask, existing or new, verify:
 
 - [ ] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
-- [ ] `brew audit --cask --online <cask>` is error-free.
 - [ ] `brew style --fix <cask>` reports no offenses.
+- [ ] `brew install --cask <cask>` worked successfully.
+- [ ] `brew uninstall --cask <cask>` worked successfully.
```
bevanjkay, line 10: Let's move these back to the bottom.

0xdevalias:
> It is under a heading that reads "Additionally, **if adding a new cask:**"

The annoyance is more that if I am adding a new cask; I don't want/need to run the audit twice; but the way this currently laid out; it suggests I need to run the 'always' audit, and then additionally also run the 'new' audit.

.github/PULL_REQUEST_TEMPLATE.md
bevanjkay, line 14: Replace this with the suggestion, and then move this audit line to the bottom checklist

```suggestion
- [ ] `brew audit --cask --online <cask>` is error-free.
```

bevanjkay (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
0xdevalias: 👌🏻 https://github.com/Homebrew/homebrew-cask/pull/162382/commits/561968749263de32369f9c398be1c1352b991631

.github/PULL_REQUEST_TEMPLATE.md
0xdevalias: @bevanjkay I'm confused what you're asking for here? 'move to the bottom checklist' would seem to imply you're saying this?

> **If adding a new cask**:
> 
> ..snip..
> - [ ] `brew audit --cask --online <cask>` is error-free.
> - [ ] `brew audit --cask --new <cask>` worked successfully.

bevanjkay: Sorry for the confusion, using GitHub's suggestion tool can be a bit tricky sometimes.

Essentially;

```
**If updating an existing cask**:

**If adding a new cask***;
...
- [ ] `brew audit --cask --new <cask>` is error-free.
...
```

0xdevalias: Right, that makes sense.

But why not `--strict` when updating? As per:

> And for the case of updating a cask, based on the description of `--strict`, there seems to be no reason not to include that flag; the worse case would be that it catches issues before the CI does, which seems like a positive.

bevanjkay: CI does not run the audits with `--strict` for updates, as the `--strict` audits only apply to new casks.

0xdevalias: Fair enough. Then yeah, can remove it again I guess.

> The terminology differs sometimes between homebrew-core and homebrew-cask due to the codebase being somewhat separate in homebrew/brew

Yeah.. that's another super frustrating point of friction. Having one say a thing is required, and the other not; and then wondering whether that's because it isn't actually required, or one or the other isn't up to date, etc.

Curious if there's any harm/extra effort in the strict checks being done on updates as well? Any reason they couldn't just be enabled on CI/everywhere and then maybe even simplify by deprecating `--strict` in general?

bevanjkay: I'm not sure of the total effect but there are certain checks that only matter when a cask is new (such as a token conflict). 

But the discussion for that would need to be a separate issue in Homebrew/brew or a discussion in Homebrew/brew - as it is a separate issue.

0xdevalias: Makes sense. Will try and remember to look into it and/or open a discussion there.

0xdevalias: Accepted your change, it already seemed to be in the right place as per https://github.com/Homebrew/homebrew-cask/pull/162382#discussion_r1425278554

p-linnane:
Closing since this user has been permanently banned from Homebrew.

